PR #27, opened 2016-04-26
Adding show command (useful for containers with same name) to see ins…

remyd1:
Hi,

This pull request allows a user to use the "show" command which is simply looking in the container directory to display the content of summary, installed, build_..., (...) files.

For instance, it can be useful if you have many containers with the same name and you want to display specific informations about a container from UUID.

Best regards,

Rémy

### DIFF
--- a/libexec/cli/show.exec
+++ b/libexec/cli/show.exec
@@ -1,0 +1,94 @@
+#!/bin/bash
+#
+# Copyright (c) 2015-2016, Gregory M. Kurtzer. All rights reserved.
+#
+# “Singularity” Copyright (c) 2016, The Regents of the University of California,
+# through Lawrence Berkeley National Laboratory (subject to receipt of any
+# required approvals from the U.S. Dept. of Energy).  All rights reserved.
+#
+# If you have questions about your rights to use or distribute this software,
+# please contact Berkeley Lab's Innovation & Partnerships Office at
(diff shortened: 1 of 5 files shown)


libexec/cli/show.exec
```diff
@@ -0,0 +1,94 @@
...
+export SAPPNAME SINGULARITY_COMMAND CONTAINERPATH
+shift
+
+message 2 "Showing '$SAPPNAME'\n"
+
+RETVAL=1
+
+NAME="undefined"
+SUMMARY="undefined"
+SIZE="0"
```
remyd1, line 69: I forget to add
NB_FILES=0
here...

gmkurtzer:
I'll check this out as soon as I get to a real computer. Thanks!

Sent from my iPhone

gmkurtzer:
I really like this feature, thanks!

remyd1:
You are welcome. It is useful to me too. I did not find any checksum on the
whole container. So I also thought that a hash/checksum (e.g. md5 or
sha1...) on the whole container could be useful. It would be display in the
'show' command then.
If you plan to make a registry for the sapp applications this feature would
also be important.

Best,
Remy.
Le 26 avr. 2016 20:52, "Gregory M. Kurtzer" notifications@github.com a
écrit :

PR #40, opened 2016-05-31
fixing typos (maybe grammar too)

kellyrowland:
please leave open for a bit; i'll poke around the code/docs/etc. and see what else i can find.

### DIFF
--- a/libexec/cli/copy.summary
+++ b/libexec/cli/copy.summary
@@ -1,1 +1,1 @@
-Copy files from your host into to the container
+Copy files from your host into the container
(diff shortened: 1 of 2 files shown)


kellyrowland:
The above two commits were all that I could find in the help text. I have one other nitpicky comment that you should totally feel free to ignore:

```
krowland@WARP-Drive:~/singularity$ singularity -v copy
increasing verbosity level (2)
USAGE: singularity (options) copy [Singularity container image] [source file(s)] [dest]
```

"Destination" is abbreviated here, though it doesn't seem like there are any abbreviations in any of the other command messages.

gmkurtzer:
Merged, thank you!

PR #52, opened 2016-06-04
Add gray shadow for <p> tag and stick the menu bar on top

yfxia:


gmkurtzer:
I took a look on http://yfxia.github.io/singularity/ and for some reason the content wasn't loading. I'm just on my iPhone right now but it works on the main page. Can you check that?

The navbar was working nicely though!

Sent from my iPhone

gmkurtzer:
I think the problem is a missing curly bracket on the load content function.

yfxia:
Thanks Greg! I found it!

PR #60, opened 2016-06-07
various fixes

loveshack:
I hope these are obvious enough.
I think  the strerrors are worthwhile, as I've had to debug failures previously.
The Debian package won't work without the fixperms change (mea culpa).


gmkurtzer:
Fantastic updates, thank you! I rely on spell check for documentation, but spell checking source code is almost impossible. hah

loveshack:
You wrote: 

> Fantastic updates, thank you! I rely on spell check for documentation,
> but spell checking source code is almost impossible. hah

There's M-x flyspell-prog-mode and M-x ispell-comments-and-strings
if Emacs isn't illegal in Berkeley.

gmkurtzer:
Haha, Emac's isn't illegal but granted I've always been a vi(m) user! As usual, thanks @loveshack!!

PR #69, opened 2016-06-13
bootstrap/create changes

loveshack:
Sorry, these should probably have been separated.
The linux_build_redhat.smod change avoids "no package" errors when target != host distribution and possible issues from plugins (observed with subscription-manager and etckeeper on RHEL).


gmkurtzer:
Looks great, thanks!

PR #72, opened 2016-06-16
Ensure directory exists before trying to write to a file in it

loveshack:
This resolves a "no such file or directory" for /etc//mtab when bootstrapping rpms.
Maybe the extra / should be dropped, i.e. "$SINGULARITY_BUILD_ROOT$i" etc.


gmkurtzer:
Thanks!

PR #76, opened 2016-06-21
Doc improvement for mount

loveshack:
I realized I was under a misapprehension about mounts from the reply to a previous issue.

### DIFF
--- a/libexec/cli/mount.help
+++ b/libexec/cli/mount.help
@@ -1,9 +1,10 @@
 USAGE: singularity (options) mount [container path]
 
-Use this command to implement a standard loop mount on the container
-image.
+Start a shell with a standard loop mount on the container image.  The
+mount is in the container namespace, i.e. not visible outside that
+shell.
 
-note: This command must be executed as root.
+Note: This command must be executed as root.
 
 EXAMPLES:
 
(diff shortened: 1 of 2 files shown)


loveshack:
I'm confused about pull requests (amongst other things!).  I was hoping this was a separate one, providing a temporary fix for #77

gmkurtzer:
Did commenting out `dnf` fix the read only issue you saw in #77?

loveshack:
You wrote: 

> Did commenting out `dnf` fix the read only issue you saw in #77?

Yes.

loveshack:
This is in relation to #77, but very messy and probably should be started anew.
It needs the change for --tolerant, which dnf doesn't support.

I still don't understand what's happening in 2.0.

gmkurtzer:
I will merge this PR, and take a look at it in more detail from work. 

What about 2.0 is troublesome? The bootstrapping?

loveshack:
You wrote: 

> I will merge this PR, and take a look at it in more detail from work. 
> 
> What about 2.0 is troublesome? The bootstrapping?

Yes.  I haven't yet found what change in master made the dnf change work
with el6.

PR #87, opened 2016-06-23
Fix RunCmd for multi-word args

loveshack:
I tried something like "RunCmd ln -sf something /singularity" and it failed.  This makes that work, but may not be the best solution due to quoting.  Maybe either stipulate a single arg or pipe quoted components of "$@" into sh if it must use sh.  I wonder if it's better to require an exec-able command for the chroot.


gmkurtzer:
Good catch, I'll play with this later today!

PR #92, opened 2016-06-24
Fix megabyte/mebibyte confusion in help

loveshack:
Also fix expand default size.

### DIFF
--- a/libexec/cli/create.exec
+++ b/libexec/cli/create.exec
@@ -109,7 +109,7 @@ if ! touch "$IMAGE_FILE" 2>/dev/null; then
     exit 1
 fi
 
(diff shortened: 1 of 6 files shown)


gmkurtzer:
Ooops,.... pedantics! Thanks!

loveshack:
You wrote: 

> Ooops,.... pedantics! Thanks!

Well, it's a 7% difference for the default image size.

remyd1:
Hum... Interesting, that could explain a bit the size problem conversion
from docker to singularity...

loveshack:
DERNAT R�my notifications@github.com writes:

> Hum... Interesting, that could explain a bit the size problem conversion
> from docker to singularity...

I don't think so, but it would help, if anything.

PR #97, opened 2016-06-28
Change example to use a '#' prompt, since create requires you to be root

jmstover:


gmkurtzer:
Can you recreate this to be "$ sudo " instead of "# " which is how the rest of the documentation reflects?

Thank you!

jmstover:
No problem. I'll throw out another PR after that.

PR #100, opened 2016-06-28
Fix build without setns

loveshack:
You better check this is right, but it builds on el6.

### DIFF
--- a/libexec/docker-import.sh
+++ b/libexec/docker-import.sh
@@ -172,9 +172,6 @@ if ! cmd=$(docker inspect --format='{{json .Config.Cmd}}' "$dock"); then
     exit 1
 fi
 
-entry='["/bin/sh","-c"]'
-cmd='["df","-h"]'
-
 if [[ $cmd != null ]]; then
     cmd=$(IFS='[],'; echo $cmd)
     cmd=${cmd:1}            # no leading blank
--- a/libexec/docker-import.sh
+++ b/libexec/docker-import.sh
@@ -1,0 +1,211 @@
+#!/bin/bash
+
+# Converting docker images to singularity images.
+
+# Copyright 2016  Dave Love, University of Liverpool
+# Licence: BSD-3-Clause
+
+# The basic technique is:
+#   * Create a docker container from the image
+#   * Figure out the singularity container size from that
+#   * Create the singularity container
+#   * docker export | singularity import
+#   * Clean up
+# It's partly obscured by error checking and recovery in verbose
+# if blocks in Singularity style, and determining the size and possible
+# /singularity contents is complex.
+
+# Fixme: Check docker "Os" is "linux" [can it be anything else?]
+#          Are there any guarantees of contents, like coreutils?
+#        Maybe accept docker container instead of image?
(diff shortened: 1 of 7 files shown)


gmkurtzer:
Heya Dave, this looks good. About my only question I have so far is with the debootstrap min base variant. Inconsistently during my testing with different versions I noticed that debootstrap would sometimes omit apt so I added that in the include list. Have you seen similar?

loveshack:
git(hub) is confusing me no end.  I didn't mean to push the debootstrap change as it had developed a conflict, and I've just resolved that while trying to make another request for docker import that looks as if it's now mixed in.

I'm surprised if apt isn't implied. With which version of Debian? It's definitely "essential" in jessie, and requires 'Yes, do as I say!' to remove, but maybe that depends on debootstrap -- I haven't checked how it works.  There could be a case for not including it if you didn't want to pull more packages, of course, and it's a pity there isn't something like yum's --installroot as far as I know.

loveshack:
I was going to comment on the import change.  Grr.

It implements the suggestion to have import support different inputs, but has some rough edges, like wanting format-specific args.  I don't know whether it's the best approach but don't have strong feelings.
I'm still puzzled by the required size of the singularity container relative to the tar size or filesystem usage and will try to figure that out.

gmkurtzer:
If I recall correctly, it was specific to a particular debootstrap version that would omit apt if I specified minbase. I don't think it hurts to specify it as an include just to be sure. 

Yeah, some of the fork/pull/remote features of github I'm still struggling with as well. My background is almost all with single master repos like CVS and SVN. 

As far as the size goes, there will just need to be a fudge factor. Also considering the file system overhead plus you don't want it to be exact, here should always be a bit of wiggle room after the import is done. 

Do you want me to accept this pull request now, or wait for you to adjust a few more things?

Oh one last bit. I added a documentation page on contributing code and as it pertains to asserting copyright. I agree with your choice here but the only thing my legals told me to always make sure is that the projects copyright is also included in a contributed file.

Thanks!

loveshack:
I've pushed a version with the header.
I won't do any more on the import soon.  Feel free to grab it if you think it's OK-ish.

libexec/docker-import.sh
```diff
@@ -0,0 +1,227 @@
...
+# scratch file, but perhaps the file would be better.
+# Note that the Docker semantics for the entrypoint and cmd aren't too
+# clear <https://docs.docker.com/engine/reference/builder/>.
+
+if ! cmd=$(docker inspect --format='{{json .Config.Cmd}}' "$dock"); then
+    message ERROR "docker inspect failed\n"
+    exit 1
+fi
+
+entry='["/bin/sh","-c"]'
```
gmkurtzer: Why do you statically define the entry point and command here?

loveshack: [Sorry this didn't get sent.  Yesterday must have been a while ago!]

You wrote: 

> > +entry='["/bin/sh","-c"]'
> 
> Why do you statically define the entry point and command here?

Sorry, I checked in a version with that for testing.  I pushed a change
yesterday, but didn't get round to the pull request -- though I don't
understand how to refer to specific changes.

PR #106, opened 2016-06-29
Minor changes to bootstrap.html

nl384:
Just some minor errors I've encountered on the bootstrap documentation page. Both relate to the list of singularity keywords under the example bootstrap file. One is that InstallFile is incorrectly listed as InstallFiles. Second is a missing bullet point for RunScript.


gmkurtzer:
Thank you!

PR #110, opened 2016-07-05
Fix default expansion size in images.html

nl384:
From what the example text seems to say, as well as what singularity tells me when I expand an image, it seems the default is actually 512 MiB, not 1024 MiB.

Also despite the facts that I know MiB is more accurate than MB and that the newer version of singularity outputs the message with MiB not MB, it seems strange to use "MiB" when the example above uses MB; but I'll leave that choice up to you.

### DIFF
--- a/content/images.html
+++ b/content/images.html
@@ -99,7 +99,7 @@ $ ls -l container.img
 
 <p>
 Similar to the create sub-command, you can override the default size (which is
-1024MiB) by using the --size option.
+512MiB) by using the --size option.
 </p>
 
 <h2>Copying, sharing, branching, and distributing your image</h2>


gmkurtzer:
Hiya, thanks for the correction!

What example is still using MB?

nl384:
Well the example on that page (images.html) regarding expanding images is:

```
$ sudo singularity expand container.img 
Expanding sparse image by 512MB...
Checking image (/sbin/mkfs.ext4)
e2fsck 1.42.9 (28-Dec-2013)
Growing file system
resize2fs 1.42.9 (28-Dec-2013)
Done. Image can be found at: container.img
$ ls -l container.img 
-rwxr-xr-x. 1 root root 1610612769 Jun  1 08:40 container.img
```

In fact all of the examples on this page (images.html), as well as the rest of the web page besides the location I just changed, uses MB. I think this makes sense as Singularity 2.0 uses MB. So my question was whether or not it made sense to also have the part that I just correct say MB as well, in order to match the rest of the examples and the rest of the text (even though the master branch, and probably Singularity 2.1 uses MiB).

gmkurtzer:
Ahh, I see what you mean now! That was a copy/paste from the command output before it was patched to use MiB. Yeah, we can probably just change the units in the pasted code. Good catch!

PR #111, opened 2016-07-05
Changing a few MB/GB => MiB/GiB in images.html

nl384:
Just changing a few more MB/GB to MiB/GiB for better uniformity on the page :)


gmkurtzer:
Thanks again!

PR #116, opened 2016-07-06
Update rhel_ubuntu recipe

nl384:
Added a tips section (which I will hopefully add to as I come up with more tips!); made the example more robust (installs TensorFlow now); also did some general spelling and grammar fixes.


nl384:
I just saw that you fixed InstallPkgs to take more than one package in a line, I'll add that as another tip next time!

gmkurtzer:
Thanks!

Issue #117, opened 2016-07-06
Contentious legalese in web Contributing section


loveshack:
After eventually reading the Contributing section on the web site, I'm afraid I've stopped contributing.

It seems to require accepting an unusual legal theory about "patches" and derivative works (in US terms?) generally -- i.e. that contributors don't have any copyright in changes even if they are copyright-significant according to, say, the FSF's legal advice.  Without accepting that, the only way to make what would normally be considered copyrightable contributions seems to be to disclaim or assign copyright.  If I wanted to do so I'd need agreement from my employer to the extent it has a copyright interest.  (I don't understand the term "assert copyright", since copyright is automatic, even in the US now.)

I think the text at least needs to make clear what its status is as legal advice/opinion since it appears to disagree with what I've seen and been told.  Also this information should be distributed with the software, as it implies that the "separate written license agreement" mentioned in the licence could be void, depending on LBL's  determination of something as a "patch".

The section also suggests that the licence is the standard "BSD3", but it's actually a distinct one called "LBNL BSD" in Fedora https://fedoraproject.org/wiki/Licensing/LBNLBSD or "BSD-3-Clause-LBNL" in SPDX, as used by Debian https://spdx.org/licenses/BSD-3-Clause-LBNL.html#licenseText.  (I missed that initially, as did the automated licence check.)  It is a potential trap for people publishing changes who may not realize about the all-permissive default licence, especially if the licence is described as simply BSD3.

rhc54:
I’m honestly rather surprised by all the angst expressed by a couple of folks on the list regarding the license. There is nothing specifically onerous about it, and I’ve seen this language in the combination of license + contributor agreement rather frequently. The split between what appears in the license itself vs the contributor agreement varies, but the overall content is the same.

You cannot “own” copyright in something you contribute to a 3-clause BSD project. Once you contribute it, the project must remain free to distribute it without asking your permission. Likewise, anyone who uses it must be free to do so without fear of subsequent demands for royalties.

The “LBNL” clause to which you refer is also pretty standard, though mostly seen in the contributor agreement. Basically, it simply states that you cannot hold LBNL liable if you contribute a code, and then later decide you shouldn’t have done so. I see this in most projects to which I contribute.

You should certainly have the approval of your employer before contributing to any open source project, regardless of license or agreement. Employers are pretty fussy about that, in my experience.

loveshack:
rhc54 notifications@github.com writes:

[...]

PR #121, opened 2016-07-08
ENH: various tune ups to debian packaging

yarikoptic:
Description still needs some tune up to make it easy to understand
purpose of this package within 1 paragraph.

policy is newer now

since never entered Debian proper yet, release should be UNRELEASED

BSD-2-clause license header should be present verbatim as well


gmkurtzer:
Any changes to the license or copyrights have to go through Berkeley Lab legals. While I don't think that would cause a problem as I think the first two clauses of the lbl license are verbatim, I already have some pending requests with them (and can add this to the list). But just so I can make the case... Why do you say the BSD 2 clause license needs to be appended?

yarikoptic:
In this PR I am not proposing any license change. I just removed trailing whitespace in its text and provided verbatim of bsd-2 license Dave listed as the one for Debian packaging - it is required by the Debian copyright file format referenced on top of that file

gmkurtzer:
The white space fixes of course aren't a problem but the copyright files (along with the source file headers) were given to me explicitly from my legals. I think the addition of the BSD2 license should be fine (but I may have to remove it after confirming with legals). 

Before I accept this PR can you reiterate why this is needed? You mentioned that it is because Dave listed it as the license for the Debian packaging? Is that fixable (rather then adjusting licensing wording)?

Thanks!

yarikoptic:
I guess you are asking solely about why I have added full text of BSD-2 license?  
Then it is just because it is needed to fulfill the debian/copyright format, similarly to as how it already includes verbatim your other-BSD license for the main codebase.  Dave chose to release his debian packaging under BSD-2 -- a compatible but different from your other-BSD -- license, which should be alright, correct? and I just listed it (a sample BSD-2 license text I took from other project) verbatim.  If you want Dave to change license to be the same as your other-BSD -- it is all alright with me but it is not for me to decide ;) @loveshack ?

yarikoptic:
alternative, if you really want to distribute here only the code under your other-BSD license, could be is to create a clone with a dedicated 'debian' branch (advisable generally anyways) which will be the one containing `debian/` directory with content under a different license.

gmkurtzer:
Is the motivation to also use the BSD2 license for the packaging because the Singularity license is incompatible (or not approved) by the Debian packaging standards? Sorry if the question is naive, just trying to understand why the BSD2 license was added. 

Thanks!

yarikoptic:
it is not naive at all question @gmkurtzer  ;-)  the main body of the license seems to be a typical BSD-3 license, totally ok with debian free software guidelines. Additional clauses following the disclaimer sounds like providing copyleft-style additional guarantees for Lawrence Berkeley National Laboratory... probably that is just fine... on a quick search I see similar statement available in another package within Debian:  http://sources.debian.net/src/zsi/2.1~a1-4/debian/copyright/?hl=237#L237 but without "contact the office" statement.  But that one should also be alright whenever accompanying the full license text which does state permitted uses already (unlike in the headers where no mentioning of license/permitted uses is done).   Well -- we could also ask on debian-legal mailing list on what ppl think about it, but I think it should not be of a problem.  So my only concern is the headers without pointing to the license/allowed uses (ie what I have whined about in #123)

yarikoptic:
and I can only speculate that Dave chose  that more permissive BSD-2 license because "why not be more permissive"? ;)

gmkurtzer:
Ahh, I understand your point now and great to see another LBNL license already in Debian, nice find! There have been a few discussions recently regarding the Singularity license, so I already have a meeting with legals scheduled for today. I will add your proposals/requests to the agenda and let you know.

Thanks again!

loveshack:
You wrote: 

> I guess you are asking solely about why I have added full text of
> BSD-2 license?
> Then it is just because it is needed to fulfill the debian/copyright
> format, similarly to as how it already includes verbatim your
> other-BSD license for the main codebase.

Yes, I'd realized that was missing, and published an update earlier.
That uses the SPDX licence name and amends COPYING to try to nullify the
default contribution clause, pending a reply from Fedora legal.

Satisfying Debian rules specifically could only be relevant to Debian
maintainers, though.  The packaging was just a quick go at allowing
people to build a dpkg, though I'm happy if it's more useful.  Is there
some intention to include it in Debian?

> Dave chose to release his
> debian packaging under BSD-2 -- a compatible but different from your
> other-BSD -- license, which should be alright, correct? and I just
> listed it (a sample BSD-2 license text I took from other project)
> verbatim.

I thought it was meant to refer to SPDX, in case it's actually
different.  I took the text from there.

> If you want Dave to change license to be the same as your
> other-BSD -- it is all alright with me but it is not for me to decide
> ;) @loveshack ?

The packaging may or may not pass a copyrightability test in the US or
UK, but LBNL claim that's irrelevant.  That might be a concern for a
potential Debian maintainer, especially given my experience of Debian
rules.  (It probably means I have to orphan the Fedora package.)  See #117.

loveshack:
You wrote: 

> it is not naive at all question @gmkurtzer ;-) the main body of the
> license seems to be a typical BSD-3 license, totally ok with debian
> free software guidelines. Additional clauses following the disclaimer
> sounds like providing copyleft-style additional guarantees

It doesn't sound so to me, but I don't think that matters.

> for Lawrence Berkeley National Laboratory...

The default contribution clause, which makes it a distinct licence
according to Fedora and LBNL, doesn't seem to treat LBNL specially in my
reading.

> probably that is just
> fine... on a quick search I see similar statement available in another
> package within Debian:
> http://sources.debian.net/src/zsi/2.1~a1-4/debian/copyright/?hl=237#L237
> but without "contact the office" statement.  But that one should also
> be alright whenever accompanying the full license text which does
> state permitted uses already (unlike in the headers where no
> mentioning of license/permitted uses is done).  Well -- we could also
> ask on debian-legal mailing list on what ppl think about it, but I
> think it should not be of a problem.

Debian may not always agree, but the licence is free according to
Fedora, as I've said elsewhere.  I couldn't find any previous discussion
in Fedora or Debian, though I've asked within Fedora, and would be
interested in any references.

loveshack:
You wrote: 

> and I can only speculate that Dave chose that more permissive BSD-2
> license because "why not be more permissive"? ;)

Specifically, because why should I worry about the promotion clause?  I
don't know what Debian advice is apart from including a licence, but
Fedora packaging is recommended to be treated as the "MIT" default.
That's similar, and it seemed appropriate just to drop that clause from
the overall licence.

[I'm only referring to Fedora because I'm a maintainer due to having to
run RHEL-like systems.  I assume Debian and Fedora broadly agree, and
I'm typing this at Debian.]

yarikoptic:
cool cool ;)  SPDX/"update earlier" -- I have 'git grep'ed the sources and do  not see SPDX.  So the update to headers of the files is coming too, right?
May be so we are all in clear  I will just (after any adjustments are made) run the license by debian-legal mailing list (they aren't lawyers but rather enthusiasts who know the matter ;) )
Yeah -- Fedora and Debian should have similar if not identical treatment of what is free -- here is Debian's: https://www.debian.org/social_contract#guidelines , which in turn were later adopted for OSI definition

gmkurtzer:
Btw, I was just granted permission to fix the source file headers as you
proposed (point to license rather hen IPO@lbl. I will implement the fix
tomorrow.

Thanks!

On Monday, July 11, 2016, Yaroslav Halchenko notifications@github.com
wrote:

> cool cool ;) SPDX/"update earlier" -- I have 'git grep'ed the sources and
> do not see SPDX. So the update to headers of the files is coming too, right?
> May be so we are all in clear I will just (after any adjustments are made)
> run the license by debian-legal mailing list (they aren't lawyers but
> rather enthusiasts who know the matter ;) )
> Yeah -- Fedora and Debian should have similar if not identical treatment
> of what is free -- here is Debian's:
> https://www.debian.org/social_contract#guidelines , which in turn were
> later adopted for OSI definition
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> https://github.com/gmkurtzer/singularity/pull/121#issuecomment-231737103,
> or mute the thread
> https://github.com/notifications/unsubscribe/ANVkxYaufYnsMfhNVtRpwEiue5We_TQLks5qUkfygaJpZM4JHsed
> .

## 

Gregory M. Kurtzer
High Performance Computing Services (HPCS)
University of California
Lawrence Berkeley National Laboratory
One Cyclotron Road, Berkeley, CA 94720

PR #122, opened 2016-07-08
Make test.sh executable + rudimentary .travis.yml to run it

yarikoptic:
To take advantage of it, just login to travis-ci.org, and enable testing for your repository.  I did for my clone and you can see the result here:
https://travis-ci.org/yarikoptic/singularity/builds/143229601


gmkurtzer:
I have not played with Travis before. I just had someone else integrate one of my other projects with CircleCI. I haven't done the research yet but I would guess they are both similar systems? Do you have the ability to push back products from Travis into a repo like nightly builds or binary objects or containers?

Thanks!

yarikoptic:
Yes, they are largely similar. Yes, you can setup various push backs etc, but this or is just to automate running the test script you have, which would then be executed for every pr so you could get some assurance that nothing gets broken
Enabling travis is easy, but let me know if you need further directions

gmkurtzer:
Excellent, thank you!

PR #127, opened 2016-07-08
Update rhel_ubuntu.html

nl384:
Sorry for so many pull requests! Was just looking at this and saw a couple of mistakes to fix.


gmkurtzer:
No need for an apology, updates and fixes are good!

PR #130, opened 2016-07-12
Changed error message format

aduffy19:
Noticed that one error message was still in fprintf() and return(1) format. Changed it to use message() and ABORT(255) for consistency.


gmkurtzer:
Oh, great catch! Thanks!

PR #133, opened 2016-07-13
Redhat module change

jmstover:
Change InstallPkgs to look for a conf file in the build root and use that if it exists.

**\* I would just take the diff and apply instead of the mess that this change request is from merging with upstream and having a conflict on that file .... for some reason. ***


gmkurtzer:
Squashed and merged just the diff, thanks Jason!

Issue #134, opened 2016-07-13
consider making wrapper scripts fail "early" and upon any "undefined variable" via adding `set -eu`


yarikoptic:
singularity ships a number of shell scripts, but unfortunately none of them is "paranoid enough" to enforce that they exit as soon as error occurs (in a pipe or a compound command) or some undefined variable is used.  Not sure which are invoked/used while running root-suided sexec, but I think it could be a good "stability" and possibly "security" measure if behavior of those scripts would become more "deterministic" in some sense, by forcing the assumptions that all errored executions are failures (unless specifically handled) and that all variables must be defined after some point in their execution life.

gmkurtzer:
I'm not sure I exactly follow... Is this something you could help with or give me some specific examples to help me see exactly what you are thinking of?

Thanks!

yarikoptic:
Oh sure... will do later today
Cheers

gmkurtzer:
I think this is pretty much done at this point. I have set -u in functions, and set -e in bootstrap'ing (and fixed the bugs associated with doing that). Do you feel there is more I should do here, or am I clear to close this issue?

Thanks!

yarikoptic:
since indeed set-ings get inherited and now present in main functions, I guess set -u would do its role in other scripts.  But you think it is not worthwhile of similar `-e`  "protection" all throughout the scripts?

gmkurtzer:
Implementing -e across the board had a lot of unintended consequences, for example I am currently trying to catch errors and print a decent notification. With -e set, it would just stop because the controlling shell would exit before it could print a notification/error/warning.

I figured though that some things like the bootstrapping should indeed quit on error. Thoughts?

yarikoptic:
could you point me to specific lines?  it is just a matter I guess again of explicit statements/assessment of result of execution of the line in question, e.g.

``` shell
$> bash /tmp/1.sh   
/bin/ls: cannot access 'alskjdf': No such file or directory
ls failed. too bad

$> cat /tmp/1.sh
#!/bin/bash

set -e
set -u

if ! /bin/ls alskjdf; then
   echo "ls failed. too bad"
   exit 1
fi

echo "the end"
```

gmkurtzer:
I found that the test.sh was just stop at random places when -e was set, and it made debugging very difficult. If you want to see for yourself add 'set -e' right above where it sources the functions, and then run test.sh. Here is my output:

```
 + singularity mount -h                                         (retval=0) OK
[...]

PR #135, opened 2016-07-13
ENH: set -e and set -u in sample bash scripts under libexec

yarikoptic:
Just for a "demo" of #134 

Whenever a variable wants to be consumed even if not defined `${VAR:-}` could be used to explicitly assume empty value if not defined under the `set -u` mode.

Benefits of set -u -- code becomes more robust without continuing running if some variable was not checked to be not empty explicitly.   of set -e: script will exit right away if any error happens, instead of continuing plowing through possibly causing a disaster ;-)  

Possible side-effects:  if PS\* env variables use some variables loosely allowing them to be assumed empty if not defined, and script relies on using the corresponding PS (e.g. PS4 in set -x mode), could lead to failures if `set -u`

### DIFF
--- a/libexec/bootstrap.sh
+++ b/libexec/bootstrap.sh
@@ -19,12 +19,15 @@
 # 
 # 
 
+set -e
 
 ## Basic sanity
 if [ -z "$SINGULARITY_libexecdir" ]; then
     echo "Could not identify the Singularity libexecdir."
     exit 1
 fi
+
+set -u
 
 ## Load functions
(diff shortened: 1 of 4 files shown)


gmkurtzer:
Ahh, that is very cool!

Considering functions gets sourced by all other files, would it suffice to put `set -u -e` at the top of that single file alone? (and then fix all of the undeclared variable tests that will occur?)

yarikoptic:
since alternative approach was taken

PR #137, opened 2016-07-14
BF: clarify "modified BSD" license in about.html

### LINKED ISSUE #117: Contentious legalese in web Contributing section

yarikoptic:
"modified BSD" name is commonly used to refer to BSD-3 license
see eg https://en.wikipedia.org/wiki/BSD_licenses#3-clause_license_.28.22Revised_BSD_License.22.2C_.22New_BSD_License.22.2C_or_.22Modified_BSD_License.22.29
which is not the entirety of the license terms here, since BSD-3 doesn't have any additional statements about terms of how derived work (Enhancements) must be licensed.

So I took my shot at clarifying the wording.  IMHO #117 should be reopened and closed only whenever some adjustment to wording of that paragraph online was done

Closes #117

### DIFF
--- a/content/about.html
+++ b/content/about.html
@@ -73,8 +73,11 @@ reproducibility.
 
 <h4 class="loud">License</h4>
 <p>
-Singularity is released under a modified BSD license. The only modification
+Singularity is released under a BSD license with 2 modifications. One of the modifications
 being that you may not use the names of any project members or contributors
 (namely UC or the US Department of Energy) to endorse or promote products or
-services derived from or using this software.
+services derived from or using this software. And the other is that if you
+share modified code publicly (or with Berkeley National Laboratory), you will
+grant similar license terms on your "Enhancements"
+(see <a href="https://github.com/gmkurtzer/singularity/blob/master/LICENSE">LICENSE</a> file for more details).
 </p>


gmkurtzer:
This wording works for me, but I may change it to be exactly inline with what legals sent me WRT to the source file header copyright lines ("customized 3-clause BSD license").

As I followed, #117 was referring to different text in part of the contributing page which was actually removed before the ticket was closed. I had text in there about when I personally felt it was appropriate to add (assert) a copyright line (new features and functionality), and when not to (not original work). I'm not sure if it was my vision of what deserves copyright that was contentious or my layman's wording of it, but in either case it is removed.

Thanks!

yarikoptic:
Thank you @gmkurtzer for clarification. just for completeness -- you are referring to the commit   bba0fff93b7f27f2d5cd6bda08cefdcdadcf4a7c, and now I see the picture better ;-)  Related to this I guess is the Debian ITP (intent to package) discussion I have CCed you own https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=828970 
Overall I still hope that we would just all stay friends (@loveshack included ;-)) and manage to avoid the clutter of patching laying around ;-)

PR #140, opened 2016-07-14
ENH: use flawfinder in test.sh, install flawfinder in travis env

yarikoptic:
locally: fails to run test.sh before flawfinder run 

```
Building test container...
 + sudo singularity create -s 568 container.img                 (retval=0) OK
 + sudo singularity bootstrap container.img /home/yoh/deb/persp (retval=1) ERROR
Setting RELEASE=7
ERROR: Neither yum nor dnf in PATH!
Full output in: /tmp/tmp.RY842ZBOox

```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ compiler:
 before_install:
   # so that adjusted PATH propagates into sudo
   - sudo sed -i -e 's/^Defaults\tsecure_path.*$//' /etc/sudoers
+  - sudo apt-get install flawfinder
 
 script:
   - ./test.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,8 @@ before_install:
   # so that adjusted PATH propagates into sudo
   - sudo sed -i -e 's/^Defaults\tsecure_path.*$//' /etc/sudoers
   - sudo apt-get install flawfinder
+  # necessary ATM. See https://github.com/gmkurtzer/singularity/pull/140
+  - sudo apt-get install yum
 
 script:
   - ./test.sh
--- a/test.sh
+++ b/test.sh
@@ -118,6 +118,14 @@ stest 0 sudo rm -rf "$TEMPDIR"
 stest 0 make maintainer-clean
 
 /bin/echo
+if `which flawfinder > /dev/null``; then
+    /bin/echo "Testing source code with flawfinder"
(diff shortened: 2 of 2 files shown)


yarikoptic:
travis build shows the same error :  https://travis-ci.org/yarikoptic/singularity/builds/144812709

gmkurtzer:
Maybe also do a "apt-get install yum"? I just checked the busybox.net site again, still seems to be down. I just email'ed Erik and Rob to see if it will come back soon. As soon as it does, I will fix the link back to busybox.

yarikoptic:
oh... I was still waiting for travis to run it ;) not sure if it all works... ok, will fix later if it doesn't ;-)  whenever travis gets enabled, you will receive notifications right here in the pull requests so you would know if tests ran/pass/failed

yarikoptic:
still didn't make it to the end:

```
 + sudo singularity copy container.img singularity /            (retval=2) ERROR
/tmp/singularity-test.usj8gk/libexec/singularity/copy.sh: 40: /tmp/singularity-test.usj8gk/libexec/singularity/copy.sh: Bad substitution
Full output in: /tmp/tmp.70y6xkziq8
The command "./test.sh" exited with 1.
```

yarikoptic:
I think I saw such failures whenever that PS4 for set -x mode was too fancy for the shell... but I thought all that was fixed up, so might be smth else ;)

gmkurtzer:
I can't seem to replicate that error locally. Is there a more reliable/portable shell method of extracting arg ranges that you know of?

yarikoptic:
re replicate -- what is your `/bin/sh` pointing to?  if bash, try to change to `dash`

gmkurtzer:
Dash on RHEL/Centos?! LOL....

We should probably rewrite that then to not be so bash'y. How is your shell-fu? ;-)

yarikoptic:
When I do shell I do it in bash :-)
I do fix some bashisms when ppl complain though, but the foo is low ;-)

yarikoptic:
Can you run singularity within singularity? :-)

gmkurtzer:
Oh, well.. the interpreter is set to /bin/bash.. I wonder why it isn't working then. Is /bin/bash linked to /bin/dash? I would doubt that...

Yes, you can sort of run Singularity within a Singularity. You have to be root because Singularity doesn't allow SUID within a container.

PR #144, opened 2016-07-15
Fix backwards sense of conditional

jjohnson42:
The wrong DEBUG message was output, and unsetenv was called
when it wasn't set.  Also revise a log message to not say
'read/write only'.


gmkurtzer:
Looks good, thank you!

PR #146, opened 2016-07-21
Add support for slave mounts.

bbockelm:
Previously, the entire set of mounts within the container were
marked 'private': this means any changes in the host OS would not
show up in the container.

However, in places that use autofs (e.g., distributing software),
we want the autofs daemon (which lives in the host OS) to be able
to make mounts that show up in the container.  To do that, we must
mark the mounts as 'slave': changes in the host show up in the
container, but not vice-versa.

Unfortunately, it seems that it's all-or-nothing: we can't do slave
mounts _only_ for a single bind mount.  Accordingly, I set this to
be a global, container-wide option as opposed to only setting it for
a single bind mount at-a-time.  The default is to use private mounts
(the pre-existing behavior).

### DIFF
--- a/etc/singularity.conf
+++ b/etc/singularity.conf
@@ -28,6 +28,12 @@ mount sys = yes
 # attempt to mount it's base path into the container?
 mount home = yes
 
+# MOUNT SLAVE: [BOOL]
+# DEFAULT: no
+# Should we automatically propogate filesystem changes from the host?
+# This should be set to 'yes' when autofs mounts in the system should
+# show up in the container.
+mount slave = no
 
 # BIND PATH: [STRING]
 # DEFAULT: Undefined
--- a/src/sexec.c
+++ b/src/sexec.c
@@ -499,10 +499,11 @@ int main(int argc, char ** argv) {
                 ABORT(255);
             }
 
+            int slave = config_get_key_bool(config_fp, "mount slave", 0);
             // Privatize the mount namespaces
-            message(DEBUG, "Making mounts private\n");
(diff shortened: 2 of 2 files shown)


gmkurtzer:
Have you tested this with multiple containers running simultaneously? It has been a little while, but early on in the development I found that MS_PRIVATE was required to properly unshare the mount namespaces from the host (IIRC without it, the container mounts were visible and seen from the host).

bbockelm:
Well, there are three things to consider:
- `MS_PRIVATE`: makes the mount namespace inside of the container independent from the host.
- `MS_SHARED`: mount changes inside the container are visible outside _and vice versa_.
- `MS_SLAVE`: mount changes outside the container are visible inside.  Mount changes inside _are not_ visible outside.

The traditional Linux default is `MS_PRIVATE`.  Starting with RHEL7 (actually, it's done by systemd), the root filesystem was mounted with `MS_SHARED` by default.

So yes - if you completely drop the `MS_PRIVATE` flag, you completely mess up the host's mounts.  However, `MS_SLAVE` is closer to the semantics most people would want.

gmkurtzer:
Can we do both MS_SLAVE | MS_PRIVATE? I will check on that when I get out of my string of meetings today (10-5pm continuous meetings today... sheesh)!!!

bbockelm:
I'm not sure I understand the last comment.  It seems, by definition of the respective flags, SLAVE and PRIVATE are mutually exclusive.  No?

gmkurtzer:
Ahh, apologies. I think you are right, they are mutually exclusive. This is what happens when I try to do technical things in meetings... LOL

PR #153, opened 2016-07-21
Allow chroot to be based from a directory tree.

bbockelm:
This allows the user to point at a directory instead of a loopback-
device-based image.  This is _only_ enabled if `NO_NEW_PRIVS` is also
enabled in order to prevent setuid-binaries from being accessible
via the chroot image.

(Note: most of the diff is from whitespace changes.  The patch is easier to review with `diff -w`).

### DIFF
--- a/src/sexec.c
+++ b/src/sexec.c
@@ -89,8 +89,8 @@ int main(int argc, char ** argv) {
     char *command;
     char *sessiondir;
     char *sessiondir_prefix;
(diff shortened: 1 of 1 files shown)


gmkurtzer:
I want to spend some time going through this in more detail and it is hard for me to see the full context via this interface, but at first glance it looks good.

Thanks!

gmkurtzer:
This looks good.

Thanks!!

PR #157, opened 2016-07-22
Mark conf file as no-replace

bbockelm:
Files marked as `%config` have their changes overridden on upgrade by the package version.  It's traditional on RHEL platforms to mark package config files as `%config(noreplace)`, preventing nasty surprises during upgrades.

### DIFF
--- a/singularity.spec.in
+++ b/singularity.spec.in
@@ -38,7 +38,7 @@ rm -rf $RPM_BUILD_ROOT
 %defattr(-, root, root)
 %doc AUTHORS COPYING ChangeLog INSTALL NEWS README.md
 %attr(0755, root, root) %dir %{_sysconfdir}/singularity
-%attr(0644, root, root) %config %{_sysconfdir}/singularity/*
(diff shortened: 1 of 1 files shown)


gmkurtzer:
I'm not sure that is a good idea. The general behavior of RPM for configs is that they are replaced only iff they are unmodified. I am fairly certain that this modification will make it so that even when unmodified the config will never be upgraded.

bbockelm:
Nope - here's a handy guide:  http://www-uxsup.csx.cam.ac.uk/~jw35/docs/rpm_config.html

`%config(noreplace)` files are upgraded if the sysadmin hasn't changed defaults.  If the sysadmin does change them, the new file from the RPM is dropped with the `.rpmnew` suffix.

`%config` files are always upgraded, regardless of sysadmin changes.  When the sysadmin changes them, their changes are preserved with the `.rpmsave` suffix.

In general, you almost never see `%config` files: everything is `%config(noreplace)` because that's how most admins think it should work.

gmkurtzer:
Oh, I stand corrected. But I swear they changed this!! Lol

Thanks!

PR #158, opened 2016-07-22
Provide private scratch directories

bbockelm:
There are common cases (particularly, when no MPI is involved) where one would prefer to have each job receive a unique copy of `/tmp`.  This prevents various files from leaking after a job ends and simple protection against common mistakes in usage of world-writable directories.

This PR implements such a scheme.  It creates a scratch directory as the invoking user (allowing the user some control over the location - preferring one that the batch system may have created for the job, such as `$_CONDOR_SCRATCH_DIR`), then bind-mounts the scratch directory as writable over locations listed by the sysadmin in `singularity.conf`.

After the job ends, the scratch directory is cleaned up, carefully avoiding passing filesystem boundaries and not dereferencing symlinks.


gmkurtzer:
I think this is a fantastic idea. I had something similar but removed it when I implemented the bind path config, and meant to bring it back with a slightly different implementation. To that end I would recommend the following:
1. Make the tmp/scratch directories by default reside within the sessiondir so they are automatically deleted at the end of the job. This also has the benefit of concurrent per user container runs sharing the tmpdir. 
2. If there is utility in leveraging a resource managers allocated tmp dir we should make that a variable to override the default of putting it in the sessiondir. 
3. While it is fine to have a forced behavior in the config, I think it makes sense for this to also be runtime controlled so for example an MPI run can request shared tmp with host of allowed in the master config. 
4. Is there a default license in code examples posted to stackoverflow? Is it BSD and/or copyrighted? You should also add copyright on the new source file (both the projects copyright as well as your own). 

Thoughts?

Great thinking on this, thank you!

gmkurtzer:
I broke this PR, but please don't fix yet... Still doing more changes to the source code layout.

bbockelm:
1.  What's the impact (if any?) of moving this to the session directory (owned by root)?  This directory has to be owned by the unprivileged user.
2.  Yeah - there's definite utility in keeping this in the batch-system-allocated directory:  this gets automatically cleaned up at the end of the job (in case the container got killed with a `kill -9`).  Some batch systems get quite upset if there are non-user files in the temp directory (such as the session directory), so the session directory should not move.
3.  Noted.  Will address this in a future commit.
4.  I checked - it looks like the default Stack Overflow license is MIT.  I'll add license files shortly.

gmkurtzer:
1. Yes, very good point. The sessiondir will have to be owned by the user calling it. It has the unfortunate side effect of then making files binded from the sessiondir (e.g. the container's /etc/passwd|group) being owned by the user rather then root. Obviously not critical, but not perfect either. With that said, we might have to do this anyway to support Singularity with no setuid.
2. As long as the sessiondir is only created/deleted by the user, then there is no harm letting the user override the default location (e.g. environment variable).

I have a very simple s_rmdir which basically does an nftw() to unlink files in a directory. Is clean_dir necessary, or can it be merged to extend the existing s_rmdir in file.c?

Even though user configured scratch binding works only if no_new_privs is set, it still needs to be wrapped in a singularity.conf bool so the system administrators can define if users are allowed to use this feature.

Lastly, the scratch dir config option is confusing me but maybe we come back to that in a separate PR?

Feel free to start remerging this into the new master branch.

Thank you!

bbockelm:
For `s_rmdir`: this (a) follows symlinks and (b) will cross to different devices.

I think it's equivalent to what's in `clean_dir` if you are OK with adding `FTW_PHYS|FTW_MOUNT`.

gmkurtzer:
Oh, good catches! Yes, let's add those features so it is just as smart as clean_dir.

gmkurtzer:
Looks good to me, thanks!

PR #159, opened 2016-07-22
Add user namespace support / unprivileged mode.

bbockelm:
This commit adds support for unprivileged operation _if_ we are using
a chroot instead of a raw disk image (use of loopback devices are still
banned for raw disk images).  Remapping is done so the invoking user
'looks right' in the namespace, but all other file ownership looks strange.

Should work on modern Ubuntu and Fedora variants.  RHEL7 doesn't have
enough backported patches to make this function.  As it's tricky to
determine whether a given host has support for this mode, it must be
explicitly enabled at compile time.

### DIFF
--- a/libexec/cli/exec.exec
+++ b/libexec/cli/exec.exec
@@ -83,5 +83,10 @@ SINGULARITY_IMAGE="${1:-}"
 export SINGULARITY_IMAGE
 shift
 
+if [ -d "$SINGULARITY_IMAGE" ]; then
+  export SINGULARITY_USERNS=1
+  exec "$SINGULARITY_libexecdir/singularity/sexec_nosuid" "$@" <&0
+else
+  exec "$SINGULARITY_libexecdir/singularity/sexec" "$@" <&0
+fi
 
-exec "$SINGULARITY_libexecdir/singularity/sexec" "$@" <&0
--- a/libexec/cli/exec.exec
+++ b/libexec/cli/exec.exec
@@ -88,10 +88,11 @@ SINGULARITY_IMAGE="${1:-}"
 export SINGULARITY_IMAGE
 shift
 
-if [ -d "$SINGULARITY_IMAGE" ]; then
+if [[ -e "$SINGULARITY_libexecdir/singularity/sexec" && "x${SINGULARITY_FORCE_NOSUID-}" == "x" ]]; then
+  exec "$SINGULARITY_libexecdir/singularity/sexec" "$@" <&0
+else
+  unset SINGULARITY_FORCE_NOSUID
   export SINGULARITY_USERNS=1
   exec "$SINGULARITY_libexecdir/singularity/sexec_nosuid" "$@" <&0
-else
-  exec "$SINGULARITY_libexecdir/singularity/sexec" "$@" <&0
 fi
 
--- a/src/mounts.c
+++ b/src/mounts.c
@@ -111,9 +111,11 @@ void mount_bind(char * source, char * dest, int writable) {
         ABORT(255);
     }
 
-    //  NOTE: This used to be separated into two distinct mount calls (one to mount, a second to remount
-    //  as read-only).  I changed this to a single mount() call as user-namespaces aren't allowed to
-    //  remount.
+    //  NOTE: The kernel history is a bit ... murky ... as to whether MS_RDONLY can be set in the
+    //  same syscall as the bind.  It seems to no longer work on modern kernels - hence, we also
+    //  do it below.  *However*, if we are using user namespaces, we get an EPERM error on the
+    //  separate mount command below.  Hence, we keep the flag in the first call until the kernel
+    //  picture cleras up.
     message(DEBUG, "Calling mount(%s, %s, ...)\n", source, dest);
     if ( mount(source, dest, NULL, MS_BIND|MS_NOSUID|MS_REC|(writable <= 0 ? MS_RDONLY : 0), NULL) < 0 ) {
         message(ERROR, "Could not bind %s: %s\n", dest, strerror(errno));
--- a/src/mounts.c
+++ b/src/mounts.c
@@ -107,18 +107,13 @@ int mount_bind(char * source, char * dest, int writable) {
         ABORT(255);
     }
 
+    //  NOTE: This used to be separated into two distinct mount calls (one to mount, a second to remount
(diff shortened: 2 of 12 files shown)


gmkurtzer:
This is fantastic, thank you! I am right in the middle of refactoring a major chunk of the sources so it is easier to maintain and add features. Once I finish with that, I will look at this in depth and see how to integrate it with the refactored code (unless you will want to do that).

I have some questions on exactly what is happening with the user namespace but I suspect they will be answered as I go through your code in more depth.

yarikoptic:
`whether a given host has support for this mode` -- how could I figure out if particular debian/ubuntu release has support for it?

bbockelm:
I tried writing a simple autoconf test for determining kernel support: it's not sufficient to look for kernel versions (lots of backporting occurs) or presence of constants in the header.  I gave up - it was a really tough in the end.  One really needs to check if:
1.  You can create a user namespace.
2.  You can perform bind mounts within the namespace.
For example, RHEL7 succeeds at (1) and fails at (2).  I think you can piece together things from the information in `man 7 user_namespaces`.

As for hints on which Ubuntu / Debian: I'd guess anything running 3.14 or 3.16.  Unfortunately, I really don't have much background on that platform.

gmkurtzer:
I broke this PR, but please don't fix yet... Still doing more changes to the source code layout.

bbockelm:
Ok, I think this is ready to go now.

jjohnson42:
FYI on the loop limitations, one could imagine using guestmount as a means of supporting unprivileged use of 'raw' files (and qcow2 files for that matter).

libexec/cli/exec.exec
```diff
@@ -83,5 +83,10 @@ SINGULARITY_IMAGE="${1:-}"
 export SINGULARITY_IMAGE
 shift
 
+if [ -d "$SINGULARITY_IMAGE" ]; then
+  export SINGULARITY_USERNS=1
```
gmkurtzer, line 95: Does a use-case exist where someone would want to run in a directory and still use the standard sexec?

src/mounts.c
```diff
@@ -110,13 +111,18 @@ void mount_bind(char * source, char * dest, int writable) {
         ABORT(255);
     }
 
+    //  NOTE: This used to be separated into two distinct mount calls (one to mount, a second to remount
+    //  as read-only).  I changed this to a single mount() call as user-namespaces aren't allowed to
+    //  remount.
```
gmkurtzer: Did this work for you? It was my finding that remounting was the only want to make a bind read-only.

gmkurtzer: Ahhh, you re-exec the standard sexec if this fails... Ok, still reading along. lol

libexec/cli/exec.exec
gmkurtzer, line 95: Ahh, you are re-exec'ing the SUID version if necessary from within sexec... Considering this can not be the default code pathway, I'm hesitant to make it default even if the image is a directory...

gmkurtzer:
How does this sound to you.... Considering most people interested in --with-userns probably don't even want an setuid program on their computer, let's toggle the install-exec-hook in src/Makefile.am, such that sexec doesn't get the 4755 perm. Then we don't have to conditionalize how we are getting called and satisfies people interested in using it with no SUID.

Thoughts?

bbockelm:
That approach might lead to packaging problems.  I was thinking that, with this, we could split singularity into two packages, `singularity` and `singularity-setuid` (containing only `sexec`) on platforms that support user namespaces (this avoids breaking most existing production sites).

This appeals to me, as the presence of the setuid binary is a bit more explicit.

Then, the logic would look something like:

```
if [[ -e "$SINGULARITY_libexecdir/singularity/sexec" && "x$SINGULARITY_FORCE_NOSUID" == "x" ]];
  exec "$SINGULARITY_libexecdir/singularity/sexec" "$@" <&0
else
  unset SINGULARITY_FORCE_NOSUID
  exec "$SINGULARITY_libexecdir/singularity/sexec_nosuid" "$@" <&0
fi
```

This would use the setuid binary if it's available (as that's going to always have the most features) while still providing a way for debugging / testing unprivileged mode via setting an env var.

src/mounts.c
bbockelm: I've been looking around, and it seems this is fairly kernel version dependent - with newer kernels skewing toward requiring two separate calls.

gmkurtzer:
```
if [[ -e "$SINGULARITY_libexecdir/singularity/sexec" && "x$SINGULARITY_FORCE_NOSUID" == "x" ]];
```

My concern with this is that users can then enable/disable execution through the SUID binary... However that is mitigated if we package it separately. We may want to end up having a non_SUID sexec-wrapper that just reads from the system owned config file and does the right thing, and then confirm from the SUID version that it should be allowed to run (and yes, of course or leave it unpackaged).

I am still thinking about separating the features from sexec.c to a new source called sexec_nosuid.c. There will be a lot of overlap in code, so we should move as much as we can into external source files/functions. Over time as more and more features become standard into the kernel, we can shift emphasis to sexec_nosetuid.c.

This way we can keep the current sexec focusing on all of the features we currently need escalation for and keep it as clean as possible. In parallel we can develop the future unprivileged version and have it directly address your needs first.

Thoughts?

bbockelm:
Yeah - I thought about that too.  However:
1.  If the sysadmin installs `sexec` setuid, then that binary can always be invoked directly.  I view these scripts as helpful helper scripts!
2.  If there's kernel support for unprivileged mode, then the user can always download, compile `sexec_nosuid`, and execute from his or her home directory.

So, between the two, it seems reasonable to allow the user to select the mode.  The only improvement I could think of is allowing the sysadmin to put in a "setuid kill switch" into the config file that cases sexec to abort if it's disabled in the config file.  This seems to have nonzero-but-marginal value.

The problem I see with a separate `sexec_nosuid.c` is the behavior changes for nosuid aren't limited to `sexec.c` - in fact, there's a surprisingly small amount of code changes in `sexec.c` right now.  Most of it got factored out to the privilege handling code.

gmkurtzer:
Well, with the first option, we can easily add the check to the setuid version and just exit immediately out if it is configured by the system admin not to run it.

The reason I was thinking this is because there are still a lot of changes coming that may not be necessary for the setuid version. You maybe exactly right, and I just don't have the changes or requirements fully grock'ed in my head yet, so I'll keep chugging through your PRs as I find opportunities!

Thanks!

PR #162, opened 2016-07-25
Initial implementation of user-specified bind mounts.

bbockelm:
This adds user bind-mounts _if_ NO_NEW_PRIVS is available.  Currently,
this is not hooked up to the CLI (one must provide the correct env vars).

As the user likely will specify directories that don't exist in the
underlying container, this adds a very preliminary form of creation
of mount points inside the container.  It has a major drawback: it
clobbers the top-level directory that _does_ exist.  So, if /foo/bar
exists and the user requests that /foo/bar/baz is bind-mounted from
the host, then we will do a mount on /foo/bar, wiping out the contents
of /foo/bar.  This makes a pretty poor mechanism for adding files such
as /etc/resolv.conf (as it would have a side-effect of removing the
rest of /etc).

A more thorough approach would involve an overlayfs mount: however,
I'd like to keep this more limited version as at least an option as
it would be usable in unprivileged mode (overlayfs is not available
to unprivileged users!).

Note that the helper directories for the bind mount are kept in /tmp;
I plan to move them to the session directory in the future once
directory cleaning is fixed there.

### DIFF
--- a/src/mounts.c
+++ b/src/mounts.c
@@ -30,6 +30,7 @@
 #include <string.h>
 #include <fcntl.h>
 #include <libgen.h>
+#include <limits.h>
 
 #include "config.h"
 #include "mounts.h"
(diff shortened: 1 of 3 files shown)


bbockelm:
Ok, I think everything is cleaned up now.

bbockelm:
FYI - I think this one is ready to go, but it makes a lot more sense to merge #164 and #159 first.

Once those are in, I plan to update this to prefer overlayfs-based creation of the bind mounts (these are far less problematic) unless user namespaces are used.

gmkurtzer:
Will do! I'll go through this as soon as I can get a breath up for air.. Sorry for the delay!

BTW, I had a talk with the contributors working on the tmpfs overlay for creation of arbitrary bind points. That should come along soon, but we will have to conditionally figure out where to bind to existing paths. I wrote a function to find the basedir within a container for this (not sure right now if you used it or reimplemented), but in theory we have to be careful about what we allow (e.g. a user trying to bind ontop of /usr, will have bad results LOL). 

Thanks!!

PR #164, opened 2016-07-26
Added mount_overlay

aduffy19:
Functionality designed to allow modification of readonly container images.

### DIFF
--- a/src/file.c
+++ b/src/file.c
@@ -214,9 +214,10 @@ int s_rmdir(char *dir) {
 }
 
 int copy_file(char * source, char * dest) {
+    struct stat filestat;
     int c;
-    FILE * fd_s;
-    FILE * fd_d;
+    FILE * fp_s;
+    FILE * fp_d;
 
     message(DEBUG, "Called copy_file(%s, %s)\n", source, dest);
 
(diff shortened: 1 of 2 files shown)


gmkurtzer:
@aduffy19 Perfect timing! There are a few other features that I am working on which this will greatly help! One thing we need to figure out is how to identify if the host supports this feature (overlay is a new mount option). It would be great to have this as part of the configure check and then conditionally enable this if the host supports it.

Then... if not, we will have to do a bunch of other things differently in the code. For example, bind to base directories that have to be present like we are doing now (ewwww).

Thanks!

bbockelm:
With respect to the use of `overlayfs`: this won't work in unprivileged mode. :(

For unprivileged mode, we might have to keep around the (more problematic) code in my other PR.

gmkurtzer:
@bbockelm Hrmmmmm... (sorry for the tangent off of this particular PR, but it is somewhat related) I am starting to think that we should implement the unprivileged version with a new sexec_nosuid.c source that has very minimal functionality but works with no privilege escalation on new hosts.

gmkurtzer:
@aduffy19 I just added a configure check for the host's overlay FS support in master 1f95f0b. Also, please note that it is currently failing the build check (click Details on the Travis integration test) and please send me another PR that fixes those issues. Would be good to also include some `#ifdef SINGULARITY_OVERLAYFS` conditionals as they are also now supported.

Thanks!

bbockelm:
Note -

RHEL7 supports overlayfs, but it's very non-POSIX (especially compared to the version on Ubuntu).  I expect them to pull in more patches for RHEL7.3 to bring things closer to the Ubuntu kernel patches.

However, it's probably a good idea to keep the use of overlayfs very light for now.  We're thinking of just using this to bootstrap bind mounts, right?

gmkurtzer:
Yes, that is my current thinking, using it exclusively for bind mount points. Do you know if the base RHEL7 implementation supports dealing with creation of new directories properly?

bbockelm:
New directory creation is fine: deleting directories and chmod gets dicey.  There are also a few kernel Oops's that can trigger, but I've forgotten what the conditions are for those.

gmkurtzer:
Tickling a RH kernel ooops, isn't something I'd want to do normally, but hopefully they will fix it then! lol

jjohnson42:
As mentioned though, 'mkdir -p' level functionality is the main draw, so we can have mount points that weren't originally there.  removing anything is problematic and other such limitations, but shouldn't be the case they impact this scenario.

Also might be able to meaningfully impose NOSUID on what would otherwise be a bind mount (need to double check)

gmkurtzer:
There are also issues imposing read only limitations on a bind mount and it has to be done in a second step in mount.c. I wonder if that hack will work to enable NOSUID on a bind mount too.

aduffy19:
@gmkurtzer I've seen those errors before. Looks like I committed an old version. Will get things synced up and hopefully that will fix things. Sorry!

gmkurtzer:
Done. Please let me know how you both wish to be referenced in the AUTHORS file.

aduffy19:
@gmkurtzer The same format as the rest works for us.

Amanda Duffy aduffy@lenovo.com
Jarrod Johnson jjohnson2@lenovo.com

Issue #155, opened 2016-07-21
Refactor sexec


gmkurtzer:
Sexec is begging for a refactoring which will allow us to identify the primary code paths and functionize everything else.

gmkurtzer:
This is in progress... committed a bunch of updates to this front last night.

gmkurtzer:
I did a first pass at this, but will come back and do more. At present sexec forks twice so there is 3 running process at any given time:
1. Management process: responsible for init/setup, and cleanup
2. Mount and namespace process: do the mounts, create and hold open the namespaces
3. Exec process: execute a particular command within the container namespaces

I will be moving these to separate binaries that will do the following:
1. Primary driver: init/setup and cleanup
2. Image prepare: bind loop, etc.. (this will be compiled as either SUID or non-SUID)
3. Namespace and mount process: Do the mounts and create/hold namespaces (SUID or non-SUID)
4. Exec process: set NO_NEW_PRIVS and execute the particular command within the container namespace

This is all pending I can securely pass the information from one to process to another without risk of the SUID bits being manipulated. I have some ideas on that, but still needs more thought. I may end up moving this architecture to a Google doc for easier iteration and collaboration.

bbockelm:
There aren't a lot of portable options for securely passing information between two processes - I guess the best option is to use anonymous sockets with `SCM_CREDENTIALS`.

But then you have to worry about serialization / deserialization errors and careful checking of the remote side's credentials.  Seems like it'd be subtle logic.

I think a better starting point would be:
1.  Continue to decrease the size of `sexec.c`.  I bet it could reasonably be below 500 lines of code.  That would make code review even better.
   - It would be useful to refactor the exec and payload code into a separate function instead of inside a multi-hundred-line code block.
2.  Separate current compilation targets into into child executables (non-setuid) that are either run as root or as the target user.
3.  Remove all global variables and introduce an opaque state handle.  When passing info from one process to another, we'd need to gather all this information and serialize it.  Might as well start on that part now!

gmkurtzer:
[...]

PR #166, opened 2016-07-28
Fix compilation issue from recent merges.

bbockelm:


gmkurtzer:
I saw these and was going to handle after my morning meetings... Thanks for jumping on them!

bbockelm:
Sure, no problem.  I wanted to spend some time improving the user bind mounts over lunch :)

gmkurtzer:
Thanks again! If you get a chance can you take a quick look at #155 and let me know what you think?

PR #169, opened 2016-07-28
Add CLI support for user bind mounts.

bbockelm:
With this, the user bind mounts can be specified via the `-B` command line flag.

Also moves the user bind mount code _out_ of `sexec.c`.


gmkurtzer:
Thanks, merged!

PR #171, opened 2016-07-29
Have signals forward transparently to the payload process.

bbockelm:
Signals are now forwarded correctly from parent to child; if the
namespace daemon is killed, then the exec daemon will exit immediately.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -23,8 +23,8 @@ bin_PROGRAMS = sexec image-create image-expand image-mount image-bind sexec_nosu
 
 ftrace_SOURCES = ftrace.c file.c util.c message.c
 ftype_SOURCES = ftype.c util.c file.c message.c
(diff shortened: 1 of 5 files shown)


bbockelm:
NOTE NOTE NOTE: This moves the exec process to be PID 1!!!

Will this be OK?  It seems so, as the namespace daemon does all the cleanup.

Just to be safe, I made the exec daemon drop privileges permanently.  Not clear why it doesn't already do that.

gmkurtzer:
Hi @bbockelm, there is a conflict with this merge. Did you want to check that or should I merge it in?

Thanks!

bbockelm:
Whoops, I missed the merge conflict.  Will try to get that fixed this afternoon.

bbockelm:
Ok, all fixed up.

gmkurtzer:
Hrmm, for some reason the CI is now failing:

https://travis-ci.org/gmkurtzer/singularity/builds/149864399

Wanna have a quick look when you get a chance?

PR #181, opened 2016-08-03
Add "errors=remount-ro" option to mount()

### LINKED ISSUE #180: Potential kernel panic when running a container with corrupted ext filesystem
Hi Greg!

We realized a potential issue when running a container based on a corrupted ext filesystem. 

Ext filesystems can be created (or modified) to generate a kernel panic when they encounter an error. Form `ext4(5)`:

```
errors={continue|remount-ro|panic}
              Define the behavior when an error is encountered.  (Either
              ignore errors and just mark the filesystem erroneous and
              continue, or remount the filesystem read-only, or panic and
              halt the system.)  The default is set in the filesystem
              superblock, and can be changed using tune2fs(8).
```

So if somebody ever (intentionally or not) alters an existing container with `tune2fs -e panic` and corrupts the filesystem, there is a potential for a regular user to generate a kernel panic on the host. 

This could be easily fixed by forcingly mounting the filesystem with the `errors=remount-ro` option in [mounts.c](https://github.com/gmkurtzer/singularity/blob/master/src/mounts.c#L72)

kcgthb:
Avoid specialy crafted ext images to generate kernel panics. Fixes #180

### DIFF
--- a/src/mounts.c
+++ b/src/mounts.c
@@ -70,11 +70,11 @@ int mount_image(char * loop_device, char * mount_point, int writable) {
 
     if ( writable > 0 ) {
         message(DEBUG, "Trying to mount read/write as ext4 with discard option\n");
-        if ( mount(loop_device, mount_point, "ext4", MS_NOSUID, "discard") < 0 ) {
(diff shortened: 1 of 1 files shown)


gmkurtzer:
dang, now I wish I held off on the 2.1.1 release... Guess there will be a 2.1.2 release soon! lol

PR #182, opened 2016-08-04
Overlaymount

aduffy19:
Added setup and call to mount_overlay within mount_image, if not writable and overlay is enabled. Includes addition to singularity.conf file.

### DIFF
--- a/etc/singularity.conf
+++ b/etc/singularity.conf
@@ -91,6 +91,11 @@ config group = yes
 # show up in the container.
 mount slave = no
 
+# OVERLAY DIR: [STRING] 
+# DEFAULT: (disabled) 
+# Directory used to mount the portions for overlay.  If unspecified, overlay is not 
+# used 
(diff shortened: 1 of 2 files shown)


gmkurtzer:
Heya @aduffy19! This looks great but it isn't building via travis. Can you take a look at that?

Thank you!

aduffy19:
Ah, @jjohnson42 had done some separate setup to test this and we didn't copy it all over. Will update once he's available, as I'm not sure what he did.

gmkurtzer:
It looks like it is building properly now, but we are getting errors on unsupported systems (e.g. in the above travis output) that it is failing on unsupported systems.

How are you configuring it to work automatically? That perhaps needs to be conditional at runtime or better yet checked with ./configure.

Thanks!

gmkurtzer:
Looks good! I'm going to merge this and shortly start looking at how to integrate into the devel branch (unless of course you want to do that?). :)

PR #184, opened 2016-08-05
Next round of testing work

bbockelm:
First commit is just bumping up the debug level to try and figure out what's wrong with the current master; working on it...  don't merge immediately.


bbockelm:
Figured out the test failure!

It's due to the signal handling code changing when the PID namespace was enabled.  It definitely broke unprivileged mode.  However, the signal handling code got tested with the _old_ test cases but then merged _after_ the new tests were added.

Hence, the build break wasn't apparent until after the merge.

Yay for tests though!

PR #191, opened 2016-08-18
Adding commandline argument to change the homedirectory which is bound into the image

Nekel-Seyew:
I modified the singularity.in, and mounts.c to allow end users to change the home directory which is bound into the container. This way, if there are permission issues due to HPC environments, like condor, users can still use singularity.


gmkurtzer:
Wow, you took this feature right off of my TODO list in my head! lol

I have been working for a couple of weeks on the devel branch changing the code structure and I am just about ready to merge this into master. If you want I can merge this now, and then do the migration myself into the new code structure, or you can take a stab at it.

Let me know, and thanks!

Nekel-Seyew:
I wouldn't be able to get to it until monday of next week, so if it's faster for you to do it, then that's fine. Otherwise, if me migrating these changes monday works, I can do it then, if that's more convenient for you.

gmkurtzer:
I merged this into master, and I will work on it from there. This will give me the ability to easily leverage the work you have done as well as get your name in the git logs.

I'll let you know when I've finished merging to the devel tree for your review and testing.

Thank you!

gmkurtzer:
This has been integrated into the devel tree and can be used by the -H /path argument. I also added a check for proper directory ownership and works with --contain.

Please check out what I did and let me know if it is satisfactory. :)

Thank you!

PR #198, opened 2016-08-29
Allow PID namespaces to be used in HTCondor.

bbockelm:
When run inside HTCondor (particularly if HTCondor is started as non-root), we want to have the option of PID namespaces inside the HTCondor job.

### DIFF
--- a/etc/init
+++ b/etc/init
@@ -20,7 +20,6 @@ if [ -n "${OMPI_COMM_WORLD_SIZE:-}" ] ||    # ORTE (Open MPI)
    [ -n "${PBS_JOBID:-}" ] ||               # PBS/Torque
    [ -n "${LSB_JOBID:-}" ] ||               # LSF (and Open Lava?)
    [ -n "${COBALT_JOBID:-}" ] ||            # Cobalt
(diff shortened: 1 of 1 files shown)


bbockelm:
@gmkurtzer - thoughts on this?  Otherwise, PID namespaces are unconditionally disabled inside the HTCondor job (and I'd like to utilize them as a security feature if HTCondor is running as an unprivileged process).

gmkurtzer:
Heya @bbockelm... If you checkout the devel branch, you will see I've gotten rid of this conditional in etc/init. I've also changed the default behavior to _not_ separate the PID namespace as it requires the generation of a new child process. The behavior can be enabled by the calling user using the --pid, --contain or the SINGULARITY_UNSHARE_PID environment variable and the behavior can be globally disabled by the configuration/system admin.

Does this sound reasonable to you?

bbockelm:
Yes, that sounds reasonable!

Do you mean to say that one would need to pass `--pid --contain` or that `--contain` implies `--pid`?

gmkurtzer:
Yes, sorry for not being clear. The --contain option also implies --pid.

Thanks!

PR #202, opened 2016-08-30
Arch Linux bootstrapping for 2.x

czka:
I think it's pretty much ready. I can help with updating this to a new boostrapping backend in 2.2 in the future.

### DIFF
--- a/examples/arch.def
+++ b/examples/arch.def
@@ -3,7 +3,7 @@
 # Minimal installation process is defined in mods/linux_build_arch.smod. A
 # couple extra actions are called from here using RunCmd and InstallPkgs
 # Singularity bootstrap keywords [1]. Adjust them to your liking as needed.
-# Arch Linux installation guide [3] may come in handy.
+# Arch Linux installation guide [2] may come in handy.
 #
 # [1] http://singularity.lbl.gov/#bootstrap
 # [2] https://wiki.archlinux.org/index.php/Installation_Guide
--- a/examples/arch.def
+++ b/examples/arch.def
@@ -1,14 +1,54 @@
+# By Maciej Sieczka, 2016.
+#
+# Minimal installation process is defined in mods/linux_build_arch.smod. A
+# couple extra actions are called from here using RunCmd and InstallPkgs
+# Singularity bootstrap keywords [1]. Adjust them to your liking as needed.
+# Arch Linux installation guide [3] may come in handy.
+#
+# [1] http://singularity.lbl.gov/#bootstrap
+# [2] https://wiki.archlinux.org/index.php/Installation_Guide
+
 DistType "arch"
 
-# ODO:
-# You can set your preferred Arch mirror here. If it's not set, bootstrap
-# process will try to pick the fastest one using `rankmirrors'.
-MirrorURL 'https://mirrors.kernel.org/archlinux/$repo/os/$arch'
+# "Setup" keyword doesn't do anything anymore, at least as of Singularity
+# 2.1.2. Undocumented "PreSetup" replaced it and is called at each
+# `singularity bootstrap' run.
+# Setup
 
-Setup
 Bootstrap
 
-#InstallPkgs vim
+# Set time zone. Use whatever you prefer instead of UTC.
+RunCmd ln -s /usr/share/zoneinfo/UTC /etc/localtime
+
+# WARNING: Make sure you prefix the STDOUT redirection target (ie. that guy
+# after `>' or `>>`) with $SINGULARITY_BUILD_ROOT in RunCmd calls, as below.
+# Otherwise such command will write to host's filesystem, at least as of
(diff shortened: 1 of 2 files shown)


gmkurtzer:
Looks great! Thanks!

gmkurtzer:
@czka Do you have time to implement Arch Linux bootstrap ability in the new bootstrap model? Also, if you can send me a pull request that has your name and email appended to the AUTHORS file that would be great.

Thanks!

czka:
@gmkurtzer 

I can give it a whirl. Where exactly? There are some changes to bootstrapping on branch master and devel. Master seems to have more changes. There are _.sh files now in libexec/bootstrap/modules-v1 and modules-v2, instead of libexec/mods/_.smod.

gmkurtzer:
I will have it finished and committed hopefully tonight. Yes, you are looking in the right place. I moved things around a bit such that it still supports the old spec format (and potentially others in the future (e.g. docker/vagrant files?)... Anyway, it will be at libexec/bootstrap/modules-v2/dist-arch.sh and you will be able to copy from the dist-debian.sh and dist-redhat.sh as soon as I finish them up.

czka:
OK. Please let me know when all is set. BTW, are there plans to update the bootstrap doc (http://singularity.lbl.gov/#bootstrap)?

gmkurtzer:
Everything should be set for you to start. Let me know if you have any questions. 

Yes, I will be updating the documentation with the release. 

Thanks!

czka:
Looking into it. Don't you have something missing in front of `>` in libexec/bootstrap/modules-v2/dist-redhat.sh, line 88? Is there a reason you are not using a here-document rather than multiple echos?

czka:
Oh, and there's a bunch of typos in dist-*.sh scripts - `messge` instead of `message`.

czka:
And a couple more in prebootstrap.sh, prebootstrap.sh, setup.sh, driver-v2.sh.

gmkurtzer:
Oooops! Yes I'm bad at that particular typo, feel free to fix or I will. Thanks for letting me know. 

Regarding the redirects, it is shell shorthand to redirect nothing to a file to overwrite it with nothing (truncate to zero). I'm still working on the files and a here doc might indeed make more sense. I tend not to use them if it breaks indention and/or not completely necessary. Feel free to make changes. 

Also make sure you add yourself to the authors file!

czka:
OK. I'll fix the typos. Another Q: Is there a reason why you use `ABORT` in some places and `exit` elsewhere?

gmkurtzer:
Yes.... I haven't finished replacing the echoes with ABORTS. Lol

czka:
Erm, so the intention is to use ABORT rather than exit in exceptions?

gmkurtzer:
Yes, that was my thinking... Because then (when necessary) we can also invoke cleanup and relevant output messaging (and also mimic the C Singularity API).

czka:
Makes sense. Under what circumstances are exit code 255 and 1 meant to be used? I can see both in your shell scripts, but can't figure out the pattern.

gmkurtzer:
I want to go through all of the exit codes and create a mapping of exit codes to more detailed messages, but for now what I've been using is 1 = user error, 255 = system error.

czka:
Would an undefined $SINGULARITY_ROOTFS and $SINGULARITY_BUILDDEF be user or system error?

gmkurtzer:
Now that I think about it... Probably system and user errors respectively.

czka:
So I thought as well. Has to be true ;).

czka:
Another Q: in some places you check for failures with `if ! eval "a command"; then` and sometimes you omit the `eval` part like `if ! "a command"; then`. Is there a pattern to this?

gmkurtzer:
Generally I will use `eval` if a command is being represented by a variable (e.g. `eval $MKFS ...`) or if there are multiple layers of variable substitution with an expression.

czka:
OK. And why do you single-quote '$OSVERSION', '$SINGULARITY_ROOTFS', '$MIRROR' in dist-debian.sh bootrap command, but $REQUIRES and $ARCH are not quoted like this? And in dist-redhat.sh bootstrap command no variable is single-quoted.

czka:
I think I get it - single quotes around a `$var` inside an `eval` expression surrounded with double quotes, like `if ! eval "'$var'"`, are equivalent to `if ! eval "\"$var\""` (which I would normally use), and used in order to gracefully handle any weird chars within the `$var`, eg. if it was a filesystem path with spaces in it.  Did I get it right?

gmkurtzer:
Yep, exactly!

czka:
I initially assumed that `'$var'` inside `eval` expression would not be parsed but taken literally, like in `var=ASTRING; echo '$var'` which prints `$var` rather than `ASTRING`, unlike `echo "$var"` and `echo $var` :D. `eval` always gives me hard time. Moving on...

PR #208, opened 2016-09-09
Arch Linux support for v2 format and a bunch of corrections here and there

czka:
@gmkurtzer

Please pay close attention to my "eval" calls before merging. It will be great help. I hope them right in all cases...

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -12,4 +12,5 @@ Contributors:
     Brian Bockelman <bbockelm@cse.unl.edu>
     Amanda Duffy aduffy@lenovo.com
     Jarrod Johnson jjohnson2@lenovo.com
+    Maciej Sieczka <msieczka at sieczka dot org>
 
--- a/examples/arch.def
+++ b/examples/arch.def
@@ -1,0 +1,48 @@
+#!/bin/bash
+#
+# Copyright (c) 2015-2016, Maciej Sieczka, Gregory M. Kurtzer. All rights
+# reserved.
+#
+# "Singularity" Copyright (c) 2016, The Regents of the University of California,
+# through Lawrence Berkeley National Laboratory (subject to receipt of any
+# required approvals from the U.S. Dept. of Energy).  All rights reserved.
+
+# Minimal installation process is defined in
+# libexec/bootstrap/modules-v2/dist-arch.sh. A couple extra actions are called
+# from here in `%post' section. Adjust them as needed.
+# https://wiki.archlinux.org/index.php/Installation_Guide may come in handy.
+
+OSType: arch
+
+%runscript
+    echo "This is what happens when you run the container..."
+
+%post
+echo "Hello from inside the container"
+
+# Set time zone. Use whatever you prefer instead of UTC.
+ln -s /usr/share/zoneinfo/UTC /etc/localtime
+
+# Set locale. Use whatever you prefer instead of en_US.
+echo 'en_US.UTF-8 UTF-8' > /etc/locale.gen
+# Add more locales as needed, eg:
+# echo 'pl_PL.UTF-8 UTF-8' >> /etc/locale.gen
+locale-gen
+echo 'LANG=en_US.UTF-8' > /etc/locale.conf
+# Mind that Singularity's shell will use host's locale no matter what
+# anyway, as of version 2.1.2. This may change in a future release.
+
+# Set the package mirror server(s). This is only for the output image's
+# mirrorlist. `pacstrap' can only use your hosts's package mirrors.
+echo 'Server = https://mirrors.kernel.org/archlinux/$repo/os/$arch' > /etc/pacman.d/mirrorlist
+# Add any number of fail-over servers, eg:
+echo 'Server = https://archlinux.honkgong.info/$repo/os/$arch' >> /etc/pacman.d/mirrorlist
+
+# I need VIM and Bash completion. Specify your extra packages as needed.
+echo "Hello from inside the container #2"
+pacman -Sy vim bash-completion
+echo "Hello from inside the container #3"
+# Remove the packages downloaded to image's Pacman cache dir.
+echo "Hello from inside the container #4"
+paccache -r -k0
+echo "Hello from inside the container #5"
--- a/examples/arch.def
+++ b/examples/arch.def
@@ -15,7 +15,7 @@
 OSType: arch
 
 %runscript
-    echo "This is what happens when you run the container..."
+echo "This is what happens when you run the container..."
 
 %post
 echo "Hello from inside the container"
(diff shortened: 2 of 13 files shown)


gmkurtzer:
Looks great!

gmkurtzer:
I fixed your copyright lines in [master be8a6e9]. I also think you should add MirrorURL to your definition file and possibly OSVersion too...

Thoughts?

czka:
Thanks for accepting my contribution. That was fun :).

You are right about the copyright lines.

As for the backtick vs apostrophe - I'm using the former on purpose for tool names, sometimes code pieces etc. For other things I would use double quotes usually. A matter of style probably. See eg. `man find`, `man iptables`. This convention has long history - https://www.emacswiki.org/emacs/EmacsDocumentationNotation.

MirrorURL will not be used for anything by `pacstrap`. The tool uses host's mirrorlist only, and by default it even tries to copy them over to a new system. It's a rather simple shell script - 
https://git.archlinux.org/arch-install-scripts.git/tree/pacstrap.in. Also, it's expected on Arch system to have multiple mirror servers defined. Eg.:

```
$ cat /etc/pacman.d/mirrorlist
################################################################################
################# Arch Linux mirrorlist generated by Reflector #################
################################################################################

# With:       reflector --protocol https --sort rate --save /etc/pacman.d/mirrorlist
# When:       2015-11-02 22:16:46 UTC
# From:       https://www.archlinux.org/mirrors/status/json/
# Retrieved:  2015-11-02 22:16:29 UTC
# Last Check: 2015-11-02 22:11:45 UTC

Server = https://mirror.pseudoform.org/$repo/os/$arch
Server = https://ftp.lysator.liu.se/pub/archlinux/$repo/os/$arch
Server = https://archlinux.honkgong.info/$repo/os/$arch
Server = https://archlinux.limun.org/$repo/os/$arch
Server = https://mirrors.uni-plovdiv.net/archlinux/$repo/os/$arch
Server = https://ftp.myrveln.se/pub/linux/archlinux/$repo/os/$arch
Server = https://ftp.fau.de/archlinux/$repo/os/$arch
Server = https://mirror.fluxent.de/archlinux/$repo/os/$arch
Server = https://mirror.i3d.net/pub/archlinux/$repo/os/$arch
Server = https://arch.jensgutermuth.de/$repo/os/$arch
Server = https://mirror.fs.cx/archlinux/$repo/os/$arch
Server = https://arch.localmsp.org/arch/$repo/os/$arch
Server = https://mirror.js-webcoding.de/pub/archlinux/$repo/os/$arch
Server = https://archlinux.my-universe.com/$repo/os/$arch
Server = https://ftp.heanet.ie/mirrors/ftp.archlinux.org/$repo/os/$arch
Server = https://foss.aueb.gr/mirrors/linux/archlinux/$repo/os/$arch
Server = https://mirror.bitjungle.info/archlinux/$repo/os/$arch
Server = https://k42.ch/mirror/archlinux/$repo/os/$arch
Server = https://lug.mtu.edu/archlinux/$repo/os/$arch
Server = https://mirror.lty.me/archlinux/$repo/os/$arch
Server = https://mirrors.ocf.berkeley.edu/archlinux/$repo/os/$arch
Server = https://kwk.pw/arch/$repo/os/$arch
Server = https://www.mailtunnel.eu/archlinux/$repo/os/$arch
Server = https://archlinux.mirror.pkern.at/$repo/os/$arch
Server = https://mirrors.kernel.org/archlinux/$repo/os/$arch
Server = https://fooo.biz/archlinux/$repo/os/$arch
Server = https://mirrors.ustc.edu.cn/archlinux/$repo/os/$arch
Server = https://arch.static.lu/$repo/os/$arch
```

Given that, I can't see a use for MirrorURL in arch.def.

The concept of OSVersion does not fit Arch world. It's a rolling release ("OSVersion") by design. There are "tesing" and "unstable" repos (https://wiki.archlinux.org/index.php/Official_Repositories#testing), but the way they are enabled/disabled by the user does not mean anything in the context of the intended OSVersion use, as I get it at least.

 Is there some fundamental reason why it's better to have these set for Singularity anyway?

gmkurtzer:
OK, that makes perfect sense. No fundamental reason why I want them set aside to keep in alignment with the other bootstraps. Would you be open to this?

```
OSType: arch
OSVersion: Null
MirrorURL: Null
```

Also, what about the "Include" directive which automatically installs additional packages. Is there a way to integrate that?

I have another question. Is it possible to make it such that Arch can be bootstrapped from a non-Arch system? For example, I think Arch has a bootstrap that can be used for this purpose.. You first download a minimal chroot, and then run the packstrap from there into a directory. Can we use that if someone wanted to run this from lets say... Ubuntu or Red Hat?

If not, it is fine, I just want to make sure that there is an Arch Linux base container that people can download. Do you know any of the Arch maintainers? Do you think they would be willing to also distribute a Singularity container from their download page?

Thanks again, and if you see anything else that needs fixing, please don't hesitate to jump in and send me a pull request!

czka:
> Would you be open to this?

OK. But with a few words explaining what thoses `Null` entries mean. Otherwise it may look kindoff "intriguing" ;).

> what about the "Include" directive which automatically installs additional packages. Is there a way to integrate that?

Sure, but I decided it's an overhead, as one can install whatever extra packages he needs in `%post` with plain `pacman -Sy --noconfirm package1 package2 package3`. I believe that bootstrapping should be just it - installing a minimum required to get the system rolling. Anything above that is not a part of bootstrapping as I see it, and the `%post` section is a right place for that. If supporting the `Include` keyword in all bootstrap scripts is eg. crucial to further Singularity development please add whatever's needed in your opinion.

> Is it possible to make it such that Arch can be bootstrapped from a non-Arch system?

Never done it myself, but they are saying it's rather easy: https://wiki.archlinux.org/index.php/Install_from_Existing_Linux#From_a_host_running_another_Linux_distribution.

> Do you know any of the Arch maintainers?

Nope, sorry.

I was wondering about another thing. I have just succesfully built an i686 Arch image on an x86_64 Arch host. And it seems to run OK on that 64bit host. Should I expect any issues with such setup? Maybe Singularity bootstrap process could take such combination into account, if it's meant to be supported?

gmkurtzer:
The point about the Null entries is just to show that these are standard keywords but this particular bootstrap definition doesn't take them. Maybe a comment or two.

Understood about the "Include" point. Maybe we can also add that as a Null option as well.

You bring up a really good point about arch! Maybe we should add a bootstrap keyword for OSArch!  And no, I do not anticipate any issue running 32bit on a 64bit kernel that supports multi-arch.

czka:
> Maybe we should add a bootstrap keyword for OSArch!

I like the idea. Submitted in #210.

PR #215, opened 2016-09-14
Add some simple bash completion scripts.

bbockelm:
Ok, I have to admit - improve bash tab-completion isn't probably the most high-priority project in the world.

OTOH, I find it really annoying when they aren't there...


gmkurtzer:
Oh, I know exactly what you mean! This is excellent, thank you!!

PR #218, opened 2016-09-20
Add/bootstrap docker

vsoch:
This is early / initial work to add bootstrap functionality for Docker images. Several things that need to be worked out:
1. The docker registry API (version 2.0) I cannot get to work with the instructions they have on their docs. The call that works for the tags, for example:
   
     https://registry.hub.docker.com/v1/repositories/library/ubuntu/tags

does not include latest. If we change v1 to v2:

```
 https://registry.hub.docker.com/v2/repositories/library/ubuntu/tags
```

The image_ids are null, and then we can't look up layers here:

```
  https://registry.hub.docker.com/v1/repositories/library/ubuntu/images
```

We need to figure out this new API spec, or better, get help from Docker to get it working, so we can properly retrieve the layers for any tag.
1. The bootstrap command requires sudo. @chrisfilo suggested extracting all layers into a single .tar.gz, and I have not tested this yet. Ideally the user should be able to bootstrap an image without needing sudo so it can be done on a research cluster.
2. the "import" functionality for docker needs to be extended into these python modules, I can do this next.

@gmkurtzer  - we really should move the main repo into singularityware... let me know your thoughts and how you think would be best to do this! My suggestion is simply to start using it as the base, and when the new docs are ready (also hosted here) we point users to it.


gmkurtzer:
Looks great, thank you!

PR #219, opened 2016-09-21
Fixing "get_manifest" to show list of available tags if provided tag fails

vsoch:
This is a quick PR to add a line in to the except loop of get_manifest that will show the user a list of acceptable tags given that the provided tag failed. I meant to include yesterday, wrote the function, and just forgot to add the line :)


gmkurtzer:
Got it, thanks!

PR #225, opened 2016-09-23
add python to the libexec dir in singularity.spec

DrDaveD:
Without this line in singularity.spec, I am unable to do an rpmbuild on el7.

### DIFF
--- a/singularity.spec.in
+++ b/singularity.spec.in
@@ -60,6 +60,7 @@ rm -rf $RPM_BUILD_ROOT
 %{_libexecdir}/singularity/image-create
 %{_libexecdir}/singularity/image-expand
 %{_libexecdir}/singularity/image-mount
+%{_libexecdir}/singularity/python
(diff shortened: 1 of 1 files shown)


gmkurtzer:
You beat me to it! Thanks!

bbockelm:
@gmkurtzer - would it be possible to do a `rpmbuild` per-commit in Travis?

gmkurtzer:
That's a good idea. Hrmm.. should we do it in their default image, or do you know if there is a way to have it also run a test in Centos?

bbockelm:
I don't know of a way to get a CentOS VM from Travis - in other projects, I've just done a CentOS Docker container inside the VM.  See:

https://github.com/opensciencegrid/htcondor-ce/blob/master/tests/setup_tests.sh
https://github.com/opensciencegrid/htcondor-ce/blob/master/tests/test_inside_docker.sh

That allows us to build the RPMs.

PR #226, opened 2016-09-23
re-adding fix to remove targz after layer extraction (oups)

vsoch:
ehehhe, oups. Apologies to your computer!


gmkurtzer:
We could also in theory not cache them at all, and run them into tar via a pipe, right?... But then we miss out on Chris' future RFE of keeping them.

PR #229, opened 2016-09-23
Scratch directory fixups

bbockelm:
A few improvements for the scratch directories:
- Allows scratch directories to be used even if `overlayfs` is disabled.  If the directory doesn't exist inside the container, it is a fatal error.
- Allows multiple scratch directories to be specified.
- Add bash completion updates for `--scratch`.


gmkurtzer:
Looks good, thanks!

PR #230, opened 2016-09-23
Add support for user-specified PWD from the command line.

bbockelm:
When the `$PWD` is not available inside the container, singularity currently defaults to starting the process inside the user's `$HOME`.

It's not a bad strategy, particularly for interactive use.  However, when invoked from other tools (such as a batch system), it would be useful to be able to specify this directly.

### DIFF
--- a/src/lib/action/action.c
+++ b/src/lib/action/action.c
@@ -104,25 +104,28 @@ int singularity_action_do(int argc, char **argv) {
     singularity_priv_drop_perm();
 
     singularity_message(DEBUG, "Trying to change directory to where we started\n");
-    if ( chdir(cwd_path) < 0 ) {
-        struct passwd *pw;
-        char *homedir;
-        uid_t uid = singularity_priv_getuid();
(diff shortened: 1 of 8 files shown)


src/lib/action/action.c
```diff
@@ -104,25 +104,28 @@ int singularity_action_do(int argc, char **argv) {
...
-
-        errno = 0;
-        if ( ( pw = getpwuid(uid) ) != NULL ) {
-            singularity_message(DEBUG, "Obtaining user's homedir\n");
-
-            homedir = pw->pw_dir;
-
-            if ( chdir(homedir) < 0 ) {
-                singularity_message(WARNING, "Could not chdir to home directory: %s\n", homedir);
+    char *target_pwd = getenv("SINGULARITY_TARGET_PWD");
```
gmkurtzer, line 107: I wrote some envar*() wrappers to getenv() to check sanity of a few things, and limit exposure to some exploits. I will fix this after merge.

gmkurtzer:
I had the one comment, but aside from that, looks good!

PR #236, opened 2016-09-29
Enable slave mounts by default

bbockelm:
I haven't seen any issues when utilizing mounts, I'm not aware of any security concerns, and it does allow `autofs` to work by default.

Let's toggle the default configuration value.


gmkurtzer:
I literally just had someone ask me about this.... Does slave enable an automount from within the container? BTW, I am accepting this. Thanks!

PR #243, opened 2016-10-04
Fix home directory autocreation permission

jjohnson42:
If the home directory is a subdirectory of a directory
not already in image, use 755 as permission for the new
directory.  This means a parent directory will not be
unreadable by the user.


gmkurtzer:
Good catch, thanks!

PR #244, opened 2016-10-04
Create r_python_julia.def

cjprybol:
example for generating a container with 3 commonly used languages for data analysis


gmkurtzer:
Looks good, thanks!

PR #245, opened 2016-10-04
Create linuxbrew_and_non-root_software_example.def

cjprybol:
Example of how to install linuxbrew, which makes it easy to install many scientific software libraries. Example can be extended to install other software that refuses to be installed as root (although I can't think of any others off hand)


gmkurtzer:
Thanks!

PR #248, opened 2016-10-05
Null terminate envar result, and don't lose track of malloc

DrDaveD:
When passing a -B on to a singularity exec with -C and without without -W, I was getting fatal errors that it was trying to bind something that didn't exist.  I tracked it down to a missing null terminator in the envar function.  This pull request fixes the problem.  While there I noticed it was also possible for the function to malloc some space and lose track of it.

### DIFF
--- a/src/util/util.c
+++ b/src/util/util.c
@@ -43,7 +43,7 @@
 
 
 char *envar(char *name, char *allowed, int len) {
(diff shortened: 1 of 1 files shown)


gmkurtzer:
Looks good, thank you!

PR #251, opened 2016-10-06
Support setting home directory inside of container

DrDaveD:
When only using bind mounts (no overlayfs) then if the top level directory in the user's home directory path happens to not be in the image, mounting the user's home directory fails.  This PR enables setting a different home directory for the user inside the container.   I could have added a separate option, but I thought it fit pretty naturally as an addition to the -H option when a second colon-separated component is added.  It's quite a bit like the -B option, although slightly different.

One question I had is if there should be more sharing of code when looking up the home directory, but it's not clear to me exactly what a shared function should have in it nor where it should go.  I suppose it would look at both the environment variable and the passwd file, but have an option to not look up the passwd file because there's one case where that has already been done.


gmkurtzer:
I think you are right, we should create a function which returns the home directory location and for that, it might make sense to make it a feature not of -H, but a separate option.

Thoughts?

DrDaveD:
I don't think using a function need be connected with having a separate option -- it is just easy for the function to read from an environment variable after a colon as it is to read from a whole environment variable.  I rather like extending the purpose of the -H option, I think it's easier to explain than to have one option for the source of the home directory and a separate one for the destination.

So if there should be a function, where would it be?  util?  If there, in util.c or does it need a new source file?

gmkurtzer:
I think util.c is a good home for it. :)

DrDaveD:
Ok, I moved looking up the home directory to a common function in util.c, please tell me what you think of it now.

Note the subtle difference between -B and -H semantics, in that with -B if you leave out the colon it applies to both the bind source and destination, where with -H it applies only to the source and the destination is left alone.  Now that getting the home directory is in a common function, if we wanted to change the functionality of -H to apply to both source and destination when there's no colon, in order to make it behave more like -B, it could be implemented by a trivial change in the get_homedir() function.  It would however change the meaning of the old -H function in an incompatible way so I don't know whether or not you want to do that.  Personally I think it would make sense to do it, but I don't know who added -H or why, or if they're depending on the current functionality.

gmkurtzer:
Apologies for the slow responses on these (this is seriously a crazy week, 3 more presentations still to do...). I promise I will look through this in detail and merge it after the release (which should be later today).

Thanks!

DrDaveD:
Greg, can you look at this in more detail?

gmkurtzer:
Yep, thanks for the ping. On it now!

gmkurtzer:
Looks good, thanks!

DrDaveD:
@gmkurtzer Please consider this feature for inclusion in version 2.2.1.  When I was chatting with @bbockelm a couple weeks ago I think I convinced him that this is needed for integration with HTCondor for use on the LHC Computing Grid.  He didn't realize that the home directory was otherwise getting mapped even with the -C/--containall option, and it needs to be controlled for security reasons.

gmkurtzer:
Hi @DrDaveD,

Point releases should not include new features or API changes or additions (this PR technically does both). While I appreciate the request this will probably have to wait until 2.3 which is still at least a couple of months out. 

Can you describe what you meant about the home directory getting mapped even with the -C/--containall option? Security implications can be used as justification for inclusion in point releases. lol

Greg

bbockelm:
Hi Dave -

Given our various timelines, we can probably wait until 2.3 or simply backport ourselves.

Brian

DrDaveD:
> Can you describe what you meant about the home directory getting mapped even with the -C/--containall option? Security implications can be used as justification for inclusion in point releases. 

Even when using the -C option, the user's home directory by default gets mapped into the container, and that means that a payload job inside the container can do all sorts of undesirable things including modifying .profile and .ssh settings.  Brian had thought that -C eliminated all default mappings, but that's not the case.  When overlayfs is not in use, the top level directory (e.g. /home) has to already exist inside the container in order to bind onto something, so sometimes the mapping fails, but sometimes it does succeed and that's a problem.  With my addition to -H, we can set both the source of a scratch home directory for the user and choose a destination that we know exists inside the container.

Is that justification enough for including it in 2.2.1?  If not, as Brian says we'll probably add the patch to the version we distribute for the Open Science Grid, or wait for 2.3.  An alternative to address just the security concern would be to change -C to disable the default home directory mapping.

gmkurtzer:
By default the -C option should create an empty directory in the sessiondir, and bind that to home instead of the user's real home. Are you saying that there is a failure mode which allows the user's real home directory to still be used even when -C is given?

If so, can you send me a reproducer so I can experiment?

Thanks!

DrDaveD:
Ah, I never actually saw the home directory mapped in because my home directory's top level directory did not have a corresponding bind path inside the container.  I just assumed it was the real home directory because it attempted to map to the real path.  So it's probably not a security issue, it's just that in many cases there will be no home directory.

gmkurtzer:
Ahh, no. It should not use the user's real home directory when you use the -C option. It maps in the empty directory within the sessiondir prefix or scratchdir if that is defined.

There is a request to change the usage of the -H option such that it always changes the user's home directory path to be that of the virtualized directory for example:

```bash
$ singularity shell -H /home/gmk/virthome container.img
Singularity.container.img> echo $HOME
/home/gmk/virthome
Singularity.container.img> pwd
/home/gmk/virthome
```

This will get rid of a conditional second portion of the `-H /path1:/path2`. Thoughts?

DrDaveD:
I think it's a good idea to have that as the default without a colon, and to still have the ability with the colon to separate out source and destination.  That will make it the same semantics as -B.  As I mentioned above, it could be implemented with a trivial change to the get_homedir() function that this pull request added.

DrDaveD:
Hey @gmkurtzer, one of my commits from this pull request, d31856c, was silently included in singularity version 2.2.1.   See [the 2.2.1 home.c](https://github.com/singularityware/singularity/blob/2.2.1/src/lib/mount/home/home.c).  Was that intentional? The help message update 7d778b2 was not included, nor the rest of the commits having to do with reading the home directory in a common function.  I was surprised to find out in a presentation today by @bbockelm that we are in fact now using this feature in production at about 15 sites.

gmkurtzer:
It was both intentional and missed (respectively). I apologize for that @DrDaveD and thank you for pointing it out!

@bbockelm You are presenting today and I wasn't invited?! I feel gypped!

PR #257, opened 2016-10-14
Image definition file for ROOT with singularity

bauerm97:
Provides a definition file for building ROOT scientific library in a singularity container.


gmkurtzer:
Looks good, thanks!

PR #258, opened 2016-10-18
initial port for alpinelinux

truatpasteurdotfr:
the 2.1.2 works fine but 2.2 and master are failing during the installation phase of "abuild -r", the manual installation procedure works fine.. 

imho, it is caused by the "install-data-hook: make_suid" changed, imho it should be reverted to "install-exec-hook: make_suid"


gmkurtzer:
Thanks!

PR #259, opened 2016-10-18
make_suid fix for alpine linux

truatpasteurdotfr:
allow the abuild wrapper to chmod 4755 sexec-suid

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -43,7 +43,7 @@ get_section_LDADD = lib/libsingularity.la
 
 #dist_suidPROGRAM_INSTALL = ${INSTALL} -m 640
 
-install-data-hook: make_suid
+install-exec-hook: make_suid
 
 make_suid:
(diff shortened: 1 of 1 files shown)


gmkurtzer:
If I remember correctly, I used to have it as this, but it was failing under some circumstances. Have you tested on different Centos versions too?

gmkurtzer:
@truatpasteurdotfr ping

truatpasteurdotfr:
pong :P 
it works for me on centos7 and centos6 build host compiling singularity 2.2.

gmkurtzer:
Heya @truatpasteurdotfr, did you mean to close the PR? I'm happy to implement it as long as it doesn't break builds on other OSs. I seem to recall that I used to have it defined as such, and it didn't work for one reason or another somewhere, someplace.. but I really don't recall the reason right now. I could dig if necessary...

truatpasteurdotfr:
my bad :P re-opening. If you have time I would also like to know  why the perms settings was pushed to the data hook instead of exec. :)

Issue #256, opened 2016-10-14
WARNING: Ignoring user bind request: host does not support PR_SET_NO_NEW_PRIVS


soichih:
I am using singularity 2.2 on Cray XE6 HPC system. SLES11-SP3 (x86_64). Linux 3.0.101

It has cudatoolkit installed on it and I am trying to use it by binding it from the container.

> singularity shell -B /opt/nvidia/cudatoolkit7.0/7.0.28-1.0502.10742.5.1:/usr/local/cuda soichi.img

I am seeing following warning message, and mounting fails. 
WARNING: Ignoring user bind request: host does not support PR_SET_NO_NEW_PRIVS
WARNING: Not mounting current directory: host does not support PR_SET_NO_NEW_PRIVS
Singularity: Invoking an interactive shell within container...
..

Does this mean I can not do use bind on our system?

Thank you!

gmkurtzer:
According to that yes, it seems that prctl () didn't work at compile time. Can you send me the output of configure and also tell me what version of kernel it is running? I am kind of surprised SLES would fail on that.

Thanks!

soichih:
@gmkurtzer 

```
$ uname -a
Linux login2 3.0.101-0.46-default #1 SMP Wed Dec 17 11:04:10 UTC 2014 (8356111) x86_64 x86_64 x86_64 GNU/Linux
```

I've asked the person who installed it on our cluster for configure output, but I believe it should look similar to following.

```
checking build system type... x86_64-suse-linux-gnu
checking host system type... x86_64-suse-linux-gnu
checking target system type... x86_64-suse-linux-gnu
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... /bin/mkdir -p
checking for gawk... gawk
checking whether make sets $(MAKE)... yes
checking for a BSD-compatible install... /usr/bin/install -c
checking for style of include used by make... GNU
checking for gcc... gcc
checking for C compiler default output file name... a.out
checking whether the C compiler works... yes
checking whether we are cross compiling... no
checking for suffix of executables... 
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking dependency style of gcc... gcc3
checking for a sed that does not truncate output... /usr/bin/sed
checking for grep that handles long lines and -e... /usr/bin/grep
[...]

PR #260, opened 2016-10-18
Implemented ability for config file to specify allowable bind directories

bauerm97:
In #256 there was an issue reported that users could never bind when PR_SET_NO_NEW_PRIVS is not enabled on the kernel. @bbockelm proposed a solution where an admin could list directories in the config file that are always allowed to be bind mounted. This pull request implements this suggestion and provides new functionality:

New config options:

```
# ALLOW SELECT BINDS: [BOOL]                                                                                                             
# DEFAULT: no                                                                                                                            
# Limit user bind options to the specified source and destination                                                                       
# directories in singularity.conf                                                                                                        
allow select binds = no


# USER BIND SOURCE: [STRING]                                                                                                             
# DEFAULT: Undefined                                                                                                                     
# Specify locations that users are always allowed to bind from, regardless of                                                            
# USER BIND CONTROL. USER BIND SOURCES is a list of locations on the host                                                                
# file system.                                                                                                                           
#user bind source =                                                                                                                      


# USER BIND DESTINATION: [STRING]                                                                                                        
# DEFAULT: Undefined                                                                                                                     
# Specify locations that users are always allowed to bind to, regardless of                                                              
# USER BIND CONTROL. USER BIND SOURCES is a list of locations on the container                                                           
# file system.                                                                                                                           
#user bind destination =
```

user bind source/destination configs function like bind path with one entry per line. Any subdirectory of the directories specified in user bind source/destination will also be accepted. If user bind control and allow select binds are both set to yes, users will still only be able to bind on the specified directories. Attempting to bind a location not specified in the config will trigger a WARNING to the user and continue as normal without mounting, however an attempt will still be made to bind mount other locations if multiple are present.

Current default for allow select binds is no, however it could be changed to yes to allow for more security. If you set allow select binds default to yes and revert user bind control default to no, any administrator must manually switch on user bind control, possibly preventing admins from accidentally leaving open a security hole.

### DIFF
--- a/etc/singularity.conf
+++ b/etc/singularity.conf
@@ -118,24 +118,24 @@ user bind control = yes
 # ALLOW SELECT BINDS: [BOOL]
 # DEFAULT: no
 # Limit user bind options to the specificed source and destination
-# directories specificed in singularity.conf
+# directories in singularity.conf
 allow select binds = no
 
 
-# USER BIND SOURCES: [STRING]
+# USER BIND SOURCE: [STRING]
 # DEFAULT: Undefined
 # Specify locations that users are always allowed to bind from, regardless of
 # USER BIND CONTORL. USER BIND SOURCES is a list of locations on the host
 # file system.
-user bind sources = 
+#user bind source = 
 
 
-# USER BIND DESTINATIONS: [STRING]
+# USER BIND DESTINATION: [STRING]
 # DEFAULT: Undefined
 # Specify locations that users are always allowed to bind to, regardless of
 # USER BIND CONTORL. USER BIND SOURCES is a list of locations on the container
 # file system.
-user bind destinations = 
+#user bind destination = 
 
 
 # MOUNT SLAVE: [BOOL]
--- a/etc/singularity.conf
+++ b/etc/singularity.conf
@@ -115,6 +115,29 @@ bind path = /etc/hosts
 user bind control = yes
 
 
+# ALLOW SELECT BINDS: [BOOL]
+# DEFAULT: no
+# Limit user bind options to the specificed source and destination
+# directories specificed in singularity.conf
+allow select binds = no
+
+
+# USER BIND SOURCES: [STRING]
+# DEFAULT: Undefined
+# Specify locations that users are always allowed to bind from, regardless of
+# USER BIND CONTORL. USER BIND SOURCES is a list of locations on the host
+# file system.
+user bind sources = 
+
+
+# USER BIND DESTINATIONS: [STRING]
+# DEFAULT: Undefined
+# Specify locations that users are always allowed to bind to, regardless of
+# USER BIND CONTORL. USER BIND SOURCES is a list of locations on the container
+# file system.
+user bind destinations = 
+
+
 # MOUNT SLAVE: [BOOL]
 # DEFAULT: yes
 # Should we automatically propagate file-system changes from the host?
--- a/src/lib/mount/userbinds/userbinds.c
+++ b/src/lib/mount/userbinds/userbinds.c
@@ -92,7 +92,7 @@ void singularity_mount_userbinds(void) {
 	      }
 	      if ( tmp_path == NULL ) {
 		singularity_message(WARNING, "Ignoring user bind request: %s is not in allowed sources.\n", source);
-		return;
+		continue;
 	      } else {
 		singularity_message(DEBUG, "Proceeding since %s is subdir of %s\n", source, tmp_path);
 	      }
@@ -104,7 +104,7 @@ void singularity_mount_userbinds(void) {
 	      }
 	      if ( tmp_path == NULL ) {
 		singularity_message(WARNING, "Ignoring user bind request: %s is not in allowed destinations.\n", dest);
-		return;
+		continue;
 	      }
 	    }
 	      
--- a/src/lib/mount/userbinds/userbinds.c
+++ b/src/lib/mount/userbinds/userbinds.c
@@ -46,7 +46,7 @@ void singularity_mount_userbinds(void) {
     if ( ( bind_path_string = envar_path("SINGULARITY_BINDPATH") ) != NULL ) {
 
         singularity_message(DEBUG, "Checking for 'user bind control' in config\n");
-        if ( singularity_config_get_bool("user bind control", 1) <= 0 ) {
+        if ( ( singularity_config_get_bool("user bind control", 1) <= 0 ) && ( singularity_config_get_bool("allow select binds", 1) <=0 ) ) {
             singularity_message(WARNING, "Ignoring user bind request: user bind control is disabled by system administrator\n");
             return;
         }
--- a/src/lib/mount/userbinds/userbinds.c
+++ b/src/lib/mount/userbinds/userbinds.c
@@ -51,6 +51,7 @@ void singularity_mount_userbinds(void) {
             return;
         }
 
+	singularity_config_rewind();
 	if (singularity_config_get_bool("allow select binds", 0) <= 0) {
 	  singularity_message(WARNING, "allow select binds set to off\n");
 	}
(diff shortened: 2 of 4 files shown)


bbockelm:
(Note - out of office for a few days...)

The is_subdir implementation is problematic because:
- doesn't normalize path (I.e., handle "/../../")
- there are various TOCTOU issues in walking the directory path and making sure filesystem boundaries / symlink issues are handled correctly.
  - not sure the level of paranoia is needed here; for the worst case, see the "safefile" paper / library by Kupsch et al.  It can get really nasty to "do it right".

But yes, once we can get the is_subdir method nailed down, the rest is likely correct.

bauerm97:
@bbockelm Some comments on your points

> - doesn't normalize path (I.e., handle "/../../")

For now this shouldn't be a huge issue, if you pass relative paths it will fail out. A later commit could implement functionality to parse the relative paths. I think this could be implemented by first passing the path's through some hypothetical function get_absolute_path() before any parts of is_subdir() run, however the get_absolute_path() code is the bulk of the work.

> - there are various TOCTOU issues in walking the directory path and making sure filesystem boundaries / symlink issues are handled correctly.

I understand the issues here but I'm not sure how we can properly handle them. If you know of a way to fix these issues please let me know. One thing we could do is during the handling of the path comparison, make sure that if any of the paths are symlinks to just break out, something like this might work

```
while ( ( is_link(test_subpath) != 0 ) && ( strcmp(test_subpath, "/") != 0 ) ) {
        if ( strcmp(test_subpath, test_path) == 0 ) {
            singularity_message(DEBUG, "%s is subdir of %s\n", subpath, path);
            return(1);
        } else {
            test_subpath = dirname(strdup(test_subpath));
        }
    }
```

> - not sure the level of paranoia is needed here; for the worst case, see the "safefile" paper / library by Kupsch et al.  It can get really nasty to "do it right".

Again, also not sure about this one. An easy "solution" to these issues for the time being would be to simply remove the ability to mount subdirectories at all, and only allow the specified directories to be mounted. Then later go back and re-implement the subdirectory functionality once the security is solved properly.

bauerm97:
I pushed an updated is_subdir implementation that properly handles relative paths and symlinks. It turns out there is a function that we can use to expand symlink locations and relative paths, realpath(). All that was necessary was swapping

`test_path = strndup(path, strlen(path)-1);`

with

`test_path = realpath(strndup(path, strlen(path)-1), NULL);`

Assuming that the is_subdir still passes (after resolving symlinks and relative paths), the original paths specified with the -B flag will be mounted as if user binds is on. This does not re assign the fully resolved path to source/dest in userbinds.c loop.

bbockelm:
I've been thinking on this one more.

First, there's no point in controlling the destination directory.  The user manages the image - hence, by playing hardlink tricks (possibly by corrupting the image), they can likely get any directory to appear anywhere they want.  So, the point of the exercise is to prevent _any_ =setuid= binaries from entering the runtime.  So, start with the assumption that there are no =setuid= binaries before we hit the user bind code.

Next, for the =is_subdir= function - I think we could make it something like this (sketching pseudocode in python):

```
# Returns -1 on failure; otherwise, returns a fd for the testdir.
def is_subdir(testdir, basedir):
   real_fname = os.path.split(os.path.realpath(testdir))[-1]
   base_st = os.stat(basedir)
   test_st = os.stat(testdir)
   matches = ((base_st.st_dev == test_st.st_dev) and (base_st.st_ino == test_st.st_ino))
   if matches:
     parent_fd = os.open(basedir) # add O_CLOEXEC | O_NOFOLLOW
   else:
     parent_fd = is_subdir(testdir + "/..", basedir)
   test_fd = posix.openat(parent_fd, real_fname, O_CLOEXEC)
   test_fd_st = os.fstat(test_fd)
   os.close(parent_fd)
   if (test_fd_st.st_dev != test_st.st_dev) or (test_fd_st.st_ino != test_st.st_ino):
     os.close(test_fd_st)
     return -1
   return test_fd_st
```

Then, the logic looks like:
1. Bind mount specified source into a temp directory _not_ visible in the image.
2. Drop privileges to the target user.
3. Run the `is_subdir` routine above.  If the source is actually a file, pass the parent directory to `is_subdir` and repeat the `openat` test to make sure the file is actually in the parent directory.
4. If `is_subdir` passes, check that the inode/device of the final FD still matches the temporary directory.
5. If everything checks out, return privileges to the superuser and bind mount the temporary directory into the image.

This covers every possible race condition I could think of.

bauerm97:
@bbockelm OK I think I'm following what you're getting at here, I took quick shot at implementing the updated is_subdir in C. I think this should work, it's slightly different from the way yours is. Let me know what you think.

```
int is_subdir(char *testpath, char *basepath) {
  struct stat testpath_st;
  struct stat basepath_st;
  struct stat test_fd_st;
  char *testpath_real;
  int parent_fd;
  int test_fd;
  realpath(testpath, testpath_real);
  char *real_fname = basename(testpath_real);

  if ( ( stat(testpath, &testpath_st) == -1 ) || ( stat(basepath, &basepath_st) == -1 ) ) {
    return(-1);
  }

  if ( (basepath_st.st_dev == testpath_st.st_dev) && (basepath_st.st_ino == testpath_st.st_ino) ) {
    parent_fd = open(basepath, O_CLOEXEC);
  } else {
    parent_fd = is_subdir(dirname(testpath), basepath);
  }

  test_fd = openat(parent_fd, real_fname, O_CLOEXEC);
  fstat(test_fd, &test_fd_st);
  close(parent_fd);

  if ( (test_fd_st.st_dev != testpath_st.st_dev) || (test_fd_st.st_ino != testpath_st.st_ino) ) {
    close(test_fd)
    return(-1);
  }

  return(test_fd);
}

```

Returns -1 if fails, otherwise returns the int test_fd. Let me know how that looks

gmkurtzer:
@bbockelm Why do you mention implement a bind mount?

We should be able to take device and inode IDs from the directory listed as `user bind source` and then walk the paths via a loop backwards on the full paths dirname() from the user requested bind location to see if it exists within a directory matching the device inode ID of the `user bind source`. Why throw another bind mount into the mix?

BTW, nice talk today! It was good to see and hear you! :)

gmkurtzer:
I was talking to both @bauerm97 and @bbockelm (separately), but here is what I'm proposing:

```
is_subdir(source, path) {
    source_st = stat(source)
    testpath = strdup(path);
    while(testpath) {
        test_st = stat(testpath);
        if ( test_st == source_st ) {
            return(TRUE);
        }
        testpath = dirname(testpath);
    }
    return(-1);
}
```

And if we want to be really careful, the bind mount of the `user bind source` and then use that location for the source just to ensure that it doesn't get changed in between testing and bind mounting. But I personally don't think this is worth the complication as the `user bind source` should be a trusted location by the system administrators. We might want to add a quick test to make sure that the directory/path is owned by root and error out if not, but even that might be overkill.

Thoughts?

bauerm97:
@gmkurtzer I think that loop is mostly acceptable, except for one line of code. We should not crawl directories using dirname, because it doesn't always resolve relative paths perfectly. A better solution would be to append "../" to the end of the testpath string. Something like this should work.

```
is_subdir(source, path) {
    source_st = stat(source)
    testpath = strdup(path);
    while(testpath) {
        test_st = stat(testpath);
        if ( test_st == source_st ) {
            return(TRUE);
        }
        strcat(testpath, "../");
    }
    return(-1);
}
```

After talking with my coworker I came to the conclusion that all the extra red tape that @bbockelm suggests is unnecessary. His suggestion works to prevent TOCTOU attacks or swapping out a directory for a hardlink, but it shouldn't be necessary. The MS_NOSUID bit set on the mount will prevent any SUID files (provided we remove the ability to directly mount a file, since MS_NOSUID doesn't work properly directly on a file mount according to @gmkurtzer) from making their way into the container. Since there is no user namespace isolation in the container, we don't need to worry about any other issues that could arise, such as editing a config file used by a SUID binary. Users in the container will still need the proper permissions to do anything to a file that is mounted in the container, and therefore won't be able to edit any config files for SUID programs that the user can't already edit on the host. All of this combined seems to point to us not even needing this feature for kernels without PR_SET_NO_NEW_PRIVS. It may be a nice addition for other sysadmins for whatever reason so we could still polish it up and include it.

gmkurtzer:
This won't make it into 2.3, and if we end up rebasing this to 2.4, then we are going to need to rebase and then reimplement. Closing (for now).

PR #262, opened 2016-10-24
Commented functions in config_parser.c for readability

bauerm97:
Added some pretty comments in config_parser.c so that anybody digging through the code base knows exactly how the config parsing works. I intend to add other comments as I dig more into the code, however this was a file that I spent a bit of time examining as I was working on #260


gmkurtzer:
Thank you!

PR #266, opened 2016-10-25
py2/py3 compatibility for urllib/urllib2

alaindomissy:
Documentation at http://singularity.lbl.gov/docs-exec shows running this example
$ singularity exec docker://python:latest /usr/local/bin/python hello.py
Doing so with python3 as default python interpreter brings up an error as urllib2 is python2 only:
ImportError: No module named 'urllib2'
This is a change to make the code py2/py3 compatible

### DIFF
--- a/libexec/python/utils.py
+++ b/libexec/python/utils.py
@@ -28,10 +28,13 @@ import shutil
 import subprocess
 import sys
 import tarfile
-import urllib
-import urllib2
-
-from urllib2 import HTTPError
+try:
+    from urllib.parse import urlencode
+    from urllib.request import urlopen, Request
+    from urllib.error import HTTPError
+except ImportError:
+    from urllib import urlencode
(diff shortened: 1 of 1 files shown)


gmkurtzer:
Hi @alaindomissy, this looks good to me but it is @vsoch's code so I want her thumbs up too.

Thanks!

alaindomissy:
Actually, I still have a problem with python3. I now get this error : 
Error getting pull token for repository library/python.latest, exiting.
ERROR: Aborting with RETVAL=0

vsoch:
I can fix this up this weekend - just did this exact thing for another project actually!

alaindomissy:
There was another python3 issue. It is now working for me in both py2 and py3.

vsoch:
ok, this looks good. It would be good to have some tests for this, but it doesn't hurt to fix the incompatibility even without. Thanks for your help doing this!

PR #269, opened 2016-10-27
QL: d/rules - New version needs sexec-suid set suid

rfehren:
Hi,

this PR makes the debian package work again with version 2.2

Best,

Roland


gmkurtzer:
Looks good, thanks!

Issue #267, opened 2016-10-26
ERROR  : User namespace not supported, and program not running privileged.


soichih:
I am finally testing singularity on our HTC cluster; Indiana University Karst (running on RHEL6), and I am seeing following error messages.

```
$ singularity -v shell docker://ubuntu bash
increasing verbosity level (2)
Exec'ing: /N/soft/rhel6/singularity/2.2/libexec/singularity/cli/shell.exec docker://ubuntulibrary/ubuntu:latest
Downloading layer: sha256:a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4
Downloading layer: sha256:668604fde02e75dddb4b44c80d4ce20baaac4832c41c3a945f4a236cd7d2f164
Downloading layer: sha256:2879a7ad31445fe2cea410b8ba76704003c11ee05c0a4d32d1113009ea1a1aae
Downloading layer: sha256:de413bb911fd848383ef2e5068a42c258c898d6ee869fb441fb2391eb327b576
Downloading layer: sha256:fc19d60a83f11bbddc7bd2dfca6095b49100314bfde61d83729112a6b6e11d48
Downloading layer: sha256:6bbedd9b76a496816d86a0af731ea984f40467ef8fb23be752f801cb80436ac6
VERBOSE: Set messagelevel to: 2
VERBOSE: Running NON-SUID program workflow
VERBOSE: Opening configuration file: /N/soft/rhel6/singualrity/2.2/etc/singularity/singularity.conf
VERBOSE: Not invoking SUID mode: SUID sexec not installed
ERROR  : User namespace not supported, and program not running privileged.
ABORT  : Retval = 255
VERBOSE: Cleaning sessiondir: /tmp/.singularity-session-740536.2050.10354693
VERBOSE: Cleaning run directory: /tmp/singularity-rundir.Q1oe5W65
```

Our sysadmin confirmed that the singularity was cleanly installed, and I see sexec-suid installed with SUID

```
hayashis@h1(karst):/N/soft/rhel6/singularity/2.2/libexec/singularity $ ls -lrt
total 1536
drwxr-sr-x 4 root root   4096 Oct 14 16:20 bootstrap
drwxr-sr-x 2 root root   4096 Oct 14 16:20 cli
drwxr-sr-x 2 root root   4096 Oct 14 16:20 helpers
drwxr-sr-x 3 root root   4096 Oct 14 16:20 python
[...]

PR #274, opened 2016-10-27
Added VERBOSE message to always output path of sexec_suid_path

bauerm97:
This is a pull request to add a VERBOSE message that will output the path that sexec binary is searching at for sexec-suid binary. This was brought to light in #267 and should allow easier debugging if this happens again.

Also added separate check if sexec-suid exists to distinguish between improper permissions and an uninstalled binary file.


gmkurtzer:
Looks perfect, you rock! Thanks!

Issue #279, opened 2016-10-28
[docker][python] runscript: wrong shebang + wrong encoding


flx42:
```
$ mkdir ubuntu
$ singularity import ubuntu docker://ubuntu
$ singularity run ubuntu
/.run: line 4: /singularity: cannot execute binary file: Exec format error
/.run: line 4: /singularity: Succes
```

Two problems:
- The shebang is invalid:

```
cat ubuntu/singularity 
!#/bin/sh

/bin/bash
```

Easy fix :)
https://github.com/singularityware/singularity/blob/97e965134589661ff63cd996d90d8b11d68a1b8b/libexec/python/docker/api.py#L43
- The file is in unicode, hence the `cannot execute binary file` above:

```
$ cat -e ubuntu/singularity
!^@^@^@#^@^@^@/^@^@^@b^@^@^@i^@^@^@n^@^@^@/^@^@^@s^@^@^@h^@^@^@$
^@^@^@$
^@^@^@/^@^@^@b^@^@^@i^@^@^@n^@^@^@/^@^@^@b^@^@^@a^@^@^@s^@^@^@h^@^@^@
```

One fix could be to add a `.encode('ascii')` here or earlier:
https://github.com/singularityware/singularity/blob/97e965134589661ff63cd996d90d8b11d68a1b8b/libexec/python/docker/api.py#L191

vsoch:
ok I'm walking through this now. My first instinct is that it's some weird python3 thing with bytes, so I'm testing with python3 first!

The issue doesn't seem to be where the command is returned - the manifest returns strings in json, and the thing that is returned is a string:

![image](https://cloud.githubusercontent.com/assets/814322/19824542/4865195c-9d25-11e6-9714-29445f92fa1f.png)

The next call is passing the output of that (the command) to `create_runscript`, and we still have string:

![image](https://cloud.githubusercontent.com/assets/814322/19824593/b3ed348e-9d25-11e6-9f6d-5c9db80552c6.png)

TLDR, I **think** the issue is just having the [default write mode](https://github.com/singularityware/singularity/blob/master/libexec/python/utils.py#L218) as `wb`. I changed this in my master, it seems to do the trick, do you want to give it a go and then I'll PR?  As for the shebang, I have no idea what I was thinking! !#!#!#

https://github.com/vsoch/singularity

flx42:
@vsoch: I tested your patches briefly, looks good, thanks!

vsoch:
[...]

PR #280, opened 2016-10-28
Fix runscript parsing error

vsoch:
The current runscript is being encoded incorrectly in python3, thanks to the updates with bytes. This makes the default write_file function mode "w" instead of "wb" and also fixes the incorrect shabang ordering of the runscript.


vsoch:
This will close #279

PR #282, opened 2016-10-30
Add SLURM integration plugin

bbockelm:
This PR creates a SLURM plugin.  With this installed, a user can add an `--image` argument such as:

```
srun --image /cvmfs/cms.cern.ch/rootfs/x86_64/centos7/latest/ /usr/bin/ls /opt
```

With this, `singularity` does not need to be explicitly called by the user - instead, the user can just state the intent and have the batch system figure it out.

This utilizes the "SPANK" API provided by SLURM.  SPANK allows us to call code as root, after forking but prior to `exec`ing the job -- but does not allow us to actually prefix the job with a call to `singularity` (manipulating the `argv` string).  Hence, we must do the image and namespace initialization inside the plugin instead of as part of the `singularity` executable.  While this is more complex, it _does_ eliminate the need to have a `setuid` binary.

This contains support for file- and directory-based images.  Docker can be done as part of the followup.

### DIFF
--- a/src/slurm/Makefile.am
+++ b/src/slurm/Makefile.am
@@ -1,0 +1,23 @@
+
+plugindir = $(libdir)/slurm
+
+if WITH_SLURM
+plugin_LTLIBRARIES = singularity_spank.la
+singularity_spank_la_SOURCES = singularity.c
+singularity_spank_la_LIBADD = ../lib/libsingularity_internal.la
+singularity_spank_la_LDFLAGS = -module -no-undefined -avoid-version -export-symbols-regex '^slurm_spank_|^plugin_'
+endif
+
+install-data-hook: cleanup_plugin
+
+cleanup_plugin:
+	@if test -e "$(DESTDIR)$(plugindir)/singularity_spank.so"; then \
+		echo "cp $(DESTDIR)$(plugindir)/singularity_spank.so $(DESTDIR)$(plugindir)/backup_singularity_spank"; \
+		cp "$(DESTDIR)$(plugindir)/singularity_spank.so" "$(DESTDIR)$(plugindir)/backup_singularity_spank"; \
+		echo "rm -f $(DESTDIR)$(plugindir)/singularity_spank.*"; \
+		rm -f "$(DESTDIR)$(plugindir)/"singularity_spank.*; \
+		echo "mv $(DESTDIR)$(plugindir)/backup_singularity_spank $(DESTDIR)$(plugindir)/singularity_spank.so"; \
+		mv "$(DESTDIR)$(plugindir)/backup_singularity_spank" "$(DESTDIR)$(plugindir)/singularity.so"; \
+	fi
+
+
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,7 @@ src/.libs/
 src/lib*.la
 src/*.o
 src/*.lo
+src/singularity_spank.la
 
 src/*/.deps/
 src/*/.libs/
--- a/.gitignore
+++ b/.gitignore
@@ -45,7 +45,7 @@ src/.libs/
 src/lib*.la
 src/*.o
 src/*.lo
-src/singularity_spank.la
+src/slurm/singularity_spank.la
 
 src/*/.deps/
 src/*/.libs/
--- a/configure.ac
+++ b/configure.ac
@@ -267,6 +267,7 @@ AS_IF([test "x$with_slurm" != "xno"],
     ]
 )
 AM_CONDITIONAL([WITH_SLURM], [test "$with_slurm" = "yes"])
+AC_SUBST(with_slurm)
 
 
 AC_MSG_CHECKING([if suid should be enabled])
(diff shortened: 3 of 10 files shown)


vsoch:
Awesome! I have some tight deadlines for Tues this week but I definitely will be able to take a peek after that. Thanks for doing this, it's important I think. But I already went on about that ;)

bbockelm:
Sounds great!  Note that there is a `README.md` in the added `src/slurm` sub-directory which explains how to use and configure the plugin.

I'm a total SLURM newbie, but it was pretty easy to get going even for me.

gmkurtzer:
Hi @bbockelm,

Two things.. It seems as though there is a merge conflict and can you create/add a Makefile in src/slurm/ and move the relevant Make bits there?

Thanks!

gmkurtzer:
It looks like this is has some redundant code with sexec.c. Is it possible to use this to wrap the non-SUID version of sexec instead of having the redundant bits?

bbockelm:
@gmkurtzer -

Nope, no merge conflict.  That was just a note from github that someone needs to review / +1 the code.

Moved the automake code into `src/slurm` as requested.

Unfortunately, there's no way to call out to a separate binary: the SLURM plugin invokes this after `fork`'ing a sub-process and before final environment setup and `exec`'ing.  Since we have no means to actually `exec` the final process itself, everything has to stay inside the plugin - no external executables.

gmkurtzer:
Yep, I noticed the review note after my muscle memory and pattern recognition concluded prematurely what was going on. lol...

I assume you already know, but what about using the singularity_fork_exec() function? It seems like that _must_ work? <crosses fingers>?

bbockelm:
Nah - `singularity_fork_exec` doesn't work because SLURM hasn't finished setting up the user environment yet.  So, we can't do the `exec` ourselves - SLURM must do it.

src/slurm/Makefile.am
```diff
@@ -0,0 +1,23 @@
+
+plugindir = $(libdir)/slurm
```
gmkurtzer, line 2: This plugindir seems to be dependent on the --prefix or --libdir passed to Singularity's build which could in theory clash with Slurm's own library directory. While this may not cause any specific problems, I would suggest staying out of the Slurm internal library/plugin directory. Additionally, this will have to be added to the Slurm plugin include file anyway. How about `plugindir = $(libdir)/singularity/slurm` instead?

bbockelm: I think the SLURM intent is to have third-party libraries install into their plugin directory.  This is how other plugins appear to do it, e.g., 

https://github.com/grondo/slurm-spank-plugins/blob/master/Makefile

With the plugin installed in the plugin directory, then the admin only has to specify the plugin name:

```
required singularity.so
```

as opposed to the full path.

Looking at the SLURM RPM, _only_ plugins are put into this directory: none of the core libraries or executables live here.

gmkurtzer: Oh, understood. Seeing that was the intention, is it possible to figure out what the SLURM plugin directory is so the documentation can be simplified?

bbockelm: Hm - I'm not sure I understand the question.  The SLURM plugin directory - for the SLURM RPMs - is `/usr/lib64/slurm`.

Are you suggesting that we make the SLURM plugin directory path a `configure.ac` option?

bbockelm:
@gmkurtzer - I think I addressed all the above questions in the review.  Anything else catch your eye?

Thanks!

gmkurtzer:
The rest looks good to me, I'll merge hopefully later today. Just kinda overwhelmed a bit at the moment (mostly SC stuffs) and I'll get to it shortly!

Thanks and sorry for the latency!

vsoch (changes requested):
One quick comment - I think here we would also do well to talk about a standard for "default places to look for images." For example, the user might want to cache an image, and call it by its name instead of specifying the whole path. See this other comment: https://github.com/singularityware/singularity/pull/297/files/f2f89607430733a237ed69aa5da3e9de9b161506#diff-096a3732fd022f0d09636eddf62cf716 

Would it make sense to have an address that is in reference to the user cache, akin to docker:// ? Perhaps just singularity:// would look for ~/.singularity/images/singularity, and Docker cache would be under ~/.singularity/images/docker @gmkurtzer  what do you think?

It would be great to test this out - when we do the above I'll ping some of my team to see if we can!

gmkurtzer:
@vsoch: Just to make sure I understand properly, you are asking for a URI to reference cached containers? I don't think this will invoke the image handler functions, which means we would have to rewrite the image-handler and throw it into C for inclusion of this module. Not impossible, but not something we can block on this PR for as it will require a much larger subset of changes.

vsoch:
So just to doubly confirm - we need a set of larger changes to make a quasi URI namespace to reference different images (which then modules, etc would work with). If the two things are (mostly) separate, and if we can update this addition when we have the proper URI, then definitely we don't need to block this one!

gmkurtzer:
Yep, that's what I'm thinking too. :)

PR #289, opened 2016-11-01
Better chomp

bauerm97:
Updated chomp() in util/util.c to remove all leading and trailing spaces.


gmkurtzer:
Have you removed the places where we are individually ++'ing over whitespace to be sure this is working properly?

bauerm97:
Errr no I forgot to do so. I'll go sort through them and then commit an update

gmkurtzer:
Sounds great, thanks!

PR #297, opened 2016-11-04
another script to convert docker to container

truatpasteurdotfr:
Fixes #

 -
 -
 -

@singularityware-admin

### DIFF
--- a/examples/contrib/docker-image-2-singularity-container.sh
+++ b/examples/contrib/docker-image-2-singularity-container.sh
@@ -1,6 +1,7 @@
 #!/bin/sh
 # convert docker images in singularity container
 # requirements:
+# docker and sudo privileges to run docker
 # singularity and sudo privileges to run singularity
 # 2016/11/04 Tru Huynh <tru@pasteur.fr>
 
--- a/examples/contrib/docker-image-2-singularity-container.sh
+++ b/examples/contrib/docker-image-2-singularity-container.sh
@@ -1,0 +1,48 @@
+#!/bin/sh
+# convert docker images in singularity container
+# requirements:
+# singularity and sudo privileges to run singularity
+# 2016/11/04 Tru Huynh <tru@pasteur.fr>
+
+# where the containers will be installed
+SINGULARITY_CONTAINER_DIR=~/singularity-img
+
+if [ $# -lt 1 ]; then
+echo usage: $0 docker_image_names
+echo
+exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


gmkurtzer (changes requested):
Can you integrate this into the `import` subcommand? perhaps via a URI like `local-docker://...`?

examples/contrib/docker-image-2-singularity-container.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/sh
+# convert docker images in singularity container
+# requirements:
+# docker and sudo privileges to run docker
+# singularity and sudo privileges to run singularity
+# 2016/11/04 Tru Huynh <tru@pasteur.fr>
+
+# where the containers will be installed
+SINGULARITY_CONTAINER_DIR=~/singularity-img
+
```
vsoch, line 10: We should have a more group discussion about where we plan to cache singularity containers, because there are going to be a few types!
- docker layers
- singularity image layers
- (and support for something else)

My suggestion is to have a `~/.singularity` folder, and inside we would have images, with subdirectories for each type. Otherwise, the layers/images are going to get confusing.

This structure would need to be checked for at each run of some kind of bootstrap, and additionally have options for the user to enable or disable the cache, or customize this location. That would be great to add to this PR!

vsoch (changes requested):

examples/contrib/docker-image-2-singularity-container.sh
```diff
@@ -0,0 +1,49 @@
...
+for DOCKER_IMAGES in "$@"
+do
+CONVERT_ID=${DOCKER_IMAGES}.${TEMP_TAR}
+SINGULARITY_CONTAINER_NAME=${SINGULARITY_CONTAINER_DIR}/`echo ${DOCKER_IMAGES}|tr '/:' '_-'`.img
+echo "converting docker image ${DOCKER_IMAGES} to ${SINGULARITY_CONTAINER_NAME}"
+if [ -f  ${SINGULARITY_CONTAINER_NAME} ]; then
+echo "*** ${SINGULARITY_CONTAINER_NAME} exists, aborting ***"
+echo
+else
+sudo docker pull ${DOCKER_IMAGES} && \
```
vsoch, line 29: is there a need for a check that Docker is installed first?

gmkurtzer:
I am closing this PR because we have new Docker interfaces...

PR #298, opened 2016-11-07
Unified simage binary

### LINKED ISSUE #294: [Proposal] Refactor of image-X binaries into unified simage binary
Hi all, in my never ending quest to remove all root restrictions for Singularity I've come upon a way that I think we could improve the workflow of the application (which will make it easier in the long run to implement new features). Currently the relevant .exec file in `libexec/cli/`calls upon the relevant binary. For instance, all commands that pertain with running the image call upon `sexec`. However, the commands pertaining with image modification all call upon separate binaries to do different work. Some call `image-mount`, others `image-create`, you get the idea.

I propose that we unify the image modifying binaries into one `simage` binary defined in `src/simage.c` and move image-X.c source files into a new `src/img/` directory (perhaps a better location would be in `lib/img/`, therefore giving access to image modification functions to all parts of the program). This directory can then house the code that I have been writing at `src/lib/bootstrap/` containing the bootstrap port to C. The relevant .exec files in `libexec/cli/` will simply change from:

`exec "$SINGULARITY_libexecdir/singularity/image-mount" "$SINGULARITY_libexecdir/singularity/image-bootstrap" "$DEFINITION" "$@"`

to:

`exec "$SINGULARITY_libexecdir/singularity/simage" "mount" "bootstrap" "$DEFINITION" "$@"`
(Or something similar, depending on how we finalize the implementation)

At the start `simage` will simply check argv[1] and then run the appropriate part of the code. However, because we have merged all of the binaries into one unified binary we now have a lot of flexibility in terms of modularity and portability of code. For instance, some of the non-root features require SUID and namespace features going forward. As it stands now, each binary will (likely) at some point need SUID in order for all features to work non-priv, therefore somebody would need to go in and write this code for `image-mount.c`, `image-create.c` etc.. individually. With a unified binary we can make use of the same workflow as sexec.c: write the SUID workflow once and forget about it.

I don't see any particular reasons off the top of my head why this wouldn't work, or why we shouldn't do this, however if you can think of anything I would appreciate it. I am willing to do this since it's basically an idea to help my other ideas easier to implement, but if anybody else wants to tackle this feel free. Please let me know any thoughts.

bauerm97:
Fixes #294 

As discussed with @gmkurtzer I have implemented the merge of all `image-*` binaries into a unified `simage` binary. Currently the simage-suid binary is commented out from `src/Makefile.am` as none of the workflows can properly take advantage of SUID features yet. This is a precursor to a wider refactoring, where we can better take advantage of SUID functionality. Please note that there are no changes to functionality present, therefore no documentation or test updates are required.

Please let me know if there is a better solution to `src/lib/image/image.c` being necessary to build `libimage.la`.

 - Merged `image-*` binaries into `simage` binary
 - Created `src/simage.c` to build `simage` binary
 - Moved `image-*.c` files into `lib/image/*/*.c`
 - Updated `configure.ac`, `src/Makefile.am`, and `src/lib/Makefile.am` to reflect these changes
 - Updated `libexec/cli/*.exec` to call `simage *` instead of `image-*`

@singularityware-admin


gmkurtzer:
I will check in more detail as soon as I can, but has all of the previous functionality been migrated (e.g. does `make test` work)?

bauerm97:
Yes, all functionality is preserved, and all tests pass. I wrote simage.c and lib/image in such a way that all functionality remains unchanged, the workflow simply passes through simage binary. This gives us a centralized point to make changes to all workflows at once. Going forward this makes it easier to implement new features into the codebase that otherwise might take extra work or require us repeating code.

gmkurtzer:
Ahh, very cool. I am looking forward to checking out what you did in much more depth.

Thanks!

PR #306, opened 2016-11-10
Removed assumption that certain programs are located in /bin/ for NixOS

### LINKED ISSUE #300: Singularity assumes existance of coreutils in /bin
Singularity assumes that coreutils exists at /bin. Here's an example:

    $ singularity create -s 100 -F myimg.img 
    /nix/store/7md6r7n4c89ql6vxjrg7n300sjcadi5n-singularity-2.2/libexec/singularity/cli/create.exec: line 95: /bin/rm: No such file or directory
    Creating a new image with a maximum size of 100MiB...
    ERROR  : Will not overwrite existing file: myimg.img
    ABORT  : Retval = 255
    ERROR: Failed creating image: myimg.img

rm is available in my path, it just doesn't exist in /bin. It would be better if singularity did not assume rm existed at /bin/rm, and even better if it unlinked files using unlink(2) instead of executing another process.

bauerm97:
Fixes #300 

 - Certain scripts in `libexec/*` assumed that programs would be located in `/bin/`. These assumptions are removed

@singularityware-admin

### DIFF
--- a/libexec/cli/create.exec
+++ b/libexec/cli/create.exec
@@ -92,7 +92,7 @@ fi
 if [ -f "$SINGULARITY_IMAGE" ]; then
     if [ -n "${OVERWRITE:-}" ]; then
         message 2 "Removing existing file\n"
-        /bin/rm -f "$SINGULARITY_IMAGE"
+        rm -f "$SINGULARITY_IMAGE"
     else
         message ERROR "Image file exists, not overwriting.\n"
         exit 1
--- a/libexec/functions
+++ b/libexec/functions
@@ -22,12 +22,12 @@
 set -u
 
 if [ -z "${SINGULARITY_libexecdir:-}" ]; then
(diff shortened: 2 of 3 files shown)


vsoch (approved):
This looks good to me - should be a good workaround for NixOS just while the code is being converted.  This only wouldn't work if there was an original reason to target the specific executables in /bin/ - @gmkurtzer?

bauerm97:
@vsoch I'm not sure why these assumptions were made in the first place. If you take a look at `libexec/functions`, you'll notice that there are 2 calls to /bin/echo and many calls to echo. So, this just unifies the styling I guess?

vsoch:
Probably just random :) @gmkurtzer will likely confirm and merge soon, he might be traveling. Thanks for your help with fixing this up! :)

gmkurtzer:
I am a little worried about `echo` as we may get unpredictable results (e.g. bash's internal version of `echo` getting used instead), but only testing will confirm.

PR #310, opened 2016-11-10
Fix unbound variable in image import.

markeganfuller:

 - Fix unbound variable in image import.

@singularityware-admin

When importing via a file path and not specifying `file:///` as the
protocol the `LOCAL_FILE` variable is not defined resulting in an
`unbound variable` error.

This replaces the use of `LOCAL_FILE` with `IMPORT_URI`.


gmkurtzer:
Looks good to me, thank you!

PR #311, opened 2016-11-11
Adding proper python logging

vsoch:
Fixes #304 

Changes proposed in this pull request

 - Addition of logging module to report logging

@gmkurtzer - can we chat about how the MESSAGELEVEL variable is passed around? Currently, my logger is spitting out everything regardless of the --debug or --quiet commands, and I can't figure out why. I will look at this again tomorrow with fresh eyes, but any information about what might be up would be super helpful (and anyone else that could take a look - would appreciate greatly!) I just stumbled on python logging last weekend. :)

Issue #283, opened 2016-10-30
Cache Docker images in a reasonable manner


bbockelm:
We really need a place to cache generated Docker image before the first person tries to utilize the Docker image URLs in a large workflow (and ends up pulling the same image 1000 times for 1000 jobs).

Here's what I'm thinking:
1.  Cache image `foo:bar` in `~/.singularity/docker_cache/foo:bar.img`.
2.  If image is greater than X hours old (perhaps 48 by default, configurable), then `singularity` will regenerate the image.
3.  If `singularity` decides to update the image, it takes a POSIX lock on the file.
4.  Once the lock is acquired, double-check the timestamp on the image file.
5.  If the timestamp is still out-of-date, then generate a new image to `foo:bar.img.tmpXXXXX` (a file made with `mkstemp`.  Once complete, rename to the permanent filename.

I think this can all go into the python code; for reference, here's an example I've done previously in using POSIX locks in python: https://github.com/opensciencegrid/gratia-probe/blob/master/common/gratia/common/GratiaWrapper.py#L69

Besides the normal `singularity` client, my hope would be then to integrate this into the SLURM plugin.  This way, the Docker file would be built at submit time client-side and all nodes in the job will already have the pre-made binary.

vsoch:
@gmkurtzer how does this sound to you? We could have a global cache with subdirectories, one of them being Docker as @bbockelm has suggested! I could write the python code to handle the Docker cache, and then (other kinds) could be written to where appropriate!

A few things to talk about:
- if the user is generating a Singularity image via a Docker bootstrap (meaning some custom install or runscript) then running the command to produce an image..
  - 1. with the same name: would not overwrite even with changes if < 48 hours, which would be very confusing
  - 2. with a different name: would generate something new in the cache that might be exactly the same

so we have to decide what is the true "unique id" and how to make sure usage with cache does not mislead the user about the image being used.

What would be most appropriate I think is to support what we had chat about, some kind of singularity "overlay" or "import" so that whenever the Docker registry pulls layers, a base image is created (with the Docker layers, says user custom inputs) and then that image is used/bootstrapped from the cache. This means that the additional commands (runscript and installs) would never be cached, unless they are part of the original image being bootstrapped. Thoughts?

vsoch:
And in the future when we have singularity hub, the additional commands (runscript and installs) would be cached in a .singularity/singularity_cache as they are pulled from singularity hub, thought of akin to Docker images.

bbockelm:
[...]

PR #314, opened 2016-11-14
Add/caching and dockerfile

### LINKED ISSUE #312: Conversion of Dockerfile to Singularity
We should be able to bootstrap going from a Dockerfile immediately to Singularity, and this seems logical by first converting the Dockerfile to Singularity, and then executing as normal. Likely this "sniffing" of the filename will need to happen in the core bootstrap code (which is being converted to C I believe) however I will write a python-based conversion to either support this, or provide the functionality from other python stuffs.

vsoch:
Fixes #312 and #283 

- The first part of this PR is adding a cache for docker layers (see details below)
- The second component is adding a python module to convert a Dockerfile to Singularity file (or string), as promised. The use case for this is when a user wants to generate a Singularity image from a Dockerfile they have locally, as not all images are available on the hub (eg, those in progress, private). See details below.

## Cache for Docker Layers
Currently, when a Docker image is bootstrapped, meaning layers are pulled from the registry and dumped into an image, we download and then remove them. This works fine, but the software should have the ability to cache layers, to either save time, or some other context where the download isn't available. To this end, to the python code I have added the following functionality:

- The environment variable `SINGULARITY_CACHE` determines the base of the cache directory.
- If not defined, the default (set in [defaults.py](https://github.com/vsoch/singularity/blob/add/caching-and-dockerfile/libexec/python/defaults.py#L28) is to use a hidden folder called `.singularity` in the user's home directory. 

### Notes on the default `SINGULARITY_CACHE`
I must be careful with the term USER, as when it is run by root that becomes root's home, and the user doesn't actually see it. This is why this variable should likely be defined/exported as the true user's home (not sure how, but it would need to happen before the python script is called) OR we set the default to be somewhere else). The problems I see with having the cache be in some install folder is that it can be lost given a new install, but maybe this isn't an issue. Definitely open (and hoping for) suggestions here!

- Next, with anticipation of wanting to cache other kinds of image layers and things, the function to [get_cache](https://github.com/vsoch/singularity/blob/add/caching-and-dockerfile/libexec/python/utils.py#L194) takes an optional subfolder, and it is defined to be "docker" in the command line client that handles the operation:

        cache_base = get_cache(subfolder="docker", 
                               disable_cache = args.disable_cache)

### Ability to disable cache
Finally, you will notice the `disable_cache` variable, which comes in as a command line argument for the client script. The default is set to False, and I would suggest we also set a default somewhere in the core software that can be tweaked by the user/admin setup. This is currently [called via bash](https://github.com/vsoch/singularity/blob/add/caching-and-dockerfile/libexec/bootstrap/modules-v2/build-docker.sh#L119) in the `build-docker.sh` script, and given this current setup, would be passed as an environmental / command line variable. Again, given that someone is transitioning the entirety of bootstrap to C, I didn't touch this part. When this is parsed out, let me know and the python portion can be adjusted appropriately.

## Dockerfile to Singularity file
This part of the PR will add a new module file, [converter.py](https://github.com/vsoch/singularity/blob/add/caching-and-dockerfile/libexec/python/docker/converter.py), to the python/docker folder. It currently is not "plugged in" anywhere to the singularity core software, as it is my understanding that the bootstrap is being converted to C, however this PR will add the functionality to be available to the core when it's ready. This will need a lot of testing with different Dockerfiles, because when you've seen one Dockerfile... you've seen one Dockerfile :) I went off of their [standard spec here](https://docs.docker.com/engine/reference/builder/) and [have provided here](https://gist.github.com/vsoch/4192e93b0597bee9610ece87e42edf5d) example Dockerfile and the generated Singularity file, along with how to test the code:

      # How to test Dockerfile --> Singularity generation
      # checkout the caching-and-dockerfile branch from https://github.com/vsoch/singularity
      # cd singularity/libexec/python/docker

      # Here in python

      from converter import dockerfile_to_singularity  
      import tempfile

      # You can specify an output directory, to save a file called "Singularity"
      output_dir = tempfile.mkdtemp() # (optional)
      dockerfile = "/home/vanessa/Documents/Dropbox/Code/singularity/singularity-hub/Dockerfile"
      singularity_spec = dockerfile_to_singularity(dockerfile_path=dockerfile,
                                                                           output_dir=output_dir)

      ## -- End pasted text --
      # Singularity spec written to /tmp/tmpe7gnhe9k
 
## Next steps
Given that the second part of the PR is not integrated into the software, we likely don't need extensive testing until it's added and available on the command line. If others want / are able to do testing like the above, it would be greatly appreciated, but no harm will come if it's not perfectly suited for every use case yet. 

The first part of the PR, the cache, is very important so that I can next add an endpoint for Singularity Hub, and those images to be cached. Additionally, this PR will need to be updated with logging, when #311 is merged. Thanks in advance for good discussion and feedback! Save me some stickers at SC :)

@singularityware-admin

### DIFF
--- a/libexec/python/cli.py
+++ b/libexec/python/cli.py
@@ -25,7 +25,7 @@ perform publicly and display publicly, and to permit other to do so.
 '''
 
 from docker.api import get_layer, create_runscript, get_manifest, get_config, get_images
-from utils import extract_tar, change_permissions
+from utils import extract_tar, change_permissions, get_cache
 import argparse
 import os
 import re
@@ -70,6 +70,14 @@ def main():
                         dest='notoken', 
                         action="store_true",
                         help="boolean to specify that the CMD should be included as a runscript (default is not included)", 
+                        default=False)
+
+
+    # Flag to disable cache
+    parser.add_argument("--no-cache", 
+                        dest='disable_cache', 
+                        action="store_true",
+                        help="boolean to specify disabling the cache.", 
                         default=False)
 
     
@@ -154,26 +162,32 @@ def main():
 #  DOWNLOAD LAYERS -------------------------------------------
 # Each is a .tar.gz file, obtained from registry with curl
 
-        # Create a temporary directory for targzs
-        tmpdir = tempfile.mkdtemp()
+        # Get the cache (or temporary one) for docker
+        cache_base = get_cache(subfolder="docker", 
+                               disable_cache = args.disable_cache)
+
         layers = []
 
         for image_id in images:
 
-            # Download the layer
-            targz = get_layer(image_id=image_id,
-                              namespace=namespace,
-                              repo_name=repo_name,
-                              download_folder=tmpdir,
-                              registry=args.registry,
-                              auth=doauth) 
+            # Download the layer, if we don't have it
+            targz = "%s/%s.tar.gz" %(cache_base,image_id)
+ 
+            if not os.path.exists(targz):
+                targz = get_layer(image_id=image_id,
+                                  namespace=namespace,
+                                  repo_name=repo_name,
+                                  download_folder=cache_base,
+                                  registry=args.registry,
+                                  auth=doauth) 
 
             layers.append(targz) # in case we want a list at the end
                                  # @chrisfilo suggestion to try compiling into one tar.gz
 
             # Extract image and remove tar
             extract_tar(targz,singularity_rootfs)
-            os.remove(targz)
+            if args.disable_cache == True:
+                os.remove(targz)
                
      
     # If the user wants to include the CMD as runscript, generate it here
@@ -190,8 +204,9 @@ def main():
             # change permission of runscript to 0755 (default)
             change_permissions("%s/singularity" %(singularity_rootfs))
 
-    # When we finish, change permissions for the entire thing
-    #change_permissions("%s/" %(singularity_rootfs))
+    # When we finish, clean up images
+    if args.disable_cache == True:
+        shutil.rmtree(cache_base)
 
 
 if __name__ == '__main__':
--- a/libexec/python/cli.py
+++ b/libexec/python/cli.py
@@ -91,7 +91,7 @@ def main():
     if args.rootfs != None:
        singularity_rootfs = args.rootfs
     else:
-       singularity_rootfs = os.environ.get("SINGULARITY_ROOTFS",None)
+       singularity_rootfs = os.environ.get("SINGULARITY_ROOTFS", None)
        if singularity_rootfs == None:
            print("ERROR: root file system not specified or defined as environmental variable, exiting!")
            sys.exit(1)
--- a/libexec/python/Makefile.am
+++ b/libexec/python/Makefile.am
@@ -2,7 +2,7 @@ SUBDIRS = docker
 
 scriptlibexecdir = $(libexecdir)/singularity/python
 
-dist_scriptlibexec_SCRIPTS = cli.py __init__.py utils.py
+dist_scriptlibexec_SCRIPTS = cli.py __init__.py utils.py defaults.py
 
 MAINTAINERCLEANFILES = Makefile.in
 
--- a/libexec/python/defaults.py
+++ b/libexec/python/defaults.py
@@ -1,0 +1,28 @@
+#!/usr/bin/env python
+
+'''
+defaults.py: python helper for singularity command line tool. This script stores
+default variables for core python modules for singularity, akin to __init__.py.
+
+Copyright (c) 2016, Vanessa Sochat. All rights reserved. 
+
+"Singularity" Copyright (c) 2016, The Regents of the University of California,
+through Lawrence Berkeley National Laboratory (subject to receipt of any
+required approvals from the U.S. Dept. of Energy).  All rights reserved.
+ 
+This software is licensed under a customized 3-clause BSD license.  Please
+consult LICENSE file distributed with the sources of this project regarding
+your rights to use or distribute this software.
+ 
+NOTICE.  This Software was developed under funding from the U.S. Department of
+Energy and the U.S. Government consequently retains certain rights. As such,
+the U.S. Government has been granted for itself and others acting on its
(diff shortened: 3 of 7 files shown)


bauerm97:
@vsoch thank you for implementing the docker file conversion! It should be very easy to implement this now

vsoch:
Sure thing! When the rest is ready, I'd like to do more testing. For example, I anticipate a bug given a case where you have a line with a list (for RUN or CMD) followed by a comment of some sort, because it does a try to read as json, and a catch to just treat the entire string as the command. We can easily address this by stripping comments first, of course, I just didn't do this out proper (and looking over Dockerfiles,it's not a hugely common thing because people tend to write comments above lines, and not use the list syntax.

Anyhoo, let me know when we are ready for some more tests / tweaking!

gmkurtzer:
I think the logger PR caused conflicts in this PR. Can you update this PR?

Thanks!

vsoch:
Yep, I expected that, will do this right away! So great to see all these Merging emails! vroomvroom!

vsoch:
ok just a heads up - I'm first going to update the PR with JUST fixes to the merge conflict. But don't merge that - I want to also add logging for the (now merged) caching stuffs.

vsoch:
btw - @gmkurtzer is `SINGULARITY_CACHE` an appropriate environmental variable to specify the cache? We should discuss this - I noticed that before we get to the python functions, the script looks for [SINGULARITY_CACHEDIR](https://github.com/singularityware/singularity/blob/765522d36cbf526b1aee3823316b60a3904af752/libexec/image-handler.sh#L54) and sets it to be `/tmp` if it's not defined. This of course means it's not really cached in the way you think about cache, something that endures after you restart the system. This is why I had chosen `~/.singularity` for the `SINGULARITY_CACHE`, because it wouldn't be cleared / cleaned up. It comes down to the question - can we dump docker layers in a subdirectory called docker under whatever folder the user has defined to be the cache?

I think likely the fix is to:

## If we can share the same cache:
- remove the SINGULARITY_CACHE variable and have both the python and bash/C functions use SINGULARITY_CACHEDIR. 
- Instead of the python script defaulting to `~/.singularity` if nothing is provided, just make a temporary directory (as is done now). 

## If we can't share the cache
We need to clearly define the difference between the two caches (eg, one is for saving image layers, the other is for building the image itself).

I sort of prefer the second solution, because I think there is something fundamentally different about a place to keep finished images, and a place to build them. We would just need to make sure it's clearly documented / defined as to not be confusing. What do you think?

gmkurtzer:
I think your plan on using `SINGULARITY_CACHEDIR` is fine, and defaulting to ~/.singularity. We just need a way of separating out docker caches from image caches, etc. and then leveraging some commonalities between them (e.g. easily achieved with directory namespaces).

In a nutshell, yes... Share the cache, and lets work towards a unified caching system! :D

vsoch:
werd! Will do.

vsoch:
ok let me look over the comparison on here - want to make sure I didn't miss anything...

libexec/python/cli.py
```diff
@@ -77,6 +76,14 @@ def main():
...
                         default=False)
 
+
+    # Flag to disable cache
+    parser.add_argument("--no-cache", 
+                        dest='disable_cache', 
+                        action="store_true",
+                        help="boolean to specify disabling the cache.", 
+                        default=False)
+
```
vsoch, line 86: ok - here is something to discuss! This command basically means that the layers that are downloaded are removed after running (with shutil.rmtree(cache_base)). This is different from the scenario when a user doesn't specify a cache, and then it goes to /tmp (and the docker portion doesn't clean it up). If the post python portion cleans up the files by removing the directory, then we are ok. If not, or if there is doubt, I think the base software should set this --no-cache variable on the getgo, that way the python always is in charge of making sure the files are cleaned up. 

In the current implementation, If the pre-docker portion always exports SOMETHING as SINGULARITY_CACHEDIR, this command isn't useful, because it would find that variable and use /tmp. However, given that this env variable is defined, this means that those files aren't cleaned up after.  So it comes down to this question:

- Does the user not specifying an environmental variable for the cache mean he/she wants to save the layers in /tmp? If so, we leave as is
- Does the user not specifying a cache mean he/she would prefer them to be pulled and then deleted? If so, we probably want the calling functions to add `--no-cache`.

gmkurtzer:
Done. I will work on the image-handler.sh :)

Thanks!

vsoch:
Thank you!!!

stevekm:
Is this "Dockerfile to Singularity file" conversion feature implemented somewhere in Singularity now? Haven't seen it documented anywhere else, and not sure how to run it from the details here.

vsoch:
hey @stevekm ! You actually can use docker natively with Singularity now, eg:

```
singularity pull docker://ubuntu
```
or

```
sudo singularity build image.simg docker://ubuntu
```

stevekm:
@vsoch thanks I have seen that documented. However, I think those commands require you to have built a Docker image already or have the Docker container already built on Docker Hub? I am trying to build a Singularity container directly from a Dockerfile, as described in this PR, since that is what many workflows ship with.

vsoch:
Yep, you have a few options then!

 1. Host it on Docker hub (you would create the repository then do docker push username/repository --tag versionxxx
2. Run a local docker registry (https://blog.docker.com/2013/07/how-to-use-your-own-registry/) and then pull the singularity image from there, note you would need to set the variable `SINGULARITY_NOHTTPS` eg:

```
SINGULARITY_NOHTTPS=yes singularity pull docker://localhost:5000/username/repository
``` 

 3. or use the docker2singularity image --> https://github.com/singularityware/docker2singularity (and note the 2.4 branch will give a squashfs 2.4 image, the others are the old ext3 filesystem) to (after you build an image and have it locally) generate your image.

stevekm:
Thanks for the help, option 3 was what I was looking for, I put together a demo workflow here: https://github.com/stevekm/docker2singularity-demo

vsoch:
Awesome! Would you mind if I added a link to your demo in the docker2singularity repository? It might be useful to others that want to do the same.

stevekm:
Sure of course!

PR #317, opened 2016-11-19
fixing docker command and runscript specification

### LINKED ISSUE #316: Runscript: doesn't seem to be obtained from Docker image
need to test this further, it doesn't appear to be working as expected.

vsoch:
Fixes #316 

 - Makes the default behavior (if nothing specified) to use ENTRYPOINT as an executable. There is a weird distinction between CMD and ENTRYPOINT, but [I believe Entrypoint](http://stackoverflow.com/questions/21553353/what-is-the-difference-between-cmd-and-entrypoint-in-a-dockerfile) is more suited to the executable, and should be the default. If the user specifies IncludeCmd then it changes to use the CMD from in the manifest.
 - The runscript message was printing to save, but wasn't actually saving to the container. The reason was because the code in postbootstrap.sh dumped the output from trying to get the runscript from the boostrap spec (even if it was empty) into a singularity file, and then deleted the file if it was empty. The behavior should be:

1. if there is a runscript defined in the bootstrap, write it to file. This takes priority over Docker, etc.
2. if there is no runscript defined in the bootstrap spec, the default uses the Dockerfile ENTRYPOINT
3. if the user specifies includeCmd, then use the CMD argument instead.

- I also removed unneeded changing of permissions in the docker cli.py, as the runscript permissions are handled by postbootstrap after.
- The original script to create_runscript originally appended the /bin/sh as the executor, but we should not bias, and so I removed this so it looks like:

     exec /usr/local/run.sh "$@"

The above is basically the command from the runscript, prefixed by exec, followed by "$@"  to capture any other input arguments from the user.

@singularityware-admin

### DIFF
--- a/libexec/bootstrap/modules-v2/postbootstrap.sh
+++ b/libexec/bootstrap/modules-v2/postbootstrap.sh
@@ -158,7 +158,7 @@ if [ -f "$SINGULARITY_BUILDDEF" ]; then
 
     # If the command is not empty, write to file.
     if [ ! -z "$runscript_command" ]; then
-        cat $runscript_command > "$SINGULARITY_ROOTFS/singularity"    
+        cat "$runscript_command" > "$SINGULARITY_ROOTFS/singularity"    
     fi
 
     # If we have a runscript, whether docker, user defined, change permissions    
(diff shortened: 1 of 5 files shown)


gmkurtzer:
This PR conflicts with the logger update that I already implemented. Can you update this PR?

Thanks!

vsoch:
I'll wait to do this one until the last (for the dockerfile and cache) is merged. This one is super important, I should have done it first! /bonk

gmkurtzer:
Freakin' perfect!

PR #318, opened 2016-11-21
Bind mount devices instead of using mknod

flx42:
There are multiple issues with the option `dev=minimal`.

Permissions are not right, you can't use `/dev/null`:
```
$ singularity shell ubuntu
Singularity.ubuntu> echo foo > /dev/null
/bin/sh: 6: cannot create /dev/null: Permission denied
Singularity.ubuntu> ls -l /dev/null
crw-rw-r-- 1 root root 1, 3 Nov 21 21:26 /dev/null
```
The permissions passed to `mknod` are [correct](https://github.com/singularityware/singularity/blob/3a46a326e006b39afa711c38f031413e62a693d1/src/lib/mount/dev/dev.c#L57), but it doesn't take `umask` into account, which is often `0002` or `0022` by default.

`dev=minimal` can't be combined with `setuid=no`, mknod is not allowed in user namespace:
```
singularity -vvvv shell ubuntu
[...]
DEBUG   [U=1000,P=31830]   privilege.c:142:singularity_priv_escalate(): Not escalating privileges, user namespace enabled
DEBUG   [U=1000,P=31830]   dev.c:55:singularity_mount_dev()           : Checking container's /dev/null
VERBOSE [U=1000,P=31830]   dev.c:58:singularity_mount_dev()           : Can not create /dev/null: Operation not permitted
[...]
Singularity.ubuntu> ls -l /dev/
total 0
```

Bind-mounting the devices is a solution that should always work.
With the patch and `dev=minimal setuid=no`
```
$ singularity shell ubuntu
Singularity.ubuntu> ls -l /dev/
total 0
crw-rw-rw- 1 nobody nogroup 1, 3 Nov 18 22:57 null
crw-rw-rw- 1 nobody nogroup 1, 8 Nov 18 22:57 random
crw-rw-rw- 1 nobody nogroup 1, 9 Nov 18 22:57 urandom
crw-rw-rw- 1 nobody nogroup 1, 5 Nov 18 22:57 zero
```

I admit the empty `fileput` is a bit weird, but I don't think we have a function to `touch` a file.

@singularityware-admin

### DIFF
--- a/src/lib/mount/dev/dev.c
+++ b/src/lib/mount/dev/dev.c
@@ -36,6 +36,28 @@
 #include "lib/ns/ns.h"
 #include "lib/rootfs/rootfs.h"
 
+static int mount_dev(const char *dev) {
+    char *container_dir = singularity_rootfs_dir();
+    char *path = joinpath(container_dir, dev);
+
+    singularity_message(DEBUG, "Mounting device %s at %s\n", dev, path);
+
+    if ( is_chr(path) == 0 ) {
+        return(0);
+    }
+
+    if ( fileput(path, "") < 0 ) {
+        singularity_message(VERBOSE, "Can not create %s: %s\n", dev, strerror(errno));
+        return(-1);
+    }
+
+    if ( mount(dev, path, NULL, MS_BIND, NULL) < 0 ) {
+        unlink(path);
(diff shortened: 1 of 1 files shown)


gmkurtzer:
Looks good, thanks!

PR #321, opened 2016-11-22
New feature: let singularity execute containers not directly readable by calling user, uses privileged fopen

serious-steve:
Fixes #

 - new cfg option to enable/disable privileged fopens in image.c
 - new cfg options to define container groups (groups of containers singularity can execute via privileged fopen)
 - new behaviour is as follows:
1.) try to open container.image as calling user (needs 'r' permission)
2.) if failed: check if privileged fopen is allowed
if yes: 
 get fh
 get stat(fh)
 if user(fh) has 'r' or 'x' permission OR
 if group(fh) == group(calling_user) && group has 'r' or 'x' permission OR
 if other(fh) has 'x' permission OR
 if (group(fh) is in a group enabled by 'group container' && group has 'r' or 'x' permission
 SUCCEED

else FAIL

Please tell me if you accept the implementation and if not what needs to be fixed in order to be acceptable. As soon as you give your okay I will write docs/tests.
Thanks: )

@singularityware-admin

### DIFF
--- a/etc/singularity.conf
+++ b/etc/singularity.conf
@@ -17,6 +17,11 @@ allow setuid = yes
 # Should we allow users to request the PID namespace?
 allow pid ns = yes
 
+# ALLOW PRIVILEGED FOPEN: [BOOL]
+# DEFAULT: no
+# Should singularity try to fopen container images in privileged mode
+# if the calling user does not have sufficient permissions?
+allow privileged fopen = no
 
 # ENABLE OVERLAY: [BOOL]
 # DEFAULT: yes
@@ -140,4 +145,11 @@ container dir = /var/singularity/mnt
 # /tmp/ does not work for your system, /var/singularity/sessions maybe a
 # better option.
 #sessiondir prefix = /var/singularity/sessions/
+
+# CONTAINER GROUP: [STRING]
+# DEFAULT: Undefined
+# Specifies a group a container must belong to in order to be executable by singularity.
+# The calling user is not required to belong to that group but the container must have the group executable or group readable permission set.
+# This feature requires 'allow privileged fopen' to be set.
+#container group = singularity
 
--- a/src/lib/rootfs/image/image.c
+++ b/src/lib/rootfs/image/image.c
@@ -87,6 +87,12 @@ int rootfs_image_init(char *source, char *mount_dir) {
         if ( ( image_fp = fopen(source, "r") ) != NULL ) { // Flawfinder: ignore
             singularity_message(VERBOSE, "Opened image (read only, without privileges) %s\n", source);
         } else {
+            singularity_config_rewind();
+            if ( singularity_config_get_bool("allow privileged fopen", 0) == 0 ) {
+                singularity_message(ERROR, "Privileged fopen is not enabled in config\n" );
+                ABORT(255);
+            }
+
             singularity_priv_escalate();
             if ( ( image_fp = fopen(source, "r") ) == NULL ) { // Flawfinder: ignore
 	            singularity_message(ERROR, "Could not open image (read only, with privileges) %s: %s\n", source, strerror(errno));
--- a/src/lib/rootfs/image/image.c
+++ b/src/lib/rootfs/image/image.c
@@ -28,6 +28,7 @@
 #include <sys/mount.h>
 #include <unistd.h>
 #include <stdlib.h>
+#include <grp.h>
 
 #include "util/file.h"
 #include "util/util.h"
@@ -50,6 +51,10 @@ static int read_write = 0;
 
 int rootfs_image_init(char *source, char *mount_dir) {
     singularity_message(DEBUG, "Inializing container rootfs image subsystem\n");
+    struct group *file_grp;
+    struct group *config_grp;
+    struct stat image_stat;
+    char *config_gname;
 
     if ( image_fp != NULL ) {
         singularity_message(WARNING, "Called image_open, but image already open!\n");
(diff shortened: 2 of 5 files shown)


bauerm97:
Reading through your logic here, I'm not sure why we need the config option for allow privileged fopen. I think this logic will be enough:

1) Privileged fopen
2) Check perms on fstat from fopen
3) If user already has appropriate permissions
4) If user doesn't, then check if container is in the group

This seems to me to be more simple and a bit easier to implement.Thoughts @gmkurtzer @c1t4r ?

serious-steve:
@bauerm97 While I agree with your suggestion I remember @gmkurtzer did not like a change in default behaviour.
I agree that if we do not mind a privileged fopen in either case the bool option can be omitted.
So... are there no drawbacks on privileged fopen vs unprivileged?

gmkurtzer:
I like having the boolean toggle for this feature in the configuration file, but I don't like statically defining a single group for all containers to bypass POSIX ACLs.  Perhaps just make sure that the user is a member of the container's group? Thoughts or other options?

serious-steve:
@gmkurtzer what you suggest is covered by my implementation, too.
If I was asked to make a choice I'd vote for the possibility to execute containers with 'x' permissions (either 'others' or 'group' && calling user is in 'group')
It can be, however, inconvenient to add users to groups for any SW suite they are entitled to use.
That is why I added the 'container group' way as well.

I'm fairly new to github collaboration. 
How is the workflow? 
Do I commit further changes to the branch of the PR?
@bauerm97  @gmkurtzer I would like to come up with a consensus soon, please advise me what in your opinion needs to be done to finalize the PR.

vsoch:
@c1t4r any pushes that you make to your branch (c1t4r:group-containers-3rdtry) will update the PR.

vsoch:
Could you explain the use case for this? If the calling use doesn't have root permissions, why would we want to potentially give them? If we want to have shared permissions for a group, it's probably not going to be root (given a cluster HPC environment) and couldn't we just treat the container as a file, and manage permissions like that? I don't think I fully understand the use case, and it's making me a bit nervous. For example, if a container on a cluster falls within some group, and then a user removes it and makes changes, and puts it back, couldn't it potentially do harm? If there is a group that singularity will give root access to run containers, they essentially get it by way of the container.

serious-steve:
@vsoch Thanks for asking...
The use case is to allow users execute singularity containers on which they have 'x' permissions but do not have 'r' permissions on. This should protect licensed containers from being copied.

Privileges are being dropped after 'fopen' .
We are not giving root access to the calling user, we just allow to fopen the container image in cases where standard permissions will not allow read access. Correct me if I oversee a flaw for I am no security expert :)

vsoch:
ah ok, understood. Thanks for the explanation @c1t4r  - I'm no security expert either, just aware and overly concerned about these kind of things, haha.

src/lib/rootfs/image/image.c
```diff
@@ -79,9 +84,64 @@ int rootfs_image_init(char *source, char *mount_dir) {
...
+            if ( fstat(fileno(image_fp), &image_stat) != 0 ) {
+                singularity_message(ERROR, "Could not obtain stat on image %s: %s\n", source, strerror(errno));
+                ABORT(255);
+            }
+
+            if ( ( file_grp = getgrgid(image_stat.st_gid) ) == NULL ) {
+                singularity_message(ERROR, "Could not obtain gid of image %s: %s\n", source, strerror(errno));
+                ABORT(255);
+            }
+            singularity_priv_drop();
```
bbockelm, line 113: Please drop privilege around `L101`; it is not needed for `fstat` or `getgrgid`.  As the latter is pretty complex code inside glibc, I'd rather do it unprivileged.

src/lib/rootfs/image/image.c
```diff
@@ -79,9 +84,64 @@ int rootfs_image_init(char *source, char *mount_dir) {
...
+
+            singularity_config_rewind();
+            do {
+                config_gname = singularity_config_get_value("container group");
+
+                if ( ( has_perm(4, image_stat) == 0 ) || ( has_perm(1, image_stat) == 0 ) ) {
```
bbockelm, line 119: Can you add a comment here as to what is being tested?  It's not clear what `4` and `1` are... if they are modes, maybe use the equivalent mode macros?

etc/singularity.conf
```diff
@@ -17,6 +17,11 @@ allow setuid = yes
 # Should we allow users to request the PID namespace?
 allow pid ns = yes
 
+# ALLOW PRIVILEGED FOPEN: [BOOL]
+# DEFAULT: no
+# Should singularity try to fopen container images in privileged mode
+# if the calling user does not have sufficient permissions?
```
bbockelm, line 23: Since this is a pretty scary comment, can you please elaborate more about what protections are still in place?

src/lib/rootfs/image/image.c
```diff
@@ -79,9 +84,64 @@ int rootfs_image_init(char *source, char *mount_dir) {
...
+            }
+
+            singularity_message(VERBOSE, "Opened image (read only, with privileges) %s\n", source);
+
+            if ( fstat(fileno(image_fp), &image_stat) != 0 ) {
+                singularity_message(ERROR, "Could not obtain stat on image %s: %s\n", source, strerror(errno));
+                ABORT(255);
+            }
+
+            if ( ( file_grp = getgrgid(image_stat.st_gid) ) == NULL ) {
```
bbockelm, line 109: What's the point of `file_grp`?  The only thing it appears used for is to compare against the user's GID, which we could use directly from `image_stat.st_gid`.

src/lib/rootfs/image/image.c
```diff
@@ -79,9 +84,64 @@ int rootfs_image_init(char *source, char *mount_dir) {
...
+                        singularity_message(ERROR, "Calling user does not have proper permissions to access image, aborting...\n");
+                        ABORT(255);
+                    }
+
+                    if ( (config_grp = getgrnam(config_gname)) == NULL ) {
+                        singularity_message(WARNING, "Unusable container group %s\n", config_gname);
+                        continue;
+                    }
+                    if ( config_grp->gr_gid == file_grp->gr_gid )  {
+                        if ( (( image_stat.st_mode & 11 )== 0) && ((image_stat.st_mode & 44 ) == 0) ) {
```
bbockelm, line 133: Use existing macros or a comment to describe precisely what you're testing against (especially as this specifies things in _decimal_, whereas modes are traditionally done in _octal_).

src/lib/rootfs/image/image.c
```diff
@@ -79,9 +84,64 @@ int rootfs_image_init(char *source, char *mount_dir) {
...
+                            singularity_message(ERROR, "Image does not have proper container group permissions to access image, aborting...\n");
```
bbockelm, line 134: If we tell the user there is something wrong, we should say how to fix it.  For example,

```
Image does not have proper container group permissions (current permissions are XYZ; needs to be at least ABC) to access image, aborting...
```

src/lib/rootfs/image/image.c
```diff
@@ -79,9 +84,64 @@ int rootfs_image_init(char *source, char *mount_dir) {
...
+                            ABORT(255);
+                        } else {
+                            singularity_message(VERBOSE, "Image access is permitted by container group %s specified in config\n", config_gname);
+                            break;
+                        }
+                    }
+                }
+            } while( config_gname != NULL );
```
bbockelm, line 142: Why is a do-while loop used here?  It appears the control flow is correct, but confusing to follow (each loop iteration either aborts the program or calls `break`).

Can't this just be rewritten without the loop?

src/util/file.h
```diff
@@ -19,6 +19,7 @@
  */
 
 
+int has_perm(int level, struct stat file_stat);
```
bbockelm, line 22: For new functions, inline documentation in the header would be appreciated (yes, I realize the prior authors didn't...).

src/util/file.c
```diff
@@ -40,6 +40,29 @@
 #include "lib/message.h"
 #include "lib/privilege.h"
 
+int has_perm(int level, struct stat file_stat) {
```
bbockelm, line 43: It's not clear to me how this can be correct.  Everything is done in decimal, but permissions are done in octal.

Also not clear what permission this is testing for.  It doesn't seem to be looking for a single permission, but at least one of several permissions.  Maybe rename it to something more descriptive?

bbockelm (changes requested):
Several nitpicks, but critically, it doesn't look like `has_perm` works as designed.

Maybe provide a few additions to `test.sh` to validate the security features work as designed?

src/util/file.c
bauerm97: @bbockelm So I actually wrote this function just really quickly as an example for what it *could* look like, however I never tested it at all. The idea of the function would be that it checks whether or not a user has the permission specified by level (4 = read; 2 = write; 1 = execute). Given one of these numbers as a level, it would then compare it to the owner field, the group field, and the other field from the files permission bitfield. I wasn't sure if the comparison using bitwise and would automatically convert to octal or not, but this was never meant to necessarily be merged as is, rather as me just throwing out some example code.

bbockelm:
@c1t4r @bauerm97 - can you check out https://github.com/bbockelm/singularity/tree/protected_mode ?

That accomplishes about the same thing (I think) but doesn't rely on us re-implementing POSIX checks.  I'd rather go that direction...

bauerm97:
@bbockelm To be honest, I'm not really sure what features @c1t4r is requesting to be implemented at this point. I've passed over your code and it looks good to me at first glance, however you'll have to ask @c1t4r if this will fulfill his use case. Furthermore, I think that defending the reasoning behind some of these implementation decisions would be useless, since certain parts of this code (such as the do-while loop, has_perm function, etc) appear to be taken from a branch I put up on my fork as an example not intended to be merged.

bbockelm:
My (re-)implementation was cleaned up and submitted as #350.

serious-steve:
Thanks @bauerm97 @bbockelm, in particular for the exhaustive analysis - I can learn from that. Guess I didn't issue my PR as intended so @bauerm97 isn't to blame. I am about to check the new implementation and whether it can satisfy my requirements...

bbockelm:
@gmkurtzer @c1t4r -

Since we merged the "protected image mode" (which hits a lot of the same functionality), I am going to close the PR for now.  Please re-open if you'd like to revisit

PR #325, opened 2016-11-22
Implement basic authentication for Docker registries

flx42:
There are a few problems currently:
- The Docker registry to use cannot be modified for `singularity import`
- The `--no-token` option is tricky. A token might be necessary for a registry, even if you don't need an account. We should attempt without a token, then try to get a token if we receive a 401.
- Basic authentication is not supported, so I don't think there is a way to fetch private repos.
- Token caching might not be a good idea: the token might expire when downloading large images over a slow connection. I think it's safer to get a new token per layer, and it makes the code simpler imo.

To specify the registry for `singularity import`:
```
export SINGULARITY_DOCKER_REGISTRY='--registry myrepo'
export SINGULARITY_DOCKER_AUTH='--username fabecassis --password [mygreatpassword]'
```
Having to export the password as an environment variable might not be great, I'm open to better suggestions. Maybe an authentication file like Docker does.

I tested my patch with “singularity import” on the following:
- DockerHub: a public and a private repo
- Google Container Registry: gcr.io/tensorflow (public)
- Githost.io (GitLab): a private repo 

It seemed to work, please test!

### DIFF
--- a/libexec/bootstrap/modules-v2/build-docker.sh
+++ b/libexec/bootstrap/modules-v2/build-docker.sh
@@ -85,19 +85,12 @@ fi
 
 
 ### Does the registry require authentication?
-SINGULARITY_DOCKER_AUTH=`singularity_key_get "Token" "$SINGULARITY_BUILDDEF"`
-if [ -n "${SINGULARITY_DOCKER_AUTH:-}" ]; then
-    message 1 "Token: $SINGULARITY_DOCKER_AUTH\n"
-
-    # A command of "no" means don't add token auth header
-    if [ "$SINGULARITY_DOCKER_AUTH" == "no" ]; then
-        SINGULARITY_DOCKER_AUTH="--no-token"
-
-    # Anything else, we do authentication
-    else
-        SINGULARITY_DOCKER_AUTH=""
-    fi
-
+SINGULARITY_DOCKER_USERNAME=`singularity_key_get "Username" "$SINGULARITY_BUILDDEF"`
+SINGULARITY_DOCKER_PASSWORD=`singularity_key_get "Password" "$SINGULARITY_BUILDDEF"`
+if [ -n "${SINGULARITY_DOCKER_USERNAME:-}" ] && [ -n "${SINGULARITY_DOCKER_PASSWORD:-}" ]; then
+    message 1 "Username: $SINGULARITY_DOCKER_USERNAME\n"
+    message 1 "Password: [hidden]\n"
+    SINGULARITY_DOCKER_AUTH="--username $SINGULARITY_DOCKER_USERNAME --password $SINGULARITY_DOCKER_PASSWORD"
 else
     SINGULARITY_DOCKER_AUTH=""   
 fi
--- a/libexec/helpers/image-import.sh
+++ b/libexec/helpers/image-import.sh
@@ -54,7 +54,7 @@ fi
 case "$IMPORT_URI" in
     docker://*)
         CONTAINER_NAME=`echo "$IMPORT_URI" | sed -e 's@^docker://@@'`
-        SINGULARITY_IMPORT_GET="$SINGULARITY_libexecdir/singularity/python/cli.py --rootfs '$SINGULARITY_ROOTFS' --docker '$CONTAINER_NAME' --cmd"
+        SINGULARITY_IMPORT_GET="$SINGULARITY_libexecdir/singularity/python/cli.py --rootfs '$SINGULARITY_ROOTFS' --docker '$CONTAINER_NAME' ${SINGULARITY_DOCKER_REGISTRY:-} ${SINGULARITY_DOCKER_AUTH:-}"
     ;;
     http://*|https://*)
         SINGULARITY_IMPORT_GET="curl -L -k '$IMPORT_URI'"
--- a/libexec/python/cli.py
+++ b/libexec/python/cli.py
@@ -33,7 +33,7 @@ from docker.api import (
     get_manifest 
 )
 
-from utils import extract_tar, change_permissions, get_cache
+from utils import extract_tar, change_permissions, get_cache, basic_auth_header
 from logman import logger
 import argparse
 import os
@@ -76,13 +76,15 @@ def main():
                         help="boolean to specify that CMD should be used instead of ENTRYPOINT as the runscript.", 
                         default=False)
 
+    parser.add_argument("--username",
+                        dest='username',
+                        help="username for registry authentication",
+                        default=None)
 
-    # Flag to indicate a token is not required
-    parser.add_argument("--no-token", 
-                        dest='notoken', 
-                        action="store_true",
-                        help="boolean to specify that the CMD should be included as a runscript (default is not included)", 
-                        default=False)
+    parser.add_argument("--password",
+                        dest='password',
+                        help="password for registry authentication",
+                        default=None)
 
 
     # Flag to disable cache
@@ -112,11 +114,11 @@ def main():
            sys.exit(1)
        logger.info("Root file system defined by env variable as %s", singularity_rootfs)
 
-    # Does the registry require a token?
-    doauth = True
-    if args.notoken == True:
-       doauth = False
-    logger.info("Do registry authentication set to %s", doauth)
+    # Does the registry require authentication?
+    auth = None
+    if args.username is not None and args.password is not None:
+        auth = basic_auth_header(args.username, args.password)
+        logger.info("Username for registry authentication: %s", args.username)
 
     # Does the user want to override default Entrypoint and use CMD as runscript?
     includecmd = args.includecmd
@@ -169,12 +171,12 @@ def main():
                                 namespace=namespace,
                                 repo_tag=repo_tag,
                                 registry=args.registry,
-                                auth=doauth)
+                                auth=auth)
 
         # Get images from manifest using version 2.0 of Docker Registry API
         images = get_images(manifest=manifest,
                             registry=args.registry,
-                            auth=doauth)
+                            auth=auth)
         
        
 #  DOWNLOAD LAYERS -------------------------------------------
@@ -185,12 +187,6 @@ def main():
                                disable_cache = args.disable_cache)
 
         layers = []
-
-        # Let's get a token to use across image downloads
-        token = get_token(repo_name=repo_name,
-                          namespace=namespace,
-                          permission="pull")
-
         for image_id in images:
 
             # Download the layer, if we don't have it
@@ -202,8 +198,7 @@ def main():
                                   repo_name=repo_name,
                                   download_folder=cache_base,
                                   registry=args.registry,
-                                  auth=doauth,
-                                  token=token) 
+                                  auth=auth)
 
             layers.append(targz) # in case we want a list at the end
                                  # @chrisfilo suggestion to try compiling into one tar.gz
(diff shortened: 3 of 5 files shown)


gmkurtzer:
I will import once @vsoch has had a chance to sign-off on the Python portions of the code.

Thanks @flx42, awesome feature addition!

vsoch:
It looks good at first glance - I'd like to go through it carefully. I should have time in the next day or so, just not at this very moment. Huge thanks for making this addition!

flx42:
@vsoch sure!
If that helps, here's a command-line version of how authentication can be used to access a private repo on DockerHub:
```
$ CREDENTIAL=$(echo -n flx42:[password] | base64)
$ TOKEN=$(http 'https://auth.docker.io/token?service=registry.docker.io&scope=repository:flx42/cuda-samples:pull' Authorization:"Basic $CREDENTIAL" | jq -r '.token')
$ http https://index.docker.io/v2/flx42/cuda-samples/tags/list Authorization:"Bearer $TOKEN"
```

libexec/bootstrap/modules-v2/build-docker.sh
```diff
@@ -85,19 +85,12 @@ fi
...
-        SINGULARITY_DOCKER_AUTH="--no-token"
-
-    # Anything else, we do authentication
-    else
-        SINGULARITY_DOCKER_AUTH=""
-    fi
-
+SINGULARITY_DOCKER_USERNAME=`singularity_key_get "Username" "$SINGULARITY_BUILDDEF"`
+SINGULARITY_DOCKER_PASSWORD=`singularity_key_get "Password" "$SINGULARITY_BUILDDEF"`
+if [ -n "${SINGULARITY_DOCKER_USERNAME:-}" ] && [ -n "${SINGULARITY_DOCKER_PASSWORD:-}" ]; then
```
vsoch, line 90: I like these keys `Username` and `Password` as they could be generally used for different bootstrap options. @gmkurtzer - is there an easy way to make it so capitalization doesn't matter? USERNAME should work equivalently to username or Username.

libexec/helpers/image-import.sh
```diff
@@ -54,7 +54,7 @@ fi
 case "$IMPORT_URI" in
     docker://*)
         CONTAINER_NAME=`echo "$IMPORT_URI" | sed -e 's@^docker://@@'`
-        SINGULARITY_IMPORT_GET="$SINGULARITY_libexecdir/singularity/python/cli.py --rootfs '$SINGULARITY_ROOTFS' --docker '$CONTAINER_NAME' --cmd"
+        SINGULARITY_IMPORT_GET="$SINGULARITY_libexecdir/singularity/python/cli.py --rootfs '$SINGULARITY_ROOTFS' --docker '$CONTAINER_NAME' ${SINGULARITY_DOCKER_REGISTRY:-} ${SINGULARITY_DOCKER_AUTH:-} --cmd"
     ;;
```
vsoch, line 58: I didn't look at this part when @gmkurtzer implemented - it looks like it doesn't include logic for including / not including the command as runscript. Right now, the default (without `--cmd`) is to use the entrypoint, which is the more proper of the two (CMD and ENTRYPOINT) as it is the container's executable. Adding the `--cmd` argument overrides this default, and changes the runscript to be CMD. For import, we would likely want the default to be the same (not having CMD) and akin to the docker credentials, have it be an envrionmental variable. Thoughts @gmkurtzer @flx42 ?

libexec/python/cli.py
```diff
@@ -112,11 +115,12 @@ def main():
...
        logger.info("Root file system defined by env variable as %s", singularity_rootfs)
 
-    # Does the registry require a token?
-    doauth = True
-    if args.notoken == True:
-       doauth = False
-    logger.info("Do registry authentication set to %s", doauth)
+    # Does the registry require authentication?
+    auth = None
+    if args.username is not None and args.password is not None:
```
vsoch, line 119: Can we move this guy into a function in docker/utils.py that takes a username,password, and returns an updated auth object in the same fashion? This will be useful in case we need to sneak it in other places at some point - we'd want them all to change based on pinging the same function.

libexec/bootstrap/modules-v2/build-docker.sh
```diff
@@ -85,19 +85,12 @@ fi
...
+    message 1 "Username: $SINGULARITY_DOCKER_USERNAME\n"
+    message 1 "Password: [hidden]\n"
+    SINGULARITY_DOCKER_AUTH="--username $SINGULARITY_DOCKER_USERNAME --password $SINGULARITY_DOCKER_PASSWORD"
 else
```
vsoch: I like this :)

vsoch, line 94: I like this :)

libexec/python/cli.py
```diff
@@ -185,12 +189,6 @@ def main():
                                disable_cache = args.disable_cache)
 
         layers = []
-
```
vsoch, line 188: I have mixed feelings about this because it seems redundant, but you are correct that a large or timely download could mean an invalid token. It will probably be annoying from a user perspective to have the token printed each time, a fix could be to obtain it each time (but only have the message appear in the debug logs).

libexec/bootstrap/modules-v2/build-docker.sh
gmkurtzer: They shouldn't be capitalization specific already:

```
    if OUT=`egrep -i "^$KEY:" $FILE`; then
        echo "$OUT" | head -n 1 | sed -e "s@^$KEY:\s*@@i" | sed -e "s@\s*#.*@@"
        return 0
    fi
```

libexec/python/docker/api.py
```diff
@@ -47,24 +52,46 @@ def create_runscript(cmd,base_dir):
     return output_file
 
 
-def get_token(repo_name,namespace="library",scope="repository",permission="pull"):
+def get_token(registry,repo_name,namespace,scope="repository",permission="pull",auth=None):
     '''get_token will use version 2.0 of Docker's service to return a token with given permission and scope - this
```
vsoch: What is the reason for requiring the registry here? Akin to the other functions, why not have the default as None, and then use api_base if it is? We might want to turn those three lines into a proper function.

libexec/python/docker/api.py
```diff
@@ -47,24 +52,46 @@ def create_runscript(cmd,base_dir):
...
     function does work, but the token doesn't seem to work when used with other functions below for authentication
```
vsoch: oups, need to update this line since we fixed the "doesn't work" part :P

libexec/helpers/image-import.sh
gmkurtzer: Ahh, good catch. Probably best to leave it off then.

libexec/python/docker/api.py
```diff
@@ -157,11 +184,11 @@ def get_manifest(repo_name,namespace,repo_tag="latest",registry=None,auth=True):
     logger.info("Obtaining manifest: %s", base)
     
     # Format the token, and prepare a header
-    token = None
-    if auth == True:
-        token = get_token(repo_name=repo_name,
-                          namespace=namespace,
-                          permission="pull")
+    token = get_token(registry=registry,
```
vsoch, line 187: I like this logic much better.

vsoch:
@gmkurtzer and @flx42 see my comments on the code - once we fix up those details this should be good to go! Then let's rebase and I'll peek at the other small PR. These additions are great, I'm really grateful you did them @flx42 !

libexec/bootstrap/modules-v2/build-docker.sh
vsoch: ok, awesome. The last time I tried was a long time ago, I remember specifying from instead of From didn't work, glad that's fixed!

libexec/helpers/image-import.sh
vsoch: Agreed - we can stick to default (entrypoint) for now, and then if someone says "heeeey I want CMD on import too!" then it should be pretty easy to add.

flx42: It should be a different patch set anyway, so I'll leave it this way :)

vsoch: Would you mind removing the --cmd bit at the end? If it's a hassle I'd be glad to submit a PR to your PR, haha. I think it's important to have this behavior be consistent asap between bootstrap and import (which it currently isn't).

libexec/python/cli.py
flx42: Actually, I removed your `logger.info("Obtaining token: %s", base)` line.
A line will only be printed if a token is **not** required: `logger.info("No token required for registry %s", registry)`. But we can remove it, no need to be overly verbose if everything is fine...

vsoch: Excellent. Yes, this line only makes sense if we did it once.

libexec/python/docker/api.py
flx42: I think it makes less sense here, I think functions like `get_config`, `get_manifest` or `get_layer` are part of the actual public API of your module. But it would be weird to have `get_token` part of the public API, you always need a token for something :)
But I can change it for the sake of consistency if you want.

vsoch: Do you mean having get_token be provided at all? For organization - I don't see any reason that someone might want to use it, you are right that it might be more appropriate as a utility or authentication function - if we had a larger code base it would be appropriate to have an auth.py in the docker folder, but I didn't think this level of organization was necessary (yet anyway).  For what my comment was about - Is there a reason for not having the registry default to index.docker.io as we do in the other functions? That should be easy to add, even it's not a complete function (eg, the same three lines of code are repeated in multiple places in the file!)

flx42: Yeah, I believe `get_token` is more an utility function. It should only be called from `get_config`, `get_manifest`, `get_layer`, etc. And these functions can indeed default to `index.docker.io`, but no need to do it twice IMO.
But yeah, it's easy to fix :)

vsoch: cool, thanks!

flx42: So I should add the default?

libexec/python/docker/api.py
flx42: Ah yes, correct, the doc needs updating. Actually parameters `scope` and `permission` are gone. We shouldn't need to pass these values to a function, they should be inferred from the HTTP response after we get a `401`.

libexec/python/docker/api.py
vsoch: Yes please!

libexec/python/docker/api.py
vsoch: +1. Also - I'll take care of updating the online docs to give details about using the command line token for import once this is merged. I'll ping you with the link to check over for a second set of eyes.

flx42:
It should be good now, please take a look @vsoch!
I amended and rebased my patch.

libexec/python/cli.py
flx42: Done

vsoch:
Looks great! Thanks for your hard work on this.

vsoch:
Thanks @flx42 - here are the [updated docs](http://singularity.lbl.gov/docs-docker#custom-authentication) if you have some time to check them out and tell me what I messed up :)

PR #326, opened 2016-11-22
Added a check for ENOENT for getgrgid and getpwuid

### LINKED ISSUE #319: Non-Existant GID handling
There are cases where users will have GIDs added to their user, but no actual groups will exist in /etc/groups. Singularity errors out at this. It would be handy to have a way to ignore the issue, and just warn the user of skipped GIDs.

Nekel-Seyew:
Fixes #319 

fixes missing ENOENT check for both getgrgid and getpwuid calls
 -
 -
 -

@singularityware-admin

…etpwuid calls.

### DIFF
--- a/src/lib/file/group/group.c
+++ b/src/lib/file/group/group.c
@@ -90,7 +90,7 @@ int singularity_file_group(void) {
     errno = 0;
     if ( ! pwent ) {
         // List of potential error codes for unknown name taken from man page.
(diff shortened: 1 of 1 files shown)


bbockelm (approved):
Looks good to me.

I hate enumerating a whitelist of error codes, but I think it's better than risking letting unexpected behavior past.

gmkurtzer:
Looks good to me too, thanks guys!

PR #337, opened 2016-11-24
First attempt at enabling pylint support. (#335)

### LINKED ISSUE #335: [python] plugin for singularity-bot to check python
http://docs.sonarqube.org/display/PLUG/Python+Plugin

bbockelm:
Fixes #335 

This installs `pylint` and makes the project not `C`-specific, which should be sufficient to get a python code analysis going.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,9 @@ addons:
   sonarqube: true
   apt:
     packages:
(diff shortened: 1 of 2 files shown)


bbockelm:
Hm.  SonarQube client doesn't like this PR.  Will close and try again.

Issue #331, opened 2016-11-23
Docker layer cache is not fail-safe or concurrency-safe


flx42:
There are a few shortcomings with the current Docker layer caching code. It doesn't handle failures well, for instance if I kill a `singularity import` during a layer download, the next run will believe the layer is cached:
```
$ singularity import tensorflow docker://tensorflow/tensorflow:nightly-devel-gpu
Cache folder set to /root/.singularity/docker
Downloading layer sha256:a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4
Extracting /root/.singularity/docker/sha256:a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4.tar.gz
Downloading layer sha256:247d4da3141d6674f423f02a06b43d09c39a5cb504c8295ca007d547f368b858
^CTraceback (most recent call last):
  File "/usr/local/libexec/singularity/python/cli.py", line 241, in <module>
    main()
  File "/usr/local/libexec/singularity/python/cli.py", line 206, in main
    token=token) 
  File "/usr/local/libexec/singularity/python/docker/api.py", line 237, in get_layer
    return api_get(base,headers=token,stream=download_folder)
  File "/usr/local/libexec/singularity/python/utils.py", line 163, in api_get
    chunk = response.read(chunk_size)
  File "/usr/lib/python2.7/socket.py", line 384, in read
    data = self._sock.recv(left)
  File "/usr/lib/python2.7/httplib.py", line 612, in read
    s = self.fp.read(amt)
  File "/usr/lib/python2.7/socket.py", line 384, in read
    data = self._sock.recv(left)
  File "/usr/lib/python2.7/ssl.py", line 756, in recv
    return self.read(buflen)
  File "/usr/lib/python2.7/ssl.py", line 643, in read
    v = self._sslobj.read(len)
KeyboardInterrupt
ERROR: Aborting with RETVAL=1
```
And then the next run:
```
$ singularity import tensorflow docker://tensorflow/tensorflow:nightly-devel-gpu
Cache folder set to /root/.singularity/docker
Extracting /root/.singularity/docker/sha256:a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4.tar.gz
Extracting /root/.singularity/docker/sha256:247d4da3141d6674f423f02a06b43d09c39a5cb504c8295ca007d547f368b858.tar.gz

[...]

PR #346, opened 2016-11-27
Addition of Singularity Hub (shub) endpoint

### LINKED ISSUE #301: Enhancement: Integration with Singularity Hub
The user should be able to shell / bootstrap etc using another Singularity image, deployed by singularity hub, e.g.:

      singularity run shub://vsoch/singularity-images

I will start developing this.

vsoch:
Fixes #301 

This PR adds the Singularity Hub (shub) endpoint, with uri `shub://`. I have added it to `import`, `run`, `bootstrap` and `shell`, as those are the commands that I'm familiar with (can't remember the last time I typed start, mount, etc) - if the shub:// endpoint needs to be added elsewhere let's discuss!

### Addition of `pull`
I have introduced a new command, "pull" for the user to download a "ready for use" image from Singularity Hub. It works like this:

      singularity pull shub://7

`shub://` refers to the endpoint / URL for singularity hub, and 7 refers to the container ID assigned by singularity hub, which the user would have obtained via browsing the interface. In the future we can expand this functionality to include the collection/name of an image, akin to Docker, something like this:

      singularity pull shub://vsoch/singularity-images:latest

and I will add this for some future PR, after the endpoints of Singularity Hub themselves are fine tuned to handle things like query, search, etc.

### Why pull is different from import/bootstrap
I think we need a command that is stupid and just pulls an image from an endpoint, without needing sudo. The use case is (of course) having someone in an environment that doesn't allow sudo, and he/she may not want to bootstrap or import an image, but just obtain (download) one as is from Singularity Hub. It's not quite the same as run or shell, because these are going to execute things.

#### Pulling

      singularity pull shub://7 
      Cache folder set to /home/vanessa/.singularity/shub
      Found image vsoch/singularity-images
      Downloading image... 61b4721e3583d1c84b524696c08771dd6fa188bc.tar.gz

If I try to do it again, It tells me that I already have it:

      singularity pull shub://7
      Cache folder set to /home/vanessa/.singularity/shub
      Image already exists at /home/vanessa/.singularity/shub/61b4721e3583d1c84b524696c08771dd6fa188bc.tar.gz, skipping download.

If I want to change the cache, I just change the environmental variable

      export SINGULARITY_CACHEDIR=/home/vanessa/Desktop
      singularity pull shub://7
      Cache folder set to /home/vanessa/Desktop/shub
      Found image vsoch/singularity-images
      Downloading image... 61b4721e3583d1c84b524696c08771dd6fa188bc.tar.gz

#### Shell
Shell uses the exact same python script, except it returns the container image path (a .tar.gz file) to the calling script, so it looks like this:

      singularity shell shub://7
      Cache folder set to /home/vanessa/Desktop/shub
      Image already exists at /home/vanessa/Desktop/shub/61b4721e3583d1c84b524696c08771dd6fa188bc.tar.gz, skipping download.
      Opening gzip compressed archive, stand by...
      Singularity.61b4721e3583d1c84b524696c08771dd6fa188bc.tar.gz> 

#### Run
This is the same as shell, but we run the runscript of course. In the example below, my image is crappy because I call a python executable that doesn't exist. Heh.

      singularity run shub://7
      Cache folder set to /home/vanessa/Desktop/shub 
      Image already exists at /home/vanessa/Desktop/shub/61b4721e3583d1c84b524696c08771dd6fa188bc.tar.gz, skipping download.
      Opening gzip compressed archive, stand by...
      /singularity: line 2: /usr/bin/python: No such file or directory

#### Bootstrap
Finally, we can have a file called `Singularity`

      bootstrap: shub
      From: 7

      %post
            echo "Hello World!"

and the same bootstrap functionality should work, again just pulling the image, and then exporting as a `.tar.gz`.  I can't say this is implemented optimally - this was me poking around what Greg did, and finding skeethy tricks to get things working. You might want to change this, but here is what I wound up doing:

- run the [python/cli/py script](https://github.com/vsoch/singularity/blob/add/shub-endpoint/libexec/python/cli.py) as before, it downloads the `.tar.gz` file as an image to the cache. If the command is `pull`, it stops here.
- In the case of the other functions, we provide a `singularity_rootfs` to the python command, and the python script doesn't extract the image here, it returns the downloaded image to the calling script and let's it handle things. This means saving this information in temporary files in the `singularity_rootfs` that are removed before the image is finalized (yes, this could get messy with some shared cache) and users running / shelling into the same shub image at the same time, and we will address this with #331. 
- The image is then treated as a standard `.tar.gz` is for each of import, run, and shell.

### Cache
The reason to have different cache folders (as we currently do) is because of contents and (possibly) permissions. The python defaults to saving things in:

      $HOME/.singularity/docker 
      $HOME/.singularity/shub

unless the user overrides `SINGULARITY_CACHEDIR` with something else, in which case you get:

      $SINGULARITY_CACHEDIR/docker 
      $SINGULARITY_CACHEDIR/shub

#### Stuff inside
The distinction should be made between layers and images. The docker cache always has layers. The shub cache (currently) always has images. While it might be some future development to also download shub in layers, I gave that a good effort and thinking this weekend, and I always run into some line of discomfort that we break reproducibility. So for now, let's work with entire images - the compressed don't take too long to download. It does become redundant if you have shared things between images. If we do come up with something that is akin to a layer, it needs to be carefully thought out and a deterministic algorithm that can take any image and always break it into the same pieces (that are identifiable across images). Too much discussion for this PR :)

#### Permissions
The first (docker) is definitely being used for import/bootstrap, and this requires sudo. So the user (not as sudo) by default cannot write to this folder. The second (shub) supports functions that can be run as root or the user. I am concerned about missing a case when the user will get a PermissionError writing to a folder created by root. I ran into this error once and did some testing and haven't since, so let's keep an eye out for it because I tend to suck at things involving fixing permissions :)

### Singularity Hub Representation
Each container on Singularity Hub is built from a particular commit of a Github repo. This means that a group of containers, called a container collection, is a repo itself. For example, here is the repo [vanessa/singularity-images](http://www.github.com/vsoch/singularity-images), and here is the associated collection [vanessa/singularity-images](https://singularity-hub.org/collections/1/) on singularity hub. The singularity commend line client is just grabbing an [image manifest](https://singularity-hub.org/api/containers/7/) that has links to all of these things.  The string that is being used to name the image (e.g. `61b4721e3583d1c84b524696c08771dd6fa188bc`.tar.gz ) is the commit id. I will note that the name of the image itself in Google Storage is the MD5 sum. Yes, this means we have a few ids:

 - singularity hub unique id (`7`): links to...
 - commit id (`61b4721e3583d1c84b524696c08771dd6fa188bc`): the github (or other version control) identifier
 - md5 sum: generated for the file when it is built, and how the image is named. You haven't seen this yet, see the Google Storage image at the bottom for an example)

I did it in this fashion because one could imagine some (future) strategy of building an image from layers, and I think we would want the layers to be named based on a file content hash, and then the image still named corresponding to the commit where the build file is. And since we can't be sure about using Github, or some other thing, I think singularity hub should have it's own central identifier. 

Looking forward to your feedback on this! The rest below here is more information about Singularity Hub for those interested, and my intended next steps after this initial functionality is added.

#### Singularity Hub Build
Singularity Hub is responsible for making sure that the image is built when the user pushes to Github, and that particular image produced via build of the [Singularity file](https://github.com/vsoch/singularity-images/blob/master/Singularity). The metadata (what is built, installed, etc) and the image itself are a pair that are locked together, and this is important for reproducibility. (Sidenote: this is one of the primary reasons I changed my initial strategy of using Google Drive to Google Storage to keep files - I was hugely uncomfortable with the chance that a user could delete things easily, and then Singularity Hub loses them.) Here is what it looks like on the storage side:

image

First notice the "folder address" goes from:

      bucket://singularity-hub  --> user --> repo --> commit

Folders aren't really folders on Google Storage, but having files named with this structure makes organization nice for the Google Cloud user interface, if/when any particular bucket gets large. Technically, all you would need is the image itself (the `.tar.gz) but the builder is also sending some other small files up to storage to make the image a more viewable, explorable thing. For example, the `files.txt` and `folders.txt` are flat files that Singularity Hub grabs to make [interactive visualizations](https://singularity-hub.org/containers/7/view). That same view extends to calculating comparisons (differences and similarities, and corresponding scores) between containers (not yet implemented in the interface) because I should be able to click on two versions, and then easily understand the difference. The VERSION is just the commit id, and the Singularity file is another copy of the build spec, in case the github repo is deleted. The `.zip` is just a compressed object with all the tiny files is some external application wants to build their own stuff, and wants to download one thing.

#### Singularity Hub Workflow
The intended use is based on the idea that (most) will likely use Singularity based on some kind of build file. I personally prefer this workflow because it guarantees a humanly readable thing, vs. a user running random commands on the command line (and then forgetting said commands), rendering the container essentially a black box that couldn't be made again. The researcher will ideally develop this file with version control (Github) and then linking a repo to Singularity Hub just means signing up and clicking a tab, eg clicking one of these:

image

Activating a repo creates a webhook that pings Singularity Hub at each push, so then any particular Singularity file (e.g. one used for some analysis) will have a version with an id (linked to an image file and the build spec) that can be included in the publication, or linked to open science framework. I'd like to give the user total control - meaning links when wanted, or the ability to freeze and export everything to some other desired location. None of this is added, but will be as needed!

#### Cache contents
As is now, the cached image is named according to the commit id. If a user changes the location of the cache or renames the image to something else, it's up to him or her to remember versioning. This makes me slightly uncomfortable because it introduces opportunity to break a reproducible workflow. However, given that the easiest thing to do is to run / shell / import images with singularity command line tool, the ideal use case is that the user doesn't need to do that. he/she can update an image in a pipeline by editing the build file, pushing to Github, getting the updated container id, and then running the command again.

Yes - I know what you are thinking - ideally the user wouldn't need to change the container id (a new build), and for this reason I am thinking of having an endpoint that will return the "latest" of an image, likely corresponding with the container collection name. We would also want an endpoint to look up and retrieve complete image information based on the commit id, in the case that a user finds an image on a file system lying around. Thoughts?

#### Metadata
We talked some a while back about having some preservation of metadata about an image within the image (header? file in root?) but this discussion was sort of lost, and so for the purposes of academic reproducibility, I'm taking the approach that Singularity Hub can (at least try) to keep information about the build file, versions, etc.

### Next Steps
I'd like to get this simple functionality finalized, and added to singularity. Then we will do the following:
 - resolve the docker and shub cache issue #331 
 - add the additional endpoints I listed above
 - do testing of singularity hub (with shub://) with a simple analysis across several cloud providers and clusters. We already have a plan for this, and I'm excited!

Looking forward to hearing everyone's thoughts! I'm going to work on something else for a bit.

@singularityware-admin


bbockelm:
I see you do some additions to the command line interface - could you please also update the bash completion script to reflect these?

vsoch:
Definitely - I'll get back to this guy ones the bootstrap et al PRs are gone through, and likely after fixing the argparse --> optparse issue as well!

vsoch:
Unless you think the bootstrap will take a long time - should I update this PR and merge sooner/before?

vsoch:
Hey @bbockelm I'm about to update this PR - could you be more specific about the bash completion script? Which one is that?

vsoch:
@gmkurtzer et. al this base endpoint addition is ready for review!

gmkurtzer:
Does `pull` always refer to a Singularity Hub location? If so, is the URI format optional?

vsoch:
Yes, right now `pull` is just for shub, and if you enter another endpoint it will send you a message that it's invalid. I think we should still require the URI to:

1. be consistent across commands in using it
2. allow for (future) support for additional pull endpoints

vsoch:
@gmkurtzer the singularity-hub builder code now will generate compressed images, not compressed tars (exported from the image). [Image 8](https://singularity-hub.org/containers/8/view). It works the same, it's just a different format that doesn't have the weird issues you outlined. Sorry I missed that, I just didn't know there was a difference!

vsoch:
Ug, I'm getting an error that it's not a valid image... can you give me a set of commands that, starting with a spec file or other, will generate a valid image?

gmkurtzer:
Dangit, did I merge it too soon?

bbockelm:
@vsoch - the bash completion stuff isn't a blocker.  Can you create a fresh issue?

This is the affected file:

https://github.com/singularityware/singularity/blob/master/etc/bash_completion.d/singularity

bash completion is pretty esoteric to write, but I think users really appreciate it.

vsoch:
@bbockelm this is so awesome! I'd love to do this.

PR #350, opened 2016-11-27
Extended permission model for image access

bbockelm:
This PR introduces two new features for _read-only_ containers (writable containers are unaffected):

1.  "Protected mode".  Allows access to images readable by a dedicated user account.   This has three possible settings
      * `none` (default): This feature is disabled.
      * `group`: The image must be readable by the dedicated user account *and* its group ownership must overlap with the invoking user's.  So, if user `bbockelm` is in group `foo` and `bar`, then the user would be able to access an image with mode `600` and ownership `singularity:foo`, `singularity:bar`, but not `singularity:baz`.
      * `user`: The image must be readable by the dedicated user account.  Given the user `bbockelm` above, they would be able to access an image mode `600` and ownership `singularity:baz` or mode `640` and ownership `root:singularity`.
      The dedicated user account is named `singularity` by default(when switching to this account, we set to the corresponding UID and primary GID).
2.  A boolean flag (`allow user image`) on whether users can provide their own images (defaults to `yes`).  When disabled, user-provided images (i.e., anything readable by the invoking user) are not allowed.

The default configuration values leave the permission model unchanged from the current code.

A few example use cases:
* Sysadmin only wants `singularity` to execute images she provides for the user.  In that case, set `allow user image = no` and `protected user mode = user`.
* Admin wants to provide a group-ownership based model for the users she maintains.  Users in group `physics` should not access the images available to group `chemistry`.  In that case, set `protected user mode = group`.
   * As a follow-up, we could put together a more fine-grained model that a user is only allowed to execute a fixed command inside the image (would prevent the user from doing arbitrary reads of the software inside the image).
* Admin wants only users belonging to a given group to create `singularity` images.  Set `allow user image = no` and `protected user mode = user`, and add these users to the `singularity` group.
   * In this case, users put their images in a publicly-accessible directory and allow other users to execute them.

### DIFF
--- a/etc/singularity.conf
+++ b/etc/singularity.conf
@@ -17,8 +17,30 @@ allow setuid = yes
 # Should we allow users to request the PID namespace?
 allow pid ns = yes
 
+# ALLOW USER IMAGE: [BOOL]
+# DEFAULT: yes
+# Should we allow users to run images they own?  If set to 'no', then the image
+# can only be executed through 'protected image mode' (see below).
+#
+# Note that if 'allow user image' is no and 'protected image mode' is none,
+# then no image file may be used.
+allow user image = yes
 
-# ENABLE OVERLAY: [BOOL]
+
+# PROTECTED IMAGE MODE: [STRING]
+# DEFAULT: none
+# The 'protected image mode' allows users to start singularity inside images
+# that they do not own.  There are three allowed modes:
+# - none: Protected image mode is disabled.  Users must own any image they run.
+# - group: The image file must be readable by the user 'singularity' AND the
+#          invoking user must be in the same group as the image file.
+# - user: The image file must be readable by the user 'singularity'.
+# The name of the user 'singularity' can be controlled via the
+# 'singularity user' option below.
+protected image mode = none
+
+
+# # ENABLE OVERLAY: [BOOL]
 # DEFAULT: yes
 # Enabling this option will make it possible to specify bind paths to locations
 # that do not currently exist within the container. Some limitations still exist
@@ -140,4 +162,15 @@ container dir = /var/singularity/mnt
 # /tmp/ does not work for your system, /var/singularity/sessions maybe a
 # better option.
 #sessiondir prefix = /var/singularity/sessions/
+
+
+# SINGULARITY USER: [STRING]
+# DEFAULT: singularity
+# The protected image mode will restrict Singularity images to those owned by
+# the 'singularity user'; this option allows admins to change the username we
+# use as a test.
+#
+# If user namespaces are enabled, the singularity user setting is ignored and
+# instead the invoking user is utilized for all access privilege checks.
+singularity user = singularity
 
--- a/etc/singularity.conf
+++ b/etc/singularity.conf
@@ -19,8 +19,9 @@ allow pid ns = yes
 
 # ALLOW USER IMAGE: [BOOL]
 # DEFAULT: yes
-# Should we allow users to run images they own?  If set to 'no', then the image
-# can only be executed through 'protected image mode' (see below).
+# Should we allow users to run images they can access according to normal
+# filesystem privileges?  If set to 'no', then the image can only be executed
+# through 'protected image mode' (see below).
 #
 # Note that if 'allow user image' is no and 'protected image mode' is none,
 # then no image file may be used.
@@ -29,12 +30,17 @@ allow user image = yes
 
 # PROTECTED IMAGE MODE: [STRING]
 # DEFAULT: none
-# The 'protected image mode' allows users to start singularity inside images
-# that they do not own.  There are three allowed modes:
-# - none: Protected image mode is disabled.  Users must own any image they run.
+# The 'protected image mode' allows users to launch singularity with images
+# that they cannot access according to filesystem privileges.
+# There are three allowed modes:
+# - none: Protected image mode is disabled.  The typical filesystem rules
+#         are applied.
 # - group: The image file must be readable by the user 'singularity' AND the
 #          invoking user must be in the same group as the image file.
 # - user: The image file must be readable by the user 'singularity'.
+# In both 'group' and 'user' cases, the image must be either owned by the
+# singularity user or be owned by the singularity user's primary group.
+#
 # The name of the user 'singularity' can be controlled via the
 # 'singularity user' option below.
 protected image mode = none
--- a/src/get-section.c
+++ b/src/get-section.c
@@ -64,7 +64,7 @@ int main(int argc, char ** argv) {
         ABORT(255);
     }
 
-    singularity_message(DEBUG, "Iterating through file looking for sections matching: \%%s\n", section);
+    singularity_message(DEBUG, "Iterating through file looking for sections matching: %s\n", section);
     while ( fgets(line, MAX_LINE_LEN, input) != NULL ) {
         if ( strncmp(line, strjoin("%", section), strlength(section, 128) + 1) == 0 ) {
             toggle_section = 1;
--- a/src/get-section.c
+++ b/src/get-section.c
@@ -64,7 +64,7 @@ int main(int argc, char ** argv) {
         ABORT(255);
     }
 
-    singularity_message(DEBUG, "Iterating through file looking for sections matching: %s\n", section);
+    singularity_message(DEBUG, "Iterating through file looking for sections matching: %%%s\n", section);
     while ( fgets(line, MAX_LINE_LEN, input) != NULL ) {
         if ( strncmp(line, strjoin("%", section), strlength(section, 128) + 1) == 0 ) {
             toggle_section = 1;
--- a/src/lib/config_parser.c
+++ b/src/lib/config_parser.c
@@ -129,6 +129,11 @@ char *singularity_config_get_value(char *key) {
     return(NULL);
 }
 
+char *singularity_config_get_value_default(char *key, const char *def) {
+    char *value = singularity_config_get_value(key);
+    return value ? value : strdup(def);
+}
+
 /*
  * Gets the associated boolean value of key from config_fp. Passes
  * key into singularity_get_config_value() and then checks if that
--- a/src/lib/privilege.c
+++ b/src/lib/privilege.c
@@ -49,10 +49,10 @@ static struct PRIV_INFO {
     gid_t *gids;
     size_t gids_count;
     int userns_ready;
-    int disable_setgroups;
     uid_t orig_uid;
     uid_t orig_gid;
     pid_t orig_pid;
+    int dropped_groups;
     int target_mode;  // Set to 1 if we are running in "target mode" (admin specifies UID/GID)
 } uinfo;
 
@@ -163,6 +163,13 @@ void singularity_priv_escalate(void) {
         ABORT(255);
     }
 
+    singularity_message(DEBUG, "Clearing supplementary GIDs.\n");
+    if ( setgroups(0, NULL) == -1 ) {
+        singularity_message(ERROR, "Unable to clear the supplementary group IDs: %s (errno=%d).\n", strerror(errno), errno);
+        ABORT(255);
+    }
+    uinfo.dropped_groups = 1;
+
 }
 
 // Note that we don't initialize the singularity user at the beginning of the program;
@@ -173,6 +180,18 @@ void singularity_priv_escalate(void) {
 static void cache_singularity_user(void) {
 
     if (sinfo.ready == 1) {
+        return;
+    }
+
+    if ( uinfo.ready != 1 ) {
+        singularity_message(ERROR, "User info is not available\n");
+        ABORT(255);
+    }
+
+    if ( uinfo.userns_ready == 1 ) {
+        singularity_message(DEBUG, "User namespaces enabled; setting singularity user as the invoking user.\n");
+        sinfo.uid = uinfo.uid;
+        sinfo.gid = uinfo.gid;
         return;
     }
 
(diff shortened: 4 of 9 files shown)


vsoch:
Stupid question about permissions for a single user - don't the images carry permissions akin to files, so if someone would want to limit permissions, they would just change the permissions of the image file?

bauerm97:
@vsoch Correct. In fact, images are just another file and carry identical permissions as any other file.

etc/singularity.conf
```diff
@@ -17,8 +17,30 @@ allow setuid = yes
...
-# ENABLE OVERLAY: [BOOL]
+# ALLOW USER IMAGE: [BOOL]
+# DEFAULT: yes
+# Should we allow users to run images they own?  If set to 'no', then the image
+# can only be executed through 'protected image mode' (see below).
+#
+# Note that if 'allow user image' is no and 'protected image mode' is none,
+# then no image file may be used.
+allow user image = yes
+
```
vsoch, line 29: ok, so this is sort of like a controller for extending permissions for singularity image? For example, by quickly changing the values in the config file (protected image mode to none and allow user image to no) then no containers can be used. This would be very useful in some kind of situation where singularity needed to be "frozen" but not uninstalled / removed. 

Are these permissions living / changing or do they apply to all images created after being set?

etc/singularity.conf
```diff
@@ -17,8 +17,30 @@ allow setuid = yes
...
+
+# PROTECTED IMAGE MODE: [STRING]
+# DEFAULT: none
+# The 'protected image mode' allows users to start singularity inside images
+# that they do not own.  There are three allowed modes:
```
vsoch: what do you mean start singularity inside images? Does that actually work? (reminds me of this ;P https://github.com/jpetazzo/dind)

etc/singularity.conf
```diff
@@ -141,3 +163,14 @@ container dir = /var/singularity/mnt
...
 #sessiondir prefix = /var/singularity/sessions/
 
+
+# SINGULARITY USER: [STRING]
+# DEFAULT: singularity
+# The protected image mode will restrict Singularity images to those owned by
+# the 'singularity user'; this option allows admins to change the username we
+# use as a test.
+#
+# If user namespaces are enabled, the singularity user setting is ignored and
```
vsoch, line 179: I don't understand this. Is singularity the user that owns the singularity process? What kind of tests would be done and why would admins want to change this? Why does it have to be just one user?

bbockelm:
@vsoch - while you could _allow_ access based on permissions, without this patch, there's no way to _remove_ someone's access from executing singularity.

And - doh! - now that I read through the original designs, I need to add back in the ownership check for user `singularity`.  Without it, any user can bypass the `allow user image = no` (they can always create an image readable by user `singularity`...).

vsoch:
One more stupid question. How does this move between systems? For example, if I'm working on some cluster that gives me a lot of freedom with doing what I want with my images, and then I move my image to another cluster, do I lose some functionality? I *think* I like this idea because it gives admins some power to do things in cases of emergency, but it seems like something that if done to a degree too high, it goes against one of the doctrine of singularity that the user has the same ability to do stuff on the command line as in the image. But most likely admins would be less conservative, and just use this as a security fallback, and I'm overthinking it? What do you think?

etc/singularity.conf
bbockelm: The permissions examined are those of the image file.  Since they're enforced by Linux, it's based on whenever `singularity` is invoked.

etc/singularity.conf
bbockelm: `singularity` is the user that must own the image file (when utilizing protected image mode).  The test performed is literally whether `singularity` can `fopen()` the file (and whether it owns the image file).  I don't foresee admins changing the name of the user, but I dislike hardcoding usernames (what if the site already has a user named `singularity`, used for other purposes?).

Because we perform the checks as the `singularity` user (instead of looking at the mode / ownership and "recreating" POSIX), we:
* Are less likely to introduce bugs as we are using the Linux kernel's logic.
* Will function in the face of with less-common access control methods such as file attributes and ACLs.

Since there is the `group` mode that provides more fine-grained permissions, I don't think there's much use in allowing the admin to specify multiple `singularity` users.

bbockelm:
I would look at it in the opposite way - this allows admins who are paranoid about the image format to allow their users utilize singularity for a limited number of use cases.  Without this, they wouldn't install singularity at all.

bbockelm:
Ok, I'm happy with the functionality, implementation, and test coverage.  Rebased to make the branch's commit history more readable (basically, removed a bunch of intermediate commits from trying to wrangle the tests).

serious-steve:
I read the description and use cases, and I believe it enables me to do anything I originally proposed.
Just not simultaneously... 

How can I cover the following use case:
Admin offers executable non-readable `SW1`-containers to any user (since `SW1` is side-licensed), and executable non-readable `SW2`-containers to an exclusive group `chemistry_xyz` (since members of the group need to have a personal `SW2` license)

Could this be achieved by modifying the `protected mode=user` behaviour when a certain group such as `nogroup` is used for the container?

src/get-section.c
```diff
@@ -64,7 +64,7 @@ int main(int argc, char ** argv) {
         ABORT(255);
     }
 
-    singularity_message(DEBUG, "Iterating through file looking for sections matching: \%%s\n", section);
+    singularity_message(DEBUG, "Iterating through file looking for sections matching: %s\n", section);
     while ( fgets(line, MAX_LINE_LEN, input) != NULL ) {
```
bauerm97: Can I ask what the reasoning behind removing "%" is? It would make sense to include the % sign in the string so it reads out "%pre" or "%setup". I'm not sure if % is the correct way to escape a % sign, but I know that %% works. Maybe use that instead?

src/lib/config_parser.c
```diff
@@ -129,6 +129,11 @@ char *singularity_config_get_value(char *key) {
     return(NULL);
 }
 
+char *singularity_config_get_value_default(char *key, const char *def) {
```
bauerm97, line 132: Right now, default values are created by adding a line `key = default` into the configuration file that gets distributed. Are we sure we need to add this function? The first method also ensure that two different calls to the function will always return the same value. If we add this function, it would make sense to me that we should change every call to `config_get_value` -> `config_get_value_default`. Otherwise, simply adding the default entry into the config file should do. Agree?

src/get-section.c
```diff
@@ -64,7 +64,7 @@ int main(int argc, char ** argv) {
         ABORT(255);
     }
 
-    singularity_message(DEBUG, "Iterating through file looking for sections matching: \%%s\n", section);
+    singularity_message(DEBUG, "Iterating through file looking for sections matching: %s\n", section);
```
bauerm97: Can I ask what the reasoning behind removing "%" is? It would make sense to include the % sign in the string so it reads out "%pre" or "%setup". I'm not sure if % is the correct way to escape a % sign, but I know that %% works. Maybe use that instead?

(This is not really important or imperative, just nitpicking)

bauerm97 (changes requested):

bauerm97:
@c1t4r I believe that your use case is covered here. For instance, we have user `mibauer` in group `cdefault`. We also have user `c1t4r` in group `cdefault` & `c1`. We have two images with mode `600`: `container1` with ownership `singularity:cdefault` and `container2` with ownership `singularity:c1`. We set `protected user mode = group` in the config file. User `c1tr4` can access `container1` & `container2`. User `mibauer` can only access `container1`

etc/singularity.conf
bauerm97: @bbockelm What are your thoughts on allowing several `singularity user` lines? The use case for this might be letting user `mibauer` in group `science` bring an image into the cluster that anybody in group `science` could access.

serious-steve:
@bauerm97 Technically you're right. 
I just hoped to avoid adding every single user to each group dedicated to a side licensed SW which can be a hassle during registration/deprovisioning. 
IMO the mechanism would be nicer if coupled to the group dedicated to the SW.
However, I could probably live with that and would be just happy to have a possibility to implement the use case if the feature looks like that.

Update: OK no idea what I was thinking... it is in fact simple :D

bauerm97:
@c1t4r Just change the ownership of the image with SW1 to `singularity:sw1_group` and it will couple into that already.

src/get-section.c
bbockelm: Ah - so that's what it was trying to attempt!  (As written, it was an invalid format specifier.)  In that case, it'll need to change to `%%%s`.

src/lib/config_parser.c
bbockelm: The problem is that we distribute a _config_ file, not a _defaults_ file.  That means, on upgrade, we have no clue as to whether our updated configs are in place.  For example, on RPM-based platforms, updated configs will be saved as `.rpmnew` if the admin changes _any_ line of the config file.  (We've already seen this occur.)

Hence, we're left with the unenviable task of making sure the code defaults match the config defaults (NOTE: other projects I have worked with solve this by either auto-generating the config file from the code defaults at build-time or simply removing defaults from the config file).

We already do this for all the boolean configuration knobs; this just appears the first time we've done it for strings.

Yes, the configuration system code really needs an uplift to make it future-proof.

bauerm97: In that case would it make sense to leave this out of the PR for now, and then go in and give the config system a bit of a facelift to account for all of this in a separate PR? That way we don't mix and match what we're solving in each PR and what we're doing in the code, but rather go through and systematically change what needs to be changed.

I could maybe handle that when I'm finished with the bootstrapping code. There's not much left with the bootstrapping code (just polish and design decisions at this point!!), but I am prioritizing my time on that at the moment.

etc/singularity.conf
bbockelm: @bauerm97 -
Two options exist:
   1.  Simply keep the `allow user images = yes` default and allow user `mibauer` to use normal Unix permissions to share his images.
   2.  If `allow user images = no` is set for other reasons, you could add `mibauer` to the `singularity` group, set `protected image mode = user` and have him set those images to be owned by group `singularity`.
      * Note that this allows anyone on the host to access the image.

So (2) is close - but not exactly - to what you want.

I currently clear out the supplementary GID list when testing access from the `singularity` user.  I suppose we could keep these GIDS for the `singularity` user.  In that case, (2) would require that the image live in group-readable (not-world-readable) space and owned by the `singularity` group.

bbockelm:
@gmkurtzer - ping!

gmkurtzer:
I'm about half done on my initial look through the changes. Tomorrow (or later tonight) I will finish that and then start working with it in a branch.

Looks like everyone made great headway on this, thank you and great work!

bbockelm:
@gmkurtzer - any updates on this one?

src/lib/privilege.c
```diff
@@ -156,6 +163,83 @@ void singularity_priv_escalate(void) {
...
+    struct passwd *pw = getpwnam(username);
+    if (pw == NULL) {
+        singularity_message(ERROR, "Unable to determine UID/GID for user %s: %s (errno=%d)", username, strerror(errno), errno);
+        ABORT(255);
+    }
+    sinfo.uid = pw->pw_uid;
+    sinfo.gid = pw->pw_gid;
+}
+
+void singularity_priv_escalate_singularity(void) {
```
gmkurtzer, line 209: This is about the biggest thing I was able to find... The second mention of 'singularity' seems to be the singularity user? singularity_priv_escalate_singularityuser() perhaps? In either case, this is not worth holding up the merge.

bbockelm: Yeah - it's unfortunate naming, but it's long enough.  I guess we could do "escalate privilege to the singularity user"...

bauerm97: I know this has already been merged, but what do you think about giving singularity_priv_escalate() an input variable? Something like `singularity_priv_escalate(char *user)`. If `char *user` is NULL then we escalate to root, otherwise we escalate to the user specified by `char *user`.

@bbockelm @gmkurtzer

gmkurtzer: That is an interesting idea! My first thought is that I like it, but I would need to go back and look at the logic in singularity_priv_escalate_singularity() and be sure we don't want to mix the code logic as it is nice having them somewhat discrete now. @bbockelm, what is your take?

bbockelm: Well, I like having the API force the programmer to explicitly select one of the three relevant user states:
- The superuser.
- The "singularity user", which has special privileges for images.  Note that the singularity user is possibly distinct from the Unix user named `singularity`.
- The invoking user.

We can easily determine the programmer's intent - even via static analysis! - by looking at the function name.

`singularity_priv_escalate(char *user)` provides too much flexibility.  Stack corruption attacks could cause the semantics of the code (as `char *user` is in read-write memory while the function pointer is currently read-only) and hides the special semantics for the singularity user.

bbockelm: (It's worth noting that we aren't _just_ switching users, but also doing special handling of primary and secondary group IDs.)

gmkurtzer:
We had a discussion some time back (but I can't find the thread) which I think is what generated the "protected mode" feature above. The idea was to provide similar functionality for programs that have execute bit set, but not read for containers.

While, I do see there are numerous use-cases of the protected mode as it stands in this PR, I am still sitting on the fence on two things: Is it appropriate use of SUID, and are those features solving a problem we currently need to solve.

I am still thinking about the first and it will apply to any implementation about overriding POSIX file ACLs, but as far as the second... Going back to the initial motivation of this feature, we were trying to come up with a way to run containers without a user being able to copy them. Metaphorically, we discussed programs that have the execute bit set, but not read/write'able.

Using the same logic, the same permission model as was described for the programs would also work for images (e.g. `chmod 0111 container.img`). This would also simplify things considerably and abides closer to the principal of least surprise.

But... is it a suitable replacement for `protected mode`? @bbockelm Are you using `protected mode` currently or do you have plans to use it in a manner which `chmod 0111` would not accommodate? Does anyone else?

note: I am only discussing `protected mode` not `allow user image`.

Thanks!

bbockelm:
Hi @gmkurtzer  - can you open a separate ticket?  Seems strange to have a conversation on a merged PR.

As with the original PR, I have concerns with us re-implementing filesystem access checks.  You mention checking mode `0111` above, but I would also point out that it bypasses SELinux and POSIX ACLs.

To me, it's more surprising to repurpose one DAC mechanism and bypass other DAC mechanisms and MAC.  The protected mode rules are not as simple but are less likely to incur surprises.

serious-steve:
@gmkurtzer While I am testing the `protected mode` I don't see how the proposed `chmod 0111` would hinder us implementing proper copy protection mechanisms for our containers. However, i thought the idea about `protected mode` was also to have a dedicated non-root user who is accessing images then. With `cmod 0111` will there be a singularity user or not? 
The PR was merged does that mean the feature will appear in the upcoming release of singularity, or are you considering other ways to implement similar functionality? As long as the next official singularity release has `protected mode` or the `chmod 0111` I would be happy :)

gmkurtzer:
@bbockelm Yes, it is strange but this CC's everyone that was part of this discussion and provides context. I probably should have done this before merging, but I didn't want to stifle progress and I didn't have time to review it in gory depth until recently. Apologies for that.

@c1t4r You are right that protected mode gives us the ability to access images as non-root which is a huge advantage. I suppose with `chmod 0111` there could be a singularity user that will be used to open the images but that then brings me back to the core issue that isn't sitting well for me on this...

The fundamental issue I have with this is that it is an abusal of being SUID. For example, if we are installed by a user and run within a user namespace, none of these features will work (granted, neither will the Singularity image itself, but that is within the core functionally of Singularity and one of the primary reasons to always run as privileged (IMHO)).

So... with that said, I'm investigating how we can meet these needs with the least amount of obvious SUID privilege manipulation. Maybe it is not possible, and I am open to that, and I am also open to keeping this as a feature as it stands, but I want to make sure that any possible alternatives and caveats have been discussed.

PR #352, opened 2016-11-28
BF: var/singularity -> var/lib/singularity to please FHS

yarikoptic:
Debian lintian tool has rightfully complained:

`E: singularity-container: non-standard-dir-in-var var/singularity/`

which is indeed the case.  FHS (http://www.pathname.com/fhs/pub/fhs-2.3.html#THEVARHIERARCHY) defines a list of directories to be commonly found directly under /var, and I thought that it would be better to comply with that to not pollute name space there

@singularityware-admin

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -19,7 +19,7 @@ install-perms:
 	@echo
 
 install-data-hook:
-	install -d -m 0755 $(DESTDIR)$(localstatedir)/singularity/mnt
+	install -d -m 0755 $(DESTDIR)$(localstatedir)/lib/singularity/mnt
 
 
 ACLOCAL_AMFLAGS = -I m4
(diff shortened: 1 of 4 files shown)


yarikoptic:
FWIW -- I will hold with uploading 2.2 package into debian until we reach some conclusion on this one

gmkurtzer:
This seems reasonable to me. Let's call it concluded. But... There is also a reference in the master branch's Makefile.am in the root directory, and in the singularity.spec.in. Can you change those too and add them to this PR?

Thanks!

yarikoptic:
thanks... done... although I still feel that I was doing something "not nice" .... e.g. may be using `@runstatedir@` overall for that location would have been more kosher?  I just don't know what stuff should I expect to appear there besides mount points (for that runstatedir sounds more appropriate)...  or there will be some persistent (across reboots) stuff? (images, etc)

gmkurtzer:
It is basically just needed for container mount points. The directory structure needs to be there, but any disposable content should be in the sessiondir.

Thanks again!

PR #355, opened 2016-11-28
BF: bash_completion.d script has bashisms, so use bash as the shell, not generic posix sh

yarikoptic:
Otherwise fails the syntax check (`sh -n`)

@singularityware-admin


bbockelm:
Looks good to me.

PR #356, opened 2016-11-28
Simplify add_http() in utils.py

flx42:
Small simplification + removal of a trailing whitespace.

### DIFF
--- a/libexec/python/utils.py
+++ b/libexec/python/utils.py
@@ -32,12 +32,13 @@ import tempfile
 import tarfile
 import base64
 try:
-    from urllib.parse import urlencode
+    from urllib.parse import urlencode, urlsplit
     from urllib.request import urlopen, Request
     from urllib.error import HTTPError
 except ImportError:
     from urllib import urlencode
     from urllib2 import urlopen, Request, HTTPError
+    from urlparse import urlsplit
 
 # Python less than version 3 must import OSError
 if sys.version_info[0] < 3:
@@ -48,24 +49,21 @@ if sys.version_info[0] < 3:
 ## HTTP OPERATIONS #########################################################
 ############################################################################
 
-def add_http(url,use_https=True):
-    '''add_http will add a http / https prefix to a url, in case the user didn't
-    specify
-    :param url: the url to add the prefix to
-    :param use_https: should we default to https? default is True
+def add_http(url, default_scheme="https"):
+    '''add_http will verify the URL scheme and add a default one if needed
+    :param url: the url to add the scheme to
+    :param default_scheme: the default scheme to add (http or https)
     '''
-    prefix = "https://"
-    if use_https == False:
-        prefix="http://"
-    
-    # Does the url have http?
-    if re.search('^http*',url) == None:
-        url = "%s%s" %(prefix,url)
+    u = urlsplit(url)
+    if u.scheme is "":
+        url = "%s://%s" % (default_scheme, url)
+        # If URL didn't have a scheme, it might have been parsed incorrectly.
+        u = urlsplit(url)
 
-    # Always remove extra slash
-    url = url.strip('/')
+    if u.netloc is "" or u.scheme not in ["http", "https"]:
+        return None
 
-    return url
+    return u.geturl().rstrip("/")
 
 def api_get_pagination(url):
    '''api_pagination is a wrapper for "api_get" that will also handle pagination
--- a/libexec/python/utils.py
+++ b/libexec/python/utils.py
@@ -24,7 +24,6 @@ from defaults import SINGULARITY_CACHE
 from logman import logger
 import json
 import os
-import re
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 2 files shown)


vsoch:
This looks good! Just to keep you in the loop, and @yarikoptic since he has a PR for python in as well, I'm going to write tests for our current python bits, get those in, then we can test on your PRs and go from there. Will keep you guys updated!

vsoch:
@flx42 could you please rebase with the current master? We also want to be sure that if an http url goes in and the user specifies to use https, that it is properly changed.

flx42:
> We also want to be sure that if an http url goes in and the user specifies to use https, that it is properly changed.

Why? As I said in #366, it should only happen if the scheme was not specified. Since we have `use_https=True` and we never override this argument, we would disable all http registries with this change.

vsoch:
sure, fair enough. Please rebase with master.

flx42:
Rebased and cleaned up `extract_tar()` too because you re-added a new dependency on `re`.

vsoch:
Thanks! Although I'll note that `re` is provided standard in [python 2](https://docs.python.org/2/library/re.html) and [python 3](https://docs.python.org/3/library/re.html), meaning it comes default:

      import re
      re.__file__
      '/usr/lib/python2.7/re.pyc'

and python 3:

      import re
      re.__file__
     '/usr/lib/python3.4/re.py'

so it's not a dependency to worry about, akin to something like requests!

vsoch:
Is there a reason for concern in having re? Just curious.

flx42:
No, but it was overkill for checking prefix/suffix.

vsoch:
Why exactly? Regular expressions are simple and powerful. Are you concerned about the code being easy to understand, or being less efficient, or something else?

flx42:
Powerful sure, but clearly not simple to understand.
And yes, it might be less efficient, and it's simply not the right tool for this job IMO.

vsoch:
Thanks for the detail - I appreciate providing it to support your opinion! I too like the clarity that comes with startswith, and think the more powerful regular expressions aren't needed in this case. The tweaks look good to me, @gmkurtzer take a glance and I'm good to merge when you are!

Thanks for your contribution @flx42!

flx42:
And if we can avoid importing one module for now, we should. We can still re-add it later.

flx42:
Ok, cool, thanks!

libexec/python/utils.py
```diff
@@ -57,13 +56,13 @@ def add_http(url,use_https=True):
     prefix = "https://"
     if use_https == False:
         prefix="http://"
-    
+
     # Does the url have http?
-    if re.search('^http*',url) == None:
+    if not url.startswith('http'):
```
gmkurtzer: I would recommend `http://` and/or `https://` to pedantically identify the valid URIs. While hostnames starting with 'http(s)' are not common, it doesn't hurt to be more specific.

vsoch: I think we would want the developer using the function to be able to be specific about this, which is why I first had the `use_https` boolean. The bug here (that I mentioned earlier) is that if a url goes in that is not https, and the user specifies wanting https, then the call above using `startswith` is going to be True, and move on to return the original `http` url. My suggestion before was to do the check specifically for the prefix that was asked for by the user, and regardless of what the url looks like, the function should absolutely return what was asked for.

vsoch: Maybe something using [urlparse](https://docs.python.org/2/library/urlparse.html)?

vsoch: (and note in python 3 it is now under [urllib.parse](https://docs.python.org/3.0/library/urllib.parse.html))

gmkurtzer: Apologies for not knowing the background. I see the discussion above, but I don't have full context.

My only point is that if we are checking to see that this is a valid URI (http or https), we should probably use the appropriate URI prefix. If that isn't needed, I would suggest taking the check out. <shrugs>

vsoch: @flx42 if you don't want to change your approach, you could probably just change what is inside the parens to be the prefix selected by `use_https`

gmkurtzer: @vsoch You are the primary maintainer for the Python portions of Singularity and I am not a Python developer so I defer to you as to the specifics of implementation.

You probably already got this, but my point could easily be covered by any/all of these options (I think) and specifically in the current patched version by doing:

```
if not url.startswith('http://'):
```

(and/or check for https instead).

Hope that helps clarify my perspective. Let me know when you are ready for me to merge into master.

flx42: let's try a better approach with urlparse/urllib.parse then.

vsoch: thanks @gmkurtzer - @flx42 let me mess around a bit and I'll come up with something that you like that fits the bill! Stay tuned :)

vsoch: Exactly my thoughts! +1 +1 +1

flx42:
Ah, didn't see you just did a PR too.

vsoch:
yes it includes yours, so your contributions are included!

flx42:
Yes, but I just amended my patches to use `urlsplit`, you can take a look.

PR #358, opened 2016-11-28
Abort Docker bootstrap if image extraction fails

flx42:
First step for solving #331

### DIFF
--- a/libexec/python/cli.py
+++ b/libexec/python/cli.py
@@ -266,7 +266,10 @@ def run(args):
             layers.append(targz) # in case we want a list at the end
 
             # Extract image and remove tar
-            extract_tar(targz,singularity_rootfs)
+            output = extract_tar(targz,singularity_rootfs)
+            if output is None:
+                logger.error("Error extracting image: %s", targz)
+                sys.exit(1)
             if args.disable_cache == True:
                 os.remove(targz)
                
--- a/libexec/python/utils.py
+++ b/libexec/python/utils.py
@@ -200,11 +200,14 @@ def run_command(cmd):
     try:
         logger.info("Running command %s with subprocess", " ".join(cmd))
         process = subprocess.Popen(cmd,stdout=subprocess.PIPE)
-        output, err = process.communicate()
     except OSError as error:
         logger.error("Error with subprocess: %s, returning None",error)
         return None
-    
+
+    output = process.communicate()[0]
+    if process.returncode != 0:
+        return None
+
     return output
 
 


vsoch:
Hey @flx42 is this still applicable? I can see use cases where the cache runs out of space, or there is a permissions error (or something else we can't anticipate) in which case we may want to abort the bootstrap. If you want to rebase and add a test, I can review. Thanks! :)

flx42:
Well yeah it's always useful to abort in case of an error, instead of silently ending up with an half-baked rootfs
Note sure what kind of test you could write for this, it seems straightforward.

vsoch:
We would just want to make sure that the error is status 1 given that the extraction returns `None`

flx42:
That doesn't seem to be easy to mock in [test.sh.in](https://github.com/singularityware/singularity/blob/master/test.sh.in)

vsoch:
Oh, you are just looking in the wrong place - we have python tests! Add the guy [here](https://github.com/singularityware/singularity/blob/master/libexec/python/tests/test_client.py#L241) :)

flx42:
It's the same, I'm still unsure how you would test that, since the `exit` is in the `main` function. It would mean making this function fail using only the CLI arguments.

vsoch:
yes, so either we do that, or think about if the default of `extract_tar` should be to return None. Should the `sys.exit` be inside the function, or in other words, do you see a case when you wouldn't want a program to fail if an extraction didn't work?

flx42:
`extract_tar` is an utility function, I don't think it should make the decision of aborting. This decision should be made by the calling function, depending on the situation. Maybe the action in this case should be "download again" instead of aborting.

vsoch:
Yes I agree, and it does seem like more hassle than it might be worth to implement the test needing to call the client. however we still need a proper test to make sure `extract_tar` returns None. I just wrote and will push a PR that has yours integrated - thanks for your hard work on this!

bbockelm:
Hi - to be fair, `extract_tar` should probably be rewritten.  It currently forks a sub-process and assumes ; the built-in `tarfile` module is likely a better candidate here.

(I know the PR is already closed - I'll let @vsoch decide if this is worthwhile of a new ticket)

vsoch:
@bbockelm I'd definitely like to chat about this - I was thinking about it when looking over this PR because I've gone back and forth between using tarfile and sending the command to console, and mostly thought they were equivalent (but could we talk about the differences?). So when we fork the sub-process, even if we wait for the process to return (and return status is 0) this is not ideal because there is chance of the timing being off, and [things like this happening](http://stackoverflow.com/questions/8893359/untar-archive-in-python-with-errors)?

PR #363, opened 2016-11-29
Make Docker layer download handle failures and concurrency

flx42:
A simple attempt at fixing #331 to start a discussion, it might not be perfect.

The main drawback is that if the download is interrupted by a `kill -9` or a machine power cycle, we have no way of cleaning up the temporary file.

### DIFF
--- a/libexec/python/docker/api.py
+++ b/libexec/python/docker/api.py
@@ -28,6 +28,8 @@ from utils import api_get, write_file, add_http
 from logman import logger
 import json
 import re
+import os
+import tempfile
 try:
     from urllib.error import HTTPError
 except ImportError:
@@ -295,12 +297,24 @@ def get_layer(image_id,namespace,repo_name,download_folder=None,registry=None,au
         # Update user what we are doing
         print("Downloading layer %s" %image_id)
 
-    response = api_get(base,headers=token,stream=download_folder)
-    if isinstance(response, HTTPError):
-        logger.error("Error downloading layer %s, exiting.", base)
+    try:
+        # Create temporary file with format .tar.gz.tmp.XXXXX
+        fd, tmp_file = tempfile.mkstemp(prefix=("%s.tmp." % download_folder))
+        os.close(fd)
+        response = api_get(base,headers=token,stream=tmp_file)
+        if isinstance(response, HTTPError):
+            logger.error("Error downloading layer %s, exiting.", base)
+            sys.exit(1)
+        os.rename(tmp_file, download_folder)
+    except:
+        logger.error("Removing temporary download file %s", tmp_file)
+        try:
+            os.remove(tmp_file)
+        except:
+            pass
         sys.exit(1)
 
-    return response
+    return download_folder
 
 
 # Under Development! ---------------------------------------------------------------------------------


flx42:
Good luck automating tests for this one ;)

gmkurtzer:
Heya @flx42, sorry about the wait. Just wanted to let you know this PR didn't get overlooked and @vsoch or I will get to it as soon as possible!

Thanks!

vsoch:
hey @flx42 I'll be able to review this in the next few days - just been super busy and need to find some time to think about it. Sorry for the delay!

vsoch:
lol @gmkurtzer are you inside my head?!? :P

gmkurtzer:
Haha, that was pretty cool huh?!

vsoch:
Hey @flx42 could you rebase with master for this one? I am hoping to be able to take a look soon.

flx42:
rebased

vsoch:
awesome, thank you!

vsoch:
ok this first bug looks like an easy fix, and I was thinking it when I first looked at the code. Would it make sense to do a `sys.exit(1)` when the layer fails to download, instead of the RunTime error? We would want to pass that error back to any calling Singularity executables.

vsoch:
and if it fails, should we remove the temporary file too? Does that get handled automatically by tempfile?

flx42:
The temporary file is removed by my code in most cases.
As I mentioned, the only cases where we can't cleanup is when the program stops without giving us any chance to cleanup.

flx42:
sys.exit(1) works too, we can use that indeed.

flx42:
Fixed.

libexec/python/docker/api.py
```diff
@@ -295,12 +297,22 @@ def get_layer(image_id,namespace,repo_name,download_folder=None,registry=None,au
...
-    return response
+    try:
+        # Create temporary file with format .tar.gz.tmp.XXXXX
+        fd, tmp_file = tempfile.mkstemp(prefix=("%s.tmp." % download_folder))
+        os.close(fd)
+        response = api_get(base,headers=token,stream=tmp_file)
+        if isinstance(response, HTTPError):
+            sys.exit(1)
+        os.rename(tmp_file, download_folder)
+    except BaseException as e:
```
vsoch: This looks great! I just tested walking through a layer download manually, and then from command line with a clean install. Two quick edits before merge - could we have a sys.exit(1) here (to replace the `BaseException`) as well, and then for each of the exits, since we lose the information that a specific error would provide, could we have the logger print an error so the user can see the specific error with `--debug`. So for the first maybe something like:

      if isinstance(response,HTTPError):
          logger.error("Error downloading image layer %s, exiting.", base)
          sys.exit(1)

Then for the second, maybe:

      except:
            logger.error("Error creating temporary file %s in cache, exiting.",temp_file)
            sys.exit(1)

Feel free to tweak the language to what you think is most informative.

flx42:
Done

vsoch:
Great job! @gmkurtzer ready for merge.

gmkurtzer:
Sounds good, thanks guys!

Issue #364, opened 2016-11-29
[docker] Export environment of imported images


flx42:
When importing a Docker image, we use the environment of the Docker container. This would be useful when containers install binaries/libraries at non-standard paths.
We probably need to write the output of `get_config(manifest, "Env")` to a file in the rootfs. Maybe `/environment` or a different file.

chrisgorgo:
[docker2singularity](https://github.com/singularityware/docker2singularity) grabs the environment variables from the Docker image (for example set by `ENV` statements in Dockerfile) and exports them in the converted Singularity image. See: https://github.com/singularityware/docker2singularity/blob/master/docker2singularity.sh#L128

This might not be the case if you use singularity directly to boostrap images from docker hub (instead of using docker2singularity).

flx42:
I see, it should be only a few lines of python code to get the same behavior then.

vsoch:
+1

And generally a bigger issue to think about is how to preserve other meta data for the container

oschulz:
I'm working on something similar here: https://github.com/oschulz/singularity-utils/blob/dev/bin/docker2singularity.py

It converts Docker containers to Singularity (FS-directory or SquashFS) via either Docker itself or via docker2aci (no Docker installation/privileges required). It does parse Docker (and ACI) manifests and creates an "/environment" file. Still under development, but already usable (at your own risk ;-) ).

vsoch:
Thanks! We have this functionality in Singularity, eg:

      singularity run docker://ubuntu:latest

Which also does not need the docker daemon. We are just finishing up the environment variable stuffs.

oschulz:
@vsoch , thanks - I was aware of `singularity run docker://...`, but was missing the ENV support. One question, if I may - is there an easy way to import a docker container into a directory (not a .img) without docker and sudo? Also, it would be nice to have a unified way that can both use Docker (to use the local Docker cache and local images) and also go without (for users without Docker access).

As for the ENV handling - will this be done via Python or C? If Python, I'd be glad to contribute the code I have so far (if you still need it).

vsoch:
Hey @oschulz I hear ya! For your first question - the env support (I hope) will be coming along soon. The setup will be implemented in C, and the python bit to add docker support is actually already written, but dependent on that. You can probably guess it means just writing the Docker `ENV` to a file that the C bits know to source.

For directory vs. image, you should be able to exchange the two simultaneously. I usually use images for most things, and directories when debugging as an easy way to look around.

[...]

PR #369, opened 2016-11-29
Use environment of imported Docker images

flx42:
Simple fix for #364: just create a brand new `environment` file, I don't think it's necessary to combine it with the [default bootstrap environment](https://github.com/singularityware/singularity/blob/6932a0f2380885ac984c8d6f81479fc2ecb5787e/libexec/bootstrap/modules-v2/postbootstrap.sh#L68-L79)


vsoch:
I'm also going to close this legacy PR  - please reopen if we missed something!

PR #373, opened 2016-11-29
Fix errno checks

flx42:
Previously the tests were rather... tautological:
```
errno = 0;
if (errno == 0)
    do_something();
```


gmkurtzer:
Good catch!

BTW, I had to look it up so tautological is now the word of the day! Lol

PR #376, opened 2016-11-29
Flx42 docker image environment

### LINKED ISSUE #364: [docker] Export environment of imported images

vsoch:
Fixes #369 and closes #364 

 - adding tests for environment creation function
 - adjusting function to allow user to provide manifest or env, with the env taking preference
 - if user provides a string (eg, gives an env without splitting by newline) we check for this

@gmkurtzer there will be a print of the environment files in the tests - could you please check to see if the PS1 part looks ok? I'm not sure what that variable is, or why there is an extra `>` at the end.
@flx42 please review as well!

@singularityware-admin

### DIFF
--- a/libexec/python/cli.py
+++ b/libexec/python/cli.py
@@ -25,7 +25,8 @@ perform publicly and display publicly, and to permit other to do so.
 '''
 
 from docker.api import (
-    create_runscript, 
+    create_runscript,
+    create_envfile,
     get_config, 
     get_images,
     get_layer, 
@@ -240,6 +241,9 @@ def run(args):
         if args.disable_cache == True:
             shutil.rmtree(cache_base)
 
+        env = get_config(manifest, 'Env').split("\n")
+        create_envfile(env=env,
+                       base_dir=singularity_rootfs)
 
         logger.info("*** FINISHING DOCKER BOOTSTRAP PYTHON PORTION ****\n")
 
--- a/libexec/python/cli.py
+++ b/libexec/python/cli.py
@@ -2,7 +2,7 @@
 
 '''
 
-bootstrap.py: python helper for Singularity command line tool
+cli.py: python helper for Singularity command line tool
 
 Copyright (c) 2016, Vanessa Sochat. All rights reserved. 
 
@@ -241,8 +241,8 @@ def run(args):
         if args.disable_cache == True:
             shutil.rmtree(cache_base)
 
-        env = get_config(manifest, 'Env').split("\n")
-        create_envfile(env=env,
+        # environment is added to ~/environment in the image
+        create_envfile(manifest=manifest,
                        base_dir=singularity_rootfs)
 
         logger.info("*** FINISHING DOCKER BOOTSTRAP PYTHON PORTION ****\n")
--- a/libexec/python/docker/api.py
+++ b/libexec/python/docker/api.py
@@ -1,5 +1,3 @@
-#!/usr/bin/env python
-
 '''
 
 api.py: Docker helper functions for Singularity in Python
@@ -51,13 +49,34 @@ def create_runscript(cmd,base_dir):
     output_file = write_file(runscript,content)
     return output_file
 
-def create_envfile(env, base_dir):
-    envfile = "%s/environment" % base_dir
+
+def create_envfile(base_dir,env=None,manifest=None):
+    '''create_envfile will generate a file called environment
+    in the base_dir of an image
+    :param base_dir: the base_dir to write the file to
+    :param manifest: the manifest to generate the Env for.
+    :param env: a list of environment variables to export, if manifest
+    is not provided
+    '''
+    if env == None:
+        if manifest == None:
+            logger.error("env or manifest must be provided, exiting")
+            sys.exit(1)    
+        env = get_config(manifest, 'Env').split("\n")
+
+    # A user providing an environment not split by '/n' --> error
+    if not isinstance(env,list):
+        env = [env]
+
+    envfile = "%s/environment" %(base_dir)
+    logger.info("Generating environment file %s",envfile)
     with open(envfile, "w") as f:
         f.write("# Docker image environment\n")
         for e in env:
             f.write("export %s\n" % e)
         f.write('export PS1="Singularity.$SINGULARITY_CONTAINER> "\n')
+    return envfile
+
 
 def get_token(namespace,repo_name,registry=None,auth=None):
     '''get_token uses HTTP basic authentication to get a token for Docker registry API V2 operations
--- a/libexec/python/docker/api.py
+++ b/libexec/python/docker/api.py
@@ -51,6 +51,13 @@ def create_runscript(cmd,base_dir):
     output_file = write_file(runscript,content)
     return output_file
 
+def create_envfile(env, base_dir):
+    envfile = "%s/environment" % base_dir
+    with open(envfile, "w") as f:
+        f.write("# Docker image environment\n")
+        for e in env:
+            f.write("export %s\n" % e)
+        f.write('export PS1="Singularity.$SINGULARITY_CONTAINER> "\n')
 
 def get_token(namespace,repo_name,registry=None,auth=None):
     '''get_token uses HTTP basic authentication to get a token for Docker registry API V2 operations
--- a/libexec/python/tests/test_client.py
+++ b/libexec/python/tests/test_client.py
@@ -285,7 +285,8 @@ class TestUtils(TestCase):
         tmpfile = tempfile.mkstemp()[1]
         os.remove(tmpfile)
         write_json(good_json,tmpfile)
-        content = json.load(open(tmpfile,'r'))
+        with open(tmpfile,'r') as filey:
+            content = json.load(filey)
         self.assertTrue(isinstance(content,dict))
         self.assertTrue("Wakkawakkawakka" in content)
 
--- a/libexec/python/tests/test_docker.py
+++ b/libexec/python/tests/test_docker.py
@@ -53,6 +53,7 @@ class TestApi(TestCase):
         with some command
         '''
         from docker.api import create_runscript
+        print("Testing api.create_runscript to create runscript...")
         cmd = "echo 'Hello World'"
         base_dir = tempfile.mkdtemp()
         runscript = create_runscript(cmd=cmd,
@@ -62,6 +63,48 @@ class TestApi(TestCase):
         # Commands are always in format exec [] "$@"
         # 'exec echo \'Hello World\' "$@"'
         self.assertEqual('exec %s "$@"' %cmd,generated_cmd)
+
+
+    def test_create_envfile(self):
+        '''test_create_envfile should test if a list of environment commands
+        obtained from a manifest return a proper file
+        '''
+        from docker.api import create_envfile
+        from docker.api import get_manifest
+        from docker.api import get_config
+        
+        print("Testing generation of environment file...")
+        print("Case 1: Generating environment from manifest")
+        manifest = get_manifest(repo_name = self.repo_name,
+                                namespace = self.namespace)
+        
+        # Env is a variable in the manifest config
+        envfile = create_envfile(manifest=manifest,
+                                 base_dir=self.tmpdir)
+        self.assertTrue(os.path.exists(envfile))
+        generated_env = read_file(envfile)
+        print("".join(generated_env))
+        
+        print("Case 2: Generating environment from user string")        
+        env = get_config(manifest,spec="Env")      
+        envfile = create_envfile(env=env,
+                                 base_dir=self.tmpdir)
+        self.assertTrue(os.path.exists(envfile))
+        generated_env = read_file(envfile)
+        print("".join(generated_env))
+
+        print("Case 3: Generating environment from user list")        
+        env = get_config(manifest,spec="Env").split('\n')      
+        envfile = create_envfile(env=env,
+                                 base_dir=self.tmpdir)
+        self.assertTrue(os.path.exists(envfile))
+        generated_env = read_file(envfile)
+        print("".join(generated_env))
+
+        print("Case 4: User does not supply manifest or env")        
+        with self.assertRaises(SystemExit) as cm:
+            envfile = create_envfile(base_dir=self.tmpdir)
+        self.assertEqual(cm.exception.code, 1)
 
 
     def test_get_token(self):
--- a/libexec/python/utils.py
+++ b/libexec/python/utils.py
@@ -1,5 +1,3 @@
-#!/usr/bin/env python
-
 '''
 utils.py: python helper for singularity command line tool
 
--- a/libexec/python/utils.py
+++ b/libexec/python/utils.py
@@ -282,9 +282,8 @@ def write_file(filename,content,mode="w"):
     and properly close the file
     '''
     logger.info("Writing file %s with mode %s.",filename,mode)
-    filey = open(filename,mode)
-    filey.writelines(content)
-    filey.close()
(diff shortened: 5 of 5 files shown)


vsoch:
Here is the file that is generated, I'm wondering about the `PS1` and extra `>` how is this supposed to work when the image is run?

      # Docker image environment
      export PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
      export PS1="Singularity.$SINGULARITY_CONTAINER> "

vsoch:
I'm adding a fix for this:

      tests/test_client.py:281: ResourceWarning: unclosed file <_io.TextIOWrapper name='/tmp/tmpwvqmh_' mode='w' encoding='UTF-8'>
  write_json(bad_json,tmpfile)

bauerm97:
@vsoch We should discuss interaction between the new bootstrapping module and this as well. I have introduced the concept of doing `%environment` in the bootstrap file to create an environment file. This happens *after* the docker module is run. Should the user environment code overwrite this code? Should the user have to specify `IncludeEnv` like with `IncludeCmd`? I think there are probably more things to consider big-picture wise. 

Currently in bootstrapping (in my C code at least), the /environment file has those two lines that you've listed and is generated either from a default file format, or from the user specified `%environment` script. Maybe it would make sense to store the docker environment exports in a separate file, perhaps `/.docker_environment`. Then we can ask the user to specify `IncludeEnv` in their bootstrapping file, and we automatically append `source /.docker_environment` into the environment file.

vsoch:
Hmm, interesting. Ok, so I think the functionality that we want is that, given that a user bootstraps a docker image, they get with it exactly the environment that it had. They shouldn't have to specify anything additional in the bootstrap file (because think of the cases when you get into a docker image via `run` or `shell`). In this case, if there is some concern about overlap between docker and original singularity environments, what about having a folder-based strategy - in the root we have a hidden folder of (some number) of environment files, and the default behavior is to source all of them? That way, a user could even write a custom environment file and be assured that it would be sourced / added. Then we also wouldn't need to hard code the specific ones (eg, docker, singularity hub, whatever else comes in) as we add different kinds of imports. Thoughts?

vsoch:
And if we give the user control, it should be the other way around, that they must put a flag to NOT include a particular environment. This I suggest because my expectation for the default would be to include the same environment from the image I'm bootstrapping. And it shouldn't be docker specific, it should work with any file that is in the environments folder.

libexec/python/docker/api.py
```diff
@@ -51,13 +49,34 @@ def create_runscript(cmd,base_dir):
...
+    :param base_dir: the base_dir to write the file to
+    :param manifest: the manifest to generate the Env for.
+    :param env: a list of environment variables to export, if manifest
+    is not provided
+    '''
+    if env == None:
+        if manifest == None:
+            logger.error("env or manifest must be provided, exiting")
+            sys.exit(1)    
+        env = get_config(manifest, 'Env').split("\n")
```
flx42, line 65: It seems a little weird to me to have all these optional arguments everywhere, it makes the interface of the function more difficult to understand IMO. I have troubles understanding how they will be used.
Do we really need to make `env` optional? The file is `docker/api.py`, should it allow setting a custom environment unrelated to Docker?
`base_dir` + mandatory `manifest` would be enough.

libexec/python/docker/api.py
```diff
@@ -51,13 +49,34 @@ def create_runscript(cmd,base_dir):
...
+
+    # A user providing an environment not split by '/n' --> error
+    if not isinstance(env,list):
+        env = [env]
```
flx42, line 69: That's weird, see my comment above, maybe we shouldn't allow the user to pass anything she wants to this function.

libexec/python/cli.py
```diff
@@ -241,8 +241,8 @@ def run(args):
         if args.disable_cache == True:
             shutil.rmtree(cache_base)
 
-        env = get_config(manifest, 'Env').split("\n")
-        create_envfile(env=env,
+        # environment is added to ~/environment in the image
```
flx42, line 245: `/environment`, not `~/environment`, right?

flx42 (changes requested):

vsoch:
eg:

    environments/
        environment # (produced via image default environment, for example, here we want the PS1 stuff - this file should be wipable and exchangeable for an updated version without mucking up the applications inside or the user's customizations)
        environment_docker # (produced via docker)
        environment_bootstrap # (produced (and updated) via  %environment)
        environment_custom # (any other future user / custom integration)

How are the files getting sourced? It wouldn't need to be that exact list, but something along those lines?

bauerm97:
I agree that we should come up with some strategy of environment file(s) such that we maximize the degree of forward compatibility that we will have. Currently we source the /environment file in `/.run`, `/.shell`, and `/.exec` before continuing with normal operation. I think this is the right place (in the `/.run` etc... files) to inject the code. If we can settle with a format together, I can inject the code into the correct place in the bootstrapping process.

libexec/python/docker/api.py
vsoch: It's because as a developer, I don't find it intuitive to get "env" via get_config(spec="env") and then have to split that by newline before giving to the function to generate the environment. I do, however, find it intuitive to grab a manifest, and then generate an environment from it.

flx42:
A strategy for environment files seems too narrow in scope, why not a folder with a set of scripts to execute? init style.
```
/.init/00-user-environment
/.init/10-docker-environment
/.init/10-check-system-script
[...]
```

bauerm97:
@flx42 I think this is what @vsoch was suggesting (and what I meant by strategy, just come up with some consistent way to handle them)

Semantics aside, I do like this idea. We store a bunch of files to be sourced in `/.environment/*`, and I will make sure in my bootstrapping code that those files will get sourced on `run shell exec` commands.

libexec/python/docker/api.py
flx42: Fine, having the `get_config("env")` before calling `create_envfile` was not ideal. But you can still do that in `create_envfile` and removing `env` as an argument.

flx42:
@bauerm97 sure, but I'm suggesting we shouldn't limit that to just environment files. It could also store scripts for checking the container, starting services, etc.

libexec/python/docker/api.py
vsoch: If the user (developer) gives a non-list (as the default field is returned as a string from the get_config function) then this will fail. This isn't a pretty line, but it makes sure that the function is robust.

vsoch:
I really like this line of thinking! Can we extend it to include a flexible way of defining metadata for a container?

libexec/python/docker/api.py
flx42: Then fix the comment then, it's not an error, and it's not actually checking if it was splittable by `\n`.

bauerm97:
I like this line of thinking as well, but I'm not sure how this will interact with forward and backward compatibility of images. I would like to avoid breaking old images or breaking new images on old code if possible. This means that whatever we come up with, we should only need to insert code into the `/.run` `/.exec` and `/.shell` files so that the workflow remains the same. I think we can do what you're suggesting this way @flx42

vsoch:
For executable scripts, the `.init` folder seems appropriate. @flx42 the number is to distinguish an order to load, or something else? We could make it a standard to have a (three digit?) number that would specify the order to load, eg:

      /.init/000-user-environment
      /.init/001-docker-environment
      /.init/002-check-system-script

It wouldn't break things to have equivalent numbers, they would just be read in based on the ordering of the files with `ls`. Then we would want to give the user the power to disable one of these scripts from running, maybe with some header?

      %scripts
      disable 001-docker-environment

 If multiple images are bootstrapped, or we have some kind of layering of environments (or actually, any script file with the same name) you would get something like:

      /.init/000-user-environment
      /.init/001-docker-environment
      /.init/001-docker-environment-01
      /.init/001-docker-environment-02
      /.init/002-check-system-script

@bauerm97 - maybe for conserving the `.shell` and `.run` and `.exec` - those files are generated by way of compiling the different `.init` files. This way, the different builders for the final environment are preserved, and a user could find an image and build it again (eg, producing a different version of `.run`) based on enabling or disabling some scripts.

vsoch:
And for metadata, I think a stupid flat file solution might be not so crazy, given that metadata is unlikely to be more than maybe 100 things. We could have a root directory to go along with `.init` called `.meta` that has files (named by the variable) with whatever the metadata is inside.

flx42:
Yes the number prefix defines the order, it's pretty standard to have two digits, for instance:
```
$ ls /etc/rc0.d/
K10unattended-upgrades  K20gunicorn    K20munge  K20rsync           K20slurm-llnl         K25mdadm           K99sep4_0  S20sendsigs  S31umountnfs.sh  S60umountroot
K20cachefilesd          K20kerneloops  K20nginx  K20screen-cleanup  K20speech-dispatcher  K98mdadm-waitidle  README     S30urandom   S40umountfs      S90halt
```

gmkurtzer:
I am just looking at this quickly, but it seems to me that the initial problem we are solving is straight forward (but correctly me if I am oversimplifying or misinformed).

As part of bootstrapping we create a file called `/environment` which we put all of the environment variables and configurations specific to a particular container. The Docker modules also *may* have a defined series of environment variables. At first I thought that we should supersede the `/environment` created at bootstrap, but what I think we should do is append to it. 

Now where the conversation is headed now is quite different... We could go to a completely shell based modular bootstrap subsystem that is run in order of nomenclature (e.g. sysVinit). I'm fine with that and it will simplify some things considerably.

flx42:
@gmkurtzer right, this discussion is now largely off-topic.

Why append to `/environment`? The [default /environment](https://github.com/singularityware/singularity/blob/6932a0f2380885ac984c8d6f81479fc2ecb5787e/libexec/bootstrap/modules-v2/postbootstrap.sh#L68-L79) seems to bring little value.
And actually, I don't think we can append, when we're in the python code the file doesn't exist yet. So we could only prepend.

vsoch:
I'm ok with doing what the group thinks is best, but I am in support of preserving the different environments that go into an image with a (slightly more complicated) `.env/` folder. As we discussed on slack, this would allow for future integration of a container having a context, e.g., giving the user or applications the ability to specify a different context (environment, runscript, etc.) for the container to use. @bauerm97 is sleeping, let's wait for him to pick this up again :)

gmkurtzer:
@flx42 We just had an Slack conversation on the matter. We decided to use environment files within /.env/* that are bootstrap module/layer specific. Each file within /.env/* will be sourced by the respective run, shell, exec, test scripts in /. This makes it easy to overwrite and keep things clean without affecting any other environment files.

I didn't like this at first because of the overhead of globbing and sourcing files through a loop (trying to keep startup overhead down to an absolute minimum and I didn't see the point), but @vsoch and @bauerm97 made a good case for doing it this way.

bauerm97:
@gmkurtzer We might be able to do this *without* requiring searching and sorting of the files in `/.env/` at container runtime, but rather at container buildtime. Here is my proposal on how that could be accomplished:

 - Create `/.env/metadata` (or whatever we want to name it) which will just be a text file
 - Every time a new file in `/.env/` is created, the creating process is responsible for placing it into `/.env/metadata` in the proper order (IE 02-docker would be placed in between 01-* and 02-*)
 - `/.run` & co. will source that metadata file (this preserves the current format of those files better) which in turn already contains a list of precedence ordered source commands.

Obviously if we go that route, it will mean that the `/.env/` folder is "less dynamic" in nature, and whoever puts a new file in `/.env/` has to be conscious to update the metadata as well. We also need to make sure that we establish a set of rules for which files get to be updated before other files, and who gets to add environment files and the metadata file. I'm not really sure this is worth it, because the performance loss from just sorting through the `/.env/` folder each run is probably incredibly minimal.

EDIT (Best Option IMO): **Alternatively, the Singularity C code could be in charge of maintaining the metadata file. After the bootstrap is done, the C code could generate the metadata file based on what's in the container. As well as that, every time somebody runs a Singularity container with `--writable`, after they exit the container the C code will again sort through `/.env/` and recreate the metadata file. This would ensure that the `/.env/` file never changes (if you're not bootstrapping or running in `--writable`, the files can't change!) without the metadata file changing to update it.**

**I think this is actually the best option. Users don't need to know about the metadata file, they simply need to know that any file place into `/.env/` will be sourced based on some naming convention that we will determine. This also means that we do the computationally expensive work (searching/sorting) only once, while still ensuring that at runtime we will always source the exact contents of the folder**

@flx42 We could do prepend/append of docker vars, since after the docker module is done running, the C bootstrapping code will still be able to make changes. After all bootstrapping code is run, I could simply merge the docker-env into the environment file. I do think that doing it the `/.env/` way might just be simpler and better defined behavior, however.

gmkurtzer:
@bauerm97 That seems reasonable to me too. Granted, it means opening and reading in another whole file! OMG!! haha... But that is surely less expensive then a directory glob and sort, but it is also less intuitive if someone specifies the --writable option as you suggested. At this point, I'm fine with it in either way and thanks for coming up with this model!

vsoch:
@bauerm97 I'm really excited about this! I've been thinking of a use case of having "on demand containers" and then being able to record a user's work and save it. This idea fits nicely with that, because it would mean that we could have this scenario where (for example an HPC cluster) gives users access to writable containers, the writable container has a simple process running to record (some aspect) of their workflow into an executable. Then the process would just have to write this stuffs to the `.env` folder, and when the session closes it would be recorded and kept with the image! 

How would be best to move forward with developing this? I can definitely help with logic of file sourcing, etc, and we already started that a bit above. If it's in C you probably would want to implement that first, and then @gmkurtzer look it over? Once those bits are in place, I can tweak this PR to write the environment stuffs to fit this schema, and we can test to be sure that it works as expected. We might then want to add a few tests for the CI. Let me know your thoughts!

bauerm97:
@vsoch There are two pieces of this that I will need to write in C:
1) After bootstrapping is done and before the process ends, the `rootfs/.env/` directory needs to be parsed
2) After a container is run in writable mode, we also need to parse the `rootfs/.env/` directory

I will write a function in C that will handle the parsing of `rootfs/.env/` based on a set of rules. However, we need to come up with those rules. I was thinking that every file needs to follow the format `XX-*` where `XX` is a two digit number and `*` is a string descriptor. The larger the number, the later the file will be sourced. `01` will be sourced before `02` and etc... therefore a higher number has a higher priority.

The only issue with doing all of this now, is that it essentially hinges on the bootstrap being done in C (otherwise we can't cleanly execute C code after the bootstrap). I will write the code on a branch based off of my bootstrap refactor code, but many things are still changing in the code so I can't be 100% sure it's the final version.

bauerm97:
@vsoch I have pushed code that will allow the handling of this here: https://github.com/singularityware/singularity/pull/353/commits/f442dcbf3c738b4e7c170b7040e007ae2e997f48 Please feel free to check it out, however it has not actually been put to use, it just is implemented and ready to go.

vsoch:
Awesome!  I'll be able to look later today... we are hosting folk and it will eat up most of the work time today (sad face) Looking forward to working on this with power me through! :)

vsoch:
@bauerm97 let me know when you've had a chance to fix up the `.env` stuff, and no rush, I know you've been feeling under the weather!

vsoch:
I think we have this taken care of now with development - I'm going to close, please ping and reopen if needed!

PR #377, opened 2016-11-30
Rearranged src/lib/mount/cwd/cwd.c to account for edge cases

### LINKED ISSUE #362: Ignoring user bind request: host does not support PR_SET_NO_NEW_PRIVS
Hi

My IBM iDataPlex has the same issue as the Cray at IU.  RHEL 6.4 2.6.32-431.el6.x86_64
My singularity image built on a different system: CentOS Linux release 7.2.1511 
Running any command like this one:
singularity exec centos-mpi.img ls /
returns:
WARNING: Not mounting current directory: host does not support PR_SET_NO_NEW_PRIVS

So I added this to the config:
bind path = /extra
and changed this:
user bind control = no

My testing is working as the image mounts /extra but I am constantly getting:
WARNING: Not mounting current directory: user bind control is disabled by system administrator
Even if I run the command from my home directory and not /extra
Is my configuration still incorrect?  

Thanks
Chris

bauerm97:
Fixes #362 (kind of, this is just me suggesting how the logic should flow in cwd.c)

 - Rearrange the logic flow in cwd.c to account for fringe cases
 - Fix minor memory leaks here and there

### DIFF
--- a/src/lib/mount/cwd/cwd.c
+++ b/src/lib/mount/cwd/cwd.c
@@ -16,11 +16,7 @@
  * to reproduce, distribute copies to the public, prepare derivative works, and
  * perform publicly and display publicly, and to permit other to do so. 
  * 
-*/
-
-#ifndef _GNU_SOURCE
-#define _GNU_SOURCE
-#endif
+ */
 
 #include <errno.h>
 #include <fcntl.h>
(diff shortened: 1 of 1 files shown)


gmkurtzer:
I need to look at the code on a real computer (not from iPhone) but one aspect I don't like is that on hosts that don't support or set no new privs users will always see that warning all the time, no matter what. With the order I have currently, the message is only seen if it needs to mount the cwd but can not due the lack of that feature.

bauerm97:
@gmkurtzer Good point. Somehow the tests aren't passing and I have no idea why, so maybe I'll just scrap this in the end.

bauerm97:
@gmkurtzer I will actually close this for now, I think my head was in the right place (that maybe cwd.c could be tidied up a bit), but I broke something and it's probably better to just redo from the beginning later and go less overboard ;)

gmkurtzer:
Let's try and sync up in slack. There is actually a lot of things I want to cleanup and we can add this to the list. Thanks!

PR #378, opened 2016-11-30
Flx42 python add http cleanup

vsoch:
Fixes #356 and integrates and closes #356 to integrate @flx42 contributions to the functions

 - add_http function reliably returns the correct URI asked for the by the user, given that http, https, or no URI are supplied. It does this by using `urlparse` to parse the url and add the extension to what is requested to the rest. 
 - the tests have been updated to test for all these cases

@singularityware-admin

### DIFF
--- a/libexec/python/tests/test_client.py
+++ b/libexec/python/tests/test_client.py
@@ -68,27 +68,44 @@ class TestUtils(TestCase):
     def test_add_http(self):
         '''test_add_http ensures that http is added to a url
         '''
-        print("Case 1: adding https to url with nothing specified...")
 
         from utils import add_http
-        url = 'registry.docker.io'
+        url_http = 'http://registry.docker.io'
+        url_https = 'https://registry.docker.io'
+
+        print("Case 1: adding https to url with nothing specified...")
 
         # Default is https
+        url = 'registry.docker.io'
         http = add_http(url)
-        self.assertEqual("https://%s"%url,http)
+        self.assertEqual(url_https,http)
 
         # http
         print("Case 2: adding http to url with nothing specified...")
         http = add_http(url,use_https=False)
-        self.assertEqual("http://%s"%url,http)
+        self.assertEqual(url_http,http)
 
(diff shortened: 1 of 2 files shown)


vsoch:
@gmkurtzer ready for merge! Thanks @flx42 ! Good teamwork.

flx42:
You don't remove the trailing `/` anymore, and you are not preserving `http`.

vsoch:
What do you mean not preserving http? it always returns http or https, which is what we want. The trailing `/` is a good call - let me add that back in!

flx42:
You always discard the user-provided scheme, I thought we agreed not to do that?

vsoch:
@gmkurtzer mentioned that it should be controlled to http or https, so I think that's reasonable to do now, and using a boolean (and not an open string as default argument) is a good way to do that. We can easily add support for other URIs if/when we need them.

flx42:
My updated PR verifies it's `http` or `https`, rejecting other schemes, but it's also extensible to other schemes in the future by passing an open string.

vsoch:
It's generally not a good idea to use mutable objects in function arguments, because they are created at the time the function is created (not when called) which can result in weird behavior. I think this is why you tend to see defaults set as `None` and then set in the function when called. I haven't read a ton about it, but [here is a good starting point](https://pythonconquerstheuniverse.wordpress.com/2012/02/15/mutable-default-arguments/). So if we need this functionality in the future, we will find a different way to implement it. I greatly appreciate your input and general contributions @flx42. @gmkurtzer this is ready for merge.

gmkurtzer:
Will do, thanks!

flx42:
Actually, a string is **not** a mutable object, so that's fine.
If you really want to have an argument called `use_https`, then you should really enforce it and raise an error for both of these cases: `add_http("http://example.com", use_https=True)` and `add_http("https://example.com", use_https=False)` because that looks like a logic error.

flx42:
Thanks for merging, we're making good progress!

PR #381, opened 2016-12-02
Overhaul configuration system.

### LINKED ISSUE #357: Cleanup configuration and defaults
The current configuration subsystem is fairly brittle.  Ideas to make it more robust
1.  The "default value" of a parameter is in both the code and the configuration file.  Updates must be applied to both places (we've had cases where this was forgotten).  We could have the defaults in a central place in the code and have the config file be generated from the code.
2.  The defaults and the local configuration are in the same file.  That means any admin with non-standard configuration won't get updates.   A more traditional setup would be to split the two and have an `#include` command of some sort.
3.  Any time we want to read out a config value, we must call `rewind` on the file descriptor.  This is brittle: there's nothing stopping us from forgetting to call the mandatory rewind.  At the minimum, we should move the rewind to be inside the configuration `get_value` call; likely, we should parse everything to a hash table and just use values from the hash.

bbockelm:
Fixes #357

@singularityware-admin

Three major improvements:
- Parse configuration once and keep it in a hash table.
- Keep configuration default values in a header.
- Auto-generate the default configuration file from the values in
  the header.


bbockelm:
DO NOT MERGE YET - I still have to do more testing to do.  I just want @gmkurtzer and @bauerm97 to see the approach I wanted to take.

src/configure_transform.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/python
+
+import os
+import re
+import sys
+
+define_re = re.compile("#define ([A-Z_]+) (.*)")
+
```
vsoch, line 13: Why is this done in python?

src/lib/config_parser.c
```diff
@@ -28,105 +28,189 @@
...
-                    chomp(config_value);
-                    singularity_message(VERBOSE2, "Got config key %s (= '%s')\n", key, config_value);
-                    free(line);
-                    return(config_value);
-                }
-            }
-        }
+static const char *_default_entry[2];
+
+const char **_singularity_config_get_value_multi_impl(const char *key, const char *default_value)
```
bauerm97, line 245: Word of caution here: the current system will parse one value at a time, and return the second, third, etc... values on consecutive calls. Changing this to a `char **` will break any current features that expect/allow >1 value. This is obviously the proper way to do this with a has table (in fact keeping the old format is much more trouble than it's worth), but I just wanted to point out that we will have to make sure to thoroughly comb through the rest of the code to fix this!

src/util/util.c
```diff
@@ -177,36 +177,27 @@ void chomp_noline(char *str) {
...
-    // If str starts with a new line, there is nothing here
-    if ( str[0] == '\n' ) {
-        str[0] = '\0';
-    }
-
-    // Trim trailing new line characters
-    if ( str[len - 1] == '\n' ) {
-        str[len - 1] = '\0';
-    }
-
```
bauerm97, line 209: Watch out when removing this code, in my experimentation it actually broke things. That's the reason that chomp_noline exists. We could either:

1) create chomp_newline() which specifically trims the leading and trailing newline characters only
2) Leave the newline trimming in chomp and copying the whitespace logic to chomp_noline()

I think option 1 is probably the smarter choice.

src/lib/image/mount/mount.c
```diff
@@ -56,7 +57,7 @@ int singularity_image_mount(int argc, char ** argv) {
     }
 
     singularity_priv_init();
-    singularity_config_open(joinpath(SYSCONFDIR, "/singularity/singularity.conf"));
+    singularity_config_init(joinpath(SYSCONFDIR, "/singularity/singularity.conf"));
```
bauerm97, line 60: We can move remove all the config_init calls from `lib/image/*/*.c` and simply move it into one call in `src/simage.c`

src/lib/singularity.h
```diff
@@ -158,14 +158,4 @@
...
 
-
-
-    extern int singularity_config_open(char *config_path);
-    extern void singularity_config_close(void);
-    extern void singularity_config_rewind(void);
-
-    extern char *singularity_config_get_value(char *key);
-    extern int singularity_config_get_bool(char *key, int def);
-
```
bauerm97, line 169: Is there a reason we don't do the same definitions and declarations here as we do in `config_parser.h`? It's currently this way for all header files that implement `singularity_*` functions (singularity.h includes all `singularity_*` functions)

bauerm97 (changes requested):

src/lib/config_parser.c
bbockelm: Yup, this PR changes all the config APIs, so every single call site had to be manually examined and updated with the correct function (either single or multi-valued).

src/configure_transform.py
bbockelm: Why not?  Seems easier than parsing and transforming in bash...

src/util/util.c
bbockelm: I'm not sure I follow.  This still removed newlines as isspace counts \n as a space character.

src/lib/singularity.h
bbockelm: I think we can have this file include others: I just don't see the value in copy/pasting the same definition twice.  Seems like it'd lead to errors.

src/lib/config_parser.c
bauerm97: Yea, I wrote that before looking at the rest of the PR. Just wanted to point it out anyways :)

src/lib/singularity.h
bauerm97: I'm not sure what the benefit is, but whatever we do decide on we should make sure it's consistent across all files and functions

src/util/util.c
bauerm97: You are following just fine, I didn't realize that the newline character is also caught by issapce.

src/configure_transform.py
vsoch: If it's kosher to have a random python script, then I don't see why not! Two more things to think about. Would it be more appropriate to have some kind of organizational logic other than just having in `src`? If we do decide to move forward with python for this, would you mind some help to make it a bit more function based and script-ish? (Eg, using argparse instead of sys.argv, and having functions. *raises hand* I would be glad to help with this!! :OD

bbockelm: Is `argparse` available on RHEL6?

Anyhow, no problem in sprucing it up as much as you'd like (I think you should be able to commit directly to the branch); I just didn't put much emphasis on this as it is a build script and not something that gets shipped.

vsoch: ok, sounds good! I'm being a bit slow today, but will be on campus within an hour or so.

vsoch: maybe just a bug for [ipython](https://bugzilla.redhat.com/show_bug.cgi?id=874133)? We can always test it out.

vsoch: ok, working on this now! :)

bbockelm:
Alright, I think this is ready to merge!  The new config directive is:

```
%include UNIX_GLOB
```

Where the `UNIX_GLOB` can match multiple files (or just one).

vsoch:
Before you do, see https://github.com/bbockelm/singularity/pull/1

vsoch:
I think I'm missing someone's branch? There are changed files that I didn't change in there!

bbockelm:
@vsoch - this breaks the build on RHEL6 (need to separately install the `python-argparse` RPM).  Can you fixup the RPM spec file?

vsoch:
I don't understand - what is the RPM spec file? We have argparse in the docker python scripts - is this an uncaught issue as well?

gmkurtzer:
@vsoch can you use a different method to obtain the program/script arguments such that we don't have to require anything specifically.

@bbockelm I don't want to add any static requirements to the SPEC (if possible).

vsoch:
@bbockelm could we talk about what bugs / docs you run into with using argparse? The bug I linked says it's specific to using ipython, because the library [comes standard in python](https://docs.python.org/2/library/argparse.html). Under what specific OS / situation is it not found?

gmkurtzer:
@vsoch: I think @bbockelm's point was that it does not come as part of the main Python source on RHEL6. It looks like argparse was included into Python proper sometime after RHEL6's version, so it is only a problem on slightly older distributions.

vsoch:
ok, I can fix that when we get this other clump of PRs through - it's a lot easier to start from a fresh base than deal with all the merge conflicts that would likely arise! See https://github.com/singularityware/singularity/issues/389

vsoch:
If that's the only issue, I think it would be ok to not let it hold up this addition, because the bug with argparse is already distributed and running around like a wild chicken in the production version.  When I fix the argparse used in the main python client I'll do the same for this portion as well!

gmkurtzer:
This PR will take me a little bit to go through and merge. I'll try to get it done by later today!

bbockelm:
@gmkurtzer - sounds good!  Since this touches a large number of files (the call signatures of all interaction with the configuration have changed), merging it sooner rather than later will help prevent merge conflicts.

gmkurtzer:
I am seeing conflicting files for this merge. As soon as those are resolved, I'll merge.

Thanks!

bbockelm:
Merge conflicts resolved - basically, it was just limited to new invocations of the old configuration API from the protected image mode pull request.

vsoch:
As a sister/brother/sibling to this PR, could you send me in some fashion an update [to this list](http://singularity.lbl.gov/docs-config) so the docs can stay up to date with all the recent config changes?

bbockelm:
@vsoch - I even discovered a few of these that weren't even in `singularity.conf`.  (FWIW - this branch should not have changed any configuration defaults or added new keys)

What's the corresponding github page at the `docs-config` URL?  Can we auto-generate that?  In particular, it'd be handy to force the defaults in the documentation and in the code line up.

bbockelm:
Argh, some of today's merges created fresh conflicts.

@gmkurtzer - any thoughts as to when you'll get to this one?  I would prefer to not have resolve conflicts every day....

vsoch:
@bbockelm [it's this file](https://github.com/singularityware/singularityware.github.io/blob/master/pages/docs/admin-docs/advanced-config.md). Technically, we could do something that nutty! It would mean pulling that repo, making the change automatically, and then pushing it to master. I'm a bit nervous about this approach, what are your thoughts?

The docs (for the config) probably won't be updated too often - what about a solution in the other direction (eg, the static doc file is checked by a test in continuous integration, fails if they are different, tells us to update them?)

This is why sphinx is so great... *sigh*

vsoch:
haha @bbockelm on your last comment... poor @gmkurtzer ! And poor @bbockelm too... :ice_cream: anyone?

gmkurtzer:
I am on it.

bbockelm:
I like the idea of having Travis throw an error if the docs and code get out-of-date (although how do we do updates then?)

gmkurtzer:
Will take me just a few more... Found some build issues (make dist tarballs were failing) and simage.c needed some help. Testing now...

vsoch:
The @vsoch bot does the updates! Or the @gmkurtzer bot if the @vsoch bot is sleeping. She hardly ever is!

gmkurtzer:
Haha @vsoch.

BTW, this is now merged and GitHub worked this time! YAY!

vsoch:
ok, I'll make an issue to get something working for the config docs - for now @bbockelm are you the best one to generate an updated list (and I'll make sure the docs are on point)

PR #383, opened 2016-12-02
Restore 'expected' signal handling rules for the payload.

bbockelm:
@singularityware-admin

Without this, the payload process - when PID namespaces are used -
gets PID 1 rules for signal handling.  This can lead to surprising
results for end-users.

This utilizes an additional `sexec` process inside the PID namespace as a shim process for signal handling.

### DIFF
--- a/src/lib/ns/pid/pid.c
+++ b/src/lib/ns/pid/pid.c
@@ -93,6 +93,13 @@ int singularity_ns_pid_unshare(void) {
     // PID namespace requires a fork to activate!
     singularity_fork_run();
 
+    // At this point, we are now PID 1; when we later exec the payload, it will also be PID 1.
+    // Unfortunately, PID 1 in Linux has special signal handling rules (the _only_ signal that
+    // will terminate the process is SIGKILL; all other signals are ignored).  Hence, we fork
+    // one more time.  This makes PID 1 a shim process and the payload process PID 2 (meaning
+    // that the payload gets the "normal" signal handling rules it would expect).
+    singularity_fork_run();
+
     return(0);
 }
 


gmkurtzer:
Yep, that is pretty much exactly what I was thinking of doing as well... Thanks!

gmkurtzer:
Do we want to rename the sexec binary in the PID table so it doesn't show up as sexec?

If so I've had mix results rewriting argv[0], but I think the prctl() method might be better. I know processes can rename themselves reliably (e.g. in perl: $0 = "new process  name";), but not sure how to do it in C. are you familiar?

bbockelm:
Instead of just renaming the `sexec` binary in the PID table, we ought to actually have a separate executable (see #384).  This would reduce the amount of unnecessary code in the image for these "sleeper" processes.

gmkurtzer:
Hrmm, yes... that might be possible indeed! We will need to come up with a way for the sleeper process to know when the foreground process has exited and for it to exit too. signals handlers perhaps?

bbockelm:
Which process do you refer to as the "foreground" process?  I think of them as the "payload" and "invoking" process.

When the payload process, its parent (`sexec` currently) receives a `SIGCHLD` and the existing signal handlers fire (call `wait` to reap the process and then exit with the appropriate code).

gmkurtzer:
The foreground process is the application that is called from within the container. Sexec can actually be exec'ed out of existence when no session directory cleanup or PID namespace is requested. In the case of the PID namespace, we will need to fork/exec a "sleeper" process (which is a mini init) and it must be able to communicate (or at least receive) signals from sexec.

bbockelm:
Sounds good - anyhow, let's move the conversation over to #384 as this PR is closed.

PR #392, opened 2016-12-06
updating python tests to remove tag

vsoch:
This is a quick patch to update the python tests - for custom registries, it looks like the standard manifest is changing (and not including the "tag" key anymore). We don't rely on this for functionality, and since we retrieve the manifests by specifying the tag, this test isn't needed. I added a test to look at the manifest version, so as soon as (if/when) they change it on us, the test will fail and we will know, and can read about the right way to update the manifest (and hopefully avoid the situation next time)

@singularityware-admin


gmkurtzer:
Wahooooooey!

Issue #384, opened 2016-12-02
Lower privileges when in "shim" mode


bbockelm:
After a `singularity_fork`, the parent lives on as sort of a "shim" process.   It simply sits around and does signal handling.  See:

https://github.com/singularityware/singularity/blob/master/src/lib/fork.c#L90

Two potential problems (not security issues alone, but might cause a small unrelated issue to escalate):
1.  The `sexec` process is still real UID 0 but effective UID of the invoking user.  If the invoking user came up with a clever (but currently unknown) mechanism to attack the parent `sexec`, they could escalate to root.
2.  There's a _lot_ of code in `sexec` that we don't want executed in this mode.  There's no obvious way to reach it - unless someone corrupts the `sexec` memory.

To lower the risk profile, I would propose we do two things:
* Have the parent `sexec` immediately drop to real UID non-zero.  Use a pipe for communication to prevent the child from proceeding until the parent has finished dropping permissions.  This way, we are guaranteed that end-user code is never executed until _after_ the privileges are completely dropped.
* (Even better) have the parent `exec` to a separate executable that does nothing but handle signals as an unprivileged user.  Once the exec is successful, then the child can proceed.

bbockelm:
Not sure if it's feasible to have the parent `exec` to a separate executable.  In the session dir case, we need to do some cleanup action still.  Probably only applies to the PID namespace case.  I think dropping privileges permanently (and setting `NO_NEW_PRIVS`) is as good as we'll get.

gmkurtzer:
We technically don't need any privileges for the cleanup so (at least in theory) we could exec() the child processes as well.

bbockelm:
Correct.  I just liked the idea better when it was "one generic shim executable" instead of "one shim executable per use case".

I think dropping privileges permanently probably is "good enough" for now.

gmkurtzer:
At present, the cleanup thread is a separate unprivileged binary. Still need to do the same for the `init` process for the PID namespace.

GodloveD:
@bbockelm, @gmkurtzer, and @bauerm97 does the new `instance` support address these issues?  If so, can we close?

gmkurtzer:
Oh, good catch @GodloveD, I think it does! I'm gonna close, but if @bbockelm feels we need more, we can reopen. Thanks!

PR #394, opened 2016-12-07
Drop privileges in parent threads

bbockelm:
Partial fix for #384

This forces the child process (post-fork) to delay until the parent has confirmed it has permanently dropped all privileges.

When testing, I additionally noticed that the exit status is incorrect when the payload is terminated by a signal; that is fixed here.

@singularityware-admin

### DIFF
--- a/src/lib/ns/user/user.c
+++ b/src/lib/ns/user/user.c
@@ -39,11 +39,41 @@
 #include "lib/config_parser.h"
 #include "lib/privilege.h"
 
-static int enabled = -1;
+int singularity_ns_user_configured(void) {
+    singularity_message(DEBUG, "Checking if user namespaces are configured.\n");
 
+    if ( ( is_suid("/proc/self/exe") == 0 ) && ( is_owner("/proc/self/exe", 0)  == 0) ) {
+        singularity_message(VERBOSE, "Not virtualizing user namespace: running SUID root\n");
+        return -1;
+    }
+
+    if ( getuid() == 0 ) {
+        singularity_message(VERBOSE3, "Not virtualizing USER namespace: running as root\n");
+        return -1;
+    }
+
+    if ( singularity_config_get_bool(ALLOW_USER_NS) <= 0 ) {
+        singularity_message(VERBOSE2, "Not virtualizing USER namespace by configuration\n");
+        return -1;
+    }
+
+#ifdef NS_CLONE_NEWUSER
+    singularity_message(DEBUG, "User namespace will be attempted.\n");
+    return 0;
+#else
+    singularity_message(DEBUG, "User namespace support disabled at compile time; support not available on host\n");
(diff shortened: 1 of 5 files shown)


bbockelm:
Test failure is due to userns being broken post-refactor.  Will probably try to do something minimal now and open up a ticket on the subject.

bbockelm:
@gmkurtzer - I think this one is ready to go.  I created a few more notes on what I want to do with the user namespaces.  There's cleanup to do there that can happen independently of this PR.

gmkurtzer:
I am going over this in more detail now, but at first glance, this looks good!

src/lib/ns/user/user.c
```diff
@@ -39,11 +39,32 @@
 #include "lib/config_parser.h"
 #include "lib/privilege.h"
 
-static int enabled = -1;
-
 int singularity_ns_user_enabled(void) {
-    singularity_message(DEBUG, "Checking user namespace enabled: %d\n", enabled);
-    return(enabled);
+    singularity_message(DEBUG, "Checking if user namespace is enabled.\n");
```
gmkurtzer: This does not take into consideration what I was initially going for with this function. The original idea was not to return if user namespace is possible, but to return if it is currently running. There are functions in the source that are expecting the previous behavior of this function (e.g. rootfs/dir.c and fork.c possibly others).

There is also some duplicate funcationatly presently in singularity_priv_userns_ready() and singularity_priv_userns_enabled(). I see pros and cons to using singularity_priv_userns_* as well as singularity_ns_user_*. If you are suggesting we use singularity_priv_userns_*, then we will need to go back and make sure we have changed all of the other calls to singularity_ns_user_enabled() to be either expecting this new return model, or switch them to use singularity_priv_userns_enabled().

bauerm97: Maybe a solution would be to have two user namespace functions `singularity_ns_user_enabled()` and `singularity_ns_user_allowed()`? Where one checks if you're *allowed* to run with namespaces, the other checks if you are *currently* running in a namespace?

bbockelm: Yeah - we need something like `singularity_ns_user_plan_on_invoking_at_some_point`.

As I said, I think there is some cleanup to doing in the userns code - I was just trying to avoid blocking this ticket on the cleanup.

bbockelm:
@gmkurtzer - a few fixes this morning:
* Separated unrelated commits into separate PRs.
* Kept original behavior of the `_enabled` function.
* Fixed status checks of the return value of `_enabled` function (only one of three were correct!  Maybe indicates the test coverage isn't great here?).
* Refactored out a `_configured` function to indicate whether this feature is configured.

gmkurtzer:
Ahh, looks much better to me now. Thanks @bbockelm!

Issue #391, opened 2016-12-06
change argument parsing to optparse


vsoch:
see https://github.com/singularityware/singularity/issues/389#issuecomment-265058452

vsoch:
This was fixed / done today.

PR #395, opened 2016-12-07
updating tests and cli.py to use optparse

### LINKED ISSUE #389: implement simple version of argparse
argparse is supported on python 2 (default library) but not RHEL 6. We need to implement a basic argparse - like strategy that does not have argparse as a dependency, but probably (hopefully) more than a strategy that assumes arguments are provided in some specific order (e.g., sys.argv[0]). Since this is a client used only by the bootstrap, it should be pretty reasonable to not have the degree of argument parsing that one would want for a true client.

vsoch:
Fixes #389 #391 

 - changes argparse to optparse, since it is supported in older version of RHEL (6?) The package is deprecated but still valid for Python 3, so we will likely be able to transition back to argparse and (hopefully) not need to support RHEL 6 anymore
 
 I still need to find and change the code for the python executable for the config parser - if it hasn't been merged yet it will need to wait for another pr!

@singularityware-admin

### DIFF
--- a/libexec/python/cli.py
+++ b/libexec/python/cli.py
@@ -46,7 +46,7 @@ def get_parser():
 
 
     parser = optparse.OptionParser(description='bootstrap Docker images for Singularity containers',
-                                   usage="usage: %prog [options] filename",
+                                   usage="usage: %prog [options]",
                                    version="%prog 2.2")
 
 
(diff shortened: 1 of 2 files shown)


vsoch:
@gmkurtzer ready for merge - this is a pretty simple PR, and I can update @bbockelm python script either before/after merge (I didn't realize that PR wasn't merged yet!)

gmkurtzer:
Perfect, thank you!

PR #400, opened 2016-12-08
Update Administration.md

serious-steve:
Fixes #

 -Adding a section in the Admin docs on how to build a debian package from source
 -Describing work-around for failed tests when they do not find an installed singularity binary

@singularityware-admin

Adding a section on how to build a debian package from source
Describing work-around for failed tests when they do not find an installed singularity binary

### DIFF
--- a/docs/Administration.md
+++ b/docs/Administration.md
@@ -68,6 +68,23 @@ $ PREFIX=/opt/singularity
 $ rpmbuild -ta --define="_prefix $PREFIX" --define "_sysconfdir $PREFIX/etc" --define "_defaultdocdir $PREFIX/share" singularity-*.tar.gz
 ```
 
+### Building a DEB directly from source
+
+To build a deb package for Debian/Ubuntu/LinuxMint invoke the following commands:
+
+```bash
+$ fakeroot dpkg-buildpackage -b -us -uc # sudo will ask for a password to run the tests
+$ sudo dpkg -i ../singularity-container_2.2-1_amd64.deb
+```
+ 
+Note that the tests will fail if singularity is not already installed on your system. This is the case when you run this procedure for the first time.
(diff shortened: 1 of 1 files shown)


gmkurtzer:
Hi @c1t4r! Thank you for these updates. We will also need to move this to the website documentation (and actually I think it would be best for me to remove these docs from the source code so that we don't have two places to maintain it).

Can you also look into what it takes to update the website docs? @vsoch and I posted a contributing docs section to the site so hopefully it isn't too hard.

Thank you!

vsoch:
Hi @c1t4r! Huge thanks for adding this! [Here is the link](http://singularity.lbl.gov/contributing-docs#setting-up-your-development-environment) for how to contribute to the web docs. Do you have the bandwidth to give it a go? I can be of help if you have any questions.

@gmkurtzer +1 - it will get confusing pretty quickly (if it isn't already) to have two versions of the same docs. We should migrate @c1t4r 's work to the web docs, and then remove these from the repo proper.

serious-steve:
Dear @gmkurtzer  @vsoch  IMO it is fine to have detailed docs on the web page and basic setup/install instructions with the sources. 
Should I also open an issue about the failing tests in `build-dpkg` ?

We all agree that the docs need some love.
Do we even know which portions are still working as intended originally? 
Why cant we have 
1) Short, stable documentation which will not get outdated (Web) No specific tutorials it is very hard to pick out the info one is seeking.
2) A "Fixes&Workaround" section which can change frequently and is version dependent. Ideally users could comment on that.
3) Some basic Install/Setup Docs with the sources, maybe even copied-on-fly from 1)

@vsoch  I'd love to discuss with you more how docs can be made more user friendly and how users can contribute. I've managed to set up a singularity container running jekyll very easily so I'll give it a try.

PR #405, opened 2016-12-09
Stopped test.sh from failing when test user accounts / groups already exist

bauerm97:
Fixes #

 - Ensure test user accounts / groups don't already exist in test.sh

### DIFF
--- a/test.sh.in
+++ b/test.sh.in
@@ -78,9 +78,21 @@ sudo true
 
 /bin/echo
 /bin/echo "Adding users and groups for a test."
-stest 0 sudo groupadd test_singularity2
-stest 0 sudo useradd singularity
-stest 0 sudo useradd test_singularity -G test_singularity2 -d /tmp
+if [ $(getent group test_singularity2) ]; then
+  /bin/echo "- Not adding test_singularity2 group; already exists."
+else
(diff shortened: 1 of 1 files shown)


bbockelm:
Hi - I actually have a separate fix for this here:

https://github.com/singularityware/singularity/pull/394/commits/e26b3058c8a644984235beb90ecdce6240fc154f

Can we cherry-pick that instead?  I'd prefer to avoid removing system users if possible.

bauerm97:
@bbockelm Yea, that works too. I just hadn't seen that your other PR is fixing this. I think it would also work if we just removed the stest from the lines (so it doesn't stop when it hits this error). Or, tell stest that 0/9 are acceptable return values (is this even possible?). Either way, I think it's reasonable to do this in a separate PR and push this through quickly since it's a bit annoying currently.

EDIT: Also if you can hop on slack I have a question for you about something @vsoch and I have been discussing recently

gmkurtzer:
Which one of these should I merge (#405 or #406)?

bauerm97:
@gmkurtzer Your call, they're identical in every way. Maybe just go with @bbockelm s branch? I'll go ahead and close this PR

gmkurtzer:
Thanks!

PR #406, opened 2016-12-09
Make tests idempotent.

bbockelm:
@singularityware-admin

Since the tests need to add users to the system, they currently
fail if run twice in the same VM.  Make it so we don't attempt the
useradd command if it was already done in a previous run.


bbockelm:
@bauerm97 - replaces #405

bauerm97:
@gmkurtzer Let's merge this one

gmkurtzer:
Awesome, thanks guys!

PR #407, opened 2016-12-09
Fix parent exit status when container is terminated by a signal.

bbockelm:
@singularityware-admin

Currently, if a process terminates due to a signal (`SIGSEGV`, for example), then `singularity` will return exit status 0.

NOTE that the tests fail on this one.  @bauerm97 - I think this is due to an issue with the bootstrap refactor segfaulting that was previously hidden because the test improperly succeeded.

### DIFF
--- a/src/lib/sessiondir.c
+++ b/src/lib/sessiondir.c
@@ -111,7 +111,13 @@ char *singularity_sessiondir_init(char *file) {
             singularity_message(DEBUG, "Cleanup thread waiting on child...\n");
 
             waitpid(child_pid, &tmpstatus, 0);
-            retval = WEXITSTATUS(tmpstatus);
+            if (WIFEXITED(tmpstatus)) {
+                retval = WEXITSTATUS(tmpstatus);
+                singularity_message(DEBUG, "Child exited with status %d.\n", retval);
+            } else {
+                retval = WTERMSIG(tmpstatus);
+                singularity_message(DEBUG, "Child exited with signal %d.\n", retval);
+            }
 
             singularity_message(DEBUG, "Checking to see if we are the last process running in this sessiondir\n");
             if ( flock(sessiondir_fd, LOCK_EX | LOCK_NB) == 0 ) {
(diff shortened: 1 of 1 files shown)


src/lib/sessiondir.c
```diff
@@ -111,7 +111,13 @@ char *singularity_sessiondir_init(char *file) {
             singularity_message(DEBUG, "Cleanup thread waiting on child...\n");
 
             waitpid(child_pid, &tmpstatus, 0);
-            retval = WEXITSTATUS(tmpstatus);
```
vsoch, line 114: what does WEXITSTATUS mean? I only found [this](http://stackoverflow.com/questions/20465039/what-does-wexitstatusstatus-return) and it didn't tell me much.

gmkurtzer:
@vsoch From the waitpid(2) man page:

```
       WEXITSTATUS(status)
              returns the exit status of the child.  This consists of the least significant 8 bits of the status argument that the child specified in a call to exit(3) or _exit(2) or as the argument for  a  return  statement  in
              main().  This macro should be employed only if WIFEXITED returned true.
```

The rest looks good to me, but not sure why we are getting a failure from the Python unit tests.. Thoughts?

vsoch:
oh wow, I didn't know stuff like that existed! That is very cool :)

gmkurtzer:
Oh, I missed the error in the unit tests... 

```
Building test container...
 + sudo singularity create -s 568 container.img                                        (retval=0) OK
 + sudo singularity bootstrap container.img /home/travis/build/singularityware/sing  (retval=139) ERROR
```

@bbockelm can you take a look at that? If not, I'll try to get to it later this afternoon.

bbockelm:
The unit test failure is due to the bootstrap refactor.  This PR fixed the fact that segfaults result in exit code 0.

bauerm97:
@bbockelm This is a good catch, thanks for bringing it to my attention. I'll check out these changes in a local branch and see what the issue is.

EDIT: Fixed the error, it was being caused by passing a NULL pointer into strcmp. I'll update my code janitor branch, which we can merge in before merging this in and it should be fixed.

PR #410, opened 2016-12-12
Fix logic for dropping permissions in parent thread caused by ns/user/user.c

### LINKED ISSUE #409: singularity_user_ns_enabled() call breaks functionality
Commit ed07e7ced8a1b5e25180715388e702e2586c1beb breaks functionality, reverting it to the previously used `singularity_ns_user_configured()` fixes this, however I'm not sure about the logic behind what we're doing here so I can't comment on if that is an acceptable fix. I'm looking into the best way to fix it, but for now it might be best to revert back to the original form, so that PR tests will pass again.

bauerm97:
Fixes #409 

 - This reverts commit ed07e7ced8a1b5e25180715388e702e2586c1beb.

### DIFF
--- a/src/lib/fork.c
+++ b/src/lib/fork.c
@@ -248,10 +248,12 @@ pid_t singularity_fork(void) {
 
         // At this point, we have nothing to do but wait on some external action.
         // We should never again need to increase our privileges.  Drop privs
-        // permanently and then indicate the child can proceed.
-        if (singularity_ns_user_enabled() < 0) {
+        // permanently and then indicate the child can proceed. 
+        if ( singularity_priv_is_suid() == 0 ) {
+            singularity_message(DEBUG, "Reached drop_perm in fork.c\n");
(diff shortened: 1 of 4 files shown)


gmkurtzer:
Why are you reverting ed07e7c?

gmkurtzer:
Ahh, reading backlog now....

gmkurtzer:
Please see my comment on #409 and include that in this pull request.

Thank you!

bauerm97:
@gmkurtzer I updated the logic around this entire part of the code: 

 - Cleaned up a bit in `lib/ns/user/user.c`
 - Added `singularity_priv_is_suid()` to `lib/privilege.c`
 - Fixed `lib/fork.c` to drop privileges only when it has privileges to drop

gmkurtzer:
This has been merged and I went ahead and cleaned the logic in user.c.

PR #416, opened 2016-12-15
Bootdef parsing case insensitivity

bauerm97:
Fixes #

 - Add str_casecmp which is a case insensitive version of strcmp
 - Updated bootdef parsing to be case insensitive for keywords

@singularityware-admin

### DIFF
--- a/src/lib/image/bootstrap/bootdef_parser.c
+++ b/src/lib/image/bootstrap/bootdef_parser.c
@@ -25,6 +25,7 @@
 #include <sys/types.h>
 #include <errno.h>
 #include <string.h>
+#include <strings.h>
 #include <fcntl.h>
 #include <limits.h>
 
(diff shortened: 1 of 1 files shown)


gmkurtzer:
Why not use strcasecmp()?

bauerm97:
@gmkurtzer All the references on strcasecmp() I could find said that it wasn't necessarily in the standard library, so I wasn't sure if it was safe to use this. See: https://linux.die.net/man/3/strcasecmp

gmkurtzer:
It seems to be defined in strings.h on the system that I am looking.

My concern with using tolower() is that it won't work on multi-byte chars. Considering we control the names of the strings we are comparing in this particular case, this is probably a big deal, but that was why I brought up strcasecmp(). If you have reservations about it, we can use the current implementation, let's just make a note in the header that it will only work on single byte characters.

Greg

bauerm97:
@gmkurtzer other thing we could do is only define the tolower and str_casecmp if strcasecmp is not defined? That might potentially work, but I'm not sure it's worth the hassle to us. I think you're correct that it should be OK.

gmkurtzer:
Yes, we could write a wrapper and some `#ifdef`s, but being that it is in strings.h and available going back as far as RHEL5, I think it is safe.

bauerm97:
@gmkurtzer This is ready to be merged, it's a very simple change. Please push this through when you get a chance.

vsoch:
thank goodness! I've had so many builds fail because of `bootstrap` vs `Bootstrap` and then finally (the working)  `BootStrap` !

Issue #411, opened 2016-12-12
$LD_LIBRARY_PATH is unset within the container (with setuid)


alvarosimon:
This is quite simple to reproduce, singularity should be able to inherit the OS envs but `$LD_LIBRARY_PATH` is always unset within the container. 

As example  from the host:
> $ echo $LD_LIBRARY_PATH
> /apps/....

but from the container the variable is always unset (probably because singularity is executed with setuid flag and is unset for security reasons by the OS)
>$ singularity shell $VSC_DATA_VO_USER/debian_ib_drivers.img
> $ echo $LD_LIBRARY_PATH
>

This is documented in `ld.so`

The image was generated from scratch using a debian bootstrap definition file, and `/etc/singularity/init` host file looks like this:
  
> unset module
> unset BASH_ENV
> PATH="$PATH:/bin:/sbin:/usr/bin:/usr/sbin"
> HISTFILE="/dev/null"
> if [ -z "${TZ:-}" ]; then
>     TZ=`date +%Z`
> fi
> export PATH HISTFILE TZ

Any workaround to change this behaviour in singularity 2.2 to keep ` $LD_LIBRARY_PATH`? maybe we should create another variable as backup and recover the original variable within the container?

alvarosimon:
I have tried to do something like this:

```
$ diff -Naur /etc/singularity/init.orig /etc/singularity/init
--- /etc/singularity/init.orig	2016-12-12 16:21:34.667802836 +0100
+++ /etc/singularity/init	2016-12-12 16:24:42.285398331 +0100
@@ -17,5 +17,12 @@
     TZ=`date +%Z`
 fi
 
-export PATH HISTFILE TZ
+# Set LD_LIBRARY_PATH
+if [ -n "$LD_LIBRARY_PATH" ]; then
+
+    LD_LIBRARY_PATH_COPY="$LD_LIBRARY_PATH"
+
+fi
+
+export PATH HISTFILE TZ LD_LIBRARY_PATH_COPY
```
to keep a copy, but then you have to generate again LD_LIBRARY_PATH in your container, I don't know if there is a better workaround to bypass this...

gmkurtzer:
Hi @alvarosimon, you are right in pointing at SUID but something else to consider is that there is no guarantee that the LD_LIBRARY_PATH which works on the host will also be applicable via the container... As a matter of fact, it seems to me that this by definition may end up with ABI incompatibilities and severely stunt container portability.

[...]

PR #417, opened 2016-12-16
New container environment file format

bauerm97:
As discussed in #364, #376
Enables proposed solution to #411 

 - Introduces `/.env/` directory in container
   - `.metasource` file, contains a list of all files in `/.env/` directory
   - An arbitrary number of files following the naming convention: `XX-$NAME` where `XX` is a two digit number. Higher numbers take precedence over lower numbers.
 - Default `exec`, `shell`, and `run` entrypoints will source `/.env/.metasource`
 - In `src/lib/action/action.c`, `singularity_action_do()` will fork before executing the entrypoint. After the child process (the container) exits, the parent process will attempt to overwrite the `/.env/.metasource` file. This ensures that if any files were added or removed will also be added or removed from the `/.env/.metasource` file.
 - `src/lib/rootfs/rootfs.c`
     - Stores the filedescriptor of the rootfs directory (opened with O_CLOEXEC)
     - Will now fchdir to the open file descriptor, and then chroot on the current directory
 - In `src/util/file.c`, fileputat() function will write to a file when when given an fd of a directory
 - `sexec.c` will return the value of `singularity_action_do()` to handle the new forking
 - `src/lib/file/environment/environment.c` will handle sorting & creation of the `/.env/.metasource` file

@singularityware-admin

I would like to further convert more of our functions to using filedescriptors, as they are more secure than running on path name alone. I've laid out the groundwork for this in `rootfs.c` by storing the fd and using it for the `singularity_rootfs_chroot()` function. We can further improve our usage of file descriptors by calling fileputat instead of fileput, anytime we would like to write to a file inside the container. Further, it might be a good idea to store the fd of the working directory when singularity is first called, to allow files such as `src/lib/mount/cwd/cwd.c` to access the location from an fd, rather than path or `get_current_directory_name()` (the man page of which recommends instead using fd's).

### DIFF
--- a/src/lib/action/action.c
+++ b/src/lib/action/action.c
@@ -109,8 +109,6 @@ int singularity_action_do(int argc, char **argv) {
     int tmpstatus;
     int retval = 0;
 
-    singularity_priv_drop_perm();
-
     singularity_message(DEBUG, "Trying to change directory to where we started\n");
     char *target_pwd = envar_path("SINGULARITY_TARGET_PWD");
     if (!target_pwd || (chdir(target_pwd) < 0)) {
@@ -133,6 +131,7 @@ int singularity_action_do(int argc, char **argv) {
     child = singularity_fork();
 
     if ( child == 0 ) {
+        singularity_priv_drop_perm();
         if ( action == ACTION_SHELL ) {
             singularity_message(DEBUG, "Running action: shell\n");
             action_shell_do(argc, argv);
--- a/src/lib/action/action.c
+++ b/src/lib/action/action.c
@@ -35,6 +35,7 @@
 #include "util/util.h"
 #include "lib/message.h"
 #include "lib/privilege.h"
+#include "lib/singularity.h"
 #include "exec/exec.h"
 #include "shell/shell.h"
 #include "run/run.h"
@@ -160,4 +161,5 @@ int singularity_action_do(int argc, char **argv) {
         retval = WEXITSTATUS(tmpstatus);
         singularity_file_environment();
     }
+    return(retval);
 }
--- a/src/lib/action/action.c
+++ b/src/lib/action/action.c
@@ -27,6 +27,9 @@
 #include <linux/limits.h>
 #include <unistd.h>
 #include <stdlib.h>
+#include <sys/wait.h>
+#include <poll.h>
+
 
 #include "util/file.h"
 #include "util/util.h"
@@ -101,6 +104,9 @@ int singularity_action_init(void) {
 }
 
 int singularity_action_do(int argc, char **argv) {
+    pid_t child;
+    int tmpstatus;
+    int retval = 0;
 
     singularity_priv_drop_perm();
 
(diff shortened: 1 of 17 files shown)


src/lib/action/action.c
```diff
@@ -123,25 +128,37 @@ int singularity_action_do(int argc, char **argv) {
...
-    } else if ( action == ACTION_TEST ) {
-        singularity_message(DEBUG, "Running action: test\n");
-        action_test_do(argc, argv);
-    } else if ( action == ACTION_START ) {
-        singularity_message(DEBUG, "Running action: start\n");
-        action_start_do(argc, argv);
-    } else if ( action == ACTION_STOP ) {
-        singularity_message(DEBUG, "Running action: stop\n");
-        action_stop_do(argc, argv);
+    child = singularity_fork();
```
gmkurtzer, line 131: Why are we forking here?

gmkurtzer: @bauerm97 and I spoke about this on Slack. This is for the post process rebuild of the environment metadata. I asked Michael to refactor this so it 1. doesn't require a fork, 2. doesn't update the metadata after every command (even if --writable was set). I would prefer either non-state and glob at runtime or only do this after bootstrap/import.

src/lib/rootfs/rootfs.c
```diff
@@ -245,18 +254,22 @@ int singularity_rootfs_check(void) {
 int singularity_rootfs_chroot(void) {
     
     singularity_priv_escalate();
-    singularity_message(VERBOSE, "Entering container file system root: %s\n", joinpath(mount_point, OVERLAY_FINAL));
-    if ( chroot(joinpath(mount_point, OVERLAY_FINAL)) < 0 ) { // Flawfinder: ignore (yep, yep, yep... we know!)
-        singularity_message(ERROR, "failed enter container at: %s\n", joinpath(mount_point, OVERLAY_FINAL));
+
+    if ( fchdir(singularity_rootfs_fd()) < 0 ) {
```
gmkurtzer, line 258: What are you fixing here? Typically to pass lint testing, you will need to chdir() after doing a chroot().

src/util/file.c
```diff
@@ -358,6 +358,30 @@ int fileput(char *path, char *string) {
     return(0);
 }
 
+int fileputat(int fd, char *path, char *string) {
```
gmkurtzer, line 361: I am fine with this, but just curious... What is it gaining us over using fileput(joinpath(dir1, file), ...)?

bauerm97: Well, I wanted to be able to run `singularity_file_environment()` at any point of the code. At some points in the code we are chdir'd to different locations, or perhaps chroot'd  to another location. This basically ensures that we're writing into the container no matter where we call this in the program, since the rootfs_fd will remain opened.

gmkurtzer: Ahh, good thinking on that. Makes sense, thanks!

src/lib/rootfs/rootfs.c
bauerm97: I am not technically fixing anything, however it is theoretically more secure to first fchdir to the rootfs fd, and then chroot on the current directory. We don't have to worry about issues with path name strings, changing links, etc.. because of doing it this way I believe.

And if we need to chdir after the chroot, then we can just add one last call to chdir("/") after the chroot

gmkurtzer: Well, we aren't using chroot() to be secure as we will end up being the same user inside as outside, but we do need to pass lint'ing which I know generally checks for the chdir() soon after the chroot().

I don't see this step as actually gaining us anything.

src/util/file.c
bauerm97: And we can further extend this usage of file descriptors to other places in the program, letting us remove our dependence on path name strings, and switch what we can to file descriptors. This also has the added benefit of removing pathways for attempting to attack singularity and escalate

gmkurtzer: +1

src/lib/rootfs/rootfs.c
bauerm97: We can be more secure in the sense that we know that we are attempting to chroot into the directory we want to be.

src/util/file.c
vsoch: @gmkurtzer what is the status here? This is an important addition!

bauerm97: @vsoch @gmkurtzer I offered to tackle this issue, however I'm just back at uni so I wasn't working the last few weeks. I will handle the continuation of shifting to file descriptors, however I think we can merge the new environment format before that is all finished.

gmkurtzer: I would suggest we integrate this into 2.3 and use the new key/value registry to store this. So fileputat() will always just work.... Hrmm... Maybe the registry should store `void *` so it can take any sort of value (e.g. an int). I'll look into that too.

bbockelm:
As we start to beef up the metadata environment of Singularity container images (for example, maybe we should start recording when the image was created?  Version of Singularity?), maybe we should move this sort of work into a subdirectory of `/.singularity`?

vsoch:
yes! See my [thinking here](https://github.com/singularityware/singularity/issues/387#issuecomment-277482562). We might want basic stuff to live with the container, and everything else for Singularity Hub to help manage.

gmkurtzer:
This RFE has been implemented in the "development" branch. Thanks!

PR #420, opened 2016-12-19
Support for overriding variables inside the container at runtime

### LINKED ISSUE #411: $LD_LIBRARY_PATH is unset within the container (with setuid)

jtriley:
Fixes #411

This allows users to explicitly override environment variables within the container via `SENV_` prefixed environment variables. This prefix gets stripped and the result gets set within the containers's environment. This is especially useful for proxying [unsecure variables stripped by ld for suid/sgid executables ](https://sourceware.org/git/?p=glibc.git;a=blob_plain;f=sysdeps/generic/unsecvars.h;hb=HEAD)  to the container:
```
$ whoami
test
$ export LD_LIBRARY_PATH=/test/path
$ singularity exec cuda75.img env | grep LD_LIBRARY_PATH
$ export SENV_LD_LIBRARY_PATH=/test/path
$ singularity exec cuda75.img env | grep LD_LIBRARY_PATH
LD_LIBRARY_PATH=/test/path
```

 - adds a `singularity_env_override` method to `src/lib/action/action.c` called by `singularity_action_init` that looks for env variables that start with prefix, strips the prefix, and sets the resulting key/value inside the container.

@singularityware-admin

### DIFF
--- a/src/lib/action/action.c
+++ b/src/lib/action/action.c
@@ -49,19 +49,62 @@
 static int action = 0;
 static char *cwd_path;
 
+static char *LD_UNSECVARS[] = {
+    "GCONV_PATH",
+    "GETCONF_DIR",
+    "HOSTALIASES",
+    "LD_AUDIT",
+    "LD_DEBUG",
+    "LD_DEBUG_OUTPUT",
+    "LD_DYNAMIC_WEAK",
+    "LD_LIBRARY_PATH",
+    "LD_ORIGIN_PATH",
+    "LD_PRELOAD",
+    "LD_PROFILE",
+    "LD_SHOW_AUXV",
+    "LD_USE_LOAD_BIAS",
+    "LOCALDOMAIN",
+    "LOCPATH",
+    "MALLOC_TRACE",
+    "NIS_PATH",
+    "NLSPATH",
+    "RESOLV_HOST_CONF",
+    "RES_OPTIONS",
+    "TMPDIR",
+    "TZDIR",
+};
+
+static int LD_UNSECVARS_LEN = 22;
+static int LD_UNSECVARS_MAX_STRLEN = 16;
+static int LD_UNSECVARS_PREFIX_LEN = 10;
+static char * LD_UNSECVARS_PREFIX = "CONTAINER_%s";
+
+int singularity_proxy_unsecvars() {
+    int prefixedvar_max_len = LD_UNSECVARS_MAX_STRLEN + LD_UNSECVARS_PREFIX_LEN + 1;
+    char prefixedvar[prefixedvar_max_len];
+    char *varname;
+    char *value;
+    int c;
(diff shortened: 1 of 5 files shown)


vsoch:
Would it be possible to write tests for this for continuous integration?

jtriley:
Yup! Working on that now

vsoch:
Excellent.

jtriley:
OK looks like new tests passed

moskalenko:
+1 for this PR.

jtriley:
New plan: spoke with @gmkurtzer on slack about updating this to accept any variable with `PREFIX` which would allow us to drop the `unsecvars.h` list and let users set any environment variable in the container they see fit.

gmkurtzer:
Confirming @jtriley's comment about updating to accept any variable name with a particular `PREFIX`. This `PREFIX` could be dynamically set, but that will make documentation difficult, so I propose we define it.

I would like to use `SENV_` as a prefix unless anyone else has either a better suggestion or a reason why that particular prefix is a bad idea. Just to articulate, this would look like:

```bash
$ SENV_LD_LIBRARY_PATH=/path1 singularity shell container.img
Singularity.container.img> echo $LD_LIBRARY_PATH
/path1
```

jtriley:
@gmkurtzer First implementation complete. Let me know what you think.

jtriley:
I think I did a little too much here based on your earlier comment - let me know if you'd prefer to remove the `SINGULARITY_ENV_PREFIX` and change the default prefix from `_SENV` to `SENV`.

jtriley:
@gmkurtzer I removed the extra cruft (`SINGULARITY_ENV_PREFIX` and `_SENV` as the default prefix) and reworked the title and description to more accurately reflect what's going on here.

jtriley:
Grrr just realized this PR is pointing to master. Will rename branch and recreate the PR.

PR #424, opened 2016-12-21
Support for overriding variables inside the container at runtime 

### LINKED ISSUE #411: $LD_LIBRARY_PATH is unset within the container (with setuid)

jtriley:
Fixes #411

This allows users to explicitly override environment variables within the container via `SENV_` prefixed environment variables. This prefix gets stripped and the result gets set within the containers's environment. This is especially useful for proxying [unsecure variables stripped by ld for suid/sgid executables ](https://sourceware.org/git/?p=glibc.git;a=blob_plain;f=sysdeps/generic/unsecvars.h;hb=HEAD)  to the container:
```
$ whoami
test
$ export LD_LIBRARY_PATH=/test/path
$ singularity exec cuda75.img env | grep LD_LIBRARY_PATH
$ export SENV_LD_LIBRARY_PATH=/test/path
$ singularity exec cuda75.img env | grep LD_LIBRARY_PATH
LD_LIBRARY_PATH=/test/path
```

 - adds a `singularity_env_override` method to `src/lib/action/action.c` called by `singularity_action_init` that looks for env variables that start with prefix, strips the prefix, and sets the resulting key/value inside the container.

@singularityware-admin

### DIFF
--- a/src/lib/action/action.c
+++ b/src/lib/action/action.c
@@ -49,19 +49,62 @@
 static int action = 0;
 static char *cwd_path;
 
+static char *LD_UNSECVARS[] = {
+    "GCONV_PATH",
+    "GETCONF_DIR",
+    "HOSTALIASES",
+    "LD_AUDIT",
+    "LD_DEBUG",
+    "LD_DEBUG_OUTPUT",
+    "LD_DYNAMIC_WEAK",
+    "LD_LIBRARY_PATH",
+    "LD_ORIGIN_PATH",
+    "LD_PRELOAD",
+    "LD_PROFILE",
+    "LD_SHOW_AUXV",
+    "LD_USE_LOAD_BIAS",
+    "LOCALDOMAIN",
+    "LOCPATH",
+    "MALLOC_TRACE",
+    "NIS_PATH",
+    "NLSPATH",
+    "RESOLV_HOST_CONF",
+    "RES_OPTIONS",
+    "TMPDIR",
+    "TZDIR",
+};
+
+static int LD_UNSECVARS_LEN = 22;
+static int LD_UNSECVARS_MAX_STRLEN = 16;
+static int LD_UNSECVARS_PREFIX_LEN = 10;
+static char * LD_UNSECVARS_PREFIX = "CONTAINER_%s";
+
+int singularity_proxy_unsecvars() {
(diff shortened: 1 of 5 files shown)


gmkurtzer:
Hi @jtriley, just wanted to check to see if this is ready for a review.

Thanks!

jtriley:
@gmkurtzer yup this is ready, thanks!

jtriley:
@gmkurtzer I reworked this using strcmp and strtok_r as discussed. Ready for a review.

boegel:
Any progress on this? Incl. ETA for a released version with this included?

gmkurtzer:
@jtriley Can you take a look at the lib-refactor branch and rebase to that? It is gonna be a bit messy I know as things will have to move, but I would prefer you get these changes into that rather then have me respin this into that branch.

Check out https://github.com/singularityware/singularity/blob/lib-refactor/src/lib/runtime/environment/environment.c for a starting point, and you should see I used your code as a template.

Let me know if you have any questions, thanks!

jtriley:
@gmkurtzer will do

gmkurtzer:
This has been implemented in the development branch by using the environment variable syntax `SINGULARITYENV_` where everything post string will be transposed into the container. To unset a variable simply make it equal a zero length string.

PR #428, opened 2016-12-23
adding link to docs repo to encourage contribution

vsoch:
Currently, it's not clear that our docs are hosted on Github (and thus users can contribute), and the current README also presents them as a work in progress. While this is true, I think we would be ok to say "here are the docs! Please contribute!" Is this trivial change ok @gmkurtzer ?

@singularityware-admin


vsoch:
This is trivial for README, so I'm going to merge.

PR #437, opened 2017-01-03
Fix format string vulnerability in message handler.

cclerget:
Steps to reproduce:

export SINGULARITY_IMAGE=%n
singularity shell

result in:
*** %n in writable segment detected ***
Abandon (core dumped)

expected result:
ERROR  : Illegal input character '%' in: 'SINGULARITY_IMAGE=%n'
ABORT  : Retval = 255

glibc have some protection against format string vulnerabilites depending of compilation optimization options, so it should be safe without the fix but in case of old glibc or different libc ...

### DIFF
--- a/src/lib/message.c
+++ b/src/lib/message.c
@@ -106,7 +106,7 @@ void _singularity_message(int level, const char *function, const char *file_in, 
         char syslog_string[540]; // Flawfinder: ignore (512 max message length + 28'ish chars for header)
         snprintf(syslog_string, 540, "%s (U=%d,P=%d)> %s", __progname, geteuid(), getpid(), message); // Flawfinder: ignore
 
-        syslog(syslog_level, syslog_string, strlength(syslog_string, 1024)); // Flawfinder: ignore (format is internally defined)
+        syslog(syslog_level, "%s", syslog_string, strlength(syslog_string, 1024)); // Flawfinder: ignore (format is internally defined)
     }
 
     if ( level <= messagelevel ) {


bbockelm:
@vsoch - I think the unit test failures are a python issue.  Can you look into it?

vsoch:
yes, just comment out the shub (Singularity Hub) tests for now because I will replace the (now non existing) images with permanent ones for testing when shub is finished. But I think there are other (non python) related issues? See:

      /var/lib/singularity/mnt/final/bin/busybox: line 1: syntax error near unexpected token `newline'
      /var/lib/singularity/mnt/final/bin/busybox: line 1: `<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">'
      ERROR  : Container does not have a valid /bin/sh
      ABORT  : Retval = 255
      Full output in: /tmp/tmp.TbhX0AKVFm
      make: *** [test] Error 1
      The command "make test" exited with 2.
      0.01s$ jdk_switcher use oraclejdk8
      Switching to Oracle JDK8 (java-8-oracle), JAVA_HOME will be set to /usr/lib/jvm/java-8-oracle

bbockelm:
Actually, it looks like most of the of the `snprintf` usage is incorrect.  Working on a more comprehensive fix..

bbockelm:
Thanks a lot for the report @cclerget.  I am closing this PR as it is superseded by #438, which attempts a bit more bulletproofing around this code.

cclerget:
@bbockelm look at your diff, why rather than add condition on snprintf, not initialize buffer to zero's and just do snprintf(buffer, buffer_len-1 ...) to ensure buffer is always null terminated ?

bbockelm:
@cclerget - I actually did a bit of both.  I did an explicit null-termination on overflows that happen due to internal issues (such as overly-long function names in the logging header).

However, in cases where we're getting a long string (such as a path name) from the user or admin, I wanted to explicitly detect and handle the overflow case.  In some cases, I upgraded it to an abort; in another, I make a note in the log that it had been truncated.

PR #438, opened 2017-01-03
Correctly handle format strings and null-terminate.

bbockelm:
Make sure we don't utilize user-controlled input as a format string.

Verify we null-terminate fixed-length strings when `snprintf` hits the character limit.

@singularityware-admin

### DIFF
--- a/src/lib/message.c
+++ b/src/lib/message.c
@@ -118,23 +118,23 @@ void _singularity_message(int level, const char *function, const char *file_in, 
     }
 
     if ( level <= messagelevel ) {
-        char *header_string = NULL;
+        char header_string[95];
 
         if ( messagelevel >= DEBUG ) {
-            char *debug_string = (char *) malloc(25);
-            char *location_string = (char *) malloc(60);
-            char *tmp_header_string = (char *) malloc(80);
-            header_string = (char *) malloc(80);
+            char debug_string[25];
+            char location_string[60];
+            char tmp_header_string[86];
             snprintf(location_string, 60, "%s:%d:%s()", file, line, function); // Flawfinder: ignore
+            location_string[59] = '\0';
             snprintf(debug_string, 25, "[U=%d,P=%d]", geteuid(), getpid()); // Flawfinder: ignore
-            snprintf(tmp_header_string, 80, "%-18s %s", debug_string, location_string); // Flawfinder: ignore
-            snprintf(header_string, 80, "%-7s %-62s: ", prefix, tmp_header_string); // Flawfinder: ignore
-            free(debug_string);
-            free(location_string);
-            free(tmp_header_string);
+            debug_string[24] = '\0';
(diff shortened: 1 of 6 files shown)


bbockelm:
This is a more thorough handling of the problem reported by @cclerget in #437 .

@gmkurtzer - I think this is serious enough that it should be backported to the 2.2.x series.  Unsanitized `snprintf` format strings can be seriously dangerous.

bbockelm:
@vsoch - can you comment out the appropriate failing test and commit to master?  I'll rebase this PR once that's done.

vsoch:
python tests are passing, but I'm getting the same error as previously (that we see here):

https://travis-ci.org/singularityware/singularity/builds/188582489

vsoch:
I rebuilt a (previously passing) build, and it's failing with the same error:

https://travis-ci.org/singularityware/singularity/builds/187484431

so there must be some change on travis? Circle gives ssh access to the instances, that would sure be helpful now. It looks like it's failing with some newline error on [this line](https://github.com/singularityware/singularity/blob/master/test.sh.in#L139) and that the variable for some reason is an http page...

vsoch:
wait, now I'm really confused, the status is red but the error doesn't seem to be there?

gmkurtzer:
Uhmm... Why did this close when I merged #440? I'm assuming this needs to be reopened, so reopening it.

src/lib/message.c
```diff
@@ -103,30 +106,35 @@ void _singularity_message(int level, const char *function, const char *file_in,
...
+            syslog_string[559] = '\0';
+        }
 
-        syslog(syslog_level, syslog_string, strlength(syslog_string, 1024)); // Flawfinder: ignore (format is internally defined)
+        syslog(syslog_level, "%s", syslog_string);
     }
 
     if ( level <= messagelevel ) {
-        char *header_string = NULL;
+        char header_string[95];
```
gmkurtzer, line 121: Why did you prefer to statically allocate this instead of malloc()?

bbockelm: I'm sure static allocation is 1 microsecond faster, but the real reason was to prevent the possibility of future memory leaks in a refactoring.

gmkurtzer: Hahaha, well.. Good point. We need to keep performance in mind!

OK, while I don't like the motivation it is valid. Thanks!

gmkurtzer:
Looks good to me. Can you resync with master branch?

bbockelm:
Actually, you can do that from your side.  First you have to hit "approve" in the review tab.  Once you're ready to commit, you can drop-down the "merge" button and select "rebase and merge" to avoid the merge commit.

gmkurtzer:
Ahh, good to know. Let me give that a try. :)

gmkurtzer:
Thanks!

vsoch:
yeah it's pretty awesome :) Github also recently added an ability to search old commits!!

PR #443, opened 2017-01-04
Add definitions for CPU and GPU TensorFlow

eshelman:
These provide version 0.12.1 of TensorFlow installed within an Ubuntu
16.04 container. The GPU-accelerated version does require specific
versions of the NVIDIA driver and CUDA, but instructions on obtaining
them are provided.

The NVIDIA GPU support is adapted from existing Singularity definitions:
  * https://github.com/drorlab/tf-singularity
  * https://github.com/jdongca2003/Tensorflow-singularity-container-with-GPU-support

### DIFF
--- a/examples/contrib/ubuntu16-tensorflow-0.12.1-gpu.def
+++ b/examples/contrib/ubuntu16-tensorflow-0.12.1-gpu.def
@@ -1,0 +1,136 @@
+# Defines a Singularity container with TensorFlow pre-installed
+#
+
+#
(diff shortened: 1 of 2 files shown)


vsoch:
These would be great to have in the contrib folder, and (hopefully in the early year) provided on Singularity Hub for easier access. I don't have time to test today, but definitely +1

gmkurtzer:
Looks good, thanks!

PR #448, opened 2017-01-06
Convert config_parser to not use global hash instance.

bbockelm:
Should make config_parser better when embedded within a larger library.


gmkurtzer:
Thanks @bbockelm!

PR #453, opened 2017-01-14
changing second character to newline to fix parsing bug. Seems to work!

vsoch:
Not sure if this is reasonable, but after trying different stuff it seemed that the issue was using another `:` in `strtok`. I changed it to newline, and this seemed to fix the issue. What isn't clear to me is why it was chosen in the first place (is there some format of line like : : : that uses it?).

@singularityware-admin


gmkurtzer (approved):
Looks good!

PR #457, opened 2017-01-17
Changed TZ from date +%Z to use info from /etc/localtime.

tin6150:
The last line of /etc/localtime works better than +%Z,
This string is longer.  eg PST8PDT,M3.2.0,M11.1.0 instead of PST.  But the
latter doesn't actually have a matching timezone file and thus produce
wrong result.

I ran
TZ_LIST=$(  find /usr/share/zoneinfo -type f -exec tail -1 {} \; )
for TZ in $TZ_LIST; do
   echo $TZ
   date -d "TZ=\"$TZ\"  2016-02-29 13:45"
done
and confimed setting TZ like this produce desired result.

I ran sanity test on CentOS-6 and Ubuntu-14.
Both "echo $TZ" and "date" produced expected result for all 6 time zones of US, and Europe/Zurich.

Fixes #

 - Fix TZ so that date would produce desired result for timezone that doesn't have filenames matching the string from "date +%Z".   Issue was discussed in https://groups.google.com/a/lbl.gov/forum/#!topic/singularity/YWFspd3A_sg

-Tin

### DIFF
--- a/etc/init
+++ b/etc/init
@@ -12,10 +12,5 @@ PATH="$PATH:/bin:/sbin:/usr/bin:/usr/sbin"
 # Don't save the shell's HISTFILE
 HISTFILE="/dev/null"
 
-# Set the timezone if it is not already set
-if [ -z "${TZ:-}" ]; then
-    TZ=`strings /etc/localtime | tail -n 1`
-fi
-
-export PATH HISTFILE TZ
+export PATH HISTFILE
 
--- a/etc/init
+++ b/etc/init
@@ -14,7 +14,7 @@ HISTFILE="/dev/null"
 
 # Set the timezone if it is not already set
 if [ -z "${TZ:-}" ]; then
-    TZ=`date +%Z`
+    TZ=`strings /etc/localtime | tail -n 1`
 fi
 
 export PATH HISTFILE TZ


gmkurtzer (changes requested):
Why is the call to `strings` necessary, wouldn't `tail -n 1 /etc/localtime` work?

Thanks!

tin6150:
`tail -n 1 /etc/localtime` does work.  However the timezone file is a binary database file.  `cat` would produce funny non-print symbols.  Thus, I thought to play it safe and used `strings` as Jason Stover had suggested in the mailing list.

gmkurtzer:
I am very weary of introducing points of latency into the invocation pathway, and every time we have to shell out, it adds to the startup overhead. While this this isn't a lot, Singuarity is also being used for some HTC (high throughput) jobs where at 0.020 seconds, it has already been called slow.

I am thinking a better bet might be to do a bind mount over the container's /etc/localtime so we don't have to shell out at all. Are zoneinfo files generally compatible over glibc versions? I suppose that is easy to test.... Are you up for working with me on this and making this a Singularity mount module?

tin6150:
I don't know if timezone files are compatible between glibc, but I would be happy to help test.
But this brings up a good question that I had forgotten to ask: why set TZ at all?

Docker don't seems to set TZ.  Neither does CentOS-6 nor Ubuntu-14.  If you don't set TZ at all in singularity init, then date inside the container would just return time in UTC.  It is okay if it clearly say so, I just had a real problem when it said PST but the time was wrong.    For user who really care about having correct default timezone inside the container, then there could be a FAQ on how they can set TZ correctly.  Wouldn't this work?

gmkurtzer:
Yes, I agree with you about not setting it at all. Let's go with that. Will you update this PR to just remove it from /etc/singularity/init?

Thanks!

gmkurtzer:
Thanks!

PR #431, opened 2016-12-25
Use fixed image offset of zero for SquashFS images

oschulz:
SquashFS images can contain a 0x0a byte within the first 64 bytes, so the current header/offset detection yields wrong results for some SquashFS files, resulting in failure to mount the image.

Fixes #430 

Could this (or a similar solution) be backported to the 2.x branch as well?

PR #459, opened 2017-01-18
Use fixed image offset of zero for SquashFS images

oschulz:
Backport of #431 to current release branch.

### DIFF
--- a/src/lib/image-util.c
+++ b/src/lib/image-util.c
@@ -74,6 +74,11 @@ int singularity_image_offset(FILE *image_fp) {
     int ret = 0;
     int i = 0;
 
+    if ( singularity_image_check(image_fp) != 0 ) {
+        singularity_message(DEBUG, "File is not a Singularity image, returning zero offset\n");
+        return(0);
+    }
+
     singularity_message(VERBOSE, "Calculating image offset\n");
     rewind(image_fp);
 


gmkurtzer:
The more I think about this, the more I consider this is a bug in `singularity_image_offset()` as it should `return(0)` when there is no offset or when it is not a Singularity image. Can you check out what I did in lib-refactor, and see if you can do something similar here?

Thank you!

oschulz:
Should be really fix this in different ways on master and on 2.x, though?

gmkurtzer:
You are right, that it would be better if it wasn't.... But we aren't using the standard GitFlow process right now. With the 2.3 release, we will move to the standard GitFlow model, but for now, what is in lib-refactor is the basis for what will end up in development, and development will end up in master.

oschulz:
Hi @gmkurtzer , I backported your elegant fix for the issue from lib-refactor to here (2.x). I gave it a quick test, seems to work fine. :-)

oschulz:
I rebased this PR, it only contains your nice fix now, Greg (with a tiny change for 2.x).

gmkurtzer:
Thanks!

Hrmm, looks like the busybox container download is failing again. I'm gonna override the CI failure and merge.

oschulz:
Thanks!

PR #469, opened 2017-01-27
fixing container creation for tensorflow gpu definition file

satra:
- updated instructions
   - newer driver version
   - when using inside a virtual machine to create a singularity container
- updated libraries for use with ubuntu 16:10

Fixes #

 -
 -
 -

@singularityware-admin

### DIFF
--- a/examples/contrib/ubuntu16-tensorflow-0.12.1-gpu.def
+++ b/examples/contrib/ubuntu16-tensorflow-0.12.1-gpu.def
@@ -6,7 +6,7 @@
 # are present in the current directory (alongside this definition file):
 #
 #   * cuda-linux64-rel-8.0.44-21122537.run  (* see below)
-#   * NVIDIA-Linux-x86_64-367.48.run        (* see below)
+#   * NVIDIA-Linux-x86_64-375.20.run        (* see below)
 #   * cudnn-8.0-linux-x64-v5.1.tgz          (https://developer.nvidia.com/cudnn)
 #
 # * The cuda-linux64 and NVIDIA-Linux files can be obtained by downloading the
@@ -21,10 +21,15 @@
 # IF YOUR HPC SYSTEM IS USING A DIFFERENT VERSION OF CUDA AND/OR NVIDIA DRIVERS
 # YOU WILL NEED TO ADJUST THE ABOVE VERSION NUMBERS TO MATCH YOUR SYSTEM
 #
+# YOU WILL ALSO NEED TO DOWNLOAD THE APPROPRIATE DRIVER. For example,
+# cuda_8.0.44_linux.run returns driver version 367.48.
+#
+# If you use this to create a container inside a virtual machine with no access to
+# a GPU, comment out the final test.
 
 
 BootStrap: docker
-From: ubuntu:16.04
+From: ubuntu:16.10
 
 
 %runscript
@@ -43,7 +48,7 @@ From: ubuntu:16.04
 %setup
     # Runs from outside the container during Bootstrap
 
-    NV_DRIVER_VERSION=367.48
+    NV_DRIVER_VERSION=375.20
     NV_CUDA_FILE=cuda-linux64-rel-8.0.44-21122537.run
     NV_CUDNN_FILE=cudnn-8.0-linux-x64-v5.1.tgz
     NV_DRIVER_FILE=NVIDIA-Linux-x86_64-${NV_DRIVER_VERSION}.run
@@ -92,9 +97,9 @@ export PATH LD_LIBRARY_PATH
         build-essential \
         curl \
         git \
-        libcurl3-dev \
+        libcurl4-openssl-dev \
         libfreetype6-dev \
-        libpng12-dev \
+        libpng-dev \
         libzmq3-dev \
         python-pip \
         pkg-config \
@@ -132,5 +137,6 @@ export PATH LD_LIBRARY_PATH
     /usr/bin/python -c "import tensorflow as tf"
 
     # Runs in less than 30 minutes on low-end CPU; in less than 2 minutes on GPU
+    # Comment the following line if building the container inside a VM with no access to a GPU
     /usr/bin/python -m tensorflow.models.image.mnist.convolutional
 


vsoch:
Great! I'm giving this a run through now, and seeing if I can make it so the driver is obtained programatically to have a completely automated build. Of course centers that need a different driver won't be able to use that particular version, but it will be a start.

vsoch:
@truatpasteurdotfr want to give this a test run as well? And I am currently/about to test this, but it would be nice if we could programatically install the driver. I'm guessing the license agreement will be the issue... but maybe we can somehow pipe a "yes" in, heh.

satra:
@vsoch - the driver is the easier bit - but cudnn cannot be downloaded without going through nvidia's developer member area.

vsoch:
@satra also give out this one a look https://github.com/GodloveD/TensorFlow/blob/ubuntu_gpu/Singularity - from @GodloveD it's still building on singularity hub! It's [cpu sibling](https://singularity-hub.org/containers/108/) is all finished though.

GodloveD:
I anxious to see whether or not the build dies because I didn't specify the size.  It actually ends up being a little over 4GB but it needs almost 2x that amount when unpacking and installing CUDA and cuDNN.

satra:
@vsoch - @GodloveD has an interesting location to download the cuda pieces from. it would be really nice if we can convince nvidia to allow for public downloads of cudnn.

vsoch:
@GodloveD no worries about size, you can adjust in the build settings if the automatic estimation is too small.

GodloveD:
I think I am reading this right?

http://docs.nvidia.com/cuda/eula/#abstract

1.2.1.4. Linux/FreeBSD Exception
Notwithstanding the foregoing terms of Section 1.2.1.2, Section 1.2.1.1 and Section 1.2.1.3, SOFTWARE designed exclusively for use on the Linux or FreeBSD operating systems, or other operating systems derived from the source code to these operating systems, may be copied and redistributed, provided that the binary files thereof are not modified in any way (except for unzipping of compressed files).

satra:
@GodloveD - from cudnn license, different from cuda license:

```
2.1.1. Usage Rights
Licensee may install and use multiple copies of the SOFTWARE on a shared computer
or concurrently on different computers, and make multiple back-up copies of the
SOFTWARE, solely for Licensee's use within Licensee's Enterprise. "Enterprise" shall
mean individual use by Licensee or any legal entity (such as a corporation or
university) and the subsidiaries it owns by more than 50 percent. 
```

GodloveD:
@satra Well then.  Don't copy it I guess.  That version is only for me.  :-)

vsoch:
@GodloveD I think it built ok!

https://singularity-hub.org/containers/109/

@satra what does that say, in human speak? If there is an issue, then this would mean that we need to make the build private (aka, only for @GodloveD) but maybe we can figure out a way for a user to agree to the license and then build/use the equivalent image? We don't intend to do anything malicious, or make profit, and it would encourage use of their software - so I hope we can figure something out. Is cudnn part of nvidia?

satra:
@vsoch - in human speak it means @GodloveD, if using cudnn, can share it with anyone at NIH, but not outside.

i do think the bigger concern is going to be mismatch between drivers (in the build vs on the system). so that should be made explicit somehow.

vsoch:
A little background to add to this - we have a great Nvidia developer (@flx42) who is working with @gmkurtzer on a system low level library that will be a direct integration into Singularity and automatically find and mount devices and libraries.

So what that means for this issue, and long term, problems similar to it, is that Singularity will have an integration to handle these discrepancies, and given some agreement like NVidia has with Docker, there will be a way to distribute the software for users. For the short term, what it means is that we should make that gpu image private (and @GodloveD what I've just done is increase your number of private collections allowed, so this should not be an issue) and we probably should just provide the gpu spec file here.

@satra what are your goals specifically - do you need an image on singularity hub or is the spec file (with manual instructions) substantial enough for now?

For both - just for curiosities sake - does this version work for you?

https://singularity-hub.org/containers/105/

It's the same thing, but coming from their Docker image.

GodloveD:
@satra Thanks for bringing this to my attention.  I'm going to move this software to an internal location so that it can't be downloaded from a public ftp.  @vsoch I went ahead and made those containers private.  Thanks for the heads up.

satra:
@vsoch - for us spec file is fine at the moment - we plan to create a local hub as well because undoubtedly there will be variants that are going to be tied to our specific hardware and driver versions.

re: docker tensorflow: doesn't work on our cluster.

```
 $ /usr/bin/python -c "import tensorflow as tf"
I tensorflow/stream_executor/dso_loader.cc:111] successfully opened CUDA library libcublas.so locally
I tensorflow/stream_executor/dso_loader.cc:111] successfully opened CUDA library libcudnn.so locally
I tensorflow/stream_executor/dso_loader.cc:111] successfully opened CUDA library libcufft.so locally
I tensorflow/stream_executor/dso_loader.cc:105] Couldn't open CUDA library libcuda.so.1. LD_LIBRARY_PATH: /usr/local/cuda/lib64:/usr/local/cuda/lib64/stubs:
I tensorflow/stream_executor/cuda/cuda_diagnostics.cc:160] hostname: node027.cm.cluster
I tensorflow/stream_executor/cuda/cuda_diagnostics.cc:185] libcuda reported version is: Not found: was unable to find libcuda.so DSO loaded into this program
I tensorflow/stream_executor/cuda/cuda_diagnostics.cc:356] driver version file contents: """NVRM version: NVIDIA UNIX x86_64 Kernel Module  375.20  Tue Nov 15 16:49:10 PST 2016
GCC version:  gcc version 4.8.5 20150623 (Red Hat 4.8.5-4) (GCC) 
"""
I tensorflow/stream_executor/cuda/cuda_diagnostics.cc:189] kernel reported version is: 375.20.0
I tensorflow/stream_executor/cuda/cuda_gpu_executor.cc:1080] LD_LIBRARY_PATH: /usr/local/cuda/lib64:/usr/local/cuda/lib64/stubs:
I tensorflow/stream_executor/cuda/cuda_gpu_executor.cc:1081] failed to find libcuda.so on this system: Failed precondition: could not dlopen DSO: libcuda.so.1; dlerror: libcuda.so.1: cannot open shared object file: No such file or directory
I tensorflow/stream_executor/dso_loader.cc:111] successfully opened CUDA library libcurand.so locally
```

vsoch:
ok, understood. In that case, I'm ok with updating the example file in this repo, although I'll leave the final decision up to @gmkurtzer. Minimally, the improved documentation and commenting in your version is a great improvement! And you are building your own hub? I guess that puts me out of work for my Singularity Hub project, haha. But there are lots of interesting things to build, and I think someone will find it useful still, so I'm ok with this :)

satra:
@vsoch - no we plan to use singularity hub for most things, but will have a local cache so that everybody doesn't have to download an image each time. on a related note, we should consider setting up the public images over ipfs (ipfs.io).

vsoch:
If ipfs better / faster than Google storage?

vsoch:
What I was thinking that might eventually be warranted (but isn't currently set up) is to have the APi (to link to images in Google Storage) served as a separate instance from the web client where users look around. I think to start it is reasonable for a big server (with several workers) to handle, but if we can imagine some future where the load becomes too big, we would want on demand / deployable specialized workers just for connecting to the database and serving a (mode modular) API.

GodloveD:
@satra and @vsoch I just spoke with @flx42.  I think I might be able to accomplish the same thing using this as a starting point and then unpacking the driver binaries, setting the links and paths blah blah blah.  I have some other stuff to do first, so if want to try in the meantime let me know how it works.  
https://hub.docker.com/r/nvidia/cuda/

vsoch:
oh sweet!! Yeah no worries about timing, I'll be around, and might give it a go if I can finish up (my set of) other things too, haha. Happy Friday!

gmkurtzer:
This PR seems ready for merge. I skimmed the comments, and it seemed like we have thumbs up; lots of comments but no reviews positive or negative. So I'm merging...

Thanks!

GodloveD:
@satra @vsoch:  Just to follow up and finish this thread: This definition file gives me a working TensorFlow with CUDA and cuDNN installed programatically and legally.  It remains legal since it is technically still downloading both CUDA and cuDNN from an official NVIDIA source (their official DockerHub).  
```
BootStrap: docker
From: nvidia/cuda:8.0-cudnn5-devel

%post
    # add universe repo and install some packages
    sed -i '/xenial.*universe/s/^#//g' /etc/apt/sources.list
    locale-gen en_US.UTF-8
    apt-get -y update
    apt-get -y install vim wget perl python python-pip python-dev

    # download and run NIH HPC NVIDIA driver installer
    wget ftp://helix.nih.gov/CUDA/cuda4singularity
    chmod 755 cuda4singularity
    ./cuda4singularity --verbose
    rm cuda4singularity

    # install tensorflow
    pip install --upgrade pip
    pip install tensorflow-gpu
```  
The bit about installing the NIH HPC NVIDIA driver can be replaced with a script that installs whatever driver matches your setup.  That is legal because (unlike cuDNN) the Linux version of the driver is free to redistribute.   Whenever NIVIDIA gets a working Singularity version of NVIDIA Docker up and running that whole section will be able to be deleted (in theory).  Hope that helps.  :smiley:

vsoch:
@GodloveD badass! Let's definitely have this spec / build on Singularity Hub, at whatever level is allowable.

flx42:
Any reason you're not using the TensorFlow images directly?

GodloveD:
I feel like this is more general.  One of our users could start with the same definition file and install Theano or Keras or whatever just by changing the last few lines.  But that said the TensorFlow images may work fine (though @satra indicated that they don't work out of the box for him above.)

flx42:
It should work, I believe only old versions of TensorFlow had this `libcuda.so` issue.

satra:
@flx42 - which tensorflow images are your referring to? i'm happy to try again. but i don't see how a generic image can be used. 

the key problem with any image is that it needs the appropriate GPU driver that's on the existing system. are the images built for every driver released by nvidia?

@vsoch - would that be worth considering for gpu-layered images? to build specs with multiple drivers - allowing a matrix of conditions in the spec - almost like a travis or circle matrix?

satra:
@GodloveD - pulling from nvidia docker images is a great idea. i think the driver install can point directly to an nvidia driver download given the version. i'll try to create a spec for our cluster based on your spec.

GodloveD:
@satra, It was actually @flx42's idea.  :smile:   Please keep me updated on your efforts to download from NVIDIA.  I was unable to find an exact match our current driver (`367.48`) on NVIDIA's site, but I didn't do an exhaustive search.

flx42:
@satra I'm saying that instead of doing the following:
- From: nvidia/cuda
- Install TensorFlow
- Install NVIDIA drivers

You could do the following instead:
- From: tensorflow/tensorflow
- Install NVIDIA drivers

@GodloveD: major driver releases are actually at the same URL on the NVIDIA website:
```
$ curl -fsSL http://us.download.nvidia.com/XFree86/Linux-x86_64/367.57/NVIDIA-Linux-x86_64-367.57.run -O
$ curl -fsSL http://us.download.nvidia.com/XFree86/Linux-x86_64/375.26/NVIDIA-Linux-x86_64-375.26.run -O
```
The pattern is easy to find ;)

GodloveD:
@flx42  Thanks!  I don't see our minor release though.  The following doesn't exist.
```
http://us.download.nvidia.com/XFree86/Linux-x86_64/367.57/NVIDIA-Linux-x86_64-367.48.run
```
But reading between the lines, maybe you are telling me that the `367.57` binaries and libraries should be backward compatible with the `367.48` driver running on the host.  Is that true?  If so, maybe I don't have to maintain the minor release locally.

flx42:
No, there is no backward compatibility at all with our drivers. The kernel module version must match *exactly* the version of the driver libraries.
But I'm saying you should stick to the versions on the official NVIDIA website. For most users, it's the best strategy. I don't see any way to install `367.48` right now if you want to deploy it on a new cluster.

satra:
@GodloveD - in terms of driver searches, this link may help:

http://www.nvidia.com/object/linux-amd64-display-archive.html

satra:
btw, 367.48 is included in the current cuda toolkit download. it's possible that it was never released as a standalone driver download.

GodloveD:
Yes, we unpacked it from the cuda toolkit to install it onto the cluster.  Seems a little silly to download the entire toolkit and unpack it within a container just for the driver binaries and libraries.  So I guess I'm stuck hosting it on our ftp.  

But anyway thanks very much for all of your great advice @flx42!  It is always invaluable.

flx42:
In this case, you might be able to use the CUDA network repo to install `367.48` then:
http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1604/x86_64/
Package is called `nvidia-367`

On a Ubuntu host system, I recommend that you install the NVIDIA driver through this repo, you will automatically get the update to `367.58`, since Canonical is also packaging our driver:
```
$ apt-cache policy nvidia-367
nvidia-367:
  Installed: 367.57-0ubuntu0.16.04.1
  Candidate: 367.57-0ubuntu0.16.04.1
  Version table:
 *** 367.57-0ubuntu0.16.04.1 500
        500 http://us.archive.ubuntu.com/ubuntu xenial-updates/restricted amd64 Packages
        500 http://security.ubuntu.com/ubuntu xenial-security/restricted amd64 Packages
        100 /var/lib/dpkg/status
     367.48-0ubuntu1 500
        500 http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1604/x86_64  Packages
```

satra:
@flx42 and @GodloveD - this spec file worked on our system and is currently building on singularity hub. @vsoch - thanks for the help on syntax.

https://github.com/satra/om-tensorflow/blob/master/Singularity

i do get some warnings about how tensorflow was compiled without some cpu flags, but it does find my gpu and runs.

```
I tensorflow/stream_executor/dso_loader.cc:135] successfully opened CUDA library libcublas.so.8.0 locally 
I tensorflow/stream_executor/dso_loader.cc:135] successfully opened CUDA library libcudnn.so.5 locally 
I tensorflow/stream_executor/dso_loader.cc:135] successfully opened CUDA library libcufft.so.8.0 locally 
I tensorflow/stream_executor/dso_loader.cc:135] successfully opened CUDA library libcuda.so.1 locally
I tensorflow/stream_executor/dso_loader.cc:135] successfully opened CUDA library libcurand.so.8.0 locally
W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use SSE4.2 instructions, but these are available on your machine and could speed up CPU computations.
W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use AVX instructions, but these are available on your machine and could speed up CPU computations.
W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use AVX2 instructions, but these are available on your machine and could speed up CPU computations.
W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use FMA instructions, but these are available on your machine and could speed up CPU computations.
I tensorflow/core/common_runtime/gpu/gpu_device.cc:885] Found device 0 with properties: 
name: GeForce GTX TITAN X
major: 5 minor: 2 memoryClockRate (GHz) 1.076
pciBusID 0000:06:00.0
Total memory: 11.95GiB
Free memory: 11.84GiB
I tensorflow/core/common_runtime/gpu/gpu_device.cc:906] DMA: 0 
I tensorflow/core/common_runtime/gpu/gpu_device.cc:916] 0:   Y 
I tensorflow/core/common_runtime/gpu/gpu_device.cc:975] Creating TensorFlow device (/gpu:0) -> (device: 0, name: GeForce GTX TITAN X, pci bus id: 0000:06:00.0)
```

satra:
just an update that the hub build works as well.

GodloveD:
Rock and roll

PR #479, opened 2017-02-08
added an ERRNO==ENOENT clause

Nekel-Seyew:
Fixed an issue with group ids that don't actually exist, such as in the case with AFS and tokens.

 - add the clause (errno==ENOENT) to line 143 in group.c
 -
 -

@singularityware-admin


gmkurtzer:
Thanks!

PR #481, opened 2017-02-09
Let mkfs.ext3 use the image file directly instead of loopback device

wpoely86:
By avoiding the loopback device, we do not need root priviledges to create an
image. We give mkfs.ext3 an offset and image size.

This is NOT ready: it's only meant to show that it is possible.

@singularityware-admin


wpoely86:
The failing test is normal, I haven't adjusted the build system.

wpoely86:
OK, this only works with E2fsprogs 1.42.10 (May 18, 2014) and newer.

gmkurtzer:
I would support making this an optional pathway.... Probably makes the most sense to check for this feature in `configure.ac` and then conditionalize the SUID generation. Let me know if you need a hand.

wpoely86:
Okay, it took some effort but it finally works. There was an error in the overlay causing the tests to fail that I also fixed.

configure.ac
```diff
@@ -274,12 +274,25 @@ AC_ARG_ENABLE([suid],
 
 AS_IF([test "x$enable_suid" != "xno"], [
     AC_MSG_RESULT([yes])
-    BUILD_SUID="action-suid create-suid copy-suid expand-suid export-suid import-suid mount-suid"
+    BUILD_SUID="action-suid copy-suid expand-suid export-suid import-suid mount-suid"
```
gmkurtzer, line 287: Can you confirm that if the below check for new enough mkfs.ext3 fails, we need to add create-suid back into the BUILD_SUID list.

wpoely86: That happens on Line 291: it `$BUILD_SUID` is non-empty, we add `create-suid` to the list:
```
AS_IF([test -n "$BUILD_SUID"], [BUILD_SUID="$BUILD_SUID create-suid"], [])
```

wpoely86:
Btw, @gmkurtzer is there a specific reason to choose ext3 over ext4?

configure.ac
wpoely86: You can check by looking at the build on Travis: it's uses an old version of mkfs.ext3 so it needs SUID rights to work.

gmkurtzer: Ahh, right you are. Sorry for missing that!

gmkurtzer:
Yep, ext4 versions are not as stable as ext3 (stable in the sense that an ext4 system created on latest Fedora won't work with RHEL5/6).

src/create.c
```diff
@@ -74,24 +78,45 @@ int main(int argc, char **argv) {
...
+#ifdef SUID_CREATE
     mkfs_cmd[2] = strdup(image.loopdev);
     mkfs_cmd[3] = NULL;
+#else
+    mkfs_cmd[2] = strdup("-E");
+    // the offset in the file for the singularity header
+    mkfs_cmd[3] = strjoin("offset=", int2str(strlength(LAUNCH_STRING, 1024)));
+    mkfs_cmd[4] = strdup(image.path);
+    // pass the correct size of the file in KiB
+    mkfs_cmd[5] = int2str((size*1024*1024-strlength(LAUNCH_STRING, 1024))/1024);
```
gmkurtzer: Is this necessary?

wpoely86: If you don't pass it, you get (even with `-q`):
```
Warning: offset specified without an explicit file system size.
Creating a file system with 131072 blocks but this might
not be what you want.
```
So I think it's safer to add it.

wpoely86:
@gmkurtzer I just noticed that: 
```
singularity -d export image.iso > image.tar
```
doesn't work. The first debug output of singularity goes to stdout instead of stderr. I think the only clean way of solving this is not to use the redirect and do something like:
```
singularity -d export image.iso image.tar
```

wpoely86:
@gmkurtzer I've changed the target to the `development` branch.

gmkurtzer:
Thanks for the merge with the latest "development" branch code.

So I am now on the fence on this PR as the 2.3 release will be able to create images without `sudo` already without this:

```bash
[gmk@centos7-x64 ~]$ whoami
gmk
[gmk@centos7-x64 ~]$ singularity create /tmp/gmk-test.img
Initializing Singularity image subsystem
Opening image file: gmk-test.img
Creating 768MiB image
Binding image to loop
Creating file system within image
Image is done: /tmp/gmk-test.img
```

So the main value add is that we can do it without a SUID pathway, but it also creates an alternate pathway that can end up making things more complicated to debug and maintain.

Considering we can do this now, do you still see a strong need for this PR?

wpoely86:
Strong case, probably not. It's nice to not need root nor suid for a simple thing as creating an image but for the end-user this makes no difference.

I do think the overhead of the extra path is minimal: it's all contained in a single file as ifdefs.

gmkurtzer:
As of the 2.4 beta release, this is no longer necessary. We don't use the loop device anymore to create images. Thanks!

PR #486, opened 2017-02-13
this would likely handle space, comment and newline

### LINKED ISSUE #485: bootstrap: should cut comments first in From
Currently, if the user bootstraps an image an has a comment after the name:

      Bootstrap: docker
      From: tensorflow/tensorflow:awesome-version  # like omgtacos

The bootstrap looks for a newline for the tag, and parses the comment as the image version. This gets passed incorrectly to the python and as a result, the "latest" version is incorrectly used.

The workaround for now is to not have comments on the same line as the `From`

      Bootstrap: docker
      # like omgtacos
      From: tensorflow/tensorflow:awesome-version

but it should be reasonable for the user to have comments on the same line - and they should be parsed / snipped off before extracting the tag, to be passed on. This will currently affect both docker and singularity hub, passing of tags. Thanks @satra for finding this bug!

satra:
Fixes #485

 - add additional delimiters for determining end of token.

is there a directory for adding tests?

@singularityware-admin

### DIFF
--- a/src/lib/image/bootstrap/bootdef_parser.c
+++ b/src/lib/image/bootstrap/bootdef_parser.c
@@ -109,7 +109,7 @@ char *singularity_bootdef_get_value(char *key) {
         if ( ( bootdef_key = strtok(line, ":") ) != NULL ) {
             chomp(bootdef_key);
             if ( strcasecmp(bootdef_key, key) == 0 ) {
-                if ( ( bootdef_value = strdup(strtok(NULL, "\n")) ) != NULL ) {
+                if ( ( bootdef_value = strdup(strtok(NULL, " #\n")) ) != NULL ) {
                     chomp(bootdef_value);
                     singularity_message(VERBOSE2, "Got bootstrap definition key %s(: '%s')\n", key, bootdef_value);
                     return(bootdef_value);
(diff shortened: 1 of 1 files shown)


bauerm97:
>is there a directory for adding tests?

Tests go into `test.sh.in`. If you need to add a definition file to be used with the test, you can add one to `examples/contrib`.

bauerm97 (approved):
Simple fix to the problem, good catch and thanks for submitting the PR.

gmkurtzer:
@satra Do you want to add a test, if so I will hold off on merging.

Thanks!

satra:
@gmkurtzer - let me add a test.

satra:
@gmkurtzer - so i looked into the tests and this would require adding a verbose output to the bootstrap with docker step to print out the name/tag of the container and check against the intent of some example bootstrap file.

so something that would print the contents of this variable:

https://github.com/singularityware/singularity/blob/master/src/lib/image/bootstrap/docker/docker.c#L47

would adding another print be ok?

gmkurtzer:
Ohhh, I didn't even notice this before... You are in the master branch! Sorry... Check out the lib-refactor branch. That is the direction we are going in for bootstrapping.

sorry for the confusion!

satra:
@gmkurtzer - will do. a related question, should i ignore master and focus on lib-refactor even for users who want to be on the blood-sucking edge ? for example, currently master can download from `shub://` but `rel 2.2.1` cannot.

gmkurtzer:
Yes, master will actually end up going away. it is a dead end branch right now. I know this sucks, and I apologize for that. It took me longer then it should have to understand the GitFlow development model.

Right now all of the latest devel is going into lib-refactor. Once that is feature ready/equiv with master, then it will be merged into the branch development. From there we will work on stability, and will merge with master as soon as 2.3 is released.

satra:
thanks @gmkurtzer - perhaps the README on master can redirect people to lib-refactor. i'll close this pull request then and send one for lib-refactor.

PR #487, opened 2017-02-14
Use fallocate to allocate the images when creating.

wpoely86:

 -  Switch to using `posix_fallocate` for allocation of image files
 -  Handle EINTR
 -  Fix warnings by Clang

@singularityware-admin


src/lib/image/bind/bind.c
```diff
@@ -99,10 +99,7 @@ int _singularity_image_bind(struct image_object *image) {
 #endif
 
     singularity_message(DEBUG, "Calculating image offset\n");
-    if ( ( lo64.lo_offset = singularity_image_offset(image) ) < 0 ) {
-        singularity_message(ERROR, "Could not obtain message offset of image\n");
-        ABORT(255);
-    }
+    lo64.lo_offset = singularity_image_offset(image);
```
wpoely86, line 90: `lo_offset` is unsigned and `singularity_image_offset` never returns anything less then 0.

gmkurtzer: fair enough. I left that in there though incase we update singularity_image_offset() to return a negative number on error. But in that case, we can look for calls to it and update accordingly.

src/lib/image/create/create.c
```diff
@@ -57,20 +53,42 @@ int _singularity_image_create(struct image_object *image, long int size) {
...
     singularity_message(VERBOSE2, "Growing image to %ldMB\n", size);
-    // TODO: there are likely better ways to do this (falloc?); further, we should really handle
-    // EINTR here.
-    for(i = 0; i < size; i++ ) {
-        if ( fwrite(buff, 1, BUFFER_SIZE, image_fp) < BUFFER_SIZE ) {
-            singularity_message(ERROR, "Failed allocating space to image: %s\n", strerror(errno));
-            ABORT(255);
+    while(1)
+    {
+        retval = posix_fallocate(image->fd, sizeof(LAUNCH_STRING), size*BUFFER_SIZE);
```
gmkurtzer: I forget (and the man page didn't seem very clear on this), does posix_fallocate() zero out the allocated disk space, or is it sparse? Thanks!

wpoely86: As far as I understand, it's filled with zeros. A quick test seems to confirm this.

wpoely86:
As reference: http://www.lucas-nussbaum.net/blog/?p=331#comment-21515

So, `posix_fallocate` should not create a sparse file.

gmkurtzer:
This has been merged into the development branch. Thank you!

PR #503, opened 2017-03-04
updating python tests

vsoch:
This should at least be a start to fixing up the python tests. There were a few issues:

- the actual testing for the import/add/pull for both of docker and shub weren't written. It looks like I started and then promptly forgot, so I've added a few completely new files. This is a more challenging test case because the intended use is to call from command line but we are running tests from within python. They also expect environment variables to be defined, and folders to be created. To address this, I do these exports/creations upon setup, and then use subprocess to pipe commands to the command line (as the real calling singularity would) and then check for an exit code of 0. I also print the terminal output for the viewer in case it's needed.
- I fixed a few imports / header details
- the runscript function was updated to not need a base directory (as it gets it from the environment) and this wasn't updated in the tests. I also saw a potential bug if the user tried using the function without defining the `SINGULARITY_METADATA_FOLDER` in the environment - it would try to write to a path like 

      /None/.singularity/runscript

So I do a check for this first. If the metadata folder is not defined, it returns the runscript as a string (and gives the user a warning).

Finally, I moved all the tests into a bash script to be run instead of using the unittest discovery tool. The reason is pretty silly, although I don't see a workaround for it other than calling the commands separately. To test a lot of the functions that require changing environment variables, it means we need totally separate python processes. The cache functions were failing because we had set them to be something upon init, and then our changes didn't take effect.This method works for me locally so I'm hoping it works on travis.

I'll push updates to address any/all total failures, lol.

@singularityware-admin

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ script:
   - make test
   - jdk_switcher use oraclejdk8
   - sonar-scanner
-  - cd $TRAVIS_BUILD_DIR/libexec/python && python2 -m unittest discover --start-directory tests
-  - cd $TRAVIS_BUILD_DIR/libexec/python && python3 -m unittest discover --start-directory tests
+  - cd $TRAVIS_BUILD_DIR/libexec/python/tests && ./run_tests_python2.sh
+  - cd $TRAVIS_BUILD_DIR/libexec/python/tests && ./run_tests_python3.sh
   - pylint --version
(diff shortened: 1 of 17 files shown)


gmkurtzer:
Hi @v, thanks for handling this! Is it ready for merge?

vsoch:
This part is good, but let's add the other additions for the shub layerfile and adjusted name, so I can just push to this PR :)

gmkurtzer:
ok, awesome. Thanks!

vsoch:
@gmkurtzer this is ready for merge, and we probably should so I can push another PR to show you the proposed changed to the python libexec structure!

PR #508, opened 2017-03-10
Read only userbinds rebased

### LINKED ISSUE #451: Ability to bind as READ-ONLY
Ability to bind as READ-ONLY, example:

--robind /path/to/host/dir:/path/to/container/READ-ONLY/dir/
--robind /path/to/host/dir

robind == read only bind

jobh:
Fixes #451

 - Allow specifying additional --bind option ":ro" to perform a read-only bind mount.

@singularityware-admin

### DIFF
--- a/libexec/cli/exec.help
+++ b/libexec/cli/exec.help
@@ -4,9 +4,12 @@ This command will allow you to execute any program within the given
 container image.
 
 EXEC OPTIONS:
-    -B/--bind <spec>    A user-bind path specification.  spec can either be a
-                        path or a src:dest pair, specifying the bind mount to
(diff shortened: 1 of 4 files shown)


jobh:
Note: Recursive mounts may not work as expected. Whether submounts are remounted read-only is an open question, and possibly depends on the kernel version (a google search implies it doesn't, or didn't, recurse). Even if it does, the final check for access(..., W_OK) would be painful to repeat for submounts.

Note 2: CI build fails, but the error seems to be the same as on the lib-refactor branch,

gmkurtzer:
Thanks!

PR #521, opened 2017-03-20
Use correct variable for secondary groups

wpoely86:
The wrong variable was used.

@singularityware-admin


gmkurtzer:
Nice catch! Thanks! ... Makes me wonder how this has worked so far!

wpoely86:
Thank the Clang static analyser 😉 
It found a couple of other issues but all memory related (memory not being free'd and such). I will try to fix them and send patches.

gmkurtzer:
Awesome, thank you!

PR #534, opened 2017-03-23
Catch weird RPM dbpath values

### LINKED ISSUE #241: singularity (git version) bootstrapping CentOS-7 container from debian jessie or ubuntu trusty
reported on #singularity irc channel and reproduced locally.

```
sudo singularity create a.img
sudo singularity bootstrap a.img ~/git/singularity/examples/centos.def
singularity exec a.img rpm -q centos-release
```

bootstrap fails during %post when yum install vim-minimal with:

```
+ yum -y install vim-minimal
Loaded plugins: fastestmirror


 One of the configured repositories failed (Unknown),
 and yum doesn't have enough cached data to continue. At this point the only
 safe thing yum can do is fail. There are a few ways to work "fix" this:

     1. Contact the upstream for the repository and get them to fix the problem.

     2. Reconfigure the baseurl/etc. for the repository, to point to a working
        upstream. This is most often useful if you are using a newer
        distribution release than is supported by the repository (and the
        packages for the previous distribution release still work).

     3. Disable the repository, so yum won't use it by default. Yum will then
        just ignore the repository until you permanently enable it again or use
        --enablerepo for temporary usage:

            yum-config-manager --disable <repoid>

     4. Configure the failing repository to be skipped, if it is unavailable.
        Note that yum will try to contact the repo. when it runs most commands,
        so will have to try and fail each time (and thus. yum will be be much
        slower). If it is a very temporary problem though, this is often a nice
        compromise:

            yum-config-manager --save --setopt=<repoid>.skip_if_unavailable=true

Cannot find a valid baseurl for repo: base/$releasever/x86_64
Could not retrieve mirrorlist http://mirrorlist.centos.org/?release=$releasever&arch=x86_64&repo=os&infra=stock error was
14: curl#56 - "Recv failure: Connection reset by peer"
ERROR: Aborting with RETVAL=255
package centos-release is not installed
```

imho /var/lib/rpm is not populated correctly, shell inside the container hosted on jessie/trusty verify that content of /var/lib/rpm is smaller than what I get from a centos-7 hosted host.

akkornel:
Hello!

The yum bootstrap mechanism has an issue when running on a Debian-family (Debian, Ubuntu, etc.) host: Their packaged `rpm` and `yum` place the RPMDB—which should live in `/var/lib/rpm`—into a weird location: `~/.rpmdb`.

Here is the original Debian bug for the change: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=551669#34
And here is an RPM project bug pointing back to the Debian bug, when people started wondering why they had weird `.rpmdb` directories appearing everywhere: https://bugs.launchpad.net/rpm/+bug/1069350

This patch recognizes when the RPMDB path is weird, and aborts the bootstrap.  Instructions are provided to the user on how to fix the problem.

Fixes #241

The following changes are made to the yum bootstrap mechanism:

- Now explicitly requires `rpm`, and aborts if it is not in `$PATH`.
- Aborts the bootstrap if the RPM variable `_dbpath` is not set to `%{_var}/lib/rpm`.  That is the default value from CentOS 6-7 and Fedora 25 (those are the ones I tested).
- If aborting the bootstrap, provides the user instructions on which RPM config file needs to be changed, and what changes must be made.

Tested bootstrapping on an Ubuntu 16.10 host.

@singularityware-admin

### DIFF
--- a/libexec/bootstrap-scripts/deffile-driver-yum.sh
+++ b/libexec/bootstrap-scripts/deffile-driver-yum.sh
@@ -65,6 +65,26 @@ else
     ABORT 1
 fi
 
+# Check for RPM's dbpath not being /var/lib/rpm
+RPM_CMD=`singularity_which rpm`
+if [ -z "${RPM_CMD:-}" ]; then
+    message ERROR "rpm not in PATH!\n"
+    ABORT 1
+fi
+RPM_DBPATH=$(rpm --showrc | grep -E ":\s_dbpath\s" | cut -f2)
+if [ "$RPM_DBPATH" != '%{_var}/lib/rpm' ]; then
+    message ERROR "RPM database is using a weird path: %s\n" "$RPM_DBPATH"
+    message WARNING "You are probably running this bootstrap on Debian or Ubuntu.\n"
+    message WARNING "There is a way to work around this problem:\n"
+    message WARNING "Create a file at path %s/.rpmmacros.\n" "$HOME"
+    message WARNING "Place the following lines into the '.rpmmacros' file:\n"
+    message WARNING "%s\n" '%_var /var'
+    message WARNING "%s\n" '%_dbpath %{_var}/lib/rpm'
+    message WARNING "After creating the file, re-run the bootstrap.\n"
+    message WARNING "More info: https://github.com/singularityware/singularity/issues/241\n"
+    ABORT 1
+fi
+
 if [ -z "${OSVERSION:-}" ]; then
     if [ -f "/etc/redhat-release" ]; then
         OSVERSION=`rpm -qf --qf '%{VERSION}' /etc/redhat-release`


akkornel:
I forgot to include asciinemas!

This shows the problem: https://asciinema.org/a/88rnb31zl6qjtb7xrvyi6jyeb
This shows my patch in action: https://asciinema.org/a/59dbj39cxwnymlrk5haj9yujj

gmkurtzer:
Ahh, very cool @akkornel! Thank you!

BTW, in the development branch, you don't need sudo to create the image. :)

akkornel:
@gmkurtzer Thanks very much!

BTW, I actually do, because of a weirdness: I set up this VM with ZFS as the root, and their instructions have you set the home directory mounts as `nosuid`.  So, I did have to `sudo` in that case. 8-)

gmkurtzer:
Ahhhhh, and uh-oh... Root file systems mounted with `MS_NOSUID` may cause some problems. We may want to document this!

Thanks for the heads up!

PR #536, opened 2017-03-24
Python and Numerous updates to Development

### LINKED ISSUE #446: Roadmap to unprivileged Docker Hub image import
As for many users my typical workflow is as follows:
1. develop and test code inside a docker container on a local machine (in my case that's Windows, but for many it's Mac)
2. convert the image using docker2singularity
3. transfer the image to the cluster I'm using

The main problem is that container images can get big and with this workflow even the smallest change requires recreating and copying the full container image. I would much prefer the following workflow:

1. develop and test code inside a docker container on a local machine (in my case that's Windows, but for many it's Mac)
2. push the image to Docker hub (provides extra backup and can be sped up because of layer caching)
3. on the cluster import the image from docker hub (taking advantage of local cache of previously downloaded layers)

Step 3 has some constraints: it should not require root privileges or special kernel features.

It seems that all of the building blocks are in place, but they have not been put together. In fact @oschulz created a handy [docker2singularity.py](https://github.com/oschulz/singularity-utils/blob/master/bin/docker2singularity.py) script that creates singularity compatible images (in "directory" or "squashfs" formats) using nothing else (not even singularity itself) than `docker2aci` and not requiring root privileges. Since @vsoch implemented support for docker registry API and layer caching the same functionality could be replicated in singularity for the convenience of users. There are a few things that need to be cleaned up:

- [x] permissions: exclude /dev to avoid permission issues when untarring layers as recommended by @oschulz in https://github.com/singularityware/singularity/issues/364#issuecomment-269201488; make all files world readable and all executable files world executable to make the new container more robust to non-root users
- [x] environment variables: extract variables from the manifest I believe there is a PR for this already https://github.com/singularityware/singularity/pull/376/files
- [ ] mount points: by default we should create mount points available on the system the container is created/imported, but it would be nice to give an option to create additional mount points

Ideally, I would like to do something like: `singularity create tensorflow.tar.gz --from-docker-hub tensorflow/tensorflow:0.12.1`. This would THE killer feature for Singularity 2.3!

**Update 1**: added mention of the mount points

### LINKED ISSUE #528: renaming of metadata folder for development branch
This is my mistake - but @satra suggested (and thinking it through I agree) that we should not have the meta data folder be hidden. Instead, something along the lines of this:

```bash
/singularity.d
  /environment
	10-docker.sh
	99-userdef.sh
	..etc
  labels.json
```

@gmkurtzer thoughts?

### LINKED ISSUE #532: Bootstrap fails when using a Docker sha256 image digest
Docker image can be referenced by using immutable sha256 digest eg. 

```
docker pull cbcrg/dkrbase@sha256:dee861c31a99556685e91a2e00f3fdace770132d573e11d216d30950c35b93b6
```

This is useful because it guarantees that the image cannot change over time as it can happen when referencing it by using a tag. 

Singularity it's able to bootstrap an image from a Docker one but it fails when such image is specified by using the digest identifier. For example: 

```
Bootstrap:docker
From:nextflow/rnatoy@sha256:ffe9ec577f731b3c4479537aa5d37da97c954610a25e30900a33a53a278cc036
```

```
$ sudo singularity bootstrap my_image singularity_file 
Adding label: 'SINGULARITY_CONTAINER_UUID' = 'ba2c0c5b-a3c4-43e1-a462-25e21f2c908b'
Adding label: 'SINGULARITY_DEFFILE' = 'Singularity'
Adding label: 'SINGULARITY_BOOTSTRAP_VERSION' = '2.2.99'
Adding label: 'SINGULARITY_DEFFILE_BOOTSTRAP' = 'docker'
Adding label: 'SINGULARITY_DEFFILE_FROM' = 'nextflow/rnatoy@sha256:ffe9ec577f731b3c4479537aa5d37da97c954610a25e30900a33a53a278cc036'
```

vsoch:
This is a pretty substantial PR that fixes several issues.

## Addition of Python API Client
The idea here is, instead of passing around the same variables for every function, we generate a client (instantiated from a [common class](https://github.com/vsoch/singularity/blob/updates/development/libexec/python/base.py) for shub and docker) and then we can use functions off of that. Then we can do stuff like the following:

```python
from docker.client import DockerApiConnection
client = DockerApiConnection(image="ubuntu:latest")
image.get_manifest()
```
etc.
This allows for the next thing,
 - Cleaner authentication flow. We now automatically try refreshing the token given a 401 error, instead of the manual force of it we had previously. I've also cleaned up the functions into a client.get, client.prepare_request, and client.submit_request.
 - Addition of allowing a sha digest to close #532. This comes down to, given finding a digest in the URI, using it in preference of the tag.
- Renaming metadata folder to close #528. I liked `singularity.d` best so I chose that one. It took me over 2 hours to find that the extracted environment base was in a `environment.tar` file, which isn't even a real tar, it's a compressed gzip! And that was after changing all those file paths! @gmkurtzer you owe me a :cookie: !! or more stickers. I can be pursuaded :)
- Addition of Docker environment and labels without sudo import to finally close #446. Note that I tested this manually, and can confirm that the in memory tar gets written to the cache:

metadata

and then the layer is passed with the others in the layerfile:

```bash
/home/vanessa/.singularity/docker/sha256:d54efb8db41d4ac23d29469940ec92da94c9a6c2d9e26ec060bebad1d1b0e48d.tar.gz
/home/vanessa/.singularity/docker/sha256:f8b845f45a87dc7c095b15f3d9661e640ebc86f42cd8e8ab36674846472027f7.tar.gz
/home/vanessa/.singularity/docker/sha256:e8db7bf7c39fab6fec91b1b61e3914f21e60233c9823dd57c60bc360191aaf0d.tar.gz
/home/vanessa/.singularity/docker/sha256:9654c40e9079e3d5b271ec71f6d83f8ce80cfa6f09d9737fc6bfd4d2456fed3f.tar.gz
/home/vanessa/.singularity/docker/sha256:6d9ef359eaaa311860550b478790123c4b22a2eaede8f8f46691b0b4433c08cf.tar.gz
/home/vanessa/.singularity/docker/metadata-index.docker.io-library-ubuntu:latest.tar.gz
```
but I am waaay too zonked and tired to give this proper testing - others help would be appreciated! It basically works by generating a list of files to add, one for labels, one for environment:

```python
[{'content': 'export HELLOKITTY=yes',
  'name': './singularity.d/env/10-docker.sh',
  'permission': '0755'},
 {'content': '{"label1":1}',
  'name': './singularity.d/labels.json',
  'permission': '0755'}]
```
and the default ownership is root root. It then creates the tar in memory, and saves it as a .tar.gz with the other layers that are to be imported:

```python
tar_file = create_tar(files,output_file)
tar_file
 '/home/vanessa/.singularity/docker/metadata-index.docker.io-library-ubuntu:latest.tar.gz
```
I am a bit concerned about the future of Docker manifests - for this version I updated the calls to enforce version 2 of their schema, and learned very unfortunately that they removed the metadata for environment, cmd, and entrypoint. However, they claim that version 1 will continue to be available, and so our functions are adopted to use the respective version for each.

Ok, I'm just going to assume that something is going to fail for these tests because there are so many changes, and I've been coding so long since last night through now that I think I'm going to throw up. I'll get this PR posted, and will definitely be making fixes and pushing toward merging this into development this weekend.  These changes are very important!

@gmkurtzer the addition of PULL for docker (combining create and import without sudo) is (I think) in yur court, because Python doesn't do anything special but be used. Let me know if you want/need any additional endpoints to make that easier!

@singularityware-admin


gmkurtzer:
Merged and pushed. Not sure why GitHub isn't seeing it automatically. Once you confirm, can you close this?

Thanks!

vsoch:
I think when the last commit is merged, if Github has its wheaties this morning, it will hopefully self close!

vsoch:
yeaaaah dawg!!

PR #539, opened 2017-03-27
Fix permissions docker

### LINKED ISSUE #538: Fix permissions docker
After untaring docker image, some files and folders are not writable. This is a problem in modified images, which contains more than 2 tar archives. The third archive then fails to extract due to permissions.

hra0031:
Fixes #538

 - Fix file permissions for custom docker images

@singularityware-admin


vsoch:
hey so with this current branch (if you look at my vsoch/development-tweak) we have actually made it so that all extraction is handled by the C code (and not python) and the permissions function was going to be removed entirely. However I had concern that we would run into this issue (the python extraction did quite a  bit of work to fix permissions) so I'll chat with @gmkurtzer later today to see if this is the same problem.

gmkurtzer:
Hi @hra0031, Do you have a reproducer that elicits a particular bug you are fixing?

Thanks!

vsoch:
I think this is the bug @gmkurtzer 

https://github.com/singularityware/singularity/issues/538#issuecomment-289534014

:bug:

vsoch:
Could it be an issue about the order of extraction?

vsoch:
hey @hra0031 we had a bug where we were extracting the layers in the wrong order (reverse!) could you try once more with the development branch? If not, could you  put the image on a public registry / or somewhere I could access to reproduce your case? Thanks much!

vsoch:
@gmkurtzer to give this some context - we no longer have the fix_permissions functions because it is assumed this is done in the C code. However, if you are unable to do this, then we need to consider moving this back to python, and somehow dealing with this wrt the content file, etc. Let us know how things look on the C end! This PR will be un-broken given that those functions are added back. If you get it working with C, then this PR will be fixed by removing the fix_permissions command. @hra0031  - if we do go this route, it will be desired to not have fix_permissions happen on the level of run_command (because as  you can imagine, there are many commands for which we wouldn't need/want to run it) but on the level of the import/pull etc. function that would be extracting the layers. Right now, since the python isn't extracting anything, adding it anywhere would not do anything.

hra0031:
Thank you for all information. We need this functionality currently, so we stay at current development build for now. Can you point me, where in the C code is the extraction done? Maybe we can write some temporary fix for our purposes.
This PR can be therefore closed.

PS: Sorry for the 3fbae8d commit, it was a mistake...

vsoch:
Let's keep it open until we are sure that your fix isn't needed! It's actually done in bash (I misspoke) and here is the link for you!

https://github.com/singularityware/singularity/blob/development/libexec/bootstrap-scripts/main-dockerhub.sh#L49

Do you have any ideas about how to fix permissions? Here is how we used to do it:

https://github.com/singularityware/singularity/blob/master/libexec/python/utils.py#L248

Basically looking at every file, and fixing those that were missing set permissions. I still think we might need to do something like this in python, which is why I want to keep this thread alive until @gmkurtzer has a chance to look (he's busy this week!)

hra0031:
@vsoch Why can not we do that like in version 2.2.1?
https://github.com/singularityware/singularity/blob/master/libexec/image-handler.sh#L231

vsoch:
We used to have two pathways for import - one supported sudo (the python extraction and changing permissions, and this was sudo import / bootstrap) and the other not (returning a layerfile and having the bash do it, not touching anything). I think the detail comes down to that change - we can't do it just as the user, but let's wait for @gmkurtzer to chime in. I'm hopeful that we can figure out a non sudo way to essentially do the same thing, and be consistent across functions to not need to maintain multiple "version" the same thing.

vsoch:
so to explicitly answer - I don't think we can do that because a user doing it on a file with higher up permissions would return denied.

hra0031:
But all files are extracted under the user, and user is the owner of this files. So ```chmod -R +w "$CONTAINER_DIR" ``` should work.

vsoch:
I haven't tried it! My thinking was that many docker layers come with root owned files, so a user on a system extracting wouldn't be able to do that. I'm going for a run quickly and then have a meeting before work - if you have some moments do you want to give it a try? We would want that fix to pass all tests for sudo/non sudo, and resolve your issue! If not I can try later today. It would be great if the fix were that easy!

hra0031:
Bingo! It's working...
```bash
...
Exploding layer: sha256:987a379c753bed9e668c9d572ea981813db45a19f8495dcb3e95c2135d8271b5.tar.gz
Fixing permissions...
Exploding layer: sha256:17385548ba54dd00b2b73dc106749019fcc0cc5641317c97766f14647d06ad6e.tar.gz
Fixing permissions...
Exploding layer: sha256:160bc4292c4667f07a3e3ae30d0ac7ccc02c122e6813172b9149e96a658b4b0a.tar.gz
Fixing permissions...
VERBOSE [U=4573,P=2235]    message_init()                            Set messagelevel to: 5
...
```
In my case the extraction of archive is here: https://github.com/singularityware/singularity/blob/development/libexec/image-handler.sh#L83

So i made a little change:
```diff
--- a/libexec/image-handler.sh
+++ b/libexec/image-handler.sh
@@ -84,9 +84,11 @@ case "$SINGULARITY_IMAGE" in
             name=`basename "$i"`
             message 1 "Exploding layer: $name\n"
             zcat "$i" | (cd "$SINGULARITY_ROOTFS"; tar --exclude=dev/* -xf -) || exit $?
+            message 1 "Fixing permissions...\n"
+            chmod -R +w "$SINGULARITY_ROOTFS"
         done
```
I've tried it on a machine, where I haven't root privileges.

vsoch:
wicked!! Let me do a separate PR to test this out against all the different tests.

gmkurtzer:
Hiya guys... This is resolved at this point now via a different PR, correct?

Thanks!

vsoch:
Yes this should have been closed, my bad!

PR #543, opened 2017-03-28
Development: Fixing name conflict, and adding size function

### LINKED ISSUE #542: Python namespace collision on Cray XC40 SMW for `make dist`
So, `make dist` will run the script `configure_transform.py` script, which does this:

```
sys.path.append('../libexec/python')

from utils import (
    get_fullpath,
    read_file,
    write_file
)
```
On a cray XC40 SMW, `utils` aready exists:
```
>>> import utils
>>> help(utils)

NAME
    utils

FILE
    /opt/cray/install-support/3.3.2568-2.8/utils/__init__.py
```

Obviously, this results in an import error:

```
Traceback (most recent call last):
  File "./configure_transform.py", line 8, in <module>
    from utils import (
ImportError: cannot import name get_fullpath
```

We developed a temporary workaround by adding `sys.path.remove('/opt/cray/install-support/3.3.2568-2.8'` but this is obviously not a good long-term solution; path.insert() may work as well, to put `../libexec/python` at the beginning of the path.

vsoch:
Fixes #542 

 - renaming the utils to sutils to deal with a naming conflict on some crusty version of cray
 - adding size function to write to the CONTENTFILE the size of a singularity hub or docker image, by way of using the Docker version 2 manifest, or singularity hub manifest. @gmkurtzer - we can use this for pull! We must have a pull, lol.
 - added tests for size, each for docker and shub
 - adding a "quiet" command to several functions that might be called once during an import/pull etc, so the debug is not printed twice.

@singularityware-admin


vsoch:
@gmkurtzer ready for review!

vsoch:
I just added to this PR a fix for an observation that a single variable was missing in the Docker environment - the reason was because we were using the LAST of the layers extracted from version 1 of the manifest, and we should have used the first in the list (the top layer). I've simply added a break to the loop to stop once it's found. If there is some issue with previous layers having environmental variables that should be included (and for some reason is not in the top) we can further parse and use a dictionary as a lookup, but I think this should be sufficient for now.

vsoch:
@gmkurtzer this is (still) ready :)

PR #547, opened 2017-03-30
Check/permissions fix!

vsoch:
Fixes #539 

This is to test the fix by @hra0031 for docker permissions, also includes the other tweaks to development to make sure nothing breaks. We would want to see if this works for non sudo functions.

 - adding chmod +x after image extraction in handler.

@singularityware-admin


libexec/image-handler.sh
```diff
@@ -80,10 +80,12 @@ case "$SINGULARITY_IMAGE" in
 
         eval_abort "$SINGULARITY_libexecdir/singularity/python/import.py"
 
+         
         for i in `cat "$SINGULARITY_CONTENTS"`; do
             name=`basename "$i"`
             message 1 "Exploding layer: $name\n"
             zcat "$i" | (cd "$SINGULARITY_ROOTFS"; tar --exclude=dev/* -xf -) || exit $?
+            chmod -R +w "$SINGULARITY_ROOTFS"
```
vsoch: @hra0031 let me know if this looks akin to your change - I removed the extra message to make it a little less verbose.

vsoch:
sweeet! And here is the command that wasn't working before:

```bash
singularity shell docker://blacklabelops/centos 
Cache folder set to /home/vanessa/.singularity/docker
Exploding layer: sha256:32acb9e864dfc5f21a59231e8f87b239a2584a4b354f9f6fd35d55def4384055.tar.gz
Exploding layer: sha256:45a2e645736c4c66ef34acce2407ded21f7a9b231199d3b92d6c9776df264729.tar.gz
Exploding layer: sha256:965814b5d3438f511d6e9fcfeb2ec09dd720053d9925946cac9fe92a516cb2ec.tar.gz

Singularity centos:~/Desktop> 

```

are there others we can test?

libexec/image-handler.sh
hra0031: I've added the message for a check if this part of the code is actually running.

vsoch:
hey @hra0031 I chat with @gmkurtzer and we discussed the issue with the above is that it is applying the command globally to all of the image,and we really just want to target files in the tar. It can't all be done in python because it needs to be done with a certain level of permission by the caller. But I think we can get this to work, because I think I can actually do the change of permissions to an in memory tar first in python, and then write and hand the fixed files equivalently to the calling function! This is actually sort of how we are doing the metadata folder.

I'm going to bed soon and traveling a big chunk of tomorrow (yuckers!) but I'll be on this definitely by tomorrow and we will get the fix in soon. Thanks again for helping with this!

hra0031:
I absolutely agree that this is not a solution. In fact, we need only to change permission on folders, which is contained in more archives. After extraction of all archives can be the permissions restored. But in another hand, this could be a problem for some provisioning scripts too (limited access to some folders).

vsoch:
hey @gmkurtzer and @hra0031 please take a look! I removed the blind `chmod` and instead wrote a little modified check_tar_permissions function that will go through the members, and if it's a folder, file, or (not) symlink, check the permission and change to the one we want given it's not set, or below, eg:

```python
def check_tar_permissions(tar_file,permission=None):
    '''check_tar_permissions changes a permission if 
    any member in a tarfile file does not have it
    :param file_path the path to the file
    :param permission: the stat permission to use
    '''
    tar = tarfile.open(tar_file, "r:gz")

    # Add all content objects to file
    fd, tmp_tar = tempfile.mkstemp(prefix=("%s.tmp." % tar_file))
    os.close(fd)
    fixed_tar = tarfile.open(tmp_tar, "w:gz")

    if permission == None:
        permission = stat.S_IWUSR

    # Check permissions for folders, not symlinks
    for member in tar:  
        if member.isdir() or member.isfile() and not member.issym():
            if not has_permission(member,permission):
                logger.debug("Fixing permission for %s" %(member.name))
                member.mode = permission | member.mode
            extracted = tar.extractfile(member)        
            fixed_tar.addfile(member, extracted)
        else:    
            fixed_tar.addfile(member)

    fixed_tar.close()
    tar.close()
 
    # Rename the fixed tar to be the old name
    os.rename(tmp_tar, tar_file)
    return tar_file        
```

The "fixed tar" is originally a tmp file so we can be sure we don't write and replace until it's done being written. This should hopefully avoid the situation of an operation bugging out / being canceled, as the temporary fixed tar will not be found and assumed to be the finished real one. The one potential bug could come in if the user breaks an operation between the finish of a download, but before the permissions have been checked, here:

```python
        targz = "%s/%s.tar.gz" %(cache_base,image_id)
        if not os.path.exists(targz):
            targz = client.get_layer(image_id=image_id,
                                     download_folder=cache_base)
            targz = check_tar_permissions(targz)
        layers.append(targz)
```

If the targz doesn't exist, then we are ok, it is downloaded and then checked. The issue would be that it gets downloaded, NOT checked, and then found and not checked. Ack, there could be some weird issues in here, let me see if I can fix this.

gmkurtzer:
This looks like a good approach, well done!

vsoch:
woot! Thanks @gmkurtzer !

hra0031:
@vsoch If it is downloaded and not checked, the sha256 will be the same as is the name of the file. If it is checked, the sha256 will differ. Or we can run check_tar_permissions() every time. Both will decrease performance, but controlling sha256 of archives can be good for security.

PR #550, opened 2017-04-01
adding start of checking for python modules

### LINKED ISSUE #549: configure.ac should check for python library dependencies (modules)

vsoch:
Fixes #549 

This is a test PR to see if my configure.ac changes function as they should to check for python module dependencies. If this works, then next I'll want to figure out how to do a check for version specific packages (eg, urllib, urllib2, etc)

@singularityware-admin


gmkurtzer:
Looks great, thanks!

PR #575, opened 2017-04-09
Adding permission o+w to all in memory tars, and verbose progress bar

vsoch:
This PR will modify the in memory tar extraction to add permission "write" to the owner (o+w) - the function is renamed to `change_tar_permissions` from `check_tar_permissions` because we are no longer checking anything. I also went through and removed unnecessary imports, and added a progress bar for the user given a verbose or higher level. A demo is here: 

https://asciinema.org/a/112081?speed=3

Note that adding the permission doesn't seem to make the images just work seamlessly, for example for the tensorflow image shown in the asciinema above, I cannot run it, because `/run/user` doesn't have execute, or something like that:

```bash
vanessa@som-ctp:/tmp$ singularity shell tensorflow.img 

Singularity tensorflow.img:/tmp>
```
Should that work? Does the changing permission step need to check if the file has o+x permission, and not change those first?

@singularityware-admin

### DIFF
--- a/libexec/python/base.py
+++ b/libexec/python/base.py
@@ -93,7 +93,7 @@ class ApiConnection(object):
         if 'Content-Length' in response.headers and response.code not in [400,401]:
             progress = 0
             content_size = int(response.headers['Content-Length'])
-            bot.show_progress(progress,content_size,length=50)
+            bot.show_progress(progress,content_size,length=50,suffix=bot.coffee)
 
         chunk_size = 1 << 20
         with open(file_name, 'wb') as filey:
@@ -102,14 +102,15 @@ class ApiConnection(object):
                 if not chunk: 
                     break
                 try:
+
                     filey.write(chunk)
                     if content_size is not None:
                         progress+=chunk_size
                         bot.show_progress(iteration=progress,
                                           total=content_size,
                                           length=50,
-                                          min_level=0,
-                                          carriage_return=False)
+                                          carriage_return=False,
+                                          suffix=bot.coffee)
                 except Exception as error:
                     bot.error("Error writing to %s: %s exiting" %(file_name,error))
                     sys.exit(1)
--- a/libexec/python/base.py
+++ b/libexec/python/base.py
@@ -93,7 +93,7 @@ class ApiConnection(object):
         if 'Content-Length' in response.headers and response.code not in [400,401]:
             progress = 0
             content_size = int(response.headers['Content-Length'])
-            bot.show_progress(progress,content_size,length=50,suffix=bot.coffee)
+            bot.show_progress(progress,content_size,length=40,suffix="downloading layer")
 
         chunk_size = 1 << 20
         with open(file_name, 'wb') as filey:
@@ -108,9 +108,9 @@ class ApiConnection(object):
                         progress+=chunk_size
                         bot.show_progress(iteration=progress,
                                           total=content_size,
-                                          length=50,
+                                          length=40,
                                           carriage_return=False,
-                                          suffix=bot.coffee)
+                                          suffix="downloading layer")
                 except Exception as error:
                     bot.error("Error writing to %s: %s exiting" %(file_name,error))
                     sys.exit(1)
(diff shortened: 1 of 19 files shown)


gmkurtzer:
I love the progress bar so much, can we make that the normal output mode (e.g. not `--verbose`)? Of course `-q` (`SINGULARITY_MESSAGELEVEL=0`) should hide it, but that is totally awesome!

vsoch:
Updates to the PR:

- the progress bar now lives with the SingularityMessage class, as this seemed the most appropriate place for it
- untarring of layers from dockers changes all files (not symlinks) to o+rw and folders (not symlinks) to o+rwx. Just to make sure I did this ok:

```python
    # Owner read, write (o+rw)
    if file_permission == None:
        file_permission = stat.S_IRUSR | stat.S_IWUSR

    # Owner read, write execute (o+rwx)
    if folder_permission == None:
        folder_permission = stat.S_IRUSR | stat.S_IWUSR | stat.S_IXUSR  
```

- The progress bar has been added to downloading files, with a default verbosity to print for info (and fixing permissions still only prints for verbose). I would suggest that we print one for both, however we would need to change the statement about fixing permissions to also be info. No, we can't combine them because they are two distinct loops/processes.
- I estimate the size of the download from the Content-Length header of the response. and then add up chunks as I extract, updating the progress bar with the chunk sum / total content length. Given that the content length includes some extra (and it's hard to consistently estimate this) I made a small tweak to the progress bar to cap at 100%. This shouldn't be terribly misleading (thinks of Windows...) because in testing it never went over much more than 100.8 or 101.3ish. It's a trivial rounding that the user shouldn't notice, and it shouldn't mislead.

vsoch:
one more thing I want to test - I use the stdout to test the encoding, and I want to see if it can be different from stderr!

gmkurtzer:
Oooo, I love the progress bar for downloads! I see there are two bars, Is the first the download, and the second the tarfile modification?

```
Downloading layer sha256:6d9ef359eaaa311860550b478790123c4b22a2eaede8f8f46691b0b4433c08cf
Progress |██████████████████████████████████████████████████| 100.0% \u26FE  
Progress |██████████████████████████████████████████████████| 100.0% \u26FE  
```

Thanks!

vsoch:
The first progress bar should be the layers, and the second the permissions. Are you running that on your local computer? It should use utf-8 and print the characters correctly, but it looks like that isn't the case. The message to print layers should also be showing up, what mode are you running it in?

vsoch:
If there is some use case when the second message doesn't show (eg quiet) and we still want the progress bar (which it shouldn't show, actually) then we could change the icons to be something like:

```
Downloading layer sha256:6d9ef359eaaa311860550b478790123c4b22a2eaede8f8f46691b0b4433c08cf
Progress |█████████████████████████████████████████| 100.0% download layers  
Progress |█████████████████████████████████████████| 100.0% fixing permissions  
```

What do you think?

vsoch:
ok, we now have the command written next to the progress bar:

```bash
Downloading layer sha256:6d9ef359eaaa311860550b478790123c4b22a2eaede8f8f46691b0b4433c08cf
Progress |████████████████████████████████████████| 100.0% downloading layer  
VERBOSE Fixing permission for /home/vanessa/.singularity/docker/sha256:6d9ef359eaaa311860550b478790123c4b22a2eaede8f8f46691b0b4433c08cf.tar.gz.5xtl99kh
Progress |████████████████████████████████████████| 100.0% changing permissions  
```

You know, we don't really need the information prints before if we have the bar, what if it were just like:

```bash
Progress |████████████████████████████████████████| 100.0% downloading layer  
Progress |████████████████████████████████████████| 100.0% changing permissions  
```

vsoch:
ok, for shub download looks like:

```bash
singularity pull shub://vsoch/singularity-images
Progress |████████████████████████████████████████| 100.0% downloading image  
Done. Container is at: ./vsoch-singularity-images-mongo.img
```

chrisgorgo:
I just realized that[ `o` in `o+w` stands for "others" not "owner"](https://ss64.com/bash/chmod.html). `u+w` would be the owner. Not sure if this matter (I'm not even sure which issue is this PR solving, but I'm a bit out of the loop). I thought I would bring this up since it might be an indication of more confusion.

gmkurtzer:
@chrisfilo Haha, I make that same mistake all the time. Sorry if I did that too, I meant owner or rather "user".

vsoch:
PR has been updated with ascii characters for progress bar:

```bash
Downloading layer sha256:d54efb8db41d4ac23d29469940ec92da94c9a6c2d9e26ec060bebad1d1b0e48d
Progress |========================================>| 100.0% downloading layer
```

vsoch:
In the Python it should be  ok - we use [these permissions](https://docs.python.org/2/library/stat.html#stat.S_IRUSR) for owner and not user. @chrisfilo can you confirm that looks ok?

```python
    # Owner read, write (o+rw)
    if file_permission == None:
        file_permission = stat.S_IRUSR | stat.S_IWUSR

    # Owner read, write execute (o+rwx)
    if folder_permission == None:
        folder_permission = stat.S_IRUSR | stat.S_IWUSR | stat.S_IXUSR  
```

from https://github.com/vsoch/singularity/blob/development/libexec/python/sutils.py#L311

gmkurtzer:
Done and awesomeness! YAY!

PR #579, opened 2017-04-09
Allow hyphen in TERM environment variable

### LINKED ISSUE #577: Rejecting valid $TERM strings which contain a hyphen
It seems that the latest dev. Singularity is rejecting $TERM values which include a hyphen.

This is happening in the `development` branch.  I'm at commit 167857d5983aad17b84269666c8e48d255903d67.

Here's the verbose output from a bootstrap:

```
akkornel@blargh-yakkety-typical:~/workflow$ singularity -vv bootstrap image.img bootstrap.def
increasing verbosity level (3)
Exec'ing: /usr/local/libexec/singularity/cli/bootstrap.exec image.imgVERBOSE: Set messagelevel to: 3
VERBOSE: Checking input from environment: 'LANG'
VERBOSE: Obtained input from environment 'LANG' = 'en_US.UTF-8'
VERBOSE: Checking input from environment: 'TERM'
ERROR  : Illegal input character '-' in: 'TERM=xterm-256color'
ABORT  : Retval = 255
```

I don't think the details of the bootstrap script matters here, because it never gets to the point of parsing the bootstrap script.

Looking on a stock Ubuntu 16.10 system, I see 22 different terminal types which include a hyphen character:

```
akkornel@blargh-yakkety-typical:~/workflow$ ls /lib/terminfo/* | grep \-
cons25-debian
Eterm-color
mach-bold
mach-color
mach-gnu
mach-gnu-color
rxvt-basic
rxvt-m
rxvt-unicode
screen-256color
screen-256color-bce
screen-bce
screen-s
screen-w
xterm-256color
xterm-color
xterm-debian
xterm-mono
xterm-r5
xterm-r6
xterm-vt220
xterm-xfree86
```

In my case, the terminal type I'm using is `xterm-256color`.  This is the default that I got from an Ubuntu 16.10 Desktop installation.

akkornel:
Running `ls /lib/terminfo/* | grep \-` on a stock Ubuntu 16.10 Desktop system
shows 22 terminal types that use the hyphen character.  When using the Terminal
in the Unity desktop, the terminal type is set to `xterm-256color`.

A manual scan of the list shows that the hyphen is the only non-alphanum
character in the list of terminal types.

Fixes #577.

 - Add hyphen to the list of allowed non-alphanum characters in the `TERM` environment variable.

@singularityware-admin

### DIFF
--- a/src/bootstrap.c
+++ b/src/bootstrap.c
@@ -47,7 +47,7 @@
 int main(int argc, char **argv) {
     struct image_object image;
     char *lang = envar_get("LANG", "_-=+:,.%", 128);
-    char *term = envar_get("TERM", NULL, 128);
+    char *term = envar_get("TERM", "-", 128);
 
     singularity_config_init(joinpath(SYSCONFDIR, "/singularity/singularity.conf"));
     singularity_registry_init();


akkornel:
Re-wrapped commit message.

gmkurtzer:
Good catch!

PR #580, opened 2017-04-10
Add GPG support to yum bootstrap method

akkornel:
This enhances the yum bootstrap mechanism by allowing the author to specify a GPG key to be used by the initial run of `yum` (or `dnf`).

The bootstrap author uses the "GPG" field to place either a path or a URL to the GPG key.  The GPG key can either be specified as a path, or as a URL.  If an HTTPS URL is used, the bootstrap will check for curl (which RPM uses), and will fail if curl is not found.  If an HTTP URL is used, the bootstrap will fail unconditionally.  Only one key may be specified.

The field's value (after checking for HTTP URLs) is passed directly to `rpm --import`.  It is not placed in `/etc/pki/rpm-gpg`; that is left to the appropriate `*-release` package (such as `centos-release`).

Changes proposed in this pull request:

- In the temporary yum config file, use `gpgcheck=1` (instead of `gpgcheck=0`) if the `GPG` field is set.
- If the `GPG` field is set, and the field's value starts with `http://`, fail the bootstrap.
- If the `GPG` field is set, and the field's value starts with `https://`, check if the `curl` command is installed.  If not, fail the bootstrap.
- If the `GPG` field is set, and neither of the two prior conditions are met, run `rpm` to import a GPG key from [insert value of the `GPG` field here].

@singularityware-admin


vsoch:
You're a bad@ss @akkornel, these contributions and your notes are great! 📝

gmkurtzer:
This looks fantastic, thanks @akkornel!

PR #587, opened 2017-04-11
Fix fopen loop dev

flx42:
No need to use fopen(3) on /dev/loopN, especially if you're always using `fileno` afterwards.


flx42:
Sorry, wrong branch!

gmkurtzer:
Haha, I was just noticing that! Your patch looks good though!

PR #588, opened 2017-04-11
Use open(2) instead of fopen(3) for /dev/loop devices

flx42:
No need to use fopen(3) on /dev/loopN, especially if you're always using fileno afterwards.


flx42:
@gmkurtzer now on the right branch ;)

PR #592, opened 2017-04-12
Revert "Tweaking output of progress bar"

vsoch:
Reverts singularityware/singularity#591

There is some strange bug now - it's printing on every newline! I'm reverting back to the working version, in case anyone tries to use development.

### DIFF
--- a/libexec/python/base.py
+++ b/libexec/python/base.py
@@ -66,8 +66,7 @@ class ApiConnection(object):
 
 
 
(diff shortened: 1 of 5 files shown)


vsoch:
ok I just tested this again - @gmkurtzer it won't hurt to revert to a version that was working - I usually check with you first on this, but this is a special case of un-doing a PR that was tiny and seems to have broken output that others might clone and be in horrible shock at, lol.

gmkurtzer:
Understood, thanks for the explanation!

PR #595, opened 2017-04-12
Remove the need to generate environment.tar as root

flx42:
The environment.tar archive is now generated as part of the build
process, and is not checked in the repository any more.

Signed-off-by: Felix Abecassis <fabecassis@nvidia.com>


flx42:
My autotools skills are rusty, there might be a better way to do that than  a `PHONY` rule

gmkurtzer:
Merged, thanks!

vsoch:
Oh thank goodness, thank you @flx42 ! This gave me such pain when I didn't know it existed, and then found it by chance, haha.

PR #597, opened 2017-04-13
Fix misleading typo regarding uid/gid remapping

flx42:
The log didn't match the values written to the uid_map/gid_map files


gmkurtzer:
Oh yeah, good catch! Thanks!

PR #602, opened 2017-04-13
updating copy help

vsoch:
Fixes #

 - addition of instructions to user to use -p, in both main and copy help
 - removing that copy require sudo (it doesn't)

@singularityware-admin


gmkurtzer:
Wahoo, thanks!

PR #603, opened 2017-04-14
Fix out-of-bounds writes in singularity_registry_{get,set}

flx42:


flx42:
valgrind was reporting thousands of errors.

gmkurtzer:
Great catches, thanks!

PR #605, opened 2017-04-16
Progress bar - let's do this again!

vsoch:
The issue with the progress bar was really stupid, and thanks to @satra for pointing me to a quick test that made me aware of it. The print output was simply too wide for the terminal. To fix, all I needed to do was make it a lot shorter. The preview of the fix is here:

https://asciinema.org/a/113644?speed=3

@singularityware-admin

### DIFF
--- a/libexec/python/base.py
+++ b/libexec/python/base.py
@@ -2,6 +2,8 @@
 
 python: base template for making a connection to an API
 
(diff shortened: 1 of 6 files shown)


gmkurtzer:
Looks great!

PR #606, opened 2017-04-16
Update debian rules

dexter2cz:
- location of binaries has changed
- sexec splits to more binaries - set suid bit to all with -suid


gmkurtzer:
Looks good, thanks!

PR #609, opened 2017-04-16
Quote string components before doing bash eval

### LINKED ISSUE #607: bash syntax error at end of yum bootstrap
Yo!

This happened during a yum bootstrap:

```
/usr/local/libexec/singularity/bootstrap-scripts/post.sh: eval: line 57: syntax error near unexpected token `('
/usr/local/libexec/singularity/bootstrap-scripts/post.sh: eval: line 57: `/usr/local/libexec/singularity/python/helpers/json/add.py -f --key SINGULARITY_BOOTSTRAP_VERSION --value 2.2.99 (development/388c86b) --file /var/lib/singularity/mnt/source/.singularity.d/labels.json'
```

Here's the basic bootstrap file I'm using: https://gist.github.com/akkornel/b9cd992af449f0573285f1187aabfd9d

Here's the problem appearing, right at the end of the bootstrap: https://asciinema.org/a/113783

My Git working copy is current to 388c86b3f612c64845aa479877f495c8fc183ea3.

akkornel:
When doing a bash `eval`, we need to escape any variables that come from
the "outside".

Thanks to http://stackoverflow.com/questions/2854655 for the solution!

Fixes #607 (though it'll need to be closed manually, because this isn't pushing to the default branch.)

 - Shell-quote SINGULARITY_LABELFILE, the path to `add.py` (SINGULARITY_ADD_SCRIPT), and all other values that could come in from the outside.

### DIFF
--- a/libexec/bootstrap-scripts/post.sh
+++ b/libexec/bootstrap-scripts/post.sh
@@ -27,7 +27,7 @@ fi
 
 ## Load functions
 if [ -f "$SINGULARITY_libexecdir/singularity/functions" ]; then
-    . "$SINGULARITY_libexecdir/singularity/functions"
+    . $SINGULARITY_libexecdir/singularity/functions
 else
     echo "Error loading functions: $SINGULARITY_libexecdir/singularity/functions"
     exit 1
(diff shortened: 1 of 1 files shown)


gmkurtzer:
Oh wow, that is a neat solution. Thanks!

PR #612, opened 2017-04-17
Allow the '+' character in pathnames.

hollowec:
We have several pathnames with '+' characters in our environment.
Singularity doesn't allow these paths to passed as bindmounts because
of the presence of this character.  Is there a particular reason why '+' 
characters aren't permitted?  In a cursory look at the code, I wasn't able 
to find any obvious security issues with permitting them.

@singularityware-admin

### DIFF
--- a/src/util/util.c
+++ b/src/util/util.c
@@ -102,7 +102,7 @@ int envar_defined(char *name) {
 
 char *envar_path(char *name) {
     singularity_message(DEBUG, "Checking environment variable is valid path: '%s'\n", name);
-    return(envar(name, "/._-=,:", PATH_MAX));
+    return(envar(name, "/._-+=,:", PATH_MAX));
 }
 
 


gmkurtzer:
Hi @hollowec, can you resubmit this against the 'development' branch? Thanks!

vsoch:
Is it an issue that `+` can have special context? For example, [arithmetic](http://tldp.org/LDP/abs/html/ops.html#AROPS1) or as a [regular expression operator](http://tldp.org/LDP/abs/html/x17129.html#PLUSREF) or as an option:

```
+
Option. Option flag for a command or filter.

Certain commands and builtins use the + to enable certain options and the - to disable them. In parameter substitution, the + prefixes an alternate value that a variable expands to.
```

[Reference](http://tldp.org/LDP/abs/html/special-chars.html)

gmkurtzer:
@vsoch It could, but currently doesn't.

hollowec:
Resubmitted to development.  Should I close this pull request (to master)?

Hi @vsoch - the '-' character is permitted already, however, and it can have special meaning in option parsing, arithmetic, and some regular expressions.

vsoch:
If `-` could be a security issue in some way we could anticipate, it should not be permitted for bind points either. @gmkurtzer what do you think? I don't have the hackerdome expertise to know.

gmkurtzer:
Yep, please close this one. Thanks @hollowec!

@vsoch I don't think there is a security concern with the use of the `-` or `+` characters.

hollowec:
OK, closing.  Thanks.

PR #614, opened 2017-04-17
Allow the '+' character in pathnames.

hollowec:
We have several pathnames with '+' characters in our environment.
Singularity doesn't allow these paths to passed as bindmounts because
of the presence of this character. Is there a particular reason why '+'
characters aren't permitted? In a cursory look at the code, I wasn't able
to find any obvious security issues with permitting them.

@singularityware-admin


gmkurtzer:
Thanks!

PR #626, opened 2017-04-21
chdir to image directory to keep image accessed, prevent unmount

DrDaveD:
Change proposed in this pull request:

 - When the image is a directory, chdir to that directory in the outer sexec process.  This is to fix a problem at least on some kernels (e.g. the one in Centos7.3), where there's something about singularity's use of containers that prevents the kernel from thinking that the image is still in use if this chdir is not done. That is, the kernel does not prevent the filesystem that the image is in from being unmounted.  This was first noticed in a case where the filesystem was automounted, but it is also easily demonstrated with a manual umount command.  Keeping a file descriptor open to the top level directory would probably work just as well, but this seems simpler.

@singularityware-admin


gmkurtzer:
Can you tell me more about the issue itself? Do you have a reproducer that I can test on a RHEL7.3 kernel?

BTW, The PR will need to be rebased to the `development` branch to be merged.

Thanks!

DrDaveD:
Ok on the development branch, I didn't know about that.  I will close this PR and make a new one, not only to rebase but also because it turns out that doing chdir fails when the image directory is a relative path.  I'm changing it to do open instead of chdir.

It's very easy to reproduce.  I'll include simple steps in the next pull request.

gmkurtzer:
Awesome, thank you!

PR #629, opened 2017-04-23
Fix commit 0650c9d

ickc:
See [singularity/LICENSE.md at master · ickc/singularity](https://github.com/ickc/singularity/blob/master/LICENSE.md) for demo on GitHub's recognition of the LICENSE.


gmkurtzer:
Thank you for suggesting this, but the project license is not to be changed without legal approvals.

ickc:
I don't understand, because I actually didn't change the license. It is the same as before.

But IANAL. The purpose is really just to achieve what commit 0650c9d wanted to achieve. You might try to replicate what I did in your liking so that GitHub will recognize it.

gmkurtzer:
Hi @ickc, it is not the same license. At an initial glance, clause #3 has been modified, and the locations of the files are in different places. This structure and the file contents was dictated to me by legals of DOE, UCOP, and LBNL so they can not be modified at this time. I already have plans to modify the license and contributors clause but I am waiting for approvals before I can proceed. When that happens, I will re-review this PR and try to encompass as much as possible.

PR #654, opened 2017-05-01
Add/plugins: environment variable driven custom functions for Singularity

vsoch:
This is a first go at a (simple) implementation for custom functionality to be triggered by an environment variable. In this case, we have the following issue. For most of our bootstraps, a recursive function to change permissions on the tarballs is sufficient to produce a working container. This process is much faster than our old approach (by 2-6 or 7 times depending on the image), and the ideal "default" setting. However, we have niche use cases, namely containers like `docker://blacklabelops/centos` where the default doing this, [as we've seen before](https://github.com/singularityware/singularity/issues/538#issuecomment-289534014)

```
singularity shell docker://blacklabelops/centos
```
isn't going to work. However we should not add extreme levels of custom commands on funky tarballs that require slowing down the process by 2-7 fold for all users just for these niche use cases. 

The workaround is to empower the user to ask for this permissions fix, and the slowdown, and we can do this by way of what I am calling a "plugin" or a custom function/filter to run given the preference for it. With our current base, this makes sense to specify via environment variables, and @gmkurtzer we could have some kind of command line flag, or in the case of many plugins, an additional `%plugins` or `%custom` section in the bootstrap (at some point in the future.) For example, here is how the current PR works:

```
# Runs normally, changing permissions after Python at the end  (fast)
singularity shell docker://ubuntu

# Changes permissions in memory using python
SINGULARITY_FIX_PERMS=True
export SINGULARITY_FIX_PERMS
singularity shell docker://ubuntu
```

To see how it works, I've also prepared a [DEMO](https://asciinema.org/a/118507?speed=3).

In the future, you could imagine something like:

```
Bootstrap: docker
From: blacklabelops/centos

%custom
plugin fix_perms
```

or

```
Bootstrap: docker
From: blacklabelops/centos

%plugins
fix_perms var1
add_webserver port
```
etc.etc.

@singularityware-admin


gmkurtzer:
Thanks!

PR #655, opened 2017-05-02
Making the world a better place

vsoch:
I'll just leave this here.

https://asciinema.org/a/118543?speed=3

@singularityware-admin

### DIFF
--- a/libexec/python/README.md
+++ b/libexec/python/README.md
@@ -145,6 +145,13 @@ singularity pull shub://vsoch/singularity-images
 Done. Container is at: ./vsoch-singularity-images-mongo.img
 ```
 
+### Formatting
+Formatting refers to the user interface for the command line tool.
+
+**SINGULARITY_COLORIZE**
+By default, debug messages (and other types) use ascii escape sequences to various commands with colors. This helps to distinguish them, and makes the user interface a bit more pleasant. If the output is not going to a terminal, or if the terminal does not support the ascii escape sequences, this variable is set to False. The user can always override this by setting this variable.
+
+
 ### Plugins
 Singularity plugins are custom environment variables that can be set to turn bootstrap (and other building) customizations on and off. Currently, we just have one plugin that will, when turned on, have the Python API backend change permissions for the tarballs.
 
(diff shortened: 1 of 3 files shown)


gmkurtzer:
It's very colorful! Can you set `TERM=xterm` and `TERM=vt100`? I'm just wondering what happens in legacy modes.

vsoch:
[Seems to be ok](https://asciinema.org/a/dqrl3xt07ypa0bkllh1jxezdb?speed=3) unless I did the change of term incorrectly? (I exported the `TERM` variable)

vsoch:
@gmkurtzer is there a way to test this? We can also easily disable all colors if the term is in some list, or not in some list. It really looks nice, and makes it fun and easier to distinguish the different output, I hope we can have colors :)

vsoch:
I can tell by your quick response to this PR that you are super excited about colors in the terminal @gmkurtzer ! To make it more flexible and error tolerant, I've added an optional environment variable that will completely disable color output. Eg:

```
SINGULARITY_COLORIZE=false
export SINGULARITY_COLORIZE
```
I also do a check to see if the output is not in a terminal (eg going to pipe) and it doesn't use color in those cases. I tried exporting the different TERM environment variables as you specified, but it didn't seem to cause issue with the colors. This could be that I am not able to reproduce a terminal that doesn't support ansi escape. From Googling, it seems like the biggest contender is Windows-based terminals, and we could do a check for the system type if we ever add Windows support. Which I hope is just never. I didn't say that :)

vsoch:
I've updated the PR to do the following:

- info statements are not printed with any color, even
- the progress bar is not printed with any color, ever
- this means by default, we can have colors enabled for logger prefixes for a subset of levels (but not all)

For a demo of this during runtime, [see here](https://asciinema.org/a/118922?speed=3)
For a demo of the different levels and all colors, [see here](https://asciinema.org/a/118923?speed=3)

vsoch:
:rainbow: :earth_asia: :smile:

gmkurtzer:
Wahoo, thanks!

PR #671, opened 2017-05-16
Fixing test for size, and there was no bug with the uri

vsoch:
This is essentially the same PR as before #668 that was never merged, with just a tweak to some print output for the image uri. The reason it wasn't working was because the sha digest was incorrect.

 - bla bla bla
 - nobody is reading this
 - sad dinosaur :*(

@singularityware-admin


gmkurtzer:
Read and merged!

vsoch:
happy dinosaur :O)

PR #672, opened 2017-05-16
Adds an option to override tmp directory

hcraT:
As proposed in #642 adds the -T (--temp) option. This allows to override the
/tmp directory similar to the -H (--home) option.

### DIFF
--- a/etc/bash_completion.d/singularity
+++ b/etc/bash_completion.d/singularity
@@ -75,6 +75,8 @@ _singularity() {
                 _filedir
             elif [[ ${cur} == -H || ${cur} == --home ]]; then
                 COMPREPLY=( $(compgen -f -- ${cur}) )
+            elif [[ ${cur} == -T || ${cur} == --temp ]]; then
+                COMPREPLY=( $(compgen -f -- ${cur}) )
             elif [[ ${cur} == -W || ${cur} == --workdir || ${cur} == --wdir ]]; then
                 COMPREPLY=( $(compgen -f -- ${cur}) )
             elif [[ ${cur} == -* ]] ; then
--- a/etc/singularity.conf.in
+++ b/etc/singularity.conf.in
@@ -113,8 +113,9 @@
 # DEFAULT: @MOUNT_TMP_DEFAULT@
 # Should we automatically bind mount /tmp and /var/tmp into the container? If
 # the --contain option is used, both tmp locations will be created in the
-# session directory or can be specified via the  SINGULARITY_WORKDIR
-# environment variable (or the --workingdir command line option).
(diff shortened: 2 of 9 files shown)


gmkurtzer:
Heya @hcraT,

I think this option is satisfied by the `--workdir` and a `--temp` option would be redundant. Also, to consider this, we would have to move this to the 'development' GitHub branch as opposed to master.

Thanks!

hcraT:
Hi,
you are right (of course :D ). However I have to say that I find the current options a bit confusing (both the naming and the semantics).

In the current status of the software in order to change the source tmp folder you need to set both contain and workdir. Moreover you need to:
- set the workdir option to the prefix of the path where your source /tmp folder is
- specify the -H option in order to prevent containment of the home directory (if all you need is to change the /tmp mountpoint)
- the names of both the source and destination mountpoint are hard-coded (you could of course use the -B option in that case)
- the contain option has important consequences (if all you want is to change the /tmp mountpoint)

My proposal would be:
- remove the workdir option (whose name is in my opinion misleading becasue in docker the semantics is different, that semantics would be a nice addition in singularity as well)
- make the --contain take an optional parameter (the path to be used as "workdir") if the option is present but no path is specified the workdir is set by default to the session-directory
- keep the -H option and add the -T option to the dev branch.
In case you agree with the proposal would in your opinion be useful to add a --temp_var (or --var_temp) option as well?

Thanks.

gmkurtzer:
Your logic makes sense, but I would prefer not to change so much. I think if I/we/you do this, the behavior would be in alignment with what you are hoping for:

If we move the order of operations so that user binds happen before `singularity_mount_tmp()` and check to see if `/tmp` and/or `/var/tmp` is already mounted via a user bind, and if so, skip it.

In that case, we can simply do what you originally requested:

```
$ singularity shell -B ~/tmp:/tmp container.img
```

Additionally, your point about option nomenclature is a good one, I propose to change `--workdir` to `--tmpdir` as that seems truer to intent (with of course maintaining legacy support).

hcraT:
Wouldn't make sense to apply the same reasoning to the home directory (I mean
to move the home binding after user bind and so remove the `-H` option)?

What about `--contain_dir` or `--contain_path` or `--contain_prefix` in this way the link between `--contain` and this option becomes obvious.

gmkurtzer:
So binding the home directory isn't quite as simple unless overlay is enabled. But the `-H` option is needed because it also updates the `/etc/passwd` and `/etc/group` files on demand so credentials and paths are taken care of properly.

hcraT:
Hi,
sorry for not being active with this.

Thanks for imlementing the option!

Closing this as is not needed anymore

PR #689, opened 2017-05-24
adding changes to inspect.py

vsoch:
This PR will add more robust parsing of the different environment, labels, and definition files using the new `--inspect` command. A few notes!

 - Instead of typing everything out, please [preview the functionality here](https://asciinema.org/a/122013?speed=3)
 - I wrote up [docs to explain the back end python API](https://github.com/vsoch/singularity/blob/development/libexec/python/README.md#inspect)
 - I added a new function to pretty print json, since it's now being used in several places.
 - The one thing I was ornery about was the fact that `SINGULARITY_ROOTFS` and `SINGULARITY_MOUNTPOINT` are essentially the same thing, but maintained as two separate variables. The python bits are expecting a base of `SINGULARITY_ROOTFS` so as a workaround I set that from `SINGULARITY_MOUNTPOINT` and everything flows as expected.
 - @gmkurtzer, when I was testing, I still had a docker ubuntu image producing without labels as I would expect (I made them manually for the example). I remember this wasn't working, and I'm not sure if it got lost in the nethers or still is not functioning.
 - It looks like there are some merge conflicts that need to be resolved. I did change around some current code, so this was expected. (update - I have resolved).

@singularityware-admin


gmkurtzer:
Looks good, thanks!

PR #694, opened 2017-05-26
Adding ability to read docker credentials from .docker/config.json

### LINKED ISSUE #693: Registry authentication with .docker/config
As far as I am aware it is currently not possible to authenticate with docker registry that use OAuth authentication, with only basic authentication being supported. Is this correct?

Is it possible to read authorisation tokens from .docker/config file ?

vsoch:
Fixes #693 

- This PR adds a function to the docker api client that can read in the config token from the user's config.json, which is located in `/.docker/config.json`, and generated when a user does docker login. I was first against adding this additional use case, but I think it's reasonable to minimally look for this as an option given that the authentication has failed, and the user has not provided a credential - for the subset of users that also use Docker, it will mean not having to export environment variables.
 - I also wrote a function to get the registry an actual function (since we use it many times), and with this more properly defined the `USERHOME` variable in defaults.py
 - Adding a test for this.

The config.json seems to store a base64 credential that is indexed by the registry version, and as far as I know, it works to use a version 1 credential for version 2 registry (this is how I tested). We should keep our eye on this in case there are situations when one credential doesn't work for the other. However, given that the user has gotten to the point, they would have just hit a sys.exit(1) anyway, so it doesn't hurt to try!

@singularityware-admin

### DIFF
--- a/libexec/python/defaults.py
+++ b/libexec/python/defaults.py
@@ -96,14 +96,14 @@ if COLORIZE is not None:
 # Cache
 #######################################################################
 
+USERHOME = pwd.getpwuid(os.getuid())[5]
 DISABLE_CACHE = convert2boolean(getenv("SINGULARITY_DISABLE_CACHE",
                                 default=False))
 
 if DISABLE_CACHE == True:
     SINGULARITY_CACHE = tempfile.mkdtemp()
 else:
-    userhome = pwd.getpwuid(os.getuid())[5]
-    _cache = os.path.join(userhome,".singularity") 
+    _cache = os.path.join(USERHOME,".singularity") 
     SINGULARITY_CACHE = getenv("SINGULARITY_CACHEDIR", default=_cache)
 
 if not os.path.exists(SINGULARITY_CACHE):
--- a/libexec/python/docker/api.py
+++ b/libexec/python/docker/api.py
@@ -190,25 +190,36 @@ class DockerApiConnection(ApiConnection):
         elif self.auth is not None:
             headers.update(self.auth)
 
-        # Third priority is trying to read from local docker config
-        else:
-            auth = self.basic_auth()
-            if auth is not None:
-                headers.update(auth)
+        try:
+            self.retrieve_token(headers)
+        except:
+            bot.error("Error getting token for repository %s/%s, exiting." %(self.namespace,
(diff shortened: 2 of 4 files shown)


rbrthogan:
It seems to me that a much more elegant solution would be to just rely on the existence of user credentials in `.docker/config.json`. Right now, for Basic authentication, you have a 2 stage process, with the first simply reproducing data that would be contained in `.docker/config.json` if it exists followed by using this data to get access, and separately have a fall back to read this file anyway. I think that if `.docker/config.json` exists it should be first thing used, otherwise perhaps you could look for env variables, construct the first token, etc. This way, the api will execute entirely the same code, independent of registry Auth, as long as the correct credentials are present.

I personally don't love the env variable solution because it requires me to stored my password as plain text which is irksome. Perhaps you don't want to rely on the existence of `.docker/config.json` because it is generated by the docker daemon. I do think, however, that most users who often use private docker registries (either with Basic Authentication or OAuth) likely use the docker daemon to handle credentials for them. In any case, the env variable solution can always be there as fall back if `.docker/config.json` doesn't exist.

I hope this isn't just my personal bias speaking, I would be interested to hear community feedback.

gmkurtzer:
Heya, is this ready for merge, or should we wait until after the 2.3 release?

vsoch:
see my comments at the end of this thread https://github.com/singularityware/singularity/issues/693 I don't think the docker config file should be first checked, because it's not/**can't** be on most environments where this command would be run. The only other option that I wanted to run by as an addition would be to allow the user to specify the token as an environment variable instead of the docker username and password. You could still decode it, but it might make users more comfortable not having the username/password easily in plain sight.

vsoch:
I'm going to close this for now - I don't think it would be of essential need, at least for now.

rbrthogan:
@vsoch that's a shame. Perhaps it's not essential for everyone but I think it's a pretty simple extension of the docker functionality that will be absolutely required from some enterprise users of Singularity and at least a nice to have for others.

vsoch:
I think it will be nice too! It's an easy enough bit of code that we can add it if /when the community pushes for that. Right now I think there are more pressing things to focus on, so I wanted to close the PR. Note that the issue is still not closed so we won't forget about it :)

PR #715, opened 2017-06-02
ENH: mailmap entries for Vanessa and Rémy

yarikoptic:
Just added their multiple "identities" for a nicer "git shortlog -sn" ;)

Attn: @singularityware-admin


gmkurtzer:
Thanks!

PR #739, opened 2017-06-09
Allow tracing program execution to create minimal containers

cclerget:
**Add trace command:**

Allow to trace command execution inside container and host to create minimal singularity container.
As discussed on singularity google group: https://groups.google.com/a/lbl.gov/forum/#!topic/singularity/l1k-CtXHiRM, it's a first attempt, so need some adjustments and discussion.

It implements two cases:
1. trace program execution inside container:
`singularity trace $HOME/ubuntu.img python -c 'print "hello"'`
2. trace program execution on the host directly:
`singularity trace img2create.sh python -c 'print "hello"'`

I can't include unit tests because of missing dependencies in test container and I don't know if travis host have the required packages (strace, perl)

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -23,3 +23,4 @@
     - Ralph Castain <rhc@open-mpi.org>
     - Yaroslav Halchenko <debian@onerussian.com>
     - Eduardo Arango <carlos.arango.gutierrez@correounivalle.edu.co>
+    - Cedric Clerget <cedric.clerget@univ-fcomte.fr>
--- a/bin/Makefile.am
+++ b/bin/Makefile.am
@@ -1,5 +1,5 @@
 
-dist_bin_SCRIPTS = singularity run-singularity
+dist_bin_SCRIPTS = singularity run-singularity trace2sing
 
 MAINTAINERCLEANFILES = Makefile.in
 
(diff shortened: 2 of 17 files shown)


gmkurtzer (changes requested):
There is a lot to review and test here, but one of the first things I've noticed is that the trace2sing script calls out to quite a few things. When it first runs, can you do a check to make sure that all dependencies are present. For example, it might be common for Perl or strace to not be installed in a container.

gmkurtzer:
BTW, I had written (in v1.x) a ptrace binary that will follow the execution of any program, script or anything, watching for files it touches, and then return those files in a list, which can then be used to assemble a container. I think that code would be useful here.

cclerget:
actually trace2sing check for presence of strace (and version too) and perl and exit gracefully with an error message saying to install required package.

The main reason I use strace rather than ptrace is to support more architecture without playing with stack and registers of each arch and let all the pain to strace, second reason is to not reinvent the wheel when strace does this perfectly

gmkurtzer:
Yes, you nailed one of the big issues I also found for dealing with ptrace as we were porting Singularity to different architectures. But one of the aspects that required it was that the output of strace was variable depending on version and parsing accordingly was problematic.

cclerget:
Yes, output could change. I have tested it with latest version 4.17 and worked fine. Centos 7 use strace 4.8, latest ubuntu use 4.11.
Regular expressions are flexible to adapt output parsing between versions. Finally output remains the same: syscall(arg1,...) = ret since 2009.
I don't worry too much about output changes, and this couldn't be a problem in a near future ;-)

PR #774, opened 2017-06-27
Import Complete set of NVidia libraries with --nv

mrmhodak:
Currently, several NVidia CUDA libraries are omitted with --nv flag. This imports all libraries from cuda-8.0 and, among other things, allows running tensorflow_gpu without the need for installing cuda inside the container. Also removing redundant entry "libnvcu" , already caught by "libnv".

Attn: @singularityware-admin

### DIFF
--- a/libexec/cli/action_argparser.sh
+++ b/libexec/cli/action_argparser.sh
@@ -116,7 +116,8 @@ while true; do
         ;;
         -n|--nv)
             shift
-            for i in `ldconfig -p | grep -E "/libnv|/libcuda|/libEGL|/libGL|/libnvcu|/libvdpau|/libOpenCL|/libOpenGL"`; do
+             for i in `ldconfig -p | grep -E "/libnv|/libcuda|/libEGL|/libGL|/libvdpau|/libOpenCL|/libOpenGL|\
+                /libcusparse|/libcusolver|/libcurand|/libcuinj|/libcufft|/libcudnn|/libcublas|/libnpp"`; do
                 if [ -f "$i" ]; then
                     message 2 "Found NV library: $i\n"
                     if [ -z "${SINGULARITY_CONTAINLIBS:-}" ]; then


GodloveD:
I don't think this is something you actually want to do.  You *want* to install CUDA into your container because there are many different versions of CUDA and you need the version that will work with your application.  

The way we are doing this now, you actually don't need to have CUDA installed on your host system.  That's desirable.  You also don't need to run the same version of CUDA in your container as that running on the host system.  In other words, the way we are currently doing it is portable.  This would make it less so.  

If installing CUDA is a hassle, try starting your build from one of [NVIDIA's Docker Hub CUDA containers](https://hub.docker.com/r/nvidia/cuda/).

mrmhodak:
Are you arguing for elimination of the "--nv" option?

The problem is that CUDA in the container cannot be truly portable, because it depends on the matching driver in kernel. There is no perfect solution here, you can either (i) create a container with CUDA matching the kernel driver, or (ii) use both libraries and kernel driver from the machine. 

My commit is trying to improve (ii) to work seamlessly. A user may want to just use whatever is on the machine, this actually makes the container more portable rather than thinking about whether CUDA in the container matches kernel driver.  As long as we have "--nv" it may as well work correctly. It is, after all, up to user to specify it, it is not a default.

GodloveD:
Sorry, let me try to give a concrete example.  Let's say I have a k80 with the latest greatest driver.  But my admin is lazy so we have CUDA 7.5.  I want to run the latest greatest TensorFlow, so I need CUDA 8.  I can install it in my container along with CUDA 8, but I need to either also install the driver (which means my container only works on machines with the exact same driver) or use `--nv`.  Right now, using `--nv` works.  But with your proposed change the CUDA in my container gets overwritten by 7.5 and my TensorFlow no longer works.  So now I'm out of options and I have to install both CUDA and the driver in my container and forget about the `--nv` option.  

As a note, the driver in the kernel doesn't have to match CUDA.  In fact, there is no such thing.  It just has be be sufficiently up to date to run whatever version of CUDA you are trying to use.

mrmhodak:
OK, so now I am confused. "--nv" already imports libcuda*, isn't that part of CUDA itself?

The other issue is that there is no documentation on what the option is supposed to do.  From examining source, I assumed to was to automatically include CUDA libraries alongside kernel driver stuff.

GodloveD:
This is confusing.  And NVIDIA does little to unmuddy the waters.  😜 

So CUDA and the NVIDIA driver are 2 different things.  To understand what we are trying to map into the container at runtime, try this.  First download an NVIDIA driver .run file, extract it's contents and inspect them.  These are all of the things that are specific to a particular driver version and that we are trying to map into the container.  

You will see `libcuda.so` even though this is the *driver* and not *CUDA*:
```
$ mkdir nvidia-stuffs && cd nvidia-stuffs

$ wget http://us.download.nvidia.com/XFree86/Linux-x86_64/375.66/NVIDIA-Linux-x86_64-375.66.run

$ sh NVIDIA-Linux-x86_64-375.66.run --extract-only

$ ls -l NVIDIA-Linux-x86_64-375.66/*cuda*
-rwxr-xr-x 1 ubuntu ubuntu 8240968 May  1 21:40 NVIDIA-Linux-x86_64-375.66/libcuda.so.375.66
-rwxr-xr-x 1 ubuntu ubuntu   60568 May  1 21:42 NVIDIA-Linux-x86_64-375.66/nvidia-cuda-mps-control
-rw-r--r-- 1 ubuntu ubuntu    2044 May  1 21:42 NVIDIA-Linux-x86_64-375.66/nvidia-cuda-mps-control.1.gz
-rwxr-xr-x 1 ubuntu ubuntu   46344 May  1 21:42 NVIDIA-Linux-x86_64-375.66/nvidia-cuda-mps-server
```
Now download CUDA and extract it's contents.  
```
$ wget https://developer.nvidia.com/compute/cuda/8.0/Prod2/local_installers/cuda_8.0.61_375.26_linux-run

$ sh cuda_8.0.61_375.26_linux-run --extract=~/nvidia-stuffs/

$ ls -l
-rw-rw-r-- 1 ubuntu 1.4G Feb  2 00:17 cuda_8.0.61_375.26_linux-run
-rwxrwxr-x 1 ubuntu 1.3G Jun 27 22:00 cuda-linux64-rel-8.0.61-21551265.run
-rwxrwxr-x 1 ubuntu  97M Jun 27 22:00 cuda-samples-linux-8.0.61-21551265.run
-rwxrwxr-x 1 ubuntu  73M Jun 27 22:00 NVIDIA-Linux-x86_64-375.26.run
drwxrwxr-x 7 ubuntu 4.0K Jun 27 21:48 NVIDIA-Linux-x86_64-375.66
-rw-rw-r-- 1 ubuntu  72M May  2 16:17 NVIDIA-Linux-x86_64-375.66.run
```
You'll notice that this actually only extracted 3 compressed archives (which will need to be extracted again in turn to be viewed).  One of the compressed archives is another NVIDIA driver installer!  That's NVIDIA's way of trying to confuse you.  You actually only need to run `cuda_8.0.61_375.26_linux-run` to install CUDA.  You can keep whatever driver you already have on the system.  So the contents of that file are things we *don't* want to bind mount into the container, but we do want to install into the container along with our app.  As I mentioned above, starting with an NVIDIA container from Docker Hub is an easy way to do this.  It contains CUDA and optionally cuDNN, without the driver.

mrmhodak:
Very enlightening, thanks!

However, there still appears to be an issue with automatically imported libraries. Using docker bootstrap (nvidia/cuda:8.0-cudnn5-devel-centos7 - I assume that this is what you meant as a container with CUDA, cuDNN but without driver), the following libraries are both inside the container and also auto-imported: 
libcudart.so
libnvrtc-builtins.so
libnvrtc.so
libnvToolsExt.so
libOpenCL.so

With the exception of the last one, this may be the case of grep in argparser catching more than intended. Nevertheless, this should be fixed as it can cause problems when mixing different versions of kernel drivers and CUDA libraries

flx42:
I took a quick look at this issue and @GodloveD is correct. You might be interested to read the explanations on our [nvidia-docker wiki](https://github.com/NVIDIA/nvidia-docker/wiki/NVIDIA-driver).

Also, `libcuda.so` is part of the driver, but `libcudart.so` is part of the CUDA toolkit.

GodloveD:
Hiya @flx42!  What about @mrmhodak's last point?  I think they may be onto something there.  Should we ensure that those libs are *not* bind mounted into the container?  Sounds like `libcudart.so` at the least is being erroneously scooped up.

flx42:
Yes, this file is not part of the driver so it should not be mounted. But the `-nv` was done by @gmkurtzer, not us :)

GodloveD:
Indeed.  Thanks very much for chiming in on this @flx42.

mrmhodak:
@GodloveD and @flx42: Thanks for your input. 

I will create a more narrow commit to fix the doubly included libraries.

gmkurtzer:
As @flx42 mentions:
> Also, libcuda.so is part of the driver, but libcudart.so is part of the CUDA toolkit.

Yes, we need to be more careful about which libraries we are importing into the container at runtime as our current pattern match will pull both of these, as was mentioned above.

If someone wants to take a stab at this go for it, otherwise I'll add it to my TODO list.

Thanks!

mrmhodak:
Let me take a stab at this, I will have something ready to go soon. I just need to figure out what is going on with libOpenCL

PR #775, opened 2017-06-28
Honor $TMPDIR instead of hardcoding /tmp

marcelm:
This allows a user to choose where (some) temporary directories such as
`.singularity-layers.XXXXXXXX` are placed by setting the `$TMPDIR` environment
variable. This may allow them, for example, to avoid running into limitations imposed on
`/tmp` that are not in place for `$TMPDIR`.

Notes:
* Possibly `mktemp --tmpdir template.XXX` is not portable enough. Then `mktemp ${TMPDIR:-/tmp}/template.XXX` could be used instead.
* It would be nice to set `$TMPDIR` once at the beginning if it is unset and then use only `$TMPDIR` subsequently instead of repeating `${TMPDIR:-/tmp}`. Autoconf uses `: ${TMPDIR=/tmp}` to achieve this.

**Checkoff for all PRs:**

- [ ] I have added myself as a contributor to the [Author's file](https://github.com/singularityware/singularity/blob/master/AUTHORS.md)

Attn: @singularityware-admin

### DIFF
--- a/libexec/bootstrap-scripts/deffile-driver-docker.sh
+++ b/libexec/bootstrap-scripts/deffile-driver-docker.sh
@@ -52,7 +52,7 @@ fi
 SINGULARITY_CONTAINER="docker://$FROM"
 SINGULARITY_LABELFILE="$SINGULARITY_ROOTFS/.singularity.d/labels.json"
 
-SINGULARITY_CONTENTS=`mktemp /tmp/.singularity-layers.XXXXXXXX`
+SINGULARITY_CONTENTS=`mktemp ${TMPDIR:-/tmp}/.singularity-layers.XXXXXXXX`
 export SINGULARITY_CONTAINER SINGULARITY_CONTENTS SINGULARITY_LABELFILE
 
 eval_abort "$SINGULARITY_libexecdir/singularity/python/import.py"
--- a/libexec/bootstrap-scripts/deffile-driver-docker.sh
+++ b/libexec/bootstrap-scripts/deffile-driver-docker.sh
@@ -52,7 +52,10 @@ fi
 SINGULARITY_CONTAINER="docker://$FROM"
 SINGULARITY_LABELFILE="$SINGULARITY_ROOTFS/.singularity.d/labels.json"
 
-SINGULARITY_CONTENTS=`mktemp ${TMPDIR:-/tmp}/.singularity-layers.XXXXXXXX`
+if ! SINGULARITY_CONTENTS=`mktemp ${TMPDIR:-/tmp}/.singularity-layers.XXXXXXXX`; then
+    message ERROR "Failed to create temporary directory\n"
+    ABORT 255
+fi
 export SINGULARITY_CONTAINER SINGULARITY_CONTENTS SINGULARITY_LABELFILE
 
 eval_abort "$SINGULARITY_libexecdir/singularity/python/import.py"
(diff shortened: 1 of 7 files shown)


vsoch:
It's definitely a good idea to not hard code things. Is `--tmpdir` supported across all the systems / versions that Singularity needs to run on? For example --> https://code.kodo.org.uk/dom/buildroot/commit/53223cbb4e775ce88c036ddbfcca260e23c5739e

vsoch:
Also note if this is changed, tmp is also used in some of the bootstrap scripts.

marcelm:
You’re right. The `-t` option is deprecated according to `mktemp --help`, but of course it’s better to use a deprecated option than to fail on some systems. I’ve changed that and also changed two occurrences in the bootstrap scripts. There’s one more hard-coded `/tmp/` in `tests/functions`, but I guess that doesn’t matter. Then there is also this in `libexec/cli/status.exec`:
```
SESSION_GLOB="/tmp/.singularity-session-${USERID}.*/daemon.pid"
```
As far as I understand, this is an unfinished subcommand, so I did not change it, but I believe best practice would be to put pid files into `/var/run/` anyway, not in `/tmp`.

vsoch:
I'm not sure about the pid (although I think you are right that [var/run is best practice](https://stackoverflow.com/questions/5173636/must-my-pidfile-be-located-in-var-run), I think tmp would be ok, but we can check with @gmkurtzer if somewhere in var/run is more appropriate. Perhaps it has something to do with where the file is needed in relation to the host and container? For the test stuffs we can be a little more lenient, only because we know the exact environment / file structure where it is to be run on Travis.

marcelm:
To clarify, I commented on the pid file location only because I saw that while searching for `/tmp`. I’m not suggesting it should be part of this PR.

gmkurtzer:
Hi all,

I would prefer we don't use the `-t` flag to `mktemp` but instead specify the `$TMPDIR` variable by hand within that path including a default. So change `mktemp -t .singularity.....` to `mktemp ${TMPDIR:-/tmp}/.singularity....` which would achieve the same outcome.

As far as `status.exec`, while it does make more sense to put it in `/var/run/`, users don't and should not have permission to write to that part of the file system. Moving forward with the start/stop/status work that @bauerm97 is working on, we will be putting those in the `~/.singularity/` directories within each user's home.

Thanks!

vsoch:
@gmkurtzer and @bauerm97 start/stop and general status maps really nicely to webby service things - please ping me when you need testing / feedback, etc.

marcelm:
I’ve added the requested change of not using `-t`. Also, every `mktemp` is now checked for success in an `if`. Previously, this was not done everywhere.

marcelm:
The CI failure seems to be caused by busybox.net not being reachable.

vsoch:
ok, rebuilt and we are good :)

libexec/bootstrap-scripts/deffile-driver-docker.sh
```diff
@@ -52,7 +52,10 @@ fi
 SINGULARITY_CONTAINER="docker://$FROM"
 SINGULARITY_LABELFILE="$SINGULARITY_ROOTFS/.singularity.d/labels.json"
 
-SINGULARITY_CONTENTS=`mktemp /tmp/.singularity-layers.XXXXXXXX`
+if ! SINGULARITY_CONTENTS=`mktemp ${TMPIR:-/tmp}/.singularity-layers.XXXXXXXX`; then
```
gmkurtzer: There are some typos here on the `TMPDIR` (`TMPIR`) that @bauerm97 just spotted with me. Can you correct those?

marcelm:
I fixed the typos - sorry about that.

gmkurtzer:
This looks great to me, merging into development branch. Thanks!

marcelm:
Nice, thanks for merging!

Issue #701, opened 2017-06-01
Slurm plugin missing header


gdolle:
When trying to compile the slurm plugin `--with-slurm`, I get an error with a missing header:

```
singularity.c:32:29: fatal error: lib/singularity.h: No such file or directory  #include "lib/singularity.h"
```

Maybe this include should be removed ? (like here https://github.com/singularityware/singularity/commit/e699be95cdb7c2352f22dfa844c2cec301e0adfe)

## Version of Singularity:
Release 2.3.0

kav2k:
To clarify: this only happens when trying to compile Slurm integration plugin.

moskalenko:
Same here, full error:

Making all in slurm
make[3]: Entering directory `singularity-2.3/src/slurm'
depbase=`echo singularity.lo | sed 's|[^/]*$|.deps/&|;s|\.lo$||'`;\
        /bin/sh ../../libtool --tag=CC   --mode=compile gcc -DHAVE_CONFIG_H -I. -I../../src     -I/opt/slurm/include -MT singularity.lo -MD -MP -MF $depbase.Tpo -c -o singularity.lo singularity.c &&\
        mv -f $depbase.Tpo $depbase.Plo
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I../../src -I/opt/slurm/include -MT singularity.lo -MD -MP -MF .deps/singularity.Tpo -c singularity.c  -fPIC -DPIC -o .libs/singularity.o
singularity.c:32:29: error: lib/singularity.h: No such file or directory
singularity.c: In function ‘setup_container_environment’:
singularity.c:66: warning: incompatible implicit declaration of built-in function ‘snprintf’
singularity.c: In function ‘setup_container_cwd’:
singularity.c:130: warning: initialization makes pointer from integer without a cast
singularity.c: In function ‘setup_container’:
singularity.c:148: warning: incompatible implicit declaration of built-in function ‘printf’
singularity.c:172: error: redeclaration of ‘image’ with no linkage
singularity.c:145: note: previous declaration of ‘image’ was here
singularity.c:173: warning: assignment makes pointer from integer without a cast
make[3]: *** [singularity.lo] Error 1
make[3]: Leaving directory `singularity-2.3/src/slurm'
make[2]: *** [all-recursive] Error 1
make[2]: Leaving directory `singularity-2.3/src'
make[1]: *** [all] Error 2
make[1]: Leaving directory `singularity-2.3/src'
[...]

PR #783, opened 2017-06-30
Patch to fix slurm plugin build (development)

votavap:

This patch fixes the issues with building the Slurm plugin for Singularity 2.3.1

- Ref: #701  

**Checkoff for all PRs:**

- [x ] This PR is NOT against the project's master branch
- [ ] I have added myself as a contributor to the [Author's file](https://github.com/singularityware/singularity/blob/master/AUTHORS.md)

Attn: @singularityware-admin

### DIFF
--- a/src/slurm/Makefile.am
+++ b/src/slurm/Makefile.am
@@ -4,7 +4,7 @@ plugindir = $(libdir)/slurm
 if WITH_SLURM
 plugin_LTLIBRARIES = singularity_spank.la
 singularity_spank_la_SOURCES = singularity.c
-singularity_spank_la_LIBADD = ../lib/libsingularity_internal.la
+singularity_spank_la_LIBADD = ../lib/runtime/libinternal.la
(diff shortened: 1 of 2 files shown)


gmkurtzer:
Heya @votavap,

Is this PR ready to be tested/merged?

note: I thought there would be quite a bit more that needs to be changed/fixed, but I'll take a look at it once you give me the thumbs up.

Thanks!

votavap:
Hi @gmkurtzer 

I have tested the build process, so it may be ready in that regard. I suspect that the next step will be to reconcile some of the the new API, because I think there will be few symbols that may not resolve at runtime. I may be able to test it with Slurm on Monday, and will see how it goes, but if there are problems, maybe it would be good to have the build patch in the development branch and deal with the actual plugin logic as a separate issue. What do you think?

gmkurtzer:
I would prefer to have PRs resolve to a working implementation (or at least as far as we know) generally speaking, but considering the SLURM module doesn't even build at present, I concur that this is a step in the right direction.

PR #785, opened 2017-07-01
Fix --nv option

mrmhodak:
--nv option needs to be adjusted because it "imports" several CUDA libraries, which should be provided by the container.

This commit does two things:
1. Remove two unnecessary entries from the "grep" pattern:
   a) "libnvcu", because "libnv" already catches that
   b} "libOpenCL". Strangely, both kernel driver and cuda  provide libOpenCL.so (installed in two different locations), which differ according to diff. However, the cuda one has a higher priority according to ldconfig, and so OpenCL already in the container should be used. Because of this the grep pattern should not include libOpenCL.
2. Because grep catches several cuda-related libraries, specifically libcudart.so*, libnvrtc.so, and libnvToolsExt.so* I am adding grep -v statement after the pipe to avoid this.

One way this issue manifested itself, was that libraries in /.singularity.d/libs were replicated with libraries inside the container. With this fix, it does not happen anymore.

Finally, this is related to Pull Request #774 which I am going to delete.

Attn: @singularityware-admin

### DIFF
--- a/libexec/cli/action_argparser.sh
+++ b/libexec/cli/action_argparser.sh
@@ -116,23 +116,24 @@ while true; do
         ;;
         -n|--nv)
             shift
-            for i in `ldconfig -p | grep -E "/libnv|/libcuda|/libEGL|/libGL|/libvdpau|/libOpenGL" \
-                    | grep -v -E "/libcudart|/libnvrtc|/libnvToolsExt"`; do
-                if [ -f "$i" ]; then
-                    message 2 "Found NV library: $i\n"
-                    if [ -z "${SINGULARITY_CONTAINLIBS:-}" ]; then
-                        SINGULARITY_CONTAINLIBS="$i"
-                    else
-                        SINGULARITY_CONTAINLIBS="$SINGULARITY_CONTAINLIBS,$i"
+            for lib in $(cat $SINGULARITY_sysconfdir"/singularity/nvliblist.conf" | grep -v "^#"); do
+                for i in $(ldconfig -p | grep "${lib}"); do
+                    if [ -f "$i" ]; then
+                        message 2 "Found NV library: $i\n"
+                        if [ -z "${SINGULARITY_CONTAINLIBS:-}" ]; then
+                            SINGULARITY_CONTAINLIBS="$i"
+                        else
+                            SINGULARITY_CONTAINLIBS="$SINGULARITY_CONTAINLIBS,$i"
+                        fi
                     fi
-                fi
+                done
             done
             if [ -z "${SINGULARITY_CONTAINLIBS:-}" ]; then
                 message WARN "Could not find any Nvidia libraries on this host!\n";
             else
                 export SINGULARITY_CONTAINLIBS
             fi
-            if NVIDIA_SMI=`which nvidia-smi`; then
+            if NVIDIA_SMI=$(which nvidia-smi); then
                 if [ -n "${SINGULARITY_BINDPATH:-}" ]; then
                     SINGULARITY_BINDPATH="${SINGULARITY_BINDPATH},${NVIDIA_SMI}"
                 else
--- a/libexec/cli/action_argparser.sh
+++ b/libexec/cli/action_argparser.sh
@@ -117,7 +117,7 @@ while true; do
         -n|--nv)
             shift
             SINGULARITY_NVLIBLIST=$(mktemp singularity-nvliblist.XXXXXXXX)
-            cat $SINGULARITY_sysconfdir"/singularity/nvliblist.conf" | grep -v "^#" > $SINGULARITY_NVLIBLIST
+            cat $SINGULARITY_sysconfdir"/singularity/nvliblist.conf" | grep -Ev "^#|^\s*$" > $SINGULARITY_NVLIBLIST
             for i in $(ldconfig -p | grep -f "${SINGULARITY_NVLIBLIST}"); do
                 if [ -f "$i" ]; then
                     message 2 "Found NV library: $i\n"
--- a/libexec/cli/action_argparser.sh
+++ b/libexec/cli/action_argparser.sh
@@ -116,18 +116,19 @@ while true; do
         ;;
         -n|--nv)
             shift
-            for lib in $(cat $SINGULARITY_sysconfdir"/singularity/nvliblist.conf" | grep -v "^#"); do
-                for i in $(ldconfig -p | grep "${lib}"); do
-                    if [ -f "$i" ]; then
-                        message 2 "Found NV library: $i\n"
-                        if [ -z "${SINGULARITY_CONTAINLIBS:-}" ]; then
-                            SINGULARITY_CONTAINLIBS="$i"
-                        else
-                            SINGULARITY_CONTAINLIBS="$SINGULARITY_CONTAINLIBS,$i"
-                        fi
+            SINGULARITY_NVLIBLIST=$(mktemp singularity-nvliblist.XXXXXXXX)
+            cat $SINGULARITY_sysconfdir"/singularity/nvliblist.conf" | grep -v "^#" > $SINGULARITY_NVLIBLIST
+            for i in $(ldconfig -p | grep -f "${SINGULARITY_NVLIBLIST}"); do
+                if [ -f "$i" ]; then
+                    message 2 "Found NV library: $i\n"
+                    if [ -z "${SINGULARITY_CONTAINLIBS:-}" ]; then
+                        SINGULARITY_CONTAINLIBS="$i"
+                     else
+                        SINGULARITY_CONTAINLIBS="$SINGULARITY_CONTAINLIBS,$i"
                     fi
-                done
+                fi
             done
+            rm $SINGULARITY_NVLIBLIST
             if [ -z "${SINGULARITY_CONTAINLIBS:-}" ]; then
                 message WARN "Could not find any Nvidia libraries on this host!\n";
             else
--- a/libexec/cli/action_argparser.sh
+++ b/libexec/cli/action_argparser.sh
@@ -116,7 +116,8 @@ while true; do
         ;;
         -n|--nv)
             shift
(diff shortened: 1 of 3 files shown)


mrmhodak:
@GodloveD and @flx42: Any comments?

I believe I got this one right after feedback I got from you for PR #774

mrmhodak:
Here are the libraries that are no longer included after this commit (for Cuda 8.0) along with the rpm package they are from:
LIBRARY                                RPM PACKAGE 
libcudart.so                           cuda-cudart
libcudart.so.8.0                     cuda-cudart
libnvrtc-builtins.so                cuda-nvrtc
libnvrtc-builtins.so.8.0          cuda-nvrtc
libnvrtc.so                             cuda-nvrtc
libnvrtc.so.8.0                       cuda-nvrtc
libnvToolsExt.so                    cuda-cudart
libnvToolsExt.so.1                 cuda-cudart 
libOpenCL.so                        cuda-cudart-dev
libOpenCL.so.1                     cuda-cudart-dev

As I mentioned previously, the last two (libOpenCL) are somewhat tricky as two packages install these in two different locations: There is both /usr/lib64/nvidia/libOpenCL.so and /usr/local/cuda-8.0/targets/x86_64-linux/lib/libOpenCL.so. The former is from the kernel driver (xorg-x11-drv-nvidia-devel) and the latter from cuda ( cuda-cudart-dev). The cuda one has a higher priority in ldconfig, so there is no need to "import" from the host inside the container.

GodloveD:
@mrmhodak I will look at this and verify that all bound libs appear in the NVIDIA driver .run file as soon as I can.  Thanks!

[EDIT] 
@mrmhodak Thanks very much for identifying this problem.  I'm now testing this on a different system and I see that there are a few other libraries erroneously passing the pattern match.

In addition to these (which are excluded by your fix).
```
libnvToolsExt.so
libnvrtc-builtins.so
libnvrtc.so
libcudart.so
```
I'm also getting these:
```
libGLEW.so
libGLEWmx.so
libGLU.so
libnvblas.so
libnvgraph.so
```
While `libnvblas.so` and `libnvgraph.so` are part of CUDA, the other libraries don't have anything to do with CUDA.  

I think this is going to be very system specific.   Maybe it's a bad idea to do a broad pattern-match.    Should we maintain list of exact libraries to search for instead based on the contents of the driver `.run` installers?  What do you think @gmkurtzer?

GodloveD:
I spoke with @gmkurtzer offline and I think that we are going to try to create a configuration file that contains all of the libs that should be bound by the `--nv` option.  That way we can update it easily when an new driver is released or an admin can customize it to suit their needs.  I just looked in a fairly recent driver .run file (375.66) and these are the libs that I believe need to be bound.  Can you verify (or not) with whatever driver version you are running @mrmhodak ?
```
libcuda.so
libEGL_installertest.so
libEGL_nvidia.so
libEGL.so
libGLdispatch.so
libGLESv1_CM_nvidia.so
libGLESv1_CM.so
libGLESv2_nvidia.so
libGLESv2.so
libGL.so
libGLX_installertest.so
libGLX_nvidia.so
libglx.so
libGLX.so
libnvcuvid.so
libnvidia-cfg.so
libnvidia-compiler.so
libnvidia-eglcore.so
libnvidia-egl-wayland.so
libnvidia-encode.so
libnvidia-fatbinaryloader.so
libnvidia-fbc.so
libnvidia-glcore.so
libnvidia-glsi.so
libnvidia-gtk2.so
libnvidia-gtk3.so
libnvidia-ifr.so
libnvidia-ml.so
libnvidia-opencl.so
libnvidia-ptxjitcompiler.so
libnvidia-tls.so
libnvidia-wfb.so
libOpenGL.so
```

flx42:
You might not need all these libraries, you should take a look at the README file shipped with our drivers, for instance at `/usr/share/doc/nvidia-375/README.txt.gz`, chapter 5 "Listing of Installed Components".

3XX0:
There are a lot of things that could wrong here. I suggest you leverage libnvidia-container for this. In fact, I will probably add an option to our CLI to do exactly that.

GodloveD:
@flx42 @3XX0 can either of you tell me how much these libs have changed in different releases of the driver?  Do they tend to be pretty stable or are new libraries introduced from time to time?

3XX0:
Difficult to say, it's pretty random and things get revamped all the time. This is why we have the library, to abstract these things away. Also, there are not only libraries but also binaries, IPCs, ICD loaders etc.

mrmhodak:
Adding changes by @GodloveD. 
"I've broken the nvidia libs out into a config file that can be changed by an admin. This way if there is something non-standard about the driver that is running on a particular system the admin can tailor the libs to their system."

libexec/cli/action_argparser.sh
```diff
@@ -116,22 +116,24 @@ while true; do
...
             shift
-            for i in `ldconfig -p | grep -E "/libnv|/libcuda|/libEGL|/libGL|/libnvcu|/libvdpau|/libOpenCL|/libOpenGL"`; do
-                if [ -f "$i" ]; then
-                    message 2 "Found NV library: $i\n"
-                    if [ -z "${SINGULARITY_CONTAINLIBS:-}" ]; then
-                        SINGULARITY_CONTAINLIBS="$i"
-                    else
-                        SINGULARITY_CONTAINLIBS="$SINGULARITY_CONTAINLIBS,$i"
+            for lib in $(cat $SINGULARITY_sysconfdir"/singularity/nvliblist.conf" | grep -v "^#"); do
+                for i in $(ldconfig -p | grep "${lib}"); do
```
gmkurtzer: I would prefer this being done with a single grep like `grep -f "$SINGULARITY_sysconfdir"/singularity/nvliblist.conf"` instead of nested loops.

gmkurtzer:
Aside from the minor change I mentioned, this looks great to me.

@3XX0 I'd like to hear more about the libnvidia-container CLI option you are thinking of. Would it return a list of nvidia libraries that should be included in the container? If so, that would be a fantastic addition and we can easily check for libnvidia-container, and use it if it is installed.

Thanks!

3XX0:
@gmkurtzer Yes we have an utility called `nvidia-container-cli`, I can add an option `list` to output the list of components to be mounted. It will have to be installed separately (we are working on packaging right now) but eventually it will be shipped alongside the driver.

gmkurtzer:
@3XX0 Oh, that sounds perfect. Let us know when it is ready and released, and we will update this static structure to use the output of that command instead.

libexec/cli/action_argparser.sh
GodloveD: I'm not sure if I follow what you're thinking.  The first loop iterates over the library patterns in nvliblist.conf while the second loop iterates over the matches in the ld.so.cache.  Sorry if I'm missing something easy, but I don't see how it could be done with a single loop.

gmkurtzer: By using `ldconfig -p | grep -f...` and then looping over the output of that, you can omit the outer loop completely.

mrmhodak: That would work, but the comments and empty lines need to be stripped first

GodloveD: Ah.  Thanks for the clarification.  I see what you mean.  I just implemented it.  Will fire off another pull request to @mrmhodak in a minute.

mrmhodak:
Adding a fix for nested loops by @GodloveD - it was originally committed against a wrong branch in my fork.

gmkurtzer:
Is this ready to be merged?

GodloveD:
It's good to go from my point of view.  I tested `--nv` with a tensorflow job and verified that it can see and use the GPU with this configuration.  @mrmhodak can you double check that these libs are correct and give it a test?

mrmhodak:
Hmm, I have just tried it and see 600 libraries being imported, was 50 with previous commit. Do not have time to investigate further right now, will double check it later.

@GodloveD  do you see the same?

GodloveD:
ACK!  I think I tested this with the wrong branch.  Thanks for double checking @mrmhodak!  Yes I also see way too many libs.  I'll have a look and see where things are going wrong.

libexec/cli/action_argparser.sh
```diff
@@ -116,22 +116,25 @@ while true; do
         ;;
         -n|--nv)
             shift
-            for i in `ldconfig -p | grep -E "/libnv|/libcuda|/libEGL|/libGL|/libnvcu|/libvdpau|/libOpenCL|/libOpenGL"`; do
+            SINGULARITY_NVLIBLIST=$(mktemp singularity-nvliblist.XXXXXXXX)
+            cat $SINGULARITY_sysconfdir"/singularity/nvliblist.conf" | grep -v "^#" > $SINGULARITY_NVLIBLIST
```
GodloveD: This leaves an empty line.  Which will apparently match everything (`\n') in the output of `ldconfig -p`. YUCK!

GodloveD: I think it needs to be:
```bash
cat $SINGULARITY_sysconfdir"/singularity/nvliblist.conf" | grep -Ev "^#|^\s*$"
```
I'm going to test a whole bunch more though.

mrmhodak: This looks like a right way to fix the issue and my testing confirms that it works. Adding into the branch.

gmkurtzer:
Heya guys. Just circling back on this. How are we now?

mrmhodak:
I have tested this on several machines and works  fine for me. 

What do you think @GodloveD ?

GodloveD:
I think we are good to go!

3XX0:
I added the list command we talked about in [libnvidia-container](https://github.com/NVIDIA/libnvidia-container).
For most use cases, you will probably want to do `nvidia-container-cli list -cuv`

GodloveD:
Thanks @3XX0.  I'll have a look at that as soon as I can.  Thanks too for giving us the command.  

@gmkurtzer for now I think we should just merge this because it fixes a potential bug pointed out by @mrmhodak.  But moving forward I will investigate using the libnvidia-container approach instead because it is probably much better.

gmkurtzer:
Looks great everyone, thank you!

PR #788, opened 2017-07-02
Clearer error message when download shub image that isn't ready

vsoch:
Right now, when a user attempts to download a singularity hub image for which the file doesn't exist, the manifest returns None, and the download fails and asks the user if he/she has permission to write to the directory.

```
ERROR Error downloading None. Do you have permission to write to [current folder]
```

This is not ideal, as it indicates the wrong error. Instead, we do a check first to ensure that the url exists, and if it doesn't (meaning the image isn't built) we tell this to the user off the bat and exit with error:

```
singularity pull shub://1633
ERROR YeoLab-cellranger-master.img.gz is not ready for download
ERROR please try when build is complete, or specify tag.
```

Attn: @singularityware-admin


vsoch:
@gmkurtzer this is a simple PR that should be ready for merge.

gmkurtzer:
Indeed, and looks great. Thanks!

yarikoptic:
sorry to pester -- when is new release scheduled to take advantage of the fixes like this? ;)

Issue #782, opened 2017-06-29
Feature: Packaging of Node


vsoch:
This is a summary of a feature that (is going to be awesome) and have many use cases for 2.4. It's the idea that someone can package the node that they are working on, and doesn't have to know a thing about containers to make them.

I think this could be integrated into Singularity as a bootstrap module, but for now I've implemented something to test [here](https://github.com/vsoch/singularity-python/tree/development/examples/create/package_node). Specifically, it looks something like this:

## On Node
```
from singularity.package import package_node
package = package_node()
 
package
'/tmp/tmp06ww_8_n/vanessa-testing-self-cloning.tgz'
```

optionally, the user can select a different root (default is `/`) or name for the image (the default name is the name of the machine)

## On Packaging Node

```
from singularity.package import unpack_node
image = unpack_node(image_path=package,size=9000)
```

The current issues with the above have to do with the metadata folder (eg, there is no action for the container, or anything else for that matter) but if we integrate this into bootstrap, that goes away.

## Workflow
This supports  a slightly different framework than is supported on most clusters - one that is more akin to requesting instances on demand on Google Cloud (and likely some larger resources do have with something like OpenStack). Under this design, a user is given a node on demand, disconnected from the cluster, and allowed to do whatever they please!

 - history is saved as a quasi "deffile" %post section, to record exactly what they did
- full permission is given to install, clone, etc.
 
When the user has finished, a trigger is run to package the image, and sent somewhere (a different machine with enough space) to package it into an image proper (the unpack_image function above, which would be supported through singularity hopefully to create a completely working image).

## Features
For an actual imlpementation, we would want to have the following:
 - security checks for the image before it's done
 - testing in the same way
 - capturing metadata, akin to what singularity hub does (which is pretty crappy, we can do a lot better)!
 
## TLDR

 1. user logs into node
 2. user does work, perhaps it's recorded in history (to save and mimic as some kind of deffile)
 3. when user is finished, some kind of trigger is done (or running a command for an admin to do so) to package the node
 4. the node is packaged, moved somewhere where it can be tested, measured, and turned into an image.
 5. the image is made available for the user.

truatpasteurdotfr:
- history cleanup (no trace of credentials like "echo changemenow| sudo passwd --stdin root")
- user controlled whitelist/blacklist of what should be kept
[...]

PR #789, opened 2017-07-03
Infrastructure for Security (and other) Checks

vsoch:
This is the start of thinking for a framework to support security checks. The general idea is that a "check" is a pre-approved (and root owned) function (an executable under `libexec/helpers/checks` that can be run over a mounted file system, to check for some condition, and report a status about the condition back to the user. I have provided complete details in the [README here](https://github.com/vsoch/singularity/blob/add/security/libexec/helpers/README.md#checks) and will explain my thinking here as well.

# Definition: Checks
Broadly, a check is a script that is run over a mounted filesystem, primary with the purpose of checking for some security issue. This process is tighly controlled, meaning that the script names, located in a checks folder under `libexec/helpers` are hard coded into the calling script `check.sh` that is in the helpers folder. Every check script that is added has an assigned level (or severity) and optionally, one or more tags to give the user the option to filter. The flow of checks is the following:

 - the user calls `singularity check container.img` to invoke `check.exec`
 - specification of `--low` (3), `--med` (2), or `--high` (1) sets the level to perform. The level is a filter, meaning that a level of 3 will include 3,2,1, and a level of 1 (high) will only call checks of high priority.
 - specification of `--tag` will allow the user (or execution script) to specify a kind of check. 

## Why tags?
Tags allows us to extend checks to be other types of things. Right now we are thinking about security, but a check could equally be some kind of find or query function, a counting of files, anything really. For example:
- A user could contribute one check that has a unique tag of `findR` that aims only to find where the R executable is, and the user would just run `singularity check --tag findR container.img`. 
-  Tags are also useful to reflect the degree of something desired. I could have three scripts (at different levels) that perform the same function, but each less conservative.
- A tag can also help to be specific. For example, during bootstrap I might want to perform security checks specific to the bootstrap type. A script might have tag `security` but also tag `debootstrap` to specify it should only perform checks with one or both these tags. If the user wanted both, he would do `singularity check --tag "security+debootstrap" container.img` or for an OR,  `singularity check --tag "security|debootstrap" container.img`. I'm indifferent about the symbols we use, it could be an issue if people forget quotes and it turns into a pipe :)

## Examples
In addition to the examples below, I made a [quick asciinema](https://asciinema.org/a/127429?speed=3) to see (the current PR here) in action.

In the case of having these checks:

```
libexec/helpers/checks
├── 1-hello-world.sh
├── 2-hello-world.sh
└── 3-hello-world.sh
```

associated with these levels:

```
#             [LEVEL] [SCRIPT]                                                            [TAGS]
execute_check 1       "$SINGULARITY_libexecdir/singularity/helpers/checks/1-hello-world.sh" security
execute_check 2       "$SINGULARITY_libexecdir/singularity/helpers/checks/2-hello-world.sh" security
execute_check 3       "$SINGULARITY_libexecdir/singularity/helpers/checks/3-hello-world.sh" security
```

This first example shows the default level is for tag security, and all levels (low, med, high)

```
# Perform all security checks, these are the same
$ singularity check ubuntu.img
$ singularity check --tag security ubuntu.img
```
produces equivalent output, (currently) just a message to indicate what gets run:

```
/usr/local/libexec/singularity/helpers/checks/1-hello-world.sh tags[security] level[HIGH] indicated to run
/usr/local/libexec/singularity/helpers/checks/2-hello-world.sh tags[security] level[MED] indicated to run
/usr/local/libexec/singularity/helpers/checks/3-hello-world.sh tags[security] level[LOW] indicated to run
```

This second example shows only performing the high (severe risk) level:

```
# Perform high level security checks
$ singularity check --high ubuntu.img
/usr/local/libexec/singularity/helpers/checks/1-hello-world.sh tags[security] level[HIGH] indicated to run
```

and medium:

```
singularity check --med ubuntu.img
/usr/local/libexec/singularity/helpers/checks/1-hello-world.sh tags[security] level[HIGH] indicated to run
/usr/local/libexec/singularity/helpers/checks/2-hello-world.sh tags[security] level[MED] indicated to run
```
and these are not yet implemented, but tags would look like this:

```
# All security checks for tag security AND debootstrap
$ singularity check --tag "security+debootstrap" ubuntu.img

# All security checks for tag security OR debootstrap
$ singularity check --tag "security|debootstrap" ubuntu.img
```
 
### Adding a Check
A check should be a bash script that will perform some action. The following is required:

 - The script assumes a base of `SINGULARITY_ROOTFS`. This is defined when coming from bootstrap, and when called via `check.sh`, is set via `SINGULARITY_MOUNTPOINT`.
 - The script should indicate any warning/message to the user if the check is found to have failed. If pass (don't print anything? print pass?)
 - The script should return 0 when finished.
 - The script level, path, and tags should be added to [check.sh](check.sh) in the following format:

This is pretty rough, and I expect will change. The scripts are hard coded, with their levels and tags, in `check.sh`

```
##################################################################################
# CHECK SCRIPTS
##################################################################################

#             [LEVEL] [SCRIPT]                                                            [TAGS]
execute_check 1       "$SINGULARITY_libexecdir/singularity/helpers/checks/1-hello-world.sh" security
execute_check 2       "$SINGULARITY_libexecdir/singularity/helpers/checks/2-hello-world.sh" security
execute_check 3       "$SINGULARITY_libexecdir/singularity/helpers/checks/3-hello-world.sh" security
```

The function `execute_check` will compare the level with the user specified (or default) `SINGULARITY_CHECKLEVEL` and execute the check only given it is under the specified threshold, and (not yet implemented) has the relevant tag. If no tag is specified, either security can be default, or performing all checks.

## Important Notes:
 - checks can/should be performed on demand (as in this implementation) but also with bootstrap. In the case of bootstrap, the user should be able to disable.
 - a check is NOT a fix. A check simply checks and alerts the user of some condition, and makes no changes to the image. 
 - However, the original use case for wanting a check was with the `self` bootstrap, in which case we would want to enforce a core set of fixes, for the security of the host where the image is being generated. We can discuss the best way to perform actions based on check status as a supplement / next step for checks.

### custom variables / filtering for within a function
If  a user wanted to specify a whitelist / blacklist for looking over history, for example, how would that information get there? We could use environment variables, and they would be in the format `SINGULARITY_CHECK[SCRIPTNAME]_VARIABLE` or something like that. To have a complete record of these variables, my suggestion is that the requirement of each check should be that it has a help section written in the README.md.

### Script names being important
There is a lot of information that can be capturing in the naming of scripts, and I like the convention generally for finding things easily and organizing automatically based on file name. Right now, I just have a number that corresponds with the check level, and have made the suggestion of having fixes vs checks represented in this way. Thoughts?

### Fixes
Arguably, each check should (optionally) come with a fix. For example, if there is a history check for some kind of bad string, running the check would tell the user about it. Running the fix would tell the user and then remove the string where the image is mounted (thus this function must be performed with root, which makes sense, because the user is still building/developing). Possibly running checks (with no fix) could be done with non root. 

**How do we handle checks vs. fixes?**
This might be stupid, but we could do it simply based on script name. For example, in the function that calls the check, it could look for an equivalent script name, but one that has a different prefix. Given a return of 1 for the check, the existence of the fix script, and some environment variable to indicate to do fixes, the fix script would be executed.

## Next steps
1. Let's discuss this general set up, how can we improve it? Does it need to be re-worked?
2. When we like the general setup, then I can give a go at writing the first set of checks. As suggested by @truatpasteurdotfr in #782, I have a small list of fun things to try.

@gmkurtzer @truatpasteurdotfr @bauerm97 
Attn: @singularityware-admin


vsoch:
Also @gmkurtzer see my note here https://github.com/vsoch/singularity/blob/add/security/libexec/cli/check.exec#L91

bauerm97:
Vanessa, I think this is a very interesting idea. I would like to clarify something: are these check scripts provided by people contributing to the upstream repository (i.e. submitting a pull request for the script and making it available to everyone) or would sites implement their own check scripts?

vsoch:
I think we would provide the core, but the reason to implement in this way is because we would want sites to also be able to easily add some check to the folder.

gmkurtzer:
This looks fantastic! 

regarding your question, `SINGULARITY_MOUNTPOINT` is defined statically within `src/mount.c` as follows:
```
envar_set("SINGULARITY_MOUNTPOINT", singularity_runtime_rootfs(NULL), 1);
```
Can we use another variable name like `SINGULARITY_ROOTFS`?

vsoch:
@gmkurtzer I chose mountpoint instead of ROOTFS because, as is currently implemented, it mimics the action of a mount (and then runs checks over that mounted base). It would support bootstrap because, like in inspect, I set one variable to point to the other.

vsoch:
I just added the bits to actually run the scripts, and report back based on the return value (just like stest but not with exit on fail) so now it's like:

```
vanessa@vanessa-ThinkPad-T450s:~/Desktop/test/singularity$ singularity check --high ubuntu.img
PASS: (retval=0)/usr/local/libexec/singularity/helpers/checks/1-hello-world.sh tags[security] level[HIGH] 
vanessa@vanessa-ThinkPad-T450s:~/Desktop/test/singularity$ singularity check -m ubuntu.img
PASS: (retval=0)/usr/local/libexec/singularity/helpers/checks/1-hello-world.sh tags[security] level[HIGH] 
PASS: (retval=0)/usr/local/libexec/singularity/helpers/checks/2-hello-world.sh tags[security] level[MED] 
vanessa@vanessa-ThinkPad-T450s:~/Desktop/test/singularity$ singularity check -l ubuntu.img
PASS: (retval=0)/usr/local/libexec/singularity/helpers/checks/1-hello-world.sh tags[security] level[HIGH] 
PASS: (retval=0)/usr/local/libexec/singularity/helpers/checks/2-hello-world.sh tags[security] level[MED] 
PASS: (retval=0)/usr/local/libexec/singularity/helpers/checks/3-hello-world.sh tags[security] level[LOW]
```
what we could now do is have an environment variable that, if set, will try to fix the security issue (rather than just saying PASS or FAIL). I also think instead of directing output to tmp, the message on fail should be clearly shown.

gmkurtzer:
Apologies, I forgot to mention that the mountpoint argument to `.../libexec/singularity/bin/mount(-suid)` is not longer used. That is why `SINGULARITY_MOUNTPOINT` is defined, and it must be used as that is where the file system will be mounted.

vsoch:
heyo! Just a quick note that I'll be back to work on this when I get back from PEARC. See ya tomorrow @gmkurtzer !

GodloveD:
Hi @vsoch.  This is really cool!  I've been playing with this a little and I have one question.  It looks like if you want to add a new check you have to recompile the code.  Is that correct?  What would you think about letting an administrator add a shell script to the `/checks` directory and then be able to use that script without recompiling?

vsoch:
yes I thought about that - wouldn't it be an issue if someone could maliciously add a script? Oh but you are saying they would need to be root owned right? What would the bash look like to get and parse over the files in the folder, and the determine ownership?

GodloveD:
I just thought that the subdirectory itself would be root owned so only
root could add to it.  Do you think that is a safe assumption?

vsoch:
I think so, as long as there aren't any loopholes to inject / insert something malicious, I'm ok with this. Do you want to help me come up with a tiny list of "basic" checks to implement? I'd like to finish this up this weekend, and I can start on these tweaks this afternoon!! No more conference!! lol.

GodloveD:
I would be really interested to see what kinds of checks you have in mind!  I'll have to think and see what kinds of things I can come up with too.

GodloveD:
One thing we could do is check the container for anything with an SUID bit.  But I don't know how worrisome that is anyway given that you can't use it.  This is hard because you basically have to think about how you would make a malicious container.  That's hard to do with the Singularity model.

GodloveD:
Wait.  I think I'm having some ideas... 😈 I'll PM you.

vsoch:
@GodloveD I thought through this, and I think we want an approach that allows for both. The reason for hard coding into the helpers/check.sh is to allow assigning labels (levels and tags). And given that this isn't actually compiled (it's a shell script) it's not such a crazy idea to require an admin to drop the script in a folder, and then assign it a level. I realized that if we took the "drop in folder and go" approach we would still need a way to define this metadata, and that still comes down to editing a file somewhere.

So for the "both" - we would allow someone to run a (different) check on demand, using the check tool, and that way it would not require a level or tag.

vsoch:
@gmkurtzer and @GodloveD here is the recent addition:

https://asciinema.org/a/129059?speed=3

As stated in the asciinema, I'd like to discuss:

 - how we want to integrate into bootstrap vs. on demand check (mount isn't a go anymore @gmkurtzer ?)
 - if/how we want to make a set of default checks (via some tag perhaps?) and then how to allow the user to customize tags during bootstrap vs. on demand
 - how should the user add custom scripts (eg, on demand) other than adding to the checks folder and a line to check.sh?
 - who should be allowed to run, under what permission? (Given no changes are made, it makes sense that anyone could, UNLESS it exposes something important, or the check requires sudo to run).

I realize a lot of the password stuffs probably isn't relevant, because we don't include home. What I'd like to do is figure out reasonable answers to the above, implement, and then merge, announce, and get contributions from others for all the checks that (the users) desire. I suck at security so I generally guess at these things.

vsoch:
Also see the notes I added here https://github.com/vsoch/singularity/blob/add/security/libexec/helpers/README.md#adding-a-check

vsoch:
I'm adding the tags feature now - it will run tag "default" by default

vsoch:
okay, behold tags!! --> https://asciinema.org/a/129060?speed=3

This is so cool :) We can make checks to do ANYTHING **diabolical laugh**

vsoch:
ok! (and I swear to god I'm going home after this) I just added one more check to illustrate the following:

 - checks can require sudo
 - we should have checks for best practices for things like general image file system, and common docker issues.

What I did is add the (start to) a docker check script. And I'll show you here how it demonstrates the above. First, we can run the defaults, and we don't see the docker check (it doesn't have `default` tag)

```
$ singularity check  ubuntu.img
START 1-hello-world.sh tags[security default] level[LOW]
Hello World, security level 1
PASS: (retval=0) bash /usr/local/libexec/singularity/helpers/checks/1-hello-world.sh 

START 2-cache-content.py tags[default clean] level[LOW]
PASS: (retval=0) python /usr/local/libexec/singularity/helpers/checks/2-cache-content.py 
```
Also notice that I cleaned up the print output to NOT include redundant information. Now let's try a check for the docker tag, but without sudo (note that I will need sudo to look into `/root`)

```
$ singularity check --tag docker  ubuntu.img
START 4-docker.py tags[docker] level[LOW]
You must run this test as sudo, skipping
PASS: (retval=0) python /usr/local/libexec/singularity/helpers/checks/4-docker.py 
```
It tells me I need to run as root. Now let's do that:

```
START 4-docker.py tags[docker] level[LOW]
PROBLEM:  You should not put content in roots home.
RESOLVE:  Install to /opt or /usr/local
FAIL: (retval=1)
 python /usr/local/libexec/singularity/helpers/checks/4-docker.py 
```

And when I use a base (that doesn't have anything other than `.bashrc` and `profile` the test passes:

```
START 4-docker.py tags[docker] level[LOW]
PASS: (retval=0) python /usr/local/libexec/singularity/helpers/checks/4-docker.py 
```
WICKED COOL!
okay lol going home, can't wait to talk about this!

vsoch:
[whole shabang](https://asciinema.org/a/129063?speed=3)

vsoch:
The docker tag, for example, would nicely work with a docker bootstrap (when sudo is required). We could also have tags that just show info, or check for particular libraries / dependencies.

vsoch:
ok! Just added checks to bootstrap. Here is how it works:

Normal bootstrap is the same (checks disabled by default)
```
sudo singularity bootstrap image.img Singularity 
Sanitizing environment
Docker image path: index.docker.io/library/centos:7
Cache folder set to /root/.singularity/docker
Exploding layer: sha256:7b6bb4652a1be64d39920f7d8c12c5f0bd8abfc10820177109e73817357c07e3.tar.gz
Exploding layer: sha256:60e3079474b1671bcc20b340572ee6cae735ecdd5a13d6686475ca4916eb6003.tar.gz
Adding runscript
Skipping checks
```
but note the message to skip checks. When we ask for checks, default runs level 3, default tag:

```
sudo singularity bootstrap --checks image.img Singularity 
Sanitizing environment
Docker image path: index.docker.io/library/centos:7
Cache folder set to /root/.singularity/docker
Exploding layer: sha256:7b6bb4652a1be64d39920f7d8c12c5f0bd8abfc10820177109e73817357c07e3.tar.gz
Exploding layer: sha256:60e3079474b1671bcc20b340572ee6cae735ecdd5a13d6686475ca4916eb6003.tar.gz
Adding runscript
Running checks
START 1-hello-world.sh tags[security default] level[LOW]
Hello World, security level 1
PASS: (retval=0) bash /usr/local/libexec/singularity/helpers/checks/1-hello-world.sh 
START 2-cache-content.py tags[default clean] level[LOW]
PASS: (retval=0) python /usr/local/libexec/singularity/helpers/checks/2-cache-content.py 
```
We can ask for a specific level. In this case I want `--high`, but note that no checks are defined for level `high` and tag `default`:

```
sudo singularity bootstrap --checks --high image.img Singularity 
Sanitizing environment
Docker image path: index.docker.io/library/centos:7
Cache folder set to /root/.singularity/docker
Exploding layer: sha256:7b6bb4652a1be64d39920f7d8c12c5f0bd8abfc10820177109e73817357c07e3.tar.gz
Exploding layer: sha256:60e3079474b1671bcc20b340572ee6cae735ecdd5a13d6686475ca4916eb6003.tar.gz
Adding runscript
Running checks
```

Now let's specify a level that does have checks, to make sure the level flag is working, period.
```
sudo singularity bootstrap --checks --low image.img Singularity 
Sanitizing environment
Docker image path: index.docker.io/library/centos:7
Cache folder set to /root/.singularity/docker
Exploding layer: sha256:7b6bb4652a1be64d39920f7d8c12c5f0bd8abfc10820177109e73817357c07e3.tar.gz
Exploding layer: sha256:60e3079474b1671bcc20b340572ee6cae735ecdd5a13d6686475ca4916eb6003.tar.gz
Adding runscript
Running checks
START 1-hello-world.sh tags[security default] level[LOW]
Hello World, security level 1
PASS: (retval=0) bash /usr/local/libexec/singularity/helpers/checks/1-hello-world.sh 
START 2-cache-content.py tags[default clean] level[LOW]
PASS: (retval=0) python /usr/local/libexec/singularity/helpers/checks/2-cache-content.py 
```
Now let's ask just for tag `clean` which should just run one check:

```
sudo singularity bootstrap --checks --tag clean image.img Singularit
Sanitizing environment
Docker image path: index.docker.io/library/centos:7
Cache folder set to /root/.singularity/docker
Exploding layer: sha256:7b6bb4652a1be64d39920f7d8c12c5f0bd8abfc10820177109e73817357c07e3.tar.gz
Exploding layer: sha256:60e3079474b1671bcc20b340572ee6cae735ecdd5a13d6686475ca4916eb6003.tar.gz
Adding runscript
Running checks
START 2-cache-content.py tags[default clean] level[LOW]
PASS: (retval=0) python /usr/local/libexec/singularity/helpers/checks/2-cache-content.py 
```

Finally, let's try one that requires sudo, which should work fine given we are doing bootstrap:

```
sudo singularity bootstrap --checks --tag docker image.img Singularity 
Sanitizing environment
Docker image path: index.docker.io/library/centos:7
Cache folder set to /root/.singularity/docker
Exploding layer: sha256:7b6bb4652a1be64d39920f7d8c12c5f0bd8abfc10820177109e73817357c07e3.tar.gz
Exploding layer: sha256:60e3079474b1671bcc20b340572ee6cae735ecdd5a13d6686475ca4916eb6003.tar.gz
Adding runscript
Running checks
START 4-docker.py tags[docker] level[LOW]
PROBLEM:  You should not put content in roots home.
RESOLVE:  Install to /opt or /usr/local
anaconda-ks.cfg
original-ks.cfg
.bash_profile
.bash_logout
FAIL: (retval=1)
 python /usr/local/libexec/singularity/helpers/checks/4-docker.py 
```
And this check fails because files were found in root's home! We can of course tweak details of the test, for example there are several profiles that I was allowing in root home (eg, `.bashrc` types but now I'm wondering if that is needed at all? 

There are a ton of different ones I want to write but I should probably stop and let you take a look before this gets out of hand, lol. :P

GodloveD:
> before this gets out of hand

rotfl Wow!  You're on a real tear here!  

This looks really great.  I think I'm getting the idea that these checks are useful for all kinds of things.  We could use them to encourage a bunch of best practices in addition to doing security tests and such.  

One thing I'm thinking is that it might be good to integrate this into the `pull` and `import` commands too.  If I'm a user doing a bootstrap with a definition file than chance are pretty OK that I know whats in the definition file and so I don't necessarily need to do a bunch of checks.  But if I'm doing a `pull`, I might not have any idea what is in the container, so tests may become important.

vsoch:
hmm that's an interesting idea - I guess the assumption is that it's unlikely for the user to pull/import and then use check? What do others think?

vsoch:
I think given that pull/import can happen in a non root space, it wouldn't be the right context to have checks, because likely the user wouldn't be able to do any fixes. I wonder if a suitable place would be in plugins / the registry controlled by admins, so they can essentially perform tests on outgoing images to run (eg, think SLURM jobs) and then choose to not run given a fail?

vsoch:
okay a few tweaks:

 - I made the naming of the files correspond with the level (1-3) so it's easy to visually see looking at the folder
 - I added a new check for bash history and profile files at root's home
 - I made a new tag, "bootstrap" for defaults to occur at bootstrap, including the above check
 - renamed the function to make it a bit shorter :)

gmkurtzer:
Whoa, pep8speaks has lots of feedback!

vsoch:
and here is the view of all checks working: https://asciinema.org/a/129312?speed=3

vsoch:
@gmkurtzer I did this not only for singularity, but also for pydicom this weekend - singularity was a few thousand lines of tweaking, but pydicom was over 15K :)

It was a good weekend!!

gmkurtzer:
This looks good to me, and is a fantastic starting point on checks! Gonna merge. Thank you!

Issue #777, opened 2017-06-28
SLES11 failure due to PID namespace support


AdamSimpson:
## Version of Singularity:

2.3.1

### Expected behavior

Singularity works on Cray's stuck with SLES11

### Actual behavior

Singularity fails under SLES11 due to the following combination of support and features:
```
checking for namespace: CLONE_NEWPID... yes
checking for namespace: CLONE_PID... no
checking for namespace: CLONE_FS... yes
checking for namespace: CLONE_NEWNS... yes
checking for namespace: CLONE_NEWUSER... yes
checking for namespace: CLONE_NEWIPC... yes
checking for feature: NO_NEW_PRIVS... no
```
Although CLONE_NEWPID is detected as available under SLES11, as it is defined and the configure test compile works, it is not supported in unshare() so it fails at runtime. Since NO_NEW_PRIVS is unsupported there is no way, that I can see, to short circuit the code in pid.c using singularity.conf

### Steps to reproduce behavior

Install on SLESS11 and run:
```
$ singularity -vvv --debug exec UbuntuTitan.img ls 
Increasing verbosity level (4)
Enabling debugging
Ending argument loop
Singularity version: 2.3.1-dist
Exec'ing: /usr/lib64/singularity/cli/exec.exec
Evaluating args: 'UbuntuTitan.img ls'
VERBOSE [U=0,P=29379]      message_init()                            Set messagelevel to: 5
VERBOSE [U=0,P=29379]      singularity_config_parse()                Initialize configuration file: /etc/singularity/singularity.conf
DEBUG   [U=0,P=29379]      singularity_config_parse()                Starting parse of configuration file /etc/singularity/singularity.conf
VERBOSE [U=0,P=29379]      singularity_config_parse()                Got config key allow setuid = 'yes'
VERBOSE [U=0,P=29379]      singularity_config_parse()                Got config key allow pid ns = 'no'
VERBOSE [U=0,P=29379]      singularity_config_parse()                Got config key allow user image = 'yes'
VERBOSE [U=0,P=29379]      singularity_config_parse()                Got config key protected image mode = 'none'
VERBOSE [U=0,P=29379]      singularity_config_parse()                Got config key enable overlay = 'yes'
VERBOSE [U=0,P=29379]      singularity_config_parse()                Got config key config passwd = 'no'
VERBOSE [U=0,P=29379]      singularity_config_parse()                Got config key config group = 'yes'
[...]

PR #793, opened 2017-07-05
Use clone as fallback when unshare fails with CLONE_NEWPID

bauerm97:

On some kernel versions, the CLONE_NEWPID flag is available only for `clone` rather than for both `clone` and `unshare`. On these systems, a call to `unshare(CLONE_NEWPID)` causes a runtime failure. This is especially problematic for systems that do not support `PR_SET_NO_NEW_PRIVS`, since those systems must run containers inside of a PID NS (security concerns), and will otherwise fail completely. This PR catches this failure at runtime and falls back to using the `clone` system call when necessary. 

- Ref: #777 

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/src/lib/runtime/ns/pid/pid.c
+++ b/src/lib/runtime/ns/pid/pid.c
@@ -64,8 +64,13 @@ int _singularity_runtime_ns_pid(void) {
     singularity_priv_escalate();
     singularity_message(DEBUG, "Virtualizing PID namespace\n");
     if ( unshare(CLONE_NEWPID) < 0 ) {
-        singularity_message(ERROR, "Could not virtualize PID namespace: %s\n", strerror(errno));
-        ABORT(255);
+        if ( errno == EINVAL ) {
+            singularity_registry_set("NS_CLONE", "1");
+            singularity_message(DEBUG, "unshare(CLONE_NEWPID) unavailable, falling back to clone instead\n");
+        } else {
+            singularity_message(ERROR, "Could not virtualize PID namespace: %s\n", strerror(errno));
+            ABORT(255);
+        }
     }
     singularity_priv_drop();
     enabled = 1;
@@ -91,8 +96,11 @@ int _singularity_runtime_ns_pid(void) {
 
     if ( enabled == 1 ) {
         // PID namespace requires a fork to activate!
-        singularity_fork_run();
-
+        if ( singularity_registry_get("NS_CLONE") ) {
+            singularity_fork_run_ns(CLONE_NEWPID);
+        } else {
+            singularity_fork_run();
+        }
         singularity_registry_set("PIDNS_ENABLED", "1");
     }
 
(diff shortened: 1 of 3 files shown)


gmkurtzer:
Based on a conversation that @bauerm97 and I had yesterday, we are going to move things around a bit more, but will have the same (but cleaner) effect.

@bbockelm I'm mentioning you here as I'd like to have your input on this as it develops.

bbockelm:
There seems to be a lot of copy/pasting in the `singularity_fork_run_ns` -- could that be refactored out (or is that what you meant by the last comment?)?

Other than that, I think the approach looks correct.  Note that there's a relatively small stack size here (32KB).  That's mostly just appropriate for doing a quick `fork`/`exec`: should we bump it up?

bauerm97:
@bbockelm Yes that is exactly what we discussed. We are going to remove the use of `fork()` and replace it with your `fork_ns()` implementation. Greg and I had a question about the current implementation of `singularity_fork()`: is `singularity_fork()` designed to exactly mimic the behavior of the `fork()` system call? Greg was under the impression that it was supposed to, but looking at the implementation it doesn't seem to be doing that as is. Moreover, what exactly is going on in the`do while` loop inside of `singularity_fork()`? It seems that it hangs in this section of code until the child thread exits, why is that? Sorry for all these questions, I'm trying to wrap my head entirely around what's going on there! Thanks!!

bauerm97:
@bbockelm About the stack size, does something like this seem appropriate? Basically I just grab the stack limit and we go from there. That way we don't ever need to worry about it.


```
 struct rlimit stack_limit;

  if ( -1 == getrlimit(RLIMIT_STACK, &stack_limit) ) {
      singularity_message(ERROR, "Unable to get current stack limit: %d\n", strerror(errno));
      ABORT(255);
  }

 int stack_size = stack_limit.rlim_cur;
```

bbockelm:
That would work!

Well, mostly.  The danger here is that glibc will arrange an explicit guard page to protect against things like the Stack Clash family of bugs -- this does not!

That said, even the glibc safeguards are surmountable.  We should probably review all the user inputs and do things like reject invocations where `argv` is larger than 1MB and try to avoid use of things like `alloca`.

bbockelm:
For the question about the `do`/`while` loop: indeed, the idea is to make sure:
a) Parent and child exit at the same time.
b) Make sure signals sent to the child are propagated to the parent and vice-versa.

bauerm97:
@bbockelm I parsed through the do/while yesterday and think I've wrapped my head around it now. I would like `singularity_fork` to behave identically to `fork()`. My plan to accomplish this is still maintain the workflow that you've written in, however I would like to move the idea of the `do/while` loop out to `singularity_fork_run()` [That way fork will return for the child and the parent at the same time]. I plan to still set up the signal handling within `singularity_fork()`, however what one does with that information will become the responsibility of the function that calls `singularity_fork()`. In the case of `singularity_fork_run()`, it will still ensure that your a) and b) points are upheld, however other invocations of `singularity_fork()` might want different behavior for the parent (perhaps the parent will just exit).

bauerm97:
If you have time to hop on the slack channel so I can pick your mind about this a bit, that would be useful :)

bauerm97:
See #803

Issue #794, opened 2017-07-06
Bootstrap: Mistyped Image Error Message


ax3l:
Hi there,

I just got stuck on an unclear error message while bootstrapping a singularity image from a dockerhub image.

## Version of Singularity:

$ singularity --version
2.3.1-dist

on Debian testing

### Expected behavior

Bootstrap from a non-existing docker image of an existing dockerhub account:
```
Bootstrap: docker
From: ax3l/typoseverywhereaaahhhaaands
```

should fail with a clear error message that the base docker image does not exist.

### Actual behavior

When bootstraping a singularity image from docker(hub) and mistyping the image name, the error should be more precise then throwing

```
ERROR Improperly formed manifest, layers or fsLayers must be present
```

([origin source code](https://github.com/singularityware/singularity/blob/master/libexec/python/docker/api.py))

### Steps to reproduce behavior

```bash
$ singularity create myImage.img

$ echo "Bootstrap: docker
From: ax3l/typoseverywhereaaahhhaaands" > Singularity

$ sudo singularity bootstrap myImage.img Singularity 
Sanitizing environment
Building from bootstrap definition recipe
Adding base Singularity environment to container
Docker image path: index.docker.io/ax3l/typoseverywhereaaahhhaaands:latest
ERROR Improperly formed manifest, layers or fsLayers must be present
```

vsoch:
could you show me a link to the image on docker hub?

ax3l:
@vsoch sorry, can't link an image since this issue is about the error message that singularity creates if *no image* is found for an *existing* dockerhub user :)

if a user did a typo like this, singularity does not say "image not found" but the confusing message above :)

Just type this into a bash:
```bash
singularity create myImage.img

echo "Bootstrap: docker
From: ax3l/typoseverywhereaaahhhaaands" > Singularity

sudo singularity bootstrap myImage.img Singularity 
```

(`typoseverywhereaaahhhaaands` is the *non-existing* docker image and `ax3l` is the *existing* dockerhub user)

vsoch:
oh haha, /bonk yes I understand :) I agree we should have a different error message if the image doesn't exist, versus an (actual) properly formatted manifest. I'll put in a fix to this later today!

vsoch:
ok so it looks like the root of the error (when it first triggers) is obtaining the manifest: docker reports the non existing image as needing authentication:

```
DEBUG Headers found: Content-Type,Authorization,Accept
DEBUG Http Error with code 401
[...]

PR #796, opened 2017-07-06
Docker/manifest message fix

vsoch:

Currently, when the user gives a weird and non existing image, the manifest returns with an error message, but until it is used, it is silent. As a result, when the user then tries to pull images (eg with error #794) the message indicates that the manifest is malformed. It is, but we shouldn't have gotten to this point in the first place. To address this, I've added a general function to check the return call of the docker api, and we will exit cleanly with an informative message before getting to the point of issue:

```
singularity pull docker://vanessa/aksfjsfjkjf
ERROR UNAUTHORIZED: authentication required
ERROR Check image existence, capitalization, and permissions.
```
We might need to tweak some tests, we will see :)

Attn: @singularityware-admin


gmkurtzer:
This looks good to me, merging. Thanks!

PR #803, opened 2017-07-11
Fork.c overhaul to use clone() instead of fork()

bauerm97:
As discussed in #777, on some systems a call to `unshare()` will fail in some systems that support `CLONE_NEWPID` only for `clone()`. The original fix ( #793 ) was to simply check if calling `unshare()` failed and attempt to use `clone()` as a fallback. This works as a temporary fix, but there is no reason not to always use `clone()`.

This PR does the following:

1. Implement `fork_ns()` which mimics `fork()` interface but uses `clone()`. Thanks @bbockelm 

2. Add two new functions to install the signal handlers. These are in functions so that in the future we can move to a more "object oriented" forking methodology as discussed w/ Greg.

3. Add new function `wait_child()` which listens on the pipes that were created for the signal handlers to write to, and propagates signals as expected. This function use to be inside of `singularity_fork()`

4. `singularity_fork()` now has the same behavior as `fork()`. Both the parent and child will return before blocking. It is now the parent's responsibility to call `wait_child()` after it calls `singularity_fork()` if it wishes to block for the signals. Further, if `singularity_fork()` is called with escalated privileges, we will not raise and drop privileges before and after the call to `fork_ns()` so that we preserve the ability to call binaries as root (_e.g. create.c_)

5. Update all calls to `singularity_fork*` to match the new interface [takes one additional argument, _unsigned int flags_, which are passed as flags to `clone()`]

6. src/lib/runtime/ns/pid/pid.c is much more streamlined.

- Ref: #777 

**Checkoff for all PRs:**

Attn: 
@AdamSimpson Please ensure that this PR fixes your issue still
@bbockelm Would like your thoughts on the refactoring of fork.c, maybe better stack size code?

### DIFF
--- a/src/copy.c
+++ b/src/copy.c
@@ -92,7 +92,7 @@ int main(int argc, char **argv) {
     }
 
     singularity_priv_escalate();
-    retval = singularity_fork_exec(cp_cmd);
+    retval = singularity_fork_exec(0, cp_cmd);
     singularity_priv_drop();
 
     if ( retval != 0 ) {
--- a/src/create.c
+++ b/src/create.c
@@ -98,7 +98,7 @@ int main(int argc, char **argv) {
 
     singularity_priv_escalate();
     singularity_message(INFO, "Creating file system within image\n");
-    singularity_fork_exec(mkfs_cmd);
+    singularity_fork_exec(0, mkfs_cmd);
     singularity_priv_drop();
 
     singularity_message(INFO, "Image is done: %s\n", image.path);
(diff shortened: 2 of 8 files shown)


bauerm97:
@bbockelm Could you weigh in on why the build was failing when I attempted to allocate the current max stack limit with an out of memory errnor? For now I've switched the stack size to 1MB and hopefully that should be enough, but I'm not sure

gmkurtzer:
This looks fantastic @bauerm97, but I'll wait for feedback from @bbockelm before I do the merge. Also, I'd like to merge locally, and take a closer look.

gmkurtzer:
I am going to go ahead and do this merge. @bbockelm, when you have time, we are still very interested in your feedback and can update if necessary at that point.

Thanks!

PR #805, opened 2017-07-12
Bug fixes

cclerget:
Includes:

- fix registry set key values
- nodev enforcement for mount
- restriction when no_new_privs support

Attn: @singularityware-admin


gmkurtzer:
This is fantastic, thank you @cclerget!

PR #810, opened 2017-07-16
Adding pep8 for Python

vsoch:
This is the first round of work to add pep8 parsing for Python - I haven't finished yet but want to start the PR to configure the pep8speaks (robot) and test out how this works. Carry on, nothing to see here! I'll post more details when this is ready to look at.

It also looks like I had the self bootstrap work in this branch - no worries I can finish that up and include, or just undo it :P  If we get the checks in, the self bootstrap might actually have feet!


vsoch:
look at the robot!! cool!!

okay, I will finish this up tomorrow, need to head home :) The tests will fail because of the bugs the robot found, which I expected and make me very happy :)

vsoch:
Note that there ARE still pep8 issues with the tests, need to fix those.

vsoch:
@gmkurtzer this is ready for review. Specifically, this PR does the following:

- adds a check for the python code conforming to [pep8](https://www.python.org/dev/peps/pep-0008/). This will ensure that code added is styled in a clean and concise way - it's a lot of work to do for an entire repo that already exists, but trivial when you are working on small bits. I just did the first, so things should be relatively easy moving forward!
- I tested calling pep8 in the testing vs using [pep8speaks](https://pep8speaks.com/), and I've decided to choose the second as the first introduces many potential future bugs as the organization / flow of testing changes (for example, the errors above were because it was testing an older version of singularity that hadn't been fixed.) Pep8speaks is much nicer, and tests the exact files that are touched for the specific PR. See above for his robot messages :)
- we don't have a singularity hub import (shub uri) so I removed this test. It doesn't make sense we had it (I think we discussed) and we can re-add when it is implemented.
- This branch (used) to have the self bootstrap, but I removed it to be added with another PR (after checks are implemented).

gmkurtzer:
Wow, this is a huge commit! Looks great, merging!

PR #826, opened 2017-07-23
Infrastructure for Standard Container Integration Format (SCI-F)

vsoch:
Hey @singularityware/singularity-admin ! This PR is being opened to get feedback on the development I'm doing to encourage a modular organizational schema for Singularity. This pull request will add nuts and bolts for Singularity to support the [Standard Container Integration Format](http://containers-ftw.org/SCI-F/), which is going to help us to better develop (internally) modular containers. To understand my thinking about this, read the content at the link first. And then proceed!

First, here is the current status of the PR. The basics to generate the `/apps` and `/data` bases are in place, you can watch a quick [preview of the structure here](https://asciinema.org/a/130145?speed=3).

>> [ removed previous section on absolute path, I figured this out ]

## Next steps
Here are the things I think are in the "TODO":

- [x] decide optimal way to run install commands relative to each application's base (above)
- [x] add equivalent sections for `%applabels`, `%appsetup`, `%appfiles` and `%appenvironment`, 

For setup and files, it would work equivalently to those current sections, but relative to the apps folder.

For labels and environment, these would be specific to an application. The environment variables would be sourced ONLY for the application. There are actually two ways we can go about this - we can store the environment variables and labels within the application folder (meaning they aren't available unless the user specifies/uses the app) or have a strategy that puts them somehow in the main metadata base. Does anyone see a use case for the second? My preference is for the first so an app is completely modular.

- [x] discuss the ideal interactions that a user would have with apps. My suggestions are the following:

### List
To list a container's applications, it makes sense to do an ls

```
singularity ls container.img
   foo
   bar
```

### Run
To run an application, we can do run with the name:

```
singularity run foo container.img
   "RUNNING FOO!"
```
or if we want to keep run separate from running an app, we could also do:

```
singularity apprun foo container.img
   "RUNNING FOO!"
```

If the application has no runscript, the default is to shell into the container directly into the application folder.

### Help
The user should be able to ask for help for an application, an extension of the "base" help command

```
singularity help foo container.img

This is the help for foo!
```

I realize this might be too similar to the main help and/or conflict if there is a command equivalently named to an application, so maybe something like this instead:

```
singularity apphelp foo container.img

This is the help for foo!
```

### Inspect
An inspect should read the labels for an app.

```
singularity inspect foo container.img
{
    "SINGULARITY_DEFFILE_BOOTSTRAP": "docker",
    "SINGULARITY_DEFFILE": "Singularity",
    "SINGULARITY_BOOTSTRAP_VERSION": "2.3.1-add/sci-f.gb0ff1cc",
    "SINGULARITY_DEFFILE_FROM": "ubuntu:latest"
}
```

All of the above should be surprisingly easy to implement, I just want to get feedback, etc. before doing it (and help with the chroot bit would be great!) This is super fun and exciting!! Thanks guys!! :D 

Attn: @singularityware-admin


truatpasteurdotfr:
afaik, realpath is provided by the coreutils package on CentOS/debian/ubuntu
(rpm -qf /usr/bin/realpath or pkg -S /usr/bin/realpath) and busybox on alpine-linux.

vsoch:
okay just fixed the issue with chroot / cd-ing to the correct directory, I understand that the command to chroot assumes the root / and not `singularity_rootfs`, and that anything I printed to the screen would be piped there, so I added the cd to that location via a print:

```
    # To install application, much change to its base and back so paths are relative
    SECTION="'appinstall ${APPNAME}'"
    printf "cd /apps/${APPNAME}\n"
    eval $SINGULARITY_libexecdir/singularity/bin/get-section "$SECTION" "$FILE"
```
and adjusted the call:

```
singularity_app_install_get "${APPNAME}" "$SINGULARITY_BUILDDEF" | chroot "$SINGULARITY_ROOTFS" /bin/sh -e || ABORT 255
```
and that produces the file where I would expect!

vsoch:
I realize that a blank `%appinstall` would mean that nothing happens, and there are two options for dealing with this. We could give a warning, or just tell the user what IS found. The second was easy to implement, so I adjusted the verbosity of the message that tells the user what apps are being installed:

```
Found applications APPNAME to install
```

There are a bunch of weird cases that I'm not sure we want to check for? For example:

```
# produces two different folders, should we enforce lowercase?
# Note this is really easy to do in python (and maybe c?) and more annoying in bash
%appinstall APPNAME vs %appinstall AppName

# This will be ignored, and the user knows when sees is not in the install list
%appinstall
```

vsoch:
@truatpasteurdotfr yeah I just removed it because it wasn't on the computer I'm using this morning - not an issue because it's not needed now that I figured out rootfs!

vsoch:
okay, just added a files section for apps. Adding this:

```
%appfiles APPNAME
manband.jpg
```

will result in the files added, relative to the app:

```
Copying 'manband.jpg' to 'apps/APPNAME'
Skipping checks
$ singularity shell tmp.img 

$ ls /apps/APPNAME/
filey.txt  manband.jpg	runscript.help
```

vsoch:
okay now labels are added, producing a file `labels.json` in the apps folder:

```
%applabels APPNAME
KEY1 VALUE1
AUTHOR VANESSASAUR
AVOCADOS YESPLEASE
```
produces 
```

Singularity tmp.img:~/Desktop> ls /apps/APPNAME/
filey.txt  labels.json	manband.jpg  runscript.help
Singularity tmp.img:~/Desktop> cat /apps/APPNAME/labels.json 
{
    "KEY1": "VALUE1",
    "AVOCADOS": "YESPLEASE",
    "AUTHOR": "VANESSASAUR"
}
```

vsoch:
just added custom environment:

```
%appenv APPNAME
PANCAKES=hello
```
produces

```

$ ls /apps/APPNAME/
environment  filey.txt	labels.json  manband.jpg  runscript.help
$ cat /apps/APPNAME/environment 
PANCAKES=hello
```

Notice that I used `%appenv` instead of `%appenvironment`

vsoch:
I'm going to hold off on adding `%appsetup` for now - with files and appinstall I (think) we might not need it, but we will see :)

vsoch:
The next bit is to add the user interactions, I'll re-iterate from above:

- commands the user should be able to do, and how they source / cd / relate to the app
- if the current setup of having metadata (labels, environment) in the app base folders makes sense (or if we want a metadata folder inside each application, so it's like a mini image)
- what (if any) data about the apps gets saved to `.singularity.d`

I'll hold off on adding more changes until discussion. @gmkurtzer

vsoch:
okay I changed my mind on the file location - I really like the idea of a more consistent singularity metadata folder, within each app (so it's clear what is singularity and what is the app) so I moved the location of app files:

Base app folders:
```

$ tree /apps/
/apps/
|-- ANOTHER
`-- APPNAME
    |-- environment
    |-- filey.txt
    `-- manband.jpg

2 directories, 3 files
```
But notice that environment is at the top level, because as @satra and others would mention, it's easiest to find there. Does this seem reasonable to be different from the root `/.singularity.d`. The other difference is that there is NO singularity (the runscript) file. I don't see the reason to have it there, the one we have now (the link) seems just to maintain backwards compatibility. Maybe others can comment.

Here is that metadata folder within an app, notice that the runscript IS here, along with the help file, clearly indicated to belong to it. I had originally named it runscript.exec to mirror singularity cli files, but didn't think that was necessary.

```
$ tree /apps/APPNAME/.singularity.d/
/apps/APPNAME/.singularity.d/
|-- labels.json
|-- runscript
`-- runscript.help

0 directories, 3 files
```

and finally, here is the data folder, which for now just has nice plugs for input/output for each app installed.
```

$ tree /data/                       
/data/
|-- ANOTHER
|   |-- input
|   `-- output
`-- APPNAME
    |-- input
    `-- output

6 directories, 0 files
```

vsoch:
I think we should even consider adding a `%help` section for an entire container, so this:

```
%help

This is how to run the application!

singularity exec container.img /usr/local/bin/sundaymorning --make pancakes
```
would produce this:

```
/.singularity.d/runscript.help
```
And then we could have some command to for invocation of cat to show that, maybe help for run?

GodloveD:
Wow.  There is an absolute TON of info here.  Still trying to ingest it all. 

So, my overall take is that you are trying to define some standards and best-practices for installing multiple apps within a single container.  And to encourage users to adopt these standards, you are introducing some new sections to the definition files. It's clear that you’ve spent a lot of time on this, and it’s still pretty new to me, so take my criticisms with a grain of salt.  

First, it’s not clear to me why we would encourage users to install more than one app within a container.  Obviously, if an app has dependencies they should all be installed into the same container with the app.  But what is the reason to want more than one app within a container?  To save disk space?  In [your SCI-F doc](http://containers-ftw.org/SCI-F/) you say 

_“For those familiar with container technology, it is commonly the case that an entire container is considered a module.”_

Right.  I don’t really see any arguments put forth on the document as to why this should not be the case.   

Even given that we want to install multiple apps within a single container to make them more modular on the inside, it's not clear to me that an `/apps` directory hierarchy would represent the best-practice for doing so.  Installing software in a non-standard location like `/apps` could make container generation significantly more challenging.  One of the main advantages of containers (from a user’s perspective) is the ability to install software as root, using a package manager, in a standard location.  It's also not clear to me if this approach would simplify using containers or not.  For one thing, the `$PATH` family of variables may need to be tinkered with significantly if you have a lot of software in non-standard locations.  

I *can* see how this may be useful from an admin’s perspective.  If users’ have containers with software in the `/apps` directory hierarchy an admin can easily audit the container and figure out what is in it, how to use it etc.  But only to the extent that users’ buy into this approach and utilize the `/apps` directory.  If you run an audit on a container and you don’t find a particular app does it mean it doesn’t exist?  Or that the user didn’t use the `/apps` directory when they created it?  I’m also not a big fan of making things harder for users so that they are easy for admins.  In general things should be made as easy as possible for users even if it means more work for admins.

I understand that this approach is also supposed to make apps more modular so that they can be used as building blocks to create new containers.  But I also don’t really see how this is going to work.  There is a lot that could go wrong when trying to copy an app from one container to another.  If the containers have an identical operating system OR if the apps are pre-compiled binaries that were built against on old glibc then OK that should work.  But if neither of these conditions are met you are likely going to see lots of seg-faults after copying an app from one container to another.  Another one of the main advantages to using containers is choosing whatever OS is best for the job.  So I don’t think it’s a good idea to lock your containers into a single operating system so that you can reuse modules.  

I also see that the new approach is that users’ will create mini help docs and stick those in the container so that others will see how to use the installed apps.  I have a few problems with this idea.  First, most (good) software already has help docs and man pages.  So I think we should rely on that for the documentation and not what some user decided was the best way to document an app.  Second, how many users are just going to install things to get them working and not worry about documentation?  Most of them.  

To sum all of this up, it basically sounds like we are trying to install Lmod inside of a container and use it to manage apps.  But the point of a container is that you no longer need this type of app management.  

I suppose I could come around on this if I see a compelling use case.  Maybe you could provide some examples?  I should say that the execution of this idea seems great.  I’m just not sure that the idea itself if the best way to go, but I’d like to discuss it further.

vsoch:
>> First, it’s not clear to me why we would encourage users to install more than one app within a container. Obviously, if an app has dependencies they should all be installed into the same container with the app. But what is the reason to want more than one app within a container? 

Imagine an extreme (but actually not so unlikely case). I have two containers, size 4000 each, with a full anaconda suite. Maybe there is a slight difference in packages installed, it is trivial. Each one runs one different python script. This right now is our solution for modularity - it's bloated modularity. But now let's say I ask the user to put both scripts in a container. Now we are putting responsibility / burden on him to write some kind of entrypoint (runscript) to make the distinct functions known and usable. I would argue that's already asking too much. What happens when you have a container like [this one](https://github.com/vsoch/singularity-scientific-example)? It has upwards of 25-50 different genomic tools installed. The best I could do is dump them somewhere and then tell the user to ls and decide based on name. There is no easy way to:

- export / import just one new module and have certainty I'm not going to overwrite something
- bind to some directory for data and be certain I'm not going to overwrite something important
- get help / define a runscript for one particular application
- easily issue one command that I don't need the user to implement in advance (which would vary by container) to understand the software inside.
- it's harder to build workflows/applications around containers because you can't be sure where the same software "starting" or "connectIon" point might be between two different containers.
 - and importantly, many of the future analysis/use cases that the document describes are really challenging if not impossible with the current way things are.

There are much more reasons, but I'll stop there :) I'll also point out that this is by no way required - people can choose to make and use containers like before.

>> “For those familiar with container technology, it is commonly the case that an entire container is considered a module.” Right. I don’t really see any arguments put forth on the document as to why this should not be the case.

and it still could be.

>> Even given that we want to install multiple apps within a single container to make them more modular on the inside, it's not clear to me that an /apps directory hierarchy would represent the best-practice for doing so. Installing software in a non-standard location like /apps could make container generation significantly more challenging. 

This is completely intentional, and in fact @gmkurtzer and I after discussion changed it to be a `/scif/apps` folder to not interfere with any possibly pre-existing locations. 

>> One of the main advantages of containers (from a user’s perspective) is the ability to install software as root, using a package manager, in a standard location.

This would still be the case - there would be the same `%post` section to install "global container" level dependencies.

>>  It's also not clear to me if this approach would simplify using containers or not. For one thing, the $PATH family of variables may need to be tinkered with significantly if you have a lot of software in non-standard locations.

Each container has it's own environment like the main container, so when you ask to run a particular container, you source that.

>> I can see how this may be useful from an admin’s perspective. If users’ have containers with software in the /apps directory hierarchy an admin can easily audit the container and figure out what is in it, how to use it etc. But only to the extent that users’ buy into this approach and utilize the /apps directory.

I have confidence that when something useful and intuitive is provided, usage follows.

>> If you run an audit on a container and you don’t find a particular app does it mean it doesn’t exist?

We would hope we could get to something so transparent.

 >> Or that the user didn’t use the /apps directory when they created it? I’m also not a big fan of making things harder for users so that they are easy for admins. In general things should be made as easy as possible for users even if it means more work for admins.

I'd also like to point out that this is the current way it is now - if you don't know where an executable is in a container, you don't know it exists.

>> I understand that this approach is also supposed to make apps more modular so that they can be used as building blocks to create new containers. But I also don’t really see how this is going to work. There is a lot that could go wrong when trying to copy an app from one container to another.

yes there could, but much less than could go wrong trying to dump an entire container onto a second... *wink*

>> If the containers have an identical operating system OR if the apps are pre-compiled binaries that were built against on old glibc then OK that should work. But if neither of these conditions are met you are likely going to see lots of seg-faults after copying an app from one container to another. 

Yes this is true. Repeat what I said above. However I'll also say that given that each application has a known install script, the import wouldn't be just dumping files (that would be for maybe a data container). It would be like a mini bootstrap to install the software module. Any errors / etc would be revealed just as they are with standard container post.

>> Another one of the main advantages to using containers is choosing whatever OS is best for the job. So I don’t think it’s a good idea to lock your containers into a single operating system so that you can reuse modules.

Actually this approach affords more flexibility with base OS. You create a different container for each base OS (and leave them empty like skeletons) and then you can easily install the same apps on top of the base OS, and do things like test how they are different, which is better. etc.

>> I also see that the new approach is that users’ will create mini help docs and stick those in the container so that others will see how to use the installed apps. I have a few problems with this idea. First, most (good) software already has help docs and man pages. 

We aren't dealing with good software as you describe it, for a lot of times we are dealing with scientific / research software that comes down to some Github scripts and no clear entrypoint. I will also point out that this is entirely optional. It can only help to make it easier for someone that doesn't know how to write a script to spit out help to still do it.

>> So I think we should rely on that for the documentation and not what some user decided was the best way to document an app. Second, how many users are just going to install things to get them working and not worry about documentation? Most of them.

Then I suppose things would just be as they are now? Ha. I have a little more confidence in users, I predict if we give them a `%help` section they would minimally write a sentence.

>> To sum all of this up, it basically sounds like we are trying to install Lmod inside of a container and use it to manage apps. But the point of a container is that you no longer need this type of app management.

The inside of containers is messy. We can do better.

>> I suppose I could come around on this if I see a compelling use case. Maybe you could provide some examples? I should say that the execution of this idea seems great. I’m just not sure that the idea itself if the best way to go, but I’d like to discuss it further.

Glad to hear :)

GodloveD:
Thanks very much for you're reasoned responses.  I see there may be more to the idea than I'm appreciating (containers within containers?).  

I like your example of 25-50 omics apps in a single container.  My solution would be to have 25-50 different containers.  I guess then you might need some way to string them together to make a pipeline.  As you know, I have a strategy for treating apps within containers as though they are installed on bare metal, but maybe this would not be good enough.

If more sophistication is required to make multiple containers work together I would advocate for integration with an orchestration tool like k8s or mesos.  But I'm still trying to figure those things out.

Maybe what you are talking about is to have one master container that has a bunch of sub-containers inside of it?  In that case maybe Singularity gains the capacity to orchestrate itself?  That would be pretty groovy.  

Anyway, looking forward to seeing how this develops (with the reservations I noted above). 😸

vsoch:
yeah !  to speak more generally about this, this opens up a means by which different applications and services inside containers have names that can be referred to.  it's not orchestration in and of itself ( and that is a hairball that I don't want to venture into) but having a simple organizational hierarchy that makes containers predictable and modular. This is the minimum requirement for doing things more complex than just running one container. It  is the minimum requirement for being able to find someone else's container that you didn't make and have a pretty good sense of what's inside.  that one was creating a good base for containers which we've done, and step two is making him easy to integrate into the places they need to be integrated ( orchestration or workflow tools).

 I want to be able to find a mysterious container, with one command see the different applications that are installed, and then with another command run the specific application that I want.

vsoch:
after feedback on the main doc, it was suggested that we provide a size for the different apps in the container, and so I added the following:

 - calculate of size for each app after the install (and added as a label)
 - addition of the %appinstall section as the defile for each app
 - adding the app name as the minimal label for it.

The above all happen for the app install, which is akin to post and (imho) the minimum base for any actual app. So as a quick review, here is what a folder looks like (without showing the hidden metadata folder):

```
Singularity ubuntu.img:~> tree /scif/
/scif/
|-- apps
|   |-- bar
|   |   `-- filebar.exec
|   `-- foo
|       `-- filebar.exec
`-- data
    |-- bar
    |   |-- input
    |   `-- output
    `-- foo
        |-- input
        `-- output
```

Here is metadata for an app:

```
tree /scif/apps/foo/.singularity.d/
/scif/apps/foo/.singularity.d/
|-- Singularity
|-- labels.json
|-- runscript
`-- runscript.help
```
What you can't see in the above is that the runscript is executable. Labels has the size and name, and would also have any additional labels provided by the user under `%applabels foo`

```
{
    "SINGULARITY_APP_NAME": "foo",
    "SINGULARITY_APP_SIZE": "1MB"
}
```
the deffile `Singularity` has the `%appinstall` section, so if someone wanted to create the app in a different image, they could just run it:

```
cat /scif/apps/foo/.singularity.d/Singularity 
%appinstall foo

echo "INSTALLING FOO"
touch filebar.exec
```
Right now I'm including the %appinstall line so the user could just plop it into another build recipe, but I'd be open to opinions about whether it should/shouldn't be.

satra:
@vsoch  - doesn't this assume that everything about the app `foo` is in this tree location

```
/scif/
|-- apps
|   `-- foo
```
the point of compiling many apps inside a single container is to leverage shared libraries, much like docker run shares layers. this seems to assume that apps are static and only exist inside that folder.

vsoch:
no it doesn't at all, I expect there to be install commands that install some library dependencies to system locations, or locations shared by many apps. What would exist in the folder are files and maybe local dependencies that aren't from package managers, etc, and are specific to the app.

satra:
then i'm confused about how an app will be relocated from one container to another.

vsoch:
because you have the `%appinstall` and the folder with whatever files were added from the user's computer / elsewhere!

vsoch:
You have the full installation script:

```
Singularity ubuntu.img:~> cat /scif/apps/foo/.singularity.d/Singularity 
%appinstall foo

echo "INSTALLING FOO"
touch filebar.exec
```

all metadata (labels)

```
/scif/apps/foo/.singularity.d/
|-- Singularity
|-- labels.json
|-- runscript
`-- runscript.help
```
including a calculated size of the final thing..

```
Singularity ubuntu.img:~> cat /scif/apps/foo/.singularity.d/labels.json 
{
    "APP_NAME": "foo",
    "APP_SIZE": "1MB"
}
```

and if I had defined environment, it would be there too :)

vsoch:
it's like a **tiny** container inside a larger container! Contained reproducibility, but still modular :)

satra:
i think this may require a live conversation or an example. 

let's say i compile something like ANTs in shared mode through bootstrap. this creates 50 binaries. now i want to make each one an app. i'm not sure i can see how this will be done and especially the information retained so that i can now take one of these apps/binaries and copy it into another container.

vsoch:
I would say that a suite of tools like ants is considered one app, granted they are compiled together. If you wanted to expose them all, then you would have something like the following:

The post still installs system libraries and shared software. There is no distinction between this and how it's currently done other than now I can more easily see the things (as determined by the writer of the recipe) that are needed more globally.

```
%post
apt-get update
```
I would even err on the side of redundancy to make sure that each apps install script has all dependencies defined. I'll chat more about this later.

The application install would then be relative to the ants apps folder at `/scif/apps/ants`. The user doesn't need to know this, he just writes the steps to do it assuming the right spot:

```
%appinstall ants
apt-get install -y cmake git
git clone https://github.com/stnava/ANTs.git
mkdir bin
cd ANTs
make 
cp Scripts/* ../bin/
```

(note that I didn't try the above, just guesstimating based on looking over the docs quickly). Then we would have environment variables that we only care about when we need ants

```
%appenv ants
ANTSPATH=${APPBASE}/bin
export PATH=${ANTSPATH}:$PATH
```
and actually now that I think of it, we should provide the user with some environment variable that would reference his/her application base, so he/she could write something like the above and not:

```
%appenv ants
ANTSPATH=/scif/apps/ants/bin
export PATH=${ANTSPATH}:$PATH
```

And then extra tidbits, if you are a detailed developer, would be help and labels:

```
%apphelp ants
AnTS is a series of binaries to help with registration. Find them at $ANTSHOME 

%applabels ants
MAINTAINER satra
FROMFUTURE "import satra"
```

So now how does the application get run? The user could shell into the image, but ask for it in the context of the app:

```
singularity shell ants ubuntu.img 

Singularity satra.img:/scif/apps/ants> 
```

If I were the developer, I would make a runscript that easily exposes the different ants executables. Maybe like this:

```
    if [ $# -eq 0 ]; then
        echo "\nThe following software is installed in this image:"
        ls -1 $ANTSHOME/bin   
        echo "Example usage: container.img run ants [command] [args] [options]"  
    else
        exec "$@"
    fi
```
so if the user called this (and we can talk with @gmkurtzer and others about how the ordering should go, since run doesn't usually have additional args):

```
singularity run ants container.img
```
It would spit out a message that says "here are the executables served by ants" and then the user would do like:

```
singularity run ants container.img antsRegistration --arg1
```
And you can imagine the same would work for exec, putting it in the context of the app, but do whatever the user wants.

```
singularity exec ants container.img echo $ANTSHOME
```

For the other functions (import, export, etc.) I am waiting to chat with @gmkurtzer on some of the details, but the general idea I think makes sense thus far!

So to copy the ants app into another folder, there are two chunks of things that we need. If the user had files added from somewhere on a host (and they are then in the container) we would have rationale to move the folder first. But given the potential for needing to recompile, we would then want to run the %appinstall (or some derivation of that). Yes, it's the case that if you have a yum command it's not going to be well suited for ubuntu, but this leads to another idea --- including some metadata about the requirements of the host for the app, which can be derived from the container during bootstrap time. So hypothetically if you had some app to export and let's say you tried to import into a container with the wrong OS:

```
                       [ import ]     [ from ]        [ to ]
sudo singularity copy ants ubuntu.img centos.img

Checking requirements for ants from centos.img...
ERROR: centos host is not compatible with debian.
```

(again, the syntax / ordering / terms of that might be wrong, this is just for example)

And now let's imagine we don't have to care much about the base container, if I'm just interested in sharing ants for different environments I could just write different app recipes for it, and let the user import / use the one right for him/her. The different chunks would be OS specific (the examples below are different files):

```
# Singularity.app.ants:centos
%appinstall ants
yum install git

# Singularity.app.ants:ubuntu
%appinstall ants
apt-get install git
```

and now even cooler! Imagine that somewhere like Singularity Hub has a registration for apps. So we would do like:

```
singularity importapp container.img shub://vsoch/ants:centos
```

yeah that's pretty cool :) I'm not sure if I answered your question, but is this making sense?

gmkurtzer:
This is fantastic stuff!!! I only have a few points to share:

1. I don't think we should confuse apps as containers, as they are not portable. The recipe for generating them "might" be portable, but not guaranteed (e.g. apt-get Vs. yum for dependencies).
2. The motivation for using `.singularity.d/` within the app is clear, but I don't like the hidden directory. I'd prefer this just sitting in the root of the app folder.
3. I think it is reasonable for `%appinstall FOO` to not exist, without throwing a warning, if `%apprun FOO` exists. That is because the %apprun could just be a workflow script that doesn't install anything.
4. I think we need to rename `%post`. It isn't post, and as a matter of fact, it should be run before any `%appinstall` sections.

Again, great work and thanks everyone for the feedback!

Greg

vsoch:
>> I don't think we should confuse apps as containers, as they are not portable. The recipe for generating them "might" be portable, but not guaranteed (e.g. apt-get Vs. yum for dependencies).

Agreed! I wrote a new section in the specification describing this. There are some best practices we can encourage for making them more portable (including dependency installs in `%appinstall` and then checks to avoid mismatches (eg, checking the host where the bootstrap was initially done and saving this as a label to check when the app is exported).

>> The motivation for using .singularity.d/ within the app is clear, but I don't like the hidden directory. I'd prefer this just sitting in the root of the app folder.

Hmm should we have a different kind of app metadata folder? Just `singularity.d` without the hiddenness? (@satra I think you should be gleaming right now, because you were heavily against the hidden folder!)

>> I think it is reasonable for %appinstall FOO to not exist, without throwing a warning, if %apprun FOO exists. That is because the %apprun could just be a workflow script that doesn't install anything.

Agreed! I'll do that.

>> I think we need to rename %post. It isn't post, and as a matter of fact, it should be run before any %appinstall sections.

I noticed in the code a `%pre` section that doesn't seem to be known/used - could we advocate for that instead, and post is really "after all the other stuff?"

vsoch:
The recent update changed the metadata folder to `scif` instead of `singularity.d` per request of @gmkurtzer ! I'd like to next chat about what kind of actions we want for apps, here is something to start us off:

# How do we list?

## Idea 1: use list
This will currently conflict with @bauerm97 list command, and my hope is that we can have `singularity ps` be for listing processes (like Docker, compose, etc) and `singularity list` to see applications available (the more common of the two use cases).
 
```
singularity list container.img
   foo
   bar
```
## Idea 1: apps == list of apps
If we cannot use list, then we can have a general command `apps` that does the same:

```
singularity apps container.img
   foo
   bar
 ```

# How do we run?
Run equates to executing the runscript associated with a particular application, and we have a few options.

## Idea 1: Use "app" to specify an app, followed by command
In this idea, we have the equivalent set of commands (eg `shell`, `run`, `exec`) and we add a new command for each that is prefixed with app (`apprun`, `appshell`). I'm not a huge fan (it looks kind of silly) but it's an option:

```
singularity apprun foo container.img
```

## Idea 2: just have app == run
The other idea goes along with the `singularity apps` --> list all idea, that if a user does `singularity app`  we expect the name of the app to come after, followed by an argument like run shell, etc. In this case, running the app would look like:

```
singularity app run foo container.img
```
and if we can re-arrange the arguments a bit, it sort of reads better to have one of the following:

```
singularity app run container.img foo
# or
singularity app container.img run foo
```
Do you see a more intuitive way to have commands be targeted to an app?

## Idea 3: Add another argument to run, if present, specifies an app
This is a slight derivation of the above, and I find it more intuitive, and possibly it fits better into the current implementation. Since we don't have anything that normally comes after the container name for run, we could have an app there:

```
singularity run container.img foo
```
For all of the above, whatever we decide, you can image similar for shell, exec, etc.

# How do we represent apps?
It should be the case that I can have a definition for an app in a file, for example here is a file called `Singulapp` (lol, I just came up with that! It's terrible).

```
%appinstall foo
 git clone ...
 cd ..
 make

%app labels
```
In this way, I could just have a folder of text files, one per app, to install into containers when needed. As an admin on a cluster, I could let users choose some grouping of applications to install into containers, and then generate them for my users. I could also have a shared repository of these files (eg, served by singularity hub or similar) to install on the fly

# How do we install?

Imagine a folder structure like this (thanks to @GodloveD this will actually be possible!):

```
/bases
   ubuntu.img
   centos.img
```

It should be the case that I can have one or more of these files, and then apply them to all of my bases. It's not really a bootstrap because I'm not using the app recipe as a base. It's not really an import because I'm running commands and doing more than dumping files. Thus, I want to propose introducing a new command, `install`. It would work like this:

```
sudo singularity install ubuntu.img Singulapp
Installing foo into ubuntu.img...
```
or we could have multiple
```
sudo singularity install ubuntu.img Singulapp.foo Singulapp.bar
Installing foo, bar into ubuntu.img
```

and then we could install by way of moving apps between containers. This would be the same flow as install, except we get the app metadata and recipe from the container:

```
sudo singularity install  [ source ]  [ destination ] [ app ]
sudo singularity install ubuntu.img centos.img foo 

# or again, multiple
sudo singularity install ubuntu.img centos.img foo bar
```
There are a few issues here. For an install from a file, we don't have information about if the OS will work with the app. In this case, if there is an error we will ERROR and exit with 255. It's the same as it would be now if a user issued a yum command in a debian base without it. In the other case of installing an app from one container to another, we get the bonus of having information about the container it is coming from. We can do some matching of OS as a rule, and my suggestion is the following:

 - on original install, save the container OS into the labels metadata for the application
 - on install [container1] --> [container2], match the OS. If it's good, let the user continue
 - if the OS is mismatched, exit with a warning. Tell the user to put --force to force the install

I'll stop there for now to get feedback. :)  I'd like to discuss this soon so I can start implementing what we decide!

**Comment by gmkurtzer** [comment]
Great stuff... in general, I want to stay with the standard convention of `singularity $VERB $CONTAINER $OPTS`, so I would do:

To list the apps within a container:
```
$ singularity apps container.img
```

To run a particular app within a container:
```
$ singularity apprun congtainer.img APPNAME
```

For bootstrapping, I would use the following section names:
```
%appinstall foo
...
%applabels foo
...
%apphelp foo
...
%apprun foo
...
```

**Comment by vsoch** [comment]
Okay going to keep a list here of the additions to come:

- [x] list apps with `singularity apps container.img`
- [x] run apps with `singularity apprun container.img foo`

## List Apps

Container with apps:
```
singularity apps ubuntu.img
total 8.0K
4.0K bar
4.0K foo
```
Container without apps:
```
singularity apps dino.img
No installed apps.
```
I decided to add the size of each, because I think that could be useful for the user.

image

**Comment by vsoch** [comment]
Okay, will fix the pep8 errors asap, but here is what is now added! The user can do `apprun` to run a specific application's runscript. This means that the application's environment is also sourced. Execution works as follows:

- user calls apprun with a named application
```
singularity apprun container.img foo
```
- if the container singularity version has support for apps, it executes the action. If not, exit 255
- the apprun action script checks for the application. 
  - If installed, it looks for an environment
    - if environment is found, it is sourced. 
  - If not installed, the user is shelled to the location of the application. This is akin to the assumption that a user probably wants to work with an app if it doesn't have a runscript, and wants to be in its `$PWD`

Here is a quick demo!
https://asciinema.org/a/131331?speed=3

**Comment by vsoch** [comment]
@gmkurtzer what is your preference for the following:

 - getting help for an app:
```
singularity apphelp container.img foo
```
 - should we have a shell into an app?
```
singularity appshell container.img foo
```
For some of these commands like shell, exec, it might make sense to have an `--app` flag to trigger the function vs. an equivalent `appshell`, `appexec`, etc. What do you think?

```
singularity shell --app foo container.img
singularity exec --app bar container.img
```

**Comment by gmkurtzer** [comment]
agreed on all points!

**Comment by vsoch** [comment]
@gmkurtzer what was your preference for the above? The two options are:

 - an app[cmd] equivalent for each current command (eg, appshell, appexec)
 - a flag --app that gets added to any command to do the same!

eg:

```
singularity shell --app foo container.img
singularity appshell container.img foo
```

I sort of like the first better (but your preference for apprun makes me think you might prefer the second!)

**Comment by vsoch** [comment]
Preview of exec, shell, and inspect: https://asciinema.org/a/132652?speed=3

File: libexec/bootstrap-scripts/deffile-sections.sh [hunk 1]
```diff
@@ -216,6 +224,138 @@ else
...
 
 
+##########################################################################################
+#
+# APP SECTIONS
+#
+##########################################################################################
+
+### APPINSTALL
+if [ -z "${SINGULARITY_BUILDSECTION:-}" -o "${SINGULARITY_BUILDSECTION:-}" == "appinstall" ]; then
```
gmkurtzer, line 234: It isn't clear to me how this will work with multiple %app* sections. I will test shortly.

libexec/cli/apprun.help
```diff
@@ -0,0 +1,51 @@
+USAGE: singularity [...] run [run options...] <container path> [...]
```
gmkurtzer: This looks like the help file for `singularity run ...`?

libexec/cli/apps.help
```diff
@@ -0,0 +1,14 @@
+USAGE: singularity [...] apps <container path>
+
+List applications (apps) installed in a container, located at
+    /scif/apps. See http://containers-ftw.org/SCI-F/
+
+EXAMPLES:
+    
+    $ singularity apps ubuntu.img
```
gmkurtzer, line 8: Can you show some example output of this command too?

libexec/cli/shell.exec
```diff
@@ -62,6 +59,14 @@ if [ -z "${SINGULARITY_IMAGE:-}" ]; then
     exit 0
 fi
 
+# Does the user want to target an application? 
+# shell --> appshell
+if [ ! -z "${SINGULARITY_APPNAME:-}" ]; then
```
gmkurtzer, line 64: How does this work in the context of a `singularity shell ...`?

src/action-lib/appshell.c
```diff
@@ -0,0 +1,59 @@
+/*
```
gmkurtzer, line 1: I still don't exactly understand how an app has it's own shell. Can you explain?

libexec/cli/apprun.help
vsoch: ah yes I will fix!

libexec/cli/apps.help
vsoch: Sure

libexec/cli/shell.exec
vsoch: it shells right into the app folder! If a user wants to work there, inspect, etc.

gmkurtzer:
I just watched your Asciinema, does the --app option to shell just dump you in that directory with a given shell? If so, there maybe a much simpler way to do it, but just setting the starting directory (but that option should also be created as it will be useful and clean the code considerably).

src/action-lib/appshell.c
vsoch: It's the same shell, just to a specific location!

vsoch:
I don't totally understand, but if you have a better way to implement, definitely is ok with me!

libexec/bootstrap-scripts/deffile-sections.sh
vsoch: It gives you the names for all the apps, which are then parsed over.

gmkurtzer:
Merged, awesome, rockin', and FANTASTIC! Great work!

vsoch:
woohoo!! I'll open some TODO issues for this, and get started!

Issue #811, opened 2017-07-17
Updating SLRUM plugin to v2.3.1 Singularity API


votavap:
## Version of Singularity:

2.3.1

### Expected behavior

Running integration with SLURM, such as:
`srun --singularity-image=<image> [ params ]`

### Actual behavior
While the plugin loads ok, at runtime, it can't resolve some of the symbols that are referenced in the shared library and exists with:

`slurmstepd: [17.0]: symbol lookup error: /usr/local/lib/slurm/singularity.so: undefined symbol: singularity_rootfs_init`

There seem to be number of calls within src/slurm/singularity.c that reference functions that were changed in v2.3 of Singularity and as a result can't be resolved at runtime through the shared libraries.

### Steps to reproduce behavior

- Build Singularity with --with-slurm option
- Try to run any singularity image with SLURM:
`srun --singularity-image=<image> [ params ]`

Tested with Singularity 2.3.1 (devel) and SLURM 17.02.6

jscook2345:
@votavap 

Sounds like this was already fixed and merged. Thanks!

PR #827, opened 2017-07-23
Fix for Singularity plugin for SLURM

votavap:

This should fix the Singularity plugin for SLURM. I have built it on CentOS7 with gcc 4.8.5 and it passed initial tests with SLURM 17.02.6

I have tested this separately from the "make test", because it requires SLURM up and running. There were no changes to the singularity core, only the plugin code under src/slurm.

- Ref: #811, #701  

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -28,3 +28,4 @@
     - Cedric Clerget <cedric.clerget@univ-fcomte.fr>
     - Rémy Dernat <remy.dernat@umontpellier.fr>
     - Mark Egan-Fuller <markeganfuller@googlemail.com>
+    - Petr Votava <votava.petr@gene.com>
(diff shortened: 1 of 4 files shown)


bauerm97:
What is the "command" variable being used for in the new code? The concept of "command" as I understand it in the current code is which action we want to take (shell, exec, or run). However, in this context there is no command (this function is simply setting up the process to be in the container environment). After everything is set up, the function just returns instead of executing the "command"

votavap:
@bauerm97 it is not used for anything right now, it was part of my checks. However I was hoping to possibly use it in the next iteration to distinguish between "run" and "exec". Currently on return from the function, slurm executes `execve()`, so pretty much the same as what gets executed in `src/action-lib/exec.c` without the initial checks. This means that there is currently no way to execute the "run" command (except as `singularity exec <image> /singularity [args]`) and I was hoping to add it there later on and distinguish between those two.

bauerm97:
@votavap Right, the reason that you can't do "run" or "exec" with the SLURM plugin is because the SLURM plugin is only intended to put the functions calling process into a container. What is to be executed inside the container is supposed to be left up to the rest of the SLURM submission script. Therefore I think that the concept of "run", "exec", "shell" are not meaningful in this context.

mbrt:
This one works for me. The build went fine and I did a quick test with slurm 17.02.2.

votavap:
@mbrt Thank you for testing this.

gmkurtzer:
This looks good to me, I'm going to merge this in. Thanks @votavap!

Issue #833, opened 2017-07-26
Support for bootstrapping from a local image (feature request)


GodloveD:
## Version of Singularity:

2.3

### Suggested behavior

It would be nice to be able to have one or a few images saved on your local machine that could be used as templates for bootstrapping images, in the same way that we are currently using Docker Hub containers as templates for bootstrapping.  You can already do this if you like, but it is not very clean.  I propose officially supporting this by creating a new bootstrap driver called local image.  I've already written the code to do this and will submit a pull request shortly.

vsoch:
This will be possible with apps! See https://github.com/singularityware/singularity/pull/826

vsoch:
oh I am reading the end now - awesome! Can't wait to see it.

vsoch:
could we hook the shub endpoint in as well? So it would be the same flow but first pull the shub image. How are you handling merging metadata?

GodloveD:
You are reading my mind @vsoch!  shub is the next thing I want to work on.  

Actually I didn't think about the meta-data.  It's probably getting clobbered by the second bootstrap right now.  I'll have to have a look and see.  I'm going to go ahead and post the PR in a few minutes and then maybe we can talk more about it.

vsoch:
yes we can't add something to the software that encourages bad practice for reproducibility. The clobbering must be resolved in some fashion!

shahzebsiddiqui:
can we get shub bootstrap option in singularity. 

Currently it is not supported
```
[root@amrndhl1295 ~]# singularity bootstrap /dev/shm/test.img test.def 
Sanitizing environment
Building from bootstrap definition recipe
Adding base Singularity environment to container
ERROR: 'Bootstrap' type not supported: shub

[root@amrndhl1295 ~]# cat test.def 
BootStrap: shub
From: shahzebsiddiqui/singularity-examples:singularity-2.3.1

%post



```

vsoch:
the singularity fairies are working on this as you speak :)

GodloveD:
https://yt3.ggpht.com/-zvlOt3MimBY/AAAAAAAAAAI/AAAAAAAAAAA/KtdVvraKp1Y/s900-c-k-no-mo-rj-c0xffffff/photo.jpg

vsoch:
how did you know!

GodloveD:
This is all finished as far as I know.  So I'm going to close now.  Re-open if necessary.  Thanks!

PR #834, opened 2017-07-26
Bootstrap from a local image

GodloveD:

This adds a new feature: official support for bootstrapping from a local image.  

Right now, you can do something like this in your definition file:
```
Bootstrap: docker
From: ubuntu

%post
    apt-get install some good stuff
```

But if you want to make a custom container to use as a base for bootstrapping, you have to write a Docker File (yuck).  

I propose a new ability to pull from a local image.  It works like this:

```
Bootstrap: localimage
From: /my/great/base.img

%post
    apt-get install some good stuff
```

- Ref: #833 

**Checkoff for all PRs:**

- ✅  I have read the [Guidelines for Contributing](https://github.com/singularityware/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- ✅  I have tested this PR locally with a `make test`
- ✅  This PR is NOT against the project's master branch
- ✅  I have added myself as a contributor to the [Author's file](https://github.com/singularityware/singularity/blob/master/AUTHORS.md)
- ✅  This PR is ready for review and/or merge

Attn: @singularityware-admin


libexec/bootstrap-scripts/deffile-driver-localimage.sh
```diff
@@ -0,0 +1,55 @@
...
+fi
+
+if [ ! -f "${FROM:-}" ]; then
+    message ERROR "${FROM} does not exist\n"
+    exit 1
+fi
+
+if [ -z "${SINGULARITY_BUILDDEF:-}" ]; then
+    exit
+fi
```
vsoch: There is no exit code or message here?

libexec/bootstrap-scripts/deffile-driver-localimage.sh
```diff
@@ -0,0 +1,55 @@
...
+
+########## BEGIN BOOTSTRAP SCRIPT ##########
+
+umask 0002
+
+message 1 "Exporting contents of ${FROM} to ${SINGULARITY_IMAGE}\n"
+
+cmd="${SINGULARITY_libexecdir}/singularity/bin/export ${FROM} | (cd ${SINGULARITY_ROOTFS} && tar xBf -)"
+if ! eval $cmd; then
+    message ERROR "Failed to export contents of ${FROM} to ${SINGULARITY_ROOTFS}\n"
```
vsoch, line 52: How does it change the execution flow to cd to the `SINGULARITY_ROOTFS` but not back?

vsoch:
This is awesome! I am very hesitant to allow for something like this before we have nicely integrated more modular applications, because likely people will dump containers on top of one another and then be confused why their software doesn't work. Some quick feedback:

- any reason to call it bootstrap from an `image` ? Will there be a different kind of local <thing> or image to warrant the longer string? I would just say:

```
Boostrap: image
From: /path/to/image.img
```

- essential for this to go through is a way to deal with the metadata. We have one of two options. 

We can either add more files, named to reference the imported image (and proceed making metadata for the image as useful) but this gets messy fast as you add more images. What I think needs to happen is an intelligent combination of metadata, and it's up to the user to decide order of operations. For example, if I have two container environments:

```
%environment (image base)
VARIABLE=ONE

%environment (image dump)
VARIABLE=TWO
```
Given that I am adding "dump" onto "base," this means `VARIABLE=TWO` is going to be sourced later and thus take dominance. I think what we would need, in order for this to work, is the following:

- reading of old metadata files and then handing to python to add to new
- an option for each of environment, labels, and runscript to disable any or all of those three things, again, at the decision of the user to lose information about his/her image
- warnings to the best that we can to alert a user when fields are over-written. This will already happen with python and labels (it issues a warning for fields overwritten) but we don't have a good way for the environment. Perhaps we print for the user the environment of the original image, and then show the new image, so they can check?

This should be pretty easy to add `shub://` bootstrap for, it would look like this:

```
Bootstrap: shub
From: vsoch/hello-world
```
and it would just come down to pulling the image first, and then going forward with that flow. I can code this up when this PR is ready to go!

Finally, I want to suggest that we think about doing this in the context of apps. When we are dumping a modular app that carries its own metadata/runscript, etc, all of these issues with over-writing stuff just go away. It's a way to support having base images and adding content onto them without also encouraging bad practice.  Let's see what @gmkurtzer has to say - he's been super busy and hasn't had a chance to look at all these PRs!

libexec/bootstrap-scripts/deffile-driver-localimage.sh
GodloveD: The `cd` happens in a subshell or something (don't know if that is the right terminology).  So the `$PWD` before and after the command are identical.

libexec/bootstrap-scripts/deffile-driver-localimage.sh
GodloveD: I copied this script from `deffile-driver-debootstrap.sh` which has the same check also lacking error message or exit code.  

It looks like it is actually possible to run a `bootstrap` without a definition file.  (See [this line in init.c](https://github.com/singularityware/singularity/blob/development/src/bootstrap-lib/init.c#L75) ). For instance, the following works:
```
singularity create lolcow.img
sudo singularity bootstrap lolcow.img docker://godlovedc/lolcow
```
I don't know if this helps to enable that or if it is just an error.  

I actually don't see how the code would ever get into `deffile-driver-localimage.sh` (or `deffile-driver-debootstrap.sh` for that matter) without a definition file because it has to parse the `Bootstrap` field to figure out which script to use.  So this check *may* be safely deleted.  But should probably check with @gmkurtzer to be sure.

GodloveD:
Great news!  The singularity devs are actually pretty smart folks! 😜 None of the env vars, labels, or other metadata gets clobbered with this approach.  Even if you add more metadata on top it will just add to what is already existing (with the exception of runscript which will be preserved if a new one is not written and overwritten by the second bootstrap if it includes a runscript.)

I'm guessing all of this was fixed when the option was added to a bootstrap from a docker file and that those changes also protect the metadata here.

GodloveD:
@vsoch inlining your message below so I can talk about each point in turn:
>This is awesome! 

Thank you.
>I am very hesitant to allow for something like this before we have nicely integrated more modular applications, because likely people will dump containers on top of >one another and then be confused why their software doesn't work. 

I think that is a problem for users to solve themselves.  I am extremely not a fan of restricting the power of users because we feel like they might be able to hurt themselves.  I also see this as orthogonal to the modular apps framework that your working on.  I can think of several important use cases for this that do not include installing more than one app in a container.  For instance:

I need to install 10 apps on a cluster.  Each one has 5 versions compiled with different toolchains.  So I need to install 50 apps with 5 different compiler toolchains.  Things like `gcc` and `mpi` can take a long time to compile.  (Like around an hour.)  If I have to compile `gcc` and `mpi` within each one of my containers before I compile my apps that's an additional 50 hours of bootstrap time.  But if I can create 5 local images that have all of the compiler toolchains I need (let's say that takes 5 hours) and use those as base images for my apps I save my self 45 hours of bootstrap time (a whole workweek plus 5 extra hours to drink :beer:)

Or let's say I'm troubleshooting a bootstrap script that starts with a pull of tensorflow from docker hub.  Every time I make change I have to wait for all of the layers to explode and whatnot.  Or I can just pull it down once and then copy it locally so that every time I make a change I just import via tar pipe at the beginning of my script.  That's much faster.  Then when my bootstrap script is finished I re-write the header to pull from Docker Hub again to make it portable again.  

>Some quick feedback:
>
>any reason to call it bootstrap from an image ? Will there be a different kind of local or image to warrant the longer string? I would just say:
>Boostrap: image
>From: /path/to/image.img

I like `localimage` better because technically bootstrapping from Docker Hub (and shub in the future) is also bootstrapping from an "image".  So I think `localimage` is less ambiguous.  
>essential for this to go through is a way to deal with the metadata. We have one of two options.
>We can either add more files, named to reference the imported image (and proceed making metadata for the image as useful) but this gets messy fast as you add more images. What I think needs to happen is an intelligent combination of metadata, and it's up to the user to decide order of operations. For example, if I have two container environments:
>
>%environment (image base)
>VARIABLE=ONE
>
>%environment (image dump)
>VARIABLE=TWO
>Given that I am adding "dump" onto "base," this means VARIABLE=TWO is going to be sourced later and thus take dominance. I think what we would need, in order for this to work, is the following:
>
>reading of old metadata files and then handing to python to add to new
>an option for each of environment, labels, and runscript to disable any or all of those three things, again, at the decision of the user to lose information about his/her image
>warnings to the best that we can to alert a user when fields are over-written. This will already happen with python and labels (it issues a warning for fields overwritten) but we >don't have a good way for the environment. Perhaps we print for the user the environment of the original image, and then show the new image, so they can check?

You're right, a user could shoot themselves in the foot by overwriting some variables that existed in the previous image when they do the second bootstrap.  This could become tricky too because we now have several different methods of adding variables, and the variables will take different precedence depending on how they are added.  Variables added with the new `$SINGULARITY_ENVIRONMENT` variable will take precedence over those added using the `%environment` section because they are written to `91-environment.sh`.  If the first image had variables added with `$SINGULARITY_ENVIRONMENT` and the user tried to overwrite the values with the `%environment` section in the second bootstrap they would be in for a nasty surprise.  I can try catting out the environment before and after but that doesn't mean the user would know how to fix a problem if they encountered one.  

As an aside @gmkurtzer you suggested that the `$SINGULARITY_ENVIRONMENT` variable should point to a new script in `/.singularity.d/env` rather than writing to `90-environment.sh`.  What was your reasoning on that?  Do you still think it's a good idea considering the previous paragraph?
>This should be pretty easy to add shub:// bootstrap for, it would look like this:
>
>Bootstrap: shub
>From: vsoch/hello-world
>and it would just come down to pulling the image first, and then going forward with that flow. I can code this up when this PR is ready to go!
>

I agree!  And I'm actually happy to do it too.
>Finally, I want to suggest that we think about doing this in the context of apps. When we are dumping a modular app that carries its own metadata/runscript, etc, all of these issues with over-writing stuff just go away. It's a way to support having base images and adding content onto them without also encouraging bad practice. Let's see what @gmkurtzer has to say - he's been super busy and hasn't had a chance to look at all these PRs!

Once again, I think these two ideas can exist in parallel, and I think that each has value apart from the other.  I think we should do both.

libexec/bootstrap-scripts/deffile-driver-localimage.sh
gmkurtzer: This conditional exists for legacy reasons, and can probably be removed (in both sources) and/or abort. This check occurs earlier and will trigger the `main-null.sh` invoker instead of `main-deffile.sh` (meaning that this driver should never get invoked).

libexec/bootstrap-scripts/deffile-driver-localimage.sh
gmkurtzer: @GodloveD Yes, this script is being as a subshell of the parent via `eval_abort`, so the current directory that this process is running within will not affect the calling script.

libexec/bootstrap-scripts/deffile-driver-localimage.sh
GodloveD: Cool.  That's what I figured.  I'll delete it.

libexec/bootstrap-scripts/deffile-driver-localimage.sh
GodloveD: Right on.  Also I think the way that the `cd` command is getting called it is also starting it's own subshell so that even within the script if you echoed $PWD before and after the tar command it will return the same.

gmkurtzer: Ahh, yes. You are right.

vsoch: ahh this makes sense! Thanks for clarification.

vsoch:
@GodloveD I don't think this is accurate (although maybe we can make it so with some tweaking):

>> Great news! The singularity devs are actually pretty smart folks! :stuck_out_tongue_winking_eye: None of the env vars, labels, or other metadata gets clobbered with this approach. Even if you add more metadata on top it will just add to what is already existing (with the exception of runscript which will be preserved if a new one is not written and overwritten by the second bootstrap if it includes a runscript.)

I will review the different files in detail.

 - `runscript` The issue with the runscript would be addressed by having apps with their own. But arguably in this case of using an entire image, we should take a similar approach that we do with Docker. If there exists a runscript already in the container, use it. If not, check if it's defined in the build spec. If nothing, then shell. By knowingly defining a new one, the user would be accepting responsibility that he/she is essentially deleting an old script.
- `deffile` The other forgotten piece is the original bootstrap recipe, which is stored at `cat /.singularity.d/Singularity` right now, that gets clobbered. So if I bootstrap a complicated image with just one simple change to write a new runscript, the entire history of the image generation is lost.
- `labels` : The labels are stored at `/.singularity.d/labels.json` A docker bootstrap means that they are first generated in memory, put into a tarball, and dumped into the image. This means, for certain, that anything there gets wiped. The reason one would possibly see labels only being added is in the case that the Docker image doesn't have labels (this is the much more common of the two) so the new labels are just added to the file.
- `env`: This one might actually be handled, given that @gmkurtzer has some logic to generate a differently numbered file.

>> I'm guessing all of this was fixed when the option was added to a bootstrap from a docker file and that those changes also protect the metadata here.

Sorry I don't think that's the case :(

vsoch:
>> I think that is a problem for users to solve themselves. I am extremely not a fan of restricting the power of users because we feel like they might be able to hurt themselves. 

I think software development is like any consumer product. If you make something that encourages a bad practice that it easy to do, by way of it being easy (and even outside of awareness) it's going to be done.  And then if someone mentions it ("hey, this container has no history, there is software installed not in the deffile, what did you do?") .. the blame is placed on the tool "Oh I just imported an image, it's **Singularity** that did that."  We have to assume that scientists aren't going to think twice about anything, so if they can import an image they will, even if it breaks reproducibility. In that we can't have control over what users do, but we can have somewhat control over what they are encouraged to do by way of giving them a tool to do it, we have to develop by a high standard.

>> I also see this as orthogonal to the modular apps framework that your working on. I can think of several important use cases for this that do not include installing more than one app in a container. For instance

I definitely agree after reading your use case! I would say the two meet overlapping goals, but there is a good region of difference in the sets.

>> I like localimage better because technically bootstrapping from Docker Hub (and shub in the future) is also bootstrapping from an "image". So I think localimage is less ambiguous.

Should we just call it local then? localimage seems too specific, but that's just an opinion.

>> You're right, a user could shoot themselves in the foot by overwriting some variables that existed in the previous image when they do the second bootstrap. This could become tricky too because we now have several different methods of adding variables, and the variables will take different precedence depending on how they are added. Variables added with the new $SINGULARITY_ENVIRONMENT variable will take precedence over those added using the %environment section because they are written to 91-environment.sh. If the first image had variables added with $SINGULARITY_ENVIRONMENT and the user tried to overwrite the values with the %environment section in the second bootstrap they would be in for a nasty surprise. I can try catting out the environment before and after but that doesn't mean the user would know how to fix a problem if they encountered one.

Exactly, this is my biggest concern. 

>> I agree! And I'm actually happy to do it too.

eeee :D 

>> Once again, I think these two ideas can exist in parallel, and I think that each has value apart from the other. I think we should do both.

+1

GodloveD:
@vsoch I hate to disagree with your disagreement, but I have to disagree.  😸 As I stated before, env vars, labels, and the runscript are already being handled the way you suggest.  Proof!

https://asciinema.org/a/kwTkmmF3AGb4dJ4McDjuKBp74

You are right that the forgotten piece of metadata is `/.singularity.d/Singularity`.   It's replaced with the latest definition file.  

I suppose that I could copy the old  `/.singularity.d/Singularity` file to `/.singularity.d/Singularity0`, and if 0 exists I could copy it to `/.singularity.d/Singularity1` and so on so that you always have a list of all the definition files that were used to produce the final product.  I actually think that is a pretty good idea and I'll work to implement it.  

I'm also working to implement a message that happens at bootstrap to tell users how their environment has changed when they bootstrap from an existing image.

I think we should keep the bootstrap agent name `localimage` because `local` is also ambiguous.  It could be misconstrued as `self`.

vsoch:
@GodloveD your test isn't capturing the case I am describing. You need to bootstrap from a docker image that has labels in the `Dockerfile`

I'm not sure about handling the Deffile - @gmkurtzer what do you think?

I see the rationale for localimage now! It makes sense, +1 :)

GodloveD:
I'm don't understand why it matters where the labels came from in the first image.  They are still stored in a .json in the first image and will be added to by the second image right?  Can you make an asciinema that shows the test you are describing?

vsoch:
ohh hold the phone, I don't know the use case I'm describing :OP The one I'm thinking of is two bootstraps from docker, anything the involves docker layers will produce an in tar memory, and then dump over the labels in the image. With a localimage bootstrap, of course we don't do this, we just update the labels.json. So I take it back ! No issue, carry on :)

vsoch:
I think we might consider having a deffile subfolder, and then Singularity outside as the main one. It could get messy on the upper level (and confusing) with too many SingularityN files.

vsoch:
@GodloveD how are we capturing (inside the image) that one of the included Singularity files was imported via localimage? What if we have some future way of interacting / adding /subtracting images, how would we label that, and then reproduce the entire flow?

GodloveD:
@vsoch The labels capture the bootstrap driver.  In this case:

```
SINGULARITY_DEFFILE_BOOTSTRAP": "localimage",
```

I think that is a good idea to make a subdirectory with all of the definition files.  The only missing piece is support for the `inspect` command.  It would be nice if `inspect` told you that there was more than one definition file.  Maybe it could check for the existence of the subdirectory and then print all of the .def files out in order?

vsoch:
I can do that! Let me know when the branch is ready and I can PR to your PR, lol.

GodloveD:
@vsoch I've been trying to print a message notifying the user of changes in the environment after the container bootstraps from a local image as you suggested.  It is surprisingly not easy!  I wonder if you could lend me a hand or point me in the right direction.  

Expanding on my `light.img` `dark.img` asciinema example above, I would like to do something like the following.  Here is a dumb little script called `env-diff.sh`:
```sh
#!/bin/sh
# compare environment of image1 and image2 and print variables that are unique
# to image1
IMAGE1=${1:-}
IMAGE2=${2:-}

BEFORE=$(mktemp) && AFTER=$(mktemp)
singularity exec $1 printenv | sort >$BEFORE
singularity exec $2 printenv | sort >$AFTER
REMOVED=$(comm -23 ${BEFORE} ${AFTER})
ADDED=$(comm -13 ${BEFORE} ${AFTER})
rm $BEFORE $AFTER

echo ""
echo "Environment variables were added, removed, and/or changed during bootstrap."

echo ""
echo "Variables unique to original image (${IMAGE1})"
echo "$REMOVED"

echo ""
echo "Variables unique to new image (${IMAGE2})"
echo "$ADDED"
```
If I run this on the images from the asciinema example I get this:
```
$ ./env-diff.sh light.img dark.img

Environment variables were added, removed, and/or changed during bootstrap.

Variables unique to original image (light.img)
SINGULARITY_CONTAINER=light.img
SINGULARITY_NAME=light.img

Variables unique to new image (dark.img)
SINGULARITY_CONTAINER=dark.img
SINGULARITY_NAME=dark.img
VADER=dark_side
```
Fine.  I would like to do something like this at the conclusion of bootstrap.  But it is surprisingly difficult (at least for me) to figure out how to do this.  First I tried putting some code at the end of [post.sh](https://github.com/singularityware/singularity/blob/development/libexec/bootstrap-scripts/post.sh) or [main-deffile.sh](https://github.com/singularityware/singularity/blob/development/libexec/bootstrap-scripts/main-deffile.sh) but the container is in a pretty weird state during bootstrap so you can't just step into it and ask what it's environment looks like (let alone query a second container at that point).  

Then I thought about creating a `check` using your new infrastructure.  But even then the environment is missing some crucial components (like the full path of the image and the full path of the image that was used to perform the bootstrap).  

I'm not sure how much of the code I want to touch just to print out a few helpful messages at the end of bootstrap.  Do you have any good ideas for how this could be accomplished simply?

vsoch:
yeah! Make it a check?

vsoch:
And for references to the images, we should minimally have something available in the metadata folder that you could parse from labels. We are going to soon hopefully merge the addition on label schema labels, and I know minimally that has the name of the def file. You could also do some kind of trick where you define and print references to the image before, and then refer to the numbers/reference after. The user would likely remember or just need to scroll up if not!

GodloveD:
Sounds easy enough.  But the metadata folder only has the image name and the name of the image it was bootstrapped from.  Not absolute paths which I think I need to get this working.  Like I said I've been trying to make this a check but I don't see a sane way to do it.

vsoch:
ahh I see. I think the script should not be called from the images, but from the internal variables like `SINGULARITY_ROOTFS`. If one image is dumped into the other, there must be some point in that evocation pathway where you have access to both images, and would be able to peek inside the metadata folder to compare the environments. It probably won't work with the printev /env command because the image isn't alive, so you would do it statically.

GodloveD:
@vsoch I think this adds both of the features you wanted.  

First, if it finds a file at `/.singularity.d/Singularity` it will create a new directory (`/.singularity.d/Singularity/bootstrap_history` if it doesn't exist) and move `/.singularity.d/Singularity` to `/.singularity.d/Singularity/bootstrap_history/Singularity0`.  If that exists it will move it to `/.singularity.d/Singularity/bootstrap_history/Singularity1` and so on.

Second, Singularity will now give some messages to the user if the environment changes.  Check this asciicast for details on how that works.   

https://asciinema.org/a/oxuF1wV9AXRn9mLtMU4ZK5lVS

libexec/bootstrap-scripts/deffile-driver-localimage.sh
```diff
@@ -0,0 +1,56 @@
...
+if [ -z "${SINGULARITY_ROOTFS:-}" ]; then
+    message ERROR "Singularity root file system not defined\n"
+    exit 1
+fi
+
+if [ -z "${FROM:-}" ]; then
+    message ERROR "Required Definition tag 'From:' not defined.\n"
+    exit 1
+fi
+
```
vsoch: can we make this an elseif?

libexec/bootstrap-scripts/deffile-sections.sh
```diff
@@ -226,5 +226,18 @@ if [ -s "$SINGULARITY_ROOTFS/.singularity.d/runscript" ]; then
     chmod 0755 "$SINGULARITY_ROOTFS/.singularity.d/runscript"
 fi
 
-# Copy the runscript into the container
+# Copy the definition file into the container.  If one already exists, archive.
+if [ -f "$SINGULARITY_ROOTFS/.singularity.d/Singularity" ]; then
+    message 1 "Found an existing definition file\n"
+    message 1 "Adding a bootstrap_history directory\n"
```
vsoch, line 232: There should be a check here to see if it exists already, in the case we have multiple layers of bootstrapped images!

libexec/bootstrap-scripts/deffile-sections.sh
```diff
@@ -226,5 +226,18 @@ if [ -s "$SINGULARITY_ROOTFS/.singularity.d/runscript" ]; then
...
+    mkdir -p "$SINGULARITY_ROOTFS/.singularity.d/bootstrap_history"
+    ct=0
+    while true; do
```
vsoch: This is going to be super annoying, but let's write out `count` instead of `ct`, I get shivers when I see weirdly named variables :shark:

libexec/bootstrap-scripts/main-deffile.sh
```diff
@@ -53,9 +53,23 @@ if [ -n "${BOOTSTRAP:-}" -a -z "${SINGULARITY_BUILDNOBASE:-}" ]; then
     fi
 fi
 
+# take a snapshot of the environment for later comparison
+if [ ${BOOTSTRAP} = "localimage" ]; then
+    export SINGULARITY_STARTING_ENVIRONMENT=$(eval_abort env -i ${SINGULARITY_libexecdir}/singularity/helpers/record-env.sh ${SINGULARITY_ROOTFS})
+    export SINGULARITY_STARTING_ENVSHA1=$(eval_abort sha1sum ${SINGULARITY_ROOTFS}/.singularity.d/env/*.sh | sha1sum)
+fi
```
vsoch: I think we want to avoid what @gmkurtzer  calls bashisms? So here first define the two variables, then export (especially because this is run as shell right?)

```
if [ ${BOOTSTRAP} = "localimage" ]; then
   SINGULARITY_STARTING_ENVIRONMENT=$(eval_abort env -i ${SINGULARITY_libexecdir}/singularity/helpers/record-env.sh ${SINGULARITY_ROOTFS})
   SINGULARITY_STARTING_ENVSHA1=$(eval_abort sha1sum ${SINGULARITY_ROOTFS}/.singularity.d/env/*.sh | sha1sum)
   export SINGULARITY_STARTING_ENVIRONMENT SINGULARITY_STARTING_ENVSHA1
fi 
```

vsoch: this is cool! So the first command is the environment (keys and values) and the second is a hash for it? Can you show me what the two variables look like?

vsoch: oh and I think we would want to do `${BOOTSTRAP:-}` so if it's not set the default is empty. I don't totally get it but I see @gmkurtzer  do it and he's bash shell god so I figure it's the right thing to do :) http://tldp.org/LDP/abs/html/parameter-substitution.html

libexec/bootstrap-scripts/main-deffile.sh
```diff
@@ -53,9 +53,23 @@ if [ -n "${BOOTSTRAP:-}" -a -z "${SINGULARITY_BUILDNOBASE:-}" ]; then
...
 eval_abort "$SINGULARITY_libexecdir/singularity/bootstrap-scripts/deffile-sections.sh"
 eval_abort "$SINGULARITY_libexecdir/singularity/bootstrap-scripts/post.sh"
 
+# take another snapshot and compare to see what changed
+if [ ${BOOTSTRAP} = "localimage" ]; then
+    export SINGULARITY_ENDING_ENVIRONMENT=$(eval_abort env -i ${SINGULARITY_libexecdir}/singularity/helpers/record-env.sh ${SINGULARITY_ROOTFS})
+    export SINGULARITY_ENDING_ENVSHA1=$(eval_abort sha1sum ${SINGULARITY_ROOTFS}/.singularity.d/env/*.sh | sha1sum)
+    compare_envs
+    rm $SINGULARITY_STARTING_ENVIRONMENT $SINGULARITY_ENDING_ENVIRONMENT
+fi
```
vsoch: ditto here for the exports and variable names.

libexec/functions
```diff
@@ -310,13 +310,40 @@ parse_opts() {
 }
 
 
+compare_envs() {
+    # during bootstrap, if we are bootstrapping from a local image, compare the
+    # environment in the new image to that in the old image and alert the user
+    # of any changes
+    REMOVED=$(comm -23 ${SINGULARITY_STARTING_ENVIRONMENT} ${SINGULARITY_ENDING_ENVIRONMENT})
+    ADDED=$(comm -13 ${SINGULARITY_STARTING_ENVIRONMENT} ${SINGULARITY_ENDING_ENVIRONMENT})
+
```
vsoch, line 319: ooh tell me what comm does?

libexec/functions
```diff
@@ -310,13 +310,40 @@ parse_opts() {
...
 
 
+compare_envs() {
+    # during bootstrap, if we are bootstrapping from a local image, compare the
+    # environment in the new image to that in the old image and alert the user
+    # of any changes
+    REMOVED=$(comm -23 ${SINGULARITY_STARTING_ENVIRONMENT} ${SINGULARITY_ENDING_ENVIRONMENT})
+    ADDED=$(comm -13 ${SINGULARITY_STARTING_ENVIRONMENT} ${SINGULARITY_ENDING_ENVIRONMENT})
+     
+    if [ -n "${REMOVED}" -o -n "${ADDED}" ]; then
```
vsoch: `${ADDED:-}` again I think

libexec/helpers/record-env.sh
```diff
@@ -0,0 +1,25 @@
...
+#
+
+# when calling this script, please use the following syntax to clean env
+# env -i ./record-env.sh $SINGULARITY_ROOTFS
+# because env should be clean upon entry, no sanity checking or functions
+SINGULARITY_ROOTFS=${1:-}
+for file in $(ls ${SINGULARITY_ROOTFS}/.singularity.d/env/*.sh); do
+    . ${file} > /dev/null 2>&1
+done
+tmpfile=$(mktemp)
```
vsoch, line 23: does this get cleaned up?

vsoch: maybe follow convention for other times we use `mktemp` something like `mktemp /tmp/.singularity-envcompare.XXXXXXXX`

vsoch:
hey @GodloveD ! Apologies for the slow reply, I'm a bit underwater with work to do (and I shouldn't even be breaking now!) I think the changes are great, and it satisfies the need to preserve the older environment and spec file. The one detail I am curious about is how this is intended to work given the new squashfs format? Since we won't have the equivalent of bootstrap / import, the user wouldn't ever be able to create an image and then add another one to it, it would have to be a "sandbox" image and then I suppose another one?

GodloveD:
Hi @vsoch.  Just because you did not reply in the first 5 minutes after I posted this does not mean you were slow! 😄 

I'm glad that you like the changes!  We'll see what @gmkurtzer thinks.  Some of this may be a little unorthodox?  But I couldn't find another way to do it.

As for how this will work with squashfs, I think that's a question for @gmkurtzer.  But I think that as long as the `export` command doesn't go away this should still work with squashfs (I think)?

vsoch:
I think the export bit makes sense, I'm just not sure about the import. You would only be able to import at image creation, in which case that would be like copying the image?

I'm lost in the land of dicom headers.... **tiny voice helppmeeee**

libexec/bootstrap-scripts/deffile-sections.sh
GodloveD: `-p` basically does that for you.  Not only will it create the full path if it doesn't already exist but it will refrain from creating the directory if it does exist.

libexec/bootstrap-scripts/deffile-sections.sh
GodloveD: At least I didn't name it `i` 😜 No you're right.  I'm all about verbosity in naming vars.

libexec/bootstrap-scripts/main-deffile.sh
GodloveD: So `$SINGULARITY_STARTING_ENVIRONMENT` is actually going to be the name of a temp file.  That temp file will contain the output of a `printenv` (basically the contents of the environment after sourcing everything in `/.singularity.d/env`.  And then yes `$SINGULARITY_STARTING_ENVSHA1` is just going to contain a hash.

I'm happy to export on a different line and give the vars default of empty, but I also don't know why.  😸

libexec/functions
GodloveD: `comm` is like `diff` but better.  It gives you 3 columns.  The first column has lines unique to file one, second column has files unique to file 2, and 3rd column has lines common to both.  The `-23` and `-13` omit the specified columns from the output.  I could just call it once and parse the output, but I opted to call it twice and not parse anything.

Also there is a caveat that the files must be sorted.  This thing was written by Stallman, so it's old (and GNU) and I think (hope) it should be available most everywhere.

libexec/helpers/record-env.sh
GodloveD: Yes it gets cleaned up in `main-deffile.sh`.  The relevant line is:
```
rm $SINGULARITY_STARTING_ENVIRONMENT $SINGULARITY_ENDING_ENVIRONMENT
```
I _can_ do that.  But I'm actually not crazy about hardcoding `/tmp`.  What if the user has `$TMPDIR` set to `/scratch` or something.  Would we want to respect their wishes or just stick it in `/tmp` anyway?

libexec/bootstrap-scripts/deffile-driver-localimage.sh
GodloveD: sure

gmkurtzer:
This looks fantastic, great work! About the only feature that I would suggest (for another PR) is that it should be able to obtain a remote image as well (e.g. SHub or tucked away on some HTTPD/FTPD server). Thanks, merging!

PR #836, opened 2017-07-26
Add/bootstrap minsize check

ityonemo:

Implements a "minsize" keyword for def file bootstraps which will guard against trying to use an image that isn't big enough to fit the intended contents.  It's still up to the def file author to make sure that the reported minsize is correct.  def files which don't implement minsize are still supported.

- Ref: # https://github.com/singularityware/singularity/issues/646

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/libexec/bootstrap-scripts/main-deffile.sh
+++ b/libexec/bootstrap-scripts/main-deffile.sh
@@ -1,25 +1,25 @@
 #!/bin/bash
-# 
+#
 # Copyright (c) 2017, SingularityWare, LLC. All rights reserved.
 #
 # Copyright (c) 2015-2017, Gregory M. Kurtzer. All rights reserved.
-# 
+#
 # Copyright (c) 2016-2017, The Regents of the University of California,
 # through Lawrence Berkeley National Laboratory (subject to receipt of any
 # required approvals from the U.S. Dept. of Energy).  All rights reserved.
-# 
+#
 # This software is licensed under a customized 3-clause BSD license.  Please
 # consult LICENSE file distributed with the sources of this project regarding
 # your rights to use or distribute this software.
-# 
+#
 # NOTICE.  This Software was developed under funding from the U.S. Department of
 # Energy and the U.S. Government consequently retains certain rights. As such,
 # the U.S. Government has been granted for itself and others acting on its
 # behalf a paid-up, nonexclusive, irrevocable, worldwide license in the Software
 # to reproduce, distribute copies to the public, prepare derivative works, and
-# perform publicly and display publicly, and to permit other to do so. 
-# 
-# 
+# perform publicly and display publicly, and to permit other to do so.
+#
+#
 
 ## Basic sanity
 if [ -z "$SINGULARITY_libexecdir" ]; then
@@ -38,6 +38,24 @@ fi
 if [ -z "${SINGULARITY_ROOTFS:-}" ]; then
     message ERROR "Singularity root file system not defined\n"
     exit 1
+fi
+
+## For deffile bootstraps, we can implement a "minsize" guard.  This triggers a
+## size check against the image we're trying to bootstrap into, so we don't
+## waste time building into an image file that will be too small.
+#this environment variable might not be defined, so we can make it null.
+MINSIZE=${MINSIZE:-}
+if [ -n "$MINSIZE" ]; then
+  #report that it's been included in the def file.
+  echo "Minimum image size has been set: $MINSIZE M"
+  #get the size of the image, in megabytes.
+  IMAGE_SIZE=$(du --apparent-size -m $SINGULARITY_IMAGE | cut -f 1)
+  echo "Detected size of the image: $IMAGE_SIZE"
+  #do the check.
+  if [ "$IMAGE_SIZE" -lt "$MINSIZE" ]; then
+    message ERROR "Size of $SINGULARITY_IMAGE insufficient to accomodate bootstrap minimum $MINSIZE M\n"
+    exit 1
+  fi
 fi
 
 
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -29,3 +29,4 @@
     - Rémy Dernat <remy.dernat@umontpellier.fr>
     - Mark Egan-Fuller <markeganfuller@googlemail.com>
     - Petr Votava <votava.petr@gene.com>
+    - Isaac Yonemoto <isaac.yonemoto@gmail.com>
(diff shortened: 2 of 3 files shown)


vsoch:
hey @ityonemo ! This needs to go against the development branch.  One of the boxes that you checked is not accurate

ityonemo:
Aw man, I had to leave the page and GitHub left the box checked and didn't
persist the branch selection :(

ityonemo:
ok @vsoch should be fixed.

vsoch:
hey @ityonemo ! Apologies for the delay - your PR made me realize we were a bit behind with regard to some changes to master (see #837 linked above). Once that goes through I'll give your PR a first review. I think this is a great idea - and it's a function that has a lot of use cases. For example:

 - calculating the size of a folder (an application) within the image, [see here](https://github.com/singularityware/singularity/pull/826#issuecomment-318227839)

and specifically, I use the function like this:

```
APPFOLDER_SIZE=$(singularity_calculate_size "${APPBASE}")
```

and actually the function looks very similar to your one liner:

```
singularity_calculate_size() {
     FOLDER="${1:-}"
     if [ ! -d "${FOLDER}" ]; then
         message ERROR "Folder not found ($FOLDER)\n"
     fi
 
     eval du -sm ${FOLDER} | cut -f1    # MB
 
     return 0
 }
```
Just curious - I am reading that `--apparent-size` doesn't take blocks into account, and this might imply that using blocks could overestimate the size. Would it be more safe to actually overestimate to give a little wiggle room? Do all distributions support that flag?
 
I like this idea because it is a helpful bit for the user to generate an image. It might even make sense to add the final (actual size) somewhere, maybe as a label for inspection?

vsoch:
ok, now try rebase with development. That support file should go away (I hope, lol)

libexec/bootstrap-scripts/main-deffile.sh
```diff
@@ -40,6 +40,24 @@ if [ -z "${SINGULARITY_ROOTFS:-}" ]; then
...
 fi
 
+## For deffile bootstraps, we can implement a "minsize" guard.  This triggers a
+## size check against the image we're trying to bootstrap into, so we don't
+## waste time building into an image file that will be too small.
+#this environment variable might not be defined, so we can make it null.
+MINSIZE=${MINSIZE:-}
+if [ -n "$MINSIZE" ]; then
+  #report that it's been included in the def file.
+  echo "Minimum image size has been set: $MINSIZE M"
```
bauerm97, line 50: I think you should be using message 1 instead of echo here, I could be wrong. Something like this:

`message 1 "Minimum image size has been set: $MINSIZE M\n"` 

Same as in the detected size of image line a few lines down.

vsoch: +1

vsoch: 1 is considered an info message (so it always prints unless the user specifies quiet), Should the size be more of a debug, or is it important to always show? Here are the levels: https://github.com/singularityware/singularity/blob/master/libexec/functions#L75. I would say maybe 2?

bauerm97:
@vsoch I worked with Isaac to fix the issue where there were stray commits attempting to be merged in. So we should be good to merge once this looks good.

libexec/bootstrap-scripts/main-deffile.sh
bauerm97: +1 agreed. Good catch. Maybe the "Minimum image size has been set" line should be 1, but the second line with the detected size line should be 2? I think that it might be important for the user to see the information that there is a Minsize set (maybe?)

libexec/bootstrap-scripts/main-deffile.sh
```diff
@@ -40,6 +40,24 @@ if [ -z "${SINGULARITY_ROOTFS:-}" ]; then
...
+  #get the size of the image, in megabytes.
+  IMAGE_SIZE=$(du --apparent-size -m $SINGULARITY_IMAGE | cut -f 1)
+  echo "Detected size of the image: $IMAGE_SIZE"
```
vsoch, line 53: @ityonemo could you comment on the choice to use --apparent-size? Just curious, I've never used it before. Can you confirm it has support for all old, crusty operating systems?

gmkurtzer:
Heya everyone, is this PR ready for my review?

Thanks!

vsoch:
Should the minsize be added as a label? Imagine at some future point that the image inspect contents are being served at an endpoint (as you would get now with singularity inspect --json container.img) and primarily the user might want to know what size the image is. If we have a minsize, that's a nifty helper.

vsoch:
Actually, why aren't we calculating the final size of the image after bootstrap? It seems like a no brainer!

gmkurtzer:
Yep, agreed, and added as a label.

vsoch:
I can add this to my PR.

vsoch:
I'm testing this, and I'm getting a list of sizes instead of one size (at least for my use case).

vsoch:
okay, so the one bug we are going to run into here is if the user is bootstrapping a folder. Without the `sm` (to indicate summary) arg, it will spit out a size for every file. It works above because we just have one image file, but I expect without the `s ` we will have a user bootstrap into a folder and get something along the lines of what I just got

vsoch:
So it needs to be something like:

```
IMAGE_SIZE=$(du --apparent-size -sm $SINGULARITY_IMAGE | cut -f 1)
```

vsoch:
okay, it's added to PR #843 

```
{
    "org.label-schema.usage.singularity.deffile.bootstrap": "docker",
    "MAINTAINER": "Vanessasaurus",
    "org.label-schema.usage.singularity.deffile": "Singularity.help",
    "org.label-schema.usage": "/.singularity.d/runscript.help",
    "org.label-schema.schema-version": "1.0",
    "org.label-schema.usage.singularity.deffile.from": "ubuntu:latest",
    "org.label-schema.build-date": "2017-07-28T22:59:17-04:00",
    "org.label-schema.usage.singularity.runscript.help": "/.singularity.d/runscript.help",
    "org.label-schema.usage.singularity.version": "2.3.1-add/label-schema.g00f040f",
    "org.label-schema.build-size": "715MB"
}
```

libexec/bootstrap-scripts/main-deffile.sh
gmkurtzer: I would also like for a check here. If the `du` fails, we need to be able to catch the error and handle in a logical/sane way.

gmkurtzer:
@ityonemo I am closing this PR as the new `singularity build` command which is superseding `singularity bootstrap` no longer requires a container size. Thank you, and please send us more PRs!

PR #837, opened 2017-07-27
Update Development with docs on master

vsoch:
I realized after seeing #836 that we need to update development with some tweaks that were done to master (so it's not so strange to see what we see in #836 - someone starting off of master base to PR to development, and then having extra bits). This quick PR should resolve that - just adding the SUPPORT.md

Attn: @singularityware-admin


gmkurtzer:
Looks great!

PR #838, opened 2017-07-27
Remove /boot from being mounted when hostfs=yes

### LINKED ISSUE #832: Mount hostfs bind mounts /boot
## Version of Singularity: 2.3.1

When configured with 'mount hostfs = true', containers bind mount /boot. This behavior is consistent with the 'mount hostfs' description: "Probe for all mounted file systems that are mounted on the host, and bind those into the container" This is not done with other local mounts on the hosts, e.g. /var and /var/log. It seems that Singularity ignores these.

Depending on what's installed into the mounted container, this can clobber your boot configuration.

### Expected behavior

I expect Singularity to ignore /boot when bind mounting file systems. I don't see the use in mounting standard system directories like /boot into a container

### Actual behavior

Singularity bind mounts /boot into the container while ignoring other standard directories like /var and /var/log.

Install Singularity 2.3.1.
Configure 'mount hostfs = yes' in singularity conf file
Open a shell into a container
Observe your local machines /boot file system mounted inside your container.

vsoch:
This pull request will add a line to prevent /boot from being mounted when mount hostfs = yes, and will close #832. If there is reason to automatically mount /boot we can chat about it, and then perhaps have another setting to enable mounting it. It seems unlikely that we would need it, since there isn't really a need to "boot" a container like a computer :)

Attn: @singularityware-admin


gmkurtzer:
Awesome!

Issue #830, opened 2017-07-25
[ CINSTANCE ] feedback


vsoch:
Omg I'm so excited about this! This is an issue board to post feedback for the daemon, so we can put stuff in one place while testing it out! I'll try to break things into topics as best as I can. 

## Starting the daemon
First, I was able to start a daemon ok - I had some weird issue with a singularity executable in `/usr/bin` (and I'm not sure how it got there...) but likely is me doing something stupid along the way, so let's assume it's not related to this! Here is the part of my script that creates the daemon, upon registry init:

```
singularity start /usr/local/libexec/sregistry/singularity.registry services
/usr/local/libexec/sregistry/singularity.registry  [66306.8915744]  Daemons:
     Daemon PID:   6272
     Daemon Name:  services
```
Note - the caller of the script was sudo, but I'm wondering if it makes a difference for a user vs. sudo to call singularity start?

I think it would be good to have some kind of feedback that the daemon was started, because I saw a little mounty icon on my sidebar (something ubuntu does) but there wasn't output in the terminal. 

## A Likely use case
I don't have idea about how we could (or even would want to) change this, but here is what I think will be  a common use case for service type daemons - to check if it's running, and only continue if it is:

```
    if singularity list | grep "Daemon Name:  services"; then
        "Registry services daemon is already running."
    else
        singularity start "$SINGULARITY_REGISTRY_MANAGER" services
    fi
```
It's pretty easy to do with grep, but do you think we should make it easier for the user? Something like:

```
singularity list services
6272
```
and it would return the PID or nothing (if services doesn't exist).

## Having trouble with bind
To get to the file I need (outside the image) I needed to do a bind, and I was able to figure out a workaround (calling something already in the image) but I'll leave this general question here - 
are binds allowed / still working when we join? The issue is that if I'm executing a command to the image, it needs to be to run a script that is either already in the image, or bound to it. In this case, these "helper / database" scripts for my application (that sometimes need write) are stored in the installation directory (/usr/local/libexec) and so we need the bind. Is the idea that this is default behavior because the start script must already be in the image (and not bound from outside) and instead of doing a `--join` at some point I can just define a `%startscript` in the bootstrap recipe?

vsoch:
eee! Nevermind the last one, I just added it to the code base already installed in the image. Daemon!!

```
Registry services daemon is already running.
server started, status code 0
worker started, status code 0
```
Also, those print lines are all from my scripts - we should have something added to the daemon so he announces himself!

[...]

PR #839, opened 2017-07-27
Singularity service support

bauerm97:

Currently, Singularity only properly supports running jobs for scientific workloads (container terminates after execution finishes). This PR is the basis for supporting services in addition to scientific workloads using singularity. For example, if a user wants to run a persistent web server, database, etc... this will enable them to do so. 

=========================================================================

New commands introduced:

- `singularity [...] start [start options...] <container path> <instance name>`: Start a named instance of the given container image.
- `singularity [...] list [list options...]`: List all running named instances
- `singularity [...] stop [stop options...] <container path> <instance name>`: Stop a named instance of the given container image.
- `singularity [...] stop-all [stop-all options...]`: Stop all running container instances.

=========================================================================

New `--join <instance name>` option for singularity verbs [run, exec, and shell]. This lets you use the current `singularity [verb]` commands but instead of starting a new container the process will run inside the specified container instance. For example:

```
singularity start test.img test
singularity shell --join test test.img
```

=========================================================================

There are still things to add, but this PR is a foundation for the things to be built on top of it. Here are some things on the to-do list that should still get done, but aren't 5 minute fixes:

- `%startscript` I would like the user to not have to execute their workloads in two different steps. With a `%startscript` in the bootstrap definition file, the user can provide a script that defines the behavior when the container instance starts. This would be the place where you would start MySQL in a MySQL container.
- `%stopscript` This is the same idea as the `%startscript`, except it is a script where you can define what to do when you issue a `singularity stop` command. By default, the stop command just has the init process in the PID namespace exit, which causes the kernel to reap the processes that live in the namespace. This script would first be run by the init process, and would potentially contain a script to gracefully shutdown whatever services are running in the container. 
- Tests. There are no tests. We need to write tests.
- Give the user the ability to use the init process provided inside the container instead of the default "sinit" 

=========================================================================

- Ref: #

**Checkoff for all PRs:**

Attn: @singularityware-admin


libexec/cli/list.exec
```diff
@@ -0,0 +1,110 @@
+#!/bin/bash
+# 
+# Copyright (c) 2017, SingularityWare, LLC. All rights reserved.
+#
+# This software is licensed under a 3-clause BSD license.  Please
+# consult LICENSE file distributed with the sources of this project regarding
```
vsoch, line 6: If you agree with my suggestion, this script would be called `ps.exec`. If we anticipate different functions for daemons,  (or should we call them instances?) then we may want to do something like `instance-ps.exec` or `daemon-ps.exec`

libexec/cli/list.exec
```diff
@@ -0,0 +1,110 @@
...
+        -a|--all)
+            if [ "$USERID" = 0 ]; then
+                USERID="*"
+                shift
+            else
+                message ERROR "Must be root to list all\n"
+                exit 1
+            fi
+        ;;
+        -u|--user)
```
vsoch: just for a sanity check - this means that the intended use is to do `--all` or `--user` but not both. We want to be sure that there isn't some trick to ask for a particular user (eg root) and then change the id to that, and then be able to list all (when you shouldn't be able to). I just iterated through the possibilities in my head and I think we are ok (because you must always be root to either ask for a specific user or all), but wanted to point out for others to check as well!

libexec/cli/list.exec
```diff
@@ -0,0 +1,110 @@
...
+        *)
+            break;
+        ;;
+    esac
+done
+
+
+# This is a configurable option, needs to be set accordingly!
+SESSION_GLOB="/tmp/.singularity-daemon-${USERID}/*"
+
```
vsoch: The one trip up I found with this workflow is the association of a particular user and a daemon. The reason is because a lot of the creation commands I'd be using sudo, but then later I would list and be confused to not see my processes. I anticipate this is going to trip a lot of people up. However, it is important to have that association. Is there any helpful messages we can provide the user? Or maybe we should just let it be, and then see the response, and see if anyone has suggestions for a better workflow.

vsoch: One more sanity check - given that we store the session globs in tmp, this means that any services / etc that Singularity is running will have their metadata (with the pid) lost given clearing tmp, or restarting. Is there any case that we can anticipate someone wanting to keep a container instance after a computer restart or the like?

libexec/cli/list.help
```diff
@@ -0,0 +1,32 @@
+USAGE: singularity [...] list [list options...]
+
+List all running named instances
```
vsoch, line 3: this should maybe be ps, I will post my main comments in a bit!

libexec/cli/list.help
```diff
@@ -0,0 +1,32 @@
...
+    /home/mibauer/singularity/sinstance/test.img  [2056.132716]  Instances:
+         Owner UID:      0
+         Instance PID:   16219
+         Instance Name:  test
+
+         Owner UID:      1000
+         Instance PID:   11963
+         Instance Name:  test
+
+
```
vsoch: nitpicking - get rid of extra white space because my terminal is TINY

libexec/cli/start.exec
```diff
@@ -0,0 +1,181 @@
...
+# 
+# Copyright (c) 2017, SingularityWare, LLC. All rights reserved.
+#
+# This software is licensed under a 3-clause BSD license.  Please
+# consult LICENSE file distributed with the sources of this project regarding
+# your rights to use or distribute this software. 
+#
+# 
+
+
```
vsoch, line 11: is there any reason we need to redo code here that is in the [action_argparser](https://github.com/singularityware/singularity/blob/master/libexec/cli/action_argparser.sh)?

libexec/cli/start.exec
```diff
@@ -0,0 +1,181 @@
...
+    esac
+done
+
+
+if [ -z "${SINGULARITY_IMAGE:-}" ]; then
+    SINGULARITY_IMAGE="${1:-}"
+    export SINGULARITY_IMAGE
+    shift
+fi
+
```
vsoch, line 148: Too bad we don't have the bash equivalent of haikunator to come up with funny names, if the user doesn't provide one :)

libexec/cli/status.exec
```diff
@@ -1,87 +0,0 @@
...
-# 
-# Copyright (c) 2016-2017, The Regents of the University of California,
-# through Lawrence Berkeley National Laboratory (subject to receipt of any
-# required approvals from the U.S. Dept. of Energy).  All rights reserved.
-# 
-# This software is licensed under a customized 3-clause BSD license.  Please
-# consult LICENSE file distributed with the sources of this project regarding
-# your rights to use or distribute this software.
-# 
-# NOTICE.  This Software was developed under funding from the U.S. Department of
```
vsoch, line 15: I hope we eventually add this guy back in :)

libexec/cli/stop-all.exec
```diff
@@ -0,0 +1,95 @@
...
+        -h|--help|help)
+            if [ -e "$SINGULARITY_libexecdir/singularity/cli/$SINGULARITY_COMMAND.help" ]; then
+                cat "$SINGULARITY_libexecdir/singularity/cli/$SINGULARITY_COMMAND.help"
+            else
+                message ERROR "No help exists for this command\n"
+                exit 1
+            fi
+            exit
+        ;;
+        -a|--all)
```
vsoch: hmm, since we have this shared action to parse common arguments for the daemon...
  - if it's a trivial addition, and if daemon commands are needed with others, we can add to the already existing argparser (assuming no conflict or reason not to)
 - if there are a substantial number of daemon specific commands, we could add them to a common daemon command argparser file.

The reason for this is making updates and changes. It's a lot easier to update/change one shared function than have to do the same things across multiple files.

libexec/cli/stop-all.exec
```diff
@@ -0,0 +1,95 @@
...
+    if [ -f "$i" ]; then
+        if [ ! -s "$i" ]; then
+            rm "$i"
+            continue
+        fi
+        FILE_NAME=`basename "$i"`
+        DAEMON_NAME=${FILE_NAME##*-}
+        . "$i"
+
+        echo "Stopping ${DAEMON_NAME} instance of ${DAEMON_IMAGE} (PID=${DAEMON_PID})"
```
vsoch, line 82: Should this (and other echos for the daemon scripts) be an info message in case the user wants to completely quiet it? (eg message 1)

src/lib/runtime/ns/pid/pid.c
```diff
@@ -68,9 +68,45 @@ int _singularity_runtime_ns_pid(void) {
...
+int _singularity_runtime_ns_pid_join(void) {
+    int ns_fd = atoi(singularity_registry_get("DAEMON_NS_FD"));
+    int pid_fd;
+
+    /* Attempt to open /proc/[PID]/ns/pid */
+    pid_fd = openat(ns_fd, "pid", O_RDONLY);
+
+    if( pid_fd == -1 ) {
+        /* Daemons should always have a ns/pid file. If it doesn't exist, something is wrong */
+        singularity_message(ERROR, "Could not open PID NS fd: %s\n", strerror(errno));
```
vsoch, line 92: +1

src/util/daemon.c
```diff
@@ -0,0 +1,194 @@
...
+}
+
+void singularity_daemon_path(void) {
+    char *daemon_path = (char *)malloc(2048 * sizeof(char));
+
+    int uid = singularity_priv_getuid();
+    char *dev_ino = file_devino(singularity_registry_get("IMAGE"));
+    char *name = singularity_registry_get("DAEMON_NAME");
+    
+    snprintf(daemon_path, 2048, "/tmp/.singularity-daemon-%d/%s-%s", uid, dev_ino, name);
```
vsoch: I can almost assure you cluster admins will want to be able to specify some custom/special place for the daemon's "glob" - should we have that go into the config? It's sort of similar to a session directory.

src/sinit.c
```diff
@@ -0,0 +1,77 @@
...
+int main(int argc, char **argv) {
+    char *daemon_fd_str;
+    int daemon_fd;
+    int i;
+    
+    singularity_config_init(joinpath(SYSCONFDIR, "/singularity/singularity.conf"));
+    singularity_priv_init();
+    singularity_registry_init();
+
+    /* After this point, we are running as PID 1 inside PID NS */
```
vsoch, line 54: I don't know if this is the right file for the comment, but I noticed in testing that if I did a start followed immediately by a join, I would get an exit with 255 with an error message that the daemon wasn't found. However, then it still existed. When I added a short sleep after the start, it worked seamlessly. I think we might consider adding an internal pause to make it so that users don't hit this error.

src/util/daemon.c
```diff
@@ -0,0 +1,194 @@
...
+    char *daemon_file = singularity_registry_get("DAEMON_FILE");
+    char *daemon_name = singularity_registry_get("DAEMON_NAME");
+    
+    /* Check if there is a lock on daemon file */
+    singularity_message(DEBUG, "Checking for lock on %s\n", daemon_file);
+    lock_result = filelock(daemon_file, lock_fd);
+
+    if ( lock_result == 0 ) {
+        /* Successfully obtained lock, no daemon controls this file. */
+        singularity_message(ERROR, "Unable to join daemon: %s daemon does not exist\n",
```
vsoch, line 84: I think this might have been the error I hit, but I'm not totally positive.

vsoch (changes requested):
Holy cow @bauerm97 that's a ton of work you've put into this! I've already tested in quite some detail, and I am +1 that we should get this skeleton in, and then address some of the additions (startscript, stopscript, and usage). My one quick critique would be that instead of singularity list, we follow convention with other container software and do `singularity ps`. I think `singularity list` is more logical to map to wanting to list a container's set of modules / software applications, and that is going to be the more commonly used of the two. If we do go with list, we would need another easy to remember command to see the software modules provided by the container.

See my comments - and let's discus! Also we should comment this will close #830

src/util/daemon.c
gmkurtzer: This could be safely an environment variable, but it runs the chance that if it is changed with daemons running, they may not be seen via `stop` or `list`. But because this is a simple ASCII based PID file, I'm not sure there is much need to move it.

libexec/cli/stop-all.exec
gmkurtzer: +1

libexec/cli/status.exec
gmkurtzer: Which one? funding from the US Government or the LICENSE file reference?

libexec/cli/start.exec
gmkurtzer: OMG, noooooooooo! hahaha

libexec/cli/start.exec
gmkurtzer: `singularity start` options are not exactly the same as the other actions.

libexec/cli/list.help
gmkurtzer: Actually, it maybe possible for us to add a real `singularity ps ...` at some point which will properly show a ps like output for all processes running within a container. We may even be able to write a `top` like interface too!

libexec/cli/list.exec
gmkurtzer: This is a good point. We do however need to have this writable as well as persistent. `/var/tmp/` or even `~/.singularity/daemon-pids` might also be reasonable, but it breaks the ability to easily allow root to have visibility as to what users are doing. @bauerm97, we spoke about this, do you recall why we decided against using $HOME?

GodloveD:
I know we already talked about this and you said you have some strategies but I want to put it here so that it doesn't get lost.  I think it would be good to resolve this before merging.  

Right now a *contaemon* (never gets old!) is tied to a file in `/tmp`.  We call it a PID file.  If that file gets out of sync with the actual process it can lead to some unusual behavior.  For instance, consider this ghostly  container that appears to rise from the dead.
```
02:40 PM
ubuntu@ubuntu-xenial:~/daemon-testing
$ ~/sing-daemon/bin/singularity start /cron.img cron

02:41 PM
ubuntu@ubuntu-xenial:~/daemon-testing
$ ~/sing-daemon/bin/singularity list
/cron.img  [2049.2186]  Instance:
     Instance PID:   11416
     Instance Name:  cron


02:41 PM
ubuntu@ubuntu-xenial:~/daemon-testing
$ kill -9 11416

02:41 PM
ubuntu@ubuntu-xenial:~/daemon-testing
$ ps -ef | grep sinit
ubuntu   11428 21542  0 14:41 pts/0    00:00:00 grep --color sinit

02:41 PM
ubuntu@ubuntu-xenial:~/daemon-testing
$ ll /tmp/.singularity-daemon-1000/
total 4.0K
-rw-r--r-- 1 ubuntu 113 Jul 28 14:40 2049.2186-cron

02:41 PM
ubuntu@ubuntu-xenial:~/daemon-testing
$ ~/sing-daemon/bin/singularity list
/cron.img  [2049.2186]  Instance:
     Instance PID:   11416
     Instance Name:  cron
```
Or this invisible container:
```
02:48 PM
ubuntu@ubuntu-xenial:~/daemon-testing
$ ~/sing-daemon/bin/singularity start cron.img cron

02:49 PM
ubuntu@ubuntu-xenial:~/daemon-testing
$ rm /tmp/.singularity-daemon-1000/*

02:49 PM
ubuntu@ubuntu-xenial:~/daemon-testing
$ ~/sing-daemon/bin/singularity list

02:49 PM
ubuntu@ubuntu-xenial:~/daemon-testing
$ ps -ef | grep sinit
ubuntu   11443     1  0 14:49 pts/0    00:00:00 /home/ubuntu/sing-daemon/libexec/singularity/bin/sinit /home/ubuntu/sing-daemon/var/singularity/mnt/overlay/final
ubuntu   11452 21542  0 14:49 pts/0    00:00:00 grep --color sinit
```
Given the PID files location in `/tmp` the second scenario is going to happen fairly frequently.  Add to this the fact that these containers will (I think) continue to run after the user logs out (which I think is good) and you're going to have a whole lot of `sinit` processes hanging around without the metadata to `stop` them.

libexec/cli/stop.help
```diff
@@ -0,0 +1,21 @@
+USAGE: singularity [...] stop [list options...] <container path> <instance name>
+
+Stop a named instance of the given container image.
+
+LIST OPTIONS:
+    -a/--all               If running as root, stop the specified instance
```
GodloveD: Is this working?  I can't seem to get either `--all` or `stop-all` to work as advertised when run as root.  Also maybe a very quick word of caution here so the admin knows that `--all` may have the unexpected side effect of creating nasty emails in one's inbox (depending on the system).

libexec/cli/start.help
```diff
@@ -0,0 +1,79 @@
...
+    -H/--home <spec>    A home directory specification.  spec can either be a
+                        src path or src:dest pair.  src is the source path
+                        of the home directory outside the container and dest
+                        overrides the home directory within the container
+    -n/--nv             Enable experimental Nvidia support
+    --pwd               Initial working directory for payload process inside
+                        the container
+    -S/--scratch <path> Include a scratch directory within the container that
+                        is linked to a temporary dir (use -W to force location)
+    -s/--shell <shell>  Path to program to use for interactive shell
```
GodloveD, line 25: Does `--shell` make sense in the context of start?  Should it be removed from help?

libexec/cli/list.help
GodloveD: If the name is going to change, (and I like `list` just fine) I would advocate for `status` to be more in line with the daemon mindset (that we are thinking about but avoiding).  😆

src/util/daemon.c
vsoch: okay, that's a good point. Let's have it like this for now, and wait for someone to complain they want to move it :)

libexec/cli/start.help
vsoch: I think that's probably just a copy paste artifact or something, tiny detail good catch!

libexec/cli/list.exec
GodloveD: I think the loose coupling between the process and the PID file is a pretty serious problem that needs to be resolved.  See my comment below.

libexec/cli/list.help
gmkurtzer: I was actually thinking of using status to get more detailed information about a particular container instance. lol

gmkurtzer:
I agree with @GodloveD and @vsoch's point about the risks with using the `/tmp` directory. Either we handle it by using `/var/tmp/` instead, or we have to use the user's home directory, and deal with the caveat that root won't probe through all user's containers with the `--all` option. But `--user` could still be easily leveraged.

libexec/cli/stop-all.exec
bauerm97: Yes there is a specific reason for not having this in `action_argparser.sh`. That file is only for argument parsing for action commands. This is the same reason that `stop.exec` and `stop-all.exec` user their own argument parsing methodology.

vsoch: should we have a `deamon_argparser.sh` equivalent? (or whatever you would want to call it). This way, when someone updates some argument variable for one of the calls (eg, start stop, all) the update can be done in one spot with assurance of consistency across calls.

bauerm97: The only commands that share 100% of their arguments are `stop` and `stop-all`. However, soon `stop-all` will have a few extra arguments that `stop` doesn't have. It doesn't really seem to make sense to do that then.

vsoch: If they really share 100%, I think it might be worth considering, If you look at the main argparse helper there are a lot of overlapping commands. @gmkurtzer what are your thoughts?

libexec/cli/list.exec
bbockelm: Note that doing a well-known filename in a world-writable directory opens, at-best, a denial-of-service (if I can reasonably guess your UID and image, then I can block you from being able to start it) and occasionally a serious exploit (tricking a script to write into a file I otherwise can't affect).  You probably really want `~/.singularity`.

bauerm97: @bbockelm Moved this into `~/.singularity/`, thanks for the feedback!

libexec/cli/list.help
vsoch: +1 for:

- status `singularity status container.img mysql` --> show me container instance status
- ps`singularity ps`--> show me all running processes. `ps` coincides with use of docker-compose and docker, so my preference is for ps to be associated with listing processes.
- `list` I think we should save for something in the future that will be highly used, like listing containers in some registry, or that match some filter, etc.

vsoch:
#feature-request can we name the daemon sparky? Or kerberos? Or maybe... little D? big D? hehe.

gmkurtzer:
I like "Sparky"! Little D?!?!? Big D?!?! OMG,... so work inappropriate V!

vsoch:
he's a gangsta' daemon.

gmkurtzer:
This is so done.......!!!! Thanks everyone, great work!

Issue #831, opened 2017-07-25
Label Schema


vsoch:
Let's conform to this - http://label-schema.org/rc1/ It's easy enough, and I've seen it on many Docker containers.

jscook2345:
@vsoch 

It looks like this was fixed and merged so I'm going to close. If not, leave a comment and I'll reopen. Thanks!

vsoch:
Wow this is old! Yes, it looks great ;)

```
$ singularity inspect salad_latest.sif 
==labels==
org.label-schema.usage.singularity.deffile.bootstrap: docker
org.label-schema.usage.singularity.deffile.from: vanessa/salad
org.label-schema.usage.singularity.version: 3.1.0-rc2.1154.g479352901
org.label-schema.build-date: Thursday_11_April_2019_9:13:20_EDT
org.label-schema.schema-version: 1.0
```

It definitely should have been closed - the remaining issue (represented here https://github.com/sylabs/singularity/issues/2790) is that the upper level "data" key is missing. I just tested a fresh pull:

```bash
$ singularity inspect --json salad_latest.sif 
{
	"attributes": {
		"labels": {
			"org.label-schema.build-date": "Thursday_11_April_2019_9:13:20_EDT",
			"org.label-schema.schema-version": "1.0",
			"org.label-schema.usage.singularity.deffile.bootstrap": "docker",
			"org.label-schema.usage.singularity.deffile.from": "vanessa/salad",
			"org.label-schema.usage.singularity.version": "3.1.0-rc2.1154.g479352901"
		}
	},
	"type": "container"
}
```

PR #843, opened 2017-07-29
Addition of org.label-schema and bootstrap %help section

vsoch:

This PR will bring standard labels to Singularity images (already a practice for Docker and rxc via the label schema (see #831), along with adding a `%help` section option to the bootstrap file, so a user can generate an image with a help section without needing to know how to write a more complicated argument parser for the runscript.

For a preview of how it looks, see here:

https://asciinema.org/a/131139?speed=3

I changed one detail in what is shown above, the singularity_deffile_[ ARG ] labels I replaced the prefix in favor of the label schema standard. So we have:

```
"org.label-schema.usage.singularity.deffile": "Singularity.help",
"org.label-schema.usage.singularity.deffile.from": "ubuntu:latest",
```

instead of:

```
"org.label-schema.usage.singularity.deffile.singularity_deffile": "Singularity.help",
"org.label-schema.usage.singularity.deffile.singularity_defifile_from": "ubuntu:latest",
```
which is a bit redundant!

Attn: @singularityware-admin


gmkurtzer:
Do we need Python to print help? Is it possible to print via `cat`, or other shell commands?

If possible, I'd rather keep Python out of any of the container operation pathways as it is a large dependency to require (e.g. for very light weight host runtimes), and it seems like printing a preformatted text help file should be pretty easy to do without Python.

Thanks!

vsoch:
I went this route because a user might want to ask for a formatted help, akin to how you'd ask for inspect --json and it prints json. The additional formatting wouldn't be possible with just something like cat. If it's ok to have python for inspect, then why isn't it ok for here, for which the same need is present?

vsoch:
As an alternative, we can add the help as an option for inspect (where we already have python), and then use cat. Let me try that out and then you can take a look!

vsoch:
okey doke! @gmkurtzer the python dependency is removed - the helpfile is shown with `cat`. The same programmatic accessibility to the helpfile (and getting the info in json) is now provided via inspect! Here is a demo:

https://asciinema.org/a/131343?speed=3

vsoch:
@gmkurtzer this is again ready for review.

libexec/helpers/help.sh
```diff
@@ -0,0 +1,37 @@
...
+
+if [ ! -d "${SINGULARITY_MOUNTPOINT}/.singularity.d" ]; then
+    message ERROR "The Singularity metadata directory does not exist in image\n"
+    ABORT 255
+fi
+
+SINGULARITY_ROOTFS=${SINGULARITY_MOUNTPOINT}
+export SINGULARITY_MOUNTPOINT SINGULARITY_ROOTFS
+
+if [ -f "${SINGULARITY_MOUNTPOINT}/.singularity.d/runscript.help" ]; then
```
gmkurtzer, line 35: If no help exists, should we print a message about the lack of a container help?

libexec/cli/help.exec
```diff
@@ -0,0 +1,43 @@
...
+## Load functions
+if [ -f "$SINGULARITY_libexecdir/singularity/functions" ]; then
+    . "$SINGULARITY_libexecdir/singularity/functions"
+else
+    echo "Error loading functions: $SINGULARITY_libexecdir/singularity/functions"
+    exit 1
+fi
+
+RETVAL=0
+SINGULARITY_IMAGE="${1:-}"
```
gmkurtzer: If no singularity image is defined at $1, we should either exit and error, print usage summary, or print the `.../libexec/cli/singularity.help`.

libexec/python/helpers/json/help.py
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env python
```
gmkurtzer: do we still need this file?

libexec/helpers/help.sh
vsoch: I thought about that, but then what if the runscript itself has help? I opted for blank because at least that doesn't print something incorrect. That said, I'm not hugely opinionated about this one, would you want to see something like `No runscript help is defined for this image.` or something in addition to advise the user on how to run/use?

libexec/python/helpers/json/help.py
vsoch: later we will, but not now, so I'll move it aside!

libexec/cli/help.exec
vsoch: This was missing in checks too, so I added it.

gmkurtzer:
Merged, and fixed up bash_completion! Thanks!

vsoch:
Woohoo!! Thanks @gmkurtzer, super happy to see this one go through!

PR #854, opened 2017-08-03
Security enhancements for old kernels

cclerget:

- Add security enhancements for old kernels (>= 2.6.25) to support missing features due to lack of NO_NEW_PRIVS support on these systems.
On hosts with kernel < 2.6.25, a warning is displayed at configuration time and users should explicitly set --disable-security-check with configure, a warning notice will be displayed mentioning security implications to install Singularity on these systems.
- Force OVERLAYFS_ENABLED to NULL if disabled or unsupported
- Don't remount /dev and trust host mount options to avoid clearing nodev partition's flag with symlink race

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/src/lib/runtime/mounts/dev/dev.c
+++ b/src/lib/runtime/mounts/dev/dev.c
@@ -117,12 +117,6 @@ int _singularity_runtime_mount_dev(void) {
                     singularity_message(ERROR, "Could not bind mount container's /dev: %s\n", strerror(errno));
                     ABORT(255);
                 }
-                if ( singularity_priv_userns_enabled() != 1 ) {
-                    if ( mount(NULL, joinpath(container_dir, "/dev"), NULL, MS_BIND|MS_NOSUID|MS_REC|MS_REMOUNT, NULL) < 0 ) {
-                        singularity_message(ERROR, "Could not remount container's /dev: %s\n", strerror(errno));
-                        ABORT(255);
-                    }
-                }
                 singularity_priv_drop();
         } else {
             singularity_message(WARNING, "Not mounting /dev, container has no bind directory\n");
--- a/configure.ac
+++ b/configure.ac
@@ -45,6 +45,36 @@ if test "$libdir" = "\${exec_prefix}/lib"; then
     LDFLAGS="$LDFLAGS -Wl,-rpath -Wl,\$(libdir)"
 fi
 
+AC_MSG_CHECKING([if security check is enabled])
+AC_ARG_ENABLE([security_check],
+  [AS_HELP_STRING([--disable-security-check],
+    [disable security check (WARNING: set this option expose your host to privileges escalation)])])
+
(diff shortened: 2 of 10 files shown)


src/lib/runtime/mounts/dev/dev.c
```diff
@@ -117,12 +117,6 @@ int _singularity_runtime_mount_dev(void) {
                     singularity_message(ERROR, "Could not bind mount container's /dev: %s\n", strerror(errno));
                     ABORT(255);
                 }
-                if ( singularity_priv_userns_enabled() != 1 ) {
```
gmkurtzer, line 120: @cclerget Can you describe why you are removing this bit?

cclerget: @gmkurtzer /dev remount is removed here for security reason.

Here the attack scenario:
- put a symlink in container image where dev point to /var/tmp/fakedev
- fakedev itself is a symlink pointing to /usr/local/var/singularity/mnt/container/finaldev
- finaldev is the target directory in container where /dev/ will reside

So for the first /dev mount will follow: 
- /usr/local/var/singularity/mnt/container/dev -> /var/tmp/fakedev -> /usr/local/var/singularity/mnt/container/finaldev.

The attack here is to exploit race for the second /dev (re)mount, where /var/tmp/fakedev will point to /usr/local/var/singularity/mnt/container, the target for second mount become:
- /usr/local/var/singularity/mnt/container/dev -> /var/tmp/fakedev -> /usr/local/var/singularity/mnt/container

which results in clearing MS_NODEV mount flag for "/" in container. That's why I remove the second mount completely, I think it's safe to trust host mount options for /dev

Between after some tests, the replacement of PR_SET_NO_NEW_PRIVS for old kernels is not so efficient as I though, it works for setuid binaries which manipulate groups like sudo or su, but if setgid bit is set on these files, fix become useless ... It's better than nothing since it disables file capabilities but inefficient with setuid and setgid binaries. I revert corresponding commit from this PR, so the others changes could be merged.

So as we already discussed, a more robust mount mechanism is required with old systems to safely remove NO_NEW_PRIVS checks from code. I work on it

gmkurtzer:
Hi @cclerget, how come you reverted the drop cap part of this PR? Should I still merge it as is with the other bits?

Thanks!

cclerget:
@gmkurtzer actually not entirely satisfied by drop cap fix, setuid/setgid permission bits completely bypass capability mask with bounding set for CAP_SETUID/CAP_SETGID, I consider this as a kernel bug or missing feature. So privilege escalation is always possible.

You can merge this PR as is. I will submit a PR later with drop cap fix and a more robust mount system to kill NO_NEW_PRIVS check on older systems

gmkurtzer:
ok, sounds good! Also, @bauerm97 has been working on capability testing (but will be afk for a bit) as well, so you two might want to sync up on that.

Thanks!

PR #850, opened 2017-08-01
adding optional https to docker

vsoch:
**Description of the Pull Request (PR):**

This PR will add an environment variable, `SINGULARITY_NOHTTPS` that will disable using https for interacting with a Docker registry. This option is primarily intended for users running a local registry that cannot support (or don't want to set up) a certificate, which is a very reasonable thing to do. This first snippet shows exporting the variable, and that it shows up as being found:
```
SINGULARITY_NOHTTPS=yes
export SINGULARITY_NOHTTPS
singularity --debug run docker://ubuntu:14.04
Enabling debugging
Ending argument loop
Singularity version: 2.3.1-add/https-optional.ga78190c
Exec'ing: /usr/local/libexec/singularity/cli/run.exec
Evaluating args: 'docker://ubuntu:14.04'
VERBOSE2 SINGULARITY_COMMAND_ASIS found as False
VERBOSE2 SINGULARITY_ROOTFS found as /tmp/.singularity-runtime.cTWaCjRb/ubuntu:14.04
VERBOSE2 SINGULARITY_METADATA_FOLDER found as /tmp/.singularity-runtime.cTWaCjRb/ubuntu:14.04/.singularity.d
VERBOSE2 SINGULARITY_FIX_PERMS found as False
VERBOSE2 SINGULARITY_COLORIZE not defined (None)
VERBOSE2 SINGULARITY_DISABLE_CACHE found as False
VERBOSE2 SINGULARITY_CACHEDIR found as /home/vanessa/.singularity
VERBOSE2 SINGULARITY_ENVIRONMENT found as /tmp/.singularity-runtime.cTWaCjRb/ubuntu:14.04/.singularity.d/environment
VERBOSE2 SINGULARITY_RUNSCRIPT found as /tmp/.singularity-runtime.cTWaCjRb/ubuntu:14.04/singularity
VERBOSE2 SINGULARITY_TESTFILE found as /tmp/.singularity-runtime.cTWaCjRb/ubuntu:14.04/.singularity.d/test
VERBOSE2 SINGULARITY_DEFFILE found as /tmp/.singularity-runtime.cTWaCjRb/ubuntu:14.04/.singularity.d/Singularity
VERBOSE2 SINGULARITY_ENVBASE found as /tmp/.singularity-runtime.cTWaCjRb/ubuntu:14.04/.singularity.d/env
[...]

Issue #849, opened 2017-08-01
ssl error during singularity pull of images from self-hosted registry


hcraT:
## Version of Singularity:

2.2.99 (git development branch)

### Expected behavior

Singularity image for requested docker image is generated.

### Actual behavior

The following error is produced:
```
Traceback (most recent call last):
  File "/usr/lib/python3.6/urllib/request.py", line 1318, in do_open
    encode_chunked=req.has_header('Transfer-encoding'))
  File "/usr/lib/python3.6/http/client.py", line 1239, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/usr/lib/python3.6/http/client.py", line 1285, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.6/http/client.py", line 1234, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.6/http/client.py", line 1026, in _send_output
    self.send(msg)
  File "/usr/lib/python3.6/http/client.py", line 964, in send
    self.connect()
  File "/usr/lib/python3.6/http/client.py", line 1400, in connect
    server_hostname=server_hostname)
  File "/usr/lib/python3.6/ssl.py", line 401, in wrap_socket
    _context=self, _session=session)
  File "/usr/lib/python3.6/ssl.py", line 808, in __init__
    self.do_handshake()
  File "/usr/lib/python3.6/ssl.py", line 1061, in do_handshake
    self._sslobj.do_handshake()
  File "/usr/lib/python3.6/ssl.py", line 683, in do_handshake
    self._sslobj.do_handshake()
ssl.SSLError: [SSL: WRONG_VERSION_NUMBER] wrong version number (_ssl.c:748)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/libexec/singularity/python/size.py", line 90, in <module>
    main()
  File "/usr/libexec/singularity/python/size.py", line 82, in main
    contentfile=LAYERFILE)
  File "/usr/libexec/singularity/python/docker/main.py", line 57, in SIZE
    client = DockerApiConnection(image=image,auth=auth)
  File "/usr/libexec/singularity/python/docker/api.py", line 82, in __init__
    self.load_image(kwargs['image'])
  File "/usr/libexec/singularity/python/docker/api.py", line 118, in load_image
    self.update_token()
  File "/usr/libexec/singularity/python/docker/api.py", line 129, in update_token
[...]

PR #856, opened 2017-08-03
Add/http domains

hcraT:

This PR adds the possiblity to specify a list of docker registries
that will be accessed with HTTP instead of HTTPS. It is similar
to PR #850 but in my opionion is a little bit more convenient in
the fact that it allows a user to have a fixed list of http registries
that can be added to a .bashrc or .bash_profile. Moreover it
allows to use both self-hosted and other servers without
having to act on environment variables in order to disable/enable
HTTPS.
This solution peacfully co-exists with @vsoch solution but I would
say that having both would add probably a bit of redundancy.
I had the idea after seeing @vsoch PR and this is the reason I
haven't created the PR before @vsoch's got merged.

- Ref: #849 

**Checkoff for all PRs:**

I've tried to run the tests (make test) but I always run into some problem (and I'm pretty confident that this doesn't depend on the code I wrote).

I've tested however the code on my machine and it works correctly.

Attn: @singularityware-admin

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -29,3 +29,4 @@
     - Rémy Dernat <remy.dernat@umontpellier.fr>
     - Mark Egan-Fuller <markeganfuller@googlemail.com>
     - Petr Votava <votava.petr@gene.com>
+    - Tarcisio Fedrizzi <taracisio.fedrizzi@gmail.com>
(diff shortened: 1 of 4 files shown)


gmkurtzer:
At the moment (as you see) we are using environment variables for everything, and it has already gotten unruly. Before we add anymore, which will cause us to have to consider legacy support, I would like to see a configuration interface get developed. I know that @vsoch has been thinking about this already as we have been discussing for some time. The feature that you are requesting can/should be satisfied with that new configuration paradigm, so I want to hold off on merging this until after we support for that.

Thanks!

PR #865, opened 2017-08-09
Fix userbind read only

vsoch:

This is a quick fix by @cclerget to make user binds with readonly (:ro) function correctly. Thanks @cclerget !

Attn: @singularityware-admin


gmkurtzer:
Awesome, thanks!!

PR #867, opened 2017-08-12
Bug/docker url namespaces

hartzell:
There can be multiple components in the namespace, e.g.:

```
docker://myregistry.example.com:5000/name/space/repo_name@sha256:570b5227f93f870aff05c040c5b1f27c15b5978dbf7704e877b71691ed58ede2
```

Where the registry is: `myregistry.example.com:5000`

and the namespace is `name/space`

and the repo_name is `repo_name`.

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/libexec/python/shell.py
+++ b/libexec/python/shell.py
@@ -120,9 +120,9 @@ def parse_image_uri(image, uri=None, quiet=False):
     # Now look for registry, namespace, repo
     image = image.split('/')
 
-    if len(image) == 3:
+    if len(image) > 2:
         registry = image[0]
-        namespace = image[1]
+        namespace = "/".join(image[1:-1])
         repo_name = image[2]
 
     elif len(image) == 2:
--- a/libexec/python/shell.py
+++ b/libexec/python/shell.py
@@ -123,7 +123,7 @@ def parse_image_uri(image, uri=None, quiet=False):
     if len(image) > 2:
         registry = image[0]
         namespace = "/".join(image[1:-1])
-        repo_name = image[2]
+        repo_name = image[-1]
 
     elif len(image) == 2:
         registry = default_registry
--- a/libexec/python/tests/test_core.py
+++ b/libexec/python/tests/test_core.py
@@ -149,18 +149,18 @@ class TestShell(TestCase):
         self.assertTrue(digest['namespace'] == 'mix')
 
         print("Case 8: Custom uri should use it.")
-        image_name = "catdog://meow/mix/%s:%s" % (self.repo_name,
+        image_name = "catdog://meow/mix/tenders/%s:%s" % (self.repo_name,
                                                   self.tag)
         digest = parse_image_uri(image_name, uri="catdog://")
         self.assertTrue(digest['registry'] == 'meow')
-        self.assertTrue(digest['namespace'] == 'mix')
+        self.assertTrue(digest['namespace'] == 'mix/tenders')
 
         print("Case 9: Digest version should be parsed")
-        image_name = ("catdog://meow/mix/%s:%s@sha:256xxxxxxxxxxxxxxx"
+        image_name = ("catdog://meow/mix/original/choice/%s:%s@sha:256xxxxxxxxxxxxxxx"
                       % (self.repo_name, self.tag))
         digest = parse_image_uri(image_name, uri="catdog://")
         self.assertTrue(digest['registry'] == 'meow')
-        self.assertTrue(digest['namespace'] == 'mix')
+        self.assertTrue(digest['namespace'] == 'mix/original/choice')
         self.assertTrue(digest['version'] == 'sha:256xxxxxxxxxxxxxxx')
 
 
--- a/libexec/python/tests/test_core.py
+++ b/libexec/python/tests/test_core.py
@@ -150,13 +150,13 @@ class TestShell(TestCase):
 
         print("Case 8: Custom uri should use it.")
         image_name = "catdog://meow/mix/tenders/%s:%s" % (self.repo_name,
-                                                  self.tag)
+                                                          self.tag)
         digest = parse_image_uri(image_name, uri="catdog://")
         self.assertTrue(digest['registry'] == 'meow')
         self.assertTrue(digest['namespace'] == 'mix/tenders')
 
         print("Case 9: Digest version should be parsed")
-        image_name = ("catdog://meow/mix/original/choice/%s:%s@sha:256xxxxxxxxxxxxxxx"
+        image_name = ("catdog://meow/mix/original/choice/%s:%s@sha:256xxxxxxxxxxxxxxx" # noqa
                       % (self.repo_name, self.tag))
         digest = parse_image_uri(image_name, uri="catdog://")
         self.assertTrue(digest['registry'] == 'meow')
--- a/libexec/python/tests/test_core.py
+++ b/libexec/python/tests/test_core.py
@@ -156,7 +156,7 @@ class TestShell(TestCase):
         self.assertTrue(digest['namespace'] == 'mix/tenders')
 
         print("Case 9: Digest version should be parsed")
-        image_name = ("catdog://meow/mix/original/choice/%s:%s@sha:256xxxxxxxxxxxxxxx" # noqa
+        image_name = ("catdog://meow/mix/original/choice/%s:%s@sha:256xxxxxxxxxxxxxxx"  # noqa
                       % (self.repo_name, self.tag))
         digest = parse_image_uri(image_name, uri="catdog://")
         self.assertTrue(digest['registry'] == 'meow')
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -29,3 +29,4 @@
     - Rémy Dernat <remy.dernat@umontpellier.fr>
     - Mark Egan-Fuller <markeganfuller@googlemail.com>
     - Petr Votava <votava.petr@gene.com>
+    - George Hartzell <hartzell@alerce.com>


libexec/python/tests/test_core.py
```diff
@@ -149,18 +149,18 @@ def test_parse_image_uri(self):
...
+        self.assertTrue(digest['namespace'] == 'mix/tenders')
 
         print("Case 9: Digest version should be parsed")
-        image_name = ("catdog://meow/mix/%s:%s@sha:256xxxxxxxxxxxxxxx"
+        image_name = ("catdog://meow/mix/original/choice/%s:%s@sha:256xxxxxxxxxxxxxxx"  # noqa
                       % (self.repo_name, self.tag))
         digest = parse_image_uri(image_name, uri="catdog://")
         self.assertTrue(digest['registry'] == 'meow')
-        self.assertTrue(digest['namespace'] == 'mix')
+        self.assertTrue(digest['namespace'] == 'mix/original/choice')
```
vsoch, line 163: I don't understand this uri - namespace usually refers to the thing that could be thought of like a collection (eg, "library" or for singularity hub a Github repository name. In this case you are saying the namespace has slashes in it? I would have expected the registry to encompass the first bit, and then the namespace to follow. For example:

```
catdog://meow/mix/original/choice/ubuntu:latest@sha:256xxxxxxxxxxxxxxx
[ uri ][        registry        ][ ns  ]:[ tag ]@[       digest        ]
```

vsoch: Just to be clear - you are creating images with names like `mix/original/choice` ? So you would do `docker run -it mix/original/choice bash` ? I haven't seen this kind of usage before, and it's a bit challenging because the entire parsing of the namespace is dependent on the slashes.

hartzell: I've inherited this, not created it :)

My basic problem is that the existing code doesn't deal with multiple levels in the path.  I'm not particularly well equipped to argue about whether the extra bits are registry or namespace....

We have things like this:

```
docker pull https://foo.blah.com:5001/myorg/aproject/repo_name@sha256:530b5227f92f870aff05c040c5b1f27c15b5978dbf7704e877b71691ed58ede2
```

In this case I *think* that `foo.blah.com:5001` is the registry.

In your example above, I don't think that `catdog` is a uri, I think that the entire thing is a URI and that `catdog` is the scheme (based on e.g. [Wikipedia's URI syntax example](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax)).

Stealing your neat labeling trick, I'd end up with:

```
https://foo.blah.com:5001/myorg/aproject/repo_name@sha256:530b5227f92f870aff05c040c5b1f27c15b5978dbf7704e877b71691ed58ede2
[scm]://[    registry   ][  namespace  ][  repo  ]@[digest]
```

Part of my motivation for labelling it this way is we have another "registry" at `https://foo.blah.com:5002/`.

On [the gripping hand](https://en.wikipedia.org/wiki/The_Gripping_Hand), perhaps it's all registries and repositories and "namespaces" are just FigNewtons of our imagination?

```
[vagrant@localhost singularity]$ http http://blah.moo.com:5000/v2/_catalog
HTTP/1.1 200 OK
Content-Type: application/json; charset=utf-8
Date: Sat, 12 Aug 2017 23:46:31 GMT
Docker-Distribution-Api-Version: registry/2.0
Transfer-Encoding: chunked
X-Content-Type-Options: nosniff

{
    "repositories": [
        "demo/blah-map",
        "moose/foo",
        "moose/poodle",
        "hartzelg/doh",
        "hartzelg/hello-world",
        "hartzelg/scooby-doo",
        "hartzelg-this-one-too",
        "hartzelg-hello-world",
        "userx/tuna",
        "research/sloopy/doh-bar",
        "research/sloopy/pete-moose",
[...]
```

vsoch:
Could you please share either some kind of documentation for others doing this practice, or the particular configuration setting for your registry? I'm searching around and I can't find another use case like this. You probably know of one and I'm missing it!

vsoch:
what about instead, since we do have agreement about the image name and tag / digest, instead of changing the way that the image is parsed, instead we allow for environment variables to specify the registry and namespace. So you could do:

```
SINGULARITY_DOCKER_REGISTRY=https://foo.blah.com:5001
SINGULARITY_DOCKER_NAMESPACE=myorg/aproject
export SINGULARITY_DOCKER_NAMESPACE SINGULARITY_DOCKER_REGISTRY
singularity pull docker://repo_name
```
and it would first parse the url, find `repo_name` to be the repository, and at the end of the function check for those environment variables. If found, replace them (in this case we would replace library and the docker hub registry).

hartzell:
I've never seen anyone else doing this but all of my non-trivial Docker experience has been at this gig, and it was set up by Other Folks.

The particular registry I queried above is an instance of the [Docker Registry image](https://hub.docker.com/_/registry/) that's running "normally"

```
docker run -d -p 5000:5000 --restart always --name registry registry:2
```

Most of the thingies in this registry seem to be `a/b/c`.  Here's the distribution:

```
[vagrant@localhost singularity]$ http http://moose.poodle.com:5000/v2/_catalog | jq .repositories[] | awk -F / '{print NF}'| sort | uniq -c
      3 1
      8 2
     70 3
      5 4
```

hartzell:
I'll need to think about where this is all getting hung up.  The environment variables solution is not particularly pleasant.

There's some discussion of the URI  [here in the registry API docs](https://docs.docker.com/registry/spec/api/), scan down to the Overview section.

I'm using to doing this:

```
docker run -it https://foo.blah.com:5001/myorg/aproject/repo_name@sha256:530b5227f92f870aff05c040c5b1f27c15b5978dbf7704e877b71691ed58ede2
```

and to my URL-oriented mind (where the URI's scheme actually describes how to retrieve the resource), I'd like to do (no giggling at any singlarity command line fubars, I'm still new...):

```
singularity run --docker https://foo.blah.com:5001/myorg/aproject/repo_name@sha256:530b5227f92f870aff05c040c5b1f27c15b5978dbf7704e877b71691ed58ede2
```

```
https://foo.blah.com:5001/myorg/aproject/repo_name@sha256:530b5227f92f870aff05c040c5b1f27c15b5978dbf7704e877b71691ed58ede2
[scm]://[   host   ]:port[          name         ]@[     version]
```

We need the *scheme*, *hostname*, and *port* to make the connection, we need the `<name>` (in the parlance of the official doc cited above) to specify the repository and the we might need the tag and version depending on what we're doing.

I'm unclear about the namespace/repo breakdown you've used.

I have another problem that I haven't even taken a stab at yet, several of our repositories are running `http`, *not* `https` and I have no idea how to make that work if Docker images start with `docker://` unless we move to `docker://http://` or `docker-https://` and `docker-http://` and ....

hartzell:
It might even be possible for singularity to "taste" a URL and see if it's a Docker registry or etc....

e.g. `GET /v2/` is indicative of a v2.(.1) Docker registry.

vsoch:
I'll think about this too! Sorry for the silence  - moved today and am just sitting down for the first time, kind of exhausted. With renewed energy later! :)

hartzell:
Don't rush this, re-energize yourself!

libexec/python/shell.py
```diff
@@ -120,9 +120,9 @@ def parse_image_uri(image, uri=None, quiet=False):
     # Now look for registry, namespace, repo
     image = image.split('/')
 
-    if len(image) == 3:
+    if len(image) > 2:
         registry = image[0]
-        namespace = image[1]
+        namespace = "/".join(image[1:-1])
         repo_name = image[2]
```
vsoch: can we be sure that the string at index 2 is reliably the repo name? What if it looks like:

```
In [32]: image
Out[32]: ['registry:4000', 'hubby', 'two', 'imagename']

In [33]: len(image)>2
Out[33]: True

In [34]: image[0] #registry
Out[34]: 'registry:4000'

In [35]: image[2] #repo name?
Out[35]: 'two'

In [36]: image[-1] #repo name
Out[36]: 'imagename'

In [37]: image[1:-1] #namespace
Out[37]: ['hubby', 'two']
```

maybe to be consistent with referencing the namespace from index 1:-1 we should do the same for repo_name and use index -1? If tag/digest is removed, is there any chance the last in the list wouldn't be the reponame?

hartzell: Sorry, that's a typo/git foobar on my part, I'm still learning how to build/run in the dev tree and had been working on the installed copy (:scream:).

Just pushed the fix.

FWIW, here's what I'm actually running:

```diff
[vagrant@localhost singularity]$ diff /usr/local/libexec/singularity/python/shell.py libexec/python/shell.py
125c123
<     if len(image) > 2:
---
>     if len(image) == 3:
127,128c125,126
<         namespace = "/".join(image[1:-1])
<         repo_name = image[-1]
---
>         namespace = image[1]
>         repo_name = image[2]
```

vsoch:
ok this looks good to me! Note the difference between the first for uri `registry.docker.io:4000/one/two/three/imagename:last`

```
parse_image_uri('registry.docker.io:4000/one/two/three/imagename:last')
{'namespace': 'one/two/three',
 'registry': 'registry.docker.io:4000',
 'repo_name': 'two',
 'repo_tag': 'last',
 'version': None}
```
and after the fix:

```
{'namespace': 'one/two/three',
 'registry': 'registry.docker.io:4000',
 'repo_name': 'imagename',
 'repo_tag': 'last',
 'version': None}

```

as for what constitutes a namespace vs. registry, I haven't a clue anymore, haha.

vsoch (approved):

hartzell:
Just imagine trying to explain "Docker registries" and "Docker repositories" and "Artifactory repositories" to folks.

> We have an Artifactory repository that hosts a Docker registry that's full of Docker repositories, each of which holds all of the layers/tags/... for a set of related images.

[Who's on first?](https://youtu.be/kTcRRaXV-fg?t=1m29s)

vsoch:
@gmkurtzer ready for review

vsoch:
Actually there is one issue! The case if a user provides a long namespace, but no registry:

```
parse_image_uri('doesnthaveregistry/one/two/three/imagename@hash')
{'namespace': 'one/two/three',
 'registry': 'doesnthaveregistry',
 'repo_name': 'imagename',
 'repo_tag': 'latest',
 'version': 'hash'}
```

vsoch:
It would be the case that, given a weird / long namespace, the user would have to specify the registry.

vsoch:
And this one too:

```
In [23]: parse_image_uri('localhost:5000/ubuntu')
VERBOSE Namespace: localhost:5000
VERBOSE Repo Name: ubuntu
VERBOSE Repo Tag: latest
VERBOSE Version: None
Out[23]: 
{'namespace': 'localhost:5000',
 'registry': 'index.docker.io',
 'repo_name': 'ubuntu',
 'repo_tag': 'latest',
 'version': None}

```

bauerm97:
@vsoch @hartzell Perhaps a solution would be to say that, if the image URI starts with a "/", then the proceeding string is the namespace. If the URI starts without "/", then the proceeding string segment is the registry. For instance:

```
parse_image_uri('/doesnthaveregistry/one/two/three/imagename@hash')
{'namespace': 'doesnthaveregistry/one/two/three',
 'registry': 'index.docker.io',
 'repo_name': 'imagename',
 'repo_tag': 'latest',
 'version': 'hash'}
```
versus
```
parse_image_uri('localhost:5000/one/two/three/imagename@hash')
{'namespace': '/one/two/three',
 'registry': 'localhost:5000',
 'repo_name': 'imagename',
 'repo_tag': 'latest',
 'version': 'hash'}
```

vsoch:
@bauerm97 that's an interesting idea, but are you saying that it would change the default for most containers to be a bit weird? For example, most containers that users start with (eg, ubuntu, centos, python) don't have a registry specified, and the desire is the default. If they do as they do now, and say like:

```
singularity import docker://ubuntu:latest
```
with the above, that would first have the docker:// removed, leaving `ubuntu:latest`, and then ubuntu would be interpreted as a registry? I think the user would then have to do:

```
Bootstrap: docker
From: /ubuntu:latest
```

and

```
singularity import docker:///ubuntu:latest
```

I might be totally off - do you have the code you used to produce the above so I can try it out?

bauerm97:
@vsoch I have no code that produced what I showed, that was just me giving an example. I see what you're saying. You could potentially implement this behavior only when some condition is met, for example if there are a certain number of `/` characters, you interpret the URI a different way?

Or maybe, parse the URI in two steps. First, parse the URI to produce this:
```
parse_image_uri('doesnthaveregistry/one/two/three/imagename@hash')
{'namespace': 'one/two/three',
 'registry': 'doesnthaveregistry',
 'repo_name': 'imagename',
 'repo_tag': 'latest',
 'version': 'hash'}
```
Then do a sanity check on if the registry exists (send the registry a ping/dummy request/whatever). Then, if the registry turns out to not exist, parse the URI like this:
```
parse_image_uri('doesnthaveregistry/one/two/three/imagename@hash')
{'namespace': 'doesnthaveregistry/one/two/three',
 'registry': 'index.docker.io',
 'repo_name': 'imagename',
 'repo_tag': 'latest',
 'version': 'hash'}
```

Another alternative, is you could look for colons and periods in the section of the string from the start up to the first `/` character. Then you can assume that any string with colons or periods is a registry, not part of the namespace.

vsoch:
Yes exactly - requiring the user to specify another leading `/` to distinguish not having a registry (and using default) would change almost every use case (for example that has the docker hub registry) in requiring the user to specify another leading slash:

```
docker:///name/space/repo_name
```
If it were the case that this was the predominant use case, then it might be worth it. But this is the first time I've seen these namespaces, ever, so I don't think users would like such a drastic change to the default, and one that isn't intuitive. It would be like asking someone to specify https:/// if they wanted it to default to localhost, or something like that.

@bauerm97 that's a good idea to ping - there would be two options, either send an actual ping to the os, or [implement in python using sockets](https://gist.github.com/pyos/10980172). It still seems like a lot of overhead and additional calls for a rare use case - this would mean that every time a user requests layers from docker hub, we would ping index.docker.io first. 

>> Another alternative, is you could look for colons and periods in the section of the string from the start up to the first / character. Then you can assume that any string with colons or periods is a registry, not part of the namespace.

This is another good suggestion, but unfortunately not foolproof! I've seen plenty of people put weird characters in folder names, and even (gasp) spaces.

@hartzell I think we need a solution that will make it functional for your use case, but without adding baggage of excess calls / pings, and without changing the default / common use case.

vsoch:
I think the cleanest thing is to simply add another environment variable that specifies the specific paths. The parsing can do whatever default it does, and then given the variables are detected, check for and then use those at the end.

bauerm97:
I agree with you about `docker:///name/space/repo_name` not being a good thing. I also think that this is such a rare use case that very few people will need it. Your proposal is a good idea, but I don't think that it extends well to bootstrapping. Perhaps for a bootstrap definition file there are two syntaxes. The first syntax is legacy (what we have now):

```
Bootstrap: docker
From: namespace/image:latest
```

The new syntax allows detailed specification of the different parts of the URI:

```
Bootstrap: docker
docker_image: image
docker_tag: latest
docker_registry: localhost:5000
docker_namespace: namespace
```

Either syntax can be used, the second one simply allows for finer control of the URI

vsoch:
Did we ever figure out a solution for handling environment variables during bootstrap? Down the line I see many cases like this - if we get one solution to fix that problem, we solve all of them! And we don't need to hard code custom tags to look for.

bauerm97:
I don't think passing environment variables is a good solution to the bootstrapping use case. Everything necessary for bootstrapping should be able to be expressed in the definition file.

vsoch:
but arguably for other Singularity functions, you can pass things to `SINGULARITY_ENV` and it goes through. Are those very different?

I don't mean to say that we don't want to represent these variables in the definition file (we do) I mean to say that there is a use case of wanting (needing) an environment variable not for `%post`, but to actually influence singularity behavior (think things like binding, directory locations, etc.). The specific docker registry is such a use case. So instead of having a hard coded `docker_variable` header field, I was thinking something more flexible to these general usecases. Eg:

```
%environment build
DOCKER_REGISTRY=foo.bar.io
SINGULARITY_MOUNT_TMPFS=False
```

Does that make sense? We want everything expressed in the file, +1, but via a means that doesn't require hardcoding specific variables.

gmkurtzer:
Looks good to me, thanks!

Issue #868, opened 2017-08-12
Unable to auth against docker registry, Accept headers are "Not Acceptable"


hartzell:
## Version of Singularity:

development branch, commit 50838127aff0f70750a3696d831639c098860a5e

### Expected behavior

We use a Docker registry that's served by an Artifactory instance behind an Nginx reverse proxy (per artifactory's setup).

I expected to be able to import an image from our private docker registry.

### Actual behavior

```
[vagrant@localhost singularity]$ env SINGULARITY_DOCKER_USERNAME=hartzelg SINGULARITY_DOCKER_PASSWORD=$(cat pw) singularity --debug import bar.img  docker://blah.foo.com:5001/path/to/repo@sha256:530b5227f92f870aff05c040c5b1f27c15b5978dbf7704e877b71691ed58ede2
Enabling debugging
Ending argument loop
Singularity version: 2.3.1-development.g5083812
Exec'ing: /usr/local/libexec/singularity/cli/import.exec
VERBOSE2 SINGULARITY_COMMAND_ASIS found as False
VERBOSE2 SINGULARITY_ROOTFS not defined (None)
VERBOSE2 SINGULARITY_METADATA_FOLDER found as None/.singularity.d
VERBOSE2 SINGULARITY_FIX_PERMS found as False
VERBOSE2 SINGULARITY_COLORIZE not defined (None)
VERBOSE2 SINGULARITY_DISABLE_CACHE found as False
VERBOSE2 SINGULARITY_CACHEDIR found as /home/vagrant/.singularity
VERBOSE2 SINGULARITY_ENVIRONMENT found as None/.singularity.d/environment
VERBOSE2 SINGULARITY_RUNSCRIPT found as None/singularity
VERBOSE2 SINGULARITY_TESTFILE found as None/.singularity.d/test
VERBOSE2 SINGULARITY_DEFFILE found as None/.singularity.d/Singularity
VERBOSE2 SINGULARITY_HELPFILE found as None/.singularity.d/runscript.help
VERBOSE2 SINGULARITY_ENVBASE found as None/.singularity.d/env
VERBOSE2 SINGULARITY_LABELFILE found as None/.singularity.d/labels.json
VERBOSE2 SINGULARITY_INCLUDECMD found as False
VERBOSE2 SINGULARITY_NOHTTPS found as False
VERBOSE2 SINGULARITY_PULLFOLDER found as /home/vagrant/singularity
[...]

PR #869, opened 2017-08-12
adding application/json to end of header list

vsoch:

This is a tiny fix to add an `application/json` to accept headers, which I didn't know would be needed by local registries (note docker hub has notes to specify the latter to in order to ask for a version 2 manifest). Thanks to @hartzell for looking into this and figuring out the fix, the only adjustment I made was to put `application/json` last to specify the order of preference.

Attn: @singularityware-admin

### DIFF
--- a/libexec/python/docker/api.py
+++ b/libexec/python/docker/api.py
@@ -104,7 +104,7 @@ class DockerApiConnection(ApiConnection):
         # meaning the correct order of digests
         # returned (base to child)
 
-        return {"Accept": 'application/vnd.docker.distribution.manifest.v2+json,application/vnd.docker.distribution.manifest.list.v2+json',  # noqa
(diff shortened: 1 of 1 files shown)


hartzell:
See #868 for details.

Those are indeed the media types for the manifests (see the Media Types section of the Docker Registry docs [here](https://docs.docker.com/registry/spec/manifest-v2-2/)).

But they're also being used to request a token, which makes our Artifactory backed registry unhappy.

I'll see how things work with the extra type at the end.  Seems like it should work.

vsoch:
ok cool! Since most of our functions revolve around getting a manifest (possibly not token) this is why I implemented it like this. I believe the order of preference (application/json last) will ensure that an older version isn't returned, and (fingers crossed) the header will still work to get a token for your Artifactory.

hartzell:
Having `application/json` at the end of the list Works For Me(tm).

Thanks!

gmkurtzer:
Wahoo! Thanks!

vsoch:
thanks @gmkurtzer !

Issue #757, opened 2017-06-15
Too many levels of symbolic links


fmorency:
## Version of Singularity:

2.3

### Expected behavior

Using the singularity container while the image is on a network share mounter using autofs should work. This used to work with 2.2.1

### Actual behavior

Sometime, the container usage fails with error "Too many level of symbolic links" with the path pointing to the `.img` file. 

### Steps to reproduce behavior

On our end, this happens when we run our Nextflow processing pipeline on our SLURM mini-cluster and the image is on our SMB network share mounted using autofs. I know that Slack users gshamov and maswan both had the issue when using an automounter on Lustre/NFS. Here's an excerpt of the Slack discussion:

```
gshamov [2:10 PM] 
We have /sbb filestystem that is Lustre, that gets automounted per user as /global/scratch/user. When I bind both with -B and try to access /global/scratch/user I get " Too many levels of symbolic links" error.

maswan [2:41 PM] 
ugh, we've just been fighting that
we ended up with static nfs mounts
we think the kernel gets all confused at the point where autofs tries to unmount an "idle" filesystem that only has bind mounts and then gets all confused for future access to the filesystem, or something
worked fine for some filesystems, horribly for others
etc

fmorency [2:46 PM] 
I sometime see that error with autofs + cifs (smb3)
Still not sure what's the cause

maswan [2:48 PM] 
yeah, me neither
``` 

The automounter works fine when accessing the mounts from the terminal before/after the error. I'm not sure how to debug this and any help would be appreciated.

gmkurtzer:
Can you find a simple reproducer?

fmorency:
I will try to make one. I will ask gshamov and maswan if they can help with this.

fmorency:
I found a way to replicate the issue. I am not sure all the steps/details are required here, but this is how I could reproduce the issue:

Network share
  Samba 4.5.10 (CIFS) share from FreeNAS. I think the problem should also occur with any type of network share that make use of an automounter. We make use of Kerberos for authentification and we use some fancy mount options that, I think, are not relevant to this issue.


Host: Ubuntu 14.04
```
$ sudo apt-get install autofs
$ cd /
$ sudo mkdir test
```

Edit the file `/etc/auto.master` and add at the end of the file

  `/test   /etc/auto.test --ghost --timeout=60`

Edit the file `/etc/auto.test` and add

[...]

Issue #714, opened 2017-06-02
Singularity not preventing unmounts of mounted image directory


DrDaveD:
## Version of Singularity:

2.3

### Expected behavior

If the image that singularity runs is a directory, I expect to not be able to unmount the filesystem that the directory is in while singularity is running (without some kind of forcing option)

### Actual behavior

Singularity does not hold any references to the filesystem, so the filesystem can be unmounted.  The only references are inside the container, and at least on el7.3 that is not enough to be noticed outside the container.  I had reported this earlier in pr #626, but when I tested the development branch at the time there was a file descriptor left open, but now in the singularity-2.3 release the behavior is back the way it was in singularity-2.2.1.

### Steps to reproduce behavior

On EL7.3, assuming an existing singularity image file I call 'centos7.img', make a loopback filesystem, mount it, and use it:
```
dd if=/dev/zero of=centos7.ext3 bs=1G count=1
sudo losetup /dev/loop0 centos7.ext3
sudo mkfs -t ext3 -m 1 -v /dev/loop0
sudo mkdir -p /mnt/loop0
sudo mount /dev/loop0 /mnt/loop0
singularity export centos7.img | (cd /mnt/loop0 && sudo tar xf -)
singularity exec --containall /mnt/loop0 bash
```
Then in another window while singularity is running, umount /mnt/loop0.

It can be worked around by first doing 'cd /mnt/loop0' before running singularity exec (or inside singularity), or by having singularity open a file descriptor to the image directory and leaving it open.

This example of course is for reproducing, but in reality we experience the problem with automounted cvmfs; the automounter doesn't realize the filesystem is still in use, and unmounts it.

gmkurtzer:
Does omitting the `--containall` flag change the behavior?

DrDaveD:
No, the behavior is the same without --containall.  Can you reproduce it?

DrDaveD:
I see the same behavior on an EL6 host as well.

gmkurtzer:
Singularity currently does open a file descriptor to the image directory:

```
DEBUG   [U=1000,P=25681]   singularity_image_open()                  Opening file descriptor to image: /mnt/loop0/
```
But I had added the FD_CLOEXEC option to that file descriptor. I just updated the development branch, will you retest?

Thanks!

DrDaveD:
Yes, that did it.  Thanks!

[...]

Issue #347, opened 2016-11-27
Support for autofs?


oschulz:
I'm having trouble accessing auto-mounted directories within a Singularity container (we're using auto-mounting for several storage systems). The mounts are supposed to show up under `/remote`, and I bind-mount `/remote into the container.

If the directory in question (e.g. `/remote/my-auto-mount`) is already mounted on the host, all is fine. However, if it is not mounted on the host yet, and accessed in the container, I get this error:

    $ ls /remote/my-auto-mount
    ls: cannot open directory '/remote/my-auto-mount': Too many levels of symbolic links

However, this *does* trigger auto-mounting of the directory on the host. The mount doesn't become available in the running container (a second `ls` fails, too, with the same error) - but when I exit the container and enter it again, the mount is there and works. 

This is with Singularity v2.2 on Linux kernel 3.13, FS overlay is enabled.

Is it possible, in principle, to support auto-mounts in Singularity? Or maybe I just need a newer kernel?

bbockelm:
What OS is this?

I see the same behavior for RHEL6 (kernel 2.6.32) but not RHEL7 (3.10).  I wonder if the fix is in the patches that RedHat backports?

Unfortunately, this is a kernel bug and I haven't figured out a workaround on older platforms.

Brian

oschulz:
@bbockelm Thanks for the info! This was on Ubuntu 14.04 (trusty) with Kernel 3.13.0-98-generic. I've upgraded the system to LTE kernel 4.4.0-47-generic now (still Ubuntu trusty), but the problem persists.

I wonder why it works on RHEL7 with kernel 3.10, but not on Ubuntu trusty with kernel 4.4? Does RedHat backport some patch that's not even in fairly recent kernels, here? If so, I wonder if Ubuntu could be convinced to include that patch as well?

bbockelm:
That's a great question!  I haven't tried to bisect this issue (kernel development is not my forte...).

In the autofs setup I use, the path inside and outside the container are the same.  Is that true for you too?  If not, that might be another source of potential difference...

oschulz:
> In the autofs setup I use, the path inside and outside the container are the same. Is that true for you too?

Yes, I use the same path in the container as on the host.

bauerm97:
I can't provide a solution, but I can say that running on debian 8.5 with a slightly modified kernel we are able to bind mount an autofs mounted CVMFS directory successfully into containers.

oschulz:
[...]

PR #877, opened 2017-08-18
Autofs fix

cclerget:

- Commit a44d1932631e137326b280cae94cbdc0e107d4a4 address issues #714, #757, it opens image file/directory before entering in mount namespace
- Commit 445152390173becaa7a1b3ccaaf76bcad7a69bff address issues #757, #347 and adds a configuration option "autofs bug path" in singularity.conf which allow admins to add autofs mount directory generating **"Too many levels of symbolic links"** (ex: for bind path or user bind path). autofs option timeout=0 is not required anymore

- Ref: #714, #347, #757

**Checkoff for all PRs:**

Attn: @singularityware-admin


vsoch:
hey @cclerget could we do this PR against development branch?

src/lib/image/open/open.c
```diff
@@ -60,12 +60,10 @@ int _singularity_image_open(struct image_object *image, int open_flags) {
     }
 
     singularity_message(DEBUG, "Opening file descriptor to image: %s\n", image->path);
-    if ( ( image->fd = open(image->path, open_flags, 0755) ) < 0 ) {
+    if ( ( image->fd = open(image->path, open_flags|O_CLOEXEC, 0755) ) < 0 ) {
```
gmkurtzer: @cclerget, we have had problems in the past with SuSE not supporting this option to open(). Instead could we use fcntl()?

gmkurtzer:
@cclerget Is it possible to do these opens via the various bind functions so it happens automatically (instead of having to call a fixing function that relies on a config entry)?

Thanks, and great thinking!

src/lib/image/open/open.c
cclerget: Hi @gmkurtzer , I was asking why you used fcntl ! no problem to add fcntl

cclerget:
Hi @vsoch, of course, I will pull on dev branch

@gmkurtzer open must happens before entering mount namespace to work as expected, so open in bind functions has no effect for the bug.

At first I wrote autofs function which parsed bind path, userbinds, WORKDIR, SCRATCHDIR and automatically open dir/files when a boolean option in configuration file "autofs bug = yes" is enabled. There are drawbacks with this approach, if userbinds or bind path syntax change it imply to change code twice, same thing if a singularity env DIR is added. Or potentially reach file descriptor limit.

So the option to put buggy paths in configuration file should give admins more liberty, for devs less work

gmkurtzer:
@cclerget Ohhh, I get it now, and I understand why this was a problem now! Great catch :)

We have some conflicts with the development branch that need resolution.

cclerget:
@gmkurtzer should be ok for merge

gmkurtzer:
Looks great, thank you @cclerget!

PR #878, opened 2017-08-18
split singularity rpm package into singularity and singularity-runtime subpackages

DrDaveD:

This PR splits the singularity rpm into singularity & singularity-runtime subpackages.  The singularity-runtime package can be installed without singularity if no image manipulation is needed.  @loveshack did this first for singularity-2.2.1 in [EPEL](https://admin.fedoraproject.org/pkgdb/package/rpms/singularity/) and recently @efajardo did this for singularity-2.3.1 in [OSG](https://jira.opensciencegrid.org/browse/SOFTWARE-2755), and now I have done it for the current development branch so hopefully it doesn't have to be done again.  This feature was requested by @bbockelm.

A question I have is whether the inspect command should be in runtime.  I left it out for now.  Also possibly mount might be desired at runtime.  Possibly they and the start/stop/list commands should be listed under container usage commands instead of container management.

There is still room for improvement on this, but this is a start.  For instance it would be nice if action were split into separate commands for runtime and image manipulation as libsingularity is split between libsingularity-image and libsingularity-runtime.  Also it would be nice to split the cli commands into separate directories so they wouldn't have to be individually listed in singularity.spec.in, and likewise possibly the bin commands.  Changing the help message so it only lists available commands would also be a nice improvement.

None

**Checkoff for all PRs:**

   (Not relevant since it affects rpm only)

Attn: @singularityware-admin


gmkurtzer:
Heya @DrDaveD,

Great thinking! However... I think the runtime is the core component of Singularity and that the container builder and management functions should be the optional subpackage.

Thoughts?

DrDaveD:
No, I think that by default if someone does 'yum install singularity' they should get the complete tool.  That would be simplest and friendliest, it's not terribly big, and most people don't care about how much space it takes or the extra vulnerability from the extra suid programs.  Let's let people who know what they want make the different choice.

gmkurtzer:
I'm confused @DrDaveD,... While I agree with your point, this PR specifically separates the tool into two components: singularity and singularity-runtime. My point was just that if we go this route, I want the runtime to be the default singularity package, and then the build tools to be the subpackage, so we would end up with two packages:

    * singularity: this includes all necessary components for runtime
    * singularity-build: subpackage for all container maintainer/packaging scripts

But again, just to reiterate, I agree that I don't see the huge necessity to create subpackages.

DrDaveD:
@gmkurtzer The problem with splitting the packages that way is that if someone just installs "singularity" they won't get everything like they do now.  The way this PR does it, if someone installs "singularity" they get both packages, because singularity Requires singularity-runtime, and if someone just wants the subset they can install "singularity-runtime".  This split is backward compatible and more user-friendly.  Does it make sense now?

DrDaveD:
Here's perhaps a helpful motivation for having two subpackages: it reduces the security vulnerability for cases that do not need the image manipulation, because there are fewer suid files. This is just like what you talk about doing in your [security documentation](http://singularity.lbl.gov/docs-security#executable-permissions).

gmkurtzer:
Heya @DrDaveD,

I totally agree with the logic of why you did it this way as it doesn't break current updates, but I still think that the model isn't right. `yum install singularity` should get you the base essential functions, and a base package should never (according to RPM legacy and best practices) should not require a sub-package unless it is truly a runtime requirement.

Thanks!

bbockelm:
@gmkurtzer - talking to Dave about the pros / cons of the different approaches, one item we weren't able to resolve is the fact there *is* a legacy package here -- EPEL (https://koji.fedoraproject.org/koji/buildinfo?buildID=951232) and Fedora.  Even worse, it's effectively outside the singularity project's control.

Fedora / EPEL have the split between `singularity` and `singularity-runtime` analogous to what's in this PR.  If we want to keep a clean upgrade route for Fedora users to an RPM from upstream (really, a driving motivation here...), I'm not sure if there's a good alternate choice.

The "second best" one I can think of is have the `singularity` package contain all files and `singularity-runtime` contain only the minimum necessary for a worker node, then have the two packages be mutually exclusive (i.e., you cannot have both installed on the same system).  Provides the same level of choice to the paranoid sysadmin -- but also means `yum install singularity` has all the functionality a user expects.

Thoughts?

gmkurtzer:
Heya @DrDaveD and @bbockelm,

After some additional thinking about this, I see there is a benefit to packaging it as such. It is defying my packaging experience, but it does make sense for someone to use `yum install singularity` to obtain all components via a `Requires: ` tag to pull in the runtime, and if someone wants something atypical, they can specifically request it.

On the other hand, if someone just installs the `singularity-runtime` package (or any subpackage that doesn't have all of the command bits), the help output will not be in alignment with their installation, so some commands will error out as the UI is not designed at present to be broken into pieces. I would suggest then that the `libexec/cli/*.exec` files and help will somehow need to be "aware" of their surroundings and error out accordingly.

So, I would like to see that we fix the command output to be smarter or tolerant about dealing with uninstalled functions before we default to splitting the package via our SPEC.

Hopefully that makes sense. :)

Thanks!

DrDaveD:
@gmkurtzer Yes it makes sense.  I'll try to fix up the help messages.  I'm excluding the irrelevant libexec/cli/*.exec files so I don't think they need to be modified.  I suppose it's an alternative to include them and modify them, I'll see whether or not I think that's worth doing.

DrDaveD:
Ok @gmkurtzer I made the help smarter to show "(not installed)" on all the commands that are not installed.  How's that?  I think that's better than not showing them at all, because it will make people realize that they need to install another package if they want to use one of those commands.

PR #882, opened 2017-08-21
adding test for apps, fixing bug with sinit as well

vsoch:

This PR will add basic functionality for testing apps, along with fixing the bug with the missing sinit in the Makefile (this did not trigger an error on Travis as it was a warning during ./autogen.sh). Specifically, I test that:

- the structure of files / folders for apps and data is as expected
- the help command works for the container
- the help command works for an app when a `runscript.help` exists
- the help command works for an app when a `runscript.help` doesn't exist
- listing of apps contains the installed apps foo and bar
- inspecting an app returns labels specific to it
- running an app with a runscript runs it
- running an app without a runscript tells the user it doens't have one.

Attn: @singularityware-admin

### DIFF
--- a/examples/apps/Singularity
+++ b/examples/apps/Singularity
@@ -1,0 +1,37 @@
+Bootstrap: docker
+From: ubuntu:14.04
+
+%setup
+    echo "SETUP"
+
+%appinstall foo
+
+    echo "INSTALLING FOO"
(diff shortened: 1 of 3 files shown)


gmkurtzer:
Looks great, thank you!

Issue #897, opened 2017-08-25
Bootstrap from shub 


GodloveD:
Right now you can do something like this:
```
Bootstrap: docker
From: centos
```
or this:
```
Bootstrap: localimage
From: /my/local/image
```
Wouldn't it be great to do this?
```
Bootstrap: shub
From: singularityhub/centos
```
Will submit a pull request shortly to make it so.

PR #898, opened 2017-08-25
Btstrp from shub

GodloveD:

This PR allows users to bootstrap from shub using the following syntax in a def file:
```
Bootstrap: shub
From: author/image

%post
    echo "add more stuff here"
``` 
Just like the recent addition allowing users to bootstrap from an image on the local host, this addition will keep track of any previous definition files and store them with the metadata so that the users should be able to rebuild the container from scratch (assuming nothing important about the internet has changed 😜.  It will also warn the user if any env vars have changed during the bootstrap.   

- fixes Ref: #897 

**Checkoff for all PRs:**
✅   I have tested this PR locally with a `make test`
✅  I have added myself as a contributor to the [Author's file](https://github.com/singularityware/singularity/blob/master/AUTHORS.md)

Attn: @singularityware-admin


libexec/bootstrap-scripts/main-deffile.sh
```diff
@@ -50,7 +50,7 @@ if [ -n "${BOOTSTRAP:-}" -a -z "${SINGULARITY_BUILDNOBASE:-}" ]; then
 fi
 
 # take a snapshot of the environment for later comparison
-if [ ${BOOTSTRAP:-} = "localimage" ]; then
+if [ ${BOOTSTRAP:-} = "localimage" -o ${BOOTSTRAP:-} = "shub" ]; then
     SINGULARITY_STARTING_ENVIRONMENT=$(eval_abort env -i ${SINGULARITY_libexecdir}/singularity/helpers/record-env.sh ${SINGULARITY_ROOTFS})
```
vsoch: ok cool, so this deals with environment / labels in the same way as a localimage?

GodloveD: Yepper!  I think I might be able to add another `-o docker` and make it work in the same way for docker too.  But I haven't tried it yet.

vsoch:
hmm, well docker has different handling of metadata, labels, along with many layers and not one image. So just be careful, lol.

vsoch:
Okay first point of discussion! We need to decide on default behavior to do for labels. Right now, overwrite is set to False when it finds an equivalent error, and the user sees the error:

```
sudo singularity bootstrap hw.img Singularity.app 
Sanitizing environment
Exporting contents of shub://vsoch/hello-world to hw.img
Adding environment to container
Adding deffile section labels to container
ERROR MAINTAINER found in /usr/local/var/singularity/mnt/container/.singularity.d/labels.json and overwrite set to False.
Adding runscript
Found an existing definition file
Adding a bootstrap_history directory
Calculating final size for metadata...Environment variables were added, removed, and/or changed during bootstrap.
Variables unique to new image (hw.img)
FOO=BAR
Skipping checks
```
It is assumed (I think?) that the user would want to override the labels in the base with custom, so overwrite should be set to True (until someone complains for a custom environment variable).

Looking in the metadata folder - could you remind me again why we have the base Singularity file, one under "bootstrap" and a "bootstrap_history?"

```
Singularity hw.img:~> tree /.singularity.d/
/.singularity.d/
|-- Singularity
|-- actions
|   |-- exec
|   |-- run
|   |-- shell
|   `-- test
|-- bootstrap
|   `-- Singularity
|-- bootstrap_history
|   `-- Singularity0
|-- env
|   |-- 01-base.sh
|   |-- 10-docker.sh
|   |-- 90-environment.sh
|   |-- 95-apps.sh
|   `-- 99-base.sh
|-- environment -> env/90-environment.sh
|-- labels.json
|-- libs
`-- runscript

5 directories, 15 files
```
If bootstrap is indeed going to go away, we are going to have a lot of bootstrap folders that don't make sense! Do we need all three of those?

The labels look great! (minus the rewrite error)

```
{
    "org.label-schema.usage.singularity.deffile.bootstrap": "shub",
    "MAINTAINER": "vanessasaur",
    "SINGULARITY_DEFFILE_FROM": "ubuntu:14.04",
    "SINGULARITY_DEFFILE_BOOTSTRAP": "docker",
    "WHATAMI": "dinosaur",
    "SINGULARITY_DEFFILE": "/tmp/tmpz39taw4q/Singularity",
    "ANOTHERLABEL": "vanessasaur",
    "org.label-schema.build-date": "2017-08-25T18:25:39-04:00",
    "org.label-schema.schema-version": "1.0",
    "org.label-schema.usage.singularity.deffile.from": "vsoch/hello-world",
    "org.label-schema.usage.singularity.version": "2.3.1-GodloveD-btstrp-from-shub.ge35f610",
    "org.label-schema.build-size": "787MB",
    "SINGULARITY_BOOTSTRAP_VERSION": "2.3-master.gadf5259",
    "org.label-schema.usage.singularity.deffile": "Singularity.app"
}
```
So much good info there :)

GodloveD:
Hey @vsoch thanks for looking this over and let me apologize for the delay in getting back to you.  I'm happy to set overwrite to True in the driver file, but should it just be set globally to True?  I think overwriting is probably the expected behavior in most situations right?

As for the multiple bootstrap files, `bootstrap_history` does not house duplicate files.  It keeps a running tally of every bootstrap file that was used to create the current image.  So if you created an image on shub, and then used it to create another image, and then used that one to create another, and repeated 10x there would be 10 Singularity files in that directory all in order so that you could (in theory) recreate the container from start to finish.  

As for `bootstrap` I have no idea why that is there.  lol

hansolo

vsoch:
hey @GodloveD ! The overwrite still happens, but I set the warning to show because otherwise a metadata value would be changed without the user realizing it. The reason for setting that parameter is actually to go against the default and hide the warning message, specifically for use cases just like this one.

I do like `bootstrap_history`! It reminds me of `bash_history`. @gmkurtzer could you explain the difference between the bootstrap and Singularity files? I think it might be the case that we can get rid of bootstrap.

libexec/bootstrap-scripts/deffile-sections.sh
```diff
@@ -167,7 +167,12 @@ if [ -z "${SINGULARITY_BUILDSECTION:-}" -o "${SINGULARITY_BUILDSECTION:-}" == "l
 
         singularity_section_get "labels" "$SINGULARITY_BUILDDEF" | while read KEY VAL; do
             if [ -n "$KEY" -a -n "$VAL" ]; then
-                $ADD_LABEL --key "$KEY" --value "$VAL" --file "$SINGULARITY_ROOTFS/.singularity.d/labels.json"
+                if [ "${SINGULARITY_DEFFILE_BOOTSTRAP:-}" = "shub" -o "${SINGULARITY_DEFFILE_BOOTSTRAP:-}" = "localimage" ]; then
+                    $ADD_LABEL --key "$KEY" --value "$VAL" --file "$SINGULARITY_ROOTFS/.singularity.d/labels.json" -f "False"
+                else
```
vsoch: Hey @GodloveD ! The force argument is a boolean, so you just need to add the flag (without any True/False), eg, here is the definition in python:
```
    parser.add_option('-f', dest="force",
                      help="force add (overwrite if exists)",
                      default=False, action='store_true')
```
and then it would just work like:

```
 $ADD_LABEL --key "$KEY" --value "$VAL" --file "$SINGULARITY_ROOTFS/.singularity.d/labels.json" -f
```

vsoch: Did you try it with the "True" and does it error? Or just ignore it?

GodloveD: Of course you are right.  duh.  Thanks.

vsoch: haha no worries! There isn't clear documentation about this... it's all IN MY HEAD. :scream:

PR #901, opened 2017-08-29
Squash from localfile

GodloveD:
Trying to add support for building from local file (which now that I think about it, is basically just the `cp` command).


libexec/bootstrap-scripts/deffile-driver-localimage.sh
```diff
@@ -46,9 +46,7 @@ fi
 umask 0002
 
 message 1 "Exporting contents of ${FROM} to ${SINGULARITY_IMAGE}\n"
-
-cmd="${SINGULARITY_libexecdir}/singularity/bin/export ${FROM} | (cd ${SINGULARITY_ROOTFS} && tar xBf -)"
-if ! eval $cmd; then
+if ! eval "${SINGULARITY_bindir}"/singularity image.export "${FROM}" | (cd "${SINGULARITY_ROOTFS}" && tar xBf - -C "${SINGULARITY_ROOTFS}"); then
```
gmkurtzer, line 49: I believe we can get rid of the `cd ...` now that we are using the `-C` option to tar.

libexec/bootstrap-scripts/main-deffile.sh
```diff
@@ -53,7 +53,7 @@ if [ -n "${BOOTSTRAP:-}" -a -z "${SINGULARITY_BUILDNOBASE:-}" ]; then
 fi
 
 # take a snapshot of the environment for later comparison
-if [ ${BOOTSTRAP:-} = "localimage" ]; then
+if [ "x${BOOTSTRAP:-}" = "xlocalimage" ]; then
```
gmkurtzer: nitpick here,... if the variable is quoted, there is no need to append a character ('x') to the strings.

libexec/cli/bootstrap.exec
```diff
@@ -0,0 +1,132 @@
...
+# See the COPYRIGHT.md file at the top-level directory of this distribution and at
+# https://github.com/singularityware/singularity/blob/master/COPYRIGHT.md.
+#
+# This file is part of the Singularity Linux container project. It is subject to the license
+# terms in the LICENSE.md file found in the top-level directory of this distribution and
+# at https://github.com/singularityware/singularity/blob/master/LICENSE.md. No part
+# of Singularity, including this file, may be copied, modified, propagated, or distributed
+# except according to the terms contained in the LICENSE.md file.
+#
+# This file also contains content that is covered under the LBNL/DOE/UC modified
```
gmkurtzer: Because this is a (re)new file, and it has no LBNL copyright, we don't need to include this blurb.

src/builddef.c
```diff
@@ -136,6 +136,7 @@ int main(int argc, char **argv) {
     envar_set("PATH", "/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin", 1);
     envar_set("SINGULARITY_ROOTFS", CONTAINER_MOUNTDIR, 1);
     envar_set("SINGULARITY_libexecdir", singularity_registry_get("LIBEXECDIR"), 1);
+    envar_set("SINGULARITY_bindir", singularity_registry_get("BINDIR"), 1);
```
gmkurtzer, line 139: We could do this another way too... If you look at src/Makefile.am, you may notice the AM_CPPFLAGS make it possible to define this statically to be included within the source file (e.g. can't be externally modified). When we spoke about this earlier, I didn't consider making it completely static, but I think now it would be a good idea. Let's do that instead of setting this through the registry (which can be modified).

GodloveD: OK.  I've tried a bunch of stuff, but I'm not seeing how to do this (although it seems like it should be trivial).  Can you give me another hint?

gmkurtzer: This isn't really a hint, more so an implementation, but check out https://github.com/singularityware/singularity/commit/49465e51256625d3cd719237503481f9eb028867. Hope that helps for next time!

gmkurtzer:
This looks awesome, thanks!

gmkurtzer:
Oh, I meant to ask, let's make some some additional tests that make sure that this is working properly in the `tests/`. Let me know if you have any questions there.

PR #904, opened 2017-09-05
making tweaks to appbase to better expose paths as variables

vsoch:

These are a few tweaks to scif apps to make working with them easier. Likely I'll have more to come, but these are some essential that should be added right away! Specifically:

## Environment
We need complete exposure of different application locations, otherwise it's not so useful to have the standard. We also should be consistent in the format of the metadata folder. Toward this goal, I first made the metadata folder in the app (`/scif`) have similar structure to the one at the base of the container (`.singularity.d`) meaning that the environment scripts live in a subfolder, `env`. I also separated out the user specific application environment variables (`90-environment.sh`) from the base of the app (`01-base.sh`) into which each app has defined the following:

```
SINGULARITY_APPOUTPUT=/scif/data/example/output
SINGULARITY_APPDATA=/scif/data/example
SINGULARITY_APPBASE=/scif/apps/example
SINGULARITY_APPINPUT=/scif/data/example/input
SINGULARITY_APPROOT=/scif/apps/example
SINGULARITY_APPNAME=example
SINGULARITY_APPMETA=/scif/apps/example/scif
```

Thus, the initialization of an application now creates a `01-base.sh` environment file, and a `90-environment.sh` file if the user has an `%appenvironment foo` section. 

As an example, I can now write a very simple application (only has an `%apprun` section) and easily get an environment variable in the container without any pipe / shell nonsense. Here is the section in the recipe:

```
%apprun env
#!/bin/sh
exec printenv $@
```

and then use:

```
singularity run --app env lg.img SINGULARITY_APPDATA
/scif/data/env
```
There are many different functions / modules / apps (etc) that one might right similar to this, just to make interaction with containers easier! I'll hopefully have some cool examples of what apps can do in the next few weeks.

## Lib and Bin
I hadn't initially thought of adding an app's bin to the `$PATH` and lib to `$LD_LIBRARY_PATH` but loved the idea after @gmkurtzer suggested and implemented. This is a great idea because it will encourage users to develop in a simple manner to put executables in the bin, and libraries in lib. And then they can be predictably found (and parsed) by a user that doesn't know how the container was made! To further encourage this, and make installation ridiculously easy, I decided to create these directories automatically on an application init. Thus, the entire application creation process could be as simple as dynamically creating a file:

```
%appinstall example
    echo "echo 'Hello World!' $@" > bin/run-example.sh
    chmod u+x bin/run-example.sh
```
or by adding the file via files (which also assumes the app base as the `$PWD`

```
%appfiles example
run-example.sh bin/run-example.sh
```
That's pretty awesome :)

Attn: @singularityware-admin

### DIFF
--- a/libexec/bootstrap-scripts/deffile-sections.sh
+++ b/libexec/bootstrap-scripts/deffile-sections.sh
@@ -294,6 +294,12 @@ if [ -z "${SINGULARITY_BUILDSECTION:-}" -o "${SINGULARITY_BUILDSECTION:-}" == "a
                 chmod 0755 "$SINGULARITY_ROOTFS/scif/apps/${APPNAME}/scif/runscript"  
             fi
 
+            # Make sure we have metadata
+            APPBASE="$SINGULARITY_ROOTFS/scif/apps/${APPNAME}"
+            APPFOLDER_SIZE=$(singularity_calculate_size "${APPBASE}")
+            $ADD_LABEL --key "SINGULARITY_APP_SIZE" --value "${APPFOLDER_SIZE}MB" --file "$APPBASE/scif/labels.json"
+            $ADD_LABEL --key "SINGULARITY_APP_NAME" --value "${APPNAME}" --file "${APPBASE}/scif/labels.json"
+
         done
     fi
 fi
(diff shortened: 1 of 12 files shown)


vsoch:
@gmkurtzer ready for review!

vsoch:
@singularityware/singularity-admin @gmkurtzer this is still ready for review!

gmkurtzer:
Excellent, thanks @vsoch!

PR #911, opened 2017-09-07
Add UTS namespace

cclerget:

It adds UTS namespace. After some tests to launch /sbin/init as root within an daemon instance show init (systemd) modifying "host" hostname based on container's /etc/hostname. So it could be useful and required for this use case. UTS is enabled by default even if it runs as root or normal user.

**Checkoff for all PRs:**

Attn: @singularityware-admin


src/lib/runtime/ns/uts/uts.c
```diff
@@ -0,0 +1,90 @@
...
+ * your rights to use or distribute this software.
+ * 
+ * NOTICE.  This Software was developed under funding from the U.S. Department of
+ * Energy and the U.S. Government consequently retains certain rights. As such,
+ * the U.S. Government has been granted for itself and others acting on its
+ * behalf a paid-up, nonexclusive, irrevocable, worldwide license in the Software
+ * to reproduce, distribute copies to the public, prepare derivative works, and
+ * perform publicly and display publicly, and to permit other to do so. 
+ * 
+*/
```
bauerm97: This should be changed to the new copyright header format

src/lib/runtime/ns/uts/uts.h
```diff
@@ -0,0 +1,30 @@
...
+ * your rights to use or distribute this software.
+ * 
+ * NOTICE.  This Software was developed under funding from the U.S. Department of
+ * Energy and the U.S. Government consequently retains certain rights. As such,
+ * the U.S. Government has been granted for itself and others acting on its
+ * behalf a paid-up, nonexclusive, irrevocable, worldwide license in the Software
+ * to reproduce, distribute copies to the public, prepare derivative works, and
+ * perform publicly and display publicly, and to permit other to do so. 
+ * 
+*/
```
bauerm97: As stated in the source file, this copyright header should be updated to the new version

bauerm97 (changes requested):
Everything looks good to me, besides the copyright headers. I'm wondering if we always want to be invoking the UTS namespace or if it should only be invoked on the users request. If it should only be at user request, we should add a --uts option to the action_argparser.sh in libexec/cli/action_argparser.sh to allow the user to toggle this

cclerget:
This namespace is required as long container is ran as root, so maybe a conditional statement if uid == 0 ?

cclerget:
@bauerm97 up, finally enforce UTS namespace when container is run as root and add --uts option for user request

cclerget:
Close PR, I'll merge it with later

PR #916, opened 2017-09-08
Squash docs

GodloveD:
Bringing docs and usage into closer alignment.  There are still some things to be done before merging (fixing issues using spaces for subcommands) but we can start discussing the proposed changes right now.


gmkurtzer:
Looks great, merging. Thanks!

PR #923, opened 2017-09-13
quick fix for docker import / pull

vsoch:

In case this is needed, it's a quick fix to force the older manifest and correctly get the layers. Note that this doesn't parse the "lists of manifests" but forces return of the older version, the one that doesn't have size information. @gmkurtzer we talked about this yesterday - bug if we have some logic for choosing a particular architecture we can use the newest "manifest list." With this current workaround (forcing version 1) we may run into the issue of losing the automatically calculated size for images where this applies, but since we are moving to squashfs this probably isn't a huge issue. I'm also not sure the extent to which we want to have a fix if the new release is coming out soon? (given that users and admins would need to re-install anyway). For those that have dire need, we might even recommend development branch, and for others, to wait until 2.4. @GodloveD and @gmkurtzer @bauerm97 let me know your thoughts!

Attn: @singularityware-admin

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,9 @@ src/get-section
 src/stamp-h1
 src/include/stamp-h1
 src/.dirstamp
+src/sinit
+src/start
+src/start-suid
 
 src/.deps/
 src/.libs/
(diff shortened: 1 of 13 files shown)


vsoch:
looks like shub pull has a bug one sec.

gmkurtzer:
It looks good to me, and passing my dumb user tests. Thank you!

PR #926, opened 2017-09-13
adding support for manifest lists, and version specification

vsoch:

This is an update to the initial fix to allow for the new version of the manifest. Specifically, the following happens:

 **1**. The `update_manifests` functions is expected to provide two manifests:
   - manifestv1: the oldest version, which notably is the only one with environment / labels (and other metadata), and also has layers
   - manifest (version 2+): This version has the layers **and** sizes.

Our old default was to retrieve the version 2, and ask it for layers (and by default our header asks for version 2+). With the update, the API now returns a [LIST of manifests](https://docs.docker.com/registry/spec/manifest-v2-2/#manifest-list). This broke most import / shell / exec because we needed to first choose one in the list, then retrieve the manifest, then grab the layers. Thus,

**2**. The update_manifest function uses the environment variables (`SINGULARITY_DOCKER_OS` and `SINGULARITY_DOCKER_ARCHITECTURE` with defaults `linux` and `amd64`) as keys to find the manifest digest. The user can also export these variables to ask for a different one.

**3**. If the manifest is found and obtained, great, we use it for layers and it also (still has size). Note that for the quick fix fallback, we just resorted to version 1 anyway, and didn't have size.

**4**. If the manifest isn't found, we fall back to the version 1. It will provide metadata and layers, but not size.

At runtime, it looks like this:

```
DEBUG 
*** STARTING DOCKER IMPORT PYTHON  ****
```
This is the initialization of the Docker client, it's parsing the image name provided
```
DEBUG Starting Docker IMPORT, includes env, runscript, and metadata.
VERBOSE Docker image: ubuntu:14.04
VERBOSE2 Specified Docker ENTRYPOINT as %runscript.
DEBUG Headers found: Content-Type,Accept
VERBOSE Namespace: library
VERBOSE Repo Name: ubuntu
VERBOSE Repo Tag: 14.04
VERBOSE Version: None
```
The initial poke to Docker hub asks for tags, to determine if we need some kind of token
```
VERBOSE Obtaining tags: https://index.docker.io/v2/library/ubuntu/tags/list
DEBUG GET https://index.docker.io/v2/library/ubuntu/tags/list
```

401 means that we do, and with the response the API sends back the scope and other details we need to make to request it

```
DEBUG Http Error with code 401
```

Here we are requesting it
```
DEBUG GET https://auth.docker.io/token?service=registry.docker.io&expires_in=9000&scope=repository:library/ubuntu:pull
DEBUG Headers found: Content-Type,Authorization,Accept
```
This means that we sent the correct header this time, and this is a public image
```
VERBOSE3 Response on obtaining token is None.
```
Here is the new bit. Instead of blindly doing one call, we have a function to update two versions of the manifest - the older (with metadata) and some version of the newer (with layers and size) with a fallback to use the version 1
```
DEBUG Updating manifests.

# Here is version 2+
DEBUG MANIFEST (Primary): not found, making initial call.
VERBOSE Obtaining manifest: https://index.docker.io/v2/library/busybox/manifests/latest
DEBUG GET https://index.docker.io/v2/library/busybox/manifests/latest

# Here is version 1 (Metadata)
DEBUG MANIFEST (Metadata): not found, making initial call.
VERBOSE Obtaining manifest: https://index.docker.io/v2/library/busybox/manifests/latest
DEBUG GET https://index.docker.io/v2/library/busybox/manifests/latest
```
Notice that the two calls are the same - that's because the headers are actually different, to request different ones.

And here is the (new) indication that we found a list
```
DEBUG Image manifest version 2.2 list found.
```
Here is the default architecture / os
```
DEBUG Obtaining architecture: amd64, OS: linux
```
And the specific call to get it
```
VERBOSE Obtaining manifest: https://index.docker.io/v2/library/busybox/manifests/sha256:030fcb92e1487b18c974784dcc110a93147c9fc402188370fbfd17efabffc6af
DEBUG GET https://index.docker.io/v2/library/busybox/manifests/sha256:030fcb92e1487b18c974784dcc110a93147c9fc402188370fbfd17efabffc6af
```

So - with this fix - we can still have a working `singularity pull docker://busybox` that can get the size automatically. 

Attn: @singularityware-admin

### DIFF
--- a/libexec/python/README.md
+++ b/libexec/python/README.md
@@ -72,16 +72,22 @@ For each, the details of required arguments are detailed in the scripts, and dis
 
 
 ### Defaults
-The following variables in [defaults.py](defaults.py) are static values that do not change. You probably don't care much about these, but they are included for reference.
-
+The following variables in [defaults.py](defaults.py) are a combination of static values, and variables that can be customized by the user via environment variables at runtime. 
 
 #### Docker
 
-**API_BASE** 
-Set as `index.docker.io`, which is the name of the registry. In the first version of Singularity we parsed the Registry argument from the build spec file, however now this is removed because it can be obtained directly from the image name (eg, `registry/namespace/repo:tag`)
+**DOCKER_API_BASE** 
+Set as `index.docker.io`, which is the name of the registry. In the first version of Singularity we parsed the Registry argument from the build spec file, however now this is removed because it can be obtained directly from the image name (eg, `registry/namespace/repo:tag`). If you don't specify a registry name for your image, this default is used.
 
-**API_VERSION**
+**DOCKER_API_VERSION**
 Is the version of the Docker Registry API currently being used, by default now is `v2`.
+
+**DOCKER_OS**
+This is exposed via the exported environment variable `SINGULARITY_DOCKER_OS` and pertains to images that reveal a version 2 manifest with a [manifest list](https://docs.docker.com/registry/spec/manifest-v2-2/#manifest-list). In the case that the list is present, we must choose an operating system (this variable) and an architecture (below). The default is `linux`.
+
+**DOCKER_ARCHITECTURE**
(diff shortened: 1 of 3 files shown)


gmkurtzer:
WAHOOOEY!

PR #937, opened 2017-09-18
Merge/devtomaster

vsoch:

Okay, the other one might be right too, because this looks the same. This is merging development into master, and then now a PR to (the old) development. I think that's what we wanted @gmkurtzer ?

Attn: @singularityware-admin


libexec/python/helpers/json/main.py
```diff
@@ -89,6 +89,12 @@ def GET(key, jsonfile):
 def ADD(key, value, jsonfile, force=False, quiet=False):
     '''ADD will write or update a key in a json file
     '''
+
+    # Check that key is not empty
```
vsoch, line 93: @GodloveD this should resolve the "empty labels" issue. By the time we get here, the calling function (add.py) has ensured we don't have a value of None, and the arg type is string so the "strip" shouldn't trigger an error. Then given that the value is # or "" (empty), instead of continuing we exit without error.

PR #938, opened 2017-09-19
tweaking pull folder for build

vsoch:

This will change the build.exec to, when using a shub image to build from, import it to the following places, ordered by preference:

 - if the user has defined the SINGULARITY_PULLFOLDER, honor it
 - if not, default to cache subfolder "shub"

Attn: @singularityware-admin


gmkurtzer:
Merged, and rockin'! Thank you!

PR #942, opened 2017-09-19
Unset (and reset) before sourcing environment

### LINKED ISSUE #941: Unbound variable error while sourcing the environment during the build process
## Version of Singularity:

2.3.9-development.gad1efde

### Expected behavior

Be able to `export FOO=/bar:$FOO` in the `%environment` section even if `FOO` is undefined. This behaviour is working on 2.3

### Actual behavior

An `unbound variable` error is raised during the build process if the variable is undefined

Add `export FOO=/bar:$FOO` in the `%environment` section and `singularity build` the image

### Notes

After chatting with @vsoch and @GodloveD on Slack, we came to the conclusion that `set -u` should be disabled (`set +u`) in the section of the code where the environment is sourced.

vsoch:

This puts a set +u and set -u wrapper around sources to the environment, which we didn't do before but now are essential so that custom user environments (with undefined variables) don't spit out an error. This will close #941 

Attn: @singularityware-admin


vsoch:
@gmkurtzer this is a tiny enough fix that you should be able to glance it over quickly, I either did it correctly or not, and then approve/deny.

gmkurtzer:
@vsoch Yep, the fix looks good and I suppose I can not enforce pedantics on our users! lol

Thanks!

PR #944, opened 2017-09-19
Continue if there's an EINVAL error from mounting devpts

DrDaveD:

This makes singularity continue silently if mounting devpts fails with an EINVAL error.  This can happen on RHEL7.4 when running unprivileged.

None.

**Checkoff for all PRs:**

Attn: @singularityware-admin


gmkurtzer:
Looks good! Thanks!

olifre:
Hi!
Isn't this "fix" the wrong approach, though? 
As discussed for example here:
https://github.com/opencontainers/runc/issues/225
mounting `devpts` inside a user namespace works just fine, it's just (expectedly) not allowed to specify an unmapped `gid` when performing the mount, hence the `EINVAL`. 
I tried this on a fresh system with Kernel 4.14, also here, it fails - but when I patch the singularity sources to not specify `gid=<tty>`, it works as expected. I did not test on CentOS 7.4 yet, but it would be strange if the cause of the issue would be a different one. 

So the correct fix would be to drop the `gid` enforcement in case we run an unprivileged container. 
Of course, this means things like `sshd` which explicitly check for ownership of the devices will still fail - even though there is no permission issue. But this is probably rather an issue with `sshd` and other apps. 
A workaround which could be used even for unprivileged containers is documented here:
https://github.com/hpc/charliecloud/issues/67

PR #951, opened 2017-09-21
scratch to warn without infinite loop

GodloveD:

The error is described in the Google Group [here](https://groups.google.com/a/lbl.gov/forum/#!topic/singularity/BCcV4DiLU04).  The short story is that if you set `enable overlay = no` in `singularity.conf` and then attempt to use the `-S` option, you will get caught in an infinite loop of warning messages.  I think this should fix the issue.

Attn: @singularityware-admin


gmkurtzer:
Great find, thanks!

PR #960, opened 2017-09-25
Help fixups

GodloveD:

These are mostly just a bunch of stuff to fix the help docs but there are also a few small bug fixes as well.  This PR does not address the proposed changes that resulted from discussion of https://github.com/singularityware/singularity/issues/958.  Those will be added in a separate PR for clarity.

Attn: @singularityware-admin


libexec/cli/bootstrap.exec
```diff
@@ -47,7 +47,7 @@ while true; do
             SINGULARITY_NOTEST=1
             export SINGULARITY_NOTEST
         ;;
-        -f|--force)
+        -f|-F|--force)
             shift
```
vsoch, line 51: lol, so I might use CAPITAL F when I'm REALLY ANGRY AT BOOTSTRAP.

libexec/cli/bootstrap.help
```diff
@@ -29,14 +29,17 @@ CREATE OPTIONS:
...
     -c|--checks    enable checks
     -t|--tag       specify a check tag (not default)
-    -l/--low       Specify low threshold (all checks, default) 
-    -m/--med       Perform medium and high checks
-    -h/--high      Perform only checks at level high
+    -l|--low       Specify low threshold (all checks, default) 
+    -m|--med       Perform medium and high checks
+    -h|--high      Perform only checks at level high
 
 See singularity check --help for available tags
```
vsoch, line 44: I'm looking forward to getting back to these after SCi-F :)

libexec/cli/bootstrap.help
```diff
@@ -101,6 +104,22 @@ DEFFILE SECTION EXAMPLES:
...
+    %apphelp app1
+        This is a help doc for running app1
+
+    %apprun app1
+        echo "this is a runscript for app1"
+
+
 EXAMPLES:
 
     $ singularity create /tmp/Debian.img
```
vsoch, line 125: should we remove the create from examples? Are we doing that anymore?

libexec/cli/build.help
```diff
@@ -118,6 +118,22 @@ DEFFILE SECTION EXAMPLES:
...
+    %appinstall app1
+        echo "These are steps to install an app using the SCI-F format"
+
+    %appenv app1
+        APP1VAR=app1value
+        export APP1VAR
+
+    %apphelp app1
+        This is a help doc for running app1
+
```
vsoch, line 132: for this an the above, if you are doing a full example, might make sense to include `%applabels` and `%appfiles`. Also included (and not likely to be used) are for setup and test too. Maybe we can put the major ones here, and link to your cowsay example (with the more "advanced" usage there?

libexec/cli/bootstrap.exec
GodloveD: lol.  It's just for consistency.  In other commands the short version of `--force` is `-F`, so I don't know why it was lowercase to begin with.

libexec/cli/image.expand.help
```diff
@@ -4,7 +4,7 @@ USAGE: singularity [...] image.expand [expand options...] <container path>
 Grow the Singularity disk image by the given amount.
 
 EXPAND OPTIONS:
-    -s/--size   Specify a size for an operation in MiB, i.e. 1024*1024B
+    -s|--size   Specify a size for an operation in MiB, i.e. 1024*1024B
                 (default 768MiB)
```
vsoch, line 8: Does this make sense in context of squashfs? Is it only for old format, or are we allowing (somehow) images to be expanded?

libexec/cli/bootstrap.help
GodloveD: We are still doing create before a bootstrap.  At least that is the way I set it up with the `--exists` option.  This may change when I remove the option.  But anyway, the idea is that many folks use scripts to `--create` and `--bootstrap` containers and we want to give them warnings before we jerk the rug out from under them.

libexec/cli/build.help
GodloveD: Yes.  Can you point me to a place where I can see an example of these sections too in the docs so that I can add to the help?

libexec/cli/image.expand.help
GodloveD: Old format.  Note that there is now an `image` command group (and that there are now "command groups").  The `image` command group is for working with legacy, writable images.  

Also, note that the writable image format is not really legacy in the sense that it is backward compatible.  You can't open an writable image in an old version of singularity, so it is not backward compatible.  But it does preserve the same behavior.

gmkurtzer:
Thank you!

Issue #970, opened 2017-09-25
apps bases need to be sourced in environment BEFORE app specific files


vsoch:
Right now, the base exports are appended to the apps file. This isn't ideal because we need them sourced before.

gmkurtzer:
This has been merged, can we close?

PR #971, opened 2017-09-25
ensuring app bases are sourced before specific app

vsoch:

This is a quick fix to ensure that app bases (for data and executables) are sourced in the environment BEFORE the app specific environments. This is because it is common to want to reference another app in the environment, eg, here I want to use a file from `reference` during `mapping` and the file should predictibly be at `$APPROOT_reference`

```
%appenv mapping
    all_reads=${SINGULARITY_DATA}/all_reads.fastq
    reference_genome=${APPROOT_reference}/lambda_ecoli.fa
    bwa_threads="${CSEQ_BWATHREADS:-1}"
    q_score="${CSEQ_QSCORE:-9}"
    output_folder="${SINGULARITY_APPDATA}"
    export all_reads reference_genome bwa_threads q_score output_folder
```

but as it is now, we haven't defined the `APPROOT_reference` until after calling the above, so we don't see it. The variable is empty and it looks like the file is at the room:

```
singularity exec --app mapping carrierseq.img env | grep reference
APPDATA_reference=/scif/data/reference
APPROOT_reference=/scif/apps/reference
reference_genome=/lambda_ecoli.fa
```
 but with the fix, we do!

```
singularity exec --app mapping carrierseq.img env | grep reference
APPDATA_reference=/scif/data/reference
APPROOT_reference=/scif/apps/reference
reference_genome=/scif/apps/reference/lambda_ecoli.fa
```

- Ref: #970 

Attn: @singularityware-admin

### DIFF
--- a/libexec/bootstrap-scripts/deffile-sections.sh
+++ b/libexec/bootstrap-scripts/deffile-sections.sh
@@ -378,9 +378,9 @@ for app in ${SINGULARITY_ROOTFS}/scif/apps/*; do
     if [ -d "$app" ]; then
         app="${app##*/}"
         app=(`echo $app | sed -e "s/-/_/g"`)
-        echo "APPDATA_$app=/scif/data/$app" >> "$SINGULARITY_ROOTFS/.singularity.d/env/95-apps.sh"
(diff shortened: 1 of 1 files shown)


gmkurtzer:
Looks good!

Issue #977, opened 2017-09-26
pull permissions errors


vsoch:
The Singularity pull command is no longer working, there are weird permissions errors. Likely it didn't get updated with build, or uses a component of it that no longer works?

## Version of Singularity:

singularity --version
2.3.9-development.g3dafa39

singularity pull docker://ubuntu
/.ubuntu.img

```
singularity pull docker://ubuntuCreating empty 226MiB image file: ./ubuntu.img
Formatting image with ext3 file system
Image is done: ./ubuntu.img
Docker image path: index.docker.io/library/ubuntu:latest
Cache folder set to /home/vanessa/.singularity/docker
[5/5] |===================================| 100.0% 
Importing: base Singularity environment
tar: ./etc: Cannot mkdir: Permission denied
tar: ./etc: Cannot mkdir: Permission denied
tar: ./etc/hosts: Cannot open: No such file or directory
tar: ./etc: Cannot mkdir: Permission denied
tar: ./etc/resolv.conf: Cannot open: No such file or directory
tar: ./environment: Cannot create symlink to ‘.singularity.d/env/90-environment.sh’: Permission denied
tar: ./.shell: Cannot create symlink to ‘.singularity.d/actions/shell’: Permission denied
tar: ./dev: Cannot mkdir: Permission denied
tar: ./sys: Cannot mkdir: Permission denied
tar: ./.test: Cannot create symlink to ‘.singularity.d/actions/test’: Permission denied
tar: ./.exec: Cannot create symlink to ‘.singularity.d/actions/exec’: Permission denied
tar: ./root: Cannot mkdir: Permission denied
tar: ./.run: Cannot create symlink to ‘.singularity.d/actions/run’: Permission denied
tar: ./var: Cannot mkdir: Permission denied
tar: ./var: Cannot mkdir: Permission denied
tar: ./var/tmp: Cannot mkdir: No such file or directory
tar: ./proc: Cannot mkdir: Permission denied
tar: ./tmp: Cannot mkdir: Permission denied
tar: ./singularity: Cannot create symlink to ‘.singularity.d/runscript’: Permission denied
tar: ./.singularity.d: Cannot mkdir: Permission denied
tar: ./.singularity.d: Cannot mkdir: Permission denied
tar: ./.singularity.d/env: Cannot mkdir: No such file or directory
tar: ./.singularity.d: Cannot mkdir: Permission denied
tar: ./.singularity.d/env/95-apps.sh: Cannot open: No such file or directory
[...]

PR #979, opened 2017-09-27
changing pull to a wrapper for build (docker hub)

GodloveD:

Because `import` and `export` no longer escalate privs, `pull` from Docker Hub no longer works properly as a non-privileged user.  You can now `build` from Docker Hub rendering the `pull` command obsolete, and it uses a slightly different method.  Instead of importing a tar pipe into a pre-made image, build unpacks to a temporary directory and then uses `mksquashfs` to create an image.  This path results in fewer errors when used by a non-prived user.  

So this PR will deprecate pull from docker by making it a wrapped build command instead.  The only trade-off is that images built from Docker Hub will now be squashfs instead of writable image format.  But this is what we want anyway.  If a user wants to convert the squashfs image to writable they can also do that with the build command, but will need to be root. 

- Ref: #977

**Checkoff for all PRs:**

:white_check_mark: I have tested this PR locally with a `make test`
:white_check_mark:  This PR is NOT against the project's master branch
:white_check_mark:  have added myself as a contributor to the [Author's file](https://github.com/singularityware/singularity/blob/master/AUTHORS.md)
:white_check_mark: This PR is ready for review and/or merge

Attn: @singularityware-admin


vsoch:
So currently this gives a warning but doesn't prevent it? Might make sense to change this line too:

```
 message ERROR "pull is only supported for docker and shub URIs\n"
```
This is definitely not going to be taken well, I think we need to still have support for pull for this release, get up the next version of singularity hub, and then only remove it when singularity hub / registry can meet the need of a non sudo pull. I'm treading lightly around this one because it's an explosion waiting to happen...boum!

bauerm97:
@vsoch I think you are misreading the error message and code combination. The error "pull is only supported for docker and shub URIs\n" only is shown when the user enters an invalid URI (i.e. not a shub or docker URI).

vsoch:
Are you sure? It  used to be that, and now it just lists "shub". I think it was changed since I commented. But whatever,  I don't care.

gmkurtzer:
Looks good, thanks!

PR #1007, opened 2017-10-03
Moderate warnings for unprivileged users in build

### LINKED ISSUE #1006: Moderate warning messages
## Version of Singularity:

2.3.9-development.gf44e68d

### Expected behavior

Warnings in `build` explain the situation to the user without causing alarm.

### Actual behavior

Warnings are too scary. 

If you `build` as an unprivileged user (from docker, shub, local image, or tarball) tar is used to copy files to staging area.  If files in the original container are root owned and non-readable by user they will not be copied to new image.  But this is actually not as big of a deal as you may think.  

If you can't read a file then you also can't execute it.  So it is basically useless to you.  In a non-suid environment you will never be able to do anything useful with these files.  So if their absence breaks your container your container is already broken because you can't access them anyway. 

`build` from shub and watch the warnings.

GodloveD:

Our warnings in `build` are too scary given the actual severity of the situation and should be moderated.  This PR:

1. Gives a more accurate warning messages about the problems that a user may encounter when building as an unprivileged user.
2. Sends redundant `tar` messages to `/dev/null` so they don't alarm the user unnecessarily.  

- Fixes #1006 

**Checkoff for all PRs:**

✅  I have added myself as a contributor to the [Author's file](https://github.com/singularityware/singularity/blob/master/AUTHORS.md)

Attn: @singularityware-admin

### DIFF
--- a/libexec/cli/build.exec
+++ b/libexec/cli/build.exec
@@ -160,7 +160,7 @@ if [ "$USERID" != "0" ]; then
 
         # Sandbox with anything else is ok
         else
-            message WARNING "Building sandbox as non-root will result in wrong file permissions\n"
+            message WARNING "Building sandbox as non-root may result in wrong file permissions\n"
         fi
     fi
 fi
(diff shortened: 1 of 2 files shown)


gmkurtzer:
Awesome, thank you!

PR #1011, opened 2017-10-03
Don't delete sandbox when building from a sandbox

GodloveD:

`build` is basically the swiss army knife of containers.  In addition to being able to build from just about anything you can also use `build` to convert your container from one format to another.
```
      squasfs
      /     \
writable -- sandbox
```
You can convert your image from a writable (ext3) to a sandbox (directory) for example.  If you do so, you now have 2 containers.  But if you convert from a sandbox, we automatically delete the sandbox afterward.  This may be unexpected and is probably unsafe.  When I `tar` a directory `tar` does not delete the directory when it is finished.  Why should we do so when creating images.  

This PR fixes the problem (in an admittedly very hacky way).  

Attn: @singularityware-admin


gmkurtzer:
Thanks!!!

Issue #1000, opened 2017-10-02
singularity freezes when I run docker container


antonkulaga:
## Version of Singularity:

Master

### Expected behavior
When I run:
```
singularity run docker://quay.io/ucsc_cgl/fastqc@sha256:86d82e95a8e1bff48d95daf94ad1190d9c38283c8c5ad848b4a498f19ca94bfa
```
it should run the container

### Actual behavior

It freezes with:
```
Docker image path: quay.io/ucsc_cgl/fastqc@sha256:86d82e95a8e1bff48d95daf94ad1190d9c38283c8c5ad848b4a498f19ca94bfa
Cache folder set to /home/antonkulaga/.singularity/docker
[1/14] ||----------------------------------|   0.0% 
```

### Steps to reproduce behavior
```
singularity run docker://quay.io/ucsc_cgl/fastqc@sha256:86d82e95a8e1bff48d95daf94ad1190d9c38283c8c5ad848b4a498f19ca94bfa
```
on Ubuntu-based Linux (in my case I use Linux Mint)

vsoch:
It's likely just downloading huge layers. Could you test with a smaller image? eg ubuntu

GodloveD:
1000!!!

vsoch:
omg. Can we give @antonkulaga a prize?

vsoch:
:100: + 0

vsoch:
Oh, also please still try this command with a smaller image. Even if it wasn't an issue you should still get a prize!!

gmkurtzer:
I just tested it, and while it did take a bit to get it going it worked for me.

BTW, WAHOOOOO!!! 1000+!

gmkurtzer:
Oh, but apparently the package is broken. Not sure if this is the same upstream, but permissions are wrong:
```
$ singularity run docker://quay.io/ucsc_cgl/fastqc@sha256:86d82e95a8e1bff48d95daf94ad1190d9c38283c8c5ad848b4a498f19ca94bfa
Docker image path: quay.io/ucsc_cgl/fastqc@sha256:86d82e95a8e1bff48d95daf94ad1190d9c38283c8c5ad848b4a498f19ca94bfa
Cache folder set to /home/gmk/.singularity/docker
[14/14] |===================================| 100.0% 
Creating container runtime...
[...]

PR #1012, opened 2017-10-04
fixing layer ordering - hopefully for good!

vsoch:

It turns out the version 1 manifest has images listed in the OPPOSITE order of version 2. This can get more confusing given that if a list is returned first, the manifest obtained from the list might have a different version. Our current method of getting / updating manifests uses version 1 for metadata, and version 2 for everything else (we ask for it specifically). So what we do is request version 2, check the version returned (it could be 1) and then reverse the layers given that it's version 1. In either case, the base should be extracted first (the base OS) up to the most recent changes in the images.

We are also removing the unique check on the list - before we didn't add layers that were already included (most of the time it's an empty ENV) but in the rare chance that isn't the case, redundancy might be worth it. 

Attn: @singularityware-admin

### DIFF
--- a/libexec/python/docker/api.py
+++ b/libexec/python/docker/api.py
@@ -73,6 +73,8 @@ class DockerApiConnection(ApiConnection):
         self.auth = None
         self.token = None
         self.token_url = None
+        self.schemaVersion = None
+        self.reverseLayers = False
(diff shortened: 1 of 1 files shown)


gmkurtzer:
Thank you! (ref: #1000)

PR #1028, opened 2017-10-06
fixing issue with download links

vsoch:

Singularity Hub (version 1) built ext3 images. This means that they were huge, and to be efficient with storage, we stored them as gzip compressed (9) and then extracted on pull. So it looked something like this:

build image on cloud builder
```
container.img
```
compress and send to storage

```
container.img.gz
```

download to local filesystem
```
singularity-hub-regional%2Fgithub.com%2Fvsoch%2Fhello-world%2Ff9cd90a7b030a9c794bfe76c123da5a518ad9962%2Fabba5a1f91b2e98ebe125fc7a5151b33.img.gz
```

knowing extension is .img and format ext3, download an image with that extension (named based on repo, or user custom name, or hash, or commit and extract with right command to gzip
```
vsoch-hello-world-master.img.gz -->  vsoch-hello-world-master.img
```

So what happened (yesterday) is that it turned out to be the case that some images didn't need to be extracted. And we could do this based on the extension. So I made a function to check and then only extract if the appropriate type:

```
        # Compressed ext3 images need extraction
        if is_gzip(image_file):
            extract = True
```

My oversight was not seeing that the image name was manipulated from the original download link to always ensure that a .img.gz was returned (as this at the time was the only option). This meant that, regardless of what file was being requested, it would be called .img.gz and then extracted. This of course breaks with squasfs, or any format other than .img.gz of course.

So the fix is to remove this hard coding of the format, and instead use `.simg` but always check for the file type gzip first.

Importantly, extraction only occurs given that we find `gzip` with the file command, and given the builders always building squashfs, I don't see needing to extend this list until something else is allowed. The one potential issue is that (current) .img files will be named with .simg. I don't see an easy way around that.

## Example use cases:

Pulling Registry Image
```
singularity pull shub://sregistry.randomroad.net/examples/docker:latest
Done. Container is at: /home/vanessa/Documents/Dropbox/Code/singularity/singularity/examples-docker-latest.simg

$ ./examples-docker-latest.simg 
This is what happens when you run the container..
```
Note the above was compressed, we look at file to determine, and then decompress.

Sinularity Hub (current version) is the same:

```
singularity run shub://vsoch/hello-world
RaawwWWWWWRRRR!! Avocado!
```

and here is an actual squashfs. Note that you need to specify NOHTTPS because this is a development server:

```
SINGULARITY_NOHTTPS=yes singularity pull shub://doc.fish/singularityhub/hello-world
Done. Container is at: /home/vanessa/Documents/Dropbox/Code/singularity/singularity/singularityhub-hello-world-master.simg
$ ./singularityhub-hello-world-master.simg 
Tacotacotaco
```
It would be good to get other sets of eyes on this - this is an extremely important detail @GodloveD @bauerm97 @gmkurtzer. I'm working around the clock on trying to test out all these details, and I definitely apologize this has to come up so late! I couldn't get to them any faster.

Attn: @singularityware-admin

### DIFF
--- a/libexec/python/shub/api.py
+++ b/libexec/python/shub/api.py
@@ -37,7 +37,7 @@ from shell import (
 from sutils import (
     add_http,
     clean_up,
-    is_gzip,
+    get_image_format,
     read_file,
     run_command
 )
@@ -170,10 +170,9 @@ class SingularityApiConnection(ApiConnection):
                                               show_progress=True)
 
         # Compressed ext3 images need extraction
-        if is_gzip(image_file):
-            extract = True
+        image_type = get_image_format(image_file)
+        if image_type == "GZIP" or extract is True:
 
-        if extract is True:
             if not image_file.endswith('.gz'):
                 os.rename(image_file, "%s.gz" % image_file)
 
--- a/libexec/python/shub/api.py
+++ b/libexec/python/shub/api.py
@@ -37,7 +37,7 @@ from shell import (
 from sutils import (
     add_http,
     clean_up,
-    get_image_format,
+    is_gzip,
     read_file,
     run_command
 )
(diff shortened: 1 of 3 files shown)


gmkurtzer:
Can you check the Travis when you get a chance? Thanks!

vsoch:
huh I  had done one more change and I thought I pushed it but I don't think I see it here

vsoch:
it's just a bug with the test that needs to be fixed for the new code

gmkurtzer:
Looks great, thanks!

PR #1037, opened 2017-10-10
removing self from docs

GodloveD:
Subject says it all.

Attn: @singularityware-admin


vsoch:
This is an easy one - +1 from me when the tests pass.

PR #1047, opened 2017-10-12
Packaging: Add "python" as build requirement.

olifre:
Fixes building of RPMs.

Unclear to me: Since "checkoff for all PRs" includes "this is not against MASTER branch", I did now base it on the development-branch. Still, this is a fix that should go both into dev and stable (i.e. master). 
The "CONTRIBUTING.md" rules state to file against master and development in this case. Is this really what I should do? 
Maybe additionally against 2.x?
Or do you cherry-pick the fix as-needed?

Signed-off-by: Oliver Freyermuth <o.freyermuth@googlemail.com>

Building singularity without having python installed fails.
This can be easily tested by building in a clean environment, e.g. openSUSE's free Open Build Service, on which building will fail without adding python as build requirement. 

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -29,6 +29,7 @@
     - Mark Egan-Fuller <markeganfuller@googlemail.com>
     - Nathan Lin <nathan.lin@yale.edu>
     - Oleksandr Moskalenko <om@rc.ufl.edu>
+    - Oliver Freyermuth <freyermuth@physik.uni-bonn.de>
     - Petr Votava <votava.petr@gene.com>
     - Ralph Castain <rhc@open-mpi.org>
     - Rémy Dernat <remy.dernat@umontpellier.fr>
--- a/singularity.spec.in
+++ b/singularity.spec.in
@@ -34,6 +34,7 @@ URL: http://singularity.lbl.gov/
 Source: %{name}-%{version}.tar.gz
 ExclusiveOS: linux
 BuildRoot: %{?_tmppath}%{!?_tmppath:/var/tmp}/%{name}-%{version}-%{release}-root
+BuildRequires: python
 Requires: squashfs-tools
 
 Requires: %{name}-runtime = %{version}-%{release}


vsoch:
hey @olifre ! Is this different than what we do here?

https://github.com/singularityware/singularity/blob/master/configure.ac#L276

olifre:
Hey @vsoch , 

yes, and very much so: 

The check in configure just checks "is python installed?" and then may stop the build if not. 

The `BuildRequires` in the spec-file tells the build infrastructure building the RPM file to ensure the package "python" is installed on the build host before even running `configure`. 
In automated environments, this information is used to do all sorts of necessary automations. To give an example, a clean infrastructure (such as the open build service, but also build infra of other distributions) does the following nowadays:

- Boot a VM with a minimal image of the OS, almost nothing installed (that is, no `python`). . 
- Read the spec-file (or for Debian, the .dsc) and install all `Requires` / `BuildRequires`. 
- Only then, trigger the actual build procedure (configure, make, make install etc.). 

So this means if the `BuildRequires` are incomplete, the spec will not build (in fact, your configure check will fail claiming python is not installed, which is expected, because the spec-file did not declare it's needed to build singularity).

vsoch:
ah I see! So it's one more level of automation to add to building. Instead of configuring, seeing the missing python (and installing it) you figure that out beforehand and it doesn't fail. This seems like a reasonable addition to me. How have other package managers dealt with this so far, @yarikoptic didn't you maintain a build? Ideally we can have one package source with a working solution, and then for other systems they could use it. Would there be preference for cloning the repo and building from scratch vs. using a package manager?

olifre:
> How have other package managers dealt with this so far...?

I have no idea. Checking the spec-file e.g. Fedora Rawhide is using, which seems to be an independently written file, they added `BuildRequires` for python already, and also packaged a separate slurm plugin. 
They actually even specify:
```BuildRequires: automake libtool chrpath python2-devel```
amongst other things. So if you want to reach a real production level packaging, it might be best to start from that. Sadly, they are still at Singularity 2.2.1, and apply 19(!) downstream patches in total to fix security and quality issues in Singularity. 
You can find more in their SRPM living here:
https://dl.fedoraproject.org/pub/fedora/linux/development/rawhide/Everything/source/tree/Packages/s/







> Would there be preference for cloning the repo and building from scratch vs. using a package manager?

For us, that's a no-go - doing that, you end up at "development level", not at "production level". 
We would like to use Singularity on Debian-based OS and CentOS, and synchronize and time the updates. That's why we choose the open build service, which offers a common infrastructure to provide builds for both. You can see that here: 
https://build.opensuse.org/project/show/home:ubn:singularity

Apart from this need, building from scratch is essentially unmaintainable. If at any time an OS library to which something "self-built" has linked is updated requiring a relinking of this "self-build" package, this will happen automagically if build infrastructure and package managers are used. We can't afford the manpower to do such things manually (and I guess nobody in production can).

gmkurtzer:
Good catch, thanks!

yarikoptic:
re Debian packaging and python: yes, we build- and run-time depend on Python

re "security" patches: ATM we have non downstream patch in Debian.  If I introduce any patch downstream which could be applied upstream -- I file a PR pretty much right away.  At some point we did review some patches from CentOS, ... don't remember what % of them was applied upstream, but at that point I felt that upstream was in good shape.  I would encourage you to submit any downstream patches upstream for review/feedback/etc.

olifre:
> I would encourage you to submit any downstream patches upstream for review/feedback/etc.

I certainly will (this is also why I opened this issue report). I'm also not the Fedora developer who collected all those patches, and I hope most of them are in any case obsolete with newer Singularity releases. I just looked at their `.spec`-file to compare it to the one provided here. 

Thanks for taking care of the Debian packaging! 👍

Issue #1060, opened 2017-10-16
Multi-layer docker image error


hra0031:
### Expected behavior
```
Importing: base Singularity environment
Exploding layer: sha256:364f9b7c969aed6ff05b7ced04b1b53b824386288dffb1e6aec41d5a706dd51e.tar.gz
Exploding layer: sha256:de28b9f03aa4e757c01c9beecc72d1acfcb27cfff53e6ff8d0b4f5073a48df75.tar.gz
Exploding layer: sha256:3f23b445366629b3beaf4d41a971c0d2cebabe6da54c0e5d8f7113691b4bb110.tar.gz
```

### Actual behavior
When first layer of docker image is extracted, some folders don't have write permission. Another layer don't have access and fail to extract.
```
Importing: base Singularity environment
Exploding layer: sha256:364f9b7c969aed6ff05b7ced04b1b53b824386288dffb1e6aec41d5a706dd51e.tar.gz
Exploding layer: sha256:de28b9f03aa4e757c01c9beecc72d1acfcb27cfff53e6ff8d0b4f5073a48df75.tar.gz
tar: usr/lib64/libICE.so.6: Cannot create symlink to `libICE.so.6.3.0': Permission denied
tar: usr/lib64/libICE.so.6.3.0: Cannot open: Permission denied
tar: usr/lib64/libSM.so.6: Cannot create symlink to `libSM.so.6.0.1': Permission denied
tar: usr/lib64/libSM.so.6.0.1: Cannot open: Permission denied
tar: usr/lib64/libX11-xcb.so.1: Cannot create symlink to `libX11-xcb.so.1.0.0': Permission denied
tar: usr/lib64/libX11-xcb.so.1.0.0: Cannot open: Permission denied
tar: usr/lib64/libX11.so.6: Cannot create symlink to `libX11.so.6.3.0': Permission denied
tar: usr/lib64/libX11.so.6.3.0: Cannot open: Permission denied
tar: Exiting with failure status due to previous errors
Exploding layer: sha256:3f23b445366629b3beaf4d41a971c0d2cebabe6da54c0e5d8f7113691b4bb110.tar.gz
```

### Steps to reproduce behavior

Any Multi-layer docker image with overlaying files in layers. For example: ` singularity --debug shell docker://jdeathe/centos-ssh`

GodloveD:
Hi @hra0031 .  Thanks very much for reporting this.  

[...]

PR #1061, opened 2017-10-16
Multi-layer docker image fix

hra0031:
This fixes the problem with files and folders privileges when exploding multilayer docker image.

Individual docker layers have to be extracted to the same folder in right order. To do this, all folders have to be writable so files can be added to it. TAR command is outputting extracted file names unbuffered. File and folder permissions are set immediately after extraction. When TAR extract another file to the same folder, permissions of the folder are restored and files from another layer can't be extracted. This patch fixes it by buffering extracted file names and setting permissions after TAR close.

- Ref: #1060

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -34,3 +34,4 @@
     - Ralph Castain <rhc@open-mpi.org>
     - Rémy Dernat <remy.dernat@umontpellier.fr>
     - Yaroslav Halchenko <debian@onerussian.com>
+    - Josef Hrabal <josef.hrabal@vsb.cz>
--- a/libexec/handlers/image-docker.sh
+++ b/libexec/handlers/image-docker.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# 
+#
 # Copyright (c) 2017, SingularityWare, LLC. All rights reserved.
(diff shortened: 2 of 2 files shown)


GodloveD (approved):
This looks good to me.  I'm going to ask @vsoch to look it over too.

vsoch:
A few comments (and these are possibly outside the scope of this PR so please disregard if you think so)
 - do we have a way to checksum the files before decompress, and exit on failure if there is a bad file (telling the user)? 
- can we add the flag to ignore the warning for old timestamps? It was a bug that just jumped on the list and seems easy enough, granted that its a valid command for older versions of tar.

GodloveD:
@hra0031 do you feel like adding these to your PR?  I think either way is fine.  I will merge if I don't hear from you.

hra0031:
Sorry for late response, I was on vacation last week.
@GodloveD Thank you for PR.
@vsoch I'm not really sure, what you mean.
We can do sha256sum of docker tar file and compare it with filename sha256. Next we can test the gz archive by `zcat -t archive.tar.gz` and try to extract the tar archive to /dev/null `tar xOf archive.tar &> /dev/null` or just list files in tar by ` tar -tvf archive.tar`.
The Timestamp warning can be suppressed by `--warning=no-timestamp` parameter. Do you mean this?

Issue #917, opened 2017-09-11
--nv option broken with 384?


ryanolson:
## Version of Singularity:
2.3.1-dist

### Expected behavior

nvidai-smi output similar to host:

```
ubuntu@runc:~$ nvidia-smi
Mon Sep 11 03:58:22 2017
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 384.66                 Driver Version: 384.66                    |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  Tesla V100-PCIE...  On   | 00000000:00:06.0 Off |                    0 |
| N/A   27C    P0    25W / 250W |      0MiB / 16152MiB |      0%      Default |
+-------------------------------+----------------------+----------------------+

+-----------------------------------------------------------------------------+
| Processes:                                                       GPU Memory |
|  GPU       PID  Type  Process name                               Usage      |
|=============================================================================|
|  No running processes found                                                 |
+-----------------------------------------------------------------------------+
```

### Actual behavior

nvidia-smi not found.

### Steps to reproduce behavior

Install singularity with apt package as described in docs.  Run the follwing:

```
ubuntu@runc:~$ singularity --version
2.3.1-dist
ubuntu@runc:~$ singularity exec --nv docker://nvidia/cuda:8.0-runtime nvidia-smi
Docker image path: index.docker.io/nvidia/cuda:8.0-runtime
Cache folder set to /home/ubuntu/.singularity/docker
Creating container runtime...
/.singularity.d/actions/exec: 8: exec: nvidia-smi: not found
```

Similar to #785

ryanolson:
@GodloveD - Any luck with the `nvidia-container-cli list` option?

```
ubuntu@runc:~$ nvidia-container-cli list -cu
/usr/lib/nvidia-384/bin/nvidia-smi
/usr/lib/nvidia-384/bin/nvidia-debugdump
/usr/lib/nvidia-384/bin/nvidia-persistenced
/usr/lib/nvidia-384/bin/nvidia-cuda-mps-control
/usr/lib/nvidia-384/bin/nvidia-cuda-mps-server
/usr/lib/nvidia-384/libnvidia-ml.so.384.66
[...]

PR #1082, opened 2017-10-20
Rework --nv option

GodloveD:

This adds 3 enhancements to the `--nv` option.  

1. Singularity will now check to see if [`nvidia-container-cli`](https://github.com/NVIDIA/libnvidia-container) is installed and accessible on the host system.  If so, Singularity will use it to determine the nvidia libs and binaries that it should bind at runtime.  It will fall back to using `nvliblist.conf` if `nvidia-container-cli` is not present, is returning errors or bad data, or if the user sets the following environment variable to any value.
```
    export SINGULARITY_USE_NVLIBLIST_CONF=1
```

2. `nvliblist.conf` has a new section allowing an administrator to list the locations of nvidia binaries that should be bind mounted within containers at runtime.  This means that previously missing binaries will be bind mounted into the container, and `nvidia-smi` can be bind mounted even if it is not in a standard location.  By default the following will now be bound using `nvliblist.conf`
    - /usr/bin/nvidia-smi
    - /usr/bin/nvidia-debugdump
    - /usr/bin/nvidia-persistenced
    - /usr/bin/nvidia-cuda-mps-control
    - /usr/bin/nvidia-cuda-mps-server

3. There is a configuration option added to `singularity.conf` so that the `--nv` option will be passed to all action commands by default.  This may be useful in a situation where GPUs are assumed such as an all GPU cluster.  Users can override this behavior with the `$SINGULARITY_NV_OFF` variable if need be.  

- Ref: #917 

**Checkoff for all PRs:**

[ ]  I have tested this PR locally with a `make test`
✅  I have added myself as a contributor to the [contributors's file](https://github.com/singularityware/singularity/blob/master/CONTRIBUTORS.md)

Attn: @singularityware-admin


vsoch:
cool I have a few questions!

1. If the nvidia libraries are installed in some location that a non root person can somehow manipulate, wouldn't it be trivial to put a different file there (like a bird raised in a nest by another mother).
2. I think we want to possibly talk about our use of the file "functions. I think we should maybe take the minimalist approach of only sourcing what is absolutely needed. With all the functions in one file, all of them are active regardless if needed are not. What about a functions/ folder, with then different groups under each, and calling the group? So for nvidia:

```
functions/
      functions-base  #  (or main)
      functions-nvidia
```
That way, we can be more specific about what is active.

libexec/functions
```diff
@@ -441,6 +441,66 @@ is_tar() {
...
+        message ERROR "Failed to create temporary nvliblist file.\n"
+        ABORT 255
+    fi
+
+    # try to use nvidia-container-cli if installed and functional
+    if singularity_which nvidia-container-cli > /dev/null 2>&1 && SINGULARITY_NV_FILES=`nvidia-container-cli list -cguv`; then
+        # bins should have full paths
+        SINGULARITY_NVBINLIST=`echo "$SINGULARITY_NV_FILES" | grep -Ev "\.so"`
+        # libs should just be basenames ending in ".so"
+        for i in `echo "$SINGULARITY_NV_FILES" | grep -E "\.so"`; do
```
vsoch: is this absolutely, always entirely the case to end in so?

libexec/functions
```diff
@@ -441,6 +441,66 @@ is_tar() {
...
+        done
+
+    else # fall back to nvliblist.conf
+        message 2 "nvidia-container-cli not installed or returning error\n"
+        message 2 "falling back to $SINGULARITY_sysconfdir/singularity/nvliblist.conf\n"
+        # strip comments and spaces and bins do NOT end in .so
+        SINGULARITY_NVBINLIST=`cat $SINGULARITY_sysconfdir"/singularity/nvliblist.conf" | grep -Ev "^#|^\s*$" | grep -vE "\.so$"`
+        # strip comments and spaces and libs DO end in .so
+        cat "$SINGULARITY_sysconfdir/singularity/nvliblist.conf" | grep -Ev "^#|^\s*$" | grep -E "\.so$" > $SINGULARITY_NVLIBLIST
+    fi
```
vsoch: OCD dinosaur kicking in... can we do:

```
cat "${SINGULARITY_sysconfdir}/ "...etc
```
instead?

libexec/functions
GodloveD: It's even worse than that.  I'm just grepping to see if the lines **contain** `.so`. 

I'm doing this because I need to process libraries and binaries separately and I'm relying on the presence or absence of `.so` to determine if the file in question is a lib or bin.  

I think that the libraries are always guaranteed to have `.so` in the name.  I think the only danger is that one of the binaries ends up having `.so` in the name.  If that happens it will be included in this list, which is no big deal because this list is only used for pattern matching within the `ld.so` cache and it just won't match any pattern.  But the worse news is that it will be excluded from the binary list and won't be mounted into the container.  But that still probably isn't that big of a deal for most applications.  These bins are not being mounted into the container right now.  If a user really needs them they have to install them on their own.

The real problem is that `nvidia-container-cli` just outputs a bunch of paths with nothing to differentiate a library from a binary.  Even if they did have some kind of filtering option I wouldn't want to call `nvidia-container-cli` more than once because it is slow.

So I guess my answer here is I think there is a very small chance this could fail.  And if it fails there is an additional very small chance that failure could cause the container to fail.  But I don't know any better way to do it.

libexec/functions
GodloveD: Yes we can!

GodloveD:
Hi @vsoch.  Thank you very much for reviewing this.  

Let me reply to your comments:

> If the nvidia libraries are installed in some location that a non root person can somehow manipulate, wouldn't it be trivial to put a different file there (like a bird raised in a nest by another mother).

Yeah that could be bad.  Only the administrator should be able to manipulate the `nvidia.conf` file.  So I guess I should put a note in the comments letting the admin know that they should not include any non-root owned binaries in that file or they could be asking for trouble.  

> I think we want to possibly talk about our use of the file "functions. I think we should maybe take the minimalist approach of only sourcing what is absolutely needed. With all the functions in one file, all of them are active regardless if needed are not. What about a functions/ folder, with then different groups under each, and calling the group?

I can understand the reasoning here, but I'm not really crazy about the idea.  Here's why:
- **What problem does it fix?**
I think the main fix is reducing overhead.  If you only have to source what you need the code can be faster because every file doesn't have to source every function.  That sounds good in theory, but I think that the sourcing introduces a trivial amount of overhead in practice.  It also might fix the problem of needing to read through a long file to find the fx you want.  But I think that is easier then traversing a directory structure to find the fx.
- **What problem does it introduce?**
I think we are more likely to write 2 functions with the same name if they are spread out across a directory structure.  Then if we source 2 different files with the two functions in question we have a difficult to debug problem.  I also don't want to have to figure out which files I should source at the top of every file to get the functions I want.  But that's just me.

libexec/functions
vsoch: I think that's reasonable, given what you have to work with! As long as nothing terrible, security wise could happen (other than an issue posted that we resolve) then this seems ok. Who made the `nvidia-container-cli` and why is it so doopy?

vsoch: Maybe the container client could be changed, or a command added to indicate which you want?

libexec/functions
vsoch: https://www.youtube.com/watch?v=RF5ubSoXZGM

vsoch:
Thanks @GodloveD ! For the second I was thinking of a security standpoint - like with API authentication tokens you only ask for the absolute minimum you need. The speed up would be trivial, but down the line as Singularity grows in size, would we really want one huge@ss file of functions, sourced always? What if 1/3 of the functions aren't needed, but there is some escalation pathway to reference once. I'm not good enough with security to know how, but it's been  on my radar for a while that the general image handlers / functions are turning into big hairballs, and maybe we want to nip that in the butt. I don't mean to say it's for this PR specifically, but in general something to talk about.

libexec/functions
GodloveD: > Maybe the container client could be changed, or a command added to indicate which you want?

You mean instead of defaulting to `nvidia-container-cli` the user could specify that they want to use `nvliblist.conf` instead?  I think that is a good idea.  How do you suggest I implement that?  Option + env var?

GodloveD: Oh and I don't want to badmouth `nvidia-container-cli`.  It's not doopy and it's filling a big gap for us.  It is a little slow but I think that is because it is loading and unloading kernel modules?  Anyway, it is a big improvement to the guessing game that I've been playing.

GodloveD:
Yeah.  I guess.  Then we will have to start every bash script with a bunch of `include` like statements similar to some other languages I know...

etc/nvliblist.conf
```diff
@@ -1,9 +1,12 @@
 # NVLIBLIST.CONF
 # This configuration file determines which NVIDIA libraries to search for on 
 # the host system when the --nv option is invoked.  You can edit it if you have
-# different libraries on your host system.
+# different libraries on your host system.  You can also add binaries and they
+# will be mounted into the container when the --nv option is passed.
 
 # put binaries here
+# In shared environments you should ensure that permissions on these files 
+# exclude writing by non-privileged users.
```
vsoch: +1!

libexec/functions
vsoch: yes, **or** continue to use `nvidia-container-cli` but ask its developers to make it more customizable? Is that possible?

etc/nvliblist.conf
vsoch: >> Yeah. I guess. Then we will have to start every bash script with a bunch of include like statements similar to some other languages I know...

But not python, because python uses "from" and "import" and because python is just flawless! :P

libexec/functions
ryanolson: Is the ask for nvidia-container-cli to have new flags like `--libraries` and `--executables` to help separate out the list of host files that should be present in the container to provide the requested capabilities?

@flx42 @3XX0

GodloveD: That would be one way to do it.  But then we would have to call `nvidia-container-cli` 2x which would be slow.  Another way would be to return some indication as to the file type along with the file location in the text output.  Maybe in csv or tab delimited format or something?

vsoch: Yes this is what I had in mind! I think the two calls to be absolutely sure about the files would be worth it - the grepping will likely work for most, but I would definitely predict some error from it in the long run. Is it really significantly slow to call it once vs. twice?

GodloveD: Each call to `nvidia-container-cli` takes like a second or so.  There is a noticeable pause.  So two calls will create an even more noticeable pause.  

I think it would be nice if the nvidia devs can add something like this for us to use in the future, but I don't feel it's necessary to add this to merge the PR.

3XX0: @GodloveD it shouldn't take this long, your driver is probably not setup correctly  (e.g. `nvidia-persistenced` isn't running) which forces the library to go through the full GPU initialization.

GodloveD: @3XX0 Ah thanks.  That is probably the issue.  Any chance we could get some filters set up to differentiate binaries from libraries?  Should I raise an issue?

GodloveD: Also @3XX0, while I have you, are there 1 or a handful of core libs that I can _always_ expect to find (regardless of driver version)?  I want to add some error checking and I thought it might be a good idea to check for the presence of one or more libs that is more or less guaranteed to be present.

3XX0: I guess for the time being we can have a filter option and we'll need to filter IPCs as well. But more generally the `list` command won't be able to keep up in terms of features so we will need to fork and implement the library directly in Singularity.

Regarding your second comment, unfortunately there is no way of knowing.
This why the library has relaxed checks right now and several options to tweak the driver features. I was planning on implementing a strict mode which checks things more thoroughly though. Also, this list will likely change in future driver versions.

GodloveD:
Just added an environment variable so that a user can specify not to use `nvidia-container-cli` even if it exists on the path.  Also added more error checking so that if `nvidia-container-cli` returns anything that is not a file on the host system the output will be deemed bogus and we will fall back to `nvliblist.conf`.

libexec/functions
GodloveD: @vsoch 
>Maybe the container client could be changed, or a command added to indicate which you want?

I added an env var (`SINGULARITY_USE_NVLIBLIST_CONF`) so that users can ignore `nvidia-container-cli` and use nvliblist.conf if that better suits their needs.

jtriley:
Might also want a kill switch for this if it's enabled:

```There is a configuration option added to singularity.conf so that the --nv option will be passed to all action commands by default. This may be useful in a situation where GPUs are assumed such as an all GPU cluster.```

**Comment by GodloveD** [comment]
@gmkurtzer suggested offline that I use c to return values from the config file.  Now I've added that, so it should be good for review and merge.

File: src/get-configvals.c [hunk 4]
```diff
@@ -0,0 +1,44 @@
...
+    singularity_config_init();
+
+    /* 
+    If the key does not exist in the singularity.conf file, this is just 
+    going to return null char.  The function was originally designed to return
+    the default value if the value does not exist in the conf file, but I don't 
+    know how to do that using only strings, and the key needs to be based on 
+    user input, not hardcoded. 
+    */
+    printf("%s\n", _singularity_config_get_value_impl(key, "\0"));
```
gmkurtzer: Just use `singularity_config_get_value()` here.

GodloveD: Unless I'm mistaken, I can't use `singularity_config_get_value()`.  It only works with hard-coded variables right?  I cant just pass some arbitrary string out of bash to it.

vsoch: Isn't there some avenue to get any variable that starts with `SINGULARITY_` ? Not sure how it works in the C but I know they magically show up in the bash, haha.

vsoch: I think this one pulls from env --> https://github.com/singularityware/singularity/blob/adf5259ac94d7b213c649f073dcda32459b0c8b7/src/util/registry.c#L103 and the previous is from the config.

vsoch: oh and I think the comment above is saying to use the higher level function to that one (without the _)

GodloveD: It's not a higher level function.  It's a macro that is designed to only work with variables that are defined at build time.  Unless I'm mistaken.  That means I can't just pass an arbitrary character.  That is why I'm using `_singularity_config_get_value_impl()`.  If there is a way to use  `singularity_registry_get()` please show me.

vsoch: I don't know C well enough to be able to give a confident answer. I can say that a long time ago I started (almost the same named script) but did try using the registry function: https://github.com/vsoch/singularity/blob/961fcf50a264c8574071adb311bd78daf2674e13/src/get-config.c I needed some mentoring with regard to what to try next so it never went anywhere from there. Anyway, I'm sure @gmkurtzer will be able to give some insight here!

GodloveD: That's hilarious!  It's almost identical.  If I had to wager a guess I would say that get-config.c doesn't compile and it complains that key is not defined at compile time (or some such error in c-speak).  Am I right?

vsoch: Haha, no idea, I can't remember... this was back in July!

ryanolson:
What's the timeline on the merge into master?  Similarly, is there an ETA on the next official release?

GodloveD:
Hi @ryanolson we are waiting on green light from @gmkurtzer to merge this into development.  He told me offline that he is going to try to think of a better way to implement the config parsing than calling `_singularity_config_get_value_impl` in `get-configvals.c`, but I haven't heard from him yet.

It won't get merged into master until the next release (3.0).  We are working toward that as fast as we can right now, but there are still several things on the roadmap that we have not implemented yet.  Sorry I can't give you a more definite answer right now.

src/get-configvals.c
```diff
@@ -0,0 +1,44 @@
...
+* Copyright (c) 2017, SingularityWare, LLC., Inc. All rights reserved.
+* 
+* This software is licensed under a 3-clause BSD license.  Please
+* consult LICENSE file distributed with the sources of this project regarding
+* your rights to use or distribute this software.
+* 
+*/
+
+
+#define _GNU_SOURCE
```
yhcote: This is unneeded

src/get-configvals.c
```diff
@@ -0,0 +1,44 @@
...
+#endif
+
+int main(int argc, char **argv) {
+
+    if ( argc < 2 ) {
+        printf("USAGE: %s [key]\n", argv[0]);
+        exit(0);
+    }
+
+    char *key = strdup(argv[1]);
```
yhcote: Declare your vars at the top of the scope. In this case : char *key; at line 23.
Then here (29) just : key = strdup(argv[1]);

Now if _singularity_config_get_value_impl() doesn't change the content of the string at "key", you could just pass argv[1] to _singularity_config_get_value_impl() and avoid the declaration of key, the strdup() and the free() of this intermediate var.

yhcote (changes requested):
See my inlined comments

3XX0:
@GodloveD FYI I reworked the listing in libnvidia-container master and added filtering options

GodloveD:
Thanks for the heads up @3XX0.  I'll have a look.

yhcote (approved):
Looks good to me

3XX0:
@GodloveD I saw this was merged but the code hasn't been updated to support the latest `nvidia-container-cli` (which adds filtering). Hence, the `--nv` will fail to work properly.

vsoch:
oops, thanks @3XX0 this would definitely cause some bugs!

GodloveD:
@3XX0 can you be more specific please?  How should I be using the new filtering options. @vsoch What bugs do you expect this to cause.  Thanks.

vsoch:
@GodloveD after talking with my group users are still reporting general issues with the `--nv` flag, so I'm hypothesizing it could be related. I don't have a specific example off the top of my head :thinking:

GodloveD:
@vsoch I thought you told me your users were not using development.  So their issues could not be related to `nvidia-container-cli` right?

@3XX0 are you saying you changed the CLI so my calls to `nvidia-container-cli` won't work properly anymore?

3XX0:
@GodloveD Yes, after you asked for filtering options, we changed the capability options that are used in this commit (i.e. `-cguv`) and added those instead.

See `--help` below:
```
Usage: nvidia-container-cli list [OPTION...]
Query the driver and list the components required in order to configure a
container with GPU support.

 Options:
  -?, --help                 Give this help list
  -b, --binaries             List driver binaries
      --compat32             Enable 32bits compatibility
  -d, --device=ID            Device UUID(s) or index(es) to list
  -i, --ipcs                 List driver ipcs
  -l, --libraries            List driver libraries
      --usage                Give a short usage message
  -V, --version              Print program version
```

GodloveD:
Ah.  OK thanks.  I was assuming that you just added some filtering options on top of what was already available and didn't change the cli.  I'll have a look at the new version.  

This shouldn't be a huge issue b/c `nvidia-container-cli` should just be throwing an error with the current way the code is written and that should cause the code to fall back to the configuration file instead (which is what we are using now).  But I'll definitely have a look and update asap.  

Thanks for the heads up.

ps-account:
Not sure if this is the right place: I'd like to be able to configure --nv to chose between different cuda versions available on the host. 

In my /usr/local/ I have the following folders:
cuda-8.0/ cuda-9.0/ cuda-9.1/ cuda-9.2/

I normally activate them with module load cuda/x.x

However, I start singularity with --nv it automatically seems to take the cuda-9.0 and translate it into /usr/local/cuda in the container, no matter which module is loaded. Even if on the host I sudo -ln s /usr/local/cuda to a different version! 

Is this expected behaviour, and is there a way to change this somehow?

GodloveD:
Heya @rdwrt!  

The `--nv` option is only concerned with the nvidia driver.  cuda is supposed to be installed into the container.  And the `--nv` option binds all of the libraries it finds on the host system into the metadata directory at `/.singularity.d/libs`.  It doesn't touch `/usr/local`.  

So if I had to guess, I would say that your container already has cuda 9.0 installed.  Could you verify by shelling into the container without the `--nv` option and checking?

ps-account:
ah yes, ok, my bad my bad!!!

qhaas:
Am I using 'SINGULARITY_USE_NVLIBLIST_CONF' correctly to prevent `nvidia-container-cli list` from being leveraged?  It doesn't seem to make a difference...  Note, I added 'libnvidia-glvkspirv.so' to the list of libraries in 'nvliblist.conf' to [enable vulkan support](https://github.com/NVIDIA/libnvidia-container/issues/55#issuecomment-573871936)
```
$ cat /etc/redhat-release 
Red Hat Enterprise Linux Server release 7.6 (Maipo)
$ export SINGULARITY_USE_NVLIBLIST_CONF=1
$ tail /etc/singularity/nvliblist.conf | tail -1
libnvidia-glvkspirv.so
$ singularity --version
singularity version 3.5.2-1.1.el7
$ singularity build --sandbox centos7/ docker://ppc64le/centos:7
...
$ singularity exec --nv -C -e centos7/ ls /.singularity.d/libs/libnvidia-glvkspirv.so.418.87.00
/.singularity.d/libs/libnvidia-glvkspirv.so.418.87.00
$ sudo yum install -y libnvidia-container-tools
...
$ singularity exec --nv -C -e centos7/ ls /.singularity.d/libs/libnvidia-glvkspirv.so.418.87.00
ls: cannot access /.singularity.d/libs/libnvidia-glvkspirv.so.418.87.00: No such file or directory
```

On a related note, might not hurt to add 'libnvidia-glvkspirv.so' to 'nvliblist.conf' to enable Vulkan.  Nvidia publishes a list of most of the libraries, which can change between major driver version releases, [in this section of their readme](https://download.nvidia.com/XFree86/Linux-x86_64/440.44/README/installedcomponents.html).

jmstover:
@qhaas  - `SINGULARITY_USE_NVLIBLIST_CONF` is a 2.x option. In 3.x either `singularity-container-cli` is there and it's used or it isn't. :/  You can't have it installed, and use and env variable to not use it.

@dctrud in `pkg/util/gpu/paths.go` it'd be pretty easy to do something to control the execution flow based on that variable. Is that something that is wanted?

dtrudg:
Hello @qhaas - this PR is quite old and applies to prior versions, and as @jmstover mentions that env var is not available in `3.5` Please could you open a new issue if you'd like to have `SINGULARITY_USE_NVLIBLIST_CONF` available in future versions, given detail of the use cases? 

We hard prefer nvidia-container-cli because it's our understanding that it *should* be giving a complete list of libraries.

PR #1083, opened 2017-10-22
Make tar file creation reproducible

bmwiedemann:
in order to make package builds reproducible.
See https://reproducible-builds.org/ for why this is good.

Requires GNU tar-1.28 or later to be reproducible
Using gzip -n to not add a timestamp to the .gz header
OTOH I'm not sure if it was really intended to use gzip, because the file is just named .tar and very small anyway - then we could just call tar cf


GodloveD (commented):
This looks good to me.  I'm going to ask for a second opinion from @yhcote.

GodloveD:
@yhcote Agrees that this looks fine so I am merging.

GodloveD:
I'm sorry @bmwiedemann.  I stupidly forgot to check if this PR was against development or master and I merged into master.  Then I had to revert.  Can you please send this PR to development instead?  Thank you.

Issue #1044, opened 2017-10-12
OverlayFS default option value broken


olifre:
Since: 
https://github.com/singularityware/singularity/blob/e480d9640f5e76a51af7a9f43e87b77f008f6652/configure.ac#L257
the configure-macro uses some obscure "autodetection" on the build machine to deduce whether overlayfs is available on the execute machine. Since in any reasonable production setup (not talking about development here!), the build and execute machines do not coincide, this is a broken idea. 

Apart from the fact that this is something all distributions downstream have to patch out again (build environments are by rule kept clean from stuff not needed for build!), this detection is also broken. For example, I have OverlayFS built into my kernel, which makes this arcane algorithm fail. 

This is of course something ok for the testing environments, but not for the actual build. Actual build should do as common practice:
Add a config option like `--enable-overlayfs` / `--disable-overlayfs`. 

To let you know, we are building Singularity on Open Build Service in clean VM environments, and running then on several distributions in production. 

Edit: Since for my system that means overlayFS is now off by default, I added the `[Regression]` flag.

DrDaveD:
@olifre What version are you looking at?  The just-released singularity-2.4 tries to avoid that problem by adding a 'try' config option to let it try overlay at run time if it can't determine the issue at build time.  See pr #912.  If you select "Blame" at the link you gave, you can see that the line you link to has been there a year.

olifre:
@DrDaveD You are fully correct, it's not a regression, but actually appears to be broken since long. Sorry for misidentifying that. 

The "try" technique does not help with that at all. First of, the idea of detecting something at build time on a build machine and assuming that will be valid on the execute machine is broken by design. Build machines are minimal systems to ensure no dependencies are accidentally automagically included and hence forgotten. Heck, they may even run a different kernel. 
Execute nodes will be substantially different. 

Secondly, it will fail if overlayfs is built into the kernel. On my 4.13 kernel with overlayfs built-in, I get:
```
checking for overlayfs... no                                                                                                                                                                                                                 
```
The reason is:
I do have `modules.dep`, but there is no string matching `overlayfs.ko` in it. 

So yes, you are right, it's not a regression (and not introduced by the PR you linked), it's broken since very long time.

DrDaveD:
[...]

PR #1084, opened 2017-10-22
always use 'try' for overlayfs functionality.

olifre:
Finding the module on build host is neither a guarantee it will
exist in the runtime environment, nor is absence of the module
a guarantee overlayfs is not supported (may be compiled into
kernel).
So don't deduce runtime environment from build host,
but always use the new 'try' feature.
For testing, activate that if the module is found.

Signed-off-by: Oliver Freyermuth <o.freyermuth@googlemail.com>

- Ref: #1044

**Checkoff for all PRs:**

Attn: @singularityware-admin

Note: I did not run `make test`, that makes no sense on my machine, since I have overlayfs support compiled in.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -255,20 +255,6 @@ fi
 AC_SUBST(SETNS_SYSCALL)
 
 
-AC_MSG_CHECKING([for overlayfs])
-KVERS=`uname -r`
-if test -f "/lib/modules/$KVERS/modules.dep"; then
-    if grep -q 'overlay.ko' "/lib/modules/$KVERS/modules.dep"; then
-        AC_MSG_RESULT([yes])
-        OVERLAY_FS=1 # this adds it to testing
-    else
-        AC_MSG_RESULT([maybe])
-    fi
-else
-    AC_MSG_RESULT([maybe])
-fi
-
-
 # ---------------------------------------------------------------------
 # PYTHON
 # ---------------------------------------------------------------------
@@ -294,7 +280,6 @@ else
 
 fi
 
-AC_SUBST(OVERLAY_FS)
 AC_SUBST(SINGULARITY_DEFINES)
 
 AC_MSG_CHECKING([--with-slurm])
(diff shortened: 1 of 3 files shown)


GodloveD:
Heya @olifre.  Thanks for the PR.  Can you move it to go agains the development branch please?

olifre:
Hey @GodloveD !
Yes, of course - sorry, this was not my intent. 
Should be fine now :).

GodloveD:
Thank you!

GodloveD:
I'm giving this the "bug" tag to alert @gmkurtzer that this is a bug fix and is something we will probably want to merge into 2.4.1 before releasing.

vsoch:
:bug: :spider:

DrDaveD:
I understand the reason to "try" if the overlay.ko module is not found, but is there a good reason the default shouldn't be "yes" if the module is found?

olifre:
> but is there a good reason the default shouldn't be "yes" if the module is found?

Since the machine on which Singularity is compiled and the machine on which Singularity is later used may not be the same, there's no reason to assume that the module is present in the runtime environment. For example, there are sites running Singularity within Docker containers on the worker nodes. In there, `overlay.ko` will not be present, so unless it's loaded already on the host (or you can trigger loading it), there's no chance to use overlayfs. 
Using `try`, things will still work, but using `yes`, they will fail by default.

gmkurtzer:
I've been thinking about this some more. With this PR, I'm not sure it makes sense to keep any check in the `configure.ac` any longer. Do you want to do that @olifre in this PR branch?

Thanks!

olifre:
Hi @gmkurtzer , 
I also considered removing the check altogether, but there's still one thing dependent on it: The test-suite.
Even with my changes, the variable `OVERLAY_FS` is only set to `1` if the module is found, thus influencing what the test suite does. 

I think that there, the approach is also valid, since the test suite will run on the build machine. 
Of course, that still means the test suite will only test overlayfs if the module is there (and not, for example, if overlayfs is part of the kernel). But I have no good idea to fix this.

gmkurtzer:
How about we do the module check from within the test harness then instead?

olifre:
@gmkurtzer Sounds good! Then I'll move the test to `test.sh.in` and remove it from configure. 
I'll update the PR accordingly.

olifre:
Done!
Let me know if you prefer these two to be squashed together.

gmkurtzer:
This looks perfect, thank you @olifre!

PR #1085, opened 2017-10-23
Add/python file logging

vsoch:

# Python File Logging
This PR is intended to support @bauerm97 work in progress to allow for logging to a file with the command:

```
singularity run --log /tmp/pancakes.txt container.img
```

specifically, adding the same logic to the python logger to ensure that content is appended to the file. The logic works as follows:

1. Given that the environmentvariable `SINGULARITY_LOGFILE` is found, the logger captures it. This (I think) means that the code that @bauerm97 is working on will export it for python to find.
2. Given `SINGULARITY_LOGFILE`, the level is automatically set to `DEBUG`
3. If the logfile doesn't exist, it is created
4. Subsequent logging commands are appended to the file.

Here is an example cat of a logfile with three messages printed:
```
cat /tmp/tmp01byskrw 
DEBUG This is a message log, not a message dog.
DEBUG Another thing.
WARNING A warning!
```

## Concern - Atomicity
My primary concern is that of atomic writing to file. We want to be sure that two processes aren't trying to grab the same file at the same time. This shouldn't be a concern for the functions we have that use multiprocessing (they stream the progress bars to stdout) but it could be a concern if a single logfile is used across multiple containers. The traditional approach for doing this would be to write to a temporary version of the file, and then copy to the actual file name only when finished (this is how we download Docker layers), but this still would have race conditions if two containers using the same file copy different versions (and then some content is missing depending on time of copy). 

So we should chat about how to avoid this case. Either it might be that the user provides a prefix and then new processes (different containers) grab a different prefix, or we explicitly say the log file i intended for logging one container process (so there can't be race conditions).

## Logger History
Also notice that (and I added this a long time ago) the logger has a complete history of himself. First let's log some things:

```
bot.debug("This is a message log, not a message dog.")
bot.debug('Another thing.')
bot.warning('A warning!')
```
Regardless of where that is printing (screen or file) I can, at any point, ask to see the entire history (it's in a list)

```
print(bot.history)
  ['DEBUG This is a message log, not a message dog.\n',
   'DEBUG Another thing.\n',
    'WARNING A warning!\n']
```
which of course formats when I join the list

```
print(''.join(bot.history))
DEBUG This is a message log, not a message dog.
DEBUG Another thing.
WARNING A warning!
```

This is useful potentially for two reasons. 1 - we can drastically reduce the number of times needed to write (at least for python) if we do a dump of everything at once, after it's done. Given that any error arises that would have excited, we make sure to print right before exit.  This might help with the race condition issue, if one exists, although it wouldn't fix it. 2 - it means that we could potentially expose some "on demand" generation of a log. This probably doesn't make a lot of sense now, but imagine that you have a container instance and you want to run the equivalent of "docker-compose logs" to see logs.

## Tests
I also made a new set of tests for the message logger, and made the python tests in ABC order. It was bothering me, lol.

# Fixing bug in `test`
In this PR I'm also fixing the bug found in the "test" script this morning. After discussion, we can go about integration in any way - this branch merged before @bauerm97 will be working fine even without the bash export.

# Other Things to Think About
This idea of logging is really interesting because it opens up the door to different use cases. For example, if a container is running as a service, it might actually want to send log messages to a web endpoint, and so the user would specify the endpoint and they would be POSTed. We might want to send more of status messages about state - like "the container run is started / done / errored" or "the container is started / stopped" and that would fit very well with our current messages module too.

Attn: @singularityware-admin

### DIFF
--- a/libexec/bootstrap-scripts/environment/test
+++ b/libexec/bootstrap-scripts/environment/test
@@ -7,7 +7,7 @@ for script in /.singularity.d/env/*.sh; do
 done
 
 
-if test -z "${SINGULARITY_APPNAME:-}"; then
+if test -n "${SINGULARITY_APPNAME:-}"; then
 
     if test -x "/scif/apps/${SINGULARITY_APPNAME:-}/scif/test"; then
         exec "/scif/apps/${SINGULARITY_APPNAME:-}/scif/test" "$@"
--- a/libexec/python/message.py
+++ b/libexec/python/message.py
@@ -34,6 +34,8 @@ The following levels do nothing (quiet)
 
 0
 
+SINGULARITY_LOGFILE, when defined, outputs to the file with level DEBUG
+
 Copyright (c) 2016-2017, Vanessa Sochat. All rights reserved.
 
 "Singularity" Copyright (c) 2016, The Regents of the University of California,
@@ -55,6 +57,7 @@ perform publicly and display publicly, and to permit other to do so.
 
 import os
 import sys
+import io
 
 ABRT = -4
 ERROR = -3
@@ -71,8 +74,9 @@ DEBUG = 5
 class SingularityMessage:
 
     def __init__(self, MESSAGELEVEL=None):
-        self.level = get_logging_level()
         self.history = []
+        self.logfile = os.environ.get("SINGULARITY_LOGFILE")
+        self.level = get_logging_level()
         self.errorStream = sys.stderr
         self.outputStream = sys.stdout
         self.colorize = self.useColor()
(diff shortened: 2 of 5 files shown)


vsoch:
The last fail was on a docker pull, one of the layers had an unexpected end of archive which means either a bad download, or one of those empty files. Should we have a check of the sum to ensure the content? and/Or check the size and skip empty files? I think this would likely happen on the bash/shell side of things.

vsoch:
okey doke, @bauerm97 here you go!

vsoch:
ping @gmkurtzer @bauerm97

bauerm97:
@vsoch This looks good to me, ready to merge when @gmkurtzer gives the all clear

GodloveD:
Hi all.  What is the status on this?

vsoch:
@bauerm97 asked for the addition, and I added it. I'm not sure about the plan beyond that, but please let me know if I can be helpful.

GodloveD:
I think @bauerm97 is indicating that he has what he needs from this PR and it is in Development already.  I'm going to close it for now, but please feel free to reopen if necessary.

PR #1092, opened 2017-10-25
Add limit container groups directive

cclerget:

Add possibility to limit container by groups and resolve limit container paths in case of symlinks

**Checkoff for all PRs:**

Attn: @singularityware-admin


GodloveD:
Adding bug tag so that we know to cherry pick this into 2.4.1 before release.

gmkurtzer:
This is a new feature, not a bug. @cclerget There is a failure in the CI testing, can you take a look please?

Thanks!

cclerget:
@gmkurtzer Build failed due to the following network issue https://www.traviscistatus.com/incidents/v4jnp6nkgzqx. If you rebuild, this should pass

src/lib/image/image.c
```diff
@@ -222,6 +223,50 @@ void singularity_limit_container_owners(struct image_object *image) {
     }
 }
 
+void singularity_limit_container_groups(struct image_object *image) {
+    const char *limit_container_groups = singularity_config_get_value(LIMIT_CONTAINER_GROUPS);
+
+    if ( strcmp(limit_container_groups, "NULL") != 0 ) {
+        struct stat image_stat;
+        char *group_token = NULL;
+        char *current = strtok_r(strdup(limit_container_groups), ",", &group_token);
```
GodloveD, line 232: This may be a very naive question from a dummy who doesn't know c very well, but why the `strdup` here?  Doesn't it work just the same without the `strdup`?

GodloveD (changes requested):
Overall I think this is great.  I just have one question about the use of `strdup`.

src/lib/image/image.c
GodloveD: ping @cclerget.  Can you explain this to me when you get a chance?

cclerget: Sorry for the delay @GodloveD, strtok and strtok_r modify their first argument, that's why the string is duplicated to not modify limit_container_groups directly

GodloveD: Thank you for the clarification!

GodloveD (approved):

GodloveD:
I'm happy with this, but it should get a look from another dev who knows C better than I do.

ctmadison (approved):

PR #1093, opened 2017-10-25
Packaging: Explicitly include local state directories.

olifre:

This is critical to let namespaces with user privileges work out of the box.
If the directories are not present, Singularity will fail, since with user privileges
they can not be created at runtime anymore.

Example error message observed on CentOS 7.4:
```
ERROR  : Failed creating overlay_final directory /var/singularity/mnt/final: Permission denied
ABORT  : Retval = 255
```
The really bad thing is things `just work` if the admin has ever tested singularity once as root user,
since this will trigger the directory creation.

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/debian/singularity-container.dirs
+++ b/debian/singularity-container.dirs
@@ -1,4 +1,7 @@
 var/lib/singularity
 var/lib/singularity/mnt
-var/lib/singularity/mnt/source
+var/lib/singularity/mnt/container
+var/lib/singularity/mnt/final
+var/lib/singularity/mnt/overlay
+var/lib/singularity/mnt/session
 usr/share/bash-completion
--- a/libexec/bootstrap-scripts/environment/Makefile.am
+++ b/libexec/bootstrap-scripts/environment/Makefile.am
@@ -41,7 +41,8 @@ environment.tar:
 	ln -sf .singularity.d/actions/run newroot/.run
 	ln -sf .singularity.d/actions/shell newroot/.shell
 	ln -sf .singularity.d/actions/test newroot/.test
-	cd newroot; tar czf ../environment.tar . --owner=0 --group=0
+	[ -n "${SOURCE_DATE_EPOCH}" ] && tar --help|grep -q sort= && taropts="--sort=name --clamp-mtime --mtime @${SOURCE_DATE_EPOCH} --format=gnu" ;\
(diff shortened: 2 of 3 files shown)


olifre:
@dexter2cz I did not yet test on a Debian system, could you have a look? 
For Debian, this should then also be "downstreamed" of course...

olifre:
Tests failed due to an issue with travis infra:
https://www.traviscistatus.com/incidents/v4jnp6nkgzqx

DrDaveD:
@olifre Aren't these directories only needed for the overlay feature, which doesn't work anyway if you aren't using setuid-root?

olifre:
@DrDaveD Most of them, yes - but `final`, which was the one missing from the `.spec` / `RPM` appears to be also used if:
```
allow setuid = no
enable overlay = no
```
is configured and I run singularity as normal user. 
The error message shown above was displayed to me in exactly this situation (and I fully agree that the naming `overlay_final` is pretty misleading here).

DrDaveD:
And how can a root-owned directory make a difference to a non-root privileged process?  Isn't it a bug that singularity is trying to create this directory when running unprivileged?

olifre:
> And how can a root-owned directory make a difference to a non-root privileged process? Isn't it a bug that singularity is trying to create this directory when running unprivileged?

You are correct, that's a good point. 

I'll look into that and open an extra issue - still, the packaging should include those directories (they are created during singularity compilation and should thus also become part of the package).

olifre:
> Isn't it a bug that singularity is trying to create this directory when running unprivileged?

@DrDaveD As you can see from my referenced issue #1096 you have been completely correct. 
Still, this does not invalidate this pull request (since the directories should be part of the packaging and not be created on-the-fly, and the packaging even is even done differently for the different distros in this regard). 

Thanks for pointing that out! 👍

GodloveD:
Adding bug tag so that we remember to cherry-pick this into 2.4.1 release if we get it merged in time.

olifre:
If possible, please also ping @dexter2cz or somebody from Debian so they can fix this downstream. 
Thanks!

DrDaveD:
@olifre, can you trigger another travis build now that docker with travis is fixed again?

This one should hopefully be merged soon.

olifre:
@DrDaveD Running - just took the chance to rebase to current master and re-push ;-).

gmkurtzer:
Excellent catch, thank you!

Issue #1078, opened 2017-10-18
wrong path for mksquashfs (squashfs-tools)


A33a:
## Version of Singularity:

2.4

### Expected behavior

Build a container.

### Actual behavior

Outputs:
```
frontend:31 16:10:21 ~/sjupyter > singularity build sjupyter.simg docker://a33a/sjupyter
ERROR: You must install squashfs-tools to build images
ABORT: Aborting with RETVAL=255
```

### Steps to reproduce behavior

I use shared filesystem to run the singularity and the squashfs-tools is installed there at /sharedfs/singularity/bin/mksquashfs, right in the bin derectory of singularity executable. $PATH variable contains that path, but when I try to build the image, it doesn't see it. 

What pathes build command uses? Thank you in advance.

P.S. I haven't noticed such behaviour in Singularity version 2.3.2. Doesn't it use squashfs-tools for bootsraping or does it create a "writable" image if squashfs-tools are not found? In version 2.4 build.exec checks for mksquashfs right in the beginning and there is no way to create any other kind of container.

DrDaveD:
I noticed this too, when writing my response in #940.  The problem is that /usr/libexec/singularity/cli/build.exec uses 'which mksquashfs' and there's no which command.  It should use singularity_which instead.

A33a:
@DrDaveD Thanks for an answer. Can you explain please why is there no which? Doesn't it run in the user's shell? There is the which in the system that I am using, does it utilize other binaries? Thank you in advance.

DrDaveD:
I'm not sure what shell you're using, but build.exec uses /bin/bash and it is not a builtin in bash.

GodloveD:
Hi @A33a!  Sorry you ran into this error.  

@DrDaveD is right.  `which` could be causing your problem.  

But there is another issue too.  Singularity explicitly sets the path [here](https://github.com/singularityware/singularity/blob/master/bin/singularity.in#L40).  This is a security measure to prevent a user from slipping an untrusted binary on the path and executing it with elevated privs.  

So `mksquashfs` needs to be installed in a standard system location to be found and used by Singularity.

A33a:
Hi @GodloveD! No problem, I am trying to make the application better :) Now I see that the problem was caused by PATH, because as soon as I moved `mksquashfs` to `/usr/local/bin` it was found (but another error occurred, #1079. It might not be caused by the current issue, though). 

[...]

PR #1094, opened 2017-10-26
adding ability to config location of mksquashfs

GodloveD:

Singularity sanitizes the `$PATH` when it initiates, so if your `mksquashfs` is not in a standard location Singularity won't find it.  This PR will allow an administrator to specify the location of `mksquashfs` in the `singularity.conf` file.

- Ref: #1078 

**Checkoff for all PRs:**

✅  I have tested this PR locally with a `make test`
✅  I have added myself as a contributor to the [contributors's file](https://github.com/singularityware/singularity/blob/master/CONTRIBUTORS.md)
✅   This PR is ready for review and/or merge

Attn: @singularityware-admin


libexec/cli/build.exec
```diff
@@ -36,7 +36,13 @@ else
     exit 1
 fi
 
-if ! singularity_which mksquashfs > /dev/null 2>&1; then
+# check to see if the location of mksquashfs is in singularity.conf
+tmp=`grep "mksquashfs location" "${SINGULARITY_sysconfdir}/singularity/singularity.conf" | awk -F = '{print $2}'`
+if [ -n "${tmp// }" ]; then
+    MKSQUASHFS_PATH=`dirname "${tmp}" | sed 's/^[ \t]*//;s/[ \t]*$//'`/
+fi
```
vsoch: I think this might break if the user provides the folder where mksquashfs is located, which (even though the instructions say otherwise) would be my naive thing to do (eg `/usr/local/bin` turns to `usr/local`) ?

```
tmp=/usr/local/bin
PATHY=`dirname "${tmp}" | sed 's/^[ \t]*//;s/[ \t]*$//'`/
$ echo $PATHY
/usr/local/
```

is there any reason not to, given that it's not empty, just append it to the path's searched for mksquashfs? Are you concerned that it opens up possibility to exploit those locations and want to limit to just one executable?

GodloveD: Thanks for reviewing this.  I added an example to the config file.  Does that help?  Really I could make it so that the user specifies a directory instead of the full path to the binary.  But I thought this was most intuitive and it should signal to the user that we are just adding one binary and not the entire directory.

Yes I want to add just that single binary and not the whole directory to the search path.  For security reasons I think it is good to minimize the number of things that Singularity has access to on the host system.

vsoch: yes I think that helps a lot, we would want to say "hey user, this is a direct path to an executable, not a path as you understand it!" Likely any bug reports because they did it wrong we can just point to the documented config. We should probably add more verbosity about it to the help pages too.

GodloveD:
This PR is no longer ready to merge.  Instead of `grep`-ing the `singularity.conf` file it should use a c program to do the job.  #1082 has the right idea.  Whenever it is merged I will merge development and re-write this to work `get-configvals.c`.

GodloveD:
OK.  This is now ready to merge.

bmcgough:
We use [EasyBuild](https://easybuilders.github.io/easybuild/) which uses [Lmod](https://www.tacc.utexas.edu/research-development/tacc-projects/lmod) to manage our cluster software. In order to provide the now-required (doc update?) squashfs-tools, we need this config change in order to be able to use Singularity. Researchers are blocked on this now as we also require the autofs bug path fix, so there is no version of Singularity that works for us.

ctmadison (approved):

yhcote (approved):
Could we also remove those warning message along the whole build ? At first I thought my build failed and tried 'make' again to see what the error was.

PR #1102, opened 2017-10-27
Make tar file creation reproducible

bmwiedemann:
See also #1083 

Make tar file creation reproducible
in order to make package builds reproducible.
See https://reproducible-builds.org/ for why this is good.

Requires GNU tar-1.28 or later to be reproducible
Using gzip -n to not add a timestamp to the .gz header

fixing or enhancing.

- Ref: #

**Checkoff for all PRs:**

- [X] I have tested this PR locally with a openSUSE package build

Attn: @singularityware-admin

### DIFF
--- a/libexec/bootstrap-scripts/environment/Makefile.am
+++ b/libexec/bootstrap-scripts/environment/Makefile.am
@@ -41,7 +41,8 @@ environment.tar:
 	ln -sf .singularity.d/actions/run newroot/.run
 	ln -sf .singularity.d/actions/shell newroot/.shell
 	ln -sf .singularity.d/actions/test newroot/.test
-	cd newroot; tar czf ../environment.tar . --owner=0 --group=0
+	[ -n "${SOURCE_DATE_EPOCH}" ] && tar --help|grep -q sort= && taropts="--sort=name --clamp-mtime --mtime @${SOURCE_DATE_EPOCH} --format=gnu" ;\
+	cd newroot; tar c . --owner=0 --group=0 $$taropts | gzip -n9 > ../environment.tar
 	rm -rf newroot
 
 MAINTAINERCLEANFILES = Makefile.in


vsoch:
If this requires version 1.28 or later, will it not work on older distributions? For example, just grabbing centos 6:

```
 singularity shell centos-6.simg 

Singularity centos-6.simg:~> which tar
/bin/tar
Singularity centos-6.simg:~> tar --version
tar (GNU tar) 1.23
Copyright (C) 2010 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by John Gilmore and Jay Fenlason.
Singularity centos-6.simg:~> 
```
and centos 7:
```
singularity shell centos-7.simg 

Singularity centos-7.simg:~> tar --version
tar (GNU tar) 1.26
Copyright (C) 2011 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by John Gilmore and Jay Fenlason.
Singularity centos-7.simg:~> 
```
so you are saying it is reproducible because a timestamp isn't added? Could you elaborate a bit?

bmwiedemann:
It will still work the same as before on older tar versions...
but will generate bit-identical tarballs on never versions.

> so you are saying it is reproducible because a timestamp isn't added? Could you elaborate a bit?

tar stores timestamps of source files and those timestamps usually vary between builds, because files are created dynamically during the build. And the default 'posix' or 'pax' format even stores some extra timestamps like ctime or atime that are never really needed.

You can play with it using
`touch foo ; tar -c foo | md5sum`
and reproducible
`touch foo ; tar -c --mtime @0 --owner=0 --group=0 --format=gnu foo | md5sum`

vsoch:
ah, understood! I really like this change then, we can wait to hear others thoughts.

gmkurtzer:
This is a fantastic idea, thank you!

I would like to see this check being done in configure.ac and then we export the correct tar arguments for all called tar commands (we are calling tar directly in several places). Would you like to take a stab at that @bmwiedemann and add that to this PR branch?

Thanks!

bmwiedemann:
@gmkurtzer It was the only place that mattered for our openSUSE package builds and I'm not that good at autoconf yet, so maybe you can do a draft patch and I just test it?

bmwiedemann:
also mentioning @yhcote and @GodloveD from the original PR

There was still the open question, if the environment.tar was really meant to be a mis-named .tar.gz file

bmwiedemann:
Anything I can help to get this merged or otherwise solved?

GodloveD:
Hi @bmwiedemann.  Sorry for the delay.  I think that @gmkurtzer is too busy to draft his own patch for this and test.  Do you have any interest in trying to rewrite this PR to use configure.ac like he suggested?  Thanks!

c-ney:
I pocked around a little and found some pointers. First one should define the `SOURCE_DATE_EPOCH` [1] in `configure.ac`, maybe take the date of the current HEAD commit? The following only works when building from a git checkout and I have currently no idea how to best get this information inside release tarballs.

```
SOURCE_DATE_EPOCH=$(git show -s --format=%ct HEAD)
AC_SUBST(SOURCE_DATE_EPOCH)
```

Then one needs to look out for the PAX extended headers when using `--format=posix`, as they include by default `atime` and `ctime`. These can be set to fixed values or removed with `--pax-option` in gnu tar [2].

AFAICS there is no way to pass arguments to the tar.m4 macro, so `am__tar` (and `AMTAR`?) have to be overwritten after `AM_INIT_AUTOMAKE` in `configure.ac`. I threw together a substitution that works for gnu tar 1.28 and later. This should be extended to check for available tar implementations.

```
am__tar='tar --format=posix --pax-option=exthdr.name=%d/PaxHeaders/%f,atime:=0,ctime:=0 --clamp-mtime --mtime @${SOURCE_DATE_EPOCH} --sort=name --owner=0 --group=0 -chf - "$$tardir"'
AC_SUBST([am__tar])
```

To set gzip options globally use

```
AC_SUBST([GZIP_ENV],[-n9])
```

In `libexec/bootstrap-scripts/environment/Makefile.am` we could then use (or something similar, just copy&pasted this from the dist target)

```
cd newroot; tardir=. && $(am__tar) | eval GZIP= gzip $(GZIP_ENV) -c >../environment.tar
```

[1] https://reproducible-builds.org/specs/source-date-epoch/#idm53
[2] https://www.gnu.org/software/tar/manual/html_section/tar_70.html#SEC147

Issue #1111, opened 2017-10-28
Additional tests for consideration


justbennet:
## Version of Singularity: post 2.4 development

I encountered several times errors because a) curl was not installed (seems not to be default in some Ubuntus), b) the install path was not in the sudo secure path, and c) I chose a path on a volume mounted with 'nosuid' option.

### Expected behavior

Those are, I think, fairly easy to detect early in the game.

I'll try to gin up some tests and put them into a PR.

### Actual behavior

Errors occur at various places.

### Steps to reproduce behavior

Uninstall curl; install to partition with 'nosuid' mount option; run `make test`

dtrudg:
These installation issues should be addressed in 3.x to my knowledege. If they aren't we'd be happy to look at an issue against the current 3.x supported release.

PR #1113, opened 2017-10-28
Adding a couple of new tests

justbennet:

Adding early tests for things that need to be modified or installed prior to extensive testing.

1)  If singularity is on a disk mounted with nosuid option, it will fail later
2) If singularity is not in a directory that is in the sudo secure_path, it will fail later
3) If curl is not installed, it will fail later

These can be checked early, and as all of them require making changes to filesystems before the `make test` can continue.  No changes to the Singularity code is included.

I tried to use the framework already in use.

Light laughing at my woeful shell script is encouraged, but ROTFL should be done privately.  ;-)

- Ref: #1111

**Checkoff for all PRs:**

- [ x ] I have read the [Guidelines for Contributing](https://github.com/singularityware/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- [ x ] I have tested this PR locally with a `make test`
- [ x ] This PR is NOT against the project's master branch
- [ x ] I have not added myself as a contributor to the [contributors's file](https://github.com/singularityware/singularity/blob/master/CONTRIBUTORS.md)  -- too small a change to be a contributor.
- [ x ] This PR is ready for review and/or merge

Attn: @singularityware-admin

### DIFF
--- a/tests/01-sanity.sh
+++ b/tests/01-sanity.sh
@@ -29,4 +29,18 @@ stest 1 sudo false
 stest 0 which singularity
 stest 0 test -f "$SINGULARITY_sysconfdir/singularity/singularity.conf"
 
+# Check whether singularity binary is on a volume mounted with 'nosuid'
+# I guess one could just grep /etc/mtab, but...
+mount | grep $(df -h $SINGULARITY_PATH | grep -v File | awk '{print $1}') \
+    > $SINGULARITY_TESTDIR/singularity_fs_nosuid
+stest 1 grep "nosuid" $SINGULARITY_TESTDIR/singularity_fs_nosuid
+
+# Is the SINGULARITY_PATH going to be there when you sudo?
+sudo grep "^Defaults.*secure_path=.*$SINGULARITY_PATH" /etc/sudoers
+echo $? > $SINGULARITY_TESTDIR/singularity_secure_path
+stest 0 grep 0 $SINGULARITY_TESTDIR/singularity_secure_path
+
+# Yes, Virginia, we use curl in these tests
+stest 0 which curl
+
 test_cleanup


tests/01-sanity.sh
```diff
@@ -29,4 +29,18 @@ stest 1 sudo false
 stest 0 which singularity
 stest 0 test -f "$SINGULARITY_sysconfdir/singularity/singularity.conf"
 
+# Check whether singularity binary is on a volume mounted with 'nosuid'
+# I guess one could just grep /etc/mtab, but...
+mount | grep $(df -h $SINGULARITY_PATH | grep -v File | awk '{print $1}') \
+    > $SINGULARITY_TESTDIR/singularity_fs_nosuid
+stest 1 grep "nosuid" $SINGULARITY_TESTDIR/singularity_fs_nosuid
```
GodloveD, line 36: Is there a better way to do this?  Without creating the temporary file?

tests/01-sanity.sh
```diff
@@ -29,4 +29,18 @@ stest 1 sudo false
...
+
+# Is the SINGULARITY_PATH going to be there when you sudo?
+sudo grep "^Defaults.*secure_path=.*$SINGULARITY_PATH" /etc/sudoers
+echo $? > $SINGULARITY_TESTDIR/singularity_secure_path
+stest 0 grep 0 $SINGULARITY_TESTDIR/singularity_secure_path
```
GodloveD, line 41: What would happen if you just did something like?
```
stest 0 sudo singularity --version
```

GodloveD (commented):
Hi @justbennet.  Thanks very much for identifying these potential improvements and taking the time to code them up and send us a PR.  Sorry for the long turn around time.  We are woefully behind on reviewing PRs.  

I like this idea, but I think it could probably be implemented a bit cleaner.  Is there any easier way to check if the disk is mounted with the nosuid option?   

Also, it looks like these tests are actually not passing in Travis.  So I don't think they are working as expected right?

tests/01-sanity.sh
```diff
@@ -29,4 +29,18 @@ stest 1 sudo false
 stest 0 which singularity
 stest 0 test -f "$SINGULARITY_sysconfdir/singularity/singularity.conf"
 
+# Check whether singularity binary is on a volume mounted with 'nosuid'
+# I guess one could just grep /etc/mtab, but...
+mount | grep $(df -h $SINGULARITY_PATH | grep -v File | awk '{print $1}') \
```
jmstover, line 34: I would also be paranoid and do something like:

`mount | grep $(df -h $SINGULARITY_PATH | grep -v '^Filesystem' | awk '{print $1}')`

Just make sure the grep on the `df` output matches 'Filesystem' at the start of a line. You don't want to know some of the names I've seen mounted filesystems... Your grep would have ignored the mount point.

*Edit:*
Hrmm.... on 2nd thought, you could remove the grep entirely and just use: `tail -n 1`
Just remove the possibility of accidentally catching the mountpoint entirely.

tests/01-sanity.sh
jmstover: grepping for secure_path would fail for me as I don't have it set. So.. since the grep fails it returns 1, but running `sudo singularity` works fine.

```
$ sudo grep '^Defaults.*secure_path' /etc/sudoers
$ echo $?
1
$ which singularity
/usr/local/singularity/devel/bin/singularity
$ sudo singularity --version
2.4.1-development.g08dc27a
```

justbennet:
Unfortunately, I don't have time to get back to this right now.  I start teaching next Wed, and I have to make sure I am at least a week ahead....  I'll try to sneak a look at these again, but I can't promise when.  The basic idea was just to check the system status prior to building images that would fail for things that could be detected with some system queries and grep.

I'll make a note to try to check back on these in Feb.  Sorry for the lag.

Issue #1090, opened 2017-10-24
Missing lists of runtime and build-time dependencies


olifre:
## Version of Singularity:

Any.

I just learnt from the mailing list that `squashfs-tools' is an explicit runtime requirement of Singularity 2.4. 
Checking around, I found several issues: 
- The Debian packaging has not been adapted in-tree. Only RPM got a fix: https://github.com/singularityware/singularity/commit/78af5f54b814466fe88d362688b78e016ba4c618#diff-b3fb1fb02c192e0be6fee00045236e40
- The release notes do not mention this new requirement. 

As a consequence, packaging in downstream distributions is broken. 
For example, Gentoo (which is already shipping 2.4 as of now) did not take this new dependency into account. 
That's totally expected: How should downstream maintainers, or users, know about this? 

This issue strongly interplays with https://github.com/singularityware/singularity/issues/1051 .

olifre:
Here's the downstream Gentoo bug caused by this issue:
https://bugs.gentoo.org/635342

olifre:
@dexter2cz Also the downstream official Debian package is broken in this regard:
https://packages.debian.org/experimental/singularity-container
Since I am not familiar with Debian bug reporting, it might be best if somebody else took care of that...

dtrudg:
Closing this as a PR was merged as a fix. Please don't hesitate to reopen if there is any outstanding issue.

Issue #1057, opened 2017-10-14
Trying to build a 2.4 Debian package from source currently generates one labeled 2.3.2:


amacbride:
## Version of Singularity:

2.4

### Expected behavior

Generates a Debian package for version 2.4

### Actual behavior

Generates a package labelled 2.3.2

### Steps to reproduce behavior

```% wget https://github.com/singularityware/singularity/releases/download/2.4/singularity-2.4.tar.gz
% tar xvf singularity-2.4.tar.gz
% cd singularity-2.4
% ./configure --prefix=/usr/local
% sudo apt-get install help2man
% fakeroot dpkg-buildpackage -b -us -uc
```

The build product is: `../singularity-container_2.3.2-1_amd64.deb` (so I'm not sure what's actually in it.)

olifre:
The reason appears to be that during release, somebody forgot to update the changelog here:
https://github.com/singularityware/singularity/blob/master/debian/changelog
Debian packaging pulls the version out of the changelog ( see https://github.com/singularityware/singularity/blob/master/debian/rules ). 

Since @gmkurtzer did the tag and release, giving him a friendly ping here ;-).

gmkurtzer:
Yeah, that was totally my fault. We should probably change it such that it uses the autoconf @VERSION@ string rather then a hard coded tag. Thanks for the ping!

gmkurtzer:
Actually, I take that back. I am looking at how Debian does their building, and the only thing I've found for how it ascertains it's version number is from the changelog. Is that accurate?

olifre:
Yes, indeed, I think the version information is pulled from the Changeling only. AFAIK this is a common technique for Debian, and they also uses this to ensure that a downstream change which enters the Changelog can trigger a new downstream version (such as 2.4-2).

smoe:
Hello. I am developer for Debian and confirm that this are humans who look at the package and kind of read from the changelog or the tarball's filename what the version may be. If in doubt, they ask. As correctly described by @olifre, the number after the "-" starts with 1 for every new upstream version and is increased with every attempt by Debian folks to improve the packaging. 

There is package for Singularity shipping with Debian described at https://tracker.debian.org/pkg/singularity-container . I am not involved with its packaging but typically the maintainers are more than pleased to communicate with/help upstream with whatever issue. Just contact them.

dtrudg:
Closing this as I believe we are now confident on release numbering procedure. Thanks @smoe for this information!

PR #1114, opened 2017-10-28
Debian squashfs-tools dependency

dexter2cz:

Added squashfs-tools as debian dependency

- Ref: #1090

**Checkoff for all PRs:**

Attn: @singularityware-admin @yarikoptic

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -20,7 +20,7 @@ Vcs-Browser: https://github.com/singularityware/singularity
 # "singularity" is a packaged game (but the contents don't clash)
 Package: singularity-container
 Architecture: any
-Depends: ${misc:Depends}, ${shlibs:Depends}, python
+Depends: ${misc:Depends}, ${shlibs:Depends}, python, squashfs-tools
 Description: container platform focused on supporting "Mobility of Compute"
  Mobility of Compute encapsulates the development to compute model
  where developers can work in an environment of their choosing and


GodloveD (approved):
This looks good to me.

GodloveD:
Tagging with bug so it gets our attention and makes it into the 2.4.1 release.

vsoch:
@GodloveD we would want every PR to have a clean line added to the CHANGELOG no?

GodloveD:
I think we only want things in the changelog that impact the user interface.  For instance, if you refactor a library, but it is exactly the same from a user's point of view.  I don't think it belongs in the changelog.  Otherwise the change log is just a copy of the git log.

vsoch:
+1, and it's always good to ask/check!

yarikoptic:
changes to debian/  are subject to be noted in debian/changelog.  The purpose is to have the log describing changes to packaging. git clone of the repo might not be available.

BTW -- note that changes to debian/ in this repo aren't automagically picked up by me for the official debian package.  that one is packaged out of released source tarballs and its "main" git at https://anonscm.debian.org/git/collab-maint/singularity-container.git  .  I feel that we might better avoid such dichotomy one way or another

vsoch:
how do you  think is best to go about it?

yarikoptic:
If noone uses `debian/` within this (upstream) codebase, then the best, while also staying inline with the debian community, is just remove `debian/` here and add a note that patches to packaging should be submitted to Debian (`reportbug` tool).
But if someone sees the value of having debian packaging straight in this repo, we could come up with other approaches

vsoch:
@olifre I think you added the debian changelog, what are your thoughts on this? We could maintain one changelog here, and then @yarikoptic references to update his, or the equivalent with two. @yarikoptic I think (maybe?) the general idea was to make the packaging possible from this repo (@olifre can comment) in which case, what are  your thoughts on that?

olifre:
@vsoch I did not add the changelog, I only complained when it was not updated with the last release, and added a line to the release checklist to keep it updated ;-). 
As you can see from #1057 , people appear to use the `debian/` files within this repo to build packages. 

I can only speak for us: We generally pull packaging code from a distro (e.g. debian) and base our build to roll out on our cluster and desktops on their packaging (we do our on packaging for some things to have better control over "when updates are rolled out" and have updates in-sync between RedHat and Debian). 
On the other hand, if the packaging code in the project is well maintained and updated (e.g. with new dependencies), this allows both the distro to base their initial packaging on it, and allows users to base their packaging on it. 
So I'd think it's nice to have packaging code "someplace singularity maintained" to give all distros and users a good starting point. It's also a clean place to document new dependencies, and would allow you to offer independently built deb's and rpm's easily, if you decide to do that. 

Of course, if there is someone within the singularity team who has a short turnaround time to contribute directly to debian ( @yarikoptic ?) and feels detecting new dependencies is possible without maintained packaging code directly in Singularity, that's fine with me. Then I'll keep looking at official debian packaging and pull from there. 

Maybe, in this context, I should also mention https://github.com/singularityware/singularity/pull/1093 which still has entered neither the singularity packaging code nor debian.

vsoch:
@yarikoptic and @dexter2cz I don't have much opinion here - what is your preference?

gmkurtzer:
This looks good to me, so I'm gonna merge. Thanks!

Issue #1087, opened 2017-10-24
Cannot pull image from local docker registry at paths like docker.internal/image


ilmagico:
## Version of Singularity:

Tested on 2.3 and 2.4 and latest development branch.

### Expected behavior

To be able to build an image from a local docker registry where the image is at, say, docker.internal/image.
Note that there's no namespace/image, and this is actually not required, as shown in this page from docker's own docs: https://docs.docker.com/registry/deploying/#copy-an-image-from-docker-hub-to-your-registry
In the example from the docker documentation page above, an image is uploaded to a local registry simply as localhost:5000/my-ubuntu without any namespace.

### Actual behavior

Singularity doesn't parse the image name correctly and tries to pull it from docker's main hub, index.docker.io:

```
$ cat Singularity 
Bootstrap: docker
From: docker.internal/brewst:latest
$ sudo singularity build test.simg Singularity 
Using container recipe deffile: Singularity
Sanitizing environment
Adding base Singularity environment to container
Docker image path: index.docker.io/docker.internal/brewst:latest
ERROR UNAUTHORIZED: authentication required
ERROR Check existence, naming, and permissions
Cleaning up...
$ 
```

### Steps to reproduce behavior

See above. Basically, start a local docker registry and upload an image like in the example at https://docs.docker.com/registry/deploying/#copy-an-image-from-docker-hub-to-your-registry, then try to pull it from singularity.
Here are the full instructions, including spinning up a local registry, uploading a test image, then trying to pull it down with singularity:

```
$ sudo docker run -d -p 5000:5000 --restart=always --name registry registry:2
Unable to find image 'registry:2' locally
2: Pulling from library/registry
90f4dba627d6: Pull complete 
b3e11d7b4f5e: Pull complete 
1f032f3c8932: Pull complete 
425585e7aedb: Pull complete 
f45f535a83d2: Pull complete 
Digest: sha256:0f8fe61fa337b8ef02217702ba979b47a7d68717d4628f31592ebff85915f3ba
Status: Downloaded newer image for registry:2
fe5bb4406a0086fde42252ce4e52cbb7f853e48442938cd30c14974985c0e46d
$ sudo docker ps
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS                                            NAMES
fe5bb4406a00        registry:2          "/entrypoint.sh /e..."   5 seconds ago       Up 4 seconds        0.0.0.0:5000->5000/tcp                           registry
[...]

PR #1116, opened 2017-10-30
Allow empty namespaces for docker:// images

ilmagico:

Docker allows local/private registries to have images in the top-level namespace, as in `docker://local.registry:5000/someimage` but currently singularity enforces a default namespace, which results in errors if that is not required. The main docker hub (index.docker.io) *does* require a 'library' namespace, so it's a good default to have this, but there should be a way to override this.
This pull requests does that in a couple of ways:
* docker:// URI parsing now follows more closely docker's own rules, so that an image specified as above will parse with an empty namespace (the specific rule is that a custom registry must be present in the image)
* deffiles are allowed to have empty values, so one can use `Namespace:` (no value after `:`) to enforce an empty namespace

As to which rules one should use to parse docker:// image uris, there doesn't seem to be a lot of documentation from docker itself. I found these articles which seem to imply certain rules:
* https://docs.docker.com/registry/deploying/: shows how to set up a local registry, and the example uses the top level namespace
* https://blog.docker.com/2013/07/how-to-use-your-own-registry/: while a little dated, hopefully this quote still applies: 

    > It’s important to note that we’re using a domain containing a “.” here, i.e. `localhost.domain`. Docker looks for either a “.” (domain separator) or “:” (port separator) to learn that the first part of the repository name is a location and not a user name. If you just had `localhost` without either `.localdomain` or `:5000` (either one would do) then Docker would believe that localhost is a username, as in `localhost/ubuntu` or `samalba/hipache`

    where of course what docker calls username is what singularity calls namespace.

Based on that, I decided to enforce the same rule for the registry to contain either "." or ":", and when a registry is specified, allow the namespace to be empty. If the registry is not specified, use the default namespace ('library').

I originally tried to make simple incremental modifications to `parse_image_uri` in `shell.py`, but the code got very complicated quickly, to the point that I decide to switch using regular expressions instead. I know how they can be cryptic sometimes, so I added plenty of comments, and I believe it's still much clearer than how it would've turned out otherwise.

This is meant mostly for review, I can squash commits, or split this into multiple PRs if needed, just let me know. It is otherwise, in my opinion, ready for merge (but please review!).

- Ref: #1087

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/libexec/bootstrap-scripts/deffile-driver-docker.sh
+++ b/libexec/bootstrap-scripts/deffile-driver-docker.sh
@@ -51,7 +51,9 @@ if [ ! -z "${REGISTRY:-}" ]; then
     export REGISTRY
 fi
 
-if [ ! -z "${NAMESPACE:-}" ]; then
+# Note: NAMESPACE can be set to an empty string, and that's a valid namespace
+# for Docker (not so for shub://)
+if [ -n "${NAMESPACE+set}" ]; then
     message DEBUG "Custom Docker Namespace 'Namespace:' ${NAMESPACE}.\n"
     export NAMESPACE
 fi
--- a/libexec/python/defaults.py
+++ b/libexec/python/defaults.py
@@ -50,7 +50,8 @@ def getenv(variable_key, default=None, required=False, silent=False):
         sys.exit(1)
 
     if silent:
-        bot.verbose2("%s found" % (variable_key))
+        if variable is not None:
+            bot.verbose2("%s found" % (variable_key))
     else:
         if variable is not None:
             bot.verbose2("%s found as %s" % (variable_key, variable))
--- a/libexec/python/shell.py
+++ b/libexec/python/shell.py
@@ -208,7 +208,7 @@ def parse_image_uri(image,
     if uri == "docker://":
         uri_regexes = [ _docker_uri_re ]
 
-    image = remove_image_uri(image, uri);
+    image = remove_image_uri(image, uri)
 
     for r in uri_regexes:
         match = r.match(image)
@@ -259,7 +259,6 @@ def parse_image_uri(image,
               'namespace': namespace,
               'repo_name': repo_name,
               'repo_tag': repo_tag,
-              'version': version,
-              }
+              'version': version }
 
     return parsed
--- a/libexec/python/shell.py
+++ b/libexec/python/shell.py
@@ -59,7 +59,8 @@ def get_image_uri(image, quiet=False):
 
 
 def remove_image_uri(image, image_uri=None, quiet=False):
-    '''remove_image_uri will return just the image name
+    '''remove_image_uri will return just the image name.
+    ::note this will also remove all spaces from the uri.
     '''
     if image_uri is None:
         image_uri = get_image_uri(image, quiet=quiet)
@@ -70,24 +71,111 @@ def remove_image_uri(image, image_uri=None, quiet=False):
         image = image.replace(image_uri, '')
     return image
 
+# Quick python regex syntax reference for referring back to matched groups:
+# (?:stuff) matches stuff, just like (stuff) but won't appear in m.groups()
+# (?P<name>stuff) matches stuff, and can be retrieved by m.group('name')
+
+# Regular expression used to parse docker:// uris, which have slightly
+# different rules than shub://, namely:
+# - registry must include :port or a . in the name (e.g. docker.io)
+# - namespace is completely optional
+_docker_uri_re = re.compile(
+    # Optionally match a registry if it contains a '.' or a ':',
+    # may contain any character but '/', and terminated by '/'
+    # Note the use of (?:) to make sure the group "registry" does
+    # not contain the / at the end. Also, registry includes the port
+    "(?:(?P<registry>[^/]+[.:][^/]*)/)?"
+    # Optionally match a namespace, matched as any characters but
+    # ':' or '/', followed by '/'. Note the match will include the final /
+    "(?P<namespace>(?:[^:@/]+/)+)?"
+    # Match a repo name, mandatory. Any character but ':' or '/'
+    "(?P<repo>[^:@/]+)"
+    # Match :tag (optional)
+    "(?::(?P<tag>[^:]+))?"
+    # Match @digest (optional)
+    "(?:@(?P<version>[^@]+))?"
+    # we need to match the whole string, make sure there's no leftover
+    "$"
+    )
+
+
+# Reduced regex, matches registry:port/repo or registry.com/repo
+# (registry must include : or .) with optional tag or version.
+# Also matches repo only, e.g. reponame[:tag|@version].
+# This is tried before _default_uri_re, so that namespace/repo takes
+# precedence over registry/repo.
+_reduced_uri_no_ns_re = re.compile(
+    # match a registry, optional, may include a : or .
+    "(?:(?P<registry>[^/]+[.:][^/]*)/)?"
+    # Match a repo name, mandatory. Any character but ':' or '/'
+    "(?P<repo>[^:@/]+)"
+    # Match :tag (optional)
+    "(?::(?P<tag>[^:]+))?"
+    # Match @digest (optional)
+    "(?:@(?P<version>[^@]+))?"
+    # we need to match the whole string, make sure there's no leftover
+    "$"
+    # dummy group that will never match, but will add a 'namespace' entry
+    "(?P<namespace>.)?"
+    )
+
+# Regular expression used to parse other images and uris (e.g. shub://)
+# This version expects a namespace, and won't match otherwise
+# but the registry is optional. In cases like a/b/c, the first component
+# is taken to be the registry in this regex.
+_default_uri_re = re.compile(
+    # must be either registry[:port]/namespace[/more/namespaces]/repo
+    # or namespace/repo, at least one namespace is required
+    # Match registry, if specified
+    "(?:(?P<registry>[^/]+)/)?"
+    # Match a namespace, matched as any characters but
+    # ':' or '@', ended by a '/'. Note the match will include the final /
+    "(?P<namespace>(?:[^:@/]+/)+)"
+    # Match a repo name, mandatory. Any character but ':' or '/'
+    "(?P<repo>[^:@/]+)"
+    # Match :tag (optional)
+    "(?::(?P<tag>[^:]+))?"
+    # Match @digest (optional)
+    "(?:@(?P<version>[^@]+))?"
+    # we need to match the whole string, make sure there's no leftover
+    "$"
+    )
+
 def parse_image_uri(image,
                     uri=None,
                     quiet=False,
                     default_registry=None,
                     default_namespace=None,
                     default_tag=None):
+    '''parse_image_uri will parse a docker or shub uri and return
+    a json structure with a registry, repo name, tag, namespace and version.
+    Tag and version are optional, namespace is optional for docker:// uris.
+    URIs are of this general form:
+        myuri://[registry.com:port/][namespace/nested/]repo[:tag][@version]
+    (parts in [] are optional)
+    Parsing rules are slightly different if the uri is a docker:// uri:
+    - registry must include a :port or a . in its name, else will be parsed
+      as a namespace. For non-docker uris, instead, if there are three or
+      more parts separated by /, the first one is taken to be the registry
+    - namespace can be empty if a registry is specified, else default
+      namespace will be used (e.g. docker://registry.com/repo:tag).
+      For non-docker uris, namespace cannot be empty and default will be used
 
-    '''parse_image_uri will return a json structure with a registry,
-    repo name, tag, and namespace, intended for Docker.
     :param image: the string provided on command line for
-                  the image name, eg: ubuntu:latest
-    :param uri: the uri (eg, docker:// to remove), default uses ""
+                  the image name, eg: ubuntu:latest or
+                  docker://local.registry/busybox@12345
+    :param uri: the uri type (eg, docker://), default autodetects
     ::note uri is maintained as variable so we have some control over allowed
+    :param quiet: If True, don't show verbose messages with the parsed values
+    :default_registry: Which registry to use if image doesn't contain one.
+                       if None, use defaults.REGISTRY
+    :default_namespace: Which namespace to use if image doesn't contain one.
+                       if None, use defaults.NAMESPACE. Also, check out the
+                       note above about docker:// and empty namespaces.
+    :default_tag: Which tag to use if image doesn't contain one.
+                       if None, use defaults.REPO_TAG
     :returns parsed: a json structure with repo_name, repo_tag, and namespace
     '''
-
-    if uri is None:
-        uri = ""
 
     # Default to most common use case, Docker
     if default_registry is None:
@@ -99,63 +187,66 @@ def parse_image_uri(image,
     if default_tag is None:
         default_tag = TAG
 
-    # Be absolutely sure there are not comments
+    # if user gave custom registry / namespace, make them the default
+    if CUSTOM_NAMESPACE is not None:
+        default_namespace = CUSTOM_NAMESPACE
+
+    if CUSTOM_REGISTRY is not None:
+        default_registry = CUSTOM_REGISTRY
+
+    # candidate regex for matching, in order of preference
+    uri_regexes = [ _reduced_uri_no_ns_re,
+                    _default_uri_re ]
+
+    # Be absolutely sure there are no comments
     image = image.split('#')[0]
 
-    # Get rid of any uri, and split the tag
-    image = image.replace(uri, '')
+    if not uri:
+        uri = get_image_uri(image, quiet=True)
 
-    # Does the uri have a digest or Github tag (version)?
-    image = image.split('@')
-    version = None
-    if len(image) == 2:
-        version = image[1]
+    # docker images require slightly different rules
+    if uri == "docker://":
+        uri_regexes = [ _docker_uri_re ]
 
-    image = image[0]
-    image = image.split(':')
+    image = remove_image_uri(image, uri);
 
-    # If there are three parts, we have port and tag
-    if len(image) == 3:
-        repo_tag = image[2]
-        image = "%s:%s" % (image[0], image[1])
+    for r in uri_regexes:
+        match = r.match(image)
+        if match:
+            break
 
-    # If there are two parts, we have port or tag
-    elif len(image) == 2:
-        # If there isn't a slash in second part, we have a tag
-        if image[1].find("/") == -1:
-            repo_tag = image[1]
-            image = image[0]
-        # Otherwise we have a port and we merge the path
+    if not match:
+        bot.error('Could not parse image "%s"! Exiting.' % image)
+        sys.exit(1)
+
+    registry = match.group('registry')
+    namespace = match.group('namespace')
+    repo_name = match.group('repo')
+    repo_tag = match.group('tag')
+    version = match.group('version')
+
+    if namespace:
+        # strip trailing /
+        namespace = namespace.rstrip('/')
+
+    # repo_name is required and enforced by the re (in theory)
+    # if not provided, re should not match
+    assert(repo_name)
+
+    # replace empty fields with defaults
+    if not namespace:
+        # for docker, if a registry was specified, don't
+        # inject a namespace in between
+        if registry and uri == "docker://":
+            namespace = ""
         else:
-            image = "%s:%s" % (image[0], image[1])
-            repo_tag = default_tag
-    else:
-        image = image[0]
+            namespace = default_namespace
+    if not registry:
+        registry = default_registry
+    if not repo_tag:
         repo_tag = default_tag
+    # version is not mandatory, don't check that
 
-    # Now look for registry, namespace, repo
-    image = image.split('/')
-
-    if len(image) > 2:
-        registry = image[0]
-        namespace = "/".join(image[1:-1])
-        repo_name = image[-1]
-
-    elif len(image) == 2:
-        registry = default_registry
-        namespace = image[0]
-        repo_name = image[1]
-
-    else:
-        registry = default_registry
-        namespace = default_namespace
-        repo_name = image[0]
-
-    # if user gave custom registry / namespace, use
-    if CUSTOM_NAMESPACE is not None:
-        namespace = CUSTOM_NAMESPACE
-    if CUSTOM_REGISTRY is not None:
-        registry = CUSTOM_REGISTRY
 
     if not quiet:
         bot.verbose("Registry: %s" % registry)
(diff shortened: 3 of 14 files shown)


ilmagico:
Tests on CentOS 7 have failed because the 'docker' command was not found, and I added a test that calls docker. They all passed Ubuntu.
Is it possible to add docker to CentOS 7? I'm not too familiar how to do this with travis.
Edit: I tried running `sudo OS_TYPE=centos OS_VERSION=7 .travis/script` from a vagrant ubuntu box, and it makes the whole ubuntu VM freeze, so I'm not able to test this either.

GodloveD:
Tagging this as a bug so that we can include in 2.4.1 if we get it sorted out and merged in time.

ilmagico:
Finally! ;)

vsoch:
Totally epic @ilmagico! I will make some time to review this tomorrow. Woot!

libexec/bootstrap-scripts/deffile-driver-docker.sh
```diff
@@ -51,7 +51,9 @@ if [ ! -z "${REGISTRY:-}" ]; then
     export REGISTRY
 fi
 
-if [ ! -z "${NAMESPACE:-}" ]; then
+# Note: NAMESPACE can be set to an empty string, and that's a valid namespace
+# for Docker (not so for shub://)
+if [ -n "${NAMESPACE+set}" ]; then
     message DEBUG "Custom Docker Namespace 'Namespace:' ${NAMESPACE}.\n"
```
vsoch, line 57: Question 1: what is the difference between `[ ! -z "${NAMESPACE:-}" ]` and `[ ! -z "${NAMESPACE:set}" ]`? Is there reason that `NAMESPACE` has the fallback as `set` but not `REGISTRY` or `FROM` (or any of the checks above?) 

And very good job adding tests for these!

libexec/python/defaults.py
```diff
@@ -50,7 +50,8 @@ def getenv(variable_key, default=None, required=False, silent=False):
         sys.exit(1)
 
     if silent:
-        bot.verbose2("%s found" % (variable_key))
+        if variable is not None:
+            bot.verbose2("%s found" % (variable_key))
     else:
```
vsoch, line 55: I was first worried we would be hiding good debugging into (e.g., I would want to know when a variable is set to None) but it probably communicates that by way of not printing, so I like this approach better to produce cleaner python logs. I'm down to give it a try! The variable silent is poorly named (my bad-doing) because it's intended to hide sensitive information like passwords, tokens, while still alerting the user that it's set. This is different than performing the print action - because the message bot inherits the `SINGULARITY_MESSAGELEVEL`.

libexec/python/shell.py
```diff
@@ -99,63 +187,66 @@ def parse_image_uri(image,
...
-    image = image.split('/')
+    # docker images require slightly different rules
+    if uri == "docker://":
+        uri_regexes = [ _docker_uri_re ]
 
-    if len(image) > 2:
-        registry = image[0]
-        namespace = "/".join(image[1:-1])
-        repo_name = image[-1]
+    image = remove_image_uri(image, uri);
```
vsoch: I think the ";" snuck in from another language... :O

libexec/python/shell.py
```diff
@@ -99,63 +187,66 @@ def parse_image_uri(image,
...
+    repo_tag = match.group('tag')
+    version = match.group('version')
+
+    if namespace:
+        # strip trailing /
+        namespace = namespace.rstrip('/')
+
+    # repo_name is required and enforced by the re (in theory)
+    # if not provided, re should not match
+    assert(repo_name)
```
vsoch, line 234: This is a good test! Can we catch this `AssertionError` and give a meaningful message to the user? I think as of now it would just throw the error and exit.

libexec/python/shell.py
```diff
@@ -167,15 +258,8 @@ def parse_image_uri(image,
...
-    for fieldname, value in parsed.items():
-        if len(value) == 0:
-            bot.error("%s found empty, check uri! Exiting." % value)
-            sys.exit(1)
-
-    # Version is not required
-    parsed['version'] = version
+              'repo_tag': repo_tag,
+              'version': version,
+              }
```
vsoch: stylistic thing, can we keep:

```
    parsed = {'registry': registry,
              'namespace': namespace,
              'repo_name': repo_name,
              'repo_tag': repo_tag,
              'version': version }
```

libexec/python/tests/test_core.py
```diff
@@ -163,6 +163,218 @@ def test_parse_image_uri(self):
         self.assertTrue(digest['namespace'] == 'mix/original/choice')
         self.assertTrue(digest['version'] == 'sha:256xxxxxxxxxxxxxxx')
 
+        # now test some tricky cases
+
```
vsoch, line 167: beautiful job with the new tests.

libexec/python/shell.py
```diff
@@ -99,63 +187,66 @@ def parse_image_uri(image,
...
+        repo_tag = default_tag
+    # version is not mandatory, don't check that
 
-    # if user gave custom registry / namespace, use
-    if CUSTOM_NAMESPACE is not None:
-        namespace = CUSTOM_NAMESPACE
-    if CUSTOM_REGISTRY is not None:
-        registry = CUSTOM_REGISTRY
 
     if not quiet:
```
vsoch, line 251: It looks like if the version is not long / missing numbers it gets parsed as part of the tag instead of version:

```
parse_image_uri('docker://localhost:5000/long/name/ubuntu:latest@2.2')
Out[46]: 
{'namespace': 'long/name',
 'registry': 'localhost:5000',
 'repo_name': 'ubuntu',
 'repo_tag': 'latest@2.2',
 'version': None}
```

this is ok for docker because we won't see much outside of sha256xxxxx, but for shub, a digest can also be a commit or a branch. For example, here we are seeing the digest added to the tag:

```
parse_image_uri('shub://namespace:5000/one/two/ubuntu:latest@master')

{'namespace': 'one/two',
 'registry': 'namespace:5000',
 'repo_name': 'ubuntu',
 'repo_tag': 'latest@master',
 'version': None}
```

**returns from 40 minutes of testing**
ok, so the slight mis-parsing isn't an issue with pulling images from shub because the url squashes the tag with the digest anyway, so it's just a designation for the function. In terms of function, this should be ok because it gets the job done. I'm more concerned with users seeing the mis-parsed uri and reporting issue, and it looks like it's set to be quiet so this won't come up. I think this is acceptable for now and we can revisit if needed.

libexec/python/shub/api.py
```diff
@@ -80,6 +80,11 @@ def load_image(self, image):
                                      uri='shub://',
                                      default_registry=SHUB_API_BASE,
                                      quiet=True)
+        # parse_image_uri may return an empty namespace cause that's allowed
+        # with docker://, but not with shub://
+        if len(self.image["namespace"]) == 0:
+            bot.error("Namespace cannot be empty for shub:// url!")
+            sys.exit(1)
```
vsoch, line 88: What would be an example image name to produce an empty namespace? I am trying (and failing) to get one that doesn't result in `library`

src/builddef.c
```diff
@@ -109,30 +109,34 @@ int main(int argc, char **argv) {
...
-                }
-
-                // Cool little feature, every key defined in def file is transposed
-                // to environment
-                envar_set(uppercase(bootdef_key), bootdef_value, 1);
-                envar_set(strjoin("SINGULARITY_DEFFILE_", uppercase(bootdef_key)), bootdef_value, 1);
             }
+
+            singularity_message(VERBOSE2, "Got bootstrap definition key/val '%s' = '%s'\n", bootdef_key, bootdef_value);
+
```
vsoch, line 121: We need to be careful here in the case that the user has put some kind of secret in the recipe (definitely not good practice or recommended, but it could happen!) Is there any way to filter / guarantee this? For the python bit I know the variables in advance so I just quiet them, but this seems like a catch all sort of deal.

src/builddef.c
```diff
@@ -109,30 +109,34 @@ int main(int argc, char **argv) {
...
+            }
+
+            if ( strcasecmp(bootdef_key, "bootstrap") == 0 ) {
+                singularity_registry_set("DRIVER", bootdef_value);
+            }
+
+            // Cool little feature, every key defined in def file is transposed
+            // to environment
+            envar_set(uppercase(bootdef_key), bootdef_value, 1);
+            envar_set(strjoin("SINGULARITY_DEFFILE_", uppercase(bootdef_key)), bootdef_value, 1);
```
vsoch, line 139: here too, since these logs get shared and moved around in various places, we need to possibly be conservative and juts report that the variable is defined. Especially if it's environment (more likely to have secrets than deffile)

tests/21-build_docker.sh
```diff
@@ -60,5 +60,55 @@ stest 1 sudo singularity build -F "$CONTAINER" docker://something_that_doesnt_ex
 stest 1 singularity exec "$CONTAINER" true
 stest 1 singularity exec "$CONTAINER" false
 
+# make sure local test does not exist, ignore errors
+sudo docker kill registry >/dev/null 2>&1
+sudo docker rm registry >/dev/null 2>&1
+
```
vsoch, line 66: I'm not sure about testing, but if there are multiple processes running on the same server we should be wearing of some kind of race condition here.

tests/21-build_docker.sh
```diff
@@ -60,5 +60,55 @@ stest 1 sudo singularity build -F "$CONTAINER" docker://something_that_doesnt_ex
...
+stest 0 sudo SINGULARITY_NOHTTPS=true singularity build -F "$CONTAINER" "$DEFFILE"
+stest 0 singularity exec "$CONTAINER" true
+stest 1 singularity exec "$CONTAINER" false
+
+cat >"$DEFFILE" <<EOF
+Bootstrap: docker
+From: my-busybox
+Registry: localhost:5000
+EOF
+
```
vsoch, line 92: If the user exports an environment and deffile which takes preference? I can see arguments for both - an environment is more a run time thing and probably should take preference, but on the other hand, if there is an envar set (and the user forgot) it would produce confusing errors. What do you think?

vsoch:
@zhcf (https://github.com/singularityware/singularity/issues/932), @hisplan @lordzappo (https://github.com/singularityware/singularity/issues/920), @hcraT @mimadrid (https://github.com/singularityware/singularity/issues/849) you all had issues parsing a URI, would you like to give this PR for a test spin and report back? Thanks everyone!

mimame:
@ilmagico awesome work!
@vsoch`  It's already working fine. #849 is fixed for me. Thanks a lot for all your support and help!

`SINGULARITY_NOHTTPS=true singularity pull docker://localhost:5000/ubuntu/ubuntu`

> WARNING: pull for Docker Hub is not guaranteed to produce the
> WARNING: same image on repeated pull. Use Singularity Registry
> WARNING: (shub://) to pull exactly equivalent images.
> Docker image path: localhost:5000/ubuntu/ubuntu:latest
> Cache folder set to /home/paradise/.singularity/docker
> Importing: base Singularity environment
> Importing: /home/paradise/.singularity/docker/sha256:b73dfea5931fd500fb7f3dade5238a034f35ae265c9b46be89af73ff176f5b5c.tar.gz
> Importing: /home/paradise/.singularity/docker/sha256:2d4d2d070fb359ed9de28e488a6d305ad72de6843be35568a817cd04d50a8164.tar.gz
> Importing: /home/paradise/.singularity/docker/sha256:b0d4ff048077ae280678597863e73666712563360008fba7ee0547f6f621ac9a.tar.gz
> Importing: /home/paradise/.singularity/docker/sha256:4ba06d7271c5e146924b198fa197704a0d68fb51c51ea4d80752e0380d33d69c.tar.gz
> Importing: /home/paradise/.singularity/docker/sha256:4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1.tar.gz
> Importing: /home/paradise/.singularity/metadata/sha256:f4ec8477ced160f71a5a480b1e47a70898a8cc75ee85ff1c548854d04b8019a3.tar.gz
> WARNING: Building container as an unprivileged user. If you run this container as root
> WARNING: it may be missing some functionality.
> Building Singularity image...
> Singularity container built: ./ubuntu.img
> Cleaning up...
> Done. Container is at: ./ubuntu.img

`singularity --version`
> 2.4-1087-allow-docker-url-empty-namespace.g4640dbe

vsoch:
@mimadrid great! Thanks for testing, and I'll close the previous issue you referenced.

libexec/bootstrap-scripts/deffile-driver-docker.sh
ilmagico: The fallback is not `set`, that would be if it was written as `${NAMESPACE:-set}` (notice `:-` instead of `+`). I don't know 100% how it works, but that expands to `set` when the variable is set, even if set to an empty string, and expand to the empty string when the variable is not set, which is what I want. I found that on [stack overflow](https://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash#13864829), don't ask too much details, it works ... ;)

Basically, I allow `NAMESPACE` to be the empty string, while for the other ones it's not allowed, and it would either give an error or use a default value.

libexec/python/defaults.py
ilmagico: Thanks for explaining the meaning of `silent`, I didn't know! That said, this change is independent of that, it simply says not to print `VARIABLE found` unless the variable was in fact found (`variable is not None`, which I guess I could simplify to `if variable:`). Before, it would print `VARIABLE found` for all variables even when **not** found, which kind of defeats the purpose. I assume it was a small typo, and I needed it fixed to debug my changes.
Notice it only prints the name of the variable, not its value, so there shouldn't be any password or token leak (and I didn't write that code, I only added the if condition).

libexec/python/shell.py
ilmagico: But python accepts it too! :) C'mon I write C and C++ code all day for a living, give me a break ... but I'll fix it.

libexec/python/shell.py
ilmagico: This is a "impossible" condition, in theory. The regex should enforce that, if it matched at all, the `repo_name` should always be present, and if not it's a bug. The user should never have to deal with this, and for developers, the stack trace is probably the most useful error message.
I'm not sure if I understood what you meant correctly though, how do you normally handle these cases in singularity?

libexec/python/shell.py
ilmagico: You mean the closing brace `}` on the same line as the last item? Sure, if that's the preferred style. Normally I write it that way to make it easier to add items to the end of the list (simply copy and paste the last line), but that's not a problem, I'll change it.

libexec/python/tests/test_core.py
ilmagico: Thanks!

libexec/python/shub/api.py
ilmagico: Great, it means the parsing code is working! In an earlier (uncommited) version, I just made namespaces optional, without distinguishing between docker:// and shub://, so this check was needed.
Once I switched to using regexes, those *should* already enforce that, for shub://, the namespace is never empty, so this should be changed into an assert instead (or whichever is the preferred way to check for impossible conditions).

libexec/python/shell.py
ilmagico: No, the mis-parsing is definitely an issue, good catch! That's the kind that if you ignore it now, you'll forget about it a year from now, then you'll spend hours (or days) debugging once you add a new feature which requires the `version` to be parsed correctly...
The fix is simple enough, I need to enforce, in the regex, that the tag cannot contain a `@`.

After locally fixing the above, I also found a couple of other interesting cases, which raise these questions:
1. Can `version` contain a `@` character? This is what happens with my code, after the fix:
```python
In [2]: parse_image_uri("docker://something.com/image:tag@2.2@3")
ERROR Could not parse image "something.com/image:tag@2.2@3"! Exiting.
An exception has occurred, use %tb to see the full traceback.

SystemExit: 1
```
2. Can `registry` contain a `@` character?
```python
In [6]: parse_image_uri("docker://user@something.com/image:tag@version")
Out[6]: 
{'namespace': '',
 'registry': 'user@something.com',
 'repo_name': 'image',
 'repo_tag': 'tag',
 'version': 'version'}
```
Should that be parsed as
```python
{'namespace': 'library',
 'registry': 'index.docker.io',
 'repo_name': 'user',
 'repo_tag': None,
 'version': 'something.com/image:tag@version'}
```
instead?
Of course, I kinda biased that example to make it look like `registry` should be as it is parsed by my code, but here's another example:
```python
In [16]: parse_image_uri("shub://ubuntu@some.version/something")
Out[16]: 
{'namespace': 'library',
 'registry': 'ubuntu@some.version',
 'repo_name': 'something',
 'repo_tag': 'latest',
 'version': None}
```

These are both easy to fix in one direction or another, just asking which way to go.

src/builddef.c
ilmagico: I agree on the need to be careful, that said, I didn't write this line of code, this was here before as well. I just unindented a whole block of code after changing the condition above. I can change this, but this was here the whole time even before my changes.

src/builddef.c
ilmagico: Same as above, I didn't write this code, just unindented the whole block. In this specific case, I'm also not sure where the information leak would be ...

libexec/bootstrap-scripts/deffile-driver-docker.sh
vsoch: ok, this is good.

tests/21-build_docker.sh
ilmagico: right, you cannot have two processes running the tests at the same time, and dare you not have another docker container in your system called "registry"! ;)
How do you propose to solve this, would it be enough to rename "registry" to, say,  "singularity_test_registry", or just let docker pick a unique name for us?

libexec/python/defaults.py
vsoch: yes this is also good! I definitely like it much better - the log will be a lot cleaner, the user doesn't actually learn much to see a variable set to None.

libexec/python/shell.py
vsoch: haha :P thank you :)

tests/21-build_docker.sh
ilmagico: Not sure, and I'm not even sure which one singularity would pick right now.

libexec/python/shell.py
vsoch: Oh, what I mean is that, if the condition fails (which it probably wouldn't) the user would see a stack trace. A developer is probably ok with this, but we've had a few issues in the past of people complaining about seeing the stack trace instead of a nice clean message. so something like:

```
try:
    assert(repo_name)
except AssertionError:
    bot.error('A repository name is required.')
    sys.exit(1)
```

libexec/python/shell.py
vsoch: Thanks!

libexec/python/shell.py
vsoch: As far as I know, I haven't seen a registry with an `@` character, so let's disregard that for now. As for the version, for singularity hub, this could be the case if the user has a branch with an `@` in it. It's likely a VERY rare case. To answer your question:

```
parse_image_uri("docker://something.com/image:tag@version")
```

should come out to be (note that I've removed the `user@` - have you seen this anywhere?

```
{'namespace': '',
 'registry': 'something.com',
 'repo_name': 'image',
 'repo_tag': 'tag',
 'version': 'version'}
```

libexec/python/shub/api.py
vsoch: okay, excellent.

src/builddef.c
vsoch: ahhh, need to pull out the Github blame @GodloveD @gmkurtzer @bauerm97 !! :P

tests/21-build_docker.sh
vsoch: I'm not actually sure if there would be a race condition - it would depend on how the CI is running it (which I'm not sure of). I think we can likely leave as is (it's working!) and if something fishy comes up, look into a possible race. That's the one great and bad thing about CI - it's easy because it's a black box that just sets things up for you, and it's hard because it's a black box that sets things up for you :)

libexec/python/defaults.py
ilmagico: While that is true, I think you're missing the point. What I was seeing was `SOME_VARIABLE found` when `SOME_VARIABLE` was not found at all (but not required, else it would exit)! That's not just unnecessary noise in the log, that's actually misleading.

vsoch: Agree. I meant to say to python, it's the same. To the user, I agree it's misleading and this is a good change.

libexec/python/shell.py
ilmagico: The `user@domain` syntax is used for e.g. HTTP Basic Auth, although docker by default uses a different authentication scheme, and itself doesn't accept URIs in that format.
I've found at least [one blog post](https://ahmet.im/blog/docker-http-basic-auth/) where somebody describe doing this, and a [related docker bug](https://github.com/moby/moby/issues/8283) (mentioned in that blog post) was closed with "Probably not so much needed", so if docker doesn't support that, I guess singularity doesn't need to support it either (yet?).

So, based on this, I'll change it to allow `@` inside version, and disallow it inside `registry`.

libexec/python/shell.py
ilmagico: Well, you might as well do
```python
if not repo_name:
    bot.error('A repository name is required')
    sys.exit(1)
```
in that case, but please hear me out on what I say below.

First of all, I'm not in charge here and will ultimately do what I'm told, but this is my opinion: an end user who knows nothing about the internals of singularity will have no use for that error message, and if anything, it will mislead him/her to think they didn't specify the repository, when maybe they did but there was a bug in singularity.
If a user encounters this, what you want from them is to report it here so it can be fixed. Maybe dumping a trace in their face is not the best way to ask them to help, but maybe something like capturing the stack trace into a file and showing a message like:
```
Internal error at file.py:123. Stack trace dumped at /tmp/singularity.bugreport.
Please report a bug at https://github.com/singularityware/singularity/issues and attach the file above.
```
would serve the purpose better? And I even argue this should be caught at the top level of the program, not locally in the function itself, so it can catch all the asserts generically.

vsoch: I can give you examples of the issues that I'm referring to, if it helps:

 - https://github.com/singularityware/singularity/issues/880
 - https://github.com/singularityware/singularity/issues/751#issuecomment-308243851

(I didn't mean to target that user, I just remembered his username and having reported a few).

I think either way the users come to post an issue, and a meaningful message (and you can do what you think is best) is preferable to a dump of python, which for someone that doesn't know python, might as well be Chinese :) 

We also want to exit with status 1 so the calling script gets the message too :)

vsoch:
@GodloveD @gmkurtzer let's get this fix in for 2.4.1, please review when you have a chance!

hcraT:
@vsoch Tried the PR and it works correctly.

vsoch:
great, thanks @hcraT

gmkurtzer:
Thanks, this looks fantastic!

PR #1121, opened 2017-10-31
Allow users to specify their own shebangs in runscript and startscript

### LINKED ISSUE #1038: Allow users to select the default shell for the shebang in %runscript
The `/singularity` script that is created from the %runscript section in the definition file has a default `#!/bin/sh` shebang instruction. It would be good to be able to select other shells, like the Bourne shell, maybe through a parameter in the definition file.
My use case, in particular, is the need to activate a conda environment inside the container by running
`source activate <env_name>`
The activate script from conda only supports bash. The workaround, at the moment, is to write something like:
```
%runscript
/bin/bash <<EOF
    source activate <env_name>
    echo "some code here"
EOF 
```
But having the possibility of selecting the default shell would simplify and streamline the entry point script.

GodloveD:

Singularity will now parse the first line of a runscript or startscript to see if the first non-whitespace characters are `#!`. If so, it will use that line as a shebang.  If not it will default to using `#!/bin/sh`.  Shell options that follow `%runscript` or `%startscript` sections are still passed to shebang or can be specified in custom shebang line. 

- Fixes #1038 

**Checkoff for all PRs:**

Attn: @singularityware-admin


granek:
I **agree** with this PR!!

ctmadison (approved):

dtrudg (changes requested):
@GodloveD - this looks good to me, but could we have a test added for this? I imagine it would be easy enough to have a test that builds from a deffile with a custom shebang, and then `singularity run` it, checking output, or value of $SHELL or similar? Just keen to have a test that captures it - and can be brought into the 3.0 integration tests etc.

GodloveD:
@dctrud good idea!

GodloveD:
merged in #1677

PR #1124, opened 2017-11-02
Allow user pull/run of docker images with non-standard $HOME / missing bind paths and no overlayfs

dtrudg:

On our cluster, and likely many others, user home dirs are not under /home but some other place (e.g. /home2). If the cluster is RHEL6.x and has no overlayfs available then `singularity run docker://...` will likely fail as docker images won't have the custom location to bind the user home dir. Binding of home dir can be turned off in singularity.conf, but this breaks many things (e.g. jupyter) which expect to have write access to homes. It also means keeping 2 singularity configurations - 1 for docker images that can't bind mount /home2 and 1 that can bind /home2 to customized images where it is present.

The same issue occurs for other cluster fs such as `/project` which we need to set as `bind path` entries, but do not exist in docker containers.

This PR adds an option to `singularity.conf` called `docker inject binds`. It will create an extra layer tar that injects the home dir and `bind path` dirs into the extracted image so that binding works as expected without overlayfs on unmodified docker containers when a user performs `singularity run` or `singularity pull`.

Paths are injected by adding a hidden file `<bind path>/.singularity-bind` to the tar. This file should not be visible after the bind is made, since the bind will be mad onto `<bind path>`. The hidden file is created so we mark a directory as being a bind injection, rather than belonging to the container if the image is inspected by other means.

`bind path` entries corresponding to files, or directories not-existing on the host are not injected.

** Demo in use **

```bash
VERBOSE3 Adding home bind point to bind tar
WARNING Not injecting bind path /etc/localtime into container - not a directory on host
WARNING Not injecting bind path /etc/hosts into container - not a directory on host
VERBOSE3 Adding bind point /bob to bind tar
WARNING Not injecting bind path /lala into container - not a directory on host
VERBOSE2 Tar file with injected bind points: /home/dave/.singularity/bind/sha256:f6be9f4f6905406c1e7fd6031ee3104d25ad6a31d10d5e9192e7abf7a21e519a.tar.gz
VERBOSE3 Writing Docker layers files to /tmp/.singularity-layers.VTJ1piqm
VERBOSE2 Writing file /tmp/.singularity-layers.VTJ1piqm with mode w.
VERBOSE2 Writing file /tmp/.singularity-layers.VTJ1piqm with mode a.
VERBOSE2 Writing file /tmp/.singularity-layers.VTJ1piqm with mode a.
Creating container runtime...
Exploding layer: sha256:ae79f251470513c2a0ec750117a81f2d58a50727901ca416efecf297b8a03913.tar.gz
Exploding layer: sha256:c59d01a7e4caf1aba785eb33192fec3f96e4ab01975962bcec10f4989a6edcc6.tar.gz
Exploding layer: sha256:41ba73a9054d231e1f555c40a74762276900cc6487f5c6cf13b89c7606635c67.tar.gz
Exploding layer: sha256:f1bbfd495cc1112b503350686641ee4fa2cea8ccd13fb8a8a302c81dae61d418.tar.gz
Exploding layer: sha256:0c346f7223e24b517358f52c4a3f5f9af1c86e5ddeaee5659c8889846e46d1e2.tar.gz
Exploding layer: sha256:f6be9f4f6905406c1e7fd6031ee3104d25ad6a31d10d5e9192e7abf7a21e519a.tar.gz
Exploding layer: sha256:3167532bac2cbd24e964913314dcc16c02f26fcde198e0bc92ad8a92611a5684.tar.gz
```

**Checkoff for all PRs:**

Attn: @singularityware-admin


vsoch:
I like the idea that a cluster could specify a different path, but a few comments.

- Overall, I would say it should be more generalizable. There are many different clusters with many unique features, and some approach that allows for custom functionality for (a general folder) would have more use cases. Why just a home folder?
- why not just bind home2 to home?  `singularity run --bind /home2:/home <container>` or `singularity run --home /home2`? Oh, I'm guessing you don't have overlayfs so it can't be created on the fly?
- I don't really like the idea of an environment variable, but a config setting seems reasonable, given that a cluster is likely to have this globally. There could be a list of "allowed" directories to create if needed, those are tried first, and then spit out an error.
- there definitely should be checks to make sure the directory created is the one intended, where it's intended. It would be a root creation. I don't totally know the implications of that, but that alone makes me nervous.
- Did you try running with various `--contain` to just skip mounting?

libexec/cli/build.exec
```diff
@@ -255,6 +255,23 @@ case $SINGULARITY_BUILDDEF in
...
+        if [ ! -z "${SINGULARITY_CREATE_HOME:-}" ]; then
+
+            if [ "$(id -u)" == "0" ]; then
+               message ERROR "Cannot use SINGULARITY_CREATE_HOME as root\n"
+               ABORT 255
+            fi
+
+            ROOTFS_HOME="${SINGULARITY_ROOTFS}${HOME}"
+            if [ ! -d "$ROOTFS_HOME" ]; then
+                message 1 "SINGULARITY_CREATE_HOME set. Creating missing $ROOTFS_HOME\n"
```
vsoch: Another idea is that it could be created on the fly as one of the layers (or in the metadata tar.gz) by the python, at least that way we have assurance it's extracted with rootfs and we don't have custom mkdir -p outside of that.

dtrudg: That's a nice idea. I'll have to look into that. This was the quickest place I found to add it to satisfy a need to do a demo on Friday.

vsoch: If others think it's a good idea I can offer to do it - I spent a lot of time figuring out in memory tar creation so it wouldn't take super long. But it does bring up this larger (kind of cool?) question - can we give some greater options, at runtime, to modify what is being extracted? Without compromising safety? Is that even a good idea given that a container (is supposed to be) immutable?

dtrudg:
Hi @vsoch 

> Overall, I would say it should be more generalizable. There are many different clusters with many unique features, and some approach that allows for custom functionality for (a general folder) would have more use cases. Why just a home folder?

Agreed - we need more than just home. Put here as WIP for comments on whether this is worth expanding/generalizing.

> why not just bind home2 to home? singularity run --bind /home2:/home <container> or singularity run --home /home2? Oh, I'm guessing you don't have overlayfs so it can't be created on the fly?

Binding /home2 into /home exposes everyone's home dir there. Perms should stop any nastiness but I'd rather not do that. I've messed around with other things but didn't come up with something that was nice ans simple for users. Doing the -H /home2/dave:/home/dave works, but is confusing for users to have things in different locations on the cluster and in a container. We want to make it as transparent as possible. We plan to wrap containers as environment modules - which kinda depends on everything being in the same place for it to be workable.

```
21:38 $ singularity run -H /home2/dave:/home/dave docker://ubuntu:latest
Docker image path: index.docker.io/library/ubuntu:latest
Creating container runtime...
dave@spongebob:~
02:39 AM $
```


>    I don't really like the idea of an environment variable, but a config setting seems reasonable, given that a cluster is likely to have this globally. There could be a list of "allowed" directories to create if needed, those are tried first, and then spit out an error.
    there definitely should be checks to make sure the directory created is the one intended, where it's intended. It would be a root creation. I don't totally know the implications of that, but that alone makes me nervous.

Agreed. But - this stuff here only runs as the user - root isn't creating a dir here.

>    Did you try running with various --contain to just skip mounting?
Yup tried that:
```
21:36 $ singularity run -c docker://ubuntu:latest
Docker image path: index.docker.io/library/ubuntu:latest
Creating container runtime...
ERROR  : Base home directory does not exist within the container: /home2
ABORT  : Retval = 255
#containall
21:37 $ singularity run -C docker://ubuntu:latest
Docker image path: index.docker.io/library/ubuntu:latest
Creating container runtime...
ERROR  : Base home directory does not exist within the container: /home2
ABORT  : Retval = 255
```

libexec/cli/build.exec
dtrudg: Nick Eggleston on the google groups list would probably like this also - as he's hitting the same issue on 6.x. I'd guess it's quite widespread that RHEL 6.x is in use with a non standard home.

vsoch:
I think it also suggests we need a larger thing - I came up with this need when I was first developing sregistry (in a singularity container, not using docker). For HPC / experienced users, it's pretty trivial to know how to do custom binds and arguments. But for a user they just want to "run" the thing. What we might think about is some kind of external runscript that can live with the container (either creator by the creator of the container, but more likely where singularity is installed) that can add a snippet of code before a run, for example, and do something like:

```
if /home2/$USER exists:
  - mount /home2/$USER to /home

if (some special logic just for my cluster)
  --bind /scratch:/tmp
```
It comes down to being a lot like the config file, but potentially only being applied to a subset of containers, and also having some runtime variable logic.

libexec/cli/build.exec
vsoch: just curious, why is there /home and /home2?

dtrudg: On our systems /home is a local per-node location for some administrative accounts. Then there was /home1 where users had space, and it was replaced with /home2 when another storage system came along. On Linux environment at a previous university I recall we had people split by department like /home-dcs for CS folks /home-eng for the engineers. Not under /home is relatively common I think.

dtrudg:
To satisfy local need I'm going to rewrite this into the python that creates the metadata tar (as you suggested), enabled by an option `docker inject bind paths = yes` in `singularity.conf` that if set will inject the homedir and all `bind path = ` entries from singularity.conf as empty paths. That will solve all of our local issues fairly neatly. Code will probably be up here tomorrow.

Per-container external runscript is interesting - but I'm currently doing that kinda stuff by having a modulefile on the cluster with a wrapper or shell alias that contains the magic. E.g. we:

```bash
module add blender
blender
```

... but 'blender' is really running `singularity exec --nv -B /a/b/c:/foo ...`

vsoch:
before you do that, I would check with @gmkurtzer and @GodloveD about if this would be a good use of your time - everyone is avoiding the subject but (I think) the plan is for python to go away entirely. I am not clear about what will replace it, but if this is the case it doesn't make sense to further develop the modules.

dtrudg:
Hi @vsoch - it's a good use of my time regardless (locally), as I do need to inject some `bind path`s to make some things usable here on RHEL6.9 without mass rebuilding of docker containers to just add 2 empty folders. If it's not picked up because the python is a dead end then that's fine - just think it looks like the neatest place to do it right now after you pointed out the metadata tar.

vsoch:
okay sounds good, go for it! I would maybe separate it from the metadata tar  - possibly another cache location under `.singularity.d`

GodloveD:
> everyone is avoiding the subject but (I think) the plan is for python to go away entirely. 

Is this the case?  I'm not aware of any such talk.  Though adding python to the core is definitely a non-starter.

vsoch:
Which means no new features via python should be developed, is how I infer that.

dtrudg:
> okay sounds good, go for it! I would maybe separate it from the metadata tar - possibly another cache location under .singularity.d

Yeah - I meant to say that ;-)

dtrudg:
>Though adding python to the core is definitely a non-starter.

@GodloveD - for reference, what exactly is the core in this sense?

GodloveD:
I'm not really the person to talk to about this.  @gmkurtzer would know better.

vsoch:
I think it means everything that isn't python.

dtrudg:
5a0d1d4 has a complete implementation of home and `bind path` injections for consideration - via an additional tar layer. It's configured via new option `docker inject binds` in `singularity.conf`.

DrDaveD:
An implementation of issue #1207 might make this PR unnecessary.

dtrudg:
@DrDaveD - yes, if #1207 implementation allows non-overlay binds to arbitrary dirs in container this should be closed. Per comment on #1207 was implemented here as an essential need for our RHEL6.9 cluster, and in a spot I though was safe from squashfs / SIF type changes to images and mounting code.

dtrudg:
@vsoch - any thoughts on whether this is worth tidying up any further, or is custom bind without overlay directly with docker images not much of a need?

vsoch:
I want to give feedback - but I'm having trouble understanding the use case. You are saying that this is a custom bind that is only relevant to "on the fly" created docker images? Or a docker image that is pulled first (and how is that different from any other singularity image that isn't derived from Docker?)

dtrudg:
It's applying in the docker layer extraction code, so would apply to a pull or direct run from docker. Use case is this:

- We run RHEL6.9 which doesn't support overlayfs so can't bind from host dirs into container if dir not present in the container.
- Our user home directories are under /home2 which does not exist in docker containers (unless we made them)
- Running docker containers directly fails, as $HOME can't be bound in since /home2 isn't in the container
- Using a contained $HOME, no bind, makes the containers of little use - can't access data on cluster sensibly.
- Also we have most data at /project /work /archive - these won't bind without the change here, as those dirs don't exist in container
- Major direct use of singularity by users (not including things we've hidden as modules or in our web portal) has turned out to be users trying random dockerhub containers - we can't really rebuild all for them, using def files to adding the /home2 etc we need - takes time/resources and is a waste for 'trying out' kinda work
- Users aren't interested in building a singularity.img with the dirs themselves - just stop using singularity and ask why we don't support docker.

I wrote this as creation of an additional bind point layer in the docker extraction code as:

- Don't want it to affect anything other than docker pulls - native .simg would be curated in our registry with things added.
- Wasn't sure about changes with squashfs and SIF - so wanted to try and do this in a place where I didn't need to care about impacts there - would be stable as we update Singularity here.

libexec/python/sutils.py
```diff
@@ -528,3 +530,32 @@ def write_singularity_infos(base_dir,
                                  extension)
     write_file(output_file, content)
     return output_file
+
+
+def get_singularity_conf_value(key):
```
vsoch, line 535: Is the absolute only way to do this to read in the actual config file and then search every line for the value of interest? My intuition is that as long as there aren't many values, and primary interaction of the python client with everything else is via the environment, this should be more controlled (e.g., if a value is needed from the config it should be exported in the environment, defined in defaults.py, and then imported from there.

libexec/python/docker/tasks.py
```diff
@@ -261,3 +262,48 @@ def get_configs(manifest, keys, delim=None):
...
+    '''create_bind_tar creates a tar containing empty directories for the
+    user's $HOME and any system-wide "bind path" entries in singularity.conf
+    This allows non-overlay systems to bind these into docker containers
+    when those folders do not already exist in the container.
+    '''
+    tar_file = None
+    files = []
+
+    # Deal with home directory
+    home_bind = os.path.join(os.path.expanduser("~"),
```
vsoch, line 278: I think this might also be an issue - if you look at how we get home in defaults.py (https://github.com/singularityware/singularity/blob/master/libexec/python/defaults.py#L107) this was done per a discussion with @gmkurtzer to not use standard `$HOME` for some security reason I can't remember (this was more than a year ago easily!). Regardless of how it's done, this also enforces that using the feature always adds this user home, what if this isn't desired?

libexec/python/docker/tasks.py
```diff
@@ -261,3 +262,48 @@ def get_configs(manifest, keys, delim=None):
...
+                             '.singularity-bind')
+    home_bind = home_bind.lstrip('/')
+    bot.verbose3('Adding home bind point to bind tar')
+    template = get_template('tarinfo')
+    template['name'] = home_bind
+    template['content'] = ("This is a bind point injected by singularity with "
```
vsoch, line 284: the content portion of the template is intended for content of files - so here you would be creating a file with name matched to `home_bind` and texty content (and not a directory). Is this what is wanted? I would think a directory would be wanted? I think you would use the equivalent template, but do an extra change to the tarinfo to make it directory type, e.g.

```
tarinfo.type = tarfile.DIRTYPE
tar.addfile(t)
```
I think adding as a file would work, but it doesn't seem exactly like what is wanted.

vsoch: So probably a check would be needed here for having content, period https://github.com/dctrud/singularity/blob/master/libexec/python/sutils.py#L180 and then another here for a type https://github.com/dctrud/singularity/blob/master/libexec/python/sutils.py#L169

vsoch:
hey @dctrud ! Okay I remember this clearly now. I made some detail comments and here will add some more high level discussion.

## Is it broadly useful?
One could argue that for a very specific use case of an operating system, and specific directory, it doesn't seem broadly useful. *But* it seems like a deal breaker given a particular setup, and if entire clusters are prevented from using Singularity because of it, it cannot be ignored. On the other hand, when you think about it this as a general solution for creating content on the fly without needing overlayfs, it does sound very useful. I can imagine something during extraction like:

 - 1. if no overlayfs, check for permission to create via tar (from config in environment)
 - 2. if given, create directories only also only defined some avenue by root user (probably also config)
 - 3 . create folders for metadata tar, and add (I think they should be added first, in case content is later extracted over). 

## Is it safe?
But then this is sort of the bigger issue - we are creating content on the fly, to be extracted with a root permission, only the user file system. This has me a bit concerned - because any manipulation could give the user total write access to do whatever. I want to ping @cclerget who tends to be a badass at this security stuff. @cclerget - where are the holes with this kind of setup?

## Is there an existing workaround?
If it's the case that #1207 works, and is tested (and also works) for this issue, we might kill two birds with one stone. @DrDaveD is there significant issue with that PR getting merged?

If it's the case that this is safe to do, and that it's not solved by #1207, then I think it could be valuable for clusters that have this particular issue. It also makes me wonder if there is some larger need to allow a cluster to implement some custom procedures to edit images at pull / assembly (after build). For example, imagine some kind of recipe in the config file to be applied to every image, pertinent to permissions, addition and deletion? It would bring a sort of overlayfs support without having it, and allow for clusters to implement their custom needs (whatever they may be).

vsoch:
Just resolved the CONTRIBUTORS.md conflict, not sure why it didn't work the first time but looks ok now!

libexec/python/docker/tasks.py
dtrudg: Hi - thanks for the thoughts on this PR.

I could use that code you referenced or expanduser("~username") here so it never pulls from $HOME, but uses the pwd. Doing "~username" enforces the lookup and ignores $HOME too.

I did actually want to use $HOME for something here.... but it's not generally relevant and should be stripped for consistency.

R.E. always adding the home - I'm trying to keep it simple here - say a user request binding an alternate home at the singularity command line then it'll just be adding a dir here that shouldn't have any impact.

libexec/python/docker/tasks.py
dtrudg: It's intentional that I'm creating a file here at . If I need bind point `/home2/dtrudgian` I'm creating the file `/home2/dtrudgian/.singularity-bind` with content that explains what that is. I wanted some kind of marker so that if I extract/inspect the image I can see the paths that were added/acted on by the inection code. When the container is run and bind is made at `/home2/` dtrudgian that file is masked as my real home dir is mounted over it.

dtrudg:
Hi @vsoch - thanks for your comments on this. I've replied to the code review bits above, but for the more general thoughts:

>I can imagine something during extraction like:
if no overlayfs, check for permission to create via tar (from config in environment)

Right now it's not checking if overlayfs isn't present - but there is a `docker inject binds` flag in singularity.conf

>if given, create directories only also only defined some avenue by root user (probably also config)

In this case we are only creating directories defined by the root user (though home is computed) - since we are pulling bind points from `singularity.conf`

>3 . create folders for metadata tar, and add (I think they should be added first, in case content is later extracted over).

I'd argue it does need to be done last to guarantee injected bind points really will be there. Consider if we do it first we could end up with this:

- Inject a bind point for `/work`
- Docker layer 3 extracted, creates `/work' for something remporary in the container build (already present so no effect).
- Docker layer 6 extracted, removes '/work' (via whiteout)
- Try to run container - `/work` no longer there - cannot bind to it.

>Is it safe?
But then this is sort of the bigger issue - we are creating content on the fly, to be extracted with a root permission, only the user file system. This has me a bit concerned - because any manipulation could give the user total write access to do whatever. I want to ping @cclerget who tends to be a badass at this security stuff. @cclerget - where are the holes with this kind of setup?

So with this PR we are at most adding `bind-dir/.singularity-bind` files to the image, where the bind-dir comes from the list of bind-points in the admin-maintained singularity.conf. There is no other content that is added. Since at runtime singularity is going to bind things from the host onto these dirs, the content we do put there isn't actually seen by the container at runtime - unless we take the image and run it with a different binding config. In that case you'd just see these few extra `.singularity-bind` files where dirs were injected.

AFIAK If it's possible to do anything malicious here then it's also possible to do it by crafting a docker-layer to inject similar content - and getting someone to pull a docker container with that layer in it.

>Is there an existing workaround?
If it's the case that #1207 works, and is tested (and also works) for this issue, we might kill two birds with one stone. @DrDaveD is there significant issue with that PR getting merged?

A fix for issue #1207 could present a solution, yes - but I'm not aware of a PR yet? 

> It also makes me wonder if there is some larger need to allow a cluster to implement some custom procedures to edit images at pull / assembly (after build).

I thought about that, but being able to add arbitrary things to a docker pull moves us further from replicating reproducibly what a 'docker run' would do/see for a given container. This bind point case is a 
specific thing that I'm guessing will hinder Singularity usage for people running RHEL6.x clusters.

cclerget:
@vsoch I don't see any concerns about security, as @dctrud mentionned, that concerned bind path, and homedir seems not sensitive in this case.

libexec/python/sutils.py
dtrudg: Yeah, maybe. I did it this way as I didn't at the time want to touch the C code and stick more stuff out into the environment. I'm a little confused what the preferred thing is to do to get config settings in different parts of the code base? Are the C libexec binaries always taking from environment, or sometimes reading from conf file? Wasn't there some discussion of this on the C side in another PR where .@GodloveD  and you had both implemented something to read config - might be going crazy and imagining though.

DrDaveD:
@vsoch Yes, as @dctrud said, #1207 so far is only a design described in an issue, not yet a PR.  I was hoping to inspire someone else to implement it since at this point I have higher priority things I have to spend my time on.

dtrudg:
I'm going to close this PR as I no longer work at the institution where I wrote it, and it hasn't been picked up by a colleague there. To the best of my knowledge there haven't been any explicit requests for this, and it's a bit of a hack - injecting bind points only for docker containers, and using tar files.

If anyone is watching this and sees a need, please ping me here.

vsoch:
@dctrud it was a really cool idea / implementation, for what it's worth! :)

Issue #1103, opened 2017-10-27
Duplicate bootstrap definition key found: '#UpdateURL'


DirkdeDraak:
## Version of Singularity:

2.4

### Expected behavior

Line starts with a '#', so I would think it should be seen as a comment.

### Actual behavior

It shows this error:
Duplicate bootstrap definition key found: '#UpdateURL'

### Steps to reproduce behavior

Have a definition file with 2x #UpdateURL: lines
```
BootStrap: yum
OSVersion: 7
#MirrorURL: http://mirror.centos.org/centos-%{OSVERSION}/%{OSVERSION}/os/$basearch/
MirrorURL: http://mirror.centos.org/centos-7/7.4.1708/os/x86_64/
Include: yum

#UpdateURL: http://mirror.centos.org/centos-%{OSVERSION}/%{OSVERSION}/updates/$basearch/
#UpdateURL: http://mirror.centos.org/centos-7/7.4.1708/updates/x86_64/
```
Having one commented out and the other not, it works fine.
Having both uncommented also gives the 'duplicate key found' error, which is correct.

ilmagico:
This one looks easy enough, I'll take a look at it.

PR #1127, opened 2017-11-02
Fix commented lines parsing in recipe file

ilmagico:

Implement proper parsing of comments in recipe file. Remove comments before splitting line into key/value pair, and use `strchr()` to find comment character`#` instead of `strtok`, as the latter ignores a '#' as the first character of the string.

- Ref: #1103 

**Checkoff for all PRs:**

    - I have added myself in another pull request, which is not merged yet. This change is small enough that it wouldn't matter much anyway. Didn't do it to avoid conflicts later. Edit: other PR is now merged.

Attn: @singularityware-admin


vsoch:
what happens if a user has an interpreter in a runscript? eg:

```
%runscript
#!/bin/sh

echo "Hallo!"
```

ilmagico:
@vsoch well, they're bad boys and gonna have a bad time! :-p
Didn't realise the same code was used to parse those sections too, I thought I only modified the one that parses `Key: Value` lines. I'll have another look at it and come back at you.
By the way, all tests passed, isn't there a test with a runscript like that? (if not let's add one)

vsoch:
haha :)

+1 to add the test! We actually haven't tested the runscript section because it was assumed to be a few lines of shell commands, and then when parsing we would insert the interpreter. We actually have in another PR right now to allow the user to specify the interpreter (meaning the #! would be there):

https://github.com/singularityware/singularity/pull/1121

@GodloveD do you have thoughts on how to get these to play nicely together?

src/builddef.c
```diff
@@ -101,6 +101,8 @@ int main(int argc, char **argv) {
     while ( fgets(line, MAX_LINE_LEN, bootdef_fp) ) {
         char *bootdef_key;
 
+        chomp_comments(line);
+
         if ( line[0] == '%' ) { // We hit a section, stop parsing for keyword tags
             break;
```
ilmagico: @vsoch I'm not completely familiar with this code but this check seems to imply that sections like `%runscript` are not parsed by this code, so my changes shouldn't affect them.
Do you still think this could create problems with `#!/bin/bash` inside `%runscript` or similar? Keep in mind that I haven't had time to test this yet.

GodloveD:
I would need to test it to be sure.  But I am thinking that this would not affect parsing of sections.  Check out [this line](https://github.com/ilmagico/singularity/blob/1f7e6a9552f38abb134dbbf3ef0971ab2305a2ca/src/builddef.c#L106).  @ilmagico can you test and see if a runscript get's stripped of comments with your PR?  I think that comments will end up in the runscript.

ilmagico:
@GodloveD yes, that's the same thing I noticed in my comment above. Still, I haven't had time to test it (maybe I'll have time tomorrow).

gmkurtzer:
@ilmagico This seems to be conflicting with your other PR (and your other PR is causing some problems when multiple new lines are present in the recipe header). Can you take a look in short order please?

Thanks!

ilmagico:
Yes, I know about the conflict, in fact, I originally wrote this on top of the other one, but then cherry-picked it onto 'development' to make a PR... so that should be easy.

Didn't know my other PR (for reference, we're talking about #1116) has problems with multiple empty lines, I'll double check.

ilmagico:
Ok, I solved the conflict and fixed the issue with empty lines (they are simply ignored now).

Also, I noticed the current parser is very liberal in what it accepts, even before my changes. Things like:
```
:
:::
```
are accepted and ignored, while this
```
: :
```
is parsed as key and value both empty, and is added to the dictionary as such, which results in setting `SINGULARITY_DEFFILE_` to the empty string.
In addition, these other funny looking lines are parsed "correctly":
```
:Key1:Value1:  # value is 'Value1:' including the :
:::Key2:Value2
:Key3 #empty value
```
I'm not sure if that's a bug or a feature ...

gmkurtzer:
Awesome, thank you!

PR #1133, opened 2017-11-02
Exposing labels, bin, lib, environment for SCI-F

vsoch:

This is a quick PR to add environment variables to SCIF, meaning that we expose each of an app's environment file, labels, and runscript (for use by other apps). For example, I can build this recipe:

```
Bootstrap: docker
From: ubuntu

%environment
    ALLHERE=Idontknow
    export ALLHERE

%appenv bar
    BARISHERE=yes
    export BARISHERE

%apprun foo    
    exec echo "RUNNING FOO"

%appenv foo
    . "${APPENV_bar}"
    FOOISHERE=yes
    export FOOISHERE
```

and specifically, take note of the line that sources ` . "${APPENV_bar}"` in foo's environment, meaning that bar's environment will be sourced when foo's is. This is useful so that environment dependencies (that are shared between two apps but not all apps) can easily be handled, and the user doesn't have to know the nitty gritty path to where the environment is, etc. We can see it during runtime:

```
singularity exec --app foo foo.simg env | grep ISHERE
BARISHERE=yes
FOOISHERE=yes
```

and we can look at all of the environment variables that are automatically made available to us:

```
singularity exec --app foo foo.simg env | grep APP
```

I'll walk through the different use cases. Data and root are obviously useful when we want to get the location of a file or data object for an app. Instead of needing a hard coded path, we use these variables.

```
APPDATA_foo=/scif/data/foo
APPDATA_bar=/scif/data/bar
APPROOT_foo=/scif/apps/foo
APPROOT_bar=/scif/apps/bar
```
eg:

```
%apprun lolcow
lolcat | cat ${APPDATA_bar}/message.txt
```

The direct paths to the bin is an easy shorthand for adding things to the path. 

```
APPBIN_bar=/scif/apps/bar/bin
APPBIN_foo=/scif/apps/foo/bin
APPLIB_bar=/scif/apps/bar/lib
APPLIB_foo=/scif/apps/foo/lib
```

For example, let's say I have an app that does nothing other than combine a bunch of other apps, in some unique way, adding things to the path

```
%apprun lol-animal
 PATH=${PATH}:${APPBIN_cow}
 exec runcow
```

We've already seen the environment example, and there are equivalent paths for labels metadata folder, and the runscript. 

```
APPMETA_foo=/scif/apps/foo/scif
APPMETA_bar=/scif/apps/bar/scif
APPENV_bar=/scif/apps/bar/scif/env/90-environment.sh
APPENV_foo=/scif/apps/foo/scif/env/90-environment.sh
APPLABELS_foo=/scif/apps/foo/scif/labels.json
APPRUN_foo=/scif/apps/foo/scif/runscript
```
Note how there is only one labels, or run, and this is because it won't be set if the app doesn't have any. The runscript is pretty cool, because you can run apps from other apps.

```
%apprun lol
   /bin/bash ${APPRUN_cat
   /bin/bash ${APPRUN_cow}
```

and the (active app) variables are of course still relevant!

```
SINGULARITY_APPOUTPUT=/scif/data/foo/output
SINGULARITY_APPS=/scif/apps
SINGULARITY_APPDATA=/scif/data/foo
SINGULARITY_APPINPUT=/scif/data/foo/input
SINGULARITY_APPROOT=/scif/apps/foo
SINGULARITY_APPNAME=foo
```

This was an issue on the Google Group.

Attn: @singularityware-admin


vsoch:
Note that when this is approved, I have also updated the docs for apps to describe the updated variable exports https://github.com/singularityware/singularityware.github.io/pull/152

dtrudg:
👍 This is nice. When I saw the email on the list that inspired it I was thinking - yes, I'd probably want to do that at some point.

vsoch:
+1! You can do a bunch of tricks with different apps :) I'm pooped today but I'll come up with some cooler examples.

gmkurtzer:
Looks excellent!

PR #1137, opened 2017-11-06
Improvements to Debian/Ubuntu examples

0xaf1f:

* Remove `--force-yes`. It's not recommended to be used unless
really necessary. See [apt-get(8)](https://manpages.debian.org/stretch/apt/apt-get.8.en.html):

>     Force yes; this is a dangerous option that will cause apt to
>     continue without prompting if it is doing something potentially
>     harmful. It should not be used except in very special situations.
>     Using force-yes can potentially destroy your system! Configuration
>     Item: APT::Get::force-yes. This is deprecated and replaced by
>     --allow-downgrades, --allow-remove-essential,
>     --allow-change-held-packages in 1.1.

 The file that had it (ubuntu.def) probably needed it because it didn't
 first call `apt-get update`. That change was made.

* Use `apt-get clean` at the end to remove downloaded package archives,
  saving space in the container.

- #1160
- #1161

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -14,6 +14,7 @@
 
 #Contributors:
 
+    - Afif Elghraoui <afif.elghraoui@nih.gov>
     - Amanda Duffy <aduffy@lenovo.com>
     - Ángel Bejarano <abejarano@ontropos.com>
     - Bernard Li <bernardli@lbl.gov>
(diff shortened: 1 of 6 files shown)


0xaf1f:
The difference in image size after using `apt-get clean` is actually quite significant:

```
-rwxr-xr-x 1 root        root         63M Nov 10 15:47 ubuntu-clean.simg
-rwxr-xr-x 1 root        root        117M Nov 10 15:52 ubuntu-noclean.simg
```
This is using `examples/ubuntu/Singularity` with and without `apt-get clean`.

gmkurtzer:
Looks great, thanks for the cleanups!

dtrudg:
#1376 is a rebase into development-2.x

Issue #1156, opened 2017-11-13
--nv option does not export necessary OpenCL ICD files


Caian:
## Version of Singularity:

2.4-dist

### Expected behavior

```
$ singularity shell --nv -B /etc/OpenCL/vendors/ ubuntu.img 
Singularity: Invoking an interactive shell within container...

Singularity ubuntu.img:~/singularity> clinfo 
clinfo: /.singularity.d/libs/libOpenCL.so.1: no version information available (required by clinfo)
Number of platforms:				 1
  Platform Profile:				 FULL_PROFILE
  Platform Version:				 OpenCL 1.2 CUDA 8.0.0
  Platform Name:				 NVIDIA CUDA
  Platform Vendor:				 NVIDIA Corporation
  Platform Extensions:				 cl_khr_global_int32_base_atomics cl_khr_global_int32_extended_atomics cl_khr_local_int32_base_atomics cl_khr_local_int32_extended_atomics cl_khr_fp64 cl_khr_byte_addressable_store cl_khr_icd cl_khr_gl_sharing cl_nv_compiler_options cl_nv_device_attribute_query cl_nv_pragma_unroll cl_nv_copy_opts


  Platform Name:				 NVIDIA CUDA
Number of devices:				 1
  Device Type:					 CL_DEVICE_TYPE_GPU
  Device ID:					 4318
  Max compute units:				 28
  Max work items dimensions:			 3
    Max work items[0]:				 1024
    Max work items[1]:				 1024
    Max work items[2]:				 64
  Max work group size:				 1024
  Preferred vector width char:			 1
  Preferred vector width short:			 1
  Preferred vector width int:			 1
  Preferred vector width long:			 1
  Preferred vector width float:			 1
  Preferred vector width double:		 1
  Native vector width char:			 1
  Native vector width short:			 1
  Native vector width int:			 1
  Native vector width long:			 1
  Native vector width float:			 1
  Native vector width double:			 1
  Max clock frequency:				 1531Mhz
  Address bits:					 64
  Max memory allocation:			 3195420672
  Image support:				 Yes
  Max number of images read arguments:		 256
  Max number of images write arguments:		 16
  Max image 2D width:				 16384
  Max image 2D height:				 32768
  Max image 3D width:				 16384
  Max image 3D height:				 16384
  Max image 3D depth:				 16384
  Max samplers within kernel:			 32
  Max size of kernel argument:			 4352
  Alignment (bits) of base address:		 4096
[...]

PR #1157, opened 2017-11-13
Bind /etc/OpenCL/vendors with --nv

dtrudg:

Per #1156 singularity should bind the OpenCL `/etc/OpenCL/vendors` profiles dir into the container when binding nvidia libs with `--nv`, so that the OpenCL config seen inside the container matches the NVIDIA libs that are bound from outside.

- Ref: #1156

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ and changes prior to that are (unfortunately) done retrospectively. Critical ite
  - Fix parsing of registry (including port), namespace, tags, and version
  - Add "$@" to any CMD/ENTRYPOINT found when building from Docker
  - Added sqaushfs-tools as a dependency for building deb files
+ - Bind /etc/OpenCL/vendors when using --nv option
 
 ## [v2.4](https://github.com/singularityware/singularity/tree/v2.4) (2017-10-02)
 [Full Changelog](https://github.com/singularityware/singularity/compare/2.3.2...2.4)
--- a/libexec/cli/action_argparser.sh
+++ b/libexec/cli/action_argparser.sh
@@ -150,6 +150,16 @@ while true; do
             else
                 message WARN "Could not find the Nvidia SMI binary to bind into container\n"
             fi
+            if [ -d "/etc/OpenCL/vendors" ]; then
+                if [ -n "${SINGULARITY_BINDPATH:-}" ]; then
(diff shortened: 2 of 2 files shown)


dtrudg:
@Caian - would you be able to test this to see if it fixes your issue on #1156?

GodloveD:
@dctrud I'm not sure this is actually needed.  Are OpenCL versions strictly tied to the NVIDIA driver version on a particular host?  If not shouldn't it just be installed into the container?

dtrudg:
Per discussion on slack, the `/etc/OpenCL/vendors/nvidia.icd` file is provided by the specific NVIDIA driver install on the host. It contains the name of the OpenCL ICD (vendor implementation), in this case `libnvidia-opencl.so.1`. The .so is picked up by the --nv option, but the /etc/OpenCL/vendors/*.icd files provide a discovery mechanism to find available implementations, as used by the clinfo command that fails in the #1156 issue report.

I'd argue for bringing this in with --nv, as it is provided by the host NVIDIA driver install, and it's probably a long wait otherwise for a generic `--opencl` option that would fully support generic OpenCL implementations (AMD/intel/NVIDIA) that woudl have their own driver/library binding requirements.

Caian:
Hello @dctrud. I have tested and confirm that it solves the problem for the NVIDIA case. As you pointed out, other vendors have their own requirements, and that's what I found when trying to apply the same method to make Intel OpenCL work from inside the container.  I have updated my original issue with additional information.

I think that a generic `--opencl` option is more complicated to implement, giving that vendors may provide additional shared objects that are required by the one pointed by the ICD file.

GodloveD:
Here is what I suggested on the Slack.  I'm also writing it here since this is the discussion of record:

As I see it, the objective of the `--nv` option is to grab anything that is detected by [nvidia-container-cli](https://github.com/NVIDIA/libnvidia-container).  See my proposed PR about [using nvidia-container-cli to locate libs and bins on the host system](https://github.com/singularityware/singularity/pull/1082).

I think we should check to see if nvidia-container-cli finds the NVIDIA-related OpenCL files.  If it does, they will be added by default once that PR is merged.  If it doesn't than NVIDIA doesn't consider those files to be part of the NVIDIA driver and so they have no business being bind mounted with the `--nv` option.

PR #1173, opened 2017-11-25
Fix suid detection and instance boot

cclerget:

- Fix SUID detection
- Add missing UTS option for instance.start
- Fix privblock tests and getuid for TARGET_UID/TARGET_GID

Changelog has already been updated for boot/start instance feature

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/src/lib/runtime/files/hostname/hostname.c
+++ b/src/lib/runtime/files/hostname/hostname.c
@@ -1,0 +1,70 @@
+/* 
+ * Copyright (c) 2017, SingularityWare, LLC. All rights reserved.
+ * 
+ * This software is licensed under a 3-clause BSD license.  Please
+ * consult LICENSE file distributed with the sources of this project regarding
+ * your rights to use or distribute this software.
+ * 
+ */
+
+#include <errno.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <limits.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <grp.h>
+#include <pwd.h>
+
+#include "config.h"
+#include "util/file.h"
+#include "util/util.h"
+#include "util/registry.h"
+#include "util/config_parser.h"
+#include "util/message.h"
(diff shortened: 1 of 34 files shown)


src/lib/runtime/files/hostname/hostname.c
```diff
@@ -0,0 +1,70 @@
...
+    if ( hostname_fd == NULL ) {
+        singularity_message(ERROR, "Couldn't create hostname session file\n");
+        ABORT(255);
+    }
+
+    if ( strlen(hostname) > HOST_NAME_MAX ) {
+        hostname[HOST_NAME_MAX] = '\0';
+    }
+
+    fprintf(hostname_fd, "%s\n", hostname);
```
gmkurtzer, line 64: I am pretty sure that we need to set the hostname directly via systemcall, not via file and count on systemd/SysVinit/upstart/pick_your_favorite_init or distribution to do it all the same.

src/lib/runtime/ns/uts/uts.c
```diff
@@ -49,6 +60,10 @@ int _singularity_runtime_ns_uts(void) {
         singularity_message(ERROR, "Could not virtualize UTS namespace: %s\n", strerror(errno));
         ABORT(255);
     }
+    if ( hostname && sethostname(hostname, hostname_len) < 0 ) {
```
gmkurtzer, line 63: Ahh, I see you are doing it in both ways. I think we should prefer to do it only via systemcall.

gmkurtzer (changes requested):

src/lib/runtime/files/hostname/hostname.c
cclerget: using only syscall won't work when --boot option is given, init systems set hostname based on /etc/hostname content (container default is often localhost.localdomain), so both are required to handle this case. Or just had a check : if boot requested set and bind /etc/hostname. Make me think when no hostname is given while starting instance to set it with instance name by default, no ?

gmkurtzer (approved):

gmkurtzer:
Thank you @cclerget!

Issue #1132, opened 2017-11-02
Incorrect permissions for container pulled from dockerhub


mvdoc:
## Version of Singularity:

2.3.2 and 2.4

### Expected behavior

Pulling a container from dockerhub should allow non-root users to run the container.

### Actual behavior

Permissions are set to root within the singularity container and unprivileged user is unable to run the container.

#### Creating the image with 2.3.2

```
mvdoc@euclid ~/singularity $ singularity --version
2.3.2-heads/2.3.2.gbf5853e
mvdoc@euclid ~/singularity $ singularity pull --size 8000 docker://poldracklab/mriqc:latest
```

Running the image fails for permission errors:

```
mvdoc@euclid ~/singularity $ singularity run -e -C mriqc-latest.img
Traceback (most recent call last):
  File "/usr/local/miniconda/bin/mriqc", line 6, in <module>
    from pkg_resources import load_entry_point
  File "<frozen importlib._bootstrap>", line 961, in _find_and_load
  File "<frozen importlib._bootstrap>", line 950, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 646, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 616, in _load_backward_compatible
  File "/usr/local/miniconda/lib/python3.6/site-packages/setuptools-27.2.0-py3.6.egg/pkg_resources/__init__.py", line 2985, in <module>
  File "/usr/local/miniconda/lib/python3.6/site-packages/setuptools-27.2.0-py3.6.egg/pkg_resources/__init__.py", line 2971, in _call_aside
  File "/usr/local/miniconda/lib/python3.6/site-packages/setuptools-27.2.0-py3.6.egg/pkg_resources/__init__.py", line 2998, in _initialize_master_working_set
  File "/usr/local/miniconda/lib/python3.6/site-packages/setuptools-27.2.0-py3.6.egg/pkg_resources/__init__.py", line 651, in _build_master
  File "/usr/local/miniconda/lib/python3.6/site-packages/setuptools-27.2.0-py3.6.egg/pkg_resources/__init__.py", line 644, in __init__
  File "/usr/local/miniconda/lib/python3.6/site-packages/setuptools-27.2.0-py3.6.egg/pkg_resources/__init__.py", line 700, in add_entry
  File "/usr/local/miniconda/lib/python3.6/site-packages/setuptools-27.2.0-py3.6.egg/pkg_resources/__init__.py", line 1990, in find_on_path
[...]

Issue #884, opened 2017-08-22
whiteout files present in singularity but not in docker container


kaczmarj:
## Version of Singularity:

```shell
$ singularity --version
2.3.1-HEAD.ge214d4e
```

```shell
$ docker --version
Docker version 17.07.0-ce-rc3, build 665d244
```

### Expected behavior

Consistency in presence of whiteout files between Docker and Singularity.

### Actual behavior

Whiteout files (e.g. ,`.wh..wh..opq`) are present in the Singularity container (pulled from DockerHub) but not in the Docker container.

### Steps to reproduce behavior

**Singularity**

```shell
$ singularity pull docker://tensorflow/tensorflow:1.3.0-gpu-py3
$ singularity shell --containall --shell bash tensorflow-1.3.0-gpu-py3.img
(container)$ find / -name ".wh.*" 2>/dev/null | grep -v "Permission denied"
```

**Docker**

```shell
$ docker run --rm -it --entrypoint bash tensorflow/tensorflow:1.3.0-gpu-py3
(container)$ find / -name ".wh.*" 2>/dev/null | grep -v "Permission denied"
```

related to #571

vsoch:
Did you try the singularity commands using sudo? It could be permissions on pull / run (as Docker has write permissions). I'm not familiar with whiteout files - could you talk a bit about how/why they exist? For the longest time I thought those were python wheels :)

vsoch:
I just tried with sudo and they indeed persist, I wonder why this happens? Do you think it would be appropriate for singularity to do a sweep of a container, always, and remove?

```
find / -name ".wh.*" -exec rm -f {} \;
```

kaczmarj:
I know very little about whiteout files, but they seem to be part of the overlay filesystem. I found these resources ([linux kernel docs](https://github.com/torvalds/linux/blob/6b1c776d3efbda31085b6a9f3bc7f774511fafd9/Documentation/filesystems/overlayfs.txt#L92-L107) and [opencontainers layer spec](https://github.com/opencontainers/image-spec/blob/master/layer.md#whiteouts)) useful.

Here is a quote from [opencontainers](https://github.com/opencontainers/image-spec/blob/master/layer.md#opaque-whiteout):
> An opaque whiteout entry is a file with the name .wh..wh..opq indicating that all siblings are hidden in the lower layer.

[...]

Issue #571, opened 2017-04-07
guidelines for creating readonly docker containers compatible with singularity


satra:
since docker has the ability to overwrite files within the container, we are running into weird issues that would benefit from better container creation. turns out certain python pieces want to write things in some underlying hidden files, and if those things were not cleaned up during the container build process, singularity runs into an error, since everything is read-only.

are there any guidelines we can create that would support better container creation/testing between singularity and docker? for example, now i plan to go and make sure there is a line to remove these `.wh` files from the conda directory.

```
inside singularity container derived from docker> nipypecli 
Traceback (most recent call last):
  File "/opt/conda/bin/nipypecli", line 6, in <module>
    from pkg_resources import load_entry_point
  File "<frozen importlib._bootstrap>", line 969, in _find_and_load
  File "<frozen importlib._bootstrap>", line 958, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 664, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 634, in _load_backward_compatible
  File "/opt/conda/lib/python3.5/site-packages/setuptools-27.2.0-py3.5.egg/pkg_resources/__init__.py", line 2985, in <module>
  File "/opt/conda/lib/python3.5/site-packages/setuptools-27.2.0-py3.5.egg/pkg_resources/__init__.py", line 2971, in _call_aside
  File "/opt/conda/lib/python3.5/site-packages/setuptools-27.2.0-py3.5.egg/pkg_resources/__init__.py", line 2998, in _initialize_master_working_set
  File "/opt/conda/lib/python3.5/site-packages/setuptools-27.2.0-py3.5.egg/pkg_resources/__init__.py", line 651, in _build_master
  File "/opt/conda/lib/python3.5/site-packages/setuptools-27.2.0-py3.5.egg/pkg_resources/__init__.py", line 644, in __init__
  File "/opt/conda/lib/python3.5/site-packages/setuptools-27.2.0-py3.5.egg/pkg_resources/__init__.py", line 700, in add_entry
  File "/opt/conda/lib/python3.5/site-packages/setuptools-27.2.0-py3.5.egg/pkg_resources/__init__.py", line 1990, in find_on_path
  File "/opt/conda/lib/python3.5/site-packages/setuptools-27.2.0-py3.5.egg/pkg_resources/__init__.py", line 2399, in from_location
[...]

PR #1181, opened 2017-11-29
 Exclude aufs whiteout files from docker extraction (Alternative C/libarchive impl.)

dtrudg:

**This is an alternative implementation of PR #1106 using C code and libarchive to provide libexec/bin/docker-extract for aufs whiteout handling instead of complex shell code.**

Docker layer tars can contain .wh.* files/dirs that are whiteout files from aufs. When extracted to a singularity image this causes problems. E.g.

singularity pull docker://nvidia/digits:6.0
singularity run digits-6.0.img

... will fail with python errors related to permission issues as python tries to scan .wh* files and dirs.

The aufs whiteout files instruct container software how to remove files from previous layers, that were removed during a later Dockerfile 'RUN' step. These removals must be performed by Singularity to ensure a container contains the same content as when run with docker.

The issue is illustrated with the test container at this Git repository:

https://github.com/dctrud/docker-aufs-sanity

- Ref: #571 #884 #1132 (part of issue)

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/.travis/rpmbuild_test
+++ b/.travis/rpmbuild_test
@@ -5,7 +5,7 @@
 OS_VERSION="$1"
 
 # build and install
-yum install -y libtool rpm-build make squashfs-tools openssl-devel libuuid-devel
+yum install -y libtool rpm-build make squashfs-tools openssl-devel libuuid-devel libarchive-devel
 ./autogen.sh
 ./configure
 make dist
--- a/.travis/rpmbuild_test
+++ b/.travis/rpmbuild_test
@@ -7,7 +7,7 @@ OS_VERSION="$1"
 # build and install
 yum install -y libtool rpm-build make yum-utils
 ./autogen.sh
-yum-builddep singularity.spec
+yum-builddep -y singularity.spec
 ./configure
 make dist
 rpmbuild -ta *.tar.gz
--- a/.travis/rpmbuild_test
+++ b/.travis/rpmbuild_test
@@ -1,12 +1,13 @@
 #!/bin/bash -ex
 
-# this script runs as root under docker 
+# this script runs as root under docker
 
 OS_VERSION="$1"
 
 # build and install
-yum install -y libtool rpm-build make squashfs-tools openssl-devel libuuid-devel libarchive-devel
+yum install -y libtool rpm-build make yum-utils
 ./autogen.sh
+yum-builddep singularity.spec
 ./configure
 make dist
 rpmbuild -ta *.tar.gz
--- a/.travis/rpmbuild_test
+++ b/.travis/rpmbuild_test
@@ -5,7 +5,7 @@
 OS_VERSION="$1"
 
 # build and install
-yum install -y libtool rpm-build make yum-utils
+yum install -y libtool rpm-build make yum-utils squashfs-tools
 ./autogen.sh
 yum-builddep -y singularity.spec
 ./configure
--- a/libexec/cli/build.exec
+++ b/libexec/cli/build.exec
@@ -1,25 +1,25 @@
 #!/bin/bash
-# 
+#
 # Copyright (c) 2017, SingularityWare, LLC. All rights reserved.
 #
 # Copyright (c) 2015-2017, Gregory M. Kurtzer. All rights reserved.
-# 
+#
 # Copyright (c) 2016-2017, The Regents of the University of California,
 # through Lawrence Berkeley National Laboratory (subject to receipt of any
 # required approvals from the U.S. Dept. of Energy).  All rights reserved.
-# 
+#
 # This software is licensed under a customized 3-clause BSD license.  Please
 # consult LICENSE file distributed with the sources of this project regarding
 # your rights to use or distribute this software.
-# 
+#
 # NOTICE.  This Software was developed under funding from the U.S. Department of
 # Energy and the U.S. Government consequently retains certain rights. As such,
 # the U.S. Government has been granted for itself and others acting on its
 # behalf a paid-up, nonexclusive, irrevocable, worldwide license in the Software
 # to reproduce, distribute copies to the public, prepare derivative works, and
-# perform publicly and display publicly, and to permit other to do so. 
-# 
-# 
+# perform publicly and display publicly, and to permit other to do so.
+#
+#
 set -e
 
 ## Basic sanity
@@ -181,7 +181,7 @@ if [ "$USERID" != "0" ]; then
     if [ -n "${SINGULARITY_SANDBOX:-}" ]; then
 
         # Sandbox with deffile is no go.
-        if eval is_deffile "${SINGULARITY_BUILDDEF}"; then 
+        if eval is_deffile "${SINGULARITY_BUILDDEF}"; then
             message ERROR "You must be the root user to sandbox from a Singularity recipe file\n"
             exit 1
 
@@ -200,11 +200,11 @@ fi
 # if the container exists, build into it (or delete it if -F was passed)
 if [ -e "$SINGULARITY_CONTAINER_OUTPUT" ]; then
     if eval is_image "${SINGULARITY_CONTAINER_OUTPUT}"; then
-        if [ -z "${SINGULARITY_FORCE:-}" ]; then 
+        if [ -z "${SINGULARITY_FORCE:-}" ]; then
             message 1 "Building into existing container: $SINGULARITY_CONTAINER_OUTPUT\n"
             SINGULARITY_BTSTRP_2EXISTING=1
         elif [ -n "${SINGULARITY_FORCE:-}" ]; then
-            # this may be a directory, so rm -rf 
+            # this may be a directory, so rm -rf
             rm -rf "$SINGULARITY_CONTAINER_OUTPUT" 2> /dev/null
         fi
     else
@@ -233,7 +233,7 @@ else
     fi
     SINGULARITY_IMAGE="$SINGULARITY_ROOTFS"
 
-    # create a temporary placeholder location to build file system 
+    # create a temporary placeholder location to build file system
      if ! SINGULARITY_TEMP_FS=`mktemp "${SINGULARITY_TMPDIR:-/tmp}"/.singularity-temp-fs.XXXXXX`; then
         message ERROR "Failed to create temporary file system\n"
         ABORT 255
@@ -265,8 +265,6 @@ case $SINGULARITY_BUILDDEF in
         for i in `cat "$SINGULARITY_CONTENTS"`; do
             name=`basename "$i"`
             message 2 "Exploding layer: $name\n"
-            echo ${SINGULARITY_ROOTFS}
-            echo $SINGULARITY_libexecdir/singularity/bin/docker-extract "$i"
             $SINGULARITY_libexecdir/singularity/bin/docker-extract "$i"
         done
 
@@ -291,7 +289,7 @@ case $SINGULARITY_BUILDDEF in
         # relying on pull.py for error checking here
         ${SINGULARITY_libexecdir}/singularity/python/pull.py
 
-        # switch $SINGULARITY_CONTAINER from remote to local 
+        # switch $SINGULARITY_CONTAINER from remote to local
         SINGULARITY_BUILDDEF=`cat $SINGULARITY_CONTENTS`
         SINGULARITY_CLEANUP=1
     ;;
@@ -316,7 +314,7 @@ if [ -f "${SINGULARITY_BUILDDEF:-}" ]; then
             ABORT 255
         fi
 
-    elif eval is_deffile "${SINGULARITY_BUILDDEF}"; then 
+    elif eval is_deffile "${SINGULARITY_BUILDDEF}"; then
         message 1 "Using container recipe deffile: $SINGULARITY_BUILDDEF\n"
         if [ "$USERID" != "0" ]; then
             message ERROR "You must be the root user to build from a Singularity recipe file\n"
@@ -365,7 +363,7 @@ if [ -z "${SINGULARITY_SANDBOX:-}" ]; then
 
         if [ -z "${SINGULARITY_BTSTRP_2EXISTING:-}" ]; then
             CONTAINER_SIZE=`du -sm ${SINGULARITY_ROOTFS} | cut -f 1`
-            # using the let builtin instead of bc which is not guaranteed 
+            # using the let builtin instead of bc which is not guaranteed
             let "IMAGE_SIZE=(${CONTAINER_SIZE:-768} * 125)/100"
             if [ "$IMAGE_SIZE" -lt 10 ]; then
                 IMAGE_SIZE=10
--- a/libexec/cli/build.exec
+++ b/libexec/cli/build.exec
@@ -263,8 +263,11 @@ case $SINGULARITY_BUILDDEF in
         zcat "$SINGULARITY_libexecdir/singularity/bootstrap-scripts/environment.tar" | tar xBf - -C "${SINGULARITY_ROOTFS}" || exit $?
 
         for i in `cat "$SINGULARITY_CONTENTS"`; do
-            message 1 "Importing: $i\n"
-            zcat "$i" | (cd "$SINGULARITY_ROOTFS"; tar --exclude=dev/* -xf - ) || exit $?
+            name=`basename "$i"`
+            message 2 "Exploding layer: $name\n"
+            echo ${SINGULARITY_ROOTFS}
+            echo $SINGULARITY_libexecdir/singularity/bin/docker-extract "$i"
+            $SINGULARITY_libexecdir/singularity/bin/docker-extract "$i"
         done
 
         SINGULARITY_BUILDDEF="$SINGULARITY_ROOTFS"
--- a/src/docker-extract.c
+++ b/src/docker-extract.c
@@ -24,20 +24,22 @@
  */
 int apply_opaque(const char *opq_marker, char *rootfs_dir) {
     int retval = 0;
-    char *token = strrchr(opq_marker, '/');
+    char *token, *opq_dir, *opq_dir_rootfs;
+    size_t buff_len;
+
+    token = strrchr(opq_marker, '/');
 
     if (token == NULL) {
         singularity_message(ERROR, "Error getting dirname for opaque marker\n");
         ABORT(255);
     }
 
-    size_t length = token - opq_marker;
-    char *opq_dir = malloc(length + 1);
-    strncpy(opq_dir, opq_marker, length);
-    opq_dir[length] = 0;
+    buff_len = token - opq_marker + 1;
+    opq_dir = malloc(buff_len);
+    snprintf(opq_dir, buff_len, "%s", opq_marker);
 
-    size_t buff_len = strlen(rootfs_dir) + 1 + strlen(opq_dir) + 1;
-    char *opq_dir_rootfs = malloc(buff_len);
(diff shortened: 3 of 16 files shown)


dtrudg:
Travis failure on CentOS7 is odd. Don't think this is due to issue in this commit, since 03346fc passed

```
 + sudo singularity build --isolated --writable /tmp/stest.6UkvFY/container2 ../examples/busy    (retval=0) OK
 + sudo singularity build --isolated /tmp/stest.6UkvFY/container /tmp/stest.6UkvFY/Singularit    (retval=0) OK
 + singularity exec /tmp/stest.6UkvFY/container true                                             (retval=0) OK
 + singularity exec /tmp/stest.6UkvFY/container false                                            (retval=1) OK
 + singularity exec /tmp/stest.6UkvFY/container test -f /.singularity.d/runscript                (retval=0) OK
 + singularity exec /tmp/stest.6UkvFY/container test -f /.singularity.d/env/01-base.sh           (retval=0) OK
 + singularity exec /tmp/stest.6UkvFY/container test -f /.singularity.d/actions/shell            (retval=0) OK
 + singularity exec /tmp/stest.6UkvFY/container test -f /.singularity.d/actions/exec             (retval=0) OK
 + singularity exec /tmp/stest.6UkvFY/container test -f /.singularity.d/actions/run              (retval=0) OK
 + singularity exec /tmp/stest.6UkvFY/container test -L /environment                             (retval=0) OK
 + singularity exec /tmp/stest.6UkvFY/container test -L /singularity                             (retval=0) OK
 + sudo singularity -x build --isolated --force --sandbox /tmp/stest.6UkvFY/container2 ../exa    (retval=0) OK
 + sudo singularity build --isolated /tmp/stest.6UkvFY/container /tmp/stest.6UkvFY/Singularit  (retval=255) ERROR
Using container recipe deffile: /tmp/stest.6UkvFY/Singularity
Building into existing container: /root/build
Using container recipe deffile: Singularity
Sanitizing environment
[0mAdding base Singularity environment to container
Exporting contents of container2 to /root/build
Found an existing definition file
Adding a bootstrap_history directory
Calculating final size for metadata...
Skipping checks
Singularity container built: /root/build
Cleaning up...
Building Singularity FS image...
Write failed because Bad file descriptor
FATAL ERROR:Failed to write to output filesystem
[91mERROR  : Image path doesn't exists
[0m[31mABORT  : Retval = 255
[0mCleaning up...
Full output in: /tmp/stest.6UkvFY
ERROR: Failed running test: 20-build.sh
make: *** [test] Error 1
```

src/docker-extract.c
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    free(orig_dir);
+
+    return (retval);
+}
+
+int main(int argc, char **argv) {
+    int retval = 0;
+    char *rootfs_dir = envar_path("SINGULARITY_ROOTFS");
```
GodloveD: You should probably get this from the registry rather than pulling it directly from the environment.  I don't know if that prevents a race condition or something, but at least for consistency's sake.

src/docker-extract.c
```diff
@@ -0,0 +1,320 @@
...
+        singularity_message(ERROR, "Environment is not properly setup\n");
+        ABORT(255);
+    }
+
+    if (is_dir(rootfs_dir) < 0) {
+        singularity_message(ERROR, "SINGULARITY_ROOTFS does not exist\n");
+        ABORT(255);
+    }
+
+    if (argc != 2) {
```
GodloveD: Seems like this check could probably be first?  If the input isn't right why bother doing anything else?

libexec/cli/build.exec
```diff
@@ -263,8 +263,11 @@ case $SINGULARITY_BUILDDEF in
         zcat "$SINGULARITY_libexecdir/singularity/bootstrap-scripts/environment.tar" | tar xBf - -C "${SINGULARITY_ROOTFS}" || exit $?
 
         for i in `cat "$SINGULARITY_CONTENTS"`; do
-            message 1 "Importing: $i\n"
-            zcat "$i" | (cd "$SINGULARITY_ROOTFS"; tar --exclude=dev/* -xf - ) || exit $?
+            name=`basename "$i"`
+            message 2 "Exploding layer: $name\n"
+            echo ${SINGULARITY_ROOTFS}
+            echo $SINGULARITY_libexecdir/singularity/bin/docker-extract "$i"
```
GodloveD: Are these echos debugging cruft?

src/docker-extract.c
```diff
@@ -0,0 +1,320 @@
...
+    free(wh_path_rootfs);
+
+    return retval;
+}
+
+/* apply_whiteouts
+ *  Process tarfile and apply any aufs opaque/whiteouts on rootfs_dir
+ */
+int apply_whiteouts(char *tarfile, char *rootfs_dir) {
+    int ret = 0;
```
GodloveD: nitpicky, but maybe name this variable `retval` instead for consistency.

src/docker-extract.c
```diff
@@ -0,0 +1,320 @@
...
+    archive_read_support_format_all(a);
+#if ARCHIVE_VERSION_NUMBER <= 3000000
+    archive_read_support_compression_all(a);
+#else
+    archive_read_support_filter_all(a);
+#endif
+    ext = archive_write_disk_new();
+    archive_write_disk_set_options(ext, flags);
+    archive_write_disk_set_standard_lookup(ext);
+    if ((r = archive_read_open_filename(a, tarfile, 10240))){
```
GodloveD, line 212: Naive question.  Can we always be sure that the block size will be 10240?

src/docker-extract.c
```diff
@@ -0,0 +1,320 @@
...
+    archive_read_finish(a);
+#else
+    archive_read_free(a);
+#endif
+    archive_write_close(ext);
+#if ARCHIVE_VERSION_NUMBER <= 3000000
+    archive_write_finish(ext);
+#else
+    archive_write_free(ext);
+#endif
```
GodloveD: Could the closes above be grouped so that only a single if statement is needed?

src/docker-extract.c
```diff
@@ -0,0 +1,320 @@
...
+            break;
+
+        if (r < ARCHIVE_OK)
+            singularity_message(WARNING, "Warning reading tar header: %s\n", archive_error_string(a));
+        if (r < ARCHIVE_WARN) {
+            singularity_message(ERROR, "Error reading tar header: %s\n", archive_error_string(a));
+            ABORT(255);
+        }
+
+        const char *pathname = archive_entry_pathname(entry);
```
GodloveD: As a general rule I think we want to avoid declaring variables in the middle of scope like this.  Some of the C programmers on the team can give more insight on best practices.

.travis/rpmbuild_test
```diff
@@ -5,7 +5,7 @@
 OS_VERSION="$1"
 
 # build and install
-yum install -y libtool rpm-build make squashfs-tools openssl-devel libuuid-devel
+yum install -y libtool rpm-build make squashfs-tools openssl-devel libuuid-devel libarchive-devel
```
GodloveD: I think you may need to remove the dep here and put it in the spec and deb files instead.  See https://github.com/singularityware/singularity/pull/1198 and https://github.com/singularityware/singularity/pull/1204 for examples of what we were doing wrong and how to fix.

src/docker-extract.c
```diff
@@ -0,0 +1,320 @@
...
+#include "util/file.h"
+#include "util/message.h"
+#include "util/util.h"
+
+/* apply_opaque
+ *  Given opq_marker as a path to a whiteout opaque marker
+ *    e.g. usr/share/doc/test/.wh..wh..opq
+ *  Make the directory containing the make opaque for this layer by removing it
+ *  if it exists under rootfs_dir
+ */
```
GodloveD, line 25: Based on the logic here I was expecting directories marked with an opq file to be deleted.  But I see that those directories still exist in a DIGITS image created using this branch.  Moreover in most cases (but not all) the contents of directories marked with a opq file is identical to the contents of the directory if the image is built with the current dist version instead.  What am I missing here?  Is the directory recreated when the current layer is applied?  Is that the expected/desired behavior?

GodloveD (changes requested):
Hi @dctrud.  Thanks again for taking the time to identify and provide a solution for this issue.  I've taken a first pass at review and I think I've identified some things that should be clarified and fixed.  Once we get through this round I'm going to ask one of our more experienced C devs to go over the code for best practices and security checking etc.  Thanks again.  I look forward to hearing more.  😄

src/docker-extract.c
```diff
@@ -0,0 +1,320 @@
...
+            r = copy_data(a, ext);
+            if (r < ARCHIVE_OK)
+                singularity_message(WARNING, "Warning handling tar header: %s\n", archive_error_string(ext));
+            if (r < ARCHIVE_WARN) {
+                singularity_message(ERROR, "Error handling tar header: %s\n", archive_error_string(ext));
+                ABORT(255);
+            }
+        }
+        r = archive_write_finish_entry(ext);
+        if (r < ARCHIVE_OK)
```
GodloveD: Sorry, meant to include this in the previous review.  I'm going to advocate that we always use curly braces for if statements even if they are just simple one liners like this.  Otherwise my brain skips a beat when I'm reading and I have to stop and do a double take. lol

libexec/cli/build.exec
dtrudg: Ahh yes. Whoops.

src/docker-extract.c
dtrudg: An opaque directory is one that was removed, but then recreated later. It`s opaque in that the files from earlier layers don’t show through. Believe I’m implementing this correctly. Directory is remove when processing the white outs then will reappear when layer tar is extracted.

src/docker-extract.c
dtrudg: Yeah this could be tidied a bit. Most of the tar stuff here is from a libarchive example quickly modified.

dtrudg:
@GodloveD thanks for the comments. Traveling light for Christmas and New Year in the UK with just an iPad right now, so will likely make changes after get back on 7th Jan.

src/docker-extract.c
GodloveD: OK.  As long as it is working as expected. Thanks for clarification. 😸

src/docker-extract.c
dtrudg: AFAIK the value here doesn't really matter, except for performance reasons.  In the IO callback docs they specify:

>To read an archive, you need to provide libarchive with a _read callback_ function. The read callback is expected to provide a _block_ of data to libarchive each time it is called.

>Blocks can be of any size or alignment at all. There are only two requirements:

>Each block must be at least one byte long. (A zero-byte block indicates end-of-file.)
When you return a pointer to a block, the memory where the block is stored must be stable until the next time your read function is called.

https://github.com/libarchive/libarchive/wiki/LibarchiveIO

From there it goes on to mention performance tuning, matching hardware block size etc. but there is no hard requirement on the block size.

For libarchive 3+ it may not use the specified value - but will inspect the archive for a block size to use:

>Note 1: Beginning with libarchive 3.0, the `archive_read_open_filename()` function inspects the file before deciding how to handle the blocksize. If the filename provided refers to a tape device, for example, it will use exactly the blocksize you specify. For other devices, it may adjust the requested blocksize in order to obtain better performance.

https://github.com/libarchive/libarchive/wiki/Examples

The 10240 is referenced elsewhere in docs as default for *write* for compatibility with old tar versions, and this is taken from the 'universal reader' example.

GodloveD: Makes sense.  Thanks for the clarification and the pointer to the docs.  😸

dtrudg:
@GodloveD - I've had a chance to make some changes here, tidying it up. Let me know if there's anything else you'd like to see.

src/docker-extract.c
```diff
@@ -0,0 +1,342 @@
...
+ */
+int extract_tar(const char *tarfile, const char *rootfs_dir) {
+    int retval = 0;
+    struct archive *a;
+    struct archive *ext;
+    struct archive_entry *entry;
+    int flags;
+    int r;
+    char *orig_dir;
+    const char *pathname;
```
GodloveD, line 191: Actually I think that pathname should be declared at the beginning of the for loop where it is used since that is the only scope it is needed for.

GodloveD (commented):
I'm mostly happy with this, although I think that variable declaration is still not happening in exactly the right place.  But I'm going to leave that for a more experienced C programmer like @yhcote to review.  If he or another of our C devs approves this I am happy to approve to and we can go ahead and get it merged. 

Also we are obviously going to need to make sure this is passing tests in Travis and that all conflicts are resolved before merging.  I'm not totally sure why I'm not seeing any passed or failed travis tests  on this 😕

dtrudg:
@GodloveD - the travis test on Ubuntu seems to be passing now, but the CentOS builds are failing due to squashfs-tools deps. Not sure exactly how best to change this so it works but is consistent with the previous stuff you mentioned on changing approach to deps?

GodloveD:
Hmm, yeah I noticed that.  squashfs-tools does need to be present on the host afaik.  Maybe adding it back in will lead to a passing build?

dtrudg:
Well 2 green lights now, but CentOS7 failed with what I think might be a docker pull net issue and not related to the changes?

```
    raise BadStatusLine(line)
httplib.BadStatusLine: ''
Cleaning up...
Full output in: /tmp/stest.FSX6za
ERROR: Failed running test: 40-privblock.sh
```

I'll leave it to the eyes of the C and packaging gods now...

jmstover:
Looks like a network, or remote server error. I've restarted the failed
CentOS7 job to see if it succeeds with a bit of time here between runs.

-J

jmstover:
Wow... there's some uhh... issues:

https://travis-ci.org/singularityware/singularity/jobs/327450386

-J

GodloveD:
OK.  Looks like MongoDB got it's act together and things are passing now.

src/docker-extract.c
```diff
@@ -0,0 +1,342 @@
...
+    char *token, *opq_dir, *opq_dir_rootfs;
+    size_t buff_len;
+
+    token = strrchr(opq_marker, '/');
+
+    if (token == NULL) {
+        singularity_message(ERROR, "Error getting dirname for opaque marker\n");
+        ABORT(255);
+    }
+
```
yhcote, line 36: To simplify and because that code doesn't run multi-threaded, I would define a static buffer for string manipulation and I would make sure snprintf() is able to NULL terminate the string. Here's what you can do instead : 

int apply_opaque(const char *opq_marker, char *rootfs_dir) {                                                                 
    int retval;                                                                                                              
    char *token;                                                                                                             
    static char buf[MAXPATHLEN];                                                                                             
                                                                                                                             
    token = strrchr(opq_marker, '/');                                                                                        
    if (token == NULL) {                                                                                                     
        singularity_message(ERROR, "Error getting dirname for opaque marker\n");                                             
        ABORT(255);                                                                                                          
    }                                                                                                                        
                                                                                                                             
    retval = snprintf(buf, sizeof(buf), "%s/%s", rootfs_dir, opq_marker);                                                    
    if (retval == -1 || retval >= sizeof(buf)) {                                                                             
        singularity_message(ERROR, "Error with pathname too long\n");                                                        
        ABORT(255);                                                                                                          
    }                                                                                                                        
                                                                                                                             
    if (is_dir(buf) == 0)                                                                                                    
        s_rmdir(buf);                                                                                                        
                                                                                                                             
    return 0;                                                                                                                
}

src/docker-extract.c
```diff
@@ -0,0 +1,342 @@
...
+    return retval;
+}
+
+/* apply_whiteout
+ *  Given wh_marker as a path to a whiteout marker
+ *    e.g. usr/share/doc/test/.wh.deletedfile
+ *  Whiteout the referenced file for this layer by removing it if it exists
+ *  under rootfs_dir
+ */
```
yhcote, line 55: For this function also, manipulating strings with snprintf and using buffers, you can use the example I wrote to make it more simple.

yhcote (changes requested):
The main concern is to make sure that a string is always null terminated with snprintf(). Restricting the copy length with the "size" param of snprintf() is not enough to guarantee that the resulting string will be null terminated if the string to be copied is larger than the buffer to contain it.

src/docker-extract.c
dtrudg: Thanks for the pointers - embarrassing how much sensible C I've forgotten living in Python etc. Just grepped through some C code from way back when I was 17 and found lots of `char buf[MAXLINE]` etc. Need to reactivate that part of my memory somehow 😨

src/docker-extract.c
dtrudg: Hopefully c8baf0d is something more like it for this too.

yhcote (approved):
Looks good to me

dtrudg:
@GodloveD - do you need any more changes? Thanks.

.travis/rpmbuild_test
```diff
@@ -1,12 +1,13 @@
 #!/bin/bash -ex
 
-# this script runs as root under docker 
+# this script runs as root under docker
 
 OS_VERSION="$1"
 
 # build and install
-yum install -y libtool rpm-build make squashfs-tools openssl-devel libuuid-devel
+yum install -y libtool rpm-build make yum-utils squashfs-tools
```
GodloveD, line 8: Remind me, did we need squashfs-tools here for the build to complete?  If not, I think we should remove it.

GodloveD: nvrmnd.  I see that we do need it.  Sorry for the noise.

vsoch: +1, I've hit this bug a bunch of times forgetting to install it :)

dtrudg: Yeah...

>GodloveD commented 8 days ago
Hmm, yeah I noticed that. squashfs-tools does need to be present on the host afaik. Maybe adding it back in will lead to a passing build?

>dctrud Add back squashfs-tools to rpmbuild_test script Verified 6ba0a1b

It was needed previously, had to add back in to make the build complete and get a travis checkmark. But... maybe this has been fixed by other things from the merges into development since then? Can't check right now if removing it is okay - but can tonight if needed.

GodloveD (approved):

.travis/rpmbuild_test
dtrudg: Ahh okay sounds like its needed  :-)

GodloveD: Things actually may have changed because @jmstover did some work recently to clean up dependencies in our test suite.  But don't worry I'm having a look myself.  Will let you know.  Thanks!

GodloveD:
OK.  squashfs-tools is totally still needed. 
https://travis-ci.org/GodloveD/singularity/jobs/330564707
Thanks for your hard work and patience on this @dctrud!  And thanks for the useful comments and reviews @yarikoptic and @vsoch !

olifre:
> OK. squashfs-tools is totally still needed.

Yeah. But installing it before building the package is not needed, so this was the wrong fix. 
As you can see in #1204 , the correct fix would have been to use `yum install -y` to install the built rpms, which automatically pulls in `squashfs-tools` instead of hardcoding the dependency and thus making the`rpmbuild` test useless. 
I'll rebase #1204 on current `development` tomorrow and revert this broken commit then as part of that PR.

GodloveD:
Great.  Thanks @olifre.  Still trying to wrap my head around the subtleties of packaging.

PR #1189, opened 2017-12-01
apprun script backslash removal fix

bleuchien:
Hi,

during image generation unescaped backslashes are removed from the %apprun script section. This behavior is different to the %runscript section which preserves backslashes. To make both script sections behave the same and make it possible to copy and paste existing shell code to the apprun section without modifications the bash read command needs the option -r to ignore (preserve) backslashes.

Example:

$ cat test-image.def
Bootstrap: docker
From: fedora:26
%runscript
  echo "foo\nbar\n"
%apprun test
  echo "foo\nbar\n"

$ singularity build test1.simg test-image.def
$ singularity exec test1.simg bash
Singularity> cat /singularity 
#!/bin/sh 

  echo "foo\nbar\n"
Singularity> cat /scif/apps/test/scif/runscript
  echo "foonbarn"

Cheers
Richard

**Checkoff for all PRs:**

  -> I've tested the new image build with success
- [-] I have added myself as a contributor to the [contributors's file](https://github.com/singularityware/singularity/blob/master/CONTRIBUTORS.md)
  -> no, for this small fix I didn't think it necessary to be listed as contributor

Attn: @singularityware-admin

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ and changes prior to that are (unfortunately) done retrospectively. Critical ite
 
 ## [v2.4.1](https://github.com/singularityware/singularity/tree/development) (development)
 
+### apprun script backslash removal fix
+ - Fixed the unwanted removal of backslashes in apprun scripts
+
 ### Security related fixes
  - Fixed container path and owner limitations (original merge was lost)
  - Check of overlay upper/work images are symlinks
--- a/libexec/bootstrap-scripts/functions
+++ b/libexec/bootstrap-scripts/functions
@@ -74,7 +74,7 @@ get_section() {
     RECIPE="$2"
     FOUND="no"
 
-    while read line
+    while read -r line
     do
 
  


GodloveD (approved):
It would be nice if we were using the same code path for the main runscript and for app specific runscripts so we could avoid discrepancies like this. But in the meantime, this is a good fix.  Thanks for identifying the problem and providing a solution @bleuchien!  If another member of the team signs off I am happy to merge.  @vsoch what do you think?

vsoch:
I agree that as a user I would expect the same functionality! Note that the same function was intended but changed because it didn't work directly with doing the same for apps: https://github.com/bleuchien/singularity/commit/b3afcbf2990c53058964446690a9e8d00f9ce598.

GodloveD:
Thanks @vsoch.  So are we good to go on this?

vsoch:
I did some additional testing to make sure the sections were still parsed correctly, and I see some strange behavior. @GodloveD can you reproduce this? The first part looks ok:

```
sudo singularity build test.simg Singularity.app
Using container recipe deffile: Singularity.app
Sanitizing environment
Docker image path: index.docker.io/library/ubuntu:latest
Cache folder set to /root/.singularity/docker
Exploding layer: sha256:50aff78429b146489e8a6cb9334d93a6d81d5de2edc4fbf5e2d4d9253625753e.tar.gz
Exploding layer: sha256:f6d82e297bce031a3de1fa8c1587535e34579abce09a61e37f5a225a8667422f.tar.gz
Exploding layer: sha256:275abb2c8a6f1ce8e67a388a11f3cc014e98b36ff993a6ed1cc7cd6ecb4dd61b.tar.gz
Exploding layer: sha256:9f15a39356d6fc1df0a77012bf1aa2150b683e46be39d1c51bc7a320f913e322.tar.gz
Exploding layer: sha256:fc0342a94c89e477c821328ccb542e6fb86ce4ef4ebbf1098e85669e051ef0dd.tar.gz
Exploding layer: sha256:c6a9ef4b9995d615851d7786fbc2fe72f72321bee1a87d66919b881a0336525a.tar.gz
Adding environment to container
Adding runscript
foo has runscript definition
bar has runscript definition
Adding custom environment to bar
Adding custom environment to foo
```
but then when we get here, the pipe is put onto another line. Is this expected?
```
Installing foo
+ cd /scif/apps/foo
+ + grep SINGULARITY
env
SINGULARITY_CHECKLEVEL=3
SINGULARITY_CHECKTAGS=bootstrap
SINGULARITY_BUILDDEF=Singularity.app
SINGULARITY_STAGE2=1
...
```
but then for bar, it still looks different (notice the single plus instead of double)

```
Installing bar
+ cd /scif/apps/bar
+ env
+ grep SINGULARITY
SINGULARITY_CHECKLEVEL=3
SINGULARITY_CHECKTAGS=bootstrap
```
Note the variable `SINGULARITY_APPROOT`. Is correct. I would suggest building some more complicated install steps, definitely with more than one app, and if that works +1 from me!

vsoch:
The other option (to be consistent with the get-section executable) is to re-write the functions in C, and ths use the same base methods for consistency.

GodloveD:
Hi @vsoch.  Thanks for the double check.  Can you post the definition file that you used to get this error so that I can see what the expected output was please?  Thanks.

vsoch:
I think it was super simple, something like:

```
Bootstrap: docker
From: ubuntu

%appinstall foo
    env | grep SINGULARITY

%appinstall bar
    env | grep SINGULARITY

%apprun foo    
    echo "RUNNING FOO"

%apprun bar         
    echo "RUNNING BAR"
```
Thanks @GodloveD !

vsoch:
And a general list of things to be careful about (based on my experience with apps):

The careful dinosaur should be watchful that...

 1. That sections with multiple commands don't also include other sections, or their own header.
 2. That sections with multiple commands aren't truncated (or missed)
 3. That the function to get a section doesn't only return the first one of each (this is what prompted to write the function in the first place)
 4.That app names with special characters (eg, slash-is-common) are parsed ok

I'll post any others that I can think of, that's it off the top of my head atm!

GodloveD:
@vsoch OK, I did a lot of testing and pulling-out-of-hair over this.  And as far as I can ascertain, the number of `+` signs preceding each command as well as the order in which piped commands are displayed is completely random and is not affected at all by the `-r` option.  For instance in 3 different runs with the `-r` option enabled, I get:
```
[...]
+ cd /scif/apps/foo
+ + env
[...]
+ cd /scif/apps/bar
+ + grep SINGULARITY
env
```
and...
```
[...]
+ cd /scif/apps/foo
+ + grep SINGULARITY
env
[...]
+ cd /scif/apps/bar
+ + grep SINGULARITY
env
```
and then...
```
[...]
+ cd /scif/apps/foo
+ grep+ env
...[]
+ cd /scif/apps/bar
+ env
```
I also get unpredictable results when I do the same without the `-r` option enabled.

All of this is simply enabled by adding the `-x` option to the `/bin/sh` call.  Because the input is identical in all cases, (I checked by echoing to `/dev/tty`) I have to conclude that this is a shell bug.  Can you confirm or deny?

GodloveD:
@vsoch actually I can confirm that this is a bug with `/bin/sh` that is unrelated to Singularity.  If you save this text to a file called `hello-world.sh` and make it executable you can probably recreate the bug yourself.  
```sh
#!/bin/sh -x

printenv | grep HOME
echo hello world!

```
```
$ ./hello-world.sh
+ printenv
+ grep HOME
HOME=/home/ubuntu
+ echo hello world!
hello world!

$ ./hello-world.sh
+ + printenv
grep HOME
HOME=/home/ubuntu
+ echo hello world!
hello world!
```
Do you want to do any more testing on this PR?  I'm happy to merge at this point if you are OK.

vsoch:
Nuts! I just tried this out, it looks like the extra `+` comes at totally unpredictable times, sort of randomly, here are the two versions that just popped up:

```
/hello-world.sh 
+ + grep HOME
printenv
HOME=/home/vanessa
+ echo hello world!
hello world!
```
and then!

```
./hello-world.sh 
+ grep HOME
+ printenv
HOME=/home/vanessa
+ echo hello world!
hello world!
```
Did you find a reporting of the bug, or even where we might report it? It's cool that we found it period :)

Yes I think I'm satisfied as well! And please let me know thoughts on #1238, there is an entire standard and paper that are on the bootstraps of that guy!

But don't worry we have some time before the revision is due!

bleuchien:
I tried to reproduce the problem shown with the hello-world.sh script but failed. It worked repeatedly on Fedora (26), Debian (9.3) and a really old Ubuntu.
Since the second plus sign is just in the wrong position (but with correct spaces around) I'm thinking it could be a race condition in the output. Piped commands are executed at the same time. If the '+' output is done and the second print is just a bit faster than 'grep' is put in place it might just be the second '+' that's displayed. The order of piped commands in the xtrace output is always random since both are run at the same time.
If you want to report it as a bug you can do so in the <bug-bash@gnu.org> mailing list. I didn't find it mentioned but google is not particularly helpful in finding plus signs...

GodloveD:
Hi @bleuchien and @vsoch.  @jmstover and I had a closer look and we realized that this bug is specific to dash (which is aliased to /bin/sh on later Ubuntu).  So I'll bet that is why you can't replicate it @bleuchien and why @vsoch and I can.  Anyway that is neither here nor there since it has nothing to do with this PR.  I'll go ahead and merge and I'll look for the appropriate place to report the bug on dash.  Thanks everyone!

GodloveD:
Thank you!

PR #1191, opened 2017-12-01
Add  json format output to instance.list

ArangoGutierrez:

Add  json format output to instance.list

Example output:

```bash
$ singularity instance.list --json
{"instances":[{"instance":web1,"pid":10925,"img":/var/lib/singularity/img/nginx.simg},{"instance":web2,"pid":10963,"img":/var/lib/singularity/img/nginx.simg},{"instance":web3,"pid":15577,"img":/var/lib/singularity/img/nginx.simg}]}
```

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/libexec/cli/instance.list.exec
+++ b/libexec/cli/instance.list.exec
@@ -33,6 +33,7 @@ if ! USERID=`id -ru`; then
     exit 255
 fi
 
+JSONFMT="false"
 
 while true; do
     case ${1:-} in
--- a/libexec/cli/instance.list.exec
+++ b/libexec/cli/instance.list.exec
@@ -105,9 +105,9 @@ for i in $DAEMONS; do
 
         if [[ "${JSONFMT}" == "true" ]]; then
           if [ "$COUNT" -lt "$DLEN" ]; then
-            printf '\t{\n\t\t"instance":%s,\n\t\t"pid":%s,\n\t\t"img":%s\n\t},\n' "${FILE_NAME}" "${DAEMON_PID}" "${DAEMON_IMAGE:-}"
+            printf '\t{\n\t\t"instance":"%s",\n\t\t"pid":"%s",\n\t\t"img":"%s"\n\t},\n' "${FILE_NAME}" "${DAEMON_PID}" "${DAEMON_IMAGE:-}"
           else
-            printf '\t{\n\t\t"instance":%s,\n\t\t"pid":%s,\n\t\t"img":%s\n\t}' "${FILE_NAME}" "${DAEMON_PID}" "${DAEMON_IMAGE:-}"
+            printf '\t{\n\t\t"instance":"%s",\n\t\t"pid":"%s",\n\t\t"img":"%s"\n\t}' "${FILE_NAME}" "${DAEMON_PID}" "${DAEMON_IMAGE:-}"
           fi
         else
             printf "%-16s %-8s %s\n" "${FILE_NAME}" "${DAEMON_PID}" "${DAEMON_IMAGE:-}"
--- a/libexec/cli/instance.list.exec
+++ b/libexec/cli/instance.list.exec
@@ -48,9 +48,9 @@ while true; do
                 exit 1
             fi
         ;;
-        --json)
-           SINGULARITY_JSONFMT="true"
-           shift
+        -j|--json)
+            SINGULARITY_PRINT_STRUCTURED="true"
+            shift
         ;;
         -*)
             message ERROR "Unknown option: ${1:-}\n"
@@ -73,10 +73,10 @@ DLEN=`echo "$DAEMONS"|wc -l`
 let DLEN--
 JSON_PRINTF='\t{\n\t\t"instance":"%s",\n\t\t"pid":"%s",\n\t\t"img":"%s"\n\t}'
 
-if [ -z "${SINGULARITY_JSONFMT:-}" ]; then
-  message 1 "%-16s %-8s %s\n" "DAEMON NAME" "PID" "CONTAINER IMAGE"
+if [ -z "${SINGULARITY_PRINT_STRUCTURED:-}" ]; then
+    message 1 "%-16s %-8s %s\n" "DAEMON NAME" "PID" "CONTAINER IMAGE"
 else
-  printf "%s\n\t%s:[\n" "{" '"instances"'
+    printf "%s\n\t%s:[\n" "{" '"instances"'
 fi
 
 for i in $DAEMONS; do
@@ -103,26 +103,26 @@ for i in $DAEMONS; do
             fi
         fi
 
-        if [ -z "${SINGULARITY_JSONFMT:-}" ]; then
-          printf "%-16s %-8s %s\n" "${FILE_NAME}" "${DAEMON_PID}" "${DAEMON_IMAGE:-}"
+        if [ -z "${SINGULARITY_PRINT_STRUCTURED:-}" ]; then
+            printf "%-16s %-8s %s\n" "${FILE_NAME}" "${DAEMON_PID}" "${DAEMON_IMAGE:-}"
         else
-          if [ "$COUNT" -lt "$DLEN" ]; then
-            printf "${JSON_PRINTF},\n" "${FILE_NAME}" "${DAEMON_PID}" "${DAEMON_IMAGE:-}"
-          else
-            printf "${JSON_PRINTF}" "${FILE_NAME}" "${DAEMON_PID}" "${DAEMON_IMAGE:-}"
-          fi
+            if [ "$COUNT" -lt "$DLEN" ]; then
+                printf "${JSON_PRINTF},\n" "${FILE_NAME}" "${DAEMON_PID}" "${DAEMON_IMAGE:-}"
+            else
+                printf "${JSON_PRINTF}" "${FILE_NAME}" "${DAEMON_PID}" "${DAEMON_IMAGE:-}"
+            fi
         fi
 
         let "COUNT++"
     fi
 done
 
-if [ -z "${SINGULARITY_JSONFMT:-}" ]; then
-  printf "\n"
+if [ -z "${SINGULARITY_PRINT_STRUCTURED:-}" ]; then
+    printf "\n"
 else
-  printf "\n\t]\n}\n"
+    printf "\n\t]\n}\n"
 fi
 
 if [ "$COUNT" == 0 ]; then
-  exit 1
+    exit 1
 fi
--- a/libexec/cli/instance.list.exec
+++ b/libexec/cli/instance.list.exec
@@ -1,12 +1,12 @@
 #!/bin/bash
-# 
+#
 # Copyright (c) 2017, SingularityWare, LLC. All rights reserved.
 #
 # This software is licensed under a 3-clause BSD license.  Please
 # consult LICENSE file distributed with the sources of this project regarding
-# your rights to use or distribute this software. 
+# your rights to use or distribute this software.
 #
-# 
+#
 
 
 ## Basic sanity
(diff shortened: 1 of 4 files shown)


libexec/cli/instance.list.exec
```diff
@@ -49,6 +110,10 @@ while true; do
                 exit 1
             fi
         ;;
+        --json)
+          jsonFmt
```
GodloveD: I would like to see this controlled by an environment variable instead of calling the function directly here.  This way you can combine your function with the code below and control whether it prints or constructs a json via the env var.

libexec/cli/instance.list.exec
```diff
@@ -33,6 +33,67 @@ if ! USERID=`id -ru`; then
...
+  if [ -z "${1:-}" ]; then
+      DAEMONS=`singularity_daemon_glob '*' | sort | uniq`
+  else
+      DAEMONS=`singularity_daemon_glob "$@" | sort | uniq`
+  fi
+
+DLEN=0
+for i in $DAEMONS; do
+	let "DLEN++"
+done
```
GodloveD: I think you could do this in a one-liner if you echo $DAEMONS with quotes to preserve newlines and use wc.  But I'm not convinced this variable is necessary at all.  I think you could change your if statement in the for loop below to look at the one-liner instead of using the variable.  Might be a slight bit less efficient but would increase readability imho.

libexec/cli/instance.list.exec
```diff
@@ -33,6 +33,67 @@ if ! USERID=`id -ru`; then
...
+                      message WARN "Removing stale daemon file: $i\n"
+                      rm -f "$i"
+                      continue
+                  fi
+              else
+                  rm -f "$i"
+                  continue
+              fi
+          fi
+
```
GodloveD: All of this code is basically just copied and pasted from the code below with the one change of constructing a json instead of printing each line.  That is not going to be very sustainable if we have to change this algorithm.  Can you just wrap the if statement below in another if statement so that it prints if the json environmental variable is not set and it creates the json if the json env var is set?

GodloveD (changes requested):
Hi @ArangoGutierrez!  Thanks for identifying this new potential feature and for taking the time to implement it!  

I like the idea, but I'd like to see some changes in it's implementation.  I'd really like if you could do this without duplicating any of the existing code.  I think that if you set an environment variable in the switch statement that parses the input options you could could add a single if statement that would control whether the code prints daemon names line by line or constructs json output.  

I'm also going to as @vsoch to look this over and tell me if she thinks the json format is optimal. 

Thanks!

ArangoGutierrez:
Sure I will attend the changes!

On Jan 4, 2018 2:54 PM, "David Godlove" <notifications@github.com> wrote:

*@GodloveD* requested changes on this pull request.

Hi @ArangoGutierrez <https://github.com/arangogutierrez>! Thanks for
identifying this new potential feature and for taking the time to implement
it!

I like the idea, but I'd like to see some changes in it's implementation.
I'd really like if you could do this without duplicating any of the
existing code. I think that if you set an environment variable in the
switch statement that parses the input options you could could add a single
if statement that would control whether the code prints daemon names line
by line or constructs json output.

I'm also going to as @vsoch <https://github.com/vsoch> to look this over
and tell me if she thinks the json format is optimal.

Thanks!
------------------------------

In libexec/cli/instance.list.exec
<https://github.com/singularityware/singularity/pull/1191#discussion_r159294869>
:

> @@ -49,6 +110,10 @@ while true; do
                 exit 1
             fi
         ;;
+        --json)
+          jsonFmt

I would like to see this controlled by an environment variable instead of
calling the function directly here. This way you can combine your function
with the code below and control whether it prints or constructs a json via
the env var.
------------------------------

In libexec/cli/instance.list.exec
<https://github.com/singularityware/singularity/pull/1191#discussion_r159729252>
:

> +  JSON=""
+  JSON_INIT='{"instances":['
+  JSON_END=']}'
+  COUNT=0
+
+
+  if [ -z "${1:-}" ]; then
+      DAEMONS=`singularity_daemon_glob '*' | sort | uniq`
+  else
+      DAEMONS=`singularity_daemon_glob "$@" | sort | uniq`
+  fi
+
+DLEN=0
+for i in $DAEMONS; do
+	let "DLEN++"
+done

I think you could do this in a one-liner if you echo $DAEMONS with quotes
to preserve newlines and use wc. But I'm not convinced this variable is
necessary at all. I think you could change your if statement in the for
loop below to look at the one-liner instead of using the variable. Might be
a slight bit less efficient but would increase readability imho.
------------------------------

In libexec/cli/instance.list.exec
<https://github.com/singularityware/singularity/pull/1191#discussion_r159738765>
:

> +          . "$i"
+          if [ -n "${DAEMON_PID:-}" ]; then
+              PROC_CMDLINE="/proc/${DAEMON_PID:-}/cmdline"
+              if [ -e "$PROC_CMDLINE" ]; then
+                  PROGNAME=$(<"${PROC_CMDLINE}" tr -d \\0)
+                  if [ "${PROGNAME}" != "singularity-instance: $USER
[$DAEMON_NAME]" ]; then
+                      message WARN "Removing stale daemon file: $i\n"
+                      rm -f "$i"
+                      continue
+                  fi
+              else
+                  rm -f "$i"
+                  continue
+              fi
+          fi
+

All of this code is basically just copied and pasted from the code below
with the one change of constructing a json instead of printing each line.
That is not going to be very sustainable if we have to change this
algorithm. Can you just wrap the if statement below in another if statement
so that it prints if the json environmental variable is not set and it
creates the json if the json env var is set?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/singularityware/singularity/pull/1191#pullrequestreview-86219900>,
or mute the thread

GodloveD:
Great!  Thanks @ArangoGutierrez!  Also, I'm no json expert, but I think there may be some issues with the json sytax as it is currently output.  

In particular I think you need to wrap the instance names and the image paths in quotes so that they are recognized as strings.  Also is it customary to add some `\n` and `\t` characters so that the json displays on multiple lines with proper indenting?   Maybe @vsoch can address this better.

ArangoGutierrez:
My bad, this PR was part of the singularity on nomad repo I created. I was
needing this functionality, I think this PR is too selfish made.(just for
my use case) I will rework it and enhance it.

On Jan 4, 2018 3:07 PM, "David Godlove" <notifications@github.com> wrote:

Great! Thanks @ArangoGutierrez <https://github.com/arangogutierrez>! Also,
I'm no json expert, but I think there may be some issues with the json
sytax as it is currently output.

In particular I think you need to wrap the instance names and the image
paths in quotes so that they are recognized as strings. Also is it
customary to add some \n and \t characters so that the json displays on
multiple lines with proper indenting? Maybe @vsoch
<https://github.com/vsoch> can address this better.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/singularityware/singularity/pull/1191#issuecomment-355385416>,
or mute the thread

GodloveD:
Oh, also, (sorry I keep thinking of things!) can you please update the CHANGELOG.md when you make these changes.  Thanks again!

vsoch:
Generally you'd want to deliver json in "print pretty" syntax, because otherwise it's terrible to make any sense of it.  I  usually do an intend of 4, and can give an example in python to show the difference! Here we have a fugly:

```
import json
content = {'items': [{'sub1':'1','sub2':2,'sub3':3},{'sub1':'1','sub2':2,'sub3':3}]}
```
If we print just as is, it's still fugly (albeit valid). This gets problematic for large arrays that might go over a line and produce a really hard file to interact with / open (at least for a human).

```
json.dumps(content)
Out[97]: '{"items": [{"sub1": "1", "sub2": 2, "sub3": 3}, {"sub1": "1", "sub2": 2, "sub3": 3}]}'
```

So instead we define separators for content and an indentation of 4, and we get something that is pretty to look at (and most inspection commands use, eg. look at docker inspect). Here is the same data structure:

```
print(json.dumps(content, indent=4, separators=(',', ': ')))
  
{
    "items": [
        {
            "sub1": "1",
            "sub2": 2,
            "sub3": 3
        },
        {
            "sub1": "1",
            "sub2": 2,
            "sub3": 3
        }
    ]
}
```

Note that without the print what we actually add are newlines:

```
son.dumps(content, indent=4, separators=(',', ': '))
'{\n    "items": [\n        {\n            "sub1": "1",\n            "sub2": 2,\n            "sub3": 3\n        },\n        {\n            "sub1": "1",\n            "sub2": 2,\n            "sub3": 3\n        }\n    ]\n}'
```
Take a look at the python functions in our suitils if needed.  Any particular reason to not use those?

https://github.com/singularityware/singularity/blob/master/libexec/python/sutils.py#L439

vsoch:
And @ArangoGutierrez here is my goto for checking if json is valid. It also formats nicely :)

https://jsonformatter.curiousconcept.com/

ArangoGutierrez:
Hi @GodloveD  take a look, and tell me if more work is needed. ;)

libexec/cli/instance.list.exec
```diff
@@ -33,6 +33,7 @@ if ! USERID=`id -ru`; then
     exit 255
 fi
 
+JSONFMT="false"
```
GodloveD: Instead of doing this we tend to do something like: 
```
"${JSONFMT:-}"
```
That way a user can set the env var and in their environment and control the format that way rather than passing the argument.

GodloveD: Also a good idea to prepend variables like this with `SINGULARITY_`.  After all, the user might want set this env var and if you export it, you want to make sure that it ends up in the registry for use in the c portion of the code.

libexec/cli/instance.list.exec
```diff
@@ -49,6 +50,10 @@ while true; do
                 exit 1
             fi
         ;;
+        --json)
+        JSONFMT="true"
```
GodloveD: Can you indent this please. 😄

libexec/cli/instance.list.exec
```diff
@@ -66,7 +71,14 @@ else
     DAEMONS=`singularity_daemon_glob "$@" | sort | uniq`
 fi
 
-message 1 "%-16s %-8s %s\n" "DAEMON NAME" "PID" "CONTAINER IMAGE"
+DLEN=`echo "$DAEMONS"|wc -l`
+let DLEN--
+
+if [[ "${JSONFMT}" == "true" ]]; then
```
GodloveD: In keeping with the conventions we use throughout the code, it would be good to change this to something like:
```
if [ -z "${SINGULARITY_JSONFMT:-}" ]; then
```
I think in most of our code, any value of an env var is taken as true.  It is only false if it is empty or not set.

libexec/cli/instance.list.exec
```diff
@@ -92,12 +104,24 @@ for i in $DAEMONS; do
             fi
         fi
 
-        printf "%-16s %-8s %s\n" "${FILE_NAME}" "${DAEMON_PID}" "${DAEMON_IMAGE:-}"
+        if [[ "${JSONFMT}" == "true" ]]; then
+          if [ "$COUNT" -lt "$DLEN" ]; then
```
GodloveD: I think we're using 4 indentations throughout right?

libexec/cli/instance.list.exec
```diff
@@ -92,12 +104,24 @@ for i in $DAEMONS; do
...
+        else
+            printf "%-16s %-8s %s\n" "${FILE_NAME}" "${DAEMON_PID}" "${DAEMON_IMAGE:-}"
+        fi
 
         let "COUNT++"
     fi
 done
 
+if [[ "${JSONFMT}" == "true" ]]; then
+printf "\n\t]\n}\n"
```
GodloveD: I can haz indent plz? 🐱

libexec/cli/instance.list.exec
```diff
@@ -92,12 +104,24 @@ for i in $DAEMONS; do
             fi
         fi
 
-        printf "%-16s %-8s %s\n" "${FILE_NAME}" "${DAEMON_PID}" "${DAEMON_IMAGE:-}"
+        if [[ "${JSONFMT}" == "true" ]]; then
+          if [ "$COUNT" -lt "$DLEN" ]; then
+            printf '\t{\n\t\t"instance":"%s",\n\t\t"pid":"%s",\n\t\t"img":"%s"\n\t},\n' "${FILE_NAME}" "${DAEMON_PID}" "${DAEMON_IMAGE:-}"
```
GodloveD: Would it be useful to save this big gnarly string in a variable so that you don't have to repeat it on 2 lines?

GodloveD (changes requested):
Hey @ArangoGutierrez.  This looks a lot better but I still think it needs a little more love. 

Also can you add the `--json` argument to the `instance.list` subcommand in [etc/bash_completion.d/singularity](etc/bash/completion.d/singularity)

vsoch:
hey @ArangoGutierrez ! A good way to think about this is extend-ability - there are a lot of times when something is going to need to take one or more key value pairs and generate json. Instead of a hard coded solution for this one time, we would want to have a general function / library that can handle this, and then the call to generate from instances.list would just use that. If there is a bug, it needs to be fixed in one place (and not a gazillion printf statements across a single file or different functions). Does that make sense? You got this!!

GodloveD (approved):
Hey @ArangoGutierrez, I took the liberty of making a few more changes to this myself.  I'm happy with this now.  If somebody else can take a look I think we're good to merge.  @carlm1234 do you want to give it a once over?

ctmadison:
yep

ctmadison (approved):

PR #1194, opened 2017-12-05
Make the build reproducible in case the SOURCE_DATE_EPOCH is known

yarikoptic:
To facilitate reproducible builds in Debian

See https://wiki.debian.org/ReproducibleBuilds

- [ ] I have added changes to the [CHANGELOG](https://github.com/singularityware/singularity/blob/development/CHANGELOG.md) and and documentation updates to the [singularityware](https://www.github.com/singularityware/singularityware.github.io) documentation base.  *Is it worth it for this one?*

Attn: @singularityware-admin

### DIFF
--- a/libexec/bootstrap-scripts/environment/Makefile.am
+++ b/libexec/bootstrap-scripts/environment/Makefile.am
@@ -41,7 +41,7 @@ environment.tar:
 	ln -sf .singularity.d/actions/run newroot/.run
 	ln -sf .singularity.d/actions/shell newroot/.shell
 	ln -sf .singularity.d/actions/test newroot/.test
-	cd newroot; tar czf ../environment.tar . --owner=0 --group=0
+	cd newroot; GZIP=-n tar czf ../environment.tar . --owner=0 --group=0 --sort=name --mtime="@${SOURCE_DATE_EPOCH:-$(date +%s)}"
 	rm -rf newroot
 
 MAINTAINERCLEANFILES = Makefile.in


libexec/bootstrap-scripts/environment/Makefile.am
```diff
@@ -41,7 +41,7 @@ environment.tar:
 	ln -sf .singularity.d/actions/run newroot/.run
 	ln -sf .singularity.d/actions/shell newroot/.shell
 	ln -sf .singularity.d/actions/test newroot/.test
-	cd newroot; tar czf ../environment.tar . --owner=0 --group=0
+	cd newroot; GZIP=-n tar czf ../environment.tar . --owner=0 --group=0 --sort=name --mtime="@${SOURCE_DATE_EPOCH:-$(date +%s)}"
```
yarikoptic, line 44: oh, apparently `--sort` is a "recent" addition to tar ... hmm

GodloveD (commented):
Hey @yarikoptic!  Sorry for the long delay on this review.  

Did you see https://github.com/singularityware/singularity/pull/1102 ?  I think it's a pretty similar PR.  @gmkurtzer suggested:
>I would like to see this check being done in configure.ac and then we export the correct tar arguments for all called tar commands (we are calling tar directly in several places).

Is that something you would like to try to do?

yarikoptic:
nope, didn't see #1102 ...  indeed should be done in configure.ac ... if nobody does and I still remember I will do.  not sure yet which one is better, so will keep this one open for now as a reminder ;)

c-ney:
See my comment https://github.com/singularityware/singularity/pull/1102#issuecomment-366633737.

GodloveD:
Hey @yarikoptic and @c-ney where does this PR stand?  Is this still worth looking into?

yarikoptic:
well, for debian builds we do patch along the lines suggested in this PR and implemented here: https://salsa.debian.org/hpc-team/singularity-container/blob/master/debian/patches/reproducible_build  

I am just waiting patiently for upstream to resolve it one way or another (or not, then we would just carry the patch forever ;) )

jmstover:
@yarikoptic does the following Makefile output look correct to you?

```
cd newroot; GZIP=-n tar czf ../environment.tar . --owner=0 --group=0 --sort=name --mtime="@1528224051"
```

The addition of sort and mtime comes from the `./configure` script. It will only add sort and mtime if the sort option is availible in tar, otherwise it falls back to previous behavior.

GodloveD:
Heya @jmstover thanks for having a look at this and trying to implement it at the configure stage.  Maybe it would be a good idea to open an new PR and then reference it here?

PR #1201, opened 2017-12-06
Support multiarch (32 and 64bits) in containlibs

hmlth:

The goal of this PR is to make containlibs work for 32 and 64 bits libraries at the same time. The use case for us is to make some old 32 bits binary only OpenGL applications work with the "--nv" option. In the current code the list of libraries is bound to `/.singularity.d/libs`, if a library is listed twice with two possible sources, it is bound only once.

With this PR, the x86_64 libraries is bound in `/.singularity.d/libs/x86_64` and i386 libraries are bound in `/.singularity.d/libs/i686`. The value of LD_LIBRARY_PATH is updated according to the used sub directories. If the architecture of a library can't be determined, it is bound to `/.singularity.d/libs` as before. This sorting only supports 'x86_64', 'i386' and 'x32', but other architecture works the same way as before and adding other architectures with multiarch should be easy.

To check for the library architecture, the PR relies on elfutils-libelf. This is added as a build dependency.

- Ref: #

**Checkoff for all PRs:**

### DIFF
--- a/src/util/binary.c
+++ b/src/util/binary.c
@@ -1,0 +1,72 @@
+/* 
+ * Copyright (c) 2017, EDF, SA. All rights reserved.
+ * 
+ * This software is licensed under a 3-clause BSD license.  Please
+ * consult LICENSE file distributed with the sources of this project regarding
+ * your rights to use or distribute this software.
+ * 
+ */
+
+
+#define _GNU_SOURCE
+#include <stdio.h>
+#include <errno.h>
+#include <libelf.h>
+
+#include "util/message.h"
+#include "util/binary.h"
+
+int singularity_binary_arch(char* path){
+    FILE* fp;
+    Elf* elf_file;
+    Elf32_Ehdr* hdr32 = NULL;
+    Elf64_Ehdr* hdr64 = NULL;
+
+    fp = fopen(path, "rb");
+    if (fp == NULL) {
+        singularity_message(WARNING, "Failed to open binary: %s (error=%d)\n", path, errno);
+        return(BINARY_ARCH_UNKNOWN);
+    }
+    if (elf_version(EV_CURRENT) == EV_NONE) {
+        singularity_message(WARNING, "Failed to initialize ELF library: %s\n", elf_errmsg(-1));
+        return(BINARY_ARCH_UNKNOWN);
+    }
+    elf_file = elf_begin(fileno(fp), ELF_C_READ, NULL);
+    if (elf_file == NULL) {
+        singularity_message(DEBUG, "Failed initialize elf parsing on file %s: %s\n", path, elf_errmsg(-1));
+        // Not an ELF BINARY
+        return(BINARY_ARCH_UNKNOWN);
+    }
+    hdr32 = elf32_getehdr(elf_file);
+    if (hdr32 == NULL) {
+        hdr64 = elf64_getehdr(elf_file);
+    }
+    fclose(fp);
+    if (hdr32 == NULL) {
+        if (hdr64 == NULL) {
+            singularity_message(DEBUG, "No ELF headers on binary file %s: (%s)\n", path, elf_errmsg(-1));
+            // Not an ELF BINARY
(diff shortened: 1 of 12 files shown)


GodloveD (changes requested):
Hi @hmlth.  Thanks very much for identifying this issue and submitting a PR!

I think it is a great idea to support both 64 and 32-bit versions of the NVIDIA libraries.  Let me suggest some changes to this implementation. 

My biggest concern is that I actually don’t think this is going to work as expected.  afaik, the LD_LIBRARY_PATH works identically to the PATH variable.  If there are two identically named libraries (one for 32-bit architecture and one for 64-bit arch), I think that this PR will ensure that all programs use the 32-bit version of the library because that is what comes first on the LD_LIBRARY_PATH.  I think what we really want to do is update the `/etc/ld.so.cache`. 

My second concern is that this is introducing a new dependency on `libelf`.  I don’t think that is really necessary.  

Let me propose the following algorithm instead:

First, when identifying libraries on the host system and creating dummy files in the container, why don’t we create the same directory structure in the container as exists on the host system.  Just replace the root with `/.singularity.d/libs/`.  That way you will get all of the libraries compiled for all the architectures, and you don’t need to use libelf to parse headers.  

Second, add all the new libs to `/etc/ld.so.conf` and call `ldconfig` (or do the c equivalent) to add all of the entries to `/etc/ld.so.cache`.  I’m pretty sure the `ld.so.cache` has info about arch, so the linker should be able to figure out which library goes with which binary.  

What do you think about trying that instead?  Thanks again for your time and effort @hmlth!

src/util/binary.c
```diff
@@ -0,0 +1,72 @@
+/*
```
yhcote, line 1: Hi there,

I think you can rewrite your singularity_binary_arch() function like this, so that the only requirement is to include the system include file 'elf.h' and avoid adding a new dependency 'libelf-dev'. This would shorten and simplify this patch quite a bit.
 
```
#include <sys/mman.h>
#include <sys/types.h>
#include <sys/stat.h>

#include <elf.h>
#include <fcntl.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <unistd.h>

#define BINARY_SYSTEM_ERROR -1

int
singularity_binary_arch(char *path)
{
	int fd, ret = BINARY_ARCH_UNKNOWN;
	char *elf;
	struct stat st;

	/* prepare file */
	if(path == NULL){
		fprintf(stderr, "Error invalid filename\n");
		return -1;
	}
	fd = open(path, O_RDONLY);
	if(fd < 0){
		perror("Error while opening file");
		return -1;
	}
	if(fstat(fd, &st) < 0){
		perror("Error calling fstat()");
		close(fd);
		return -1;
	}
	elf = mmap(NULL, st.st_size, PROT_READ, MAP_PRIVATE, fd, 0);
	if(elf == MAP_FAILED){
		perror("Error mapping file");
		close(fd);
		return -1;
	}

	/* file is ready for inspection */
	if(memcmp(elf, ELFMAG, strlen(ELFMAG)) != 0){
		fprintf(stderr, "Not an ELF file\n");
	}else if(elf[EI_CLASS] == ELFCLASS32){
		Elf32_Ehdr *hdr = elf;
		if(hdr->e_machine == EM_386)
			ret = BINARY_ARCH_I386;
		if(hdr->e_machine == EM_X86_64)
			ret = BINARY_ARCH_X32;
	}else if(elf[EI_CLASS] == ELFCLASS64){
		Elf64_Ehdr *hdr = elf;
		if(hdr->e_machine == EM_X86_64)
			ret = BINARY_ARCH_X86_64;
	}

	/* cleanup */
	munmap(elf, st.st_size);
	close(fd);

	return ret;
}
```

yhcote (changes requested):
See the embedded comment / example.

Cheers !

GodloveD:
@hmlth haven't heard back from you about any of these proposed changes.  Is this something that you would still like to work on?

Issue #1203, opened 2017-12-06
Missing dependencies on libuuid-devel, openssl-devel


olifre:
## Version of Singularity:

2.4.2

### Expected behavior

It compile using rpmbuild / debuild. 

### Actual behavior

Things break, dependencies are incomplete. 

### Steps to reproduce behavior

Use rpmbuild / debuild.

Apparently, https://github.com/singularityware/singularity/pull/1143 did not see any review at all - it adds dependencies on openssl-devel and libuuid-devel, but updates neither the spec-file nor the debian packaging. 
As a result, both are broken in all subsequent releases. 

Maybe a basic review procedure (as other projects have) and basic testing of the packaging code is due, or the fully unmaintained packaging code should be dropped from this repo.

jtriley:
Already being addressed: https://github.com/singularityware/singularity/pull/1198

olifre:
@jtriley Thanks for pointing that out!
Maybe there should also be something added to the release procedure (i.e. "test package building in clean environment before release) and travis tests for the packaging code being added? 
Otherwise, since this has now happened numerous times, I fear it will happen again and again. 

In any case, a basic review before merging things would also have prevented that.

olifre:
Also, even after fixing that, I still get:
```
Error building RPM:
    File not found: /root/rpmbuild/BUILDROOT/singularity-2.4.2-1.el7.cern.x86_64/usr/libexec/singularity/bin/prepheader
    File not found: /root/rpmbuild/BUILDROOT/singularity-2.4.2-1.el7.cern.x86_64/usr/libexec/singularity/bin/mount-suid
```

olifre:
Pinging @gmkurtzer as project leader since I am asking for:
- a change to PR merge policy (add basic review procedure)
- adding basic packaging tests
- potentially extending the release checklist (re-check packages can actually be build in a clean environment). 

Would be nice to see this discussed ;-).

GodloveD:
Hey @olifre I'm a little bit confused.  You are saying this is a problem with 2.4.2?   2.4.2 has nothing to do with #1143.  That is a SIF feature and should only be in development at this point in time.

olifre:
@GodloveD You are right! Yes, I tried to build from the development branch. 
So unlike the previous packaging issues which actually made it into releases, at least this has not happened (yet) for this one.

GodloveD:
[...]

PR #1198, opened 2017-12-06
Add missing build deps to deb and RPM: uuid and ssl

hmlth:
Build fails in a clean env (mock or cowbuilder) without it.

**Description of the Pull Request (PR):**

Since the commit https://github.com/singularityware/singularity/commit/de6a020105713a49079a2c56705dd3e43c152579, It is not possible to build a RPM or a deb package in a clean environment, the configure fails if those libraries are not present.

This was tested on Debian Jesssie and RHEL 6. Not sure if the packages have the same names on Suse.

**This fixes or addresses the following GitHub issues:**

- Ref: #

**Checkoff for all PRs:**

- [X] I have read the [Guidelines for Contributing](https://github.com/singularityware/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- [ ] I have added changes to the [CHANGELOG](https://github.com/singularityware/singularity/blob/development/CHANGELOG.md) and and documentation updates to the [singularityware](https://www.github.com/singularityware/singularityware.github.io) documentation base.
- [ ] I have tested this PR locally with a `make test`
- [X] This PR is NOT against the project's master branch
- [X] I have added myself as a contributor to the [contributors's file](https://github.com/singularityware/singularity/blob/master/CONTRIBUTORS.md)
- [X] This PR is ready for review and/or merge

Attn: @singularityware-admin

Issue #1074, opened 2017-10-17
Opensuse 42.3: WARNING: Could not create bind point file in container /etc/localtime: No such file or directory


wandadars:
## Version of Singularity:

2.3.1

### Expected behavior

Run a Docker image converted to a Singularity image

### Actual behavior

singularity run locistream_stream_v206_opensuse423_stream_v206-2017-08-15-d02bd8f254d5.img 
WARNING: Could not create bind point file in container /etc/localtime: No such file or directory

**NOTE: The /etc/localtime path is symbolic both in the Docker image and on my machine(both are OpenSuse 42.3).

### Steps to reproduce behavior
My Dockerfile:
FROM opensuse:42.3

RUN zypper update -y &&   \
    zypper install -y     \
    sudo                  \
    vim                   \
    vim-data              \
    gcc-fortran           \
    infinipath-psm-devel  \
    openmpi               \
    openmpi-devel         \
    openmpi-libs          \
    hdf5-openmpi          \
    blas-devel            \
    blas-devel-static     \
    lapack-devel          \
    iproute2              \
    curl                  \
    tar                   \
    which

USER root
RUN echo "root:codeuser_2017" | chpasswd      && \
    groupadd -g 1000 codeuser                 && \
    useradd -u 1000 -m -s /bin/bash codeuser  && \
    passwd -d codeuser

RUN mkdir -p -m0755       \
    /opt/MyDir/lib         \

USER codeuser:codeuser

I ran:
docker build -t codebase/code_v203_opensuse423:code_v203

Then I ran:
docker run                                                          \
            -v /var/run/docker.sock:/var/run/docker.sock            \
            -v /home/user/source/temp/docker:/output    \
            --privileged -t --rm                                   \
            singularityware/docker2singularity                      \
            codebase/code_v203_opensuse423:code_v203

Then I ran:
singularity run   <generated.img>

### System Info

uname -r
4.4.79-19-default

lsb_release -a
LSB Version:	n/a
Distributor ID:	openSUSE project
Description:	openSUSE Leap 42.3
Release:	42.3
Codename:	n/a

wandadars:
Seems that the Singularity container boots up when I used:

singularity exec -H /home2/MyUser/ locistream_stream_v206_opensuse423_stream_v206-2017-10-17-ea4f8c62ccf0.img /bin/bash

So the 'run' command doesn't drop me into the container. I noticed a similar behavior when running a Docker container on Linux.

[...]

PR #1204, opened 2017-12-06
travis: Fixup rpmbuild test to actually test the dependencies, add Debian testing, test distclean.

olifre:

This fixes the packaging test for RedHat systems, by actually testing the dependencies listed in the spec-file. 

- Ref: #1203 

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/.travis/rpmbuild_test
+++ b/.travis/rpmbuild_test
@@ -28,7 +28,7 @@ yum install -y libtool sudo which e2fsprogs
 sed -i 's,^prefix=.*,prefix=/usr,' test.sh
 sed -i 's,^sysconfdir=.*,sysconfdir=/etc,' test.sh
 sed -i 's,^localstatedir=.*,localstatedir=/var,' test.sh
-useradd testuser
+useradd --create-home testuser
 echo "Defaults:testuser env_keep=DOCKER_HOST" >>/etc/sudoers
 echo "testuser ALL=(ALL) NOPASSWD: ALL" >>/etc/sudoers
 
--- a/.travis/rpmbuild_test
+++ b/.travis/rpmbuild_test
@@ -5,13 +5,13 @@
 OS_VERSION="$1"
 
 # build and install
-yum install -y libtool rpm-build make yum-utils squashfs-tools
+yum install -y libtool rpm-build make yum-utils
 ./autogen.sh
 yum-builddep -y singularity.spec
 ./configure
 make dist
 rpmbuild -ta *.tar.gz
-rpm -iv ~/rpmbuild/RPMS/*/*.rpm
+yum install -y ~/rpmbuild/RPMS/*/*.rpm
 
 if [ "$OS_VERSION" = 6 ]; then
     echo "the test suite has not yet been ported to centos6 python, skipping."
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,10 @@ matrix:
       env: OS_TYPE=centos OS_VERSION=6
     - os: linux
       env: OS_TYPE=centos OS_VERSION=7
+    - os: linux
+      env: OS_TYPE=debian OS_VERSION=8
+    - os: linux
+      env: OS_TYPE=debian OS_VERSION=9
 
 before_install:
   - .travis/before_install
--- a/.travis/debian_run
+++ b/.travis/debian_run
@@ -1,0 +1,17 @@
+#!/bin/bash -ex
+
+DIST="${OS_TYPE}${OS_VERSION}"
+
+mkdir /tmp/$DIST
+
+cp -a . /tmp/$DIST/singularity
+cat > "/tmp/$DIST/script" << SCRIPT
+#!/bin/bash
+.travis/debuild_test $OS_VERSION
+SCRIPT
+
+cd /tmp/$DIST/singularity
+
+# mounting /run for /dev/shm (python multiprocessing)
+sudo singularity exec -w -B /run --keep-privs --allow-setuid docker://${OS_TYPE}:${OS_VERSION} /bin/bash -ex /tmp/$DIST/script
+sudo rm -rf /tmp/$DIST
--- a/.travis/debuild_test
+++ b/.travis/debuild_test
@@ -46,7 +46,7 @@ apt-get -y install libtool sudo e2fsprogs
 sed -i 's,^prefix=.*,prefix=/usr,' test.sh
 sed -i 's,^sysconfdir=.*,sysconfdir=/etc,' test.sh
 sed -i 's,^localstatedir=.*,localstatedir=/var,' test.sh
-useradd testuser
+useradd --create-home testuser
 echo "Defaults:testuser env_keep=DOCKER_HOST" >>/etc/sudoers
 echo "testuser ALL=(ALL) NOPASSWD: ALL" >>/etc/sudoers
 
--- a/.travis/debuild_test
+++ b/.travis/debuild_test
@@ -1,0 +1,59 @@
+#!/bin/bash -ex
+
+# this script runs as root under docker 
+
+OS_VERSION="$1"
+
+# Prepare for build
+export DEBIAN_FRONTEND=noninteractive
+export LC_ALL=C
+apt-get update
+apt-get -y dist-upgrade
+apt-get -y install devscripts build-essential lintian gdebi-core git
+
+DEB_VERS=$(LC_ALL=C dpkg-parsechangelog | grep '^Version:' | cut -d ' ' -f 2,2)
+PLAIN_VERS=$(echo ${DEB_VERS} | sed 's#-.*##')
+git archive --format=tar HEAD | gzip > ../singularity-container_${PLAIN_VERS}.orig.tar.gz
+
+# Install dependencies
+mk-build-deps
+mv singularity-container-build-deps_*.deb ..
+gdebi -n ../singularity-container-build-deps_*.deb
+
+# Clean up directory
+make maintainer-clean
+
+# Build package
+debuild -us -uc
+
+echo "Directory containing the build .deb-files:"
+ls -la ..
+gdebi -n ../singularity-container_*.deb
+
+# Install python 3 and pylint
+apt-get -y install python3 python3-pip python3-setuptools
+apt-get -y install pylint
+
+# Install docker
+apt-get -y install apt-transport-https ca-certificates curl gnupg2 software-properties-common
+curl -fsSL https://download.docker.com/linux/$(. /etc/os-release; echo "$ID")/gpg | apt-key add -
+add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/$(. /etc/os-release; echo "$ID") $(lsb_release -cs) stable"
+apt-get update
+apt-get -y install docker-ce
+
+# run the test suite
+apt-get -y install libtool sudo e2fsprogs
+sed -i 's,^prefix=.*,prefix=/usr,' test.sh
+sed -i 's,^sysconfdir=.*,sysconfdir=/etc,' test.sh
+sed -i 's,^localstatedir=.*,localstatedir=/var,' test.sh
+useradd testuser
+echo "Defaults:testuser env_keep=DOCKER_HOST" >>/etc/sudoers
+echo "testuser ALL=(ALL) NOPASSWD: ALL" >>/etc/sudoers
+
+# build image for isolated builds
+sed -i 's,^prefix=.*,prefix=/usr,' secbuildimg.sh
+sed -i 's,^sysconfdir=.*,sysconfdir=/etc,' secbuildimg.sh
+sed -i 's,^localstatedir=.*,localstatedir=/var,' secbuildimg.sh
+make secbuildimg
+
+su testuser -c "make test"
--- a/.travis/install_test
+++ b/.travis/install_test
@@ -1,6 +1,11 @@
 #!/bin/bash -ex
 
 ./autogen.sh
+
+# Test that distclean target works fine:
+./configure --prefix=/usr/local
+make distclean
+
 ./configure --prefix=/usr/local
 sudo make install
 sudo make secbuildimg
--- a/.travis/script
+++ b/.travis/script
@@ -3,5 +3,6 @@
 case $OS_TYPE in
     "") .travis/install_test ;;
     centos) .travis/centos_run ;;
+    debian) .travis/debian_run ;;
     *) echo "Unknown OS_TYPE $OS_TYPE"; exit 1;;
 esac


olifre:
Just to make sure this is clear: It's good that the tests fail, since this uncovers that #1198 is needed.

olifre:
I have also added tests for Debian 8 and 9, both failing as of now, and extended the `install-test` test to also check whether the `distclean` target works. 
This also fails expectedly.

GodloveD:
Hi @olifre!  Sorry for the long delay.  We finally got two reviews on https://github.com/singularityware/singularity/pull/1198 and got it merged.  When you are able, would you mind updating this with those changes so that we can make sure it passes Travis.  If it does I'll review it and ask another team member to look it over as well.  Thanks!

olifre:
Hi @GodloveD ! 
Many thanks for coming back to this, delays to take time for review for sure a good thing :smile:. 
I've rebased and will check the output from travis in detail. 
For sure, it's still failing due to #1206 .

jmstover:
@olifre Once #1246 is merged, it _should_ fix that error you're seeing on `make distclean` ... I hope.

olifre:
@jmstover See my comments in the other issue for details on the issues I observed when cherry-picking your changes. 

I re-pushed my commits just now since I was missing "git" for the debian testing (which is required to run "git archive").

olifre:
@jmstover Your changes appear to do the trick now, I can just call `make maintainer-clean` and debuild is happy. 
Now I still have to sort out getting the tests to execute, and then this is ready.

olifre:
Done. Packaging now works fine after applying the patches from #1246 . 
Tests still fail, though, seems the test suite has issues on Debian:
https://api.travis-ci.org/v3/job/328572899/log.txt
```
+ python2 -m unittest tests.test_docker_import                                                  (retval=1) ERROR
F
======================================================================
FAIL: test_IMPORT (tests.test_docker_import.TestImport)
test_import will test the IMPORT USER function
----------------------------------------------------------------------
Traceback (most recent call last):
  File "tests/test_docker_import.py", line 75, in test_IMPORT
    self.assertEqual(result['return_code'], 0)
AssertionError: 1 != 0
```
This appears to be a problem with the test suite, though. Sadly, output is not very informative to allow for clean debugging. @GodloveD Any idea what is going on there?

c-ney (commented):
I think `yum-builddep` doesn't pull `squashfs-tools` in because it is a `Requires` in the .spec file, not a `BuildRequires`. For runtime dependencies I suggest just using `yum install -y ~/rpmbuild/RPMS/*/*.rpm` instead of `rpm -iv`.

Edit: This should've been a comment/"review" on the specific commit only, guess the button doesn't work like that.

olifre:
> I think yum-builddep doesn't pull squashfs-tools in because it is a Requires in the .spec file, not a BuildRequires. For runtime dependencies I suggest just using yum install -y ~/rpmbuild/RPMS/*/*.rpm instead of rpm -iv.

That's what I also thought at the beginning, but apparently singularity requires mksquashfs during build:  #1246 
I'll ask a question there whether this is really the case or there's just a misunderstanding of `BuildRequires` / `Requires` on the core developers end. At least, if it's really a build requirement, the spec file should also use `BuildRequires`. 

Concerning `yum builddep`, if `squashfs-tools` was a `BuildRequires` it would still not install it due to:
https://bugzilla.redhat.com/show_bug.cgi?id=871512

olifre:
@c-ney As @jmstover has clarified in #1246 , `squashfs-tools` is not really a build dependency - thanks again for making me aware of this! 
I have dropped the explicit installation of `squashfs-tools` again from my PR here and also changed from simple `rpm` to install to `yum` as you suggested. 

Of course, Debian testing still fails (would be nice to get some input from @GodloveD or another expert of the Singularity test suite on that) and this PR still depends on #1246 (with the broken check for `mksquashfs` removed).

olifre:
I've rebased onto current development head. 
Let's see how travis does, I expect debian should still fail (due to the not-yet-understood test-suite issue) but the rest should work now.

olifre:
Yep, that went as expected. Anybody (@GodloveD  ? ) has a good idea on what is wrong with the test-suite on Debian? 
https://api.travis-ci.org/v3/job/329985807/log.txt 
Or how to make the test-suite somewhat more informative than "things have failed"?

GodloveD:
Hey @olifre for some reason it is failing on python unit tests.  No idea why.  This is the command that it failed to run:
```
python2 -m unittest tests.test_docker_import
```
I'll try to see if I can get it to repeat the test failures in a VM.  Does this spark any ideas for you?

olifre:
> Does this spark any ideas for you?

No - this matches what I could gather from the log. 
I'm really curious though why the test-suite is not outputting the actual error, but is so unhelpful. 
I only see `Full output in: /tmp/stest.ZwyjqT`, which does not help so much on travis. 

I guess to reproduce, also a Debian container should be sufficient. After all, this happens after entering a Debian container with a bootstrapped Singularity only after building and installing Singularity there.

c-ney:
The lack of debug output is due to the subprocess' stdout being "piped" into the Popen.communicate() result tuple [1]. In the unittest definition it looks like this:

```
output = Popen(testing_command, stderr=STDOUT, stdout=PIPE)
t = output.communicate()[0], output.returncode
```

Now one can either `print(t[0])` or just remove the `stdout=PIPE` kwarg altogether.

I have my own hacky rpmbuild test case for opensuse and I could then see

`ERROR Error creating path /home/testuser/.singularity/docker, exiting.`

So, the error (for my opensuse run) was the `useradd`  inside the `opensuse:42.3` docker container not creating a home directory by default . Thus, the fix is `useradd --create-home testuser` in .travis/rpmbuild.

Now I have a boatload of undefined variable errors in the pylint test, but that's for another day.

```
 + pylint /tmp/opensuse42.3/singularity/libexec/python --errors-only --ignore tests --disable    (retval=2) ERROR
************* Module python.shell
E: 48,12: Undefined variable 'image' (undefined-variable)
E: 49,46: Undefined variable 'image' (undefined-variable)
E: 51,11: Undefined variable 'match' (undefined-variable)
E: 52,15: Undefined variable 'quiet' (undefined-variable)
E: 53,59: Undefined variable 'image' (undefined-variable)
E: 55,20: Undefined variable 'match' (undefined-variable)
E: 56,15: Undefined variable 'quiet' (undefined-variable)
...
```

[1] https://docs.python.org/2/library/subprocess.html#subprocess.Popen.communicate

Edit: I checked the debian 9 docker image and it's the same, no home by default.

olifre:
@c-ney Thanks for looking into the matter! 
I have added `--create-home`, now the debian tests end at:
https://api.travis-ci.org/v3/job/331100289/log.txt
i.e. `E: 72, 4: Use of super on an old style class (super-on-old-class)`. 

@GodloveD It might be good to fix the test-suite to be more informative in case of error in general, i.e. `print(t[0])` in case of failure, as @c-ney figured out. Otherwise, it's pretty useless in travis. 

I'll fix the conflict here tomorrow, it was caused by changes in #1181 which were both wrong and completely off-topic to #1181 .

olifre:
Interesting. The Debian 9 tests work now, only Debian 8 fails. Is anybody actually using singularity on Debian 8 - is it supported? Otherwise, I may as well drop it from the tests - something like openSUSE could be of more interest in the tests.

c-ney:
This might very well be related to the version of `pylint` available in the debian 8 repository. At least for opensuse leap 42.3, `pylint3` is stuck at `1.4.4`. According to [1], a bump to `1.4.5` fixes the `undefined-variable` report. I'm currently working around this by adding `E0602` to the disable arg of pylint in `tests/05-python-units.sh`.

For debian 8, `pylint` is still on version `1.3.1`.

[1] https://stackoverflow.com/a/35986139

GodloveD:
It would be good to know why the tests started to work again on Debian 9 after the last commit.  If I try to manually run the command that results in an error during the Debian 9 testing I get the following output:
```
$ pylint /home/david/singularity/libexec/python --errors-only --ignore tests --disable=E0401
No config file found, using default configuration
************* Module python.base
E: 22, 4: No name 'parse' in module 'urllib' (no-name-in-module)
E: 23, 4: No name 'request' in module 'urllib' (no-name-in-module)
E: 24, 4: No name 'error' in module 'urllib' (no-name-in-module)
************* Module python.docker.api
E: 63, 4: No name 'error' in module 'urllib' (no-name-in-module)
E: 72, 4: Use of super on an old style class (super-on-old-class)
E:190,19: Instance of 'DockerApiConnection' has no 'get' member (no-member)
E:198,12: Instance of 'DockerApiConnection' has no 'update_headers' member (no-member)
E:269,19: Instance of 'DockerApiConnection' has no 'get' member (no-member)
E:311,18: Instance of 'DockerApiConnection' has no 'headers' member (no-member)
E:315,19: Instance of 'DockerApiConnection' has no 'get' member (no-member)
E:418,23: Instance of 'DockerApiConnection' has no 'download_atomically' member (no-member)
************* Module python.shub.api
E: 60, 4: No name 'parse' in module 'urllib' (no-name-in-module)
E: 67, 4: Use of super on an old style class (super-on-old-class)
E:113,19: Instance of 'SingularityApiConnection' has no 'get' member (no-member)
E:173,21: Instance of 'SingularityApiConnection' has no 'download_atomically' member (no-member)
```

GodloveD:
Maybe @vsoch might have a clue about this.

vsoch:
The combination of the version of pylint checking against the installed modules (I think) looks like a different version or a mismatch - urllib is one of those libraries that was hugely edited between python 2 and 3, so that's why you see things like:

```
try:
    from urllib.error import HTTPError
except ImportError:
    from urllib2 import HTTPError
```
pylint is a static parser, so it's just seeing one or the other of the lines and then spitting out the bug because it can't find the module. I would say our best bet is to disable those particular checks (the code starting with E) you add to the command via `--disable`. If there is a change between distribution it's likely just the python.

vsoch:
But +1 what @c-ney says, this is definitely related to pylint version (and different behavior between them). Before mucking with the command I would look into that.

c-ney:
@GodloveD: The pylint command is cut off in the travis log. The full one from `tests/05-python-units.sh` is

```
stest 0 pylint $PWD --errors-only --ignore tests  --disable=E0401,E0611,E1101
```

This actually disables the `no-name-in-module` (E0611) and `no-member` (E1101) errors you are seeing. E0401 is for `import-error`.

GodloveD:
Oh duh!  Thanks for the heads up @c-ney.  Yes this is what I get when I run the _complete_ command manually.  
```
$ pylint /home/david/singularity/libexec/python --errors-only --ignore tests  --disable=E0401,E0611,E1101
No config file found, using default configuration
************* Module python.docker.api
E: 72, 4: Use of super on an old style class (super-on-old-class)
************* Module python.shub.api
E: 67, 4: Use of super on an old style class (super-on-old-class)
```

olifre:
Rebased and resolved conflicts. 

Let me know what you prefer:
1. Drop Debian 8 from testing, since it appears the test-suite requires a more modern version of `pylint`. Fixing the test-suite to also work with older versions of `pylint` could go in a separate PR. 
2. Try to fix the test-suite here, but it seems @c-ney is much closer to success already, and also working on testing openSUSE, which would be nice to have. 

@c-ney I could also cherry-pick commits from you on top of this PR if you like (this will retain authorship).

olifre:
Alternatively, we may also proceed as is done for CentOS 6 for now:
```
if [ "$OS_VERSION" = 6 ]; then
    echo "the test suite has not yet been ported to centos6 python, skipping."
    exit
fi
```
However, maybe it's best to also fix this, since the largest userbase is likely to be using CentOS 6 / SL 6.

GodloveD:
Hi @olifre.  I agree that it would be better to fix the test for CentOS 6 if possible.  I wonder if it is skipping the whole test suite just because of pylint?  Would it be possible to just skip the pylint test?

And as far as that goes, the pylint tests only need to be done on a single platform right?  It's not like pylint would give us a different result (at least not one that we care about) when run using two different OS-es.

c-ney:
There is

```
if which pylint >/dev/null 2>&1; then
    stest 0 pylint $PWD --errors-only --ignore tests  --disable=E0401,E0611,E1101
else
    echo "Skipping pylint tests: not installed"
fi
```

in `tests/05-python-units.sh`. One could just not install `pylint` for the affected platforms.

Maybe it would be a good idea to define package dependencies for the various platforms centrally and then pass it into the scripts. See https://github.com/c-ney/singularity/blob/8ac9730d8931b575a152af3f1daf59ef80fe94e1/.travis/rpmbuild_test for a POC (ignore the messy commits).

olifre:
> One could just not install pylint for the affected platforms.

Good idea! 



> for a POC (ignore the messy commits).

This looks like a good plan. Maybe then it's a good idea to base your commits (when finished and squashed a bit) on top of my PR branch? Then they could make a second PR, or I could cherry pick them into this PR here?

c-ney:
@olifre Yes, we could do that! I'll wait for some feedback, maybe there is a nice way to split this up a bit. Not really fond of the `if`s plastered all over the place.

olifre:
> Not really fond of the ifs plastered all over the place.

One idea could be to have a separate file per distro, only setting the variables which are different, and sourcing that in `rpmbuild_test`. 
This could be done in the same way for `debuild_test`, where it would source `debian_8_deps` and `debian_9_deps` files, while `debian_8_deps` would not have `pylint`in `install_test_reqs` to effectively skip the `pylint` tests. It would also allow to have a generic `debian_deps` file, which is sourced in `debian_8_deps` and `debian_9_deps` so those only have to override what is different. 

So `debuild_test` and `rpmbuild_test` would contain something like:
```
DEP_FILE="${OS_TYPE}_${OS_VERSION}_deps"
if [ -e ${DEP_FILE} ]; then
  source ${DEP_FILE}
else
  echo "OS ${OS_TYPE}, version ${OS_VERSION} not supported yet, please create ${DEP_FILE}!"
  exit 1
fi
```
and all other `if`'s could be dropped. 

This would be one idea, maybe there are other better ones ;-).

c-ney:
@olifre Sounds good, will put something together.

@vsoch For the ubuntu build, I just noticed that `pylint` is installed via pip in `.travis/before_install`
```
# Install updated version of pylint
sudo pip install pylint
```
Could this be a solution for the other distributions as well?

vsoch:
It's worth a shot, for the one that is working I would do a pip freeze to see what version is being used, like

```
pip freeze | grep pylint
```

c-ney:
This is my current version of the splitting:

https://github.com/singularityware/singularity/compare/development...c-ney:travis-refactor

The default and centos targets finish correctly.

The opensuse 42.3 test is currently broken at `20-build.sh`. Every `singularity build` test without `sudo` fails with `retval=127`. No problem when the test command is changed to use `sudo singularity build`. I'd have to look some more into this to find out what it actually is that couldn't be found. The test will then fail because of pax format tar archives not being identified/wrongly identified as local FS images (fix is in #1262). There is some breakage in `21-build_docker.sh` with connecting to the docker daemon (only tested locally). `39-deffile.sh`  will fail because of this warning in `actual.txt` (cf. #1074):
```
WARNING: Non existent 'bind path' source: '/etc/localtime'
```
The next (and last) one failing is `45-targetmode.sh`, because opensuse does not like GID <500.

Not sure whether to omit the opensuse test completely from this merge request and make an exclusive one later, or just don't enable it in `.travis.yml` and then make a new request for enabling it with fixes included.

@olifre Now I only have to figure out how to rebase this on your changes.

@vsoch `pylint` is installed in version `1.8.1` with `pip`. Using this, there is no problem on both centos 7 and opensuse 42.3.

@GodloveD I looked into this (just commented out the `if`-clause) and it breaks (at least for me locally) with something dbus, /dev/urandom not found, dpkg during the setup in `make secbuildimg`.

vsoch:
@c-ney great news about pylint!

.travis/rpmbuild_test
```diff
@@ -5,13 +5,13 @@
 OS_VERSION="$1"
 
 # build and install
-yum install -y libtool rpm-build make yum-utils squashfs-tools
+yum install -y libtool rpm-build make yum-utils
 ./autogen.sh
 yum-builddep -y singularity.spec
```
jmstover, line 10: This won't work... singularity.spec is not generated until after `./configure` runs.

If you specified `yum-builddep -y singularity.spec.in` would that grab the BuildRequires entries, or would it break because it isn't "complete"?

.travis/rpmbuild_test
```diff
@@ -5,13 +5,13 @@
...
 ./configure
 make dist
 rpmbuild -ta *.tar.gz
-rpm -iv ~/rpmbuild/RPMS/*/*.rpm
+yum install -y ~/rpmbuild/RPMS/*/*.rpm
```
jmstover, line 14: IMO, doing:  `yum localinstall -y ~/rpmbuild/RPMS/*/*.rpm` is better for an on disk location.

dtrudg:
I did a separate quick and dirty batch of test related work on #1296 to make Centos6 test in Travis, except the Python stuff. Some of the things there may be relevant. E.g. we have to get `mount` into the centos6 docker container (it's not there in the official centos6 docker image) etc.

.travis/rpmbuild_test
olifre: Good point!
Any idea what's the magic trick why it still works in travis nevertheless? 
Here's the direct link to the build log:
https://api.travis-ci.org/v3/job/331238663/log.txt
Just search for `yum-builddep` and you will arrive at the place which ought not work...

.travis/rpmbuild_test
olifre: Indeed! 
I would first wait for @c-ney who is basing his work on this PR if I understood correctly,
so I would not apply this change now since it may cause conflicts for him.
But this for sure should be changed to `localinstall` before the final version, thanks for the review! 👍

.travis/rpmbuild_test
jmstover: Don't know why that succeeds. There was a build earlier, where `yum-builddep` blew up because `singularity.spec` didn't exist... and it actually _shouldn't_ exist at that point...

Where is the singularity source being grabbed from? I don't see a checkout or tarball extraction prior to the `./autogen.sh` run. A release tar extraction should work, as that would be packaged with a singularity.spec file...  But not if you're building directly from a clean git pull.

dtrudg: @jmstover see `.travis/centos_run` the source is bound into the docker container from `pwd` where it will be in the travis build environment:

```
docker run --privileged -d -ti -e "container=docker" -v /var/run/docker.sock:/docker.sock --network=host \
       -v /sys/fs/cgroup:/sys/fs/cgroup -v `pwd`:/build:rw  centos:${OS_VERSION}   /usr/sbin/init
```

.travis/rpmbuild_test
c-ney: From what I get from the man page of yum, there is no difference between `yum install file.rpm` and `yum localinstall file.rpm` [1]. Also the latter is _kinda_ marked as deprecated. But thinking about it, I did not check this in centos 6. The version there might not be new enough.

> If the name is a file, then install works like localinstall.

> Note that the install command will do a local install, if given a filename. This command is maintained for legacy reasons only.

[1] http://man7.org/linux/man-pages/man8/yum.8.html

.travis/rpmbuild_test
c-ney: At least `rpmspec` doesn't like "incomplete" spec files. I'd think that something similar would happen in `yum-builddep`?

```
rpmspec -q --buildrequires --qf "[%{REQUIRENAME} ]" singularity.spec.in
error: line 28: Illegal char '@' (0x40) in: Version: @PACKAGE_VERSION@
error: query of specfile singularity.spec.in failed, can't parse
```

So going back to manual selection of the dependencies for the configure step?

olifre: > So going back to manual selection of the dependencies for the configure step?

This has already gone wrong several times and will go wrong many times in the future, so I would strongly vote against it.

c-ney: Then for now one could just

```
sed 's/@PACKAGE_VERSION@/0/' < singularity.spec.in > singularity_tmp.spec
```

as the spec file is otherwise left alone by autotools.

Concerning the unclean build folder "problem", maybe move the `cp -a . /tmp/$DIST/singularity` from `centos_run` to `before_install`?

c-ney:
@olifre I forgot to ping you, my changes have been rebased and sitting around for a while. You can check https://github.com/singularityware/singularity/compare/development...c-ney:f63c93f and see if there's anything worthwhile. The opensuse tests are failing with (some?) fixes in current devel-branch, the other distributions seem fine (though based on an old devel-branch).

olifre:
@c-ney This is looking good! Sadly, I did not have time to look at it in detail yet, I guess it's best if I pull your branch and rebase it onto the very newest devel branch. 
Or do you want to open a fresh PR combining both or our changes?

GodloveD:
Hey @olifre.  I think we are considering this for the next release (2.5.2).  Just wanted to circle back and see where things stand.  Did you get a chance to merge @c-ney 's proposed changes?  Thanks!

olifre:
@GodloveD Sadly, I did not yet have time to come back to this yes, and was waiting for a reply from @c-ney whether he'd prefer a fresh PR combining both our changes (and continue work on it), or whether I should rather go through his changes and take missing things over. 
I can try to get back to it, but it might take a week at least with the current workload...

GodloveD:
@olifre Awesome! Would it make sense to rebase this to release-2.5 then?

c-ney:
Sorry for the late reply, the last few months were very busy. I will have some time in the coming weeks.

@olifre, I just tried a quick rebase (onto release-2.5) and failed horribly... it might actually be easier to cherry-pick the commits into a new branch based on release-2.5. If you're too busy, I can do this.

@GodloveD would you prefer a new PR or continuing this one?

GodloveD:
@c-ney it is really up to you whether you'd like to put your changes into a new PR or amend them to this.  If you need to create a new branch and cherry-pick commits, it sounds like it would probably just be easier to create a new PR.  But I leave the choice up to you.  😄

olifre:
> Sorry for the late reply, the last few months were very busy. I will have some time in the coming weeks.

@c-ney: That's good news :smile: . 


> it might actually be easier to cherry-pick the commits into a new branch based on release-2.5. If you're too busy, I can do this.

That's a good idea, I think that's the only option by now. Sadly, as you can already guess from my late reply, I don't have a lot of time at the moment - so I would very much appreciate it if you have the time to do it. I also have the feeling that a new PR would be easier - just ping me there (or here) and I'll follow along. I can also look through the changes and may have suggestions, but I currently lack development time to work on it myself. 
Cheers and thanks! :+1:  Oliver

olifre:
#1713 is looking like a good and clean replacement for this one - so I'm closing this here for good :smile:.

Issue #1186, opened 2017-12-01
devpts silently not mounted in unprivileged (user namespace) mode


olifre:
## Version of Singularity:

2.4

### Expected behavior

devpts is mounted also in unprivileged, user namespace mode.

### Actual behavior

Silently fails. 

### Steps to reproduce behavior

Start any container, have
```
allow setuid = no
mount devpts = yes
```
set. 

This issue exists since https://github.com/singularityware/singularity/pull/944 . 
The more correct fix would have been to actually analyze why it's failing. Sadly, that has not been done (lack of QA), but a silent deactivation of the feature was implemented. 

A simple analysis reveals the reason is the `gid=5` parameter used in the `devpts` mount (while the `5` is resolved from the `tty` gid): 
https://github.com/singularityware/singularity/blob/f2d7795d04bad5af371c3a4a0a6f372ba07ff810/src/lib/runtime/mounts/dev/dev.c#L114
Naturally, this is usually wanted, but can't work in user namespace containers, unless the tty-gid is mapped. Otherwise, `EINVALID` will be received. 
This has already been discussed a few years ago for more full-fledged container solutions:
https://github.com/opencontainers/runc/issues/225

So it would be better to:
- drop the `gid=5` from the mount parameters if user namespaces are used. This will not help for programms actually checking ownership of the devices (but one may argue that's a somewhat backwards idea anyways, since programs should rather check effective permissions), but at least makes `devpts` mounts work again.  
- potentially add support for the workaround described in https://github.com/hpc/charliecloud/issues/67 (i.e. map `tty` group by mapping the user's group to it). This should help for most programs doing such ownership checks, but of course may look awkward to the users. 

Pinging @hollowec , who initially (thankfully) implemented `devpts` mounting, @gmkurtzer who merged the workaround silencing the actual error, and @DrDaveD who is interested in the topic (and implemented the workaround - I don't blame him for that, but I'd rather have expected a **proper problem analysis by the core team** before merging a silencing of an error). 

I'd also like to point out it's kind of bad that the failure discussed in https://github.com/singularityware/singularity/pull/944 was actually not discovered by the test suite.

hollowec:
Hi @olifre,

[...]

PR #1209, opened 2017-12-07
Fix issue 1186

cclerget:

Force use of user namespace without requesting it when `allow setuid = no`

- Ref: #1186 

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/src/util/suid.c
+++ b/src/util/suid.c
@@ -103,15 +103,7 @@ int singularity_suid_init(void) {
         ABORT(255);
     }
 
-    singularity_message(VERBOSE2, "Checking if singularity.conf allows us to run as suid\n");
-    if ( ( singularity_config_get_bool(ALLOW_SETUID) <= 0  ) || ( singularity_registry_get("NOSUID") != NULL ) ) {
-        return(-1);
-    }
-
 #else
-    if ( is_enabled < 0 ) {
-        is_enabled = 0;
-    }
     singularity_message(VERBOSE, "Running NON-SUID program workflow\n");
 
     singularity_message(DEBUG, "Checking program has appropriate permissions\n");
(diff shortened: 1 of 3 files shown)


GodloveD (commented):
This looks like a good change to me.  Can @cclerget or @bauerm97 also have a look and either approve or request changes please?

cclerget:
@GodloveD I think @bauerm97 need to approve since I don't have any rights on the repository

GodloveD:
Sorry.  I actually meant @bauerm97 or @yhcote.  😅 But you should have admin privs to this repo @cclerget.  I'll ask @gmkurtzer about that.

src/util/suid.c
```diff
@@ -121,6 +113,13 @@ int singularity_suid_init(void) {
     }
 #endif /* SINGULARITY_SUID */
 
+    singularity_message(VERBOSE2, "Checking if singularity.conf allows us to run as suid\n");
+    if ( ( singularity_config_get_bool(ALLOW_SETUID) <= 0  ) || ( singularity_registry_get("NOSUID") != NULL ) ) {
+        envar_set("SINGULARITY_NOSUID", "1", 1);
+        singularity_registry_set("NOSUID", "1");
+        return(-1);
```
gmkurtzer: Should we ABORT() if we can't initialize from this function properly?

cclerget: @gmkurtzer No because aborting there when 'allow setuid = no' will always fail, the return value is for condition https://github.com/cclerget/singularity/blob/670c701959286c77216d708576afd1b5da86d0b5/src/wrapper.c#L90 to run non setuid workflow

ctmadison: The return values for this (and other singularity_*_* functions are never checked - at least the several I looked at - see start.c for example). It would seem like all of the places that singularity_suid_init is called need to be changes to check the error return and deal with them correctly. In many cases, I think that we would want to abort since there is no sense in continuing.

cclerget: @carlm1234 singularity_suid_init is called only in wrapper.c, the other files where this one appears are obsolete and not used anymore, no worries about that. To avoid confusion I will move this check into singularity_suid_allowed()

ctmadison (approved):

PR #1220, opened 2017-12-13
Do not change ownership

### LINKED ISSUE #1079: Build from Docker. No valid /bin/sh in container
## Version of Singularity:

2.4

### Expected behavior

Build and shell container using Docker image

### Actual behavior

After the container is build and executed,
```
singularity build sjupyter.simg docker://a33a/sjupyter
singularity shell sjupyter.simg
```
It returns:
```
ERROR  : No valid /bin/sh in container
ABORT  : Retval = 255
```

Try to build and shell the container on Scientific Linux 7.3. The Docker image uses autobuild so it is safe to test ;)

### Environment
I use Scientific Linux 7.3 and the issue might be connected to my previous [question](https://github.com/singularityware/singularity/issues/1078) and build command. Because I compiled the [squashfs-tools](https://github.com/plougher/squashfs-tools) and placed it in /local/usr/bin. I tried to build the image on Ubuntu and then moved the .simg file and it apparently worked. 

Thank you in advance.

GodloveD:
If you change your umask to 0077 and build a container as a non-root user (from Singularity Hub or something) you will find that you cannot use the container.  
```
$ umask 0077

$ singularity build lolcow.simg shub://GodloveD/lolcow
Cache folder set to /home/ubuntu/.singularity/shub
Building from local image: /home/ubuntu/.singularity/shub/GodloveD-lolcow-master.simg
WARNING: Building container as an unprivileged user. If you run this container as root
WARNING: it may be missing some functionality.
Building Singularity image...
Singularity container built: lolcow.simg
Cleaning up...

$ ./lolcow.simg
ERROR  : No valid /bin/sh in container
ABORT  : Retval = 255
```
What the what??  Turns out that `/bin/sh` is owned by root, with permissions == 700.  So a mere mortal user can't do anything with the container.

But why?  For some reason we decided to make all files owned by root when we did the squashfs conversion.  But we are not doing anything like this for sandbox containers, and those seem to work just fine and dandy.  So we should probably not be doing it here either. 

- Fixes #1079

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,10 +15,14 @@ and changes prior to that are (unfortunately) done retrospectively. Critical ite
 
 ## [v3.0](https://github.com/singularityware/singularity/tree/development)
 
+### Implemented enhancements
  - Feature sif sign #1143
  - Add capability support and secure build #934
  - Boot/start instance #1032
  - Put /usr/local/{bin,sbin} in front of the default PATH
+
+### Bug Fixes
+ - Stopped changing all file ownership to root in mksquashfs call
 
 ## [v2.4.2](https://github.com/singularityware/singularity/tree/release-2.4)
 
--- a/libexec/cli/build.exec
+++ b/libexec/cli/build.exec
@@ -380,12 +380,7 @@ if [ -z "${SINGULARITY_SANDBOX:-}" ]; then
         tar -cf - -C "$SINGULARITY_ROOTFS" . | eval_abort ${SINGULARITY_bindir}/singularity image.import "${SINGULARITY_TEMP_FS}"
     else
         message 1 "Building Singularity FS image...\n"
-        if [ "$USERID" != "0" ]; then
-            OPTS="-all-root"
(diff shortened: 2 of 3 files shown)


slefrancois:
On our end we would like to see this merged.

User pulled containers work fine on our end with `umask 002`. However, we use group permissions to share data across the network and recommend `umask 027`. Having users default to group-writable files is risky here.

GodloveD:
I spoke with @gmkurtzer about this some time ago and I don't think he really likes this idea.  He says the fact that we are not making the files in the _sandbox_ root owned is actually a bug.  He says the real fix is that we need to force the umask to be something sensible when we create the container.  So I think this should probably just be closed.

slefrancois:
Sure, any mechanism which does not make the resulting containers dependant on the user's umask during build would work.

GodloveD:
This PR has been rendered moot by the latest version that explicitly sets the umask.

PR #1221, opened 2017-12-13
Restore shim init process for proper signal handling/child reaping

DrDaveD:

This PR restores the "shim" init process inside of singularity containers that are not started with the instance start/stop command.  This is necessary for proper signal handling and reaping of zombie child processes.  The init process (PID 1) is treated specially, ignoring signals by default, and it is assigned all child processes when their parents die.  This feature was added by @bbockelm in PR #383, released in singularity version 2.3, and removed with very little explanation between releases 2.3 & 2.3.1 by @gmkurtzer in commit 2d2ab63f9942965336f91519750e4309bcd18de6.  The description of the commit was "Don't run a second fork until we support start/stop" which doesn't seem very relevant to me, and anyway instance start & stop is now supported and this behavior was not restored for cases where images are started without instance start & stop.

Please also review #384 because it isn't clear to me that it was fully addressed.  #394 said it addressed it partially.

- None

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -107,6 +107,19 @@ int main(int argc, char **argv) {
     singularity_runtime_environment();
     singularity_priv_drop_perm();
 
+    if ( singularity_registry_get("DAEMON_JOIN") == NULL  &&
+         singularity_registry_get("PIDNS_ENABLED") != NULL ) {
+
+        // At this point, we are now PID 1; when we later exec the payload,
+        // it will also be PID 1.  Unfortunately, PID 1 in Linux has special
+        // signal handling rules (the _only_ signal that will terminate the
+        // process is SIGKILL; all other signals are ignored).  Hence, we fork
+        // one more time.  This makes PID 1 a shim process and the payload
+        // process PID 2 (meaning that the payload gets the "normal" signal
+        // handling rules it would expect).
+        singularity_fork_run(0);
+    }
+
     if ( ( target_pwd = singularity_registry_get("TARGET_PWD") ) != NULL ) {
         singularity_message(DEBUG, "Attempting to chdir to TARGET_PWD: %s\n", target_pwd);
         if ( chdir(target_pwd) != 0 ) {
--- a/src/action.c
+++ b/src/action.c
@@ -29,6 +29,7 @@
 #include <errno.h>
 #include <string.h>
 #include <fcntl.h>
+#include <sys/prctl.h>
 
 #include "config.h"
 #include "util/file.h"
@@ -118,6 +119,11 @@ int main(int argc, char **argv) {
         // one more time.  This makes PID 1 a shim process and the payload
         // process PID 2 (meaning that the payload gets the "normal" signal
         // handling rules it would expect).
+
+        // Set the name of the process for ps
+        prctl(PR_SET_NAME, "shim-init", 0, 0, 0);
+        // And for ps -f
+        strncpy(argv[0], "shim-init", strlen(argv[0]));
 
         singularity_fork_run(0);
     }
--- a/src/action.c
+++ b/src/action.c
@@ -33,6 +33,7 @@
 
 #include "config.h"
 #include "util/file.h"
+#include "util/fork.h"
 #include "util/util.h"
 #include "util/daemon.h"
 #include "util/registry.h"
--- a/src/action.c
+++ b/src/action.c
@@ -108,7 +108,8 @@ int main(int argc, char **argv) {
     singularity_priv_drop_perm();
 
     if ( singularity_registry_get("DAEMON_JOIN") == NULL  &&
-         singularity_registry_get("PIDNS_ENABLED") != NULL ) {
+         singularity_registry_get("PIDNS_ENABLED") != NULL &&
+         singularity_registry_get("NOSHIMINIT") == NULL ) {
 
         // At this point, we are now PID 1; when we later exec the payload,
         // it will also be PID 1.  Unfortunately, PID 1 in Linux has special
@@ -117,6 +118,7 @@ int main(int argc, char **argv) {
         // one more time.  This makes PID 1 a shim process and the payload
         // process PID 2 (meaning that the payload gets the "normal" signal
         // handling rules it would expect).
+
         singularity_fork_run(0);
     }
 
--- a/libexec/cli/action_argparser.sh
+++ b/libexec/cli/action_argparser.sh
@@ -105,6 +105,11 @@ while true; do
             SINGULARITY_UNSHARE_PID=1
             export SINGULARITY_UNSHARE_PID
         ;;
+        --noinit)
+            shift
+            SINGULARITY_NOSHIMINIT=1
+            export SINGULARITY_NOSHIMINIT
+        ;;
         -i|--ipc)
             shift
             SINGULARITY_UNSHARE_IPC=1
--- a/libexec/cli/exec.info
+++ b/libexec/cli/exec.info
@@ -33,6 +33,7 @@ EXEC OPTIONS:
     --nv                Enable experimental Nvidia support
     -o|--overlay        Use a persistent overlayFS via a writable image
     -p|--pid            Run container in a new PID namespace
+    --noinit            Do not start shim init process with --pid
     --pwd               Initial working directory for payload process inside 
                         the container
     -S|--scratch <path> Include a scratch directory within the container that 
--- a/libexec/cli/run.info
+++ b/libexec/cli/run.info
@@ -38,6 +38,7 @@ RUN OPTIONS:
     --nv                Enable experimental Nvidia support
     -o|--overlay        Use a persistent overlayFS via a writable image
     -p|--pid            Run container in a new PID namespace
+    --noinit            Do not start shim init process with --pid
     --pwd               Initial working directory for payload process inside
                         the container
     -S|--scratch <path> Include a scratch directory within the container that 
--- a/libexec/cli/shell.info
+++ b/libexec/cli/shell.info
@@ -33,7 +33,8 @@ SHELL OPTIONS:
                         only network device active)
     --nv                Enable experimental Nvidia support
     -o|--overlay        Use a persistent overlayFS via a writable image
-    -p|--pid            Run container in a new PID namespace (creates child)
+    -p|--pid            Run container in a new PID namespace
+    --noinit            Do not start shim init process with --pid
     --pwd               Initial working directory for payload process inside
                         the container
     -S|--scratch <path> Include a scratch directory within the container that
--- a/src/lib/runtime/ns/pid/pid.c
+++ b/src/lib/runtime/ns/pid/pid.c
@@ -69,6 +69,15 @@ int _singularity_runtime_ns_pid(void) {
         singularity_fork_daemonize(CLONE_NEWPID);
     } else {
         singularity_fork_run(CLONE_NEWPID);
+
+        // At this point, we are now PID 1; when we later exec the payload,
+        // it will also be PID 1.  Unfortunately, PID 1 in Linux has special
+        // signal handling rules (the _only_ signal that will terminate the
+        // process is SIGKILL; all other signals are ignored).  Hence, we fork
+        // one more time.  This makes PID 1 a shim process and the payload
+        // process PID 2 (meaning that the payload gets the "normal" signal
+        // handling rules it would expect).
+        singularity_fork_run(0);
     }
 
     singularity_registry_set("PIDNS_ENABLED", "1");
--- a/src/lib/runtime/ns/pid/pid.c
+++ b/src/lib/runtime/ns/pid/pid.c
@@ -69,15 +69,6 @@ int _singularity_runtime_ns_pid(void) {
         singularity_fork_daemonize(CLONE_NEWPID);
     } else {
         singularity_fork_run(CLONE_NEWPID);
-
-        // At this point, we are now PID 1; when we later exec the payload,
-        // it will also be PID 1.  Unfortunately, PID 1 in Linux has special
-        // signal handling rules (the _only_ signal that will terminate the
-        // process is SIGKILL; all other signals are ignored).  Hence, we fork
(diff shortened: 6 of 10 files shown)


cclerget:
@DrDaveD Few problems with shim process forked here :
- it lives in host / since it isn't concerned by chroot/pivot_root, that breaks container isolation
- this process remains privileged for setuid execution
- it only concerns runs with PID namespace

But can you give a use case to prefer exec rather than instance.start/stop for background process to discuss if that can be handled differently ?

DrDaveD:
@cclerget I'm sorry I didn't see your response sooner.

> it lives in host / since it isn't concerned by chroot/pivot_root, that breaks container isolation

I don't know what the implications of that are.  What kind of problem would that cause?  So do we just have to move the extra fork to a different place, after the chroot/pivot_root but before the command is exececuted?

> this process remains privileged for setuid execution

What kind of problems does that cause?  Is this exactly what @bbockelm referred to in #384?  It sounds like moving the fork would also solve it.

> it only concerns runs with PID namespace

Yes, so?  That's the place where the problem exists.

> But can you give a use case to prefer exec rather than instance.start/stop for background process to discuss if that can be handled differently ?

We absolutely want to use exec instead of instance.start & stop.  We don't want to make our applications any more complicated than they have to be.  A single user will be running code from multiple other users, and we want  all of the other users to be isolated from each other.  I don't see any reason why we should have to manage a set of running instances and keep track of which one belongs to which, when an exec will do it much easier.

cclerget:
> I don't know what the implications of that are. What kind of problem would that cause?

This is problematic for setuid execution and a potential way for escaping container

> Is this exactly what @bbockelm referred to in #384? It sounds like moving the fork would also solve it

Indeed, moving it after singularity_priv_drop_perm like instance did to avoid these problems

> Yes, so? That's the place where the problem exists.

You're right, useless point, sorry

>  I don't see any reason why we should have to manage a set of running instances and keep track of which one belongs to which, when an exec will do it much easier.

But you don't need to keep track of instances by forking a script which does something like that : 
```
#!/bin/sh
jobid=$1
singularity instance.start /somewhere/softdirimg $jobid
singularity exec instance://$jobid userscript args
singularity instance.stop $jobid
```

DrDaveD:
@cclerget  What's the motivation to move to the more complicated 3 commands instead of the one simpler command?  Just to avoid the relatively subtle bug addressed by this PR?  Are there any other advantages?  Keep in mind that we would have to wait for a huge number of distributed worker nodes managed by dozens of different people to upgrade to a version of singularity that supports it, a version that doesn't exist yet because of issue #1227, or we'd have to have a script that detects when a new enough version is not yet available and fall back to the old way if it isn't.   Also the instance is run in the background as a child of PID 1, so if the process that started it gets killed, the instance becomes an orphan.  No thanks.   We could add a kill signal handler, but sometimes jobs get killed with kill -9 and there's nothing we could do to avoid the orphan.

I will try to fix this PR instead to move where the fork happens.

cclerget:
@DrDaveD I tested fix but shim process doesn't catch defunc processes, here the script used for testing :
```
#!/bin/sh

cat > /tmp/zombie.c << ZOMBIE
#include <sys/wait.h>
#include <stdlib.h>
#include <unistd.h>
#include <signal.h>

int main(void)
{
    pid_t child, parent;
    pid_t pids[10];
    int i;

    for (i = 9; i >= 0; --i) {
        pids[i] = fork();
        if (pids[i] == 0) {
            sleep(i+1);
            _exit(0);
        }
    }

    parent = getpid();
    child = fork();
    if ( child == 0 ) {
        kill(parent, SIGKILL);
    } else {
        sleep(1);
    }
    return 0;
}
ZOMBIE

gcc /tmp/zombie.c -static -o /tmp/zombie
singularity exec -u -p /tmp/userns /tmp/zombie &
```

My opinion is to make shim process for exec/shell as an option not as default since it covers case for applications creating zombie processes which means these one are buggy. An alternative could be to use a wrapper like this :

```
#define _GNU_SOURCE
#include <sys/wait.h>
#include <stdlib.h>
#include <unistd.h>
#include <signal.h>

int main(int argc, char **argv, char **envp) {
    int status;
    pid_t child = fork();
    if ( child == 0 ) {
        execvpe(argv[1], &argv[1], envp);
        exit(1);
    }
    while ( waitpid(-1, &status, 0) >= 0 );
    return WEXITSTATUS(status);
}
```

> a version that doesn't exist yet because of issue #1227

To resolve issue #1227 the following lines should go in runtime package :
```
%{_libexecdir}/singularity/handlers
%{_libexecdir}/singularity/image-handler.sh
```

DrDaveD:
@cclerget Ok on the zombie test.  Without testing I had just assumed that those child processes were reaped because they presumably were before.  Now running your test program I can see that they aren't with the current version of this PR, if I do something to keep the container from exiting too fast such as having singularity run `bash -c '/tmp/zombie;sleep 10'`.  The child reaping should be easy to fix and I will plan to address it in an update to the PR.

I don't think expecting people to use a wrapper script for something easily fixed inside singularity is reasonable.  It's also not necessarily a bug in an application to spin off a separate process like that and for it to exit before the job does.

Besides, even more important than the zombie processes is the default signal handling.  If you now run `singularity exec -p ... sleep 10` it does not respond to signals such as control-C.  My current PR fixes that.

DrDaveD:
Ok @cclerget please comment on this PR now.

cclerget:
@DrDaveD Thanks for taking time to address this bug. Few comments, if first parent is killed, child become orphaned and attached to host init process, this bug was already present because of implementation of PID namespace :

```
$ singularity exec -p /tmp/img sleep 60 &
$ kill -9 $!
```
Fixable with help of PR_SET_PDEATHSIG

shell command is not concerned by the bug since it catches childs, that make me think we could fix it easier by changing behaviour at image level in .singularity.d/actions/exec (eventually for run command too), so instead of using :

`exec "$@"`

using 

`"$@"`

to spawn a shell as PID 1 which handle defunc processes and execute command as PID 2.
And simply pass PIDNS_ENABLED / DAEMON_JOIN environment variable to script to check if shim process is required or not

DrDaveD:
@cclerget It sounds like that is an independent bug, with or without this bug fix, so how about addressing it in a separate PR?

cclerget:
@DrDaveD Yes it's a separate bug, I will submit a PR for this one

DrDaveD:
So @cclerget do you now approve this pull request?  Can we get another reviewer as well?

cclerget:
Sorry for the delay @DrDaveD ! After thinking about it, enforcing shim init process will break the case where a user want to run `singularity exec -p image /sbin/init` for testing purpose/or needs, since init systems requires to be PID 1 to run correctly, with shim process, /sbin/init will always run as PID 2 and execution will fail. IMHO shim init process should be enabled when requested via command line rather than automatically by default.

Between, the issue is resolved by spawning a shell as PID 1 which handle defunc processes like that :

`singularity exec -p image /bin/sh -c "command args ..."`

ping @bauerm97, what do you think about those changes ?

DrDaveD:
Testing init processes is a very rare case.  I would rather have an option to disable the shim init process than disable it.  I think it should be on by default.

cclerget:
But this issue is a corner case. I think the role of Singularity is to setup a container environment, not decide what to run inside container, this is left to end user, that's why I'm in favor to spawn a shim init process as an option.

olifre:
There's a nice discussion about the pro's and con's of this here:
https://hackernoon.com/the-curious-case-of-pid-namespaces-1ce86b6bc900
Notably, see `The Docker “Mistake”` (they also do not run an init process by default, but allow easy use of https://github.com/krallin/tini ). 

I would claim right now singularity gives off the vibe of a "I do everything for you" piece of software, i.e. it handles environment files, sets up paths, bind mounts the home directory by default, `cd`s to a working directory, it even makes use of compressed images. This is nice for the approach to `just fire up a container` without deeper knowledge of the internals. Adding a shim init to make signal handling and zombie reaping work fine fits well in this description. 

Another approach is the one chosen by most other container solutions such as runc, charliecloud, railcar etc. They focus on production environments and implementation of stable industry standards. Hence, they provide a very configurable, by default "lightweight" implementation - they expect users to know what they are doing and configure what they need. They also expect users to know what a shim init process is, why it might be needed for their application, and how to add / start it. 

So in summary, I would say "PID namespace by default" => "shim init by default", especially in a usability oriented implementation. Things would of course be different if the goal of Singularity was different. 
Maybe my conception is wrong here, or things did change? Is Singularity a feature-driven tool targetting end-users or is it trying to replicate Docker / runc?

DrDaveD:
I think that ignoring the special properties of pid 1 in a pid namespace is a bug, a misunderstanding of how they work.  Any container system should of course provide a meaningful default pid 1.

This is functionality that was in singularity at one time and was accidentally dropped.  It should be a no-brainer.

cclerget:
@olifre @DrDaveD Docker and co enter PID namespaces by default, Singularity don't while targeting primarily HPC use cases. Since PID namespaces are requested by user, if a user decides to enter in PID namespace, he has good reason to do so and he's certainly aware of implications to execute his process as PID 1. In this logical, my point of view was to add an option for spawning a shim init process on user request rather than by default and let Singularity be agnostic about running process and not decide for user. Maybe it's safest to assume that users requesting PID namespace are not aware of PID 1 properties.

@DrDaveD Can you add an option to disable shim process please ?

bbockelm:
Note that this:

```
singularity exec -p image /bin/sh -c "command args ..."
```

will likely NOT do the correct thing.  `/bin/sh` is not guarantee to implement `init`-like signal handling.  For example, I would expect this to ignore `SIGTERM` which would be quite surprising to most users...

cclerget:
@bbockelm I spoke about resolving @DrDaveD case issue, not about shim init process issue

DrDaveD:
> @DrDaveD Can you add an option to disable shim process please ?

Sure, will do.

DrDaveD:
@cclerget Ok I added a --noinit option, please re-review.  I also set the name of the process to "shim-init" for ps.

src/action.c
```diff
@@ -103,6 +104,26 @@ int main(int argc, char **argv) {
...
+        // signal handling rules (the _only_ signal that will terminate the
+        // process is SIGKILL; all other signals are ignored).  Hence, we fork
+        // one more time.  This makes PID 1 a shim process and the payload
+        // process PID 2 (meaning that the payload gets the "normal" signal
+        // handling rules it would expect).
+
+        // Set the name of the process for ps
+        prctl(PR_SET_NAME, "shim-init", 0, 0, 0);
+        // And for ps -f
+        strncpy(argv[0], "shim-init", strlen(argv[0]));
```
cclerget: Overflow is possible there, we can't be sure there is enough room for argv[0] since it can be set by user and contain only 1 character

cclerget (changes requested):
Remove my comment, my eyes wasn't open. Overall looks good to me

cclerget (approved):

bauerm97:
@DrDaveD Hey Dave, just trying to wrap my head around the changes to make sure I understand what's going on:

1. When running with `-p` option, is the shim-init process automatically enabled?
2. What happens when the contained proc (*i.e. pid 2*) exits? Are the signals handled properly, does the shim-init proc get cleaned up?
3. Could you write a few brief tests to ensure that the shim-init proc and associated PID namespace get cleaned up properly when the contained proc exits?

DrDaveD:
@bauerm97 Hi Michael, I'm sorry for the delay, I'm not caught up on my email since my vacation 2 weeks ago.

1. Yes, with -p, the shim-init process is enabled by default, unless --noinit is also used.
2. Yes, the shim-init process always exits when its children exit.  It's not related to signal handling.
3. I'm not sure what you mean.  Are you talking about adding something to 'make test'?  Any existing test of -p will already make sure that the shim-init is exiting, because it is the process that its parent waits on.  So I don't think a test like that is needed.  On the other hand, testing for the presence of shim-init with -p and for its absence with -p --noinit could be useful to avoid another regression.

DrDaveD:
This has been rebased against development-2.x

DrDaveD:
I rebased again against development-2.x.

GodloveD:
So what is the final assessment on this?  @cclerget do you think it is a good idea to add the init process to Singularity?  Wonder what @bauerm97 thinks?

cclerget:
@GodloveD After a long discussion, that seems the right thing to do

GodloveD (changes requested):
Hi @DrDaveD I think we are getting close to merging this.  Thanks for sticking with it.  

I'm noticing some compiler warnings.  Could you fix these up before we merge please?

```
../../util/fork.c: In function ‘singularity_fork’:
../../util/fork.c:357:14: warning: implicit declaration of function ‘singularity_suid_enabled’ [-Wimplicit-function-declaration]
         if ( singularity_suid_enabled() && !singularity_priv_dropped_perm() ) {
              ^
action.c: In function ‘main’:
action.c:128:9: warning: implicit declaration of function ‘singularity_fork_run’ [-Wimplicit-function-declaration]
         singularity_fork_run(0);
         ^
util/fork.c: In function ‘singularity_fork’:
util/fork.c:357:14: warning: implicit declaration of function ‘singularity_suid_enabled’ [-Wimplicit-function-declaration]
         if ( singularity_suid_enabled() && !singularity_priv_dropped_perm() ) {
              ^
action.c: In function ‘main’:
action.c:128:9: warning: implicit declaration of function ‘singularity_fork_run’ [-Wimplicit-function-declaration]
         singularity_fork_run(0);
         ^
util/fork.c: In function ‘singularity_fork’:
util/fork.c:357:14: warning: implicit declaration of function ‘singularity_suid_enabled’ [-Wimplicit-function-declaration]
         if ( singularity_suid_enabled() && !singularity_priv_dropped_perm() ) {
```

DrDaveD:
Ah, sorry, I added a commit for that.

GodloveD (approved):
Looks good to me!  Thanks @DrDaveD and @cclerget !

Issue #1206, opened 2017-12-07
make distclean fails


olifre:
## Version of Singularity:

development branch

### Expected behavior

`make distclean` runs fine after `configure`.

### Actual behavior

It fails.

### Steps to reproduce behavior
```
./autogen.sh
./configure --prefix=/usr/local
make distclean
```

This is tested now as part of https://github.com/singularityware/singularity/pull/1204 and usually called by `debuild`. 
The travis report linked to the mentioned PR shows the encountered errors, here a short excerpt:
```
make[5]: Leaving directory '/tmp/debian9/singularity/src/lib/runtime/autofs'
make[5]: Entering directory '/tmp/debian9/singularity/src/lib/runtime'
Makefile:540: ../../util/.deps/libinternal_la-capability.Plo: No such file or directory
Makefile:541: ../../util/.deps/libinternal_la-config_parser.Plo: No such file or directory
Makefile:542: ../../util/.deps/libinternal_la-file.Plo: No such file or directory
Makefile:543: ../../util/.deps/libinternal_la-fork.Plo: No such file or directory
Makefile:544: ../../util/.deps/libinternal_la-message.Plo: No such file or directory
Makefile:545: ../../util/.deps/libinternal_la-mount.Plo: No such file or directory
Makefile:546: ../../util/.deps/libinternal_la-privilege.Plo: No such file or directory
Makefile:547: ../../util/.deps/libinternal_la-registry.Plo: No such file or directory
Makefile:548: ../../util/.deps/libinternal_la-setns.Plo: No such file or directory
Makefile:549: ../../util/.deps/libinternal_la-suid.Plo: No such file or directory
Makefile:550: ../../util/.deps/libinternal_la-util.Plo: No such file or directory
make[5]: *** No rule to make target '../../util/.deps/libinternal_la-util.Plo'.  Stop.
make[5]: Leaving directory '/tmp/debian9/singularity/src/lib/runtime'
Makefile:696: recipe for target 'distclean-recursive' failed
```
Since I'm no automake expert (I usually use more modern and nowadays more widespread build systems such as cmake or meson), I sadly can't offer a fix.

jmstover:
This comes from the Makefile being generated in `src/lib/runtime` including lines like:

```
include ../../util/$(DEPDIR)/libinternal_la-capability.Plo
```

I'm messing around with the `Makefile.am` to see what I can come up with.

dtrudg:
Closing - this was fixed in #1246

PR #1246, opened 2018-01-09
Add message to top level Makefile.am about mksquashfs and Makefile cleanup

jmstover:

mksquashfs is a install requirement in the .spec file. After `make` is ran, output a message saying mksquashfs is required.

General cleanup of Makefiles and to resolve issue #1206

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -315,6 +315,14 @@ else
 
 fi
 
+# ---------------------------------------------------------------------
+# MKSQUASHFS
+# ---------------------------------------------------------------------
+AC_CHECK_PROG(MKSQUASHFS_CHECK,mksquashfs,yes)
+if test x"$MKSQUASHFS_CHECK" != "xyes" ; then
+    AC_MSG_ERROR([mksquashfs is required.])
+fi
+
 AC_MSG_CHECKING([--with-slurm])
 AC_ARG_WITH([slurm],
             AS_HELP_STRING([--with-slurm], [This feature will no longer be part of Singularity proper]),
--- a/configure.ac
+++ b/configure.ac
@@ -320,7 +320,7 @@ fi
 # ---------------------------------------------------------------------
 AC_CHECK_PROG(MKSQUASHFS_CHECK,mksquashfs,yes)
 if test x"$MKSQUASHFS_CHECK" != "xyes" ; then
-    AC_MSG_ERROR([mksquashfs is required.])
+    AC_MSG_ERROR([Unable to find mksquashfs, need package squashfs-tools.])
 fi
 
 AC_MSG_CHECKING([--with-slurm])
--- a/configure.ac
+++ b/configure.ac
@@ -315,14 +315,6 @@ else
 
 fi
 
-# ---------------------------------------------------------------------
-# MKSQUASHFS
-# ---------------------------------------------------------------------
-AC_CHECK_PROG(MKSQUASHFS_CHECK,mksquashfs,yes)
-if test x"$MKSQUASHFS_CHECK" != "xyes" ; then
-    AC_MSG_ERROR([Unable to find mksquashfs, need package squashfs-tools.])
-fi
-
 AC_MSG_CHECKING([--with-slurm])
 AC_ARG_WITH([slurm],
             AS_HELP_STRING([--with-slurm], [This feature will no longer be part of Singularity proper]),
--- a/Makefile.am
+++ b/Makefile.am
@@ -21,6 +21,13 @@ install-perms:
 	@echo "install-perms is no longer required"
 	@echo
 
+all-local:
+	@echo
+	@echo "******"
+	@echo "  mksquashfs from squash-tools is required for full functionality"
+	@echo "******"
+	@echo
+
 install-data-hook:
 	install -d -m 0755 $(DESTDIR)$(CONTAINER_MOUNTDIR)
 	install -d -m 0755 $(DESTDIR)$(CONTAINER_FINALDIR)
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,6 +1,6 @@
 SUBDIRS = bin etc libexec man src
 
-MAINTAINERCLEANFILES = Makefile.in aclocal.m4 configure ltmain.sh depcomp install-sh missing config.* *.m4 singularity-*.tar.gz singularity-*.rpm m4/* test.sh secbuildimg.sh
+MAINTAINERCLEANFILES = Makefile.in aclocal.m4 configure ltmain.sh compile depcomp install-sh missing config.* *.m4 singularity-*.tar.gz singularity-*.rpm m4/* test.sh secbuildimg.sh
 DISTCLEANFILES = Makefile test.sh secbuildimg.sh
 CLEANFILES = 
 EXTRA_DIST = singularity.spec autogen.sh examples debian CONTRIBUTORS.md CONTRIBUTING.md COPYRIGHT.md INSTALL.md LICENSE-LBNL.md LICENSE.md README.md tests
--- a/libexec/python/Makefile.am
+++ b/libexec/python/Makefile.am
@@ -6,4 +6,7 @@ scriptlibexecdir = $(libexecdir)/singularity/python
 dist_scriptlibexec_DATA = defaults.py __init__.py message.py shell.py sutils.py base.py templates.py
 dist_scriptlibexec_SCRIPTS = import.py pull.py size.py
 
-MAINTAINERCLEANFILES = Makefile.in
+MAINTAINERCLEANFILES = Makefile.in *.pyc
+DISTCLEANFILES = Makefile *.pyc
+CLEANFILES = *.pyc
+
--- a/man/Makefile.am
+++ b/man/Makefile.am
@@ -1,1 +1,3 @@
+MAINTAINERCLEANFILES = Makefile.in
+
 dist_man_MANS = singularity.1
--- a/src/lib/image/Makefile.am
+++ b/src/lib/image/Makefile.am
@@ -26,10 +26,12 @@ libsingularity_image_la_CFLAGS = $(AM_CFLAGS)
 # These are kludges so they don't remove the $(DEPDIR) in ../../util/ otherwise
 # the clean will fail when other Makefile tries to remove those directories
 distclean: distclean-recursive
-	-rm ./$(DEPDIR)
+	-rm -rf ./$(DEPDIR)
+	-rm -rf ./.libs
 	-rm -f Makefile
 
 maintainer-clean: maintainer-clean-recursive
-	-rm ./$(DEPDIR)
+	-rm -rf ./$(DEPDIR)
+	-rm -rf ./.libs
 	-rm -f Makefile
 
--- a/src/lib/image/Makefile.am
+++ b/src/lib/image/Makefile.am
@@ -28,10 +28,10 @@ libsingularity_image_la_CFLAGS = $(AM_CFLAGS)
 distclean: distclean-recursive
 	-rm -rf ./$(DEPDIR)
 	-rm -rf ./.libs
-	-rm -f Makefile
+	-rm -f $(DISTCLEANFILES) $(CLEANFILES)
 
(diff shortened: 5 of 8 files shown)


olifre:
It seems this still shows an error, e.g.:
```
rm ./.deps
rm: cannot remove './.deps': Is a directory
Makefile:955: recipe for target 'distclean' failed
make[4]: [distclean] Error 1 (ignored)
```
See:
https://travis-ci.org/olifre/singularity/jobs/328191995
for details.

olifre:
In addition, it seems to be incomplete. 
See:
https://travis-ci.org/olifre/singularity/jobs/328194283
which contains at the end:
```
dpkg-source: error: cannot represent change to libexec/python/defaults.pyc: binary file contents changed
dpkg-source: error: add libexec/python/defaults.pyc in debian/source/include-binaries if you want to store the modified binary in the debian tarball
dpkg-source: error: cannot represent change to libexec/python/message.pyc: binary file contents changed
dpkg-source: error: add libexec/python/message.pyc in debian/source/include-binaries if you want to store the modified binary in the debian tarball
dpkg-source: error: cannot represent change to libexec/python/sutils.pyc: binary file contents changed
dpkg-source: error: add libexec/python/sutils.pyc in debian/source/include-binaries if you want to store the modified binary in the debian tarball
dpkg-source: warning: executable mode 0755 of 'missing' will not be represented in diff
dpkg-source: error: cannot represent change to src/lib/signing/libsigning_la-crypt.o: binary file contents changed
dpkg-source: error: add src/lib/signing/libsigning_la-crypt.o in debian/source/include-binaries if you want to store the modified binary in the debian tarball
dpkg-source: error: unrepresentable changes to source
```
This is Debian packaging diffing the original source tarball and the files found after `make distclean`.

jmstover:
Grab at least `c34dfb5`. I noticed after the fact that the initial rm's didn't have a `-rf` when trying to remove a directory (though it didn't error for me :/ ). That last commit, should fix that, and catch the `.libs` directory as well.

And I'm not sure on the Python side... I guess I can add in a `rm -f *.pyc` and just be done with it.

olifre:
> That last commit, should fix that, and catch the .libs directory as well.

Yes, that's why I also cherry-picked it. However, it misses the directory I cited and you can find in the build log I mentioned. 
This is the place:
`src/lib/image/Makefile.am` which still misses the `-rf` for `.deps`. 


> And I'm not sure on the Python side... I guess I can add in a rm -f *.pyc and just be done with it.

That should fix it. The remaining object file seems to be missing distcleaning in:
`src/lib/signing/Makefile.am`.

jmstover:
@olifre Working on getting it cleaned up now.... Hopefully this next commit will straighten it out

olifre:
Nice improvement!
It seems now the `Makefile.in` files are left:
https://travis-ci.org/olifre/singularity/jobs/328233225
They are in `.gitignore`, but still I'd expect they are removed by `distclean` (and it seems Debian expects that, too).

olifre:
> They are in .gitignore, but still I'd expect they are removed by distclean (and it seems Debian expects that, too).

Or maybe they should not, since they are created by `autogen` and not `make`? I'm not sure, what is the general approach?

olifre:
Mhhhm. Running:
`make distclean && git clean -fx` appears to do the job fine. 

However, going with the definition of `distclean` from https://www.gnu.org/prep/standards/html_node/Standard-Targets.html , I would say `make distclean` alone should restore the exact state of `git archive` (or an archive downloaded from github releases). 
This would mean it should also cleanup the stuff created by autogen and configure, even though it's gitignored.

jmstover:
I just downloaded the release 2.4.2 tarball from GitHub and it has the Makefile.in files in it.  In theory, that's correct. A release should *not* have the `*.ac` or `*.am` files... as that's a maintainer thing. But GMK usually ships everything to rebuild from the original sources if wanted.

The Makefile.in files are used to build the corresponding Makefile when the configure script is ran. If you look in `configure.ac` in the `AC_CONFIG_FILE` section, those files get built from corresponding `[name].in` file. So... `make maintainer-clean` should remove some of the `*.in` files, but not `distclean`.

`make distclean` should bring it back to a point where you just need to run: `./configure` again

olifre:
> I just downloaded the release 2.4.2 tarball from GitHub and it has the Makefile.in files in it. In theory, that's correct. A release should not have the *.ac or *.am files... as that's a maintainer thing. But GMK usually ships everything to rebuild from the original sources if wanted.

Thanks for pointing this out! Yes, indeed I was unaware that e.g. https://github.com/singularityware/singularity/releases/download/2.4.2/singularity-2.4.2.tar.gz and https://github.com/singularityware/singularity/archive/2.4.2.tar.gz (or a "git archive") have different contents. Hence, I would not call the "release tarball" the "original source". Also, this means that building from the source tree and from a release tarball may yield different results, especially if the person building the package uses a different automake version than GMK. 
Maybe I'll open another issue asking for documentation of what exactly has been done between the sources and the release tarball (probably just autoreconf, but that should be documented, and also the automake version used) since it's not part of the release checklist. 



> So... make maintainer-clean should remove some of the *.in files, but not distclean.

Thanks, that explains it. Still, running `make maintainer-clean` forgets two files:
```
compile
man/Makefile.in
```
Should I open a separate issue on that?

jmstover:
> Still, running make maintainer-clean forgets two files

Thanks, I'll get those added in and push another update. Trying to get as much as possible cleaned up in the Makefiles from this one PR.

olifre:
Thanks! I'll adapt my packaging tests tomorrow and let you know if anything else pops up, but I think this should be complete. :smile:

jmstover:
> Maybe I'll open another issue asking for documentation of what exactly has 
> been done between the sources and the release tarball (probably just autoreconf

Looking at it, the releases tarballs are probably created by running something like:

`./autogen.sh && ./configure && make dist-gzip`

That creates a `singularity-[version].tar.gz` file.

The Makefiles don't really specify anything for the dist_* variables, so `make dist` pretty much just takes the structure. That includes the `Makefile.am` files, and everything else. If you want to control what gets caught up, you have to specify every file that's either included or excluded in it.

configure.ac
```diff
@@ -315,6 +315,14 @@ else
 
 fi
 
+# ---------------------------------------------------------------------
+# MKSQUASHFS
+# ---------------------------------------------------------------------
+AC_CHECK_PROG(MKSQUASHFS_CHECK,mksquashfs,yes)
+if test x"$MKSQUASHFS_CHECK" != "xyes" ; then
+    AC_MSG_ERROR([mksquashfs is required.])
```
GodloveD: Would you mind making this warning more like the other warnings in `configure.ac`.  How about this:
```
configure: error: Unable to find mksquashfs, need package squashfs-tools
```
That way a user who doesn't really know how to fix this error doesn't have to Google it.

GodloveD (changes requested):
Hey @jmstover.  I tested this and it looks like it works from what I can tell.  Let me just suggest a single change to the error message.

jmstover:
@GodloveD Output message updated.

GodloveD (approved):

GodloveD:
Groovy.  Don't know why the Travis build isn't running.  I tried stopping and restarting but to no avail.  I'll try again in the morning.  If it passes I'll go ahead and merge.

olifre:
> Groovy. Don't know why the Travis build isn't running. I tried stopping and restarting but to no avail. I'll try again in the morning. If it passes I'll go ahead and merge.

Travis is currently dead:
https://www.traviscistatus.com/incidents/c1s4dlyxd4lf

olifre:
@GodloveD @jmstover The following valid point has been raised in #1204 by @c-ney : Why is `mksquashfs` required to _build_ singularity? 
The RPM only requires it as a `Requires`, i.e. only for installation of the final package, so only as "runtime requirement". 
Is it really used at build time for anything? 

If not, this check should surely be removed, if yes, this PR should also adapt the spec file to make it a `BuildRequires`.

vsoch:
RIP Travis 😢

jmstover:
@olifre Yes, it's a Require in the RPM spec file, so would be added when the RPM is installed.

But building from source we can't force something being there as you can with a package that tracks dependencies. So, since it's used we say it needs to be there when built since we can check that.

This is where it's different between a SPEC file and the source. The SPEC file is going to force the install when the package is installed, but there's no way building from source we can do the same thing unless we bail out if we don't find it during the `./configure` stage.

The initial reason it came up is because someone was building from source and had questions on mksquashfs and where it was required (they initially didn't have it anywhere). So, this is as close to being an equivalent "Require" with autoconf as we can get.

jmstover:
Okay... it may be possible to change this to just output this information. So, something like this in the top-level `Makefile.am`;

```
all-local:
    @echo
    @echo "**** mksquashfs from squashfs-tools is required for full Singularity functionality ****"
    @echo
```

But, there's no guarantee of when the `*-local` hooks may get ran. If you run `make` it should be fine... but if you build parallel with say: `make -j8`  ... it's _possible_ that it gets output before a compile further down the line completes, so it isn't the last thing output in that case.

Generally with this structure, and the top-level being one up from any source it _should_ be fine... should being the key word though.

olifre:
> but there's no way building from source we can do the same thing unless we bail out if we don't find it during the ./configure stage.

Right. In general, it's also not the build system's job to check for any run time dependencies. 
Doing that and enforcing them to be present at build-time effectively makes them build time requirements needlessly - which means all packaging code should be adapted to have a `BuildRequires` on `squashfs-tools` if this check goes in as it is. Right now, it actively breaks all distro's packaging - so I wonder how it could pass @GodloveD s review. There's no other package I know of which does that - and if there was, distros would patch out those broken checks downstream and ask upstream to fix that in a future release, instead of adding wrong dependencies. 

In general, the way to go should be to document run time requirements. In any case, user's only have to take care if they don't use pre-built packages, and users not using packages from their distro should know what they are doing. 

The other option would be to have singularity check at runtime, but of course this comes at an additional cost, and it's questionable whether this should be done, when one can just expect users to read the docs.

GodloveD:
Ah.  That makes sense.  Thanks for the clarification @olifre.  I did not think to check distro packaging.  Only building from source.  I suppose this means Travis tests will fail momentarily.

olifre:
> I suppose this means Travis tests will fail momentarily.

Since the travis tests don't do correct packaging (c.f. https://github.com/singularityware/singularity/pull/1204 ), this might even go unnoticed. 
I'd say it's also wrong when building from source. 

Especially HPC users usually build on one build host machine, and install / deploy via a shared filesystem / cvmfs to many worker nodes. 
So checking for presence of `squashfs-tools` on the build machine is completely useless for all production usecases and just causes unnecessary installation of stuff on the build host. It may only be of interest in development environments, when build environment and usage environment coincide - but then, developers should know what they do.

jmstover:
@olifre In the latest commit, I removed the `configure.ac` check, and just did a hook on `all-local` to output a message after `make` is ran.

That should _hopefully_ satisfy everything. Something to say mksquashfs is required when doing a source build, and not breaking the dependency tree.

From what I was able to get, mksquashfs is required _everywhere_. Even some transitory images are created using mksquashfs ... so it has to be not only on the build host, but where singularity is ran at as well.

olifre:
> From what I was able to get, mksquashfs is required everywhere. Even some transitory images are created using mksquashfs ... so it has to be not only on the build host, but where singularity is ran at as well.

I believe it is not needed when sandbox images and unprivileged mode are used (main usecase of WLCG), but of course you are right: Both the "Singularity container build machine" and the "Singularity container execute machine" will need `mksquashfs`. 
Only the "Singularity compile machine", which could be a dedicated rpmbuild / debuild machine or build infrastructure in the cloud / from a Linux distribution will certainly not need it, and this is the major concern. 

So I believe now everything is fine, many thanks for your work on this! :smile:

GodloveD:
Thanks very much @jmstover for addressing this issue and to @olifre for taking the time to discuss and work through the tweaks.  Much appreciated.

Issue #1260, opened 2018-01-19
Clean environment LANG=C causing tar extraction errors - docker forces UTF8


dtrudg:
## Version of Singularity:

Development

### Expected behavior

`singularity pull docker://brainlife/freesurfer:latest` completes without errors or warnings

### Actual behavior

Spits out multiple:

`WARNING: Warning reading tar header: Linkname in pax header can't be converted to current locale.`

... from libarchive code. Container still runs despite warnings.

### Steps to reproduce behavior

Use development branch, run `singularity singularity pull docker://brainlife/freesurfer:latest`

----

This seems to be because when environment is sanitized, singularity is setting `LANG=C` in `src/lib/runtime/environment/environment.c`

Docker layer tar.gz files generally need to be extracted with a UTF8 $LANG set to avoid error. This was an issue previously observed for the docker/moby code base and fixed by forcing `LANG=en_US.UTF-8` when dockerd is started. See:

https://github.com/moby/moby/issues/355
https://github.com/moby/moby/commit/521ce07859ad09032268793b71e1e5f007e203b5

I can make a PR here - but is the preferred approach to either:

- Handle in the docker-extract.c code
- Change default sanitized $LANG to en_US.UTF-8

Not sure what the impact of not using C as the default may be?

Apologies for not catching this during the PR for the docker-extraction code. Came up almost as soon as that had been merge :cry:

dtrudg:
Ahh - on the moby issue they did some digging - when LANG is C

>bsdtar cannot convert to UTF-8 (when packing) or from UTF-8 (when unpacking) it will disregard the UTF-8 name contained in the pax header, and use the raw name (and it will work fine), but it will display a warning

Also later they moved the forced dockerd environment to a force for bsdtar calls only

https://github.com/moby/moby/commit/078a19d7255a7f9f50c4071a6b0811f4a996bd3e

We are using libarchive here, which is what bsdtar is based on.

GodloveD:
Does this just result in warnings or does it actually change something about the container?

I am also not sure about the consequences of changing LANG from C to en_US.UTF-8.  Why don't you try handling it in the docker-extract.c code and see if that works as expected.

dtrudg:
It seems to just be warnings - everything is extracted as far as I can see - and from what the docker folks found:

[...]

PR #1261, opened 2018-01-20
docker-extract - fix retval bug and locale warnings

dtrudg:

2 fixes for the libexec/docker-extract executable

1. BUG Fix - Fix retval handling bug - in some multi-stage build docker containers a whiteout marker may be present in next layer, but the whiteout target does not exist in extraction dir. docker-extract was exiting with error here, as retval contained a prior non-zero value from snprintf return code. Fix by resetting retval to 0 after string handling so that if marker present, but target missing we don't abort. This bug was observed with `nvidia/digits:5.0`

2. Fix for issue #1260 - libarchive tries to use UTF8 header filenames if present in archive. If running in non-UTF8 locale it will drop to using binary filenames and extract OK, but raises warnings. Singularity forces C locale by default for libexec runs. To silence warnings setlocale to en_US.UTF-8 (most commonly available UTF8 locale), or C.UTF-8 (available in some minimal debian installs / musl distros when en_US.UTF-8 is not available). Fixes warnings observed with `singularity pull docker://brainlife/freesurfer:latest`

- Ref: #1260 

**Checkoff for all PRs:**

- [X] I have added changes to the [CHANGELOG](https://github.com/singularityware/singularity/blob/development/CHANGELOG.md) and and documentation updates to the [singularityware] (https://www.github.com/singularityware/singularityware.github.io) documentation base. - _Covered by previous PR_

Attn: @singularityware-admin @GodloveD

### DIFF
--- a/src/docker-extract.c
+++ b/src/docker-extract.c
@@ -135,13 +135,15 @@ int apply_whiteouts(char *tarfile, char *rootfs_dir) {
             singularity_message(DEBUG, "Opaque Marker %s\n", pathname);
             errcode = apply_opaque(pathname, rootfs_dir);
             if (errcode != 0) {
-                break;
+                singularity_message(ERROR, "Error applying opaque marker from docker layer.\n");
+                ABORT(255);
             }
         } else if (strstr(pathname, "/.wh.")) {
             singularity_message(DEBUG, "Whiteout Marker %s\n", pathname);
             errcode = apply_whiteout(pathname, rootfs_dir);
             if (errcode != 0) {
-                break;
+                singularity_message(ERROR, "Error applying whiteout marker from docker layer.\n");
+                ABORT(255);
             }
         }
     }
--- a/src/docker-extract.c
+++ b/src/docker-extract.c
@@ -8,6 +8,7 @@
 #include <sys/file.h>
 #include <sys/param.h>
 #include <sys/stat.h>
+#include <locale.h>
 #include <archive.h>
 #include <archive_entry.h>
 
@@ -299,6 +300,15 @@ int main(int argc, char **argv) {
     int retval = 0;
     char *rootfs_dir = singularity_registry_get("ROOTFS");
     char *tarfile = NULL;
+
+    // Set UTF8 locale so that libarchive doesn't produce warnings for UTF8
+    // names - en_US.UTF-8 is most likely to be available
+    if( setlocale(LC_ALL, "en_US.UTF-8") == NULL ) {
+        // Fall back to C.UTF-8 for super-minimal debian and musl based distros
+        if (setlocale(LC_ALL, "C.UTF-8") == NULL ) {
+            singularity_message(WARNING, "Could not set a UTF8 locale, layer extraction may produce warnings\n");
+        }
+    }
 
     if (argc != 2) {
         singularity_message(ERROR, "Provide a single docker tar file to extract\n");
--- a/src/docker-extract.c
+++ b/src/docker-extract.c
@@ -40,8 +40,11 @@ int apply_opaque(const char *opq_marker, char *rootfs_dir) {
         ABORT(255);
     }
 
+    // Target may not exist - that's ok
+    retval = 0;
+
     if (is_dir(buf) == 0) {
-        s_rmdir(buf);
+        retval = s_rmdir(buf);
     }
 
     return retval;
@@ -86,6 +89,9 @@ int apply_whiteout(const char *wh_marker, char *rootfs_dir) {
         singularity_message(ERROR, "Error with pathname too long\n");
         ABORT(255);
     }
+
+    // Target may not exist - that's ok
+    retval = 0;
 
     if (is_dir(buf) == 0) {
         retval = s_rmdir(buf);


dtrudg:
@GodloveD - I maybe should emphasize here that this PR is important to get in to dev soon, if possible, as the (1) retval bug above makes the new docker-extract method fail on many of the containers that required the original whiteout PR.

Really sorry about this :-( - I introduced the bug when I was doing the string handling refactoring, and then got (un)-lucky with the containers I tested that with. I need to write up some notes and think about creating some artificial test-case containers to replicate this, that can reasonably be put in the travis tests. The nvidia/digits:5.0 etc. which trigger the above issue are unreasonably large to download in an automated test.

vsoch:
@dctrud you've worked so hard (on not just this!) chin up, we will get things figured out :) The dctrud pokemon is fearless! persistent!

```
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@@@@@@@@@@@@@@@@@..::::::::,,,,,:SS.++@@@@@@@@@@@@@@@@@@@@@@
@@@@@@@@@@@@,*::::::::::::::::::::.S.*..@@@@@@@@@@@@@@@@@@@@
@@@@@@@@@.,:::::.:::::::::::::::::.%@@.S@@@@+S@@@@@@@@@@@@@@
@@@@@@.,,,,,+::::::::::::::::::::::@@.@@@@:++%+.,,.@@@@@@@@@
@@@.@@@S@:,,,::::S:::::::::::::::::@.@,@,,,,,,..*.#*,S@@@@@@
@,+@@@@:,,,,.,,:::*::::::::::::::,,:@.,,,,,,++,,,,,,,,,,*@@@
@@@@@*@.@,@@,,+@+S@@@:,,,:::::::,,,,:,@,,,,,,++.*,***S@@@@@@
@@@@@@@+*@@@@@@@@@@@@@@@.,,,,::,,,,,,,+@,,,,,S.,:*@:@@@@@@@@
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@+,,,,,,,,@****.+S,:*:@@@@@@@**@
@@@@@@@@@@@@@@@@@@@@@@.,**:,,,,,*:,,,,,,*@@@@:%*,:@,.S+..+S@
@@@@@@@@@@@@@@@@@@@@@********.**+++.,,,,+@@@*,S:@+.****..@@@
@@@@@@@@@@@@@,@@S.%*..***.@@@@@@+.*@@,,::,@...*.+**.**:@@@@@
@@@@@@@@@@@@@@@,@@@@@@@@@@@@@@@@@@@@@@@****,**::,.*.%@@@@@@@
@@@@@@@@@@@@@@*...,S@@@@@@@@@@@@@@@@@@@@,**,,,,@*+@@@@@@@@@@
@@@@@@@@@@@@@@,***S@@,@@,.@@@@.+@@@@@@@,,,,,,@,+*@@@@@@@@@@@
@@@@@@@@@@@,%%**.%@@@@+,+,,,,.@,,@@@@@,,,,,,+@@@@:?:@@@@@@@@
@@@@@@@@@@+**S@@@@@@@.,,,,,,,,,,,,.:,,,,,,,,,@@@S@@,@@@@@@@@
@@@@@@@@@,*,,S@@@*+@,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,@@@@@@@@@
@@@@@@@@S:SSS.,,,::*,+::,S*,,,,,,,,,@@,,,,,,,,,,,,@@@@@@@@@@
@@@@@@.@@@@,,,,,,*,S,,,+S.*,,,@@,,,@@@@.,,,,,,,,,@@@@@@@@@@@
@@@@@@::@@,,,,.:,:::::+:***,,,,,,,,,,@,,.*.,,,,@@@@@@@@@@@@@
@@@@@@@@@@+:**...*....S****.**:,,,,,,,,*+*****+,@@@@@@@@@@@@
@@@@@@@@@@@@+*...S*....%*...*.+*::**.S*****%+*@@@@@@@@@@@@@@
@@@@@@@@@@@@@@%+...SS+..*:,*.*****S***.**:++@@@@@@@@@@@@@@@@
@@@@@@@@@@@@S*.****..,@.,,,,,,,*@******,,,,:@@@@@@@@@@@@@@@@
@@@@@@@@@@@@@@@@@@@@@@@@,,,%,,@@@@@@,+*.,,,,,@.@@@@@@@@@@@@@
@@@@@@@@@@@@@@@@@@@@@@@@@,,.,,@@@@@@@@@@S.+++++,,@@@@@@@@@@@
@@@@@@@@@@@@@@@@@@@@@@@@+,,,,,.@@@@@@@@@@@@@@@,:.:,@@@@@@@@@
@@@@@@@@@@@@@@@@@@@@@.,,,,,,,,,S@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@@@@@@@@@@@@@@@@@@@:..S,,,,,,S,::,*.@@@@@@@@@@@@@@@@@@@@@@@@
@@@@@@@@@@@@@@@@@@@@@@@@,@,+::::,.:,.@@@@@@@@@@@@@@@@@@@@@@@

dctrud
```

GodloveD:
@dctrud thanks for pinging me.  I thought this was just a fix for #1260.  I'm looking at this now.  Expect a review to appear within a day or two.

GodloveD:
I think it is very weird that whiteout files are present in one layer and the corresponding file or directory is not present in the previous layer.  Could this mean that the layers are being assembled out of order?  I know that we were having some trouble with that some time ago and I think it may have even been this container.  @vsoch are you able to shed any light on this?

dtrudg:
@GodloveD - I believe it occurs with multi-stage build and COPY in docker files. Take a look at: https://gitlab.com/nvidia/digits/blob/master/6.0/Dockerfile

Here we particularly have things like:

```
FROM ubuntu:14.04 as protobuf
<snip>
FROM nvidia/cuda:8.0-cudnn5-devel-ubuntu14.04 as caffe
COPY --from=protobuf /usr/local/protobuf /usr/local
```

.. but you're right that layer order probably needs to be double checked here. As mentioned above, I really want to try and make some artificial test containers for this issue. I'll try and verify the layer order this on this container this evening, and have a play to see if I can generate a container which exhibits the same thing, while being a lot less complex.

dtrudg:
Can confirm Singularity is using the correct order, same as docker:

## Docker layer order

```
$ docker pull nvidia/digits:6.0
6.0: Pulling from nvidia/digits
556ab54f6580: Already exists
6ca8fd6ef32a: Already exists
e64a7e7a2b21: Already exists
09ed7c05bfd3: Already exists
ec1faacd6fd9: Already exists
422dfa7049cb: Already exists
99cad67c81c1: Already exists
b46e83220e77: Already exists
1098bfe30a1f: Already exists
6db72aa10904: Already exists
6ab4736116af: Pull complete
960051667446: Pull complete
4ee1be86b477: Pull complete
d7291410cb41: Pull complete
6eb7027efbbe: Pull complete
12c848a346ba: Pull complete
2aa34da1126f: Pull complete
Digest: sha256:58f806c0b5887fcebcde6832cfcf67b3b415427d0619e6e8cea105cde111d781
Status: Downloaded newer image for nvidia/digits:6.0
```

## Singularity layer order 

Layers all extract in same order as docker

```
08:43 PM $ grep Exploding log
Exploding layer: sha256:556ab54f65806b5b518f81253c4daf5bb38dba3bf46cb6fd9b6d48e583fdd5e9.tar.gz
Exploding layer: sha256:6ca8fd6ef32a03f12d804935cf709712448d23f20308927adb635f4f3e38cc50.tar.gz
Exploding layer: sha256:e64a7e7a2b214e0085b5f7140a96e5a3beff9912ee4718390cd48579c1da08d6.tar.gz
Exploding layer: sha256:09ed7c05bfd36d5b21ca8d7a333ab91d5dcb8a8ea125b0d8b0058c82f48e7834.tar.gz
Exploding layer: sha256:ec1faacd6fd9f44056cf271350bb6f119e3f7dd7244f8a9b7d7b4cc1747b677c.tar.gz
Exploding layer: sha256:422dfa7049cba0a22f86e99259d672ff5db823fa43717a421b3dcef88f644e68.tar.gz
Exploding layer: sha256:99cad67c81c1ba6dfe2b45226747a7d6e83cfdcfe63fb3fbbdc23daa15e6e9c4.tar.gz
Exploding layer: sha256:b46e83220e77829548416918caddefb68ebb62b263407871172c019ae924404f.tar.gz
Exploding layer: sha256:1098bfe30a1f948538a2d5537e73f1d2219ae0060316f5d9503323681ce0946d.tar.gz
Exploding layer: sha256:6db72aa109047926a94f0a31a49ffb9b13c093fe3308212d3997b3ab6fff6268.tar.gz
Exploding layer: sha256:6ab4736116af95efb977bfb19fd86b586c3ec509f476f4420a04eaa256c44a06.tar.gz
Exploding layer: sha256:9600516674469b041a61fae764ca9a64fffb7965d4a46bdf655e9b369aff1de3.tar.gz
Exploding layer: sha256:4ee1be86b477d50b64cd8c88cd70b1c533e6539af1ca8ab7dca0ea35ea194acc.tar.gz
Exploding layer: sha256:d7291410cb419c73411f83ea9e35d4e42e72b1d4d9ab5eaaf32a65ed5a5f2b24.tar.gz
Exploding layer: sha256:6eb7027efbbe3dc830dd36de4c9000bef1d5cb396bf1c2d325ca67d6497528c4.tar.gz
Exploding layer: sha256:12c848a346badc1ddd9fe24f3a5851b7915428185db8df8bb8adb7e5e39a96a6.tar.gz
Exploding layer: sha256:2aa34da1126f5a7c0de6d1f16905a2363cdb604935f5128c3cb9d3104977291e.tar.gz
Exploding layer: sha256:bc9a2fa0bf4d4be8f68ef7174f52c84908d2871b7a44f6eda3b282a37b0173dc.tar.gz
```

## Error applying whiteout

Without this PR we are okay up to  422dfa704 which is layer 6 - this extracts OK.

```
DEBUG   [U=1000,P=22523]   main()                                    Extracting docker tar file /home/dave/.singularity/docker/sha256:422dfa7049cba0a22f8
6e99259d672ff5db823fa43717a421b3dcef88f644e68.tar.gz
```

Then we start handling layer 99cad67c and hit the first problem as we apply the whiteouts from this layer onto our container image which has been extracted okay up to 422dfa70

```
Exploding layer: sha256:99cad67c81c1ba6dfe2b45226747a7d6e83cfdcfe63fb3fbbdc23daa15e6e9c4.tar.gz
<snip>
DEBUG   [U=1000,P=22526]   main()                                    Applying whiteouts for tar file /home/dave/.singularity/docker/sha256:99cad67c81c1ba
6dfe2b45226747a7d6e83cfdcfe63fb3fbbdc23daa15e6e9c4.tar.gz
DEBUG   [U=1000,P=22526]   apply_whiteouts()                         Opaque Marker usr/local/cuda-8.0/.wh..wh..opq
ERROR   [U=1000,P=22526]   main()                                    Error applying layer whiteouts
ABORT   [U=1000,P=22526]   main()                                    Retval = 255
```

This is a whiteout dir opaque marker, telling us that we need to remove `usr/local/cuda-8.0/` before extraction - but `usr/local/cuda-8.0/` doesn't exist from the layers we've extracted so far, hence the error.

Here is the dockerfile for nvidia/digits:6.0 https://gitlab.com/nvidia/digits/blob/master/6.0/Dockerfile

When I extract the tar.gz for layer 99cad67 I find it contains files that you'd get from installing cuda 8.0, and cuda 8.0 only:

```
etc/
etc/ld.so.cache
etc/ld.so.conf.d/
etc/ld.so.conf.d/cuda-8-0.conf
tmp/
usr/
usr/local/
usr/local/cuda
usr/local/cuda-8.0/
usr/local/cuda-8.0/.wh..wh..opq
usr/local/cuda-8.0/LICENSE
usr/local/cuda-8.0/README
usr/local/cuda-8.0/doc/
usr/local/cuda-8.0/doc/EULA.txt
usr/local/cuda-8.0/lib64
usr/local/cuda-8.0/targets/
usr/local/cuda-8.0/targets/x86_64-linux/
-----SNIP-----
```

Then we end up going down a rabbit hole.... the CUDA stuff comes in from:

```
FROM nvidia/cuda:8.0-cudnn5-runtime-ubuntu14.04
```

Dockerfile for that is: https://gitlab.com/nvidia/cuda/blob/ubuntu14.04/8.0/runtime/cudnn5/Dockerfile - which has:

```
ARG repository
FROM ${repository}:8.0-runtime-ubuntu14.04
```

Dockerfile for that is:  https://gitlab.com/nvidia/cuda/blob/ubuntu14.04/8.0/runtime/Dockerfile

And then we finally see that layer 99cad67c8 is created in this Dockerfile by the RUN command:

```
RUN apt-get update && apt-get install -y --no-install-recommends \
        cuda-nvrtc-$CUDA_PKG_VERSION \
        cuda-nvgraph-$CUDA_PKG_VERSION \
        cuda-cusolver-$CUDA_PKG_VERSION \
        cuda-cublas-8-0=8.0.61.2-1 \
        cuda-cufft-$CUDA_PKG_VERSION \
        cuda-curand-$CUDA_PKG_VERSION \
        cuda-cusparse-$CUDA_PKG_VERSION \
        cuda-npp-$CUDA_PKG_VERSION \
        cuda-cudart-$CUDA_PKG_VERSION && \
    ln -s cuda-8.0 /usr/local/cuda && \
    rm -rf /var/lib/apt/lists/*
```

At this point I don't really know how the `usr/local/cuda-8.0/.wh..wh..opq` got into this layer. There was definitely no CUDA install earlier than this RUN - so that's why we have the error trying to remove the directory that doesn't exist.

So - the problem is actually with the container `nvidia/cuda:8.0-runtime-ubuntu14.04` - and we can reproduce from there more simply, instead of the DIGITS container:

```
09:04 PM $ singularity pull docker://nvidia/cuda:8.0-runtime-ubuntu14.04
WARNING: pull for Docker Hub is not guaranteed to produce the
WARNING: same image on repeated pull. Use Singularity Registry
WARNING: (shub://) to pull exactly equivalent images.
Docker image path: index.docker.io/nvidia/cuda:8.0-runtime-ubuntu14.04
ERROR  : Error applying layer whiteouts
ABORT  : Retval = 255
Cleaning up...
ERROR: pulling container failed!
```

The PR fixes this error, by not aborting when the dir marked for opaque whiteout is missing - but I'm none the wiser why the crazy whiteout marker is there in the docker layer file for 99cad67c8 - it doesn't seem like it should be.

Do we need to dig why this happens (docker weirdness??) - or OK to use this PR to stop it causing failures?

dtrudg:
NB - there is a good explanation of opaque whiteouts at: https://github.com/opencontainers/image-spec/blob/master/layer.md#opaque-whiteout

src/docker-extract.c
```diff
@@ -40,8 +41,11 @@ int apply_opaque(const char *opq_marker, char *rootfs_dir) {
         ABORT(255);
     }
 
+    // Target may not exist - that's ok
+    retval = 0;
+
     if (is_dir(buf) == 0) {
-        s_rmdir(buf);
+        retval = s_rmdir(buf);
```
GodloveD, line 48: I actually don't understand how this was working without the `retval = ` in this line.  I didn't appreciate before that `retval` would be expected to be >0 at this point.  So before this edit was `apply_opaque` just returning >0 on success?

GodloveD: Another question.  Right now, a `retval` !=0 will trigger a `break` in line 138 ending the `while` loop that is iterating through the tarred layer right?  As I understand it, this will just end evaluation of the current layer and move on to the next one.  Don't we actually want to `ABORT` if a non-zero value is returned from `apply_opaque`?   If not, why not?

Same question for the break on line 144.

src/docker-extract.c
```diff
@@ -294,6 +301,15 @@ int main(int argc, char **argv) {
     char *rootfs_dir = singularity_registry_get("ROOTFS");
     char *tarfile = NULL;
 
+    // Set UTF8 locale so that libarchive doesn't produce warnings for UTF8
+    // names - en_US.UTF-8 is most likely to be available
+    if( setlocale(LC_ALL, "en_US.UTF-8") == NULL ) {
```
GodloveD, line 308: Is the locale going to revert back to LC_ALL=C at the conclusion of this function, or will it remain changed until we exit the C code?  If it does not revert, how about setting it back to C when you are done?  There is an example for how to do this safely [here](https://www.gnu.org/software/libc/manual/html_node/Setting-the-Locale.html).

GodloveD (changes requested):
Hi @dctrud.  Thanks very much for ensuring that the layers are being handled in the same order as they are in docker.  No I don't think it is necessary to figure out the root cause of the weirdness on docker's side of things. 

I have some questions about this PR.  Please see them inline.  Thanks!

src/docker-extract.c
dtrudg: Yeah - when I simplified the code per Yannick's suggestions I overlooked retval was then going to be containing a non-zero snprintf return on successful string handling. It was indeed all screwed up here.

You are right on the ABORT too - we should stop layer extraction entirely.

src/docker-extract.c
dtrudg: We're in main here - so when we drop out, our short-lived docker-extract process is gone, and we're back to the shell script that called it (which will be under whatever locale it was initiated with). The UTF8 won't propagate outwards to anything outside of our current docker-extract instance.

src/docker-extract.c
dtrudg: Now ABORTs in 6ecdd72

src/docker-extract.c
GodloveD: Right.  And the two shell scripts that call this are `build.exec` and the docker image handler so we are good.  Thanks.

GodloveD (approved):

GodloveD:
OK.  I'm going to ask for a second review by @yhcote but this looks good to me.

yhcote (approved):
Look good to me

GodloveD:
ohhyeahh

dtrudg:
Thanks for getting this merged. Cheers!

PR #1295, opened 2018-02-12
Force write perms for docker builds as user

dtrudg:

Some docker layer .tar.gz files contain files or dirs that, when extracted, have -r-r-r permissions (no write for the owner). When building a singularity image as a user, we can't manipulate these dirs/files during extraction of later layers, and the build will fail.

#977 has examples from multiple users including:

```
singularity pull docker://jakirkham/centos_conda
singularity -vv build es_test.img docker://docker.elastic.co/elasticsearch/elasticsearch-oss:6.1.3
```

This patch force the perms of extracted files to include user write, if we are not running as root.

In conjunction with the previous whiteout fixes this allows the cases above to work as expected.

- Ref: #977 

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/src/docker-extract.c
+++ b/src/docker-extract.c
@@ -194,6 +194,7 @@ int extract_tar(const char *tarfile, const char *rootfs_dir) {
     struct archive *a;
     struct archive *ext;
     struct archive_entry *entry;
+    mode_t perms;
     int flags;
     int r;
     char *orig_dir;
@@ -253,6 +254,14 @@ int extract_tar(const char *tarfile, const char *rootfs_dir) {
         if (strstr(pathname, "/.wh.") || pathtype == AE_IFSOCK ||
             pathtype == AE_IFCHR || pathtype == AE_IFBLK || pathtype == AE_IFIFO) {
             continue;
+        }
+
+        // Issue 977 - Force write perms needed for user builds
+        if(getuid() != 0) {
+            perms = archive_entry_perm(entry);
+            if( (perms & S_IWUSR) != S_IWUSR) {
+               archive_entry_set_perm(entry, perms | S_IWUSR);
+            }
         }
 
         r = archive_write_header(ext, entry);
--- a/src/docker-extract.c
+++ b/src/docker-extract.c
@@ -258,10 +258,17 @@ int extract_tar(const char *tarfile, const char *rootfs_dir) {
 
         // Issue 977 - Force write perms needed for user builds
         if(getuid() != 0) {
+          #if ARCHIVE_VERSION_NUMBER <= 3000000
+            perms = archive_entry_mode(entry);
+            if( (perms & S_IWUSR) != S_IWUSR) {
+               archive_entry_set_mode(entry, perms | S_IWUSR);
+            }
+          #else
             perms = archive_entry_perm(entry);
             if( (perms & S_IWUSR) != S_IWUSR) {
                archive_entry_set_perm(entry, perms | S_IWUSR);
             }
+          #endif
         }
 
         r = archive_write_header(ext, entry);
--- a/src/docker-extract.c
+++ b/src/docker-extract.c
@@ -258,17 +258,17 @@ int extract_tar(const char *tarfile, const char *rootfs_dir) {
 
         // Issue 977 - Force write perms needed for user builds
         if(getuid() != 0) {
-          #if ARCHIVE_VERSION_NUMBER <= 3000000
+#if ARCHIVE_VERSION_NUMBER <= 3000000
             perms = archive_entry_mode(entry);
             if( (perms & S_IWUSR) != S_IWUSR) {
-               archive_entry_set_mode(entry, perms | S_IWUSR);
+                archive_entry_set_mode(entry, perms | S_IWUSR);
             }
-          #else
+#else
             perms = archive_entry_perm(entry);
             if( (perms & S_IWUSR) != S_IWUSR) {
-               archive_entry_set_perm(entry, perms | S_IWUSR);
+                archive_entry_set_perm(entry, perms | S_IWUSR);
             }
-          #endif
+#endif
         }
 
         r = archive_write_header(ext, entry);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,8 @@ and changes prior to that are (unfortunately) done retrospectively. Critical ite
 ### Bug Fixes
  - Put /usr/local/{bin,sbin} in front of the default PATH
  - Fixed bug that did not export environment variables for apps with "-" in name
- 
+ - Fix non-root build from docker containers with non-writable file/dir permissions
+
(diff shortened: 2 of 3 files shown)


dtrudg:
Hi @GodloveD - this is passing CI, and I've tested it on CentOS 6 to verify it's working as expected with the very old libarchive there, plus latest Fedora for a very new libarchive.

Your PR #1220 probably needs to be merged with this one to avoid further weirdness with some containers. E.g. the elasticsearch container mentioned in #977 will now build, but gives a permission denied error on run if only this PR is applied. With #1220 merged as well, it gets further, until failing with a 'read-only filesystem' error, which is expected.

jakirkham:
Thanks for doing this, @dctrud. This works on the image I gave and a couple others.

However am having trouble with the [`nanshe/nanshe_workflow`]( https://hub.docker.com/r/nanshe/nanshe_workflow/ ) image. The image converts fine and it also starts ok initially. However during startup it runs into some whiteout file in the image that causes the container to crash. Haven't had this issue when using [docker2singularity]( https://github.com/singularityware/docker2singularity ). Though have no idea if the issue is caused by this PR or something else introduced in `master`.

Edit: If I build the image as an ext3 writable image and run it as such, that last issue goes away. Looks like the entrypoint is modifying the files in some unintended way. Looks like `pkg_resources` may be to blame (inside the container). Not entirely sure why. 😕

dtrudg:
Hi @jakirkham - could you open an issue on this, or append some more detail to #977 if you are seeing errors? Running this PR I don't have an issue - but I do need to run in a writable sandbox, as the web app is trying to write into /run inside the container:

```
$ singularity build -s sandbox_nanshe/ docker://nanshe/nanshe_workflow

$ singularity run -w sandbox_nanshe/
[WARN  tini (28026)] Tini is not running as PID 1 .
Zombie processes will not be re-parented to Tini, so zombie reaping won't work.
To fix the problem, run Tini as PID 1.
fatal: Not a git repository (or any parent up to mount point /home)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
Copying workflows from source to destination...
    Skipped: nanshe_ipython.ipynb
Completed copying of workflows.
[I 17:55:38.161 NotebookApp] Writing notebook server cookie secret to /run/user/1000/jupyter/notebook_cookie_secret
[W 17:55:38.508 NotebookApp] WARNING: The notebook server is listening on all IP addresses and not using encryption. This is not recommended.
[I 17:55:38.526 NotebookApp] Loading IPython parallel extension
[I 17:55:38.547 NotebookApp] [jupyter_nbextensions_configurator] enabled 0.4.0
[I 17:55:38.547 NotebookApp] Serving notebooks from local directory: /home/dave
[I 17:55:38.547 NotebookApp] 0 active kernels
[I 17:55:38.547 NotebookApp] The Jupyter Notebook is running at:
[I 17:55:38.547 NotebookApp] http://[all ip addresses on your system]:8888/?token=7d208d3bd8a3b03d7e4840aa343b18515272910bd3355183
[I 17:55:38.547 NotebookApp] Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
[C 17:55:38.548 NotebookApp]

    Copy/paste this URL into your browser when you connect for the first time,
    to login with a token:
        http://localhost:8888/?token=7d208d3bd8a3b03d7e4840aa343b18515272910bd3355183
```

jakirkham:
np. Agreed, it is off topic. Wasn't obvious to me initially.

jakirkham:
Did a little more debugging. Seems `.wh*` (whiteout files) are being included in the Singularity image that were not before. Knowing better the issue, expect it is totally unrelated to this PR. Have raised as issue ( https://github.com/singularityware/singularity/issues/1301 ).

dtrudg:
Note for review - issue raised by .@jakirkham above is discussed in #1301 and is unrelated to this PR's code changes.

GodloveD:
@fearful-symmetry is there any way you could test this PR out and make sure it fixes the issue you pointed out with the elastic search container.  Thank you!  😄

src/docker-extract.c
```diff
@@ -255,6 +256,21 @@ int extract_tar(const char *tarfile, const char *rootfs_dir) {
...
+            if( (perms & S_IWUSR) != S_IWUSR) {
+               archive_entry_set_mode(entry, perms | S_IWUSR);
+            }
+          #else
+            perms = archive_entry_perm(entry);
+            if( (perms & S_IWUSR) != S_IWUSR) {
+               archive_entry_set_perm(entry, perms | S_IWUSR);
+            }
+          #endif
+        }
```
GodloveD: Could this be refactored to remove some duplicate lines like this:
```
perms = archive_entry_mode(entry);
if( (perms & S_IWUSR) != S_IWUSR) {
    #if ARCHIVE_VERSION_NUMBER <= 3000000
    	archive_entry_set_mode(entry, perms | S_IWUSR);
    #else
    	archive_entry_set_perm(entry, perms | S_IWUSR);
    #endif
}
```
Didn't test, so that might not work, but you get the idea.

And also, could we get 4 space block indents instead of 2 for consistency with the surrounding code?

GodloveD (changes requested):

src/docker-extract.c
dtrudg: Sorry about the indents... not sure how I ended up with 2 there.

r.e. refactoring, it'd then have 2 separate #if #endif blocks, like this:

```
#if ARCHIVE_VERSION_NUMBER <= 3000000
    perms = archive_entry_mode(entry);
#else
    perms = archive_entry_perm(entry);
#endif
if( (perms & S_IWUSR) != S_IWUSR) {
    #if ARCHIVE_VERSION_NUMBER <= 3000000
    	archive_entry_set_mode(entry, perms | S_IWUSR);
    #else
    	archive_entry_set_perm(entry, perms | S_IWUSR);
    #endif
}
```

I thought the single #if #else block was easier to read, but don't mind - what's your preference :-)?

GodloveD: Your absolutely correct.  I missed that detail on my first read.  

OK.  I'm good.  I think @yhcote or another of our more experienced C coders should read it too.

GodloveD (approved):

GodloveD:
@yhcote could you have a look at this too since you are familiar with this issue/code already?

src/docker-extract.c
```diff
@@ -255,6 +256,21 @@ int extract_tar(const char *tarfile, const char *rootfs_dir) {
             continue;
         }
 
+        // Issue 977 - Force write perms needed for user builds
```
yhcote, line 259: For clarity, it probably would have been 'better' to define a macro perhaps in the include file, to deal with the conditional compilation of that code, but that work.

yhcote (approved):

dtrudg:
Thanks @GodloveD and @yhcote!

PR #1296, opened 2018-02-13
Test Suite CentOS 6 support and fixes

dtrudg:

Support CentOS 6 for `make test` and travis rpmbuild tests. This is important now that we have integration of libs such as libarchive (used by docker-extract) that are quite different in older versions.

Changes:

- Let travis run test suite on CentOS 6
- Skip only python tests where python <2.7 (CentOS 6 has 2.6 and test code is not compatible)
- Skip arch linux container test if we are on a 2 series kernel (will get kernel too old on CentOS 6)
- Fix `SINGULARITY_OVERLAY_FS` check - this variable is always defined, 0 or 1 - so check for that, not `-n`
- Fix busybox non-overlay user bind test - `/opt` doesn't actually exist in the busybox container, so we have to user bind somewhere else to check functionality without overlay. Use `/var/tmp` which is available and empty.

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/tests/32-action_options.sh
+++ b/tests/32-action_options.sh
@@ -54,7 +54,7 @@ stest 0 mkdir -p "$TESTDIR"
 stest 0 touch "$TESTDIR/testfile"
 stest 0 singularity exec --home "$TESTDIR" "$CONTAINER" test -f "$TESTDIR/testfile"
 stest 0 singularity exec --home "$TESTDIR:/home" "$CONTAINER" test -f "/home/testfile"
-if [ -n "${SINGULARITY_OVERLAY_FS:-}" ]; then
+if [  "$SINGULARITY_OVERLAY_FS" = "1" ]; then
     stest 0 singularity exec --contain --home "$TESTDIR:/blah" "$CONTAINER" test -f "/blah/testfile"
 fi
 stest 0 sh -c "echo 'cd; test -f testfile' | singularity exec --home '$TESTDIR' '$CONTAINER' /bin/sh"
--- a/.travis/centos6_setup
+++ b/.travis/centos6_setup
@@ -1,0 +1,9 @@
+#!/bin/bash -ex
+
+mkdir /run
+mkdir /tmp/yum-download
+yum -y --downloadonly --downloaddir /tmp/yum-download install util-linux-ng
+cd /tmp/yum-download
+rpm2cpio util-linux-ng*.rpm | cpio -idmv
+mv bin/mount /bin/mount
+
--- a/.travis/centos_run
+++ b/.travis/centos_run
@@ -17,5 +17,12 @@ SCRIPT
 cd /tmp/$DIST/singularity
 
 # mounting /run for /dev/shm (python multiprocessing)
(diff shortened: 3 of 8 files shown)


dctrud:
Close/re-open to re-trigger travis.

jmstover (approved):
This all looks good to me.

tests/32-action_options.sh
```diff
@@ -54,7 +54,7 @@ stest 0 mkdir -p "$TESTDIR"
 stest 0 touch "$TESTDIR/testfile"
 stest 0 singularity exec --home "$TESTDIR" "$CONTAINER" test -f "$TESTDIR/testfile"
 stest 0 singularity exec --home "$TESTDIR:/home" "$CONTAINER" test -f "/home/testfile"
-if [ -n "${SINGULARITY_OVERLAY_FS:-}" ]; then
+if [  "$SINGULARITY_OVERLAY_FS" = "1" ]; then
```
GodloveD, line 57: Why this change?  Seems like the original line is more robust right?

GodloveD (commented):
Just looking this over.  Have not tested yet.  One quick question about conditional test in `action_options.sh`

tests/32-action_options.sh
dtrudg: The original line doesn't work..

> Fix SINGULARITY_OVERLAY_FS check - this variable is always defined, 0 or 1 - so check for that, not -n

In `test.sh.in` that variable is always getting defined to 0 or 1... so the original test for nonzero length always has the same outcome whether or not overlay is available. Here's the test.sh.in bit:

```bash
SINGULARITY_OVERLAY_FS="0"
KVERS=`uname -r`
if test -f "/lib/modules/$KVERS/modules.dep"; then
    if grep -q 'overlay.ko' "/lib/modules/$KVERS/modules.dep"; then
        SINGULARITY_OVERLAY_FS="1"
    fi
fi
```

GodloveD:
@dctrud can you rebase this plz.  I'm getting this:
```
configure: error: Unable to find the uuid headers, need package libuuid-devel (uuid-dev on Debian/Ubuntu)
```

dtrudg:
@GodloveD - yup, next on my list.. working through it :-)

dtrudg:
@GodloveD - this will be a by Monday if I have time thing. The travis setup did change a lot onto develpoment. May not be a 2.5.

tests/32-action_options.sh
GodloveD: Ah.  Good catch.

GodloveD:
OK.  PM me if you change your mind and want this reviewed.

dtrudg:
@GodloveD - Dropping this from the 2.5 milestone as I'm not going to be able to get the travis stuff sorted by this PM. Noting that running test content on CentOS6 locally is important++ to ensure 2.x libarchive behaviour is good.

GodloveD:
@dctrud any chance you could rebase this against release-2.5?

dtrudg:
@GodloveD - will do my best on this - it's quite a big change from the old development to 2.5 for this, prob a good half-day or more messing as the build setup is a lot different (old development had the secure build stuff).

dtrudg:
Closing this as out-of-date. Rebased PR #1621 is in progress.

PR #1305, opened 2018-02-14
Close file descriptors pointing to a directory

cclerget:

Singularity doesn't cleanup file descriptor table during runtime. This is especially important with file descriptor pointing to directory, because that could break container isolation with some scenarios

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/src/util/util.c
+++ b/src/util/util.c
@@ -453,3 +453,32 @@ struct tempfile *make_logfile(char *label) {
     
     return(tf);
 }
+
+// close all file descriptors pointing to a directory
+void fd_cleanup(void) {
+    struct stat st;
+    char *fd_path = (char *)malloc(128);
+    int i;
+
+    singularity_message(DEBUG, "Cleanup file descriptor table\n");
+
+    if ( fd_path == NULL ) {
+        singularity_message(ERROR, "Failed to allocate memory\n");
+        ABORT(255);
+    }
+
+    for ( i = 0; i <= sysconf(_SC_OPEN_MAX); i++ ) {
+        int length;
+        length = snprintf(fd_path, 127, "/proc/self/fd/%d", i);
+        fd_path[length] = '\0';
+
+        if ( stat(fd_path, &st) < 0 ) {
+            continue;
+        }
+        if ( S_ISDIR(st.st_mode) ) {
+            close(i);
+        }
+    }
+
+    free(fd_path);
+}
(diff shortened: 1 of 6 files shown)


GodloveD:
I think that @gmkurtzer reviewed this right @cclerget.  But he must have sent you comments offline.

src/util/util.c
```diff
@@ -453,3 +454,29 @@ struct tempfile *make_logfile(char *label) {
     
     return(tf);
 }
+
+// close all file descriptors pointing to a directory
+void fd_cleanup(void) {
+    char *fd_path = (char *)malloc(128);
```
jmstover: I'm doing this from memory ... so I could be forgetting something, or not remembering something right...

`char *fd_path = (char *)malloc(128);`

Instead of hard coding `128` use `PATH_MAX`

src/util/util.c
```diff
@@ -453,3 +454,29 @@ struct tempfile *make_logfile(char *label) {
...
+
+    if ( fd_path == NULL ) {
+        singularity_message(ERROR, "Failed to allocate memory\n");
+        ABORT(255);
+    }
+
+    for ( i = 0; i <= sysconf(_SC_OPEN_MAX); i++ ) {
+        int length;
+        length = snprintf(fd_path, 127, "/proc/self/fd/%d", i);
+        fd_path[length] = '\0';
```
jmstover: There was a question @yhcote had here on the `snprintf()` usage and the null termination...

yhcote:
Yes,

fd_path[length] = '\0'; should be :
fd_path[127] = '\0';   or fd_path[PATH_MAX-1] = '\0';

This is because snprintf() returns how many chars the function "would" have
copied if there was enough room, which could be greater than the actual
buffer size.

cclerget:
@jmstover I used a smaller buffer to not allocate unnecessary memory, but you're right PATH_MAX is the default so probably better to use here.

@yhcote since buffer is reused, I null terminated string at length return by snprintf. snprintf normally null terminates string, so I'm wonder if it's safe to remove fd_path[length] = \0 there ?
Between there is no risk of out of bound with %d there but you're right maybe add a check on length if length > 127 and keep fd_path[length] to ensure string is always null terminated ?

cclerget:
@jmstover @yhcote Changes applied. Thank you for your review !

yhcote (approved):
Looks good to me

PR #1317, opened 2018-02-19
Fix bad "doesn't exists" grammar, add more helpful error for unrecogized URIs

DrDaveD:

This PR removes 3 occurrences of "doesn't exists" in the singularity source code and adds a more helpful error message in image.c when some unrecognized URI is attempted.  

I started down this path by trying "docker://" on a RHEL system that had only singularity-runtime sub-rpm installed.  This doesn't completely fix the problems with that, but it's a hopefully non-controversial and simple start and helpful on its own.

- Ref: None

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/src/lib/image/image.c
+++ b/src/lib/image/image.c
@@ -61,7 +61,7 @@ struct image_object singularity_image_init(char *path, int open_flags) {
 
     real_path = realpath(path, NULL); // Flawfinder: ignore
     if ( real_path == NULL ) {
-        singularity_message(ERROR, "Image path doesn't exists\n");
+        singularity_message(ERROR, "Image path doesn't exist\n");
         ABORT(255);
     }
 
--- a/src/lib/image/image.c
+++ b/src/lib/image/image.c
@@ -52,6 +52,7 @@
 
 struct image_object singularity_image_init(char *path, int open_flags) {
     struct image_object image;
+    char *urip;
     char *real_path;
 
     if ( path == NULL ) {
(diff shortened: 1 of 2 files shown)


src/lib/image/image.c
```diff
@@ -52,16 +52,25 @@
...
     if ( path == NULL ) {
         singularity_message(ERROR, "No container image path defined\n");
         ABORT(255);
     }
 
+    urip = strstr(path, "://");
+
+    if (urip != NULL) {
+        *(urip + 3) = '\0';
+        singularity_message(ERROR, "Image paths beginning with %s are not supported\n", path);
```
dtrudg, line 67: ~~Do you mean to be using `urip` in the format string instead of `path` - from the setting of the `\0` above it looks like maybe this is intended?~~

dtrudg (dismissed):

src/lib/image/image.c
jmstover: I think using path is right ... though I don't think the output will be what's expected. This looks like it's just being sure that `path` isn't a URI (correct me if I'm wrong here).

But the message should print out the entire thing... like if somehow `docker://ubuntu:latest' got passed here, then the error message would be:

`Image paths beginning with docker://ubuntu:latest are not supported` ... if `urip` was used, the output should be:
`Image paths beginning with :// are not supported`

Since it's doing a check for `://` I'd change the output to something like:

`A URI location is invalid as an image path`

If `urip` isn't being used you don't really need to terminate it... especially aborting right there.

dtrudg: Sorry, brain fade - I think this does make sense here, as the error should be specific since not all know what a URI location is:

`Image paths beginning with docker:// are not supported` which this will do.

dtrudg (commented):
This looks good to me, with the caveat that I'm not sure if we need to be paranoid r.e. any chance that `path` could come in without a null terminator.

Would @yhcote / @bauerm97 / @cclerget be able to give an opinion on this for 2nd review? Thanks.

DrDaveD:
@dctrud  said 
> This looks good to me, with the caveat that I'm not sure if we need to be paranoid r.e. any chance that path could come in without a null terminator.

Even if it did, the PR adds a null terminator, so it should not be a problem.

DrDaveD:
This has been rebased against development-2.x.  hostname.c didn't exist in this branch so there was no typo to fix there.

GodloveD (approved):
This looks good to me.  Let's include it in 2.5.

dtrudg (approved):
Looks good to me against development-2.x

PR #1320, opened 2018-02-20
Prevent unmounts of the filesystem containing the image

DrDaveD:

This PR restores the fix in issue #714.  It was released in version 2.3.2 and promptly dropped from version 2.4, I think in this [commit](b299f704bd880c4f893e9cba336bfa8ce97f418c) as files got moved around.

The issue is that singularity does not hold open a reference to the original image outside of the container, so the filesystem holding it can be unmounted.  This particularly causes problems on autofs-mounted filesystems, as is done with cvmfs.

- Ref: None

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/src/lib/image/image.c
+++ b/src/lib/image/image.c
@@ -118,10 +118,9 @@ struct image_object singularity_image_init(char *path, int open_flags) {
         ABORT(255);
     }
 
-    if ( fcntl(image.fd, F_SETFD, FD_CLOEXEC) != 0 ) {
-        singularity_message(ERROR, "Failed to set CLOEXEC on image file descriptor\n");
-        ABORT(255);
-    }
+    // Do not CLOEXEC the image file descriptor, as we want to make sure we
+    // are holding onto it when we are running contained processes to prevent
+    // it from getting unmounted.
 
     if ( ( singularity_suid_enabled() >= 0 ) && ( singularity_priv_getuid() != 0 ) ) {
         singularity_limit_container_paths(&image);
--- a/src/lib/image/image.c
+++ b/src/lib/image/image.c
@@ -118,9 +118,25 @@ struct image_object singularity_image_init(char *path, int open_flags) {
         ABORT(255);
     }
 
-    // Do not CLOEXEC the image file descriptor, as we want to make sure we
-    // are holding onto it when we are running contained processes to prevent
-    // it from getting unmounted.
+    if ( singularity_registry_get("UNSHARE_PID") == NULL ) {
+        // If not using a separate PID namespace, do not CLOEXEC the image
+        //   file descriptor, so we keep holding onto it when we are running
+        //   contained processes, to prevent the image from getting unmounted.
+        // When there is a separate PID namepace, the parent process keeps
+        //   running and holds it open, so it isn't needed to prevent image
+        //   unmounts.  More significantly, it would be a security hole to
+        //   leave that file descriptor open to code that is supposed to be
+        //   isolated.
+        // If in the future a separate parent process keeps running even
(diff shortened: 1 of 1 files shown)


cclerget:
@DrDaveD There is a conflict with PR #1305. A master process living outside container is required to handle this buggy autofs case securely.

DrDaveD:
@cclerget I wondered about that, because it was seeing that PR that reminded me to check this issue and that's when I found that the unmounting prevention fix disappeared in singularity-2.4.

I've now done more testing and found that it depends on whether or not you use the -p option.  With singularity-2.4.2, if you use -p, which we always do on the OSG, then unmounting is still prevented, but if you don't use -p, unmounting is not prevented.  I believe that's because with -p an extra process stays running outside of the container and it holds the image open.

PR #1305 was merged 4 days ago so I built the latest development branch and it has the same behavior as 2.4.2.   However, I rebased this PR on top of that, and with this PR even without -p unmounting is still prevented. 

So, I think it's still probably a good idea to merge this PR, although since we always use -p, it's not so important to me.  Since PR #1305 was intentionally closing all the file descriptors, I wonder why this PR still works.  Also, since it listed as a security fix in the CHANGELOG, apparently somebody thinks it ought to be unavailable inside the container, although I'm not sure why.

cclerget:
@DrDaveD Conflict in term of purpose, #1305 close file descriptor pointing to directory and your PR keep image file descriptor open, since image can be a directory those PRs conflicts.

>  Since PR #1305 was intentionally closing all the file descriptors, I wonder why this PR still works

You observe same behaviour with or without #1305 because file descriptor are closed early in code and images are open after this cleanup stage, with actual code it's fine since file descriptors are closed during execve.

> Also, since it listed as a security fix in the CHANGELOG, apparently somebody thinks it ought to be unavailable inside the container, although I'm not sure why.

Exposing image file or directory inside container is bad from security point, ways to exploit this have been identified. Singularity 3.0 will have a master process living in host namespaces which could keep file descriptors open while addressing issues related to autofs in a secure way without exposing image file or directory inside container.

For security reasons I can't approve this PR

DrDaveD:
Ok I can see the security issue when using --contain --pid --ipc; with an open directory file descriptor and fchdir one can escape from the container, and with an open file descriptor to an image file one could open /proc/$$/fd/4 read-write and be able to write to the image.  So the current incarnation of this PR is insecure, I agree.

On the other hand, I don't see any leaking file descriptors in 2.4.2 so I'm not sure what problem #1305 solves. 

Also, if someone is not using --pid then there are other ways to escape so I don't see how it matters then.  Since unmounting is already prevented with --pid (because the parent process stays around and keeps open the file descriptor), I could modify this PR to only skip setting CLOEXEC if --pid is not set.  Do you see a security problem with that?  In any event, it sounds like the 3.0 plan will solve the unmounting problem by always keeping an extra process outside the container, and we always use --pid anyway, so it probably might as well wait for 3.0.

cclerget:
@DrDaveD Right, that's a problem when running inside PID namespace, it should be fine to only keep CLOEXEC when entering in PID namespace

DrDaveD:
Ok @cclerget I updated the PR to set CLOEXEC if and only if --pid is set, and included a thorough explanatory comment.

DrDaveD:
This has been rebased to development-2.x

DrDaveD:
I rebased again against development-2.x.

DrDaveD:
Rebased again against current development-2.x

DrDaveD:
@godloved Please reopen this against release-2.6, I think it is still viable.  I have not rebased & force pushed so I think you should be able to do it.

PR #1329, opened 2018-02-22
putting a framework in place for cli

GodloveD:
Here is a a very rough framework for the cli.


ArangoGutierrez:
I think this work should be on `/cli` rather than on `/pkg/cmd.../ due everything that goes on `/pkg` is intended to be exposed as a package to be imported

ArangoGutierrez (dismissed):

bauerm97:
I think it should actually either go in `pkg/cli/` or if it's not intended to be exposed as a package, then it should be `internal/pkg/cli/`

bauerm97 (approved):
Awesome, merging. Thanks Dave

PR #1343, opened 2018-02-27
Semi-buildable version of pkg build

bauerm97:

`go build cmd/cli/cli.go` now builds, and with that so does `pkg build, image, and internal/cli`

Attn: @singularityware-admin

@tri-adam @ArangoGutierrez @dctrud


dtrudg:
Builds okay for me

```
12:45 PM $ go build cmd/cli/cli.go
12:45 PM $ ./cli

USAGE: singularity [global options...] <command> [command options...] ...

GLOBAL OPTIONS:
    -d|--debug    Print debugging information
    -h|--help     Display usage summary
    -s|--silent   Only print errors
    -q|--quiet    Suppress all normal output
    --version  Show application version
    -v|--verbose  Increase verbosity +1
    -x|--sh-debug Print shell wrapper debugging information
```

pkg/build/build_definition_json.go
```diff
@@ -0,0 +1,15 @@
...
+  your rights to use or distribute this software.
+*/
+package build
+
+// build_definition_json.go:
+// Provide transformation from JSON -> BuildDefinition
+
+func BuildDefinitionFromJSON(j []byte) BuildDefinition {
+	return BuildDefinition{}
+}
```
tri-adam, line 15: I'm not totally sure this file is necessary. The structure itself can describe how to encode/decode the JSON, and then we can just use the standard packages?

ArangoGutierrez: yeah, the JSON std pkg is great for building a JSON from a data structure.

bauerm97: I agree that we use the standard JSON package, and we'll definitely tag the struct itself, but I figured just for now we separate it into its own file. Like, the functions that do the (un)marshalling can go in this file. Just so we separate implementation of the `BuildDefinition` data structure from the parsing. Especially since `build_definition_deffile.go` is really big, I don't want to clutter everything into `build_definition.go`

I'm open to changing if the prevailing opinion is different

PR #1345, opened 2018-02-27
Make recursive chmods more efficient

DrDaveD:

I noticed that the build command when not running as root does a recursive chmod whether or not any modes need to be changed.  I looked for other places with similar code and found it in the archive-tar and archive-cpio handlers.  This PR changes the scripts to use 'find' to identify which files are the wrong permissions and only chmods those files.

- None

**Checkoff for all PRs:**

   I haven't done this because I don't know which release it will get into.  It is relevant to the release-2.4 branch.

Attn: @singularityware-admin

### DIFF
--- a/libexec/cli/build.exec
+++ b/libexec/cli/build.exec
@@ -356,7 +356,7 @@ fi
 if [ "$USERID" != "0" ]; then
     # This is required as some files in a container are not readable or writable
     # by owner and thus fail to build when not root.
-    find "$SINGULARITY_ROOTFS" ! -perm -u=rw -print0|xargs -0 -r chmod u+rw
+    find "$SINGULARITY_ROOTFS" ! -type l ! -perm -u=rw -print0|xargs -0 -r chmod u+rw
 fi
 
 if [ -z "${SINGULARITY_SANDBOX:-}" ]; then
--- a/libexec/cli/build.exec
+++ b/libexec/cli/build.exec
@@ -354,9 +354,9 @@ fi
 
 
 if [ "$USERID" != "0" ]; then
-    # This is required as some files in a container are not readable by owner and
-    # thus fail to build when not root.
-    chmod u+rw -R "$SINGULARITY_ROOTFS"
+    # This is required as some files in a container are not readable or writable
+    # by owner and thus fail to build when not root.
+    find "$SINGULARITY_ROOTFS" -type d ! -perm -u=rw -print0|xargs -0 -r chmod u+rw
 fi
 
 if [ -z "${SINGULARITY_SANDBOX:-}" ]; then
--- a/libexec/cli/build.exec
+++ b/libexec/cli/build.exec
@@ -356,7 +356,7 @@ fi
 if [ "$USERID" != "0" ]; then
     # This is required as some files in a container are not readable or writable
     # by owner and thus fail to build when not root.
-    find "$SINGULARITY_ROOTFS" -type d ! -perm -u=rw -print0|xargs -0 -r chmod u+rw
+    find "$SINGULARITY_ROOTFS" ! -perm -u=rw -print0|xargs -0 -r chmod u+rw
 fi
 
 if [ -z "${SINGULARITY_SANDBOX:-}" ]; then
--- a/libexec/handlers/archive-cpio.sh
+++ b/libexec/handlers/archive-cpio.sh
@@ -37,7 +37,7 @@ case "$SINGULARITY_IMAGE" in
     ;;
 esac
 
-find "$CONTAINER_DIR" ! -perm -u=w -print0|xargs -0 -r chmod u+w
+find "$CONTAINER_DIR" ! -type l ! -perm -u=w -print0|xargs -0 -r chmod u+w
 
 SINGULARITY_IMAGE="$CONTAINER_DIR"
 SINGULARITY_CLEANUPDIR="$SINGULARITY_TMPDIR"
--- a/libexec/handlers/archive-cpio.sh
+++ b/libexec/handlers/archive-cpio.sh
@@ -37,7 +37,7 @@ case "$SINGULARITY_IMAGE" in
     ;;
 esac
 
-chmod -R +w "$CONTAINER_DIR"
+find "$CONTAINER_DIR" ! -perm -u=w -print0|xargs -0 -r chmod u+w
 
 SINGULARITY_IMAGE="$CONTAINER_DIR"
 SINGULARITY_CLEANUPDIR="$SINGULARITY_TMPDIR"
--- a/libexec/handlers/archive-tar.sh
+++ b/libexec/handlers/archive-tar.sh
@@ -30,7 +30,7 @@ message 1 "Opening tar archive, stand by...\n"
 # running as a user.
 tar -C "$CONTAINER_DIR" -xf "$SINGULARITY_IMAGE" 2>/dev/null
 
-chmod -R +w "$CONTAINER_DIR"
+find "$CONTAINER_DIR" ! -perm -u=w -print0|xargs -0 -r chmod u+w
 
 SINGULARITY_IMAGE="$CONTAINER_DIR"
 SINGULARITY_CLEANUPDIR="$SINGULARITY_TMPDIR"
--- a/libexec/handlers/archive-tar.sh
+++ b/libexec/handlers/archive-tar.sh
@@ -30,7 +30,7 @@ message 1 "Opening tar archive, stand by...\n"
 # running as a user.
 tar -C "$CONTAINER_DIR" -xf "$SINGULARITY_IMAGE" 2>/dev/null
 
-find "$CONTAINER_DIR" ! -perm -u=w -print0|xargs -0 -r chmod u+w
+find "$CONTAINER_DIR" ! -type l ! -perm -u=w -print0|xargs -0 -r chmod u+w
 
 SINGULARITY_IMAGE="$CONTAINER_DIR"
 SINGULARITY_CLEANUPDIR="$SINGULARITY_TMPDIR"


jmstover:
@DrDaveD Err... What is the use case here? In most cases this will be slower as `find` has to do a `stat()` call on every file in the tree to compare the permissions, where `chmod -R` just sets without looking.

From a quick test I did ... 50 files in the directory, 25 were set `chmod 444`:

```
$ time { find . ! -perm -u=w -print0 | xargs -0 -r chmod u+w; }
real    0m0.003s
user    0m0.001s
sys    0m0.000s

$ time { chmod -R u+w .; }
real    0m0.001s
user    0m0.000s
sys    0m0.000s
```

*Edit:*
And adding in 50 directories as well ... half set `chmod 555`. So at this point 100 total entries in directory. 50 files, 50 directories.... 25 of each need modified.
```
$ time { find . ! -perm -u=w -print0 | xargs -0 -r chmod u+w; }
real	0m0.007s
user	0m0.002s
sys	0m0.006s

$ time { chmod -R u+w .; }
real	0m0.002s
user	0m0.002s
sys	0m0.000s
```

DrDaveD:
Jason, when unpacking a multi-layer docker image, most of the time the recursion happens most of the files will not have a permission problem anymore, because they will have been fixed on a previous scan through the files.  So expect only a small percentage of the files to need the mode change on average.  Doing the find and only changing the needed files is a clear win, and it doesn't cause the churn of multiple updates to the inode change time.

dtrudg:
Note that in the docker case, we take care of the permissions during the actual layer extraction process in the `libexec/bin/docker-extract` extraction operations now (#1295) - so in theory there shouldn't be any chmod needed for the docker case.

DrDaveD:
Oh, ok, I missed that. In any case, if there's no permission change needed, that's all the more reason to use find before blindly doing a chmod -R.

jmstover:
@DrDaveD I guess it depends. For me, the operations that find does are more expensive (syscall/runtime time wise) than what doing a blind `chmod -R` is.

Find + xargs:
```
% time     seconds  usecs/call     calls    errors syscall
  0.00    0.000000           0       180        10 open
  0.00    0.000000           0        54         3 stat
  0.00    0.000000           0        70           fstat
```

Chmod:
```
% time     seconds  usecs/call     calls    errors syscall
 71.43    0.000010           0        66           fstat
  0.00    0.000000           0        20         5 open
  0.00    0.000000           0         5         3 stat
```

Run time on the find is longer. So... to you, what are you looking for when you say "efficient"?

@dctrud Do you know why `umask 022` isn't just set in the `singularity` script ... or are permissions actually wrong coming from the source?

DrDaveD:
Probably it depends on the filesystem.  Try it for example on nfs.  Of course doing a write operation every time (at least to the inode change time) is going to be more expensive than a stat.  I tried it on a copy of the singularity-2.4.2 source code, and with my VM & nfs server it is consistently taking 2 seconds to do the chmod after the cache is fully populated, but a find (that finds no files to change) takes only about 0.1 seconds.

jmstover:
> takes only about 0.1 seconds.

Probably client metadata cache in the NFS case. One way to test if that's the case is to remount with the `noac` (no attribute cache) option. Basically it's used to keep coherence across muliple clients. Is it possible for you to test if there's a difference with that mount option?

DrDaveD:
If I set noac then the find that returns nothing takes about 2 seconds, but the chmod -R takes 4 to 5 seconds.

DrDaveD:
This has been rebased to development-2.x

dtrudg:
I ran the following test on a CentOS6 VM, testing against a debian chroot,  with that chroot on a volume that was mounted into the VM via nfs.

In my hands in this situation, the `find` solution is slower then the blunt
`chmod`.

```
# Created a debian chroot
debootstrap --foreign jessie chroot

# Copied it to two test dirs:
cp -a chroot chroot_chmod
cp -a chroot chroot_find_chmod

#Drop caches:
echo 3>/proc/sys/vm/drop_caches

# Test find
time find chroot_find_chmod/ ! -perm -u=rw -print0|xargs -0 -r chmod u+rw

real	0m0.516s
user	0m0.006s
sys	0m0.275s

# Test chmod
$ time chmod u+rw chroot_chmod/

real	0m0.005s
user	0m0.000s
sys	0m0.001s
```


On local HDD, where it is so fast as to be difficult to time, find was slower

```
# Test find
time find chroot_find_chmod/ ! -perm -u=rw -print0|xargs -0 -r chmod u+rw

real	0m0.024s
user	0m0.002s
sys	0m0.024s

# Test chmod
time chmod u+rw chroot_chmod/

real	0m0.001s
user	0m0.000s
sys	0m0.001s
```

dtrudg (commented):
*Edited to reflect incorrect chmod test, which was repeated correctly below*

Given the conflicting reports on which is faster, using find vs straight chmod, I'm inclined to think this PR should not be merged.

With default builds happening on /tmp - which is assumed local disk, we have reports of things being slower and faster using `find`

Over NFS, reported increases in speed are no greater than 3-4 seconds difference which is likely not critical at build time - and I don't believe is worth strongly chasing when we have conflicting increase/decrease in speed on the default '/tmp'.

It's also not clear to me whether we can expect better performance for a `find` vs blunt `chmod` on parallel fs such as lustre - where full stat on a file via `find` can be expensive. I wouldn't want to merge the PR without knowing if there is an impact there. I'm aware of people occasionally using a lustre/gpfs location as a temp dir for builds.

I'm also keen to avoid adding any more complexity to the shell scripts unless there is a strong need.

DrDaveD:
@dctrud  You did not remember the -R option on the chmod. Even if you had it would be an unfair comparison because you did not drop the cache between the two tests; the first one had to load the caches and the second one did not.   I ran the fair comparison on my SL6 VM and NFS server and it was 6.5 times slower:
```
# echo 3>/proc/sys/vm/drop_caches
# time find chroot_find_chmod/ ! -perm -u=rw -print0|xargs -0 -r chmod u+rw

real	0m0.624s
user	0m0.022s
sys	0m0.151s
# echo 3>/proc/sys/vm/drop_caches
# time chmod -R u+rw chroot_chmod/

real	0m4.190s
user	0m0.040s
sys	0m0.429s
```

I repeated the test 3 times and the second & third times the find was around 1.775s and the chmod -R stayed nearly the same, so it's quite variable, but the find is still over twice as fast as chmod -R.

Frankly I am having a hard time believing that anyone, regardless of speed, would prefer to use chmod -R which always modifies the inode of every file instead of a command that only modifies the files that need to be modified.  In your test case, none of the files needed to be modified because they all already have the same permissions.

dtrudg:
@DrDaveD - well that is very embarrassing, sincere apologies - am going to redo now.

The cache issue between the tests is, I believe, not in effect here, as I am operating on 2 different dirs.

DrDaveD:
Ok good point about the cache.

Also by trying to time changing all files with the find, I found a bug in my find command line.  It should contain "! -type l" as well so as not to attempt to change symbolic link modes.  I will update the PR.

DrDaveD:
To be completely fair I think the command lines should start with `time bash -c` because I think currently time is only timing the find part of the command line, not the xargs chmod part.  In practice it doesn't seem to make much difference however, probably because by the time chmod runs everything is already in the cache.

Timing a case where every file does have chmod applied is slightly slower than chmod -R, as expected:
```
# echo 3>/proc/sys/vm/drop_caches
# time bash -c "find chroot_find_chmod/ ! -type l -perm -u=rw -print0|xargs -0 -r chmod u+rw"

real	0m4.274s
user	0m0.072s
sys	0m0.696s
```

jmstover:
@DrDaveD Are you changing the value of `SINGULARITY_TMPDIR` to build on the NFS? The default is to build local at `/tmp`, which is why I'm hesitant.

I don't see a conceptual issue with the PR, just that _by default_, we should be building local where, time wise, the blanket `chmod -R` wins.

dtrudg:
Re-run on my system I get...

```
#NFS
time find chroot_find_chmod/ ! -type l ! -perm -u=rw -print0|xargs -0 -r chmod u+rw

real	0m0.850s
user	0m0.002s
sys	0m0.728s

echo 3>/proc/sys/vm/drop_caches
time chmod -R u+rw chroot_chmod/

real	0m1.464s
user	0m0.005s
sys	0m0.767s

#Local
echo 3>/proc/sys/vm/drop_caches
time find chroot_find_chmod/ ! -type l ! -perm -u=rw -print0|xargs -0 -r chmod u+rw

real	0m0.024s
user	0m0.008s
sys	0m0.017s

echo 3>/proc/sys/vm/drop_caches
time chmod -R u+rw chroot_chmod/

real	0m0.028s
user	0m0.004s
sys	0m0.024s

```

So, yes, find has a small runtime for me advantage here. I will update my review above, but I'm still personally of the opinion that this a pretty small difference in performance during a build task which generally runs much longer. I don't think it's a great candidate unless there is a large difference, which can be replicated across various fs where performance may be different due to different architectures.

E.g. I'm not sure if on lustre, blindly setting perms via chmod may be faster - depending on how many calls have to go out to MDTs + OSTs if find is pulling all file info.

R.E. @jmstover comment - with the default build location on /tmp (which is my local storage case above) the difference is negligible.

DrDaveD:
@jmstover I wasn't setting SINGULARITY_TMPDIR, I was just using nfs as a performance comparison, but now that you mention it I have had cases where "singularity run docker://..." has run my root filesystem out of space so I probably should be using SINGULARITY_TMPDIR.  I really don't think it's a good idea to assume that /tmp can be used for multi-gigabytes of scratch space.

Regardless of performance, I consider using chmod -R only for one-time manual commands, never something that one does in scripts, because of the unnecessary inode writes.

On the other hand I agree this is a very minor point and not nearly as important as most of my other PRs.

jmstover:
@DrDaveD What are the full NFS mount / export options for the share? If you can pass them along.

DrDaveD:
This is what shows in /proc/mounts: `rw,relatime,vers=3,rsize=32768,wsize=32768,namlen=255,hard,proto=tcp,timeo=600,retrans=2,sec=sys,mountaddr=XXX.XXX.XXX.XXX,mountvers=3,mountport=4048,mountproto=tcp,local_lock=none,addr=XXX.XXX.XXX.XXX`

jmstover:
On the NFS server can you do a:  `exportfs -v` and show the values for that export? Thanks.

DrDaveD:
Unfortunately I can't.  It's a Network Attached Storage device that I don't have access to.

DrDaveD:
I rebased on the current development-2.x.

GodloveD:
Heya @jmstover and @dctrud can I please get a final assessment on this now that it seems that the conversational dust has settled so to speak?  To merge or not to merge.  That is the question.

dtrudg:
@GodloveD - my gut feeling is this is probably a win on some fs, and not others - but now I'm not in an HPC center I'm unable to test that. I'd still vote no on it, as the demonstrated improvement is pretty small in the context of the total time needed for a build step... and I'm generally against the extra complexity of piped find | xargs unless there is a big difference.

jmstover:
Hi @DrDaveD ...
Thanks for this. For the 2.x release we're going to leave the `chmod` as it behaves in previous releases. We will probably come back to this in the 3.0 branch though for different build methods.

Thanks,
-J

PR #1350, opened 2018-03-01
Initial commit of new definition file parser

bauerm97:

This commit puts in the initial functions for parsing through a definition file. Sample usage can be seen by running `singularity build --deffile [/path/to/deffile]`. It will spit out the sections contained in the file.

@ArangoGutierrez @tri-adam 

Attn: @singularityware-admin

### DIFF
--- a/pkg/build/definition.go
+++ b/pkg/build/definition.go
@@ -32,6 +32,7 @@ type imageScripts struct {
 // buildData contains any scripts, metadata, etc... that the Builder may
 // need to know only at build time to build the image
 type buildData struct {
+	files map[string]string
 	buildScripts
 }
 
--- a/pkg/build/definition.go
+++ b/pkg/build/definition.go
@@ -10,6 +10,7 @@ package build
 import ()
 
 type Definition struct {
+	Header    map[string]string
 	ImageData imageData
 	BuildData buildData
 }
@@ -17,8 +18,8 @@ type Definition struct {
 // imageData contains any scripts, metadata, etc... that needs to be
 // present in some from in the final built image
 type imageData struct {
-	Metadata []byte
-	Labels   []string
+	Metadata []byte   //
+	Labels   []string //
 	imageScripts
 }
 
@@ -32,7 +33,7 @@ type imageScripts struct {
 // buildData contains any scripts, metadata, etc... that the Builder may
 // need to know only at build time to build the image
 type buildData struct {
-	Files map[string]string
+	Files map[string]string //
 	buildScripts
 }
 
--- a/pkg/build/definition.go
+++ b/pkg/build/definition.go
@@ -17,27 +17,27 @@ type Definition struct {
 // imageData contains any scripts, metadata, etc... that needs to be
 // present in some from in the final built image
 type imageData struct {
-	metadata []byte
-	labels   []string
+	Metadata []byte
+	Labels   []string
 	imageScripts
 }
 
 type imageScripts struct {
-	help        string
-	environment string
-	runscript   string
-	test        string
+	Help        string
+	Environment string
+	Runscript   string
+	Test        string
 }
 
 // buildData contains any scripts, metadata, etc... that the Builder may
 // need to know only at build time to build the image
 type buildData struct {
-	files map[string]string
+	Files map[string]string
 	buildScripts
 }
 
 type buildScripts struct {
-	pre   string
-	setup string
-	post  string
+	Pre   string
+	Setup string
+	Post  string
 }
--- a/pkg/build/definition_parser_deffile.go
+++ b/pkg/build/definition_parser_deffile.go
@@ -9,12 +9,155 @@ package build
 
 import (
 	"bufio"
+	"bytes"
+	"fmt"
 	"io"
 	"log"
 	"regexp"
 	"strings"
 	"unicode"
 )
+
+// validSections just contains a list of all the valid sections a definition file
+// could contain. If any others are found, an error will generate
+var validSections = map[string]bool{
+	"help":        true,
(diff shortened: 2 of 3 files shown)


pkg/build/definition_parser_deffile.go
```diff
@@ -9,13 +9,156 @@ package build
...
+	"pre":         true,
+	"post":        true,
+	"runscript":   true,
+	"test":        true,
+}
+
+// scanSections is the SplitFunc for the scanner that will parse the deffile. It will split into tokens
+// that designated by a line starting with %
+// If there are any Golang devs reading this, please improve your documentation for this. It's awful.
+func scanSections(data []byte, atEOF bool) (advance int, token []byte, err error) {
```
ArangoGutierrez, line 39: You may only specify the output types, without the names.
```go
func scanSections(data []byte, atEOF bool) (int,  []byte,  error)
```

pkg/build/definition.go
```diff
@@ -17,26 +17,27 @@ type Definition struct {
 // imageData contains any scripts, metadata, etc... that needs to be
 // present in some from in the final built image
 type imageData struct {
-	metadata []byte
-	labels   []string
+	Metadata []byte
```
dtrudg: What form is the metadata going to take? Is it going to be completely free-form, or JSON?. If latter can we use a json.RawMessage just to signify intent?

ArangoGutierrez (changes requested):

pkg/build/definition_parser_deffile.go
ArangoGutierrez: Also I don't understand how this func is an enhanced work of the ones that were already for parsing the deffile.

pkg/build/definition.go
```diff
@@ -17,26 +17,27 @@ type Definition struct {
...
 	buildScripts
 }
 
 type buildScripts struct {
-	pre   string
-	setup string
-	post  string
+	Pre   string
+	Setup string
+	Post  string
```
dtrudg, line 43: There was some talk about snagging progress during the build line-by-line somehow? Are we going to do this at some point, in which case it'd natural to parse into []string... but thinking about that more, it's hard to parse given shell here documents etc. so maybe not.

pkg/build/definition_parser_deffile.go
```diff
@@ -9,13 +9,156 @@ package build
...
 	"io"
 	"log"
 	"regexp"
 	"strings"
 	"unicode"
 )
 
+// validSections just contains a list of all the valid sections a definition file
+// could contain. If any others are found, an error will generate
+var validSections = map[string]bool{
```
dtrudg, line 24: I get that the map to bool is needed for a nice lookup, given go has no 'in' operator... but it's kinda ugly. Is it any better just to have this in a function `sectionValid(section string) bool` which directly uses a switch/case block?

pkg/build/definition_parser_deffile.go
```diff
@@ -9,13 +9,156 @@ package build
...
+func ParseDeffile(r io.Reader) (Definition, error) {
+	s := bufio.NewScanner(r)
+	s.Split(scanSections)
+
+	sections, err := scan(s)
+
+	def := Definition{
+		ImageData: imageData{
+			imageScripts: imageScripts{
+				Help:        sections["help"],
```
dtrudg: This is not handling the case that we don't have certain sections yet - they won't be in the sections map returned by scan(s)? Or have I missed something here.

pkg/build/definition_parser_deffile.go
ArangoGutierrez: you can do name check as the actual parser is doing, without a map just for boolean types

pkg/build/definition_parser_deffile.go
dtrudg: @ArangoGutierrez  - you can name outputs... and you then get naked return, and built-in doc:

https://golang.org/doc/effective_go.html#named-results

ArangoGutierrez: I know you can, but is cleaner, that way. you may define the name after, when calling the func

dtrudg: I really think having the name in the function signature for functions with compound returns helps a lot... it's major win when you have functions returning multiple things of same type e.g. (string, string, string)... which one is which thing?

bauerm97: I don't get to decide the signature of `splitSection`, it is necessary to be of type [SplitFunc](https://golang.org/pkg/bufio/#SplitFunc), so I am just using the signature from there.

In general, I don't mind naming function return values. You can still return a variable with a different name that shares the same type, and you still define the name when calling the function and assigned the return values. However, it does two things:

1. Clarifies the intended usage for the return values
2. Allows you to do set the values of the named return values and simply call `return` at the end to return the values of those variables. 

I don't think we should shy away from naming return values.

pkg/build/definition.go
bauerm97: Yea I'm pretty sure it'll just be raw JSON, not sure how to use `json.RawMessage` but I'll take a look

pkg/build/definition_parser_deffile.go
bauerm97: If a section isn't present in the map, a dereference via `sections[key]` will return a `nil` value, so essentially it doesn't matter.

pkg/build/definition_parser_deffile.go
bauerm97: The reason I did it in a map is because a go map is actually a hash table under the hood, so lookup is O(1).

PR #1351, opened 2018-03-02
Fix test for SINGULARITYENV_PATH

jmstover:
Fix test for SINGULARITYENV_PATH. Current test looks only to see if variable is unset, and not an empty string.

- Ref: #

**Checkoff for all PRs:**

Attn: @singularityware-admin


dtrudg (approved):
This looks good to me.

etc/init
```diff
@@ -11,7 +11,7 @@ unset ml
 unset BASH_ENV
 
 # Provide a sane path within the container
-if [ -z ${SINGULARITYENV_PATH+x} ]; then
+if [ -z "${SINGULARITYENV_PATH:-}" ]; then
     SINGULARITYENV_PATH="/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin"
 else
     SINGULARITYENV_PATH="$SINGULARITYENV_PATH:/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin"
```
GodloveD: I think this whole `if` statement can be thrown away and replaced by a single line.  
```
SINGULARITYENV_PATH="{$SINGULARITYENV_PATH:-}:/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin"
```
That should do the right thing, correct?

GodloveD (changes requested):
I think this change can be simplified.  Plz correct me if I am wrong.

etc/init
dtrudg: With that formulation you'd have a lone : at start of it if the var isn't set when the test occurs.

Also, $ is outside of {

GodloveD: Ah. Right you are. (On both counts of course).

GodloveD (approved):

PR #1356, opened 2018-03-02
use cobra pointers

ArangoGutierrez:

The idea is to have a global data struct on which parse the flags , for now is a var array, but we might want to build a GobalFlags or some data struct in the future (next PR) 
```go
// Global variables for singularity CLI
var (
	debug   bool
	silent  bool
	verbose bool
	quiet   bool
)
```

```go
singularityCmd.PersistentFlags().BoolVarP(&debug, "debug", "d", false, "")
	singularityCmd.PersistentFlags().BoolVarP(&silent, "silent", "s", false, "")
	singularityCmd.PersistentFlags().BoolVarP(&quiet, "quiet", "q", false, "")
	singularityCmd.PersistentFlags().BoolVarP(&verbose, "verbose", "v", false, "")
```

now this variables can be used directly with out using cobra. 

Attn: @singularityware-admin

### DIFF
--- a/internal/pkg/cli/build.go
+++ b/internal/pkg/cli/build.go
@@ -1,48 +1,44 @@
 /*
-  Copyright (c) 2018, Sylabs, Inc. All rights reserved.
+Copyright (c) 2018, Sylabs, Inc. All rights reserved.
 
-  This software is licensed under a 3-clause BSD license.  Please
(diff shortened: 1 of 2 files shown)


bauerm97 (approved):
Looks good!

Issue #1352, opened 2018-03-02
SINGULARITYENV_PATH doesn't appear to be propagated into the runtime container environment


moskalenko:
## Version of Singularity:

 2.4.2

### Expected behavior

It should be possible to execute programs located in the path set via SINGULARITYENV_PATH variable.

### Actual behavior

Debug log shows the SINGULARITYENV_PATH is being processed into PATH, but the resulting PATH doesn't appear to apply. jms replicated the issue.

### Steps to reproduce behavior

Put a static binary into a sandbox container, say /opt/foo/myprog

Export SINGULARITYENV_PATH=/opt/foo before executing the container with 'myprog'.

The following error is shown:

/.singularity.d/actions/exec: 9: exec: prog: not found

If the correct PATH is exported from /.singularity.d/env/90-environment.sh 'prog' will be found and executed.

dtrudg:
Closing this as it was addressed by #1430 which is included in Singularity 2.5.0. Please feel free to reopen if you have any further issues related to this.

PR #1363, opened 2018-03-06
fixing issue with user custom environment not being set with docker

vsoch:
This pull request will address the issue that when a user specifies a custom environment variable at runtime, it is not honored. For example:

```
SINGULARITYENV_PATH=/lala singularity exec ubuntu-14.04.simg env | grep '^PATH='
PATH=/usr/local/bin:/usr/local/sbin:/bin:/sbin:/usr/bin:/usr/sbin
```
It should be the case that `/lala` is shown above, but it is not.

## The Cause of the Problem
The 10-docker.sh script is one in the environment folder that is sourced at runtime. Since Docker images typically have a PATH variable, exporting the PATH here overrides anything done by the C.

## Fix Logic
We must be able to source Docker custom environments while honoring the Singularity default path, not repeating anything twice, and also the user custom specified variables at runtime. We also don't want to overwrite the Docker environment, because it's usually relevant to the base container. To do this, we:

 - parse through the Docker Env variables, and given that we find a PATH:
   - we take the difference betweent the path set by the C and what is found in the Docker container. The path is hard coded as a string `DOCKER_PATH` in the defaults.py file.
   - we only write the variables that aren't set by singularity by default (prevents duplicates)
   - we then add $PATH to this set, and we know confidently that the Singularity C is setting this.

## Updated Example
With the fix, we run the same command and since the user defined environment is also handled by the C, we get that in addition to the custom path (note before this was wiped):

```
SINGULARITYENV_PATH=/lala singularity exec ubuntu-14.04.simg env | grep '^PATH='
PATH=/lala:/usr/local/bin:/usr/local/sbin:/bin:/sbin:/usr/bin:/usr/sbin
```
This was worked on / discussed with @jmstover.  I would encourage other testing of different images that have more complicated paths, etc.

## Gitignore Update
The second fix was also adding files that are compiled and not in the .gitignore. I also put them in ABC order so the universe is returned to a state of 42 :) If there are related issues, please post here since this was alerted to me in a slack chat. Ty!

Attn: @singularityware-admin

### DIFF
--- a/libexec/python/docker/tasks.py
+++ b/libexec/python/docker/tasks.py
@@ -11,7 +11,7 @@ import sys
 import os
 sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__),  # noqa
                 os.path.pardir)))  # noqa
-sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))) # noqa
+sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))  # noqa
 
 from sutils import (
     add_http,
@@ -181,7 +181,7 @@ def extract_env(manifest):
         lines = []
         for line in environ:
             line = re.findall("(?P<var_name>.+?)=(?P<var_value>.+)", line)
-            
+
             for x in line:
                 varname = x[0]
                 varval = x[1]
@@ -189,9 +189,9 @@ def extract_env(manifest):
                 # For PATH, subtract Docker set PATH from Singularity default
 
                 if varname == "PATH":
-                    components = varval.split(':')
-                    components = [x for x in components if x not in DOCKER_PATH]
-                    varval = "$PATH:%s" % ":".join(components)
+                    parts = varval.split(':')
+                    parts = [x for x in parts if x not in DOCKER_PATH]
+                    varval = "$PATH:%s" % ":".join(parts)
 
                 line = 'export %s="%s"' % (varname, varval.strip(':'))
                 lines.append(line)
--- a/libexec/python/docker/tasks.py
+++ b/libexec/python/docker/tasks.py
@@ -190,10 +190,10 @@ def extract_env(manifest):
 
                 if varname == "PATH":
                     components = varval.split(':')
-                    components = [x for x in components if x not in DOCKER_PATH]  
-                    varval = "$PATH:%s" % (":".join(DOCKER_PATH + components))
+                    components = [x for x in components if x not in DOCKER_PATH]
+                    varval = "$PATH:%s" % ":".join(components)
 
-                line = 'export %s="%s"' % (varname, varval)
+                line = 'export %s="%s"' % (varname, varval.strip(':'))
                 lines.append(line)
 
         environ = "\n".join(lines)
--- a/libexec/python/docker/tasks.py
+++ b/libexec/python/docker/tasks.py
@@ -167,16 +167,34 @@ def extract_env(manifest):
     '''extract the environment from the manifest, or return None.
     Used by functions env_extract_image, and env_extract_tar
     '''
+    from defaults import DOCKER_PATH
+
     environ = get_config(manifest, 'Env')
     if environ is not None:
         if not isinstance(environ, list):
             environ = [environ]
 
+        # Default PATH components to subtract Docker set PATH
+
+        DOCKER_PATH = DOCKER_PATH.split(':')
+
         lines = []
         for line in environ:
             line = re.findall("(?P<var_name>.+?)=(?P<var_value>.+)", line)
-            line = ['export %s="%s"' % (x[0], x[1]) for x in line]
-            lines = lines + line
+            
+            for x in line:
+                varname = x[0]
+                varval = x[1]
+
+                # For PATH, subtract Docker set PATH from Singularity default
+
+                if varname == "PATH":
+                    components = varval.split(':')
+                    components = [x for x in components if x not in DOCKER_PATH]  
+                    varval = "$PATH:%s" % (":".join(DOCKER_PATH + components))
+
+                line = 'export %s="%s"' % (varname, varval)
+                lines.append(line)
 
         environ = "\n".join(lines)
         bot.verbose3("Found Docker container environment!")
--- a/.gitignore
+++ b/.gitignore
@@ -37,19 +37,21 @@ secbuildimg.sh
 
 src/action
 src/action-suid
-src/start
-src/start-suid
 src/builddef
 src/cleanupd
+src/docker-extract
+src/get-configvals
+src/get-section
 src/image-type
 src/mount
 src/mount-suid
 src/sif
-src/sinit
-src/get-section
 src/sign
-src/verify
+src/sinit
+src/start
+src/start-suid
 src/stamp-h1
+src/verify
 src/include/stamp-h1
 src/.dirstamp
 src/wrapper
--- a/libexec/python/defaults.py
+++ b/libexec/python/defaults.py
@@ -126,6 +126,9 @@ TAG = "latest"
 # Container Metadata
 DOCKER_NUMBER = 10  # number to start docker files at in ENV_DIR
 DOCKER_PREFIX = "docker"
+
+# Default docker path defined in C, subtracted from Docker 10-docker.sh
+DOCKER_PATH = "/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin"
 SHUB_PREFIX = "shub"
 
 # Defaults for environment, runscript, labels


jmstover:
Argh ... as much as I wanted this to work... 

This doesn't work for already created images, since `10-docker.sh` is already written... you can't override it. And when building, you would need to do:

`SINGULARITYENV_PATH=/blah/blah singularity build blah.img docker://ubuntu:latest`

Correct?

vsoch:
I'm not sure how we can go back in time and make it work for images that already have that file written. Sorry to not be able to help you!

jmstover:
Yeah... I was "Hoping" ... but I still think this is the right thing to have implemented in some form.

vsoch:
Feel free to re-open if needed, I closed it so someone else wouldn't feel badly about doing that :)

GodloveD:
Hey @vsoch, can you rebase this to `development-2.x` plz?  Maybe we can get it into 2.5.  😸

vsoch:
Sure, after the SCIF one!

libexec/python/docker/tasks.py
```diff
@@ -167,16 +167,34 @@ def extract_env(manifest):
...
+            for x in line:
+                varname = x[0]
+                varval = x[1]
+
+                # For PATH, subtract Docker set PATH from Singularity default
+
+                if varname == "PATH":
+                    parts = varval.split(':')
+                    parts = [x for x in parts if x not in DOCKER_PATH]
+                    varval = "$PATH:%s" % ":".join(parts)
```
GodloveD, line 194: This has the potential to re-arrange the order of the `PATH`.  (See docker://godlovedc/lolcow for example.) But I don't see a good way around it.

GodloveD (approved):
I've looked this over in some detail today and I think we should merge it, preferably into 2.5 so that folks who are having trouble with #1352 have a "path" forward.  😸 

But in general, I think the way we are handling `PATH` is very ad hoc and hacky.   For instance, I was not aware until reviewing this PR that `SINGULARITYENV_PATH` will simple _prepend_ to the existing `PATH`.  That seems very counterintuitive to me.  I was expecting `SINGULARITYENV_PATH` to overwrite the `PATH`.  I assume @gmkurtzer probably decided to prepend for convenience, but it's actually a less flexible method if you really want to rewrite the `PATH`.  For instance you can't append to the `PATH` or overwrite.  If you could just overwrite the `PATH` with `SINGULARITYENV_PATH` you could do whatever you wanted.  

Moving forward with 3.0 I think this is the way we should handle the `PATH`.
1. A base `PATH` is set (as it is in the C code now).
2. If the image is coming from Docker Hub, and`$PATH` is set on the Docker side, it's value should be determined and then it should overwrite the base `PATH` that was set in the C code.  (Seems like this is what the python code was doing before this PR.)
3. At this point, `SINGULARITYENV_PATH` should be evaluated and used to adjust the `PATH` if necessary.  With our current architecture, this isn't possible because the `10-docker.sh` script is sourced inside the container, where there is no access to the `SINGULARITYENV_PATH` variable.  imho at this point `SINGULARITYENV_PATH` should **overwrite** (not append or prepend to) the existing `PATH`.
4. Sourcing of files in `/.singularity.d/env` should proceed and the `PATH` should be adjusted accordingly.

I think that the one thing missing from the steps above is that we don't have access to `SINGULARITYENV_PATH` inside the container at runtime.  If we did things would be much easier.

jmstover:
> But in general, I think the way we are handling PATH is very ad hoc and hacky. 

Not really. The issue popped up from the Docker handling. It pulled the PATH entry from the Docker Manifest in... and that caused the PATH that had what you told it to include no longer exist since it overwrote it. i.e. ... it just did a `export PATH=[path from Manifest]`

We don't want to overwrite the manifest PATH though. It could have entries that are needed to correctly run the container. So, how `10-docker.sh` should end up now is:

`export PATH=$PATH:[Paths from Manifest not in $PATH]`

`$PATH` comes from the C code, and includes the preprended entry/entries from `SINGULARITYENV_PATH`.

> For instance you can't append to the PATH or overwrite. 

Generally, when you're adding a PATH, you want it to be searched before other things...

>  If we did things would be much easier.

Well... we could just not `unsetenv()` the `*ENV_PATH` var so it's hanging around. I think it could possibly cause issues, but if we left that hanging around then created something like `11-pathfix.sh` ... that did:

```
export PATH=${SINGULARITYENV_PATH}:${PATH}
unset SINGULARITYENV_PATH
```

It would accomplish the same thing as what the python code should now be doing...

GodloveD:
But if it ends up like this 

```
export PATH=$PATH:[Paths from Manifest not in $PATH]
```
Then  whatever the user added with `SINGULARITYENV_PATH` will get moved to the back of the line.  I agree with you that you usually want to prepend new paths.  But this is a catch 22 I think.

vsoch:
> who are having trouble with #1352 have a "path" forward.  smile_cat

Your pun game is strong, @GodloveD !

It's good to have some discussion about handling of these variables, because it's tricky! I think it might be a good strategy to look at this from the vantage point of the user's expectations. As a user I want my container to have a sound path, for my build and runtime adjustments to not screw that up, and that those same adjustments are honored. If this PR gets us closer to that, albeit not being long term and perfect, it's a +1. For this PR discussion I would also suggest that we focus now on variables like `PATH` and `LD_LIBRARY_PATH` that are likely to be defined twice. A global strategy for handling repeated variables should be saved for the next implementation.

## Case 1: No Docker
**build time** if there is absolutely no Docker (or other imports with environments) as a user I would expect a sane default `$PATH` for my container. As a developer I would expect a secure and reliable `$PATH` default. In this case, the base is the C code that sets this default, and both needs are met.
**run time**: As a user I would expect any `$PATH` variable that I set to be "more important" than whatever was set by the container. I want want my special path to prepend, but I still would want to maintain what is set in the container (at build time) because I just don't know. As a developer, I expect the base container path to be honored, and the user to be empowered to add things to it. We do this with `SINGULARITYENV_PATH`.

## Case 2: Docker!
**build time**: Now we have the added complexity that the Docker container has a PATH defined. As a user, since Docker is mysterious and essential, I want to have confidence that all the special stuff added to the Docker container path for my machine learning container aren't lost, regardless of what I do. If I add to the `$PATH` I do make the implicit declaration that my software path is priority (and it's on me to debug if my path goes first and makes something awry). In this case we now have two potential sources, the Docker metadata, along with the Singularity defaults. We currently allow for the Docker metadata, given a path defined, to completely override the Singularity default (and in most cases this is ok) but with this PR we will do more of "merging." This definitely could break things, given that the Docker container explicitly overrides a default system path/executable, however it seems unlikely. But we have another avenue - the build recipe `%environment` section set by the user. If we set the priority to be:

```
 [ build recipe ] <-- [ Docker metadata merged with Singularity defaults ]
```  
 then the user could arguably override any weirdness as a result. @GodloveD as a toy example - does  your lolcow container break with this change? If so, are you able to define either an `%environment` or `SINGULARITY_ENV` at runtime to unbreak it? If this solution is going to mass-break containers and not offer a workaround, even if it's a small wait until 3.0, we should possibly rethink it.

**runtime** The runtime adds the additional complexity that we already have the above set, and either we are allowed to change it or not. Actually, this was the main purpose of this PR - to add the `$PATH` during build time so that anything set at runtime would be honored. As a user, if I set something at runtime, I want it to be honored and not wiped by the initial container environment that is sourced. I'd also expect it to be prepended because my variable is more important, of course :P On the other hand, I don't want my variable to wipe out something already set in the container. From the standpoint of the developer, again I want the same things - the original Singularity default to be represented in the container, the various other Docker metadata environment variables also captured, and the build recipe and runtime variables taking preference over the bases.

Overall, given that the user can always set these runtime variables, this solution, although not perfect (and much more discussion needed based on reading the above) makes the software better for the user than without it, and I agree with @GodloveD that it would be useful for 2.4.5. I would also suggest that (and @jmstover can confirm we discussed) that this is more of a quick fix than a long term, elegant solution. 

  1. because the Python is going away, and
  2. because a solution is needed that addresses more generally variables that are found to be repeated.

Moving forward, some possible ideas for thinking about working with environments:
 - it's likely that we will have many defaults and import sources of environments. If we take a strategy that maintains all the original sources, computes a default, but still let's the user pick and choose to customize, this will ensure that nothing is lost / overwritten, and give flexibility.
 -  It's hard in situations like these to choose one preference for append/prepend/overwrite (because different users will have different preferences) so with the first point above, the software could expose a more detailed controller for the user to specify environment. We usually did this with environment variables, but of course other avenues could be taken if tired of those :)

Anyway, lots to think about!

vsoch:
@GodloveD I'm confused why would `SINGULARITYENV_PATH` get moved to the back of the line? Could you give me an example to test out? I'm using / looking at the lala example above, where the user set `SINGULARITYENV_PATH` to be lala and lala is at the front! Because lala is highly important! :)

vsoch:
Speaking of lolcow, I literally have this guy in almost every folder on my computer, lol

image

We must take great care to not disturb his PATH!

jmstover:
> Then whatever the user added with SINGULARITYENV_PATH will get moved to the back of the line.

When it parses the `SINGULARITYENV_*` entries... `PATH` *should* end up being:

`${SINGULARITYENV_PATH}:[default Singularity C path]`

That will be the value of `PATH` going into the runtime and what the `10-docker.sh` script will work with.

dtrudg:
Hi @GodloveD - if you approve this for 2.5 could you read / approve #1380? That will be the one that merges into 2.x as a rebase of this. Since I created PR #1380 I'm not eligible to review it for merge Thanks.

GodloveD:
@dctrud thanks for the heads up.  I'm going to continue conversation here and this will be the PR of record, but I will make sure to do a final review on #1380 as you suggest. 

Also, I think you are just fine to review that one since all you really did is rebase it right?  😄

dtrudg:
@GodloveD - Github won't let me review it if I created the PR :-) - I'm in the midst of reviewing here, where I can. Just running some last tests.

dtrudg:
Test results (will put a review here in a second)

TLDR - This PR fixes issue that SINGULARITYENV_PATH being disregarded, but breaks the order of PATH when it was set in the Dockerfile for a container.

1) Image without any custom Dockerfile PATH, no SINGULARITYENV_PATH: *OK*

```bash
$ singularity exec docker://alpine:latest env | grep ^PATH=
PATH=/usr/local/bin:/usr/local/sbin:/bin:/sbin:/usr/bin:/usr/sbin
```

2) Image without any custom Dockerfile PATH, SINGULARITYENV_PATH: /opt  *OK*

*adds /opt to front... this works for issue PR is directly addressing, but agree with @GodloveD somewhat r.e. long term issue addressing overwriting vs ammending existing container PATH. Could be argued SINGULARITYENV_PATH should be an override, something like SINGULARITYENV_PREPEND_PATH could add to front, SINGULARITY_APPEND_PATH could add to back.*

```
$ singularity exec docker://alpine:latest env | grep ^PATH=
PATH=/opt:/usr/local/bin:/usr/local/sbin:/bin:/sbin:/usr/bin:/usr/sbin
```

3) Image with custom Dockerfile PATH, no SINGULARITYENV_PATH: *WRONG ORDER*

docker://contiuumio/miniconda sets: `ENV PATH=/opt/conda/bin:$PATH` in Dockerfile

```bash
PATH=/usr/local/bin:/usr/local/sbin:/bin:/sbin:/usr/bin:/usr/sbin:/opt/conda/bin
```

**The custom PATH set in Dockerfile here was appended, not prefixed as requested**


4) Image with custom Dockerfile PATH, custom SINGULARITYENV_PATH: *WRONG ORDER*

docker://contiuumio/miniconda sets: `ENV PATH=/opt/conda/bin:$PATH` in Dockerfile
We set SINGULARITYENV_PATH=/opt/dummy

```bash
$ export SINGULARITYENV_PATH=/opt/dummy
$ singularity exec docker://continuumio/miniconda env | grep ^PATH=
PATH=/opt/dummy:/usr/local/bin:/usr/local/sbin:/bin:/sbin:/usr/bin:/usr/sbin:/opt/conda/bin
```

**The custom PATH set in Dockerfile here was appended, not prefixed as requested**

Current release 2.4.4 does Dockerfile PATH in correct order

docker://contiuumio/miniconda sets: `ENV PATH=/opt/conda/bin:$PATH` in Dockerfile

```bash
$ singularity exec docker://continuumio/miniconda env | grep ^PATH=
PATH=/opt/conda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

GodloveD:
> @GodloveD I'm confused why would SINGULARITYENV_PATH get moved to the back of the line? Could you give me an example to test out? I'm using / looking at the lala example above, where the user set SINGULARITYENV_PATH to be lala and lala is at the front! Because lala is highly important! :)

@vsoch I was just saying if you were to rewrite `libexec/python/docker/tasks.py` line 194 to be thus:
```
varval = "%s:$PATH" % ":".join(parts)
```
Which I think is what @jmstover is suggesting (and which I actually think you should do) then `SINGULARITYENV_PATH` will not come at the front anymore.  But it's not really true that it will be at the back either.  It will just follow any custom paths that are added by docker.  

If I read the code correctly, right now, this is the order.
```
PATH=SINGULARITYENV_PATH:default_base_path_from_c:Docker_additions
```
If you made the change we are suggesting it would be:
```
PATH=Docker_additions:SINGULARITYENV_PATH:default_base_path_from_c
```
Neither is ideal.  But it seems that the second is slightly more sane than the first.  I think what we would like is something more like this:
```
PATH=SINGULARITYENV_PATH:Docker_additions:default_base_path_from_c

```
But I don't think we can get there without rewriting a bunch of code which would require a bunch of testing etc.  We'd like to get this release out in short order so the better solution may have to wait.

Now let me explain why I think that `SINGULARITYENV_PATH` should simply overwrite the value in the container (in >=3.0 of course).
1. This is what other variables do unless I'm mistaken.  Why should we have a one off, special variable that is different?
2. "Because the user needs handholding!" you may respond.  That doesn't make any sense to me.  Linux allows you to do something like this if you want to:
```
export PATH=borked
```
It doesn't say "Whoops! That's a mistake.  Let me prepend that onto `$PATH` for you so that you don't screw up".  In other words, Linux gives you the power to make your own decisions.  And it's a good thing too.  There are situations in which you really do want to remove some directory from your path.  Right now we are not allowing users to do this because we are afraid they might hurt themselves.  I will always argue for giving users more power even if that means they may use it unwisely.

dtrudg (changes requested):
I don't think this should be merged in this state.

It fixed the issue that SINGULARITYENV_PATH was ignored.

But - it has a regression for containers that set PATH in the Dockerfile. See the example from my tests toward bottom of comments. 

TLDR - `docker://continuum/miniconda` has `ENV PATH /opt/conda/bin:$PATH`. Singularity 2.4.4. give `/opt/conda/bin` at start of PATH. This PR makes the `/opt/conda/bin` move to the end of PATH. 

I'd not be comfortable merging unless we maintain the correct order for PATH set in the Dockerfile.

GodloveD:
@dctrud well... I don't know.  How much will that really break?  I think that if @vsoch tweaks the code as I suggested, both Docker paths and user defined runtime paths will precede the default PATH.  That might break some containers, but I bet we would both have a hard time finding one.

jmstover:
> Which I think is what @jmstover is suggesting

No, I was actually suggesting what was in there. My biggest issue is wrapping my pea brain around what the Python code is actually doing... 

If we can get it do do something like:

`${SINGULARITYENV_PATH}:[uniq docker paths (i.e. /opt/conda/bin)]:DOCKER_PATH`

I think that would fit everyones needs

dtrudg:
@GodloveD - If there's a docker container that installs a new version of a common system-provided tool into a custom PATH then it could get broken by the move to the end in the current implementation of the PR. I have come across such a thing in the past... a container that was based off CentOS6 , had an old Git in /usr/bin but installed a new Git into an /opt PATH. I don't think this is going to be common at all, but I'm uncomfortable with the position of the custom added PATH element being opposite to where it is in the Dockerfiile. I think your suggestion where it'd be at the front, as would a SINGULARITYENV_PATH element would be best here. And shoot for nicer PATH semantics in 3.0 for sure :-)

vsoch:
@GodloveD I had it both ways, and @jmstover asked for it to be reversed to its current state. 

image

I'm ok with whichever you guys decide - I think generally a runtime variable should take preference to something set at build. It's actually a slight tweak of what you showed, instead of `Docker_additions` it would be `custom_docker_additions` because any docker additions that are already represented are ordered as they would be in `default_base_path_from_c`. I don't see any "one way fits all" fix to the ordering. Either the Docker ordering or the user ordering has to be respected, but they can conflict of course. As I stated earlier (and @dctrud is also talking through) is to have a lever for the user to specify the desired order.)  This in my mind is not ideal:

```
PATH=Docker_additions:SINGULARITYENV_PATH:default_base_path_from_c
```
Given that the docker container is likely being used as a base for the user image, if the user sets something at runtime to override this, it wouldn't work. If I didn't want to set something at runtime to override, I wouldn't need SINGULARITYENV_PATH.

@GodloveD reading your last bit about not appending - I am totally on the same page here!  I think we *could* actually have some hand holding, something that if it finds a path already defined, instead of making a decision about it, just doing a print to the console for the user and saying "Hey, check out what you are doing, just want to check on this!" 

@dctrud what you showed would be correct. If the user at runtime wants to `/opt/dummy` to be set it **would** override the /opt/conda default that is hard coded in the container.

dtrudg:
To clarify my thoughts, I think @jmstover 's comment is correct here - given we aren't looking to completely change the nature of SINGULARITYENV_PATH handling here right now, it should come out with '${SINGULARITYENV_PATH}:[uniq docker paths (i.e. /opt/conda/bin)]:BASE_DOCKER_PATH'

vsoch:
Isn't that what is implemented in the PR? 
```
SINGULARITYENV_PATH=/lala singularity exec ubuntu-14.04.simg env | grep '^PATH='
PATH=/lala:/usr/local/bin:/usr/local/sbin:/bin:/sbin:/usr/bin:/usr/sbin
```
To step back - the function that is happening here is only manipulating the "hard coded" file in `/.singularity.d/env/10-docker.sh` to create the runtime environment. At this time there is no known list of `SINGULARITYENV_` that can be parsed to compare with, and we only know about the default (`DOCKER_PATH`) because it's always the same thing. So if you mean to say that we do some parsing with the actual variables in the runtime `$PATH` I do not believe we can do this. The only thing we have is `$PATH` which is literally the string that will get filled in at runtime by the SINGULARITYENV_ stuffs. The python is a different process and doesn't have the power to set environment variables to return, it can only set them by writing to file at build time. We would either need to:

 - handle this as a separate thing from python (or whatever library is used) so the library just reports the Docker environment it finds somewhere, and the controller then does this parsing logic **at runtime** so there isn't an issue with things like $PATH.
 - Have the python do as it does now and accept that PATH is set to something, append or prepend. There are pros and cons to each, have a sword fight and pick :crossed_swords: 
 - Don't accept any appending / prepending and leave the borked-ness up to the user, but give a nice friendly message at some point (if possible).

I'm thoroughly confused about what is wrong, desired, and I had thought based on discussion with @jmstover  that this was a quick fix, not meant to be a final or elegant solution. So so I'm just going to go off and read some man pages and have a good night :)

vsoch:
Suggestion: whomever has the solution they think is best, please do a quick edit / suggested or actual change to the PR, demonstrate the different usage, and then discussion can follow.

dtrudg:
@vsoch - as far as I am concerned, the SINGULARITYENV_PATH handling is correct in this PR, but the Dockerfile custom PATH is now not right.

In its Dockerfile, continuum/miniconda sets `ENV PATH=/opt/conda/bin:$PATH`

**Release 2.4.4** gets this correct....

`PATH=/opt/conda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`

**This PR** has a regression for this....

`PATH=/usr/local/bin:/usr/local/sbin:/bin:/sbin:/usr/bin:/usr/sbin:/opt/conda/bin`

vsoch:
ohhhh I see! Thank you for showing that... let me look more closely...

vsoch:
okay here is the environment from the manifest, no editing done:

```
['PATH=/opt/conda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
 'LANG=C.UTF-8',
 'LC_ALL=C.UTF-8',
 'TINI_VERSION=v0.16.1']
```
here is `DOCKER_PATH`
```
['/bin', '/sbin', '/usr/bin', '/usr/sbin', '/usr/local/bin', '/usr/local/sbin']
```
We would want that junk to be behind the Docker image path, and the Docker image path to be behind `SINGULARITY_ENV` stuffs.  When we do the loop as defined now, we get:

```
 'export PATH="$PATH:/opt/conda/bin"\nexport LANG="C.UTF-8"\nexport LC_ALL="C.UTF-8"\nexport TINI_VERSION="v0.16.1"'
``` 
let me print that...

```
export PATH="$PATH:/opt/conda/bin"
export LANG="C.UTF-8"
export LC_ALL="C.UTF-8"
export TINI_VERSION="v0.16.1"
```

and running in the interface this maps to:

```
PATH=/usr/local/bin:/usr/local/sbin:/bin:/sbin:/usr/bin:/usr/sbin:/opt/conda/bin
v
```
and this is because the `$PATH` variable, set at runtime, outside of this python bit, gets expanded to that. So now we could easily switch those two around, let's try that. Here is the change:

```
varval = "$PATH:%s" % ":".join(parts)
```
```
varval = "%s:$PATH" % ":".join(parts)
```
and then the list is as you would expect:

```
export PATH="/opt/conda/bin:$PATH"
export LANG="C.UTF-8"
export LC_ALL="C.UTF-8"
export TINI_VERSION="v0.16.1"
```
and the runtime too:

```
vanessa@vanessa-ThinkPad-T460s:~$ singularity exec docker://continuumio/miniconda3 env | grep ^PATH=
PATH=/opt/conda/bin:/usr/local/bin:/usr/local/sbin:/bin:/sbin:/usr/bin:/usr/sbin
```
but the reason I'm saying we cannot do that is because the entire point of the PR was to honor the runtime variables, let's say the /opt/conda/bin needed to go after some `/opt/poodle/bin`:

```
SINGULARITYENV_PATH=/opt/poodle/bin singularity exec docker://continuumio/miniconda3 env | grep ^PATH=
PATH=/opt/conda/bin:/opt/poodle/bin:/usr/local/bin:/usr/local/sbin:/bin:/sbin:/usr/bin:/usr/sbin
```
It would be after. The issue is that the custom variables come packaged in the runtime `$PATH` as things that need to be separated that cannot happen there because the only thing that the python knows (at build time) is that there is some path, and it comes before or after.

I think the issue comes down to which one is fixable by the user, and which one is not. You guys decide! lol.

jmstover:
Ohh... I better add this in before I forget....

The reason I think that `$PATH` should exist in `10-docker.sh` is if someone crated an image and they run it with:

`SINGULARITYENV_PATH=/scratch/job1/bin singularity run [...]`

Then `/scratch/job1/bin` is passed through during runtime... Next run they do:

`SINGULARITYENV_PATH=/scratch/job2/bin singularity run [...]`

Then `/scratch/job2/bin` is passed through.

So, the container runtime `PATH` can be modified by using the `SINGULARITYENV_PATH` variable as it would be if you didn't have the information from the Docker manifest.

Is there any way support this sanely? Maybe having `10-docker.sh` have something like:

`export PATH=[uniq manifest dirs]:$PATH`

But that means that the path given isn't at the start of the search...

vsoch:
The fix we did would support it, but we can't have our cake and eat it too. The thing you want at the front is also part of the thing that is automatically set (that we also want to go in front of the Docker path, for the docker container to work.

It sounds like a more elegant solution for 3.0 is in the works, and for this version, I would say that we should choose the one that doesn't prevent the user completely from doing something. As it is now, you can't set any custom PATH, ever, because the container Docker one always overrides it. It **is** true that the Docker set paths get klobbered (we lose /opt/conda/bin) BUT it also is the case that the user could see that and undo it for a quick fix with `SINGULARITYENV_PATH` set to add it. So... lesser of two evils? Is there a right way to do this other than redoing the entire flow (including the C, the python/Go, and how it is then sourced after).  Probably not in the time frame that you want. It's also totally an option to leave the bug as is, and fix it at some future point.

jmstover:
> The fix we did would support it, but we can't have our cake and eat it too.

Mmm... cake.

But yup... Unless we actually change the `10-docker.sh` to do some environment manipulation, don't unset `SINGULARITYENV_PATH` in the C code, etc... what we're going to end up with at best, it will come *after* the important `PATH` entries from the Docker manifest.

That would leave the manifest directories as coming in the path search order before what was given in `SINGULARITYENV_PATH` ...  i.e. `export PATH=[uniq manifest dirs]:$PATH`

@moskalenko Would this work for your use case that you're seeing? It's not an ideal solution, but as things stand, and the way it's currently setup, this is probably as close as we can come.

vsoch:
🎂🍰🎂🍰🎂🍰😀

GodloveD:
So as @vsoch  says, it is impossible with the current architecture to "have our 🍰 and eat it too".  Or as I said, it is a "catch 22" (if you prefer ⚾️ metaphors).  Either way we all agree that this fixes #1352, but may break other things. 

Instead of going this route, I wonder how hard it would be to create a file in the container's overlay called `/.singularity.d/env/11-user_defined_SINGULARITYENV_PATH` before the environment is sourced?  And I wonder if that would be possible even if a user has specified a persistent overlay?  This would allow us to have PATH ordered like so.

```
PATH=$SINGULARITYENV_PATH:docker_paths:base_path
```
I'll poke around a bit and try to figure out how feasible this would be.  If anyone else has a better handle on the overlay code and can chime in, please do so.

libexec/python/docker/tasks.py
moskalenko: I, for one, don't expect SINGULARITYENV_PATH to override the PATH inside the container. The previously working approach with prepending it is like what environment modules usually do and is intuitive.

If there is no access to SINGULARITYENV_PATH inside the container at runtime then what is the following?

DEBUG   [U=0,P=24127]      singularity_runtime_environment()         Transposing environment
DEBUG   [U=0,P=24127]      singularity_runtime_environment()         Converting envar 'SINGULARITYENV_HISTFILE' to 'HISTFILE' = '(null)'
DEBUG   [U=0,P=24127]      envar_set()                               Unsetting environment variable: HISTFILE
DEBUG   [U=0,P=24127]      singularity_runtime_environment()         Converting envar 'SINGULARITYENV_PATH' to 'PATH' = '/opt/blupf90:/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin'
DEBUG   [U=0,P=24127]      envar_set()                               Setting environment variable: 'PATH' = '/opt/blupf90:/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin'

The problem is that the above PATH seems to be discarded right after that and the 10-docker.sh PATH is used instead.

vsoch: @moskalenko this is probably happening elsewhere (not in the Python). The code above is written at the time of build, so you don't have access to `SINGULARITYENV_PATH` at that time.

moskalenko:
I'm fine with the docker manifest PATH bits ahead of the SINGULARITYENV_PATH bits. It'll be just like 'append_path' in environment modules. In general, taking a good look at environment modules systems (both c and the modern lua-based Lmod - http://lmod.readthedocs.io/en/latest/050_lua_modulefiles.html ) might be a good idea since they thought about such issues for way over 20 years. E.g. Lmod has 'prepend_path', 'append_path', and 'setenv' (override) for maximum flexibility. Since this PR is for a 'quick' fix for a concrete regression I'm fine with just getting the 'append_path'. After all, we're putting apps into the container environment, so if we sourced the container from docker://ubuntu:xenial or something it's ok to have the docker env PATH ahead of the application-specific paths. Thank you.

GodloveD:
OK.  Please see #1389.  I believe this should fix things the way we want.

GodloveD:
Please see #1430 for a second attempt at fixing this without tinkering with older containers at runtime.

dtrudg:
This PR is closed in favour of the approach implemented in #1430

PR #1371, opened 2018-03-08
Proposed 2.5 - restore docker-extract whiteout handling

dtrudg:

This PR restores the docker-extract whiteout handling that was introduced in 2.4.3, but removed in 2.4.4. The PR ensures that when docker layers are extracted, whiteout .wh.* files are correctly handled, resulting in a correct container fs.

At runtime, this PR adds the ability for extraction to fall-back to the previous approach (using tar) if libarchive is not available.

- libarchive (with headers) is *required* at build time\*
- libarchive at runtime enables correct whiteout handling
- missing libarchive at runtime falls back to incorrect extraction, no whiteout handling

\*Note - it is possible to run 'make -i install' on a machine without libarchive headers, if 'make' was run at a location with these*

Thanks to @DrDaveD for providing the commit which removed the unintentional libarchive linking to other binaries that was observed in 2.4.3 on some distributions.

Changes to the following files in this commit are *not* new, and have been reviewed in previous PRs:

- .gitignore
- .travis.yml
- .travis/rpmbuild_test
- bin/singularity.in
- debian/control
- singularity.spec.in
- src/Makefile.am
- src/docker-extract.c
- tests/21-build_docker.sh

Changes to the following files in this commit *are* new, and have not yet been reviewed:

- CHANGELOG.md
- configure.ac
- libexec/bootstrap-scripts/deffile-driver-docker.sh
- libexec/bootstrap-scripts/main-dockerhub.sh
- libexec/cli/build.exec
- libexec/handlers/image-docker.sh

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/src/docker-extract.c
+++ b/src/docker-extract.c
@@ -1,0 +1,370 @@
+#define _GNU_SOURCE
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <errno.h>
+#include <string.h>
+#include <fcntl.h>
+#include <sys/file.h>
+#include <sys/param.h>
+#include <sys/stat.h>
+#include <locale.h>
+#include <archive.h>
+#include <archive_entry.h>
+
+#include "config.h"
+#include "util/file.h"
+#include "util/message.h"
+#include "util/registry.h"
+#include "util/util.h"
+
+/* apply_opaque
+ *  Given opq_marker as a path to a whiteout opaque marker
+ *    e.g. usr/share/doc/test/.wh..wh..opq
+ *  Make the directory containing the make opaque for this layer by removing it
+ *  if it exists under rootfs_dir
+ */
+int apply_opaque(const char *opq_marker, char *rootfs_dir) {
+    int retval = 0;
+    char *token;
+    static char buf[MAXPATHLEN];
+
+    token = strrchr(opq_marker, '/');
+    if (token == NULL) {
+        singularity_message(ERROR, "Error getting dirname for opaque marker\n");
+        ABORT(255);
+    }
+
+    retval = snprintf(buf, sizeof(buf), "%s/%s", rootfs_dir, opq_marker);
+    if (retval == -1 || retval >= sizeof(buf)) {
+        singularity_message(ERROR, "Error with pathname too long\n");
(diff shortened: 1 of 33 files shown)


src/docker-extract.c
```diff
@@ -0,0 +1,370 @@
...
+        singularity_message(ERROR, "Error with pathname too long\n");
+        ABORT(255);
+    }
+
+    // Target may not exist - that's ok
+    retval = 0;
+
+    if (is_dir(buf) == 0) {
+        retval = s_rmdir(buf);
+    } else if (is_file(buf) == 0) {
```
cclerget: @dctrud symlinks seems not concerned by this check but need to be removed too

dtrudg: @cclerget - thanks for spotting that. Added an `|| is_link(buf) == 0` on the else if

jakirkham:
What about statically linking `libarchive`? Would that work or is that somehow problematic?

dtrudg:
Unfortunately the .a isn't available in easy distro packages on all common distros.

jakirkham:
That could be solved by vendoring `libarchive`. Not ideal, but it's an option.

dtrudg:
@jakirkham - yes it was one of the things we discussed but we are trying to get this out sooner rather than later - and there are complexities integrating into the build and requiring us to follow libarchive security updates in our tree etc.

At this point in time, lack of a libarchive .so at runtime will drop back to old extraction behaviour - which hopefully eases things for people who really can't push libarchive out to cluster nodes easily.

jakirkham:
Ok, sorry, took the removed release label above as this wasn't going to make it into 2.5.

dtrudg:
Ahh - no... I just switched from label to milestone for tracking. It's in the 2.5 milestone :-)

jakirkham:
Great! Thanks for trying to get it in. :)

Is there a particular version range of libarchive that we need? Was going to nudge our cluster admins if we don't have it already (we probably do).

cclerget (approved):
Looks good to me

GodloveD (approved):
I've been tinkering with this for a while now and it all seems good to me.

PR #1372, opened 2018-03-08
fix is_deffile assumptions

octomike:
I just created a recipe called `Singularity.tardis` (the name of our cluster) and it failed to build.
As it turns out this is because the file matches `*.tar*` and `is_deffile()` assumes that it's a tar archive. I think the whole test/guess for known file names can be safely removed.

Note that i ticked the box "I have tested this PR", but it didn't finish. It seems the test is broken in my environment - I waited 18h for the stage `+ python2 -m unittest tests.test_docker_import ` to complete.

Attn: @singularityware-admin

### DIFF
--- a/libexec/functions
+++ b/libexec/functions
@@ -396,30 +396,12 @@ nonroot_build_warning() {
 is_deffile() {
 # check if a file looks like, walks like, and quacks like a def file
     FILE2CHECK=$1
-    
-    # first make a guess based on naming convention
-    case ${FILE2CHECK:-} in
-        *.img)
-            return 1
-            ;;
-        *.tar*)
-            return 1
-            ;;
-        *.def)
-            return 0
-            ;;
-        Singularity) 
-            return 0
-            ;;
-        *)
-            ;;
-        esac
 
-    # name is ambiguous.  is it a text file with def file words in it?
+    # is it a text file with def file words in it?
     if ! `file $FILE2CHECK | grep -q "ASCII\|Unicode"`; then
         return 1
     fi
-        
+
     grep -qiE "^bootstrap:|^%setup|^%post|^%test|^%labels|^%files|^%runscript|^%environment" "$FILE2CHECK"
     RETVAL=$?
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and changes prior to that are (unfortunately) done retrospectively. Critical ite
  - Put /usr/local/{bin,sbin} in front of the default PATH
  - Adjustments to SCIF (Scientific Filesystem) integration for broader use
  - Fixed bug that did not export environment variables for apps with "-" in name
+ - Removed overeager check for known invalid recipe file extension
 
 ## [v2.4.4](https://github.com/singularityware/singularity/tree/release-2.4)
 


GodloveD (changes requested):
Heya @octomike, sorry for the delay and thanks for the changes.

I agree that we should get rid of this check.  Would you mind rebasing this agains the development-2.x branch and adding a note to the CHANGELOG.md file please?

octomike:
No worries! I applied and amended the commit on top of development-2.x now.
Hopefully my wording in the changelog is okay?

GodloveD (approved):

GodloveD:
Looks good to me.  If we can get a second positive review we can merge.

libexec/functions
```diff
@@ -396,30 +396,12 @@ nonroot_build_warning() {
...
-        Singularity) 
-            return 0
-            ;;
-        *)
-            ;;
-        esac
 
-    # name is ambiguous.  is it a text file with def file words in it?
+    # is it a text file with def file words in it?
     if ! `file $FILE2CHECK | grep -q "ASCII\|Unicode"`; then
```
dtrudg, line 401: I don't think we can rely on file if we are supporting running Singularity on minimal distros like alpine, as it is not always available:

```
$ singularity run docker://alpine:latest
Docker image path: index.docker.io/library/alpine:latest
Creating container runtime...
Exploding layer: sha256:ff3a5c916c92643ff77519ffa742d3ec61b7f591b6b7504599d95a4a41134e28.tar.gz
Exploding layer: sha256:1f6d219d0077f6af480398f264117d2e74c03ec5f12d09b7b29b1eefbba872eb.tar.gz
Singularity> file /etc/passwd
/bin/sh: file: not found
```

_NB - running alpine as a container here - but talking about an issue if it is the Singularity host_

I know the check based on `file` was previously here - but its brokenness on alpine was hidden by the simple (and also broken) name checks. 

I'd suggest trying the check calling `file` first, and dropping back to something else along the lines of a dumber filename test if `file` is not available.

dtrudg (changes requested):
Believe we needs some changes here, as `file` isn't guaranteed available on e.g. alpine.

libexec/functions
GodloveD: What about grepping the output of `strings`?  Looks like `strings` is included in alpine and busybox.

GodloveD: On second thought, I don't think that is going to work b/c strings will happily pull those keywords out of a def file that is stored in an image.  😕 

I actually think we tried that when we were first writing these tests and had problems.

dtrudg: I can't think of a nice way to do this if file isn't available. I think using the file check first, and dropping to the less-smart name check if it isn't available is probably the sensible thing. When we are in Go in 3.0 we will have an avenue to address this better.

GodloveD: meh.  Is it really worth it to make the change if we are just going to reverse the order of the checks and retain the file-name business?  Especially if this will be a moot point once 3.0 comes out?

dtrudg: Probably not - there is a straightforward workaround in file naming here.

GodloveD:
@octomike I'm sorry.  This is a good idea, but @dctrud pointed out a flaw.  Can you review our conversation above?  Do you have any ideas?

GodloveD:
@octomike and @dctrud so I guess we don't really have any forthcoming ideas?  Should we punt on this until 3.0 and close the PR?  

Let me know if you have any thoughts and I will re-open.

PR #1373, opened 2018-03-08
Bad deffile

ArangoGutierrez:
`ParseDefinitionFile` now returns error when a bad definition file.

Also fix the `_test.go` to work with the new version of the `ParseDefinitionFile`

Attn: @singularityware-admin


bauerm97 (changes requested):
Hey Ed, just going to leave this here for reference since we discussed in slack already.

I fixed the merge conflicts in the branch I pushed to my fork, go ahead and use that as the base. Make the tests a little less cyclic as we discussed and then we can merge :)

ArangoGutierrez:
@bauerm97 check out commit: 
d992b89

pkg/build/definition_parser_deffile.go
```diff
@@ -118,6 +118,28 @@ func doSections(s *bufio.Scanner, d *Definition, done chan error) {
...
+		if line = strings.TrimSpace(line); line == "" || strings.Index(line, "#") == 0 {
+			continue
+		}
+		var key, val string
+		lineSubs := strings.SplitN(line, " ", 2)
+		if len(lineSubs) < 2 {
+			key = strings.ToLower(strings.TrimSpace(lineSubs[0]))
+			val = ""
+		} else {
+			key = strings.ToLower(strings.TrimSpace(lineSubs[0]))
```
bauerm97, line 136: We can't use ToLower on these strings. File names are case sensitive. This will fail for any file that's not all lower case.

bauerm97 (changes requested):

pkg/build/definition_parser_deffile.go
ArangoGutierrez: got it. fixing....>

PR #1272, opened 2018-01-26
SCIF allow for "." in app names

vsoch:
**Description of the Pull Request (PR):**

This is a quick PR to close #1237 for a bug found by @Saford91 to allow for a user to define a SCIF app with a period (.) without breaking the environment. I also added two compiled filed to the gitignore that were missing.

Before the fix, interacting with a SCIF app looks like this:

```
./apps apps
/.singularity.d/actions/run: 1: /.singularity.d/env/94-appsbase.sh: SCIF_APPDATA_name.v1.0=/scif/data/name.v1.0: not found
/.singularity.d/actions/run: 2: /.singularity.d/env/94-appsbase.sh: SCIF_APPMETA_name.v1.0=/scif/apps/name.v1.0/scif: not found
/.singularity.d/actions/run: 3: /.singularity.d/env/94-appsbase.sh: SCIF_APPROOT_name.v1.0=/scif/apps/name.v1.0: not found
/.singularity.d/actions/run: 4: /.singularity.d/env/94-appsbase.sh: SCIF_APPBIN_name.v1.0=/scif/apps/name.v1.0/bin: not found
/.singularity.d/actions/run: 5: /.singularity.d/env/94-appsbase.sh: SCIF_APPLIB_name.v1.0=/scif/apps/name.v1.0/lib: not found
/.singularity.d/actions/run: 6: export: SCIF_APPDATA_name.v1.0: bad variable name
```
Ruhroh! The reason is the same that we cannot have a `-` in an environment variable. Now, usage is fixed with a simple adjustment to sed:

```
# Build
sudo singularity build versions.simg Singularity

...

Finished processing dependencies for scif==0.0.71

# Install the recipe with lots of bad characters in the name
+ /opt/conda/bin/scif install /recipe.scif
Installing base at /scif
+ apprun     FOO.BAR
+ appenv     FOO.BAR
+ apprun     name.v3.0
+ appenv     name.v3.0
+ apprun     name.v2.0
+ appenv     name.v2.0
+ apprun     name.v1.0
+ appenv     name.v1.0
+ apphelp     name.v1.0
Adding runscript
Finalizing Singularity container
Calculating final size for metadata...
Skipping checks
Building Singularity FS image...
Building Singularity SIF container image...
Singularity container built: versions.simg
Cleaning up...
```
Now we want to run the image to see if we get the SCIF entry point...
```
./versions.simg 

Scientific Filesystem [v0.0.71]
usage: scif [-h] [--debug] [--quiet] [--writable]
[...]

PR #1375, opened 2018-03-09
updating scif-revision to be against development-2.x branch

vsoch:

This is a rebase of PR #1272 . This should *not* be merged until the work here https://github.com/singularityware/singularity/pull/1238/files that is not existing in the current development-2.x is properly added back. It will not work without that. The function has been updated but the rest of the code base is not, as it was already pull requested, reviewed, and merged with a previous version of singularity (that looks like it's now gone?) :( 

Attn: @singularityware-admin

### DIFF
--- a/libexec/bootstrap-scripts/deffile-sections.sh
+++ b/libexec/bootstrap-scripts/deffile-sections.sh
@@ -389,34 +389,37 @@ for app in ${SINGULARITY_ROOTFS}/scif/apps/*; do
     if [ -d "$app" ]; then
 
         app="${app##*/}"
-        app=(`echo $app | sed -e "s/-/_/g"`)
+
+
+        # Replace all - or . with an underscore
+        appvar=(`echo $app | sed -e "s/-/_/g" | sed -e "s/[.]/_/g"`)
         appbase="${SINGULARITY_ROOTFS}/scif/apps/$app"
         appmeta="${appbase}/scif"
 
         # Export data, root, metadata, labels, environment
-        echo "APPDATA_$app=/scif/data/$app" >> "${APPGLOBAL}"
-        echo "APPMETA_$app=/scif/apps/$app/scif" >> "${APPGLOBAL}"
-        echo "APPROOT_$app=/scif/apps/$app" >> "${APPGLOBAL}"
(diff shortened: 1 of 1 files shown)


dtrudg:
Hi @vsoch - would it be okay with you if I pulled out the change so it is not dependent on #1272 ?

I believe it just needs the following on current codebase:

```diff
-        app=(`echo $app | sed -e "s/-/_/g"`)
+        app=(`echo $app | sed -e "s/-/_/g" | sed -e "s/[.]/_/g"`)
```

#1272 is something we've identified as being merged into development, but not present on the release branches. The concern on pulling that change for 2.5 is that is changes all the environment namings, so there is likely a fair amount of doc change to do with it, that might delay a release of the whiteout things etc. Maybe we could target that as a 2.6 instead?

Let us know how you feel about this. Thanks.

vsoch:
Making only this change does not conform to the SCIF specification. You are breaking the specification and people's build recipes if you don't. Here is the specification, for detail about the namespace:

https://sci-f.github.io/spec-v1#environment-namespace

All of these are changes that were previously integrated into Singularity, and I hope that the work that I put into that isn't just... erased?  The paper is accepted, and will be published any day, and readers will expect to be able to use SCIF natively with Singularity. I put great effort into making this happen, had done several PRs to do these changes already, and it doesn't seem fair that my options are to do it at all again, or do nothing and have Singularity look bad. I don't know what is going on over there, but please make sure that the work that I already put into this is properly integrated back into the code base. If it takes a little longer, this is still the right thing to do. This is good and fair practice, I think.

dtrudg (approved):
This looks good to me as a fix for the reported issue, but would it be better long term to check in SCIF things that app names (after substitutions) are [a-zA-Z_]+[a-zA-Z0-9_]*  which is the POSIX requirement for env-vars. That could give a more specific error upstream.

vsoch:
oh I didn't know that! That is indeed a good idea. :)

GodloveD (approved):
This looks fine to me as well.

GodloveD:
Thanks @vsoch!

vsoch:
And thank you @GodloveD and @dctrud ! :hugs:

PR #1378, opened 2018-03-09
Bring in "makeit" and start hooking up libs and programs to build

yhcote:

This pull request brings in code to build the C files found in "core/". This is not complete but all of "src/lib" currently builds + the sif and mount progtram. 

**Checkoff for all PRs:**

- [NEVER] I have added changes to the [CHANGELOG](https://github.com/singularityware/singularity/blob/development/CHANGELOG.md) and and documentation updates to the [singularityware](https://www.github.com/singularityware/singularityware.github.io) documentation base.

Attn: @singularityware-admin


core/lib/runtime/enter.c
```diff
@@ -33,7 +33,7 @@
 #include "util/util.h"
 #include "util/message.h"
 #include "util/privilege.h"
-#include "./chroot/chroot.h"
+#include "runtime/chroot.h"
```
bauerm97, line 36: I don't see a reason to have `runtime/enter.c` and `runtime/chroot.c`, we can combine them I think. Maybe we can move the stuff in `runtime/chroot.c` into `runtime/enter.c`, and then the same with `runtime/chroot.h` -> `runtime/enter.h`. Thoughts?

core/lib/runtime/file-bind.c
```diff
@@ -37,7 +37,7 @@
 #include "util/privilege.h"
 #include "util/mount.h"
```
bauerm97: We're using underscores in file names everywhere, so I think `file_bind.c` fits a bit better as this file name. Same with `file_bind.h`

bauerm97 (changes requested):

core/lib/runtime/enter.c
yhcote: We certainly can. This pull request is big enough as it stands because of the new elements brought in. I wanna do that in another iteration if possible. Better tracking of similar changes. In other words, I didn't wanna touch the C/H files to much this turn around.

core/lib/runtime/file-bind.c
yhcote: Once again, copied over. I can make it more seamless the other turn around if it's cool with you.

bauerm97 (approved):

PR #1184, opened 2017-11-30
add support for proxy auth flow and secured S3-backed registries

jtriley:
**Description of the Pull Request (PR):**

NVIDIA cloud's private docker registry flow is slightly different than the flow used in singularity's python docker registry API and is supported by the official docker client. In addition to a slightly different flow, NVIDIA cloud's private docker registry is also backed by a secured S3 bucket which singularity currently doesn't handle properly. This PR fixes both of these issues and allows converting images from NVIDIA cloud's private docker registry into singularity images.

With the proxy auth challenge only the realm is returned and the only requirement is to auth against that realm URL (see singularityware/singularity#1159 for more details). To address this the challenge parsing regex and code has been updated to handle any combination of realm, service, and scope (as long as that order is maintained) and builds the token URL accordingly. **NOTE**: to be fully robust in our challenge parsing we'd need to use something like https://github.com/alexsdutton/www-authenticate but for now this is likely close enough for most docker registry use cases.

Once you get past the auth challenge parsing and token URL generation issue the next hurdle is that fetching layers involves a redirect to pre-signed S3 urls. Currently when this happens the auth credentials are passed along with the request which causes an InvalidArgument error about using more than one auth mechanism due to the pre-signed URL already containing an `X-Amz-Algorithm` query param and singularity additionally providing the `Authorization:` header with the bearer token (see singularityware/singularity#1159 for more details). This PR employs an `HTTPHandler` and `HTTPSHandler` with urllib to intercept all requests, inspect the URL for a pre-signed S3 URL, and if present strips any `Authorization:` header from the request. This should fix things for any private registry using a secured S3-backend - not just NVIDIA's.

This approach ensures that we're only stripping the `Authorization:` header in the case of S3-backed docker registries that use pre-signed S3 URLS where this is known be an issue (e.g. NVIDIA's private registry: `nvcr.io`). In all other cases the previous behavior should be preserved.

**This fixes or addresses the following GitHub issues:**

- Ref: #1159, #1183

**Checkoff for all PRs:**

- [x] I have read the [Guidelines for Contributing](https://github.com/singularityware/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- [ ] I have added changes to the [CHANGELOG](https://github.com/singularityware/singularity/blob/development/CHANGELOG.md) and documentation updates to the [singularityware](https://www.github.com/singularityware/singularityware.github.io) documentation base.
- [x] I have tested this PR locally with a `make test`
[...]

Issue #1183, opened 2017-11-30
pull images from ngc.nvidia.com


shahzebsiddiqui:
## Version of Singularity:
2.4

### Expected behavior
Should be able to pull container via singularity pull

Write here.

### Actual behavior
```
[siddis14@amrndhl1157 ~]$ singularity pull docker://nvcr.io/nvidia/caffe:17.11
WARNING: pull for Docker Hub is not guaranteed to produce the
WARNING: same image on repeated pull. Use Singularity Registry
WARNING: (shub://) to pull exactly equivalent images.
ERROR Unrecognized authentication challenge, exiting.
Cleaning up...
```

Setting environment variables with API key as password

```
[siddis14@amrndhl1157 ~]$ env | grep SINGULARITY
SINGULARITY_DOCKER_PASSWORD=anF0cjBmaXZkZTNvdWVlYTN0NGZhcGhxYWg6MDg1YjQxNmItYTY0ZC00NjYyLThjMzAtYmFjMmNiZWU4YmEz
SINGULARITY_DOCKER_USERNAME=shahzeb.siddiqui@pfizer.com

[siddis14@amrndhl1157 ~]$ singularity pull docker://nvcr.io/nvidia/caffe:17.11
WARNING: pull for Docker Hub is not guaranteed to produce the
WARNING: same image on repeated pull. Use Singularity Registry
WARNING: (shub://) to pull exactly equivalent images.
ERROR Unrecognized authentication challenge, exiting.
Cleaning up...

```

I am not sure what needs to be done when creating a token, the documentation needs more explanation

```
CREDENTIAL=$(echo -n vanessa:[password] | base64)
TOKEN=$(http 'https://auth.docker.io/token?service=registry.docker.io&scope=repository:vanessa/code-samples:pull' Authorization:"Basic $CREDENTIAL" | jq -r '.token')
```
### Steps to reproduce behavior

Please register at ngc.nvidia.com and follow instructions accordingly.

vsoch:
We've been working on this already and @jtriley has a PR solution. The Nvidia cloud flow is completely different than standard docker so the answer is not just in generating a token.

jtriley:
Yep I'm submitting a PR as we speak. Would love some testing once it's up. @shahzebsiddiqui I would recommend regenerating your API token if that value above is real :smile:

jtriley:
@shahzebsiddiqui PR is open if you'd like to test ^^

nouiz:
[...]

Issue #1159, opened 2017-11-13
build image from private Docker registry


renganxu:
## Version of Singularity:
2.4-dist

### Expected behavior

build Singularity image successfully from Docker image

### Actual behavior

singularity --verbose build tensorflow.img Singularity
Increasing verbosity level (2)
Singularity version: 2.4-dist
Exec'ing: /usr/local/libexec/singularity/cli/build.exec
Using container recipe deffile: Singularity
VERBOSE: Set messagelevel to: 2
VERBOSE: Initialize configuration file: /etc/singularity/singularity.conf
VERBOSE: Initializing Singularity Registry
INFO   : Sanitizing environment
VERBOSE: Set messagelevel to: 2
Adding base Singularity environment to container
/usr/local/libexec/singularity/functions: line 87: [: DEBUG: integer expression expected
/usr/local/libexec/singularity/functions: line 87: [: DEBUG: integer expression expected
VERBOSE Docker image: tensorflow:17.10
VERBOSE Registry: nvcr.io
VERBOSE Namespace: nvidia
VERBOSE Repo Name: tensorflow
VERBOSE Repo Tag: 17.10
VERBOSE Version: None
VERBOSE Obtaining tags: https://nvcr.io/v2/nvidia/tensorflow/tags/list
**ERROR Unrecognized authentication challenge, exiting.**
Cleaning up...

### Steps to reproduce behavior
I am trying to build a singularity image from a Docker image in Nvidia GPU cloud.
The following is a Singularity specification file:
```Bootstrap: docker```
```From: tensorflow:17.10```
```Registry: nvcr.io```
```Namespace: nvidia ```

then execute the following command:
```$ SINGULARITY_DOCKER_REGISTRY='--registry nvcr.io'```
```$ export SINGULARITY_DOCKER_USERNAME=$oauthtoken```
```$ export SINGULARITY_DOCKER_PASSWORD="my password" ```
```$ singularity --verbose build tensorflow.img Singularity```

All nvidia GPU cloud users use the username "$oauthtoken" and their own password.

**Comment by vsoch** [comment]
hey @hfutxrg ! I think their cloud perhaps has a different manifest or API? It's definitely something up with the authentication challenge - I will have a look for you.

I'm also a bit weary of the "DEBUG integer expression expected issue` and I'll see if I can reproduce that.

**Comment by vsoch** [comment]
okay here is the challenge:

```
challenge
[...]

PR #1383, opened 2018-03-12
1184 rebase - add support for proxy auth flow and secured S3-backed registries

dtrudg:

This is a rebase of #1184 from @jtriley  against development-2.x

Original PR has 1 'accept' review at present. Cherry picked cleanly except CONTRIBUTORS.md

-----
Description of the Pull Request (PR):

NVIDIA cloud's private docker registry flow is slightly different than the flow used in singularity's python docker registry API and is supported by the official docker client. In addition to a slightly different flow, NVIDIA cloud's private docker registry is also backed by a secured S3 bucket which singularity currently doesn't handle properly. This PR fixes both of these issues and allows converting images from NVIDIA cloud's private docker registry into singularity images.

With the proxy auth challenge only the realm is returned and the only requirement is to auth against that realm URL (see singularityware/singularity#1159 for more details). To address this the challenge parsing regex and code has been updated to handle any combination of realm, service, and scope (as long as that order is maintained) and builds the token URL accordingly. NOTE: to be fully robust in our challenge parsing we'd need to use something like https://github.com/alexsdutton/www-authenticate but for now this is likely close enough for most docker registry use cases.

Once you get past the auth challenge parsing and token URL generation issue the next hurdle is that fetching layers involves a redirect to pre-signed S3 urls. Currently when this happens the auth credentials are passed along with the request which causes an InvalidArgument error about using more than one auth mechanism due to the pre-signed URL already containing an X-Amz-Algorithm query param and singularity additionally providing the Authorization: header with the bearer token (see singularityware/singularity#1159 for more details). This PR employs an HTTPHandler and HTTPSHandler with urllib to intercept all requests, inspect the URL for a pre-signed S3 URL, and if present strips any Authorization: header from the request. This should fix things for any private registry using a secured S3-backend - not just NVIDIA's.

This approach ensures that we're only stripping the Authorization: header in the case of S3-backed docker registries that use pre-signed S3 URLS where this is known be an issue (e.g. NVIDIA's private registry: nvcr.io). In all other cases the previous behavior should be preserved.

This fixes or addresses the following GitHub issues:

Ref: #1159, #1183
Checkoff for all PRs:

 I have read the Guidelines for Contributing, and this PR conforms to the stated requirements.
 I have added changes to the CHANGELOG and documentation updates to the singularityware documentation base.
 I have tested this PR locally with a make test
 This PR is NOT against the project's master branch
 I have added myself as a contributor to the contributors's file
 This PR is ready for review and/or merge
Attn: @singularityware-admin

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -29,6 +29,7 @@
     - Jason Stover <jason.stover@gmail.com>
     - Jeff Kriske <jekriske@gmail.com>
     - Josef Hrabal <josef.hrabal@vsb.cz>
+    - Justin Riley <justin_riley@harvard.edu>
     - Maciej Sieczka <msieczka@sieczka.org>
     - Mark Egan-Fuller <markeganfuller@googlemail.com>
     - Nathan Lin <nathan.lin@yale.edu>
--- a/libexec/python/base.py
+++ b/libexec/python/base.py
@@ -19,13 +19,15 @@ import os
 
 
 try:
-    from urllib.parse import urlencode, urlparse
-    from urllib.request import urlopen, Request, unquote
+    from urllib.parse import urlencode, urlparse, parse_qs
+    from urllib.request import Request, unquote
+    from urllib.request import HTTPHandler, HTTPSHandler, build_opener
     from urllib.error import HTTPError
 except ImportError:
     from urllib import urlencode, unquote
-    from urlparse import urlparse
-    from urllib2 import urlopen, Request, HTTPError
+    from urlparse import urlparse, parse_qs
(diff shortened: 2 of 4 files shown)


jtriley:
Just FYI I think there’s a more general (security) fix here that should be implemented instead which is to always strip authorization headers when being redirected to a different domain. I believe that’s how the native docker client is handling this given that the http lib in golang is doing this by default IIRC. This will also remove the need to inspect every urlopen for S3 URLs which was always ugly.

jtriley:
(working on a patch today)

dtrudg:
Closing this as it's now taken up by #1390

PR #1402, opened 2018-03-15
Add compile.sh to kick off project builds

yhcote:


bauerm97 (approved):
Looking good Yannick :) Thanks for the assistance.

Issue #1391, opened 2018-03-14
%apphelp removes whitespaces/indentation


guma44:
## Version of Singularity: 2.4.2-dist

The %apphelp section strips indentation and empty lines.

### Expected behavior

Please keep the indentation and new lines. The help for an app is not readable at all.

### Steps to reproduce behavior

For the recipe:
```
BootStrap: docker
From: ubuntu:16.04

%apprun test
echo "Test app"

%apphelp test
This is a help

   without whitespaces

%help
This is a help

   with whitespaces


```

Running:
```
singularity help test.img
```
gives:
```

This is a help

   with whitespaces

```
And running:

```
singularity help --app test test.img
```

gives:
```
This is a help
without whitespaces
```

guma44:
So the question really is if the lack of whitespaces and newlines is crucial? At least in thI could easily fix this by adding IFS="" to get_section and removing if statemen ie. going from:
```
get_section() {

    SECTION="$1"
    RECIPE="$2"
    FOUND="no"

    while read -r line
    do

 
        if echo $line | grep -q -e "^%$SECTION$"; then

            # If not found, continue
            if [[ "$FOUND" == "no" ]]; then
                FOUND="yes"
            else
                return 0
            fi

        # We didn't match this line but...
        else
            # It's the start of a new section
            if echo $line | grep -q -e "^%"; then

                # We already found it, we're done
                if [[ "$FOUND" == "yes" ]]; then
                    return 0
                fi

            else
                # We already found it, still in section
                if [[ "$FOUND" == "yes" ]]; then
                    if [ ! -z "${line}" ]; then
                        echo $line
                    fi 

                fi            
            fi
        fi

    done < $RECIPE

}
```
To:
```
get_section() {

    IFS=""
    SECTION="$1"
    RECIPE="$2"
    FOUND="no"

    while read -r line
    do
        if echo $line | grep -q -e "^%$SECTION$"; then

            # If not found, continue
            if [[ "$FOUND" == "no" ]]; then
                FOUND="yes"
            else
                return 0
            fi

        # We didn't match this line but...
        else
            # It's the start of a new section
            if echo $line | grep -q -e "^%"; then

                # We already found it, we're done
                if [[ "$FOUND" == "yes" ]]; then
                    return 0
                fi

            else
                # We already found it, still in section
                if [[ "$FOUND" == "yes" ]]; then
                        echo $line
                fi            
            fi
        fi

    done < $RECIPE
}
```

[...]

PR #1404, opened 2018-03-16
Fix/appcommand indentation

guma44:

The commands starting with the %app* strip all white spaces and empty new lines when they are copied to the corresponding files. This is unintended behavior eg. the help section (%apphelp) might look unreadable. Additionally, the %runscript section does not require to strip the spaces. This fix also will cause the empty file to be created if the empty command is specified. This is an expected behavior IMHO.

- Ref: #1391 

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ and changes prior to that are (unfortunately) done retrospectively. Critical ite
 ### Bug Fixes
  - Put /usr/local/{bin,sbin} in front of the default PATH
  - Fixed bug that did not export environment variables for apps with "-" in name
+ - Fixed stripping whitespaces and empty new lines for the app commands
  
 ## [v2.4.3](https://github.com/singularityware/singularity/tree/release-2.4)
 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -43,3 +43,4 @@
     - David Trudgian <david.trudgian@utsouthwestern.edu>
     - Tarcisio Fedrizzi <tarcisio.fedrizzi@gmail.com>
     - Diana Langenbach <dcl@dcl.sh>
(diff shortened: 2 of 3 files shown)


dtrudg (commented):
I've had a look at this, understand the issues it fixes, and it tests okay, but I don't feel able to make a call on whether it's a-ok or needs further work.

All our example Singularity def files indent sections which is nice for legibility:

```
BootStrap: docker
From: busybox:latest   # This is a comment

%runscript

%post
    echo "Install additional software here"
```

... and I'm just not sure we want to be always leaving the spaces in for all the sections. 

I know @bauerm97 has looked a build file things in 3.0 - do you have an opinion on this?

dtrudg (approved):
After investigation of the original 'get-section.c' implementation which does not strip whitespace I'm :+1: on this.

GodloveD (approved):
@dctrud and I were both looking at this a bit and I think we reached the same conclusion.  afaik it looks like white space began being (inadvertently?) stripped here:

https://github.com/singularityware/singularity/commit/a73a7ed40a5eb80419c6faf329d1f4e4d8b237b7
Doesn't seem to have been a conscious decision, so it's probably a good idea to revert to original behavior.  As for whitespace accidentally added to runscripts and such, meh.  I think it will be fine, and if for any reason a use really doesn't like it, they can stop indenting in def file.  At least this way they have a choice.

GodloveD:
Oh bummer.  @guma44 would you mind rebasing onto development-2.x?

GodloveD:
Nevermind.  I rebased in #1482.  Thank you @guma44!

guma44:
@GodloveD No problem. I see you rebased it when I was asleep. Perfect.

GodloveD:
That's the best time for work to get done right?  😸 
🌙 
💤 
🛌

PR #1407, opened 2018-03-19
"nproc" may not always be present.

yhcote:


jmstover (approved):
Works for me

yhcote:
There ya "go" !

PR #1410, opened 2018-03-20
Update dependency management from trash -> dep

bauerm97:

This fully updates the dependency management tool from `trash` -> `dep`. This update corrects a few of the version problems and removes the legacy `vendor.conf`

Attn: @singularityware-admin


Gopkg.toml
```diff
@@ -50,8 +50,8 @@
   version = "1.0.1"
 
 [[constraint]]
+  branch = "master"
   name = "github.com/spf13/cobra"
-  version = "0.0.1"
```
tri-adam, line 54: Down the road it looks like COBRA will move to regular SemVer: https://github.com/spf13/cobra/releases

Are we agreed that it's best to track `master` until then?

tri-adam (approved):

dtrudg (approved):
This is good for me. Same thought as @tri-adam on the cobra dep though. The 0.0.1 tag is Oct 24th 2017 - so master may be more appropriate.

Gopkg.toml
bauerm97: Yep that's the plan. If we don't track master it doesn't build.

PR #1413, opened 2018-03-20
This removes the dep on gpgme

yhcote:


dtrudg (approved):
👍 For ref the docs indicate verification of pulled OCI images that might have signatures is still good with this:

Supported build tags
containers_image_openpgp: Use a Golang-only OpenPGP implementation for signature verification instead of the default cgo/gpgme-based implementation; the primary downside is that creating new signatures with the Golang-only implementation is not supported.

yhcote:
Thanks Dave T. !

PR #1416, opened 2018-03-22
Runtime engine

cclerget:

- Runtime engine
- Configuration file parser
- Start OCI configuration API for CLI and runtime

**Checkoff for all PRs:**

Attn: @singularityware-admin


pkg/util/capabilities/capabilities.go
```diff
@@ -0,0 +1,367 @@
+package capabilities
```
bauerm97, line 1: Thanks for doing this Cedric, this was something I needed to get around to implementing but hadn't had time yet :-)

runtime/run.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+echo "suid"
+./cli suid /tmp/testing.simg /bin/ls /
```
bauerm97, line 4: What work is left to be done to run SIF images here? I assume we need to use SIF binary to expose/mount the underlying FS?

pkg/runtime/oci/config/solaris.go
```diff
@@ -0,0 +1,3 @@
+// +build solaris
```
bauerm97, line 1: Do we need to include Solaris and Windows files? We're not planning on supporting containers on these systems, strictly Linux. Is it necessary for other reasons to have this here?

pkg/runtime/oci/config/config_linux.go
```diff
@@ -0,0 +1,9 @@
+package config
+
+const Platform = "linux"
+
+type RuntimeOciPlatform struct {
```
bauerm97, line 5: I'm slightly confused how you're using the OCI runtime config. I see you have defined an interface, does the implementation of the interface methods happen in the OCI runtime spec package?

pkg/runtime/oci/config/root.go
```diff
@@ -0,0 +1,50 @@
...
+	GetSpec() *specs.Root
+
+	GetPath() string
+	SetPath(path string)
+
+	GetReadOnly() bool
+	SetReadOnly(enabled bool)
+}
+
+type DefaultRuntimeOciRoot struct {
```
bauerm97, line 17: Why are each of these `DefaultRuntimeOci*` structs embedding a `RuntimeOciSpec` directly instead of the proper member. For instance, here why are we not having `RuntimeOciSpecRoot *RuntimeOciSpecRoot`?

pkg/runtime/runtime.go
```diff
@@ -1,13 +1,48 @@
...
+type OciRuntime interface {
+	State(id string) *specs.State
+	Create(id string, bundle string)
+	Start(id string)
+	Kill(id string, signal int)
+	Delete(id string)
+}
+
+// Runtime operations
+type Runtime interface {
```
bauerm97, line 31: This would be a good place to start the investigation into runtime plugins.

pkg/runtime/runtime.go
```diff
@@ -1,13 +1,48 @@
...
+	config "github.com/singularityware/singularity/pkg/runtime/config"
+)
+
+// Generic runtime engine
+type RuntimeEngine struct {
+	*config.RuntimeConfig
+	Runtime
+}
+
+type RuntimeCLI struct {
```
bauerm97, line 16: Is this supposed to be `type RuntimeOCI struct` by any chance?

internal/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,44 @@
...
+		engines = make(map[string]*runtime.RuntimeEngine)
+	}
+	engines[name] = engine
+	engine.RuntimeConfig = engine.InitConfig()
+	if engine.RuntimeConfig == nil {
+		log.Fatalf("failed to initialize %s engine\n", name)
+	}
+}
+
+func init() {
```
bauerm97, line 40: I think each `RuntimeEngine` should have an init in its own `.go` file. For example, the OCI transports package has an example of how to pull off this registerEngine effect. See:

https://github.com/containers/image/blob/master/transports/transports.go For how they describe the registering of image transports
https://github.com/containers/image/blob/master/transports/alltransports/alltransports.go For a list of examples of image transports that implement this

bauerm97 (changes requested):
Hey @cclerget it's looking really good so far. I haven't had time to pull it locally and play with it yet so I don't want to approve until I do that. 

That being said, I really like the direction this is going and I think we're on the right path!

pkg/runtime/oci/config/solaris.go
cclerget: Indeed, it's not necessary, I just split OCI spec between platform to make things clearer and easier

pkg/runtime/oci/config/config_linux.go
cclerget: No it just describe the whole interface with methods to implement in order to manipulate OCI specs

pkg/runtime/oci/config/root.go
cclerget: I take the easy way there, because if you embed the proper member, somewhere you have to link it against a common specs.Spec instance to generate JSON at the end

pkg/runtime/runtime.go
cclerget: Yes, it embed OciRuntime interface. As discussed yesterday OciRuntime interface need to be on both side

internal/pkg/runtime/runtime.go
cclerget: I took this way because my first thought was to embed all engines in one binary, so to not import all engines each time, I wrote a single init which register all engines resulting in one import for all engines. If we split engines into binaries, an init by engine is the right thing to do

runtime/run.sh
cclerget: Yes, Yannick could confirm, but I believe we need sif_load in core/lib/sif/sif.c to read SIF image and get data object descriptors to identify partitions and their offsets to mount correct ones

pkg/runtime/oci/config/root.go
bauerm97: Ahhh, makes sense. Thanks for clarifying :)

pkg/runtime/runtime.go
bauerm97: @cclerget I think you have a typo then ;) This reads `RuntimeCLI` instead of `RuntimeOCI`

cclerget:
@bauerm97 I added some missing files in last commit, I don't know why they have been dropped from previous commit :/. Thanks for review !

pkg/runtime/runtime.go
cclerget: @bauerm97 not a typo, RuntimeCLI struct contains OciRuntime interface for CLI side. I need to do the same RuntimeEngine

cclerget:
@bauerm97 CLI and runtime work together now, runtime compilation is integrated in compile.sh. Runtime mount all image types. CLI support three commands shell/exec/run with all namespaces and debug/verbose flags :

- `./core/buildtree/singularity exec -p image.simg -- ps -x`
- `./core/buildtree/singularity -d shell -u /tmp/dirimg`

Small bug with CLI, when using exec with command arguments containing "-", cobra parse options and don't recognize them, actually we must specify "--" to stop command line parsing

bauerm97 (approved):
Looks good! A few clean up things to fix and I can fix that bug you've described, but I'll get those into a next PR. Awesome work

PR #1418, opened 2018-03-23
Use double quotes for log messages written to syslog

treydock:

This change will help tools like Splunk parse the key=value where single quotes are added to value but double quotes are not.

Example of new message output from running `make test`

```
Mar 27 09:09:04 build-el7 Singularity: action-suid (U=20821,P=32716)> USER="tdockendorf", IMAGE="container.img", COMMAND="exec"
```

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -150,7 +150,7 @@ int main(int argc, char **argv) {
 
     fd_cleanup(&close_fd);
 
-    singularity_message(LOG, "USER=%s, IMAGE='%s', COMMAND='%s'\n", singularity_priv_getuser(), singularity_image_name(&image), singularity_registry_get("COMMAND"));
+    singularity_message(LOG, "USER=\"%s\", IMAGE=\"%s\", COMMAND=\"%s\"\n", singularity_priv_getuser(), singularity_image_name(&image), singularity_registry_get("COMMAND"));
 
     if ( command == NULL ) {
         singularity_message(INFO, "No action command verb was given, invoking 'shell'\n");
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and changes prior to that are (unfortunately) done retrospectively. Critical ite
  - migration guidance (how to convert images?)
  - changed behaviour (recipe sections work differently)
 
+## v2.5.x
+
+ - Use double quotes for log messages written to syslog
+
 ## [v2.5.0](https://github.com/singularityware/singularity/releases/tag/2.5.0-rc1) (2018-04-04)
 
 ### Security related fixes
--- a/src/slurm/singularity.c
+++ b/src/slurm/singularity.c
@@ -231,7 +231,7 @@ static int setup_container(spank_t spank)
     envar_set("SINGULARITY_SHELL", singularity_registry_get("SHELL"), 1);
 
     command = singularity_registry_get("COMMAND");
-    singularity_message(LOG, "USER=%s, IMAGE='%s', COMMAND='%s'\n", singularity_priv_getuser(), singularity_image_name(&image), singularity_registry_get("COMMAND"));
+    singularity_message(LOG, "USER=\"%s\", IMAGE=\"%s\", COMMAND=\"%s\"\n", singularity_priv_getuser(), singularity_image_name(&image), singularity_registry_get("COMMAND"));
 
     // At this point, the current process is in the runtime container environment.
     // Return control flow back to SLURM: when execv is invoked, it'll be done from


src/action.c
```diff
@@ -145,7 +145,7 @@ int main(int argc, char **argv) {
     envar_set("SINGULARITY_SHELL", singularity_registry_get("SHELL"), 1);
     envar_set("SINGULARITY_APPNAME", singularity_registry_get("APPNAME"), 1);
 
-    singularity_message(LOG, "USER=%s, IMAGE='%s', COMMAND='%s'\n", singularity_priv_getuser(), singularity_image_name(&image), singularity_registry_get("COMMAND"));
+    singularity_message(LOG, "USER=%s, IMAGE=\"%s\", COMMAND=\"%s\"\n", singularity_priv_getuser(), singularity_image_name(&image), singularity_registry_get("COMMAND"));
```
GodloveD: Is there any reason not to wrap the USER string as well?

GodloveD (changes requested):
This seems like a sensible change to me.  I'm going to ask for just a few more changes.
1. Can you also wrap the user string in ""?
2. Can you make an entry in the CHANGELOG.md?
3. Can you rebase this against development-2.x?
Thanks for taking the time to identify this need and provide a fix!

src/action.c
treydock: For at least Splunk the double quotes are only necessary when the value in key=value contains a space.  It's possible a space may exist, I would assume, for IMAGE, but probably not for USER.

treydock:
@GodloveD 

1. Should I put quotes around USER even though it's likely not necessary given value won't have spaces?
2. Okay to just put this change under v2.5 in CHANGELOG.md?
3. Should I target this PR against development-2.x and not master?  Rebase is painful due to CHANGELOG.md not having 2.4.5 in development-2.x

GodloveD:
@treydock 
1. Good point. Your call.
2. v2.5 is good.
3. Yeah, this PR really needs to be targeted to development-2.5.  If the rebase is a pain, you could always make a new branch from development-2.5 and then reapply your changes on top of the fresh branch.  Shouldn't be too hard since these changes are minimal.  I guess you would need to make a new PR, but that is fine.

treydock:
Retargeted this pull request to development-2.x, updated CHANGELOG.md and put double quotes around USER.  Rather have double quotes and not need them than some weird edge to cause USER=<value> to not get parsed without double quotes.

treydock:
@GodloveD Let me know if PR needs any adjustments.

treydock:
@GodloveD Now that 2.5.0-rc1 is released, under which version should I added a changelog entry?

GodloveD:
Heya @treydock sorry we didn't get this reviewed and merged before the 2.5.0 release candidate.  I took the liberty of resolving the conflict on the changelog myself and putting the note where I thought it was best.  As of now, I think this looks good.  If somebody else looks it over and is happy I would be glad to merge.  Thanks!

GodloveD (approved):

treydock:
@GodloveD Anything needed before this is able to be merged?

dtrudg:
Hi @treydock - can you explain a little more why double quotes are better in this case? I'm trying to see, but I think the same issues come up with either double quotes or single quotes.

If single quotes we could end up with `COMMAND='echo 'bob''`

If double quotes we could equally end up with `COMMAND="echo "bob""`

If we are going from a parsing issue in one situation, and replacing it with a parsing issue in a different situation I'd be inclined not to make the change as it isn't gaining us anything overall.

Could you give a little more detail in case I'm missing anything here? Thanks!

treydock:
@dctrud The issue comes down to how Splunk handles key=value pairs in syslog.  Splunk will see `COMMAND='foo'` as the field `COMMAND` with the value being `'foo'` (single quotes in the value).  If the string in syslog is `COMMAND="foo"` then Splunk treats the key as `COMMAND` and the value as `foo` (no quotes). 

This parsing quirk may be unique to Splunk or may be shared with other log analysis tools, I don't have any other tools to test with.

The reason field extraction is useful is for reports and searches.  One could say `search process=Singularity IMAGE=foo` to find all log messages where `IMAGE="foo"` but if the single quotes are used the search would have to be something like `search process=Singularity IMAGE="'foo'"`.

I've worked around the issue locally but it only works when generating reports: `process=Singularity | rex field=IMAGE mode=sed "s/'//g" | rex field=COMMAND mode=sed "s/'//g" | stats count by USER,IMAGE,COMMAND | sort - count`.  Ad-hoc searches without the double quotes change still require some counter-intuitive quoting.

Let me know if this helps explain the motivation of this change.

dtrudg:
@treydock - thanks for the explanation, it does make it clear what you are trying to do here.

I'm afraid that in the absence of knowledge about what other log parsers do, and with consideration that other existing users may have already setup rules which would be broken by a switch from `'` to `"`, I don't really think this is a good change for a point release. If we have input that suggests it would broadly help others, then a behavior change here may be something to consider for a minor release.

Regarding splunk, some searches show that single quote parsing comes up a fair amount. It looks like single quotes can be seen in SYSLOG from e.g. Oracle: https://answers.splunk.com/answers/109734/oracle-11g-standard-audit-to-syslog-what-is-normal-for-single-quotes.html

Does this page replicate what you are doing with your workaround? https://answers.splunk.com/answers/342374/can-you-tell-auto-kv-to-honor-values-within-single.html

treydock:
@dctrud I had not come across the prop/transforms work around, thanks for the link.  That is a index-time extraction which is more efficient than my work around which was done at search time.  Given how disruptive this change could be to other Singularity users I am going to close this in favor of adapting my Splunk instance.

GodloveD:
Hey @treydock I feel like we've been stringing you along.  Sorry about that.  Thanks for taking the time to identify this potential enhancement and submit it even though it didn't get in.  Hope that you will submit again if you have another idea.

PR #1422, opened 2018-03-27
Framework for automagically generating man pages

GodloveD:

~This is a WIP and shouldn't be merged yet.  Just here for discussion.~
EDIT: I'd now like this to be considered for merge.

The basic idea is that we compile a second binary (called generate_docs or some such thing).  When called, this doc will use the cmd structs to create the man pages for you.  

Another idea, is that we could make a hidden option or command to work with the `singularity` parent command that would generate the docs.  In some ways, I think that is a bit messier, but it does have the added benefit of keeping everything contained in a single binary.

TODO:
1. Still need to investigate some of the formatting.  For instance, usage string is not yet working properly and I don't know why.
2. Need to write a custom help function to be called that will display these man pages instead of the current help.
3. Need to figure out how to automatically create these man pages at compile time.
4. Must have a good way to install these man pages into the `MANPATH` 

**Checkoff for all PRs:**

Attn: @singularityware-admin


src/cmd/sbuild/cli/sbuild.go
```diff
@@ -25,8 +43,9 @@ func ExecuteSbuild() {
 }
 
 var sbuildCmd = &cobra.Command{
-	Use:  "sbuild <builder type> <definition json> <image path>",
```
bauerm97, line 28: sbuild is not a command the user should *ever* be touching. It is strictly an under the hood construct for the implementation of image building.

src/cmd/singularity/cli/singularity.go
```diff
@@ -48,169 +82,3 @@ func TraverseParentsUses(cmd *cobra.Command) string {
...
-	//goflag.Parse()
-
-	singularityCmd.Flags().BoolVarP(&debug, "debug", "d", false, "Print debugging information")
-	singularityCmd.Flags().BoolVarP(&silent, "silent", "s", false, "Only print errors")
-	singularityCmd.Flags().BoolVarP(&quiet, "quiet", "q", false, "Suppress all normal output")
-	singularityCmd.Flags().BoolVarP(&verbose, "verbose", "v", false, "Increase verbosity +1")
-
-	cobra.AddTemplateFuncs(templateFuncs)
-
-	singularityCmd.SetHelpTemplate(
```
bauerm97, line 116: I'm worried that we will again be fighting against Cobra here. singularity help does not *necessarily* need to open the man page for the command. We could still have a very pretty templated help function in addition to man pages. Thoughts?

src/cmd/singularity/cli/build.go
```diff
@@ -29,23 +31,150 @@ var (
...
-	buildCmd.Flags().BoolVarP(&Sandbox, "sandbox", "s", false, "Build image as sandbox format (chroot directory structure)")
-	buildCmd.Flags().StringSliceVar(&Sections, "section", []string{}, "Only run specific section(s) of deffile")
-	buildCmd.Flags().BoolVarP(&Writable, "writable", "w", false, "Build image as writable (SIF with writable internal overlay)")
-	buildCmd.Flags().BoolVarP(&Force, "force", "f", false, "")
-	buildCmd.Flags().BoolVarP(&NoTest, "notest", "T", false, "")
-	buildCmd.Flags().BoolVarP(&Remote, "remote", "r", false, "Build image remotely")
-	buildCmd.Flags().StringVar(&RemoteURL, "remote-url", "localhost:5050", "Specify the URL of the remote builder")
-	buildCmd.Flags().StringVar(&AuthToken, "auth-token", "", "Specify the auth token for the remote builder")
+
+	manHelp := func(c *cobra.Command, args []string) {
```
bauerm97: Alright so I'm going to go against something that I believe I did say earlier here, but I'm not sure this is the right approach to the `singularity help *` command. What if the user does *not* install the man pages on their system? Then this will error out. What if for whatever reason the user does not have man installed, how will this behave? This also shoehorns the CLI into Linux only. Again, I'm rethinking if this is the best approach here

src/cmd/singularity/cli/actions.go
```diff
@@ -17,51 +17,131 @@ import (
...
 	actionCmds := []*cobra.Command{
-		execCmd,
-		shellCmd,
-		runCmd,
+		ExecCmd,
+		ShellCmd,
+		RunCmd,
+	}
+
+	// TODO : the next n lines of code are repeating too much but I don't
```
bauerm97: So *if* we decide to stick with this approach of running the man pages when a user calls help, this is not a great way to define these help functions. Within the [cobra.Command](https://godoc.org/github.com/spf13/cobra#Command) there is a field `Command.Annotations map[string]string`. This field allows (essentially) arbitrary meta data to be embedded within a `cobra.Command` type. So if (again, big if) we decide to stick with this approach, we can do something like this within the initialization of the `.Command` structures. 

```
var ExecCmd = &cobra.Command{
	DisableFlagsInUseLine: true,
 	Args: cobra.MinimumNArgs(2),
 		a := append([]string{"/.singularity.d/actions/exec"}, args[1:]...)
 		execWrapper(cmd, args[0], a)
 	},

	Use:     execUse,
	Short:   execShort,
	Long:    execLong,
 	Example: execExamples,
        Annotations: map[string]string{
                "manstring": "singularity-exec",
        },
 }
```

Then a general help function can access `c.Annotations["manstring"]` to get the man string for that specific command.

bauerm97 (changes requested):
Sorry for switching some things up on you, but I want to be very sure on the direction we go

src/cmd/singularity/cli/actions.go
GodloveD: Oh cool.  I wanted to do something like that, but I didn't know how.

src/cmd/singularity/cli/build.go
GodloveD: Well, I don't think that we need to make a decision on this right this minute.  If we decide that we want to write up templates and generate proper help separate from the man pages, then fine.  But I think that will be in addition to generating man pages for each of the commands.  We would probably want to do both.  

Here is what I suggest.  Let's go ahead and merge this and use it as a temporary measure until we can write templates for all of the help commands.  

BTW, I'm actually not crazy about using templates because they seem overly complicated and then we are back in the business of maintaining help docs even though we are using cobra.  But I'm fine using them if we really want to have help that is separate from the man pages.

src/cmd/singularity/cli/singularity.go
GodloveD: OK.  Redid everything with the template instead.  Check it out.

src/cmd/sbuild/cli/sbuild.go
GodloveD: Got it. Removed it from singularity command.

PR #1436, opened 2018-04-03
Fix for check_mounted() to check parent directories

gmkurtzer:

Fixes `check_mounted()` to properly check all parent directories.

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/src/util/mount.c
+++ b/src/util/mount.c
@@ -107,8 +107,9 @@ int check_mounted(char *mountpoint) {
                     free(test_mountpoint);
                     goto DONE;
                 }
+                free(linktarget);
             } else {
-                full_test_path = tmp_test_path;
+                full_test_path = strdup(tmp_test_path);
             }
 
             // Check to see if mountpoint is already mounted
@@ -119,7 +120,8 @@ int check_mounted(char *mountpoint) {
                 goto DONE;
             }
             test_mountpoint = dirname(test_mountpoint);
-
+            free(tmp_test_path);
+            free(full_test_path);
         }
 
         free(test_mountpoint);
--- a/src/util/mount.c
+++ b/src/util/mount.c
@@ -100,6 +100,53 @@ int singularity_mount(const char *source, const char *target,
     return ret;
 }
 
+int check_proc_mount(char *mount, char *real_mountpoint) {
+    int retval = -1;
+
+    char *dup_mountpoint = strdup(real_mountpoint);
+    char *test_mountpoint = dup_mountpoint;
+
+    while ( ( retval < 0 ) && ( strcmp(test_mountpoint, "/") != 0 ) ) {
+        char *full_test_path = NULL;
+        char *tmp_test_path = joinpath(CONTAINER_FINALDIR, test_mountpoint);
+
+        if ( is_link(tmp_test_path) == 0 ) {
+            char *linktarget = realpath(tmp_test_path, NULL); // Flawfinder: ignore
+            if ( linktarget == NULL ) {
+                singularity_message(ERROR, "Could not identify the source of contained link: %s\n", test_mountpoint);
+                ABORT(255);
+            }
+            full_test_path = joinpath(CONTAINER_FINALDIR, linktarget);
+            singularity_message(DEBUG, "Parent directory is a link, resolved: %s->%s\n", tmp_test_path, full_test_path);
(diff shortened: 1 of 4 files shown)


src/util/mount.c
```diff
@@ -79,24 +80,37 @@ int check_mounted(char *mountpoint) {
...
     }
 
-    real_mountpoint = realpath(mountpoint, NULL); // Flawfinder: ignore
-    if ( real_mountpoint == NULL ) {
-        // mountpoint doesn't exist
-        return(retval);
+    if ( is_link(mountpoint) == 0 ) {
+        real_mountpoint = realpath(mountpoint, NULL); // Flawfinder: ignore
+        if ( real_mountpoint == NULL ) {
+            // mountpoint doesn't exists
```
DrDaveD: Why does that trailing 's' in exists keep popping back up?  That's a regression from pr #1317

src/util/mount.c
```diff
@@ -79,24 +80,37 @@ int check_mounted(char *mountpoint) {
...
+           singularity_message(DEBUG, "returning, real_mountpoint == NULL\n");
```
DrDaveD: This should be indented one more space

src/util/mount.c
```diff
@@ -79,24 +80,37 @@ int check_mounted(char *mountpoint) {
...
     while ( fgets(line, MAX_LINE_LEN, mounts) != NULL ) {
         (void) strtok(strdup(line), " ");
         char *mount = strtok(NULL, " ");
 
-        // Check to see if mountpoint is already mounted
-        if ( strcmp(joinpath(rootfs_dir, real_mountpoint), mount) == 0 ) {
-            singularity_message(DEBUG, "Mountpoint is already mounted: %s\n", mountpoint);
-            retval = 1;
-            break;
+	char *test_mountpoint = strdup(real_mountpoint);
```
DrDaveD: This contains the only tab in the file, that's why the indent looks wrong.  Spaces would be better.

DrDaveD (approved):
I reviewed the code, and tested it on el7 applied to the release-2.4 branch.  Looks good.

Note that there's one underlying source code difference in the release-2.4 branch (that pesky "doesn't exists" comment) so I believe that means it will need a separate PR or at least need a manual merge.

DrDaveD:
I rebuilt at this point & tested and it also looks good here (better actually) except I noticed that the return value from dirname() is getting freed and man -s3 dirname says to not do that.

Duplication of free() statements (and possibly the gotos) could also be avoided by checking the value of retval after the free statements, using "!retval &&" in the while and if conditions..

src/util/mount.c
```diff
@@ -79,24 +82,50 @@ int check_mounted(char *mountpoint) {
...
+            }
+            test_mountpoint = dirname(test_mountpoint);
+            free(tmp_test_path);
+            free(full_test_path);
         }
 
+        free(test_mountpoint);
+
         // Check to see if path is in container root
         if ( strncmp(rootfs_dir, mount, strlength(rootfs_dir, 1024)) != 0 ) {
```
DrDaveD: I think this if and the following one don't do anything.  What's the point of doing "continue" at the end of a loop?

tri-adam:
Hey @DrDaveD , I've refactored that function and split the inner loop out into another function to make the code easier to follow (and free of `goto`s!) I think I've also addressed your comments regarding the free and the conditional `continue`s without effect at the end of the outer loop.

Let me know what you think!

yhcote (approved):

src/util/mount.c
```diff
@@ -51,22 +84,77 @@ int singularity_mount(const char *source, const char *target,
...
+}
+
 int check_mounted(char *mountpoint) {
     int retval = -1;
     FILE *mounts;
     char *line = (char *)malloc(MAX_LINE_LEN);
-    char *rootfs_dir = CONTAINER_FINALDIR;
     unsigned int mountpoint_len = strlength(mountpoint, PATH_MAX);
-    char *real_mountpoint;
+    char *real_mountpoint = strdup(mountpoint);
```
DrDaveD, line 155: I don't see this real_mountpoint getting freed anywhere.

DrDaveD: Nevermind, it was in the original file, it just doesn't show up in the changeset.

DrDaveD:
The final version looks good and it passed my el7 tests.

PR #1438, opened 2018-04-03
free strdupped temporary variable in joinpath

DrDaveD:

I noticed a strdup() in joinpath() that was not getting freed and not needed after the function returns.  This adds the free().

- None

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/src/util/util.c
+++ b/src/util/util.c
@@ -191,6 +191,8 @@ char *joinpath(const char * path1, const char * path2_in) {
         ABORT(255);
     }
 
+    free(tmp_path1);
+
     return(ret);
 }
 


src/util/util.c
```diff
@@ -191,6 +191,8 @@ char *joinpath(const char * path1, const char * path2_in) {
         ABORT(255);
```
GodloveD, line 191: Should the free also go in this if statement above?

GodloveD (changes requested):
Nice catch.  Should the free also go into the preceding `if` statement just in case the function bombs out?

src/util/util.c
DrDaveD: I didn't think it was worth doing since the program dies there anyway.

GodloveD: Your right.

GodloveD (approved):

ctmadison (approved):

tri-adam (approved):

lyklev:
I noticed that indeed where joinpath and strdup are used, most of the time it leaks memory. It was my intention to fix it in a few places, but the memory leaking is quite common. I have started working on it, and will probably fix the memory leaks in small steps, in many commits to make the fix easier to test and review.

lyklev:
Yes, it is, it is one of the many places where strdup without a free leaks memory.

It was my intention to fix the memory leaks bit by bit, mostly by using static variables where possible. In fact, some of the fixes are already in my "strings-fix" branch for which I have not made any pull requests yet - but feel free to have a look.

PR #1444, opened 2018-04-05
Bump version to 2.4.5.99rc2

jmstover:

Bump the version for the RC release.

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,5 +1,5 @@
 AC_PREREQ(2.59)
-AC_INIT([singularity],[2.4.6],[gmkurtzer@gmail.com])
+AC_INIT([singularity],[2.4.5.99rc2],[gmkurtzer@gmail.com])
 
 if test -z "$prefix" -o "$prefix" = "NONE" ; then
   prefix=${ac_default_prefix}
--- a/configure.ac
+++ b/configure.ac
@@ -1,5 +1,5 @@
 AC_PREREQ(2.59)
-AC_INIT([singularity],[2.4.5],[gmkurtzer@gmail.com])
+AC_INIT([singularity],[2.4.6-RC2],[gmkurtzer@gmail.com])
 
 if test -z "$prefix" -o "$prefix" = "NONE" ; then
   prefix=${ac_default_prefix}
--- a/configure.ac
+++ b/configure.ac
@@ -1,5 +1,5 @@
 AC_PREREQ(2.59)
-AC_INIT([singularity],[2.4.6],[gmkurtzer@gmail.com])
+AC_INIT([singularity],[2.4.6rc2],[gmkurtzer@gmail.com])
 
 if test -z "$prefix" -o "$prefix" = "NONE" ; then
   prefix=${ac_default_prefix}
--- a/configure.ac
+++ b/configure.ac
@@ -1,5 +1,5 @@
 AC_PREREQ(2.59)
-AC_INIT([singularity],[2.4.6-RC2],[gmkurtzer@gmail.com])
+AC_INIT([singularity],[2.4.6],[gmkurtzer@gmail.com])
 
 if test -z "$prefix" -o "$prefix" = "NONE" ; then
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -1,5 +1,5 @@
 AC_PREREQ(2.59)
-AC_INIT([singularity],[2.4.5],[gmkurtzer@gmail.com])
+AC_INIT([singularity],[2.4.6rc2],[gmkurtzer@gmail.com])
```
GodloveD: I think it should be `2.4.6-rc2` (<-note the dash).  Right?

GodloveD (changes requested):

configure.ac
jmstover: rpmbuild blows up when there's a '-' in the version. Seems to work fine without it.

GodloveD: Stupid rmpbuild.  😕 Guess that means that 2.5.0-rc probably isn't building right now either.  I guess it didn't get run through Travis since I just pushed directly to the repo and then tagged?

configure.ac
```diff
@@ -1,5 +1,5 @@
 AC_PREREQ(2.59)
-AC_INIT([singularity],[2.4.5],[gmkurtzer@gmail.com])
+AC_INIT([singularity],[2.4.6],[gmkurtzer@gmail.com])
```
GodloveD: I think this really needs to be 2.4.6rc2 because of all of the potential problems that others have noted.  I know that this may cause some challenges with release numbers, but at least folks will actually see what version they are running using the `--version` command.

debian/changelog
```diff
@@ -1,3 +1,8 @@
+singularity-container (2.4.6-1) unstable; urgency=high
```
GodloveD: I think this should actually be 2.4.6rc2-1 as well just so that we differentiate between an actual release here.  We can change it when we do the real release.  Thoughts?

GodloveD (changes requested):
I still don't think this is just right.  meh.

configure.ac
jmstover: Version for me is:
```
$ singularity --version
2.4.6-ver_fix.gcef599d
```

So the branch (ver_fix) is included as well as the git commit. If we have the 'rc' included, then a RPM update wont work because `2.4.6 < 2.4.6rc`.

For _testing_ it should be being built from the git tree and not a dist tarball anyway. It's not meant to stick around.

GodloveD: > So the branch (ver_fix) is included as well as the git commit. 

Well, sure.  But a soon as we make the 2.4.6-rc2 tag, this thing will return...
```
$ singularity --version
2.5.6-dist
```
That's a problem because now a user thinks they have the real deal when they actually mistakenly installed a release candidate and we can't tell the difference if they send us a bug report either.  

As far as the update issue, if somebody has created and installed an RPM from a release candidate then presumably they are savvy enough to know how to remove the release candidate and install the release.  And if they forget or don't know how the `--version` command should be able to tell them that they don't have the latest release.  Right?

debian/changelog
jmstover: Debian appears to be the same:

https://www.debian.org/doc/debian-policy/#version-numbers-based-on-dates

configure.ac
jmstover: >  And if they forget or don't know how the --version command should be able to tell them that they don't have the latest release.

No. It will show:  `2.4.6-dist` or `2.5.0-dist` when building from a tarball (i.e. creating a package). When you do `make dist` the `.git/` directory isn't there, so it doesn't look up the git information and version becomes:  `@PACKAGE_VERSION@-dist` instead of `@PACKAGE_VERSION@-[branch](-[git commit])`

Look at `configure.ac` ~line 314

dtrudg:
@jmstover - are we good to close this now since 2.4.6 release was issued?

jmstover:
Yup. Good to close.

PR #1450, opened 2018-04-11
fix some interface disparities

yhcote:
This fixes the build by updating code that no longer satisfied an interface.


bauerm97:
@yhcote I think this was resolved by #1451, can you confirm?

bauerm97:
I think this was solved by #1451, if it needs fixing re-open the issues. Thanks @yhcote

PR #1451, opened 2018-04-11
Repay technical debt in Singularity 3.0 tree

bauerm97:

This PR begins to repay technical debt in the Singularity 3.0 tree. First this moves The C code from @cclerget into the `core/` directory and integrates it into the makeit build system (courtesy @yhcote). This also fixes output handling for image builds.

Attn: @singularityware-admin


dtrudg (dismissed):
This gives a working `compile.sh` invocation and `singularity build docker://` for me.

Cannot run resulting container on Debian 9 at present, failing with: 
```
2018/04/11 10:15:04 Container setup failed
2018/04/11 10:15:04 can't open network namespace: open /proc/13983/ns/net: no such file or directory
2018/04/11 10:15:04 exit status 1
```

... but this is a good PR to continue library/builder/cli work on top off :+1:

ArangoGutierrez (dismissed):
`./compile.sh` works
`singularity build` works
`singularity action` not-working

we still depend to run on a host with 2.X installed for `singularity.config`

core/runtime/c/cgo_scontainer.c
```diff
@@ -61,33 +61,39 @@ __attribute__((constructor)) static void init(void) {
     int ret;
 
     if ( stage <= 0 ) {
-        pfatal("STAGE environement variable not set");
+        singularity_message(ERROR, "STAGE environement variable not set\n");
```
cclerget, line 64: why not use one function like pfatal or something like `singularity_abort` rather than always call exit() just after ?

core/runtime/c/wrapper.c
```diff
@@ -430,55 +460,60 @@ int main(int argc, char **argv) {
...
         }
 
         if ( WIFEXITED(status) || WIFSIGNALED(status) ) {
             if ( WEXITSTATUS(status) != 0 ) {
-                pfatal("Child exit with status %d", WEXITSTATUS(status));
+                singularity_message(ERROR, "Child exit with status %d\n", WEXITSTATUS(status));
+                exit(1);
             }
         }
         close(stage_socket[0]);
```
cclerget, line 475: Can you remove this line ? It's an oversight on my side because it can close stdin/stdou/stderr depending of initialization value. Thanks

cclerget (changes requested):
Looks good ! Just replace MESSAGELEVEL by SINGULARITY_MESSAGELEVEL line https://github.com/singularityware/singularity/pull/1451/files#diff-496121da2ac39a1f8d063cda7ca496c9R174

core/runtime/c/wrapper.c
bauerm97: I fixed `MESSAGELEVEL` issue, but not sure exactly which line to remove here. Can you submit PR on top of this with just that small change?

cclerget: Done

cclerget (approved):

PR #1452, opened 2018-04-11
Push / pull minimal code path - Depends on #1451

dtrudg:
** This PR stacks on #1451 and should not be merged until #1451 is merged **

Add minimal code path for working Golang CLI push and pull verbs.


bauerm97 (approved):
Looks good @dctrud Thanks!

PR #1453, opened 2018-04-11
Remove close on uninitialized variable - depends on #1451

cclerget:
Fix an oversight in core runtime wrapper on top of #1451


bauerm97:
Thanks for doing this for me @cclerget , merging

bauerm97 (approved):

PR #1454, opened 2018-04-11
Fix push/pull related package deps

dtrudg:
Install of the pb dep for the library push / pull was broken recently if taking it from GitHub, due to v2 work there. Switch to the versioned pb package source via gopkg.in to track v1 without introducing hard constraint.


bauerm97 (approved):
Looks really good!

PR #1456, opened 2018-04-12
Mksquashfs opts

ArangoGutierrez:

Enable users to choose comp at build time

```
--comp <comp>    select squashfs Compressors 
                    available:
                        gzip (default)
                        lzma
                        lzo
                        xz
```

example for lzo

```
sudo singularity build --comp lzo lzo_comp.img testdef
```

Comp differences:

```bash
-rwxr-xr-x.  1 root    root     36M Apr 12 15:45 default_comp.img
-rwxr-xr-x.  1 root    root     39M Apr 12 15:41 lzo_comp.img
-rwxr-xr-x.  1 root    root     30M Apr 12 15:41 xz_comp.img
```

**Checkoff for all PRs:**

Attn: @singularityware/singularity-admin

### DIFF
--- a/src/lib/image/squashfs/init.c
+++ b/src/lib/image/squashfs/init.c
@@ -29,6 +29,7 @@
 #include <errno.h> 
 #include <string.h>
 #include <fcntl.h>  
+#include <linux/types.h>
 
 #include "util/message.h"
 #include "util/util.h"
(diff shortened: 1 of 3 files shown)


src/lib/image/squashfs/init.c
```diff
@@ -76,8 +93,41 @@ int _singularity_image_squashfs_init(struct image_object *image, int open_flags)
...
+            compression_type = strdup("xz");
+            break;
+        case SQUASHFS_LZ4:
+            compression_type = strdup("lz4");
+            break;
+        default:
+            break;
+        }
+
+        if ( compression_type ) {
```
bauerm97, line 126: I'm not sure it's necessary to spit out a warning every time an image is run that was compressed. I think it makes more sense to output this to DEBUG.

src/lib/image/squashfs/init.c
```diff
@@ -76,8 +93,41 @@ int _singularity_image_squashfs_init(struct image_object *image, int open_flags)
...
+            singularity_message(WARNING, "squashfs image was compressed with %s, if it failed to run, " \
+                                         "please contact image's author\n", compression_type);
```
bauerm97, line 128: We shouldn't put things on two lines like this, I would say just make the text all one string.

bauerm97 (changes requested):

src/lib/image/squashfs/init.c
cclerget: Was discussed on slack to display a warning because that would be source of confusion for users downloading image not compressed with gzip by default, some kernel don't support all squashfs compression format by default, then display a warning here make sense to warn users about potential problem

src/lib/image/squashfs/init.c
cclerget: Ok, will apply changes

PR #1459, opened 2018-04-13
Add push and pull PGP material

yhcote:


pkg/signing/signing.go
```diff
@@ -223,6 +367,70 @@ func SifAddSignature(sinfo *image.Sifinfo, signature []byte) error {
 	return nil
 }
 
+func sykeysFetchPubkey(fingerprint string, sykeysAddr string) (openpgp.EntityList, error) {
+	url := "http://" + sykeysAddr + "/pks/lookup?op=get&options=mr&search=0x" + fingerprint
```
tri-adam, line 371: I'd recommend using a `url.Values` structure for the query params, and a `url.URL` to build up the URL itself. That way, you do not have to worry about whether the params are properly URL encoded. For example:

	v := url.Values{}
	v.Set("op", "get")
	v.Set("options", "mr")
	v.Set("search", "0x" + fingerprint)
	u := url.URL{
		Scheme: "http",
		Host:   sykeysAddr,
		Path:   "pks/lookup",
		RawQuery: 	v.Encode(),
	}

And then you can do `http.Get(url.String())`.

pkg/signing/signing.go
```diff
@@ -223,6 +367,70 @@ func SifAddSignature(sinfo *image.Sifinfo, signature []byte) error {
...
+	if err != nil {
+		log.Println("armor.Encode failed")
+	}
+
+	err = entity.Serialize(wr)
+	if err != nil {
+		log.Println("can't serialize public key", err)
+		return err
+	}
+	wr.Close()
```
tri-adam, line 413: Might be better to `defer wr.Close()` above after you're sure `Encode` has succeeded?

pkg/signing/signing.go
```diff
@@ -223,6 +367,70 @@ func SifAddSignature(sinfo *image.Sifinfo, signature []byte) error {
...
+		return err
+	}
+
+	text, err := ioutil.ReadAll(resp.Body)
+	resp.Body.Close()
+	if err != nil {
+		log.Println(err)
+		return err
+	}
+	fmt.Printf("%s", text)
```
tri-adam, line 429: I think you can replace this entire section (starting with the `ioutil.ReadAll` line) with:

```
defer resp.Body.Close()

if _, err := io.Copy(os.Stdout, resp.Body); err != nil {
	log.Println(err)
	return err
}
```

pkg/signing/signing.go
```diff
@@ -235,88 +443,82 @@ func SifAddSignature(sinfo *image.Sifinfo, signature []byte) error {
...
 
-	if err = SifAddSignature(&sinfo, signedmsg.Bytes()); err != nil {
+	if err = sifAddSignature(en.PrimaryKey.Fingerprint, &sinfo, signedmsg.Bytes()); err != nil {
 		return err
 	}
 
 	return nil
 }
 
 func Verify(cpath string) error {
```
tri-adam, line 511: From `golint`:

```
pkg/signing/signing.go:511:1: exported function Verify should have comment or be unexported
pkg/signing/signing.go:541:21: error strings should not be capitalized or end with punctuation or a newline
pkg/signing/signing.go:567:22: error strings should not be capitalized or end with punctuation or a newline
```

pkg/signing/signing.go
```diff
@@ -235,88 +443,82 @@ func SifAddSignature(sinfo *image.Sifinfo, signature []byte) error {
  * 6) record the KeyID used to sign into signature data object descriptor
  */
 func Sign(cpath string) error {
```
tri-adam, line 445: From `golint`:

```
pkg/signing/signing.go:434:1: comment on exported function Sign should be of the form "Sign ..."
```

pkg/signing/signing.go
```diff
@@ -223,6 +367,70 @@ func SifAddSignature(sinfo *image.Sifinfo, signature []byte) error {
 	return nil
 }
 
+func sykeysFetchPubkey(fingerprint string, sykeysAddr string) (openpgp.EntityList, error) {
```
tri-adam, line 370: It might make sense to put the stuff that deals with the keyserver in a separate file `sykeys.go`? This seems separate in purpose from actually signing/verifying, which is already quite complex.

pkg/signing/signing.go
```diff
@@ -223,6 +367,70 @@ func SifAddSignature(sinfo *image.Sifinfo, signature []byte) error {
...
+	url := "http://" + sykeysAddr + "/pks/lookup?op=get&options=mr&search=0x" + fingerprint
+	log.Println("url :", url)
+	res, err := http.Get(url)
+	if err != nil {
+		log.Println(err)
+		return nil, err
+	}
+
+	el, err := openpgp.ReadArmoredKeyRing(res.Body)
+	res.Body.Close()
```
tri-adam, line 380: Using `defer res.Body.Close()` recommended, close as possible to the `http.Get`.

tri-adam (changes requested):
Just a few suggestions.

pkg/signing/signing.go
yhcote: This is the intention, of course, this will require a complete rework of that code which I'd rather do when all use cases are working simply. Which is about now.

pkg/signing/signing.go
yhcote: Will do !

pkg/signing/signing.go
yhcote: Will do !

pkg/signing/signing.go
yhcote: Right !

pkg/signing/signing.go
yhcote: Maybe, will try it !

pkg/signing/signing.go
yhcote: Sure, I assume there's a lot more do to than that to be honest.

pkg/signing/signing.go
bauerm97: @tri-adam Quick question just for my own reference, why is using a `defer ...` statement generally preferred? Is this something that Go just says is best practice or is it to avoid mistakes?

tri-adam: It just makes for easier maintenance, since the statement is grouped with the thing that it is related to, you don't have to hunt for it, and if there are early `return`s, you're not at risk of forgetting to `Close` something.

For a more detailed and eloquent explanation, see Effective Go: https://golang.org/doc/effective_go.html#defer

pkg/signing/signing.go
tri-adam: Sounds good, feel free to do that in a future PR if it makes more sense.

bauerm97:
@yhcote Any status update on this PR? Would be great to get into the tree

pkg/signing/signing.go
yhcote: This cannot be done there, and got me debugging for a minute there. Closing the file at this point flushes the write buffers and allows the sending of a proper buffer to the server. With deferring the close of "wr", the buffer sent was incomplete and caused errors.

Issue #1185, opened 2017-11-30
Label string too long on instance.start cmd


ArangoGutierrez:
## Version of Singularity:

2.4-master.g718360bb

When doing the  [example](https://github.com/singularityware/singularityware.github.io/blob/master/pages/docs/user-docs/docs-instance-start.md)

### Expected behavior
```bash
$ singularity instance.start shub://GodloveD/lolcow cow1
```
### Actual behavior

```bash
$ singularity instance.start shub://GodloveD/lolcow cow1
ERROR  : Label string too long
ABORT  : Retval = 255
ERROR  : Daemon failed to start
ABORT  : Retval = -1
```

Debug + VERBOSE output attached:
[instance_start_issue.txt](https://github.com/singularityware/singularity/files/1519511/instance_start_issue.txt)

GodloveD:
I can't seem to replicate this.  Can you see if it is still an issue with the latest release-2.4 branch?

ArangoGutierrez:
Works! on latest release-2.4 branch thanks @GodloveD

GodloveD:
Great!  😸

ArangoGutierrez:
Error is back on 
```bash
$ singularity --version
2.4.1-release-2.4.g56581cf3
```

GodloveD:
Crap!  💩

yarikoptic:
indeed it is so even in 2018
```shell
$> singularity instance.start shub://GodloveD/lolcow cow1
Progress |===================================| 100.0% 
ERROR  : Label string too long
ABORT  : Retval = 255
ERROR  : Daemon failed to start
ABORT  : Retval = -1

$> singularity --version
2.4.5-dist
```

yarikoptic:
and even with 2.4.6 (just uploaded to Debian and NeuroDebian)

jmstover:
Hi @yarikoptic .  Have a PR in which should fix this issue. At least locally, I'm not hitting it with what I've tested.

jscook2345:
@ArangoGutierrez @yarikoptic 

This looks to be fixed in 3.0.0. Please re-open if the issue persists and we'll work from there. Thanks!

yarikoptic:
Thanks!

PR #1461, opened 2018-04-16
Use PATH_MAX for filename size in struct

jmstover:
Use `PATH_MAX` for filename size in struct

Also, do not use a hard coded `/tmp`. Use `SINGULARITY_TMPDIR` if set,
and fall back to `/tmp` if not.

- Ref: #1185 

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/src/util/util.c
+++ b/src/util/util.c
@@ -379,6 +379,12 @@ void free_tempfile(struct tempfile *tf) {
 struct tempfile *make_tempfile(void) {
    int fd;
    struct tempfile *tf;
+   char *tmpdir = singularity_registry_get("TMPDIR");
+
+   printf("TMPDIR is: %s\n", tmpdir);
+   if (tmpdir == NULL) {
+       tmpdir = "/tmp";
+   }
 
    tf = malloc(sizeof(struct tempfile));
    if (tf == NULL) {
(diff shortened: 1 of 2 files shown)


yarikoptic:
Great! Thank you @jmstover . FWIW For my limited C knowledge looks kosher, and it does resolve the issue for me.

src/util/util.c
```diff
@@ -379,14 +379,19 @@ void free_tempfile(struct tempfile *tf) {
 struct tempfile *make_tempfile(void) {
    int fd;
    struct tempfile *tf;
+   char *tmpdir = singularity_registry_get("TMPDIR");
+
+   if (tmpdir == NULL) {
+       tmpdir = "/tmp";
```
yarikoptic, line 385: Re ` hard coded /tmp` -- shouldn't regular, not the prefixed one from singularity_registry_get, TMPDIR env be consulted for that ideally here?

jmstover: I tried to use just `TMPDIR`. I'm thinking it's getting lost somewhere as doing `getenv("TMPDIR");` returned NULL.

So, I went with the next best thing. Haven't had a chance for one of the guys who knows the code better than I to look over it yet, but if they have an idea on how to use just the `TMPDIR` value, I'd prefer that over using `SINGULARITY_TMPDIR`.

yarikoptic: yeah, and here anyways `/tmp` was hardcoded already, so this fix doesn't have to fix that

GodloveD: My preference would be the following order of operations:
1.  Check if `SINGULARITY_TMPDIR != NULL` and use it if possible
2.  Check of `TMPDIR != NULL` and use it if possible
3.  Failing both of these, fall back to use `/tmp`

But I'm not sure if we have easy access to `TMPDIR` in these functions, and I don't think it merits substantial changes to get access to it.

GodloveD (approved):

ArangoGutierrez (approved):

PR #1462, opened 2018-04-16
Constrain docker_provision Untar to provided dst folder (3.0)

dtrudg:
In `docker_provision.go` we use the simple `Untar` function to extract the Singularity `environment.tar` into the container sandbox. Although we control the content of this tar it is good practice to constrain extraction with `Untar` to ensure files/folders extracted are within the provided `dst` destination directory.


bauerm97 (approved):
This looks great.

PR #1473, opened 2018-04-18
Fix Definition Reflection

tri-adam:
Libraries that use reflection aren't able to access all members of the `Definition` structure, since some of the structs inside it are not public. Make all members public.


dtrudg:
:+1: This is definitely necessary to avoid unnecessarily duplicating structs in services like Builder, Library (where we may also possibly have a use case to store a definition).

dtrudg (approved):

PR #1474, opened 2018-04-19
Fix compilation and execution on Pi aarch64

dtrudg:

Quick fixes for some architecture detection on compilation and execution so that the 3.0 tree is able to run on a RaspberryPi 3B with Pi64 linux install. This can now work...

```
singularity build pi_alpine.sif docker://aarch64/alpine
singularity -d exec pi_alpine.sif cat /etc/lsb-release
```

### DIFF
--- a/core/lib/sif/sif.c
+++ b/core/lib/sif/sif.c
@@ -111,7 +111,8 @@ sif_validate(Sifinfo *info)
 		currarch = SIF_ARCH_386;
 	else if(!strncmp(name.machine, "arm", 3) && sizeof(void *) == 4)
 		currarch = SIF_ARCH_ARM;
-	else if(!strncmp(name.machine, "arm", 3) && sizeof(void *) == 8)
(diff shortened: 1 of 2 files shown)


core/lib/sif/sif.c
```diff
@@ -111,7 +111,8 @@ sif_validate(Sifinfo *info)
 		currarch = SIF_ARCH_386;
 	else if(!strncmp(name.machine, "arm", 3) && sizeof(void *) == 4)
 		currarch = SIF_ARCH_ARM;
-	else if(!strncmp(name.machine, "arm", 3) && sizeof(void *) == 8)
+	else if( (!strncmp(name.machine, "arm", 3) && sizeof(void *) == 8) ||
```
yhcote: there's a extra space after if( that doesn't match the rest of the if statement.

I am not sure if we want all machines being converted to AARCH64. It seems like it's a different case.

yhcote (changes requested):

yhcote (approved):
Looks good to me !

PR #1477, opened 2018-04-20
Push and Pull PGP Material

yhcote:
This PR fixes code review best practices introduced in the previous version of this Pull Request.

### DIFF
--- a/pkg/signing/signing.go
+++ b/pkg/signing/signing.go
@@ -8,11 +8,19 @@ import (
 	"fmt"
 	"github.com/singularityware/singularity/pkg/image"
 	"golang.org/x/crypto/openpgp"
+	"golang.org/x/crypto/openpgp/armor"
 	"golang.org/x/crypto/openpgp/clearsign"
 	"golang.org/x/crypto/openpgp/packet"
+	"io/ioutil"
 	"log"
+	"net/http"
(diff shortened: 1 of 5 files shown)


pkg/signing/signing.go
```diff
@@ -96,10 +103,130 @@ func printSignatures(entity *openpgp.Entity) error {
...
+		log.Println("Error while trying to read public key ring: ", err)
+		return nil, err
+	}
+
+	return el, nil
+}
+
+func printEntity(index int, e *openpgp.Entity) {
+	for _, v := range e.Identities {
+		fmt.Printf("%v) U: %v %v %v\n", index, v.UserId.Name, v.UserId.Comment, v.UserId.Email)
```
bauerm97, line 186: I think we should avoid using any `fmt.*` functions and stick to `log.*` type functions

bauerm97 (changes requested):

pkg/signing/signing.go
yhcote: This is a "*printEntity*" debug function.

bauerm97: OK, I'll let it slide this time ;)

bauerm97 (approved):

PR #1484, opened 2018-04-25
Combined patch enhancing build system and overall structure

yhcote:


bauerm97 (approved):
Looks good and builds correctly for me. Let's merge this now and then submit a new patch to fix some of the minor things suggested by @tri-adam

src/cmd/singularity/actions.go
```diff
@@ -14,8 +14,8 @@ import (
 	"strings"
 
 	"github.com/opencontainers/runtime-spec/specs-go"
-	config "github.com/singularityware/singularity/internal/pkg/runtime/engine/singularity/config"
-	"github.com/singularityware/singularity/pkg/configs"
+	runtimeconfig "github.com/singularityware/singularity/src/runtime/workflows/workflows/singularity/config"
```
dtrudg: Nested `worklfows/workflows` is going to get confusing - is there any other sensible naming here?

src/cmd/singularity/actions.go
```diff
@@ -14,8 +14,8 @@ import (
...
+	"github.com/singularityware/singularity/src/pkg/buildcfg"
```
dtrudg: Would prefer consistent use of config or cfg c.f. this line with the above

src/pkg/sif/sifcore.go
```diff
@@ -5,9 +5,9 @@
   consult LICENSE file distributed with the sources of this project regarding
   your rights to use or distribute this software.
 */
-package image
+package sif
 
-// #cgo LDFLAGS: -lsycore -luuid
+// #cgo LDFLAGS: -lruntime -luuid
```
dtrudg, line 10: Just wondering if there is any chance of bumping into some other lib called 'runtime' in a library path?

dtrudg (commented):
Don't feel able to give a thumbs up or down on this.

Builds okay for me by running `./mconfig` - but how do I build or run an image?

`builddir/singularity` doesn't have any sub-commands available, and the new `sbuild` is asking for builder type, definition json etc. not a def file.

Have a few concerns about naming consistency which may be minor, or may be things that get annoying when things go along - but mainly, need some guidance here on what is the expected functionality that should be working, and how to test the binaries built.

tri-adam:
I'm getting a warning from `dep` on this branch:

```
Warning: the following project(s) have [[constraint]] stanzas in Gopkg.toml:

  ✗  github.com/syndtr/gocapability

However, these projects are not direct dependencies of the current project:
they are not imported in any .go files, nor are they in the 'required' list in
Gopkg.toml. Dep only applies [[constraint]] rules to direct dependencies, so
these rules will have no effect.

Either import/require packages from these projects so that they become direct
dependencies, or convert each [[constraint]] to an [[override]] to enforce rules
on these projects, if they happen to be transitive dependencies,
```

I think that constraint just needs to be removed from `Gopkg.toml`.

bauerm97 (changes requested):
Updating my review since we need to make a few changes as pointed out by @dctrud. I failed to do my due diligence when reviewing; I make installed this branch into a non-pathed location and ended up doing my testing from a previous installation.

bauerm97 (approved):

bauerm97:
I'm approving the updated changes and merging. Thanks @yhcote

PR #1486, opened 2018-04-26
fixed * expansion during app runscript creation

bleuchien:
Hi,

if an apprun script contains a * it will get expanded during the creation of the app runscript for the image build process. In particular * will be replaced with a listing of the directory content from which the singularity build process is started.

Obviously this results in an unexpected behavior when the app script is run.

A simple example definition to replicate the issue:

Bootstrap: docker
From: fedora:26

%apprun test
  ls *

singularity build test.simg test.def
singularity shell test.simg
> cat /scif/apps/test/scif/runscript
ls Desktop Documents Downloads Music Pictures Public Templates Videos

Expected behavior would be that the content of the runscript is the same as in the definition file.

The simple solution is to use printf instead of echo in the get_section function.

Cheers
Richard

**Checkoff for all PRs:**

Attn: @singularityware-admin


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ and **Merged pull requests**. All closed issued and bug fixes should be
 represented by the pull requests that fixed them. This log originated with Singularity 2.4
 and changes prior to that are (unfortunately) done retrospectively. Critical items to know are:
 
+ - fixed * expansion during app runscript creation
```
bauerm97: I think this shouldn't go here. This is intended to be a list of classes of changes to Singularity that warrant an entry into the changelog. This is not the place to list out changes!

bauerm97 (changes requested):
So in general, I really like this change. I think it makes a lot of sense that wildcards in scripts get propagated into the container as wildcards to be expanded when run (as opposed to at build time). However, I worry that changing this behavior during `2.x` might lead to some people having problems who relied on the old behavior. Perhaps this is a change better suited for the Singularity 3.0 release? We could work with you to help port those changes into the 3.0 code if that's the route we decide to go.

@GodloveD @vsoch Can either of you two think of some examples where this behavior is relied on when building an image?

vsoch:
I think this is a bug and it should be fixed! I can't imagine that apps are common enough, and within that small subset a user that relies on a hard coded ls during run time. I also don't think it hurts to add to the changelog if there is some tiny chance that functionality is changed that is important.  These parts:

> All closed issued and bug fixes should be represented by the pull requests that fixed them

and 

> changed behaviour (recipe sections work differently)

suggest that it should be included, but also with a reference to the PR.

bauerm97:
@vsoch I agree that it should be included in a Changelog, however the section that it was added to (as of present) is not where changes are supposed to be placed. The section it currently in is just describing *which types of changes* should be included. This should in the changelog, since it's a specific change rather than a class of changes.

vsoch:
oh, hehe, yes I didn't look closely! derpader, dinosaur!

+1 that's wrong! @bleuchien please re-organize the location of the changelog line and add a nice line to link to this PR :) Thank you!

bleuchien:
I hope my updates to the changelog fit your requirements. Cheers!

CHANGELOG.md
GodloveD: I respectfully disagree @bauerm97.  The last several releases have had bug fix sections in the changelog.  I believe this is a perfect candidate for that.

GodloveD:
Oh Sorry @bauerm97 I also assumed you meant that this change should not be included in the CHANGELOG.md.  Glad we are all in agreement that it should go in.

CHANGELOG.md
```diff
@@ -6,12 +6,18 @@ and **Merged pull requests**. All closed issued and bug fixes should be
...
```
GodloveD: Can you also delete this line please @bleuchien ?

CONTRIBUTORS.md
```diff
@@ -44,3 +44,4 @@
     - Tarcisio Fedrizzi <tarcisio.fedrizzi@gmail.com>
     - Thomas Hamel <hmlth@t-hamel.fr>
     - Yaroslav Halchenko <debian@onerussian.com>
+    - Richard Neuboeck <hawk@tbi.univie.ac.at>
```
GodloveD: I think we are doing alphabetical (by first name) if that is cool.

libexec/bootstrap-scripts/functions
```diff
@@ -101,7 +101,10 @@ get_section() {
             else
                 # We already found it, still in section
                 if [[ "$FOUND" == "yes" ]]; then
-                    echo $line
+                    if [ ! -z "${line}" ]; then
```
GodloveD: I have some questions about this line.  

First, isn't a `! -z` just a `-n`?  Should probably use that instead.

But second, I question why this line is being added at all.  This is going to strip empty lines right?  Don't we want to keep them.  In fact there was a recent PR to ensure that we keep them.  https://github.com/bleuchien/singularity/blame/df5e815f812c1260f59b694c1771c1efbc261d98/libexec/bootstrap-scripts/functions#L78

GodloveD (changes requested):
Sorry for the delay @bleuchien.   I would suggest some more changes.

libexec/bootstrap-scripts/functions
bleuchien: Sorry for the slow response:
- I removed the additional line from the changelog
- rearranged the contributors file entry
- and removed the empty line filter as requested

CHANGELOG.md
```diff
@@ -12,6 +12,11 @@ and changes prior to that are (unfortunately) done retrospectively. Critical ite
  - migration guidance (how to convert images?)
  - changed behaviour (recipe sections work differently)
 
+## [current master](https://github.com/bleuchien/singularity)
```
bauerm97: Hmmm, this doesn't seem right @bleuchien. I think it should maybe be `[current master](https://github.com/singularityware/singularity)` instead ;)

bauerm97 (changes requested):

CHANGELOG.md
bleuchien: Sorry! Copy and paste error. It's fixed.

GodloveD:
Merged in the form of #1683.  Thanks @bleuchien !

Issue #1447, opened 2018-04-08
fd_cleanup in branch development-2.x  contains a bug


eriksjolund:
## Version of Singularity:

git branch
development-2.x

The code I am referring to has not yet been released in any official release, but lives in the file
src/util/util.c in the git branch development-2.x.  In for instance this commit:

https://github.com/singularityware/singularity/blob/4c386b8273eddb104d1533f9036be1b093f86990/src/util/util.c

The if statment 

    if ( is_dir(fd_path) < 0 || is_sock(fd_path) < 0 )

is always true.

I would guess, this what is meant

    if ( is_dir(fd_path) < 0 && is_sock(fd_path) < 0 )

When tried to fix that bug, I also got curious if not the whole function could be rewritten:
I tried out two approaches:

Only iterating over the entries in /proc/self/fd:
https://github.com/eriksjolund/singularity/commits/optimize_fd_cleanup/src/util/util.c

Iterating over all the file descriptors up til sysconf(_SC_OPEN_MAX)
https://github.com/eriksjolund/singularity/commits/simplify_fd_cleanup/src/util/util.c

dtrudg:
Hi @cclerget - would you be able to take a look at this?

cclerget:
Hi @eriksjolund, @dctrud. Yes, I wrote a fix and modified it based on your suggestions @eriksjolund. Thank you for your help @eriksjolund !

jthiltges:
Is your fix available on github, @cclerget? I'd prefer the opendir() version above, rather than closing all possible fds.

In docker 1.12.6 (which ships with the CentOS 7 extras), the default limit for open files is 1048576. Calling sysconf(), stat(), and close() for each fd causes a significant delay in singularity 2.4.6 startup. A simple echo test inside a container took about 13 seconds to run, vs <0.1s after lowering the open file limit.

Regards,
John

```console
# Testing with the default limit of 1M open files
$ ulimit -n
1048576
$ time /usr/bin/singularity exec /cvmfs/singularity.opensciencegrid.org/opensciencegrid/osgvo-el6:latest echo "Hello World"
Hello World

real    0m13.618s
user    0m1.853s
sys     0m11.720s

# Retesting after lowering the limit
$ ulimit -n 1024
$ time /usr/bin/singularity exec /cvmfs/singularity.opensciencegrid.org/opensciencegrid/osgvo-el6:latest echo "Hello World"
Hello World

real    0m0.069s
user    0m0.021s
[...]

PR #1490, opened 2018-04-28
Call fd_cleanup after container setup

cclerget:

fd_cleanup closes socket early in code and can break identity services like SSSD.
This fix call fd_cleanup later after all functions finished to request user identity.

fd_cleanup take a callback function as argument to call a custom function which return 1 if file descriptor must be closed and 0 if not. With callback function src/start.c can call fd_cleanup and avoid long loop like reported by user here since actual code use sysconf to get max number of file descriptor : https://github.com/singularityware/singularity/issues/1447#issuecomment-380832392

- Ref: #1489 #1447 

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -49,14 +49,18 @@
 #error SYSCONFDIR not defined
 #endif
 
+int close_fd(int fd, struct stat *st) {
+    if ( S_ISDIR(st->st_mode) || S_ISSOCK(st->st_mode) ) {
+        return(1);
+    }
+    return(0);
+}
 
 int main(int argc, char **argv) {
     struct image_object image;
     char *pwd = get_current_dir_name();
     char *target_pwd = NULL;
     char *command = NULL;
-
-    fd_cleanup();
 
     singularity_config_init(joinpath(SYSCONFDIR, "/singularity/singularity.conf"));
 
@@ -101,7 +105,6 @@ int main(int argc, char **argv) {
     singularity_runtime_enter();
     
     singularity_runtime_environment();
-    
     singularity_priv_drop_perm();
 
     if ( ( target_pwd = singularity_registry_get("TARGET_PWD") ) != NULL ) {
@@ -144,6 +147,8 @@ int main(int argc, char **argv) {
     envar_set("SINGULARITY_NAME", singularity_image_name(&image), 1);
     envar_set("SINGULARITY_SHELL", singularity_registry_get("SHELL"), 1);
     envar_set("SINGULARITY_APPNAME", singularity_registry_get("APPNAME"), 1);
+
+    fd_cleanup(&close_fd);
 
     singularity_message(LOG, "USER=%s, IMAGE='%s', COMMAND='%s'\n", singularity_priv_getuser(), singularity_image_name(&image), singularity_registry_get("COMMAND"));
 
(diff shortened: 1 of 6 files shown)


src/action.c
```diff
@@ -49,15 +49,19 @@
 #error SYSCONFDIR not defined
 #endif
 
+int close_fd(int fd, struct stat *st) {
+    if ( S_ISDIR(st->st_mode) || S_ISSOCK(st->st_mode) ) {
```
jmstover, line 53: Is this right?

The open directory was because that was the `ls` run, so it was listing that directory...  `sssd` also had links to a `passwd` and `group` file, besides the socket. Which I don't know if they're required. I know nothing of `sssd` :/
```
lr-x------ 1 preney preney 64 Apr 27 21:23 3 -> /ram/var/lib/sss/mc/passwd
lr-x------ 1 preney preney 64 Apr 27 21:23 5 -> /ram/var/lib/sss/mc/group
```

cclerget: Discussed on slack, let comment for others. `close_fd` return `1` to close file descriptor, `0` otherwise. Close directory and socket was done before this fix, I just took same code wrapped in `close_fd` for `src/action.c`

I have no idea why sssd open those files with each user processes. I tested on CentOS 6, first run, a socket appears in opened file descriptors, next runs socket wasn't here, probably requesting information and at same time refresh the cache ...

jmstover (approved):
This looks good here...

tri-adam (approved):
Nice and clean!

GodloveD:
Thanks everybody.

markhahn:
I'm assisting Paul Preney on this, and with the current release-2.5 branch, still get:

> DEBUG   [U=3000566,P=15369] singularity_runtime_mount_home()         Checking that home directry is configured: /
> DEBUG   [U=3000566,P=15369] singularity_runtime_mount_home()         Checking if home directories are being influenced by user
> DEBUG   [U=3000566,P=15369] singularity_registry_get()               Returning NULL on 'HOME'
> DEBUG   [U=3000566,P=15369] singularity_runtime_mount_home()         Checking ownership of home directory source: /
> ERROR   [U=3000566,P=15369] singularity_runtime_mount_home()         Home directory is not owned by calling user: /
> ABORT   [U=3000566,P=15369] singularity_runtime_mount_home()         Retval = 255

One oddity of this system is that it's a cluster built on OneSIS, which uses a readonly NFS root for all nodes, and has a kind of funky mechanism for customizing files (by node identity or role, etc).  Specifically, things get redirected through /ram, so although sssd (libnss_sss.so really) opens fds to /var/lib/sss/mc/passwd, under OneSIS they look like /ram/var/lib/sss/mc/passwd.  I'm speculating that something in Singularity's startup chokes on this.  (Permissions on the path look OK to me, adding a "-B /ram" didn't help (seemed to simple not to try)).

I tested this on a standalone system that has the same sssd config but no OneSIS, and Singularity appears to work as expected.  Admittedly, there's a lot of other differences between that host and the cluster...

preney:
Per PR 1489, we tried the 2.5.1 release and everything appears to work now. Many thanks!

jmstover:
Hi @preney 

Thanks for the heads up. Glad that it's working for you now.

PR #1494, opened 2018-04-30
Development 3.0: wrapper use static library instead of calling executables

cclerget:

There is a problem to join an instance with current C code which call Go executables smaster/scontainer, once entered in instance mount namespace, binaries are unreachable. The solution is to wrap Go code into a static library libstartup and link it with C wrapper.

Fixes :
- Remove wrapper.mconf and compile it from Makefile.stub due to mutual dependencies between C and Go code
- go vet issues reported by Adam
- File cleanup for startup code

**Don't merge yet**


yhcote (approved):
looks good to me

PR #1496, opened 2018-04-30
allow --home to still work with "mount home = no"

DrDaveD:

This allows the --home command line option to still work even when "mount home = no" is set in singularity.conf.  The comments in singularity.conf don't say anything about this option disabling setting the home directory on the command line, so although this changes functionality, it fixes a bug in the currently documented behavior.

This affects all versions of singularity that I know about, but it was particularly noticed with 2.4.6 because before that some of my users had specified both --bind and --home for the same mapping.  With 2.4.6 the duplication was rejected, so the users removed the --bind, and then ran into one site that had set mount home = no, where the --home was silently ignored.

- None

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/src/lib/runtime/mounts/home/home.c
+++ b/src/lib/runtime/mounts/home/home.c
@@ -50,11 +50,6 @@ int _singularity_runtime_mount_home(void) {
     char *container_dir = CONTAINER_FINALDIR;
 
 
-    if ( singularity_config_get_bool(MOUNT_HOME) <= 0 ) {
-        singularity_message(VERBOSE, "Skipping home dir mounting (per config)\n");
-        return(0);
-    }
-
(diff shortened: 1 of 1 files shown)


DrDaveD:
I rebased on the current developement-2.x.

GodloveD (approved):
@dctrud, @jmstover, @gmkurtzer and I had a convo about this some time ago to determine whether or not this is something we want to do.  We determined with @gmkurtzer's input that this was indeed his original intention. Therefore, I propose that we merge.

GodloveD:
Oh, also, I tested this and it works and the code seems sensible to me as well.

tri-adam (approved):

GodloveD:
Thanks @DrDaveD !

PR #1507, opened 2018-05-04
Add tagging and other support for singularity pull

dtrudg:

This PR adds CLI support for tagging in the Container library.

You can push an image with:
- no tag (will default to tagging with 'latest'
    `singularity push myimage.sif library://dave/test/alpine`
- a single tag
   `singularity push myimage.sif library://dave/test/alpine:demo1`
- multiple tags
  `singularity push myimage.sif library://dave/test/alpine:demo1,latest,meoww`

Images can be pulled with a named tag, or their id (currently in `sha256.<sha256>` format, but native `sif.<uuid>` to be added later)

Also:
- now uses sylog for log messages
- pull won't overwrite existing image file unless `--force/-F` specified
- can pull without specifying an image name - it'll use a default `<container>_<tag>.sif` format

### DIFF
--- a/src/cmd/singularity/cli/pull.go
+++ b/src/cmd/singularity/cli/pull.go
@@ -23,9 +23,13 @@ func init() {
 }
 
 var pullCmd = &cobra.Command{
-	Use:  "pull myimage.sif library://user/collection/container:tag",
-	Args: cobra.ExactArgs(2),
+	Use:  "pull [myimage.sif] library://user/collection/container[:tag]",
+	Args: cobra.RangeArgs(1, 2),
 	Run: func(cmd *cobra.Command, args []string) {
-		libexec.PullImage(args[0], args[1], PullLibraryURI)
+		if len(args) == 2 {
+			libexec.PullImage(args[0], args[1], PullLibraryURI)
+			return
+		}
+		libexec.PullImage("", args[0], PullLibraryURI)
 	},
 }
--- a/src/cmd/singularity/cli/pull.go
+++ b/src/cmd/singularity/cli/pull.go
@@ -18,18 +18,19 @@ var (
 
 func init() {
 	pullCmd.Flags().StringVar(&PullLibraryURI, "libraryuri", "http://localhost:5150", "")
+	pullCmd.Flags().BoolVarP(&Force, "force", "F", false, "overwrite an image file if it exists")
 	singularityCmd.AddCommand(pullCmd)
 
 }
 
 var pullCmd = &cobra.Command{
-	Use:  "pull [myimage.sif] library://user/collection/container[:tag]",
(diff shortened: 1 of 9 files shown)


src/cmd/singularity/cli/pull.go
```diff
@@ -13,19 +13,25 @@ import (
...
 }
 
 var pullCmd = &cobra.Command{
-	Use:  "pull myimage.sif library://user/collection/container:tag",
-	Args: cobra.ExactArgs(2),
+	Use:  "pull [options] [myimage.sif] library://user/collection/container[:tag]",
+	Args: cobra.RangeArgs(1, 2),
 	Run: func(cmd *cobra.Command, args []string) {
-		libexec.PullImage(args[0], args[1], PullLibraryURI)
+		if len(args) == 2 {
```
GodloveD: Just a comment.  In general I think that we should push the runscripts elsewhere so that the CLI remains the CLI and they just call out to another function.  But there are several other instances of this that also need to be worked on.  I won't block b/c of it.  Just a comment.

src/pkg/library/client/push.go
```diff
@@ -130,16 +152,42 @@ func createContainer(name string, collectionID string) (id string, err error) {
...
-		Name:        name,
+		Hash:        hash,
 		Description: "No description",
 		Container:   bson.ObjectIdHex(containerID),
 	}
 
 	return apiCreate(i, baseURL+"/v1/images")
 }
 
+func setTags(containerID string, imageID string, tags []string) error {
```
GodloveD, line 165: Don't you need some documentation above functions or some such thing?

GodloveD (dismissed):
If you want a thorough review on this it will probably take me several days.  If you want this in quickly it may be best to get another reviewer I think.  I will approve this because I trust you and I know the code is necessarily rough and ready right now.  Up to you if you want to get a second review or if you need this in to keep making progress.

src/pkg/library/client/push.go
dtrudg: The general rule (applied by go lint at least) is that public functions (starting with capitals) need doc comments, and private functions like this don't. But that wouldn't preclude me adding some more here.

dtrudg:
@GodloveD - many thanks for this.  Hopefully @bauerm97 or @tri-adam can take a quick look through too, as it's a fair chunk.

Noting for this PR, also had useful discussion with Dave r.e. the library:// URI structure, and thoughts on naming for entities/collections etc. I will open an issue to revisit this and ensure that things are consistent in any user-exposed messages / interface.

tri-adam (dismissed):
Looks good to me!

dtrudg:
Hi @tri-adam @GodloveD  - merging development-3.0 back in dismissed your reviews. Are you still happy to merge this? The only change from the merge was on gopkg.lock - thanks.

tri-adam (approved):

PR #1522, opened 2018-05-09
Fix cgo path when building outside of makeit.

yhcote:
This will allow tools such as "go vet -all ./..., go fmt ./..., golint ./..." to run an any other tools that are not initiated via the 'makeit' build system.

Caveat but default, the buildtree path needs to be "./builddir" as
generated by makeit defaults for this to work.

### DIFF
--- a/src/pkg/sif/sifcore.go
+++ b/src/pkg/sif/sifcore.go
@@ -7,7 +7,8 @@
 */
 package sif
 
-// #cgo LDFLAGS: -lruntime -luuid
+// #cgo CFLAGS: -I../../runtime/c/lib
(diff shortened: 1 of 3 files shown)


GodloveD (approved):
I can verify that this works as expected.  Thanks @yhcote !

bauerm97 (approved):
I think that hard coding `builddir` as the build directory is okay, since:

1. `.travis.yml` uses default mconfig values
2. `.gitignore` has `builddir` hardcoded
3. The Makefile will still export the `CGO_LDFLAGS` & `CGO_CPPFLAGS` variables for people using a different build directory that just wish to build the source code in a custom location 

👍

PR #1528, opened 2018-05-10
Check GID in 45-targetmode.sh

jmstover:

Check GID in `tests/45-targetmode.sh` to see if we have a GID >= 1000.  Skip test if we have a primary GID < 1000.

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/tests/45-targetmode.sh
+++ b/tests/45-targetmode.sh
@@ -23,6 +23,11 @@
 
 . ./functions
 
+if [ `id -g` -lt 1000 ]; then
+    echo -e "\nGID less than 1000 ... skipping 45-targetmode\n"
+    exit 0
+fi
+
 test_init "Checking target UID/GID mode"
 
 


GodloveD:
Just for the record this allows tests to pass on SLES where the default gid for a new user is like 500 or something.

GodloveD (approved):
This works.

dtrudg (approved):
This works  for me when I setup <1000 - but before merge could you comment on exactly what we are losing in the test that don't run?

jmstover:
The test does something like:

```
SINULARITY_TARGET_GID=99 SINGULARITY_TARGET_UID=99 singularity exec $CONTAINER whoami
```
If those are set, then when the user information is set in `src/util/privlege.c`  it uses the value requested. This is used on some schedulers where jobs are submitted as one user, but run as a different user. The tests just verify that that the target_gid/uid is what is seen when the container runs.

dtrudg:
@jmstover - thanks, I'd read it as that but hadn't seen it used in the wild so wanted confirmation. Think this looks good for you to merge.

PR #1531, opened 2018-05-10
Revert "Fix mount invalid argument and restrictive directory binds"

GodloveD:
Reverts singularityware/singularity#1525


GodloveD:
Well that was a mistake.

PR #1538, opened 2018-05-11
Fix Centos7 travis build

DrDaveD:

Centos 7.5 has been released and as of now there are no source repositories available.  This PR skips the yum-builddep step which apparently needs the sources but is itself apparently unneeded.

- Ref: #

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/.travis/rpmbuild_test
+++ b/.travis/rpmbuild_test
@@ -7,9 +7,6 @@ OS_VERSION="$1"
 # build and install
 yum install -y libtool rpm-build make yum-utils squashfs-tools libarchive-devel
 ./autogen.sh
-#yum-builddep failing on 11 May 2018 because Centos 7.5
-#   Source/repodata/repomd.xml is missing, skip for now.
-#yum-builddep -y singularity.spec
 ./configure
 make dist
 rpmbuild -ta *.tar.gz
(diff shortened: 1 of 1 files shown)


GodloveD (commented):
Good catch @DrDaveD.  Can you further clarify just a bit more?  Is this a temporary measure?  Thanks.

DrDaveD:
I believe that the sources will show up eventually.  On the other hand, they don't appear to be needed.  I thought it better to leave the comment in for now in case something comes up later that does need it and the sources are there.   If you think it's better to just delete it with no comment, that's fine.

DrDaveD:
The 7.5 sources have reappeared, so travis builds that do not include this PR aren't failing anymore.  However I noticed one of them give the following non-fatal error messages:
```
error: Unable to open singularity.spec: No such file or directory
Bad spec: singularity.spec
No uninstalled build requires
```
That's because singularity.spec is not yet created by the time yum-builddep is run, because it is before configure.  It has probably always done this, making it clear that yum-builddep is unnecessary.  I have now deleted the commented out lines.  Please go ahead with this PR.

GodloveD (approved):
Looks good to me.  I'm also going to ask @dctrud to look.  @dctrud was this one of the lines that was added when I was giving you confusing (and likely wrong) guidance about testing?

dtrudg:
@GodloveD - I believe the `builddep` was brought in when I was doing the docker-extract stuff at your suggestion to adopt some of the ideas from #1204 - I'm okay with removing it so the tests work in this case. Maybe we want to consider asking those involved in #1204 about it?

dtrudg (approved):

PR #1539, opened 2018-05-13
Pgp cleanup

yhcote:
This PR takes the initial code from signing.go and separates its into 2 logical units. sypgp.go and signing.go. sypgp.go is a new package that provides a library of functions to manage PGP keys locally and allows connectivity with an upstream server.


src/pkg/sypgp/sypgp.go
```diff
@@ -0,0 +1,424 @@
...
+	for _, v := range e.Identities {
+		fmt.Printf("%v) U: %v %v %v\n", index, v.UserId.Name, v.UserId.Comment, v.UserId.Email)
+	}
+	fmt.Printf("   C: %v\n", e.PrimaryKey.CreationTime)
+	fmt.Printf("   F: %0X\n", e.PrimaryKey.Fingerprint)
+	bits, _ := e.PrimaryKey.BitLength()
+	fmt.Printf("   L: %v\n", bits)
+}
+
+func printPubKeyring() (err error) {
```
tri-adam, line 199: Since you've got a named return value (and this is a short `func`), you could do naked returns in this function.

src/pkg/sypgp/sypgp.go
```diff
@@ -0,0 +1,424 @@
...
+
+	for i, e := range pubEntlist {
+		printEntity(i, e)
+		fmt.Println("   --------")
+	}
+
+	return nil
+}
+
+func printPrivKeyring() (err error) {
```
tri-adam: Same here, you could do naked returns in this function.

src/pkg/sypgp/sypgp.go
```diff
@@ -0,0 +1,424 @@
...
+	if err := PathsCheck(); err != nil {
+		return err
+	}
+
+	fmt.Print("Enter your name (e.g., John Doe) : ")
+	scanner := bufio.NewScanner(os.Stdin)
+	scanner.Scan()
+	name := scanner.Text()
+	if err := scanner.Err(); err != nil {
+		log.Println("Error while reading name from user: ", err)
```
tri-adam: Should switch the `log.XXX` funcs to `sylog` while you're in this area.

src/pkg/sypgp/sypgp.go
```diff
@@ -0,0 +1,424 @@
...
+		err = fmt.Errorf("server returned more than one key for unique fingerprint")
+		log.Println(err)
+		return nil, err
+	}
+
+	return el, nil
+}
+
+// PushPubkey pushes a public key to a key server
+func PushPubkey(entity *openpgp.Entity, sykeysAddr string) (err error) {
```
tri-adam, line 386: As above, good candidate for naked returns.

tri-adam (changes requested):
Think we should change `log` -> `sylog`, and maybe go for naked returns on the funcs you've got named return values on.

tri-adam (approved):

Issue #1555, opened 2018-05-17
--debug option not working when pulling docker container


fordste5:
## Version of Singularity:

Singularity 2.5.1 on CentOS 6.9

### Expected behavior

When invoking singularity pull with the debug option, I expect additional debug output. Debug seems to work when pulling from shub. For example:

```
# /usr/bin/singularity -d pull shub://foo/bar
Enabling debugging
Ending argument loop
Singularity version: 2.5.1-dist
Exec'ing: /usr/libexec/singularity/cli/pull.exec
VERBOSE2 SINGULARITY_COMMAND_ASIS found as False
VERBOSE2 SINGULARITY_ROOTFS not defined (None)
VERBOSE2 SINGULARITY_METADATA_FOLDER found as None/.singularity.d
VERBOSE2 SINGULARITY_FIX_PERMS found as False
VERBOSE2 SINGULARITY_COLORIZE not defined (None)
VERBOSE2 SINGULARITY_DISABLE_CACHE found as False
VERBOSE2 SINGULARITY_CACHEDIR found as /root/.singularity
VERBOSE2 REGISTRY not defined (None)
VERBOSE2 NAMESPACE not defined (None)
VERBOSE2 SINGULARITY_DOCKER_ARCHITECTURE found as amd64
VERBOSE2 SINGULARITY_DOCKER_OS found as linux
VERBOSE2 SINGULARITY_ENVIRONMENT found as None/.singularity.d/env/90-environment.sh
VERBOSE2 SINGULARITY_RUNSCRIPT found as None/singularity
VERBOSE2 SINGULARITY_TESTFILE found as None/.singularity.d/test
VERBOSE2 SINGULARITY_DEFFILE found as None/.singularity.d/Singularity
VERBOSE2 SINGULARITY_HELPFILE found as None/.singularity.d/runscript.help
VERBOSE2 SINGULARITY_ENVBASE found as None/.singularity.d/env
VERBOSE2 SINGULARITY_LABELFILE found as None/.singularity.d/labels.json
VERBOSE2 SINGULARITY_INCLUDECMD found as False
VERBOSE2 SINGULARITY_NOHTTPS found as False
VERBOSE2 SINGULARITY_PULLFOLDER found as .
VERBOSE2 SHUB_NAMEBYHASH not defined (None)
VERBOSE2 SHUB_NAMEBYCOMMIT not defined (None)
VERBOSE2 SHUB_CONTAINERNAME not defined (None)
[...]

PR #1556, opened 2018-05-17
Pull --debug fix on docker:// URI

jmstover:

When we `pull` from a `docker://` URI, we aren't passing the debug flag if it was used. Check `SINGULARITY_MESSAGELEVEL`, and if `5` (debug), pass the `-d` flag to the `singularity build`

- Ref: #1555 

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/libexec/cli/pull.exec
+++ b/libexec/cli/pull.exec
@@ -141,7 +141,7 @@ case "$SINGULARITY_CONTAINER" in
             message WARNING "Container will be sized appropriately for Docker image.\n"
         fi
 
-         if [ -f "$SINGULARITY_IMAGE" ]; then
+        if [ -f "$SINGULARITY_IMAGE" ]; then
             if [ -n "${OVERWRITE:-}" ]; then
                 message 2 "Removing existing file\n"
                 rm -f "$SINGULARITY_IMAGE"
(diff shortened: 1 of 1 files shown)


libexec/cli/pull.exec
```diff
@@ -151,8 +151,13 @@ case "$SINGULARITY_CONTAINER" in
             fi
         fi
 
-         if [ -x "${SINGULARITY_bindir}/singularity" ]; then
-             ${SINGULARITY_bindir}/singularity build ${SINGULARITY_IMAGE} ${SINGULARITY_CONTAINER}
+        DBGFLAG=""
+        if [ "x${SINGULARITY_MESSAGELEVEL}" == "x5" ]; then
```
bauerm97, line 155: Why does this say `"x$...` and `"x5"` instead of `"$...` and `"5"`?

jmstover: Because if `SINGULARITY_MESSAGELEVEL` is not set, you _could_ end up like:
```
if [ == 5 ]; then
```
And not have a left hand side. So ... it's common practice to add an `x` in, so the check ends up as:
```
if [ x == x5 ]; then
```

dtrudg (approved):
This looks good to me - will assist in turnaround on issues :+1:

PR #1563, opened 2018-05-19
Send authorization token to library on push / pull

dtrudg:

This PR ensures that a library authentication token is sent to the library on `singularity push` and `singularity pull` operations if a token file exists.

The default location for the token file is `~/.singularity/sylabs-token` or there is a `--tokenfile` option to specify another location.

@tri-adam - I think we likely need to refactor the stuff that's here, and the stuff that's in the remote-builder code for some nice, shared token handling. Would love to discuss thoughts on this. If you are okay with this implementation sitting within the library for now, I can raise an issue to resolve this tech debt? Or if you feel strongly it should be addressed now that's cool too.


yhcote (approved):
looks good to me

tri-adam:
Didn't get a chance to review this yet, but want to say I don't think you should hold back for the token refactor. An issue tagged as tech debt seems like a good idea, and I think we should backlog that so we don't forget about it.

dtrudg:
@tri-adam - thanks and agreed. Would like to get this in then come back for the refactor. The current state is that I have code in `pkg/library` to read a file at `~/.singularity/sylabs-token` or an alternative path give by a `--tokenfile` option.

Issue #1588, opened 2018-05-22
scanDefinitionFile() produces duplicates when parsing def file header


GodloveD:
## Version of Singularity:

development-3.0

### Expected behavior

`scanDefinitionFile()` returns the lines composing the header of a def file without any duplicates.  

### Actual behavior

`scanDefinitionFile()` writes the lines composing the header to the def file buffer [here](https://github.com/singularityware/singularity/blob/development-3.0/src/pkg/build/definition_parser_deffile.go#L74-L75) and again [here](https://github.com/singularityware/singularity/blob/development-3.0/src/pkg/build/definition_parser_deffile.go#L69).  Taking [this example def file](https://github.com/singularityware/singularity/blob/development-3.0/examples/ubuntu/Singularity), this produces the following string for the header:

```
BootStrap: debootstrap
OSVersion: trusty
MirrorURL: http://us.archive.ubuntu.com/ubuntu/


BootStrap: debootstrap
OSVersion: trusty
MirrorURL: http://us.archive.ubuntu.com/ubuntu/
```

### Steps to reproduce behavior

Try printing the contents of `h` in `doHeader()`

GodloveD:
Oh I fixed this long ago.

PR #1589, opened 2018-05-22
stop writing header info to buffer twice

GodloveD:

Fixes a little bug in the way that the header is parsed in `scanDefinitionFile()`.

- fixes: #1588 

**Checkoff for all PRs:**

Attn: @singularityware-admin


tri-adam:
@GodloveD, would you be able to write a unit test that fails before/passes after this change? I see we do have code coverage of the line you are changing, but we must not be doing enough to validate results at some level.

GodloveD:
That's a good idea.  I need to learn how to write tests.

bauerm97:
Hmm I don't really understand how this change can work. If we don't write into `retbuf` then nothing will be in `retbuf`, right?

bauerm97:
@GodloveD Are you suggesting that line 72-74 take care of this and we've been returning 2x of the header this entire time?

```
 			// This line is not a section identifier
 			retbuf.Write(line)
 			retbuf.WriteString("\n")
```

GodloveD:
@bauerm97 exactly.  Did you read #1588?

bauerm97:
@GodloveD No, was looking at this on my phone the first time and didn't notice that there was an issue describing it :) Sorry for the confusion!

GodloveD:
@tri-adam OK, just had a closer look at what we are and aren't testing.  So, plz correct me if I'm wrong but here is what I'm seeing.  

The error is in `scanDefinitionFile()`, but we are actually not testing at that level.  We are testing `ParseDefinitionFile()` which calls `scanDefinitionFile()` and runs the output through `doHeader()`.  The `doHeader()` function basically hides the error from `scanDefinitionFile()`, because it takes the output and enters it as keys and values into the `d.Header[]` map.  When `doHeader()` encounters duplicates, it just enters them 2x and overwrites the initial values in the map.  This covers the duplicate errors from `scanDefinitionFile()`.

Because we are testing at too high a level to catch this, I can't just add to the existing tests.  So do you want me to write a new set of tests for `scanDefinitionFile()` itself or should we just accept the higher level of testing and move on?

bauerm97:
@GodloveD I think that writing a new test that evaluates `scanDefinitionFile()`'s behavior would be prudent

GodloveD:
@tri-adam and @bauerm97 check this out.  Right now the test only looks at the header, but hopefully it is written in such a way that it can be expanded to the other sections as well.

src/pkg/build/definition_parser_deffile_test.go
```diff
@@ -6,12 +6,76 @@
...
+			}
+			defer r.Close()
+
+			s := bufio.NewScanner(r)
+			s.Split(scanDefinitionFile)
+			for s.Scan() && s.Text() == "" && s.Err() == nil {
+			}
+			// debug
+			// fmt.Printf("%v\n", s.Bytes())
+			// fmt.Printf("%v\n", s.Text())
```
tri-adam: Remove these commented out debug lines.

src/pkg/build/definition_parser_deffile_test.go
```diff
@@ -6,12 +6,76 @@
...
+				Header string
+			}
+			var d []DefFileSections
+			if err := json.Unmarshal(b, &d); err != nil {
+				t.Fatal("failed to unmarshal JSON:", err)
+			}
+			// debug
+			// nums := []byte(d[0].Header)
+			// fmt.Printf("%v\n", nums)
+			// fmt.Printf("%s\n", d[0].Header)
```
tri-adam: Same here.

tri-adam (changes requested):
Sorry for the delay getting to this @GodloveD, it looks great. Thanks for adding those tests!  Just two tiny requests...

src/pkg/build/definition_parser_deffile_test.go
GodloveD: Oh bummer.  I was afraid you were going to ask that.  I'll forget how I did the debugging when it comes time to add the other sections!  =^D  

OK.  I can remove.

tri-adam (approved):

PR #1599, opened 2018-05-25
go build tag for sif.h lib

ArangoGutierrez:

Currently when importing 
```go
import "github.com/singularityware/singularity/pkg/build"
```
or 
```go
import "github.com/singularityware/singularity/pkg/image"
```
leaves to the following error at build time
```
# github.com/sylabs/remote-build/vendor/github.com/singularityware/singularity/pkg/image
vendor/github.com/singularityware/singularity/pkg/image/sifcore.go:20:10: fatal error: 'sif/list.h' file not found
#include <sif/list.h>
         ^~~~~~~~~~~~
1 error generated.
```
Which can be fixed via go tooling:

adding
```go
// +build singularity_sif


```
at the top of `singularity/pkg/image/sifcore.go`

-/-
 - Also updates support mail addr

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/mlocal/checks/project.chk
+++ b/mlocal/checks/project.chk
@@ -8,7 +8,7 @@ config_add_def PACKAGE_NAME \"singularity\"
 config_add_def PACKAGE_TARNAME \"singularity\"
 config_add_def PACKAGE_VERSION \"3.0\"
 config_add_def PACKAGE_STRING \"singularity 3.0\"
-config_add_def PACKAGE_BUGREPORT \"gmkurtzer@gmail.com\"
+config_add_def PACKAGE_BUGREPORT \"support@sylabs.io\"
 config_add_def PACKAGE_URL \"\"
 
 config_add_def BUILDDIR \"$builddir\"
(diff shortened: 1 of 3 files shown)


dtrudg:
One question - Could this instead be a negative build tag - i.e. set a build tag to *not* build SIF? The reason I suggest this is that this code is sitting inside the singularity git repo, where normal behavior needs SIF. Using from elsewhere kinda seems like the exception that you should set a build tag for to get.

As precedent, the `containers_image_openpgp` PGP tag we are using is something that is dropping features from `containers/image` through its use - it disables the full featured default gpgme to use a simple go implementation.

bauerm97 (changes requested):
I agree with @dctrud , this should be a negative tag. The mainline Singularity project shouldn't need to include the build tag.

ArangoGutierrez:
in discussion with @tri-adam , we agreed on the `-tag` should work the other way around. External packages should _just work_... and the singularity source itself can use a tag to enable things that don't work.

bauerm97 (approved):

bauerm97:
As discussed, I'm convinced now. The upstream Singularity tree will *have* to include this workaround either way, the question becomes does another package have to worry about this too. I guess it's better that *only* the Singularity tree has to include code for this workaround, rather than any external project having to know as well.

PR #1601, opened 2018-05-25
3.0 exported pkg build

bauerm97:

Remove sbuild binary and related code as it is no longer necessary. There will be a package created for external project use to build singularity images that doesn't depend on a separate binary.


mlocal/frags/Makefile.stub
```diff
@@ -170,7 +156,7 @@ test:
 	@echo "       PASS"
 	@echo " TEST go test"
 	$(V)export CGO_CPPFLAGS="$(cgo_CPPFLAGS)" CGO_LDFLAGS="$(cgo_LDFLAGS)" && \
-		cd $(SOURCEDIR) && go test -tags $(go_TAG) -cover -race -timeout 60s ./...
+		cd $(SOURCEDIR) && go test -count=1 -tags $(go_TAG) -cover -race -timeout 60s ./...
```
tri-adam, line 159: I'm assuming the `-count=1` flag is to disable result caching? Do we really want to do that?

src/pkg/build/builder_sif.go
```diff
@@ -112,8 +38,28 @@ func NewSifBuilderJSON(r io.Reader, imagePath string) (b *SIFBuilder2, err error
...
+		b.ImagePath = imagePath
+		b.tmpfs, err = image.TempSandbox(path.Base(imagePath) + "-")
+		if err != nil {
+			return b, err
+		}
+
+		uri := d.Header["bootstrap"] + "://" + d.Header["from"]
+		b.p, err = NewProvisionerFromURI(uri)
+
+		return b, err*/
```
tri-adam: I'd like to see the commented-out code removed from this file. It clutters things up, and doesn't aid in understanding the code.

tri-adam (changes requested):

mlocal/frags/Makefile.stub
bauerm97: Yes, to disable caching. Since I'm using `tests/dummy_test.go` as a testing framework for Singularity (rather than as a Go unit test) `go test` will not re-run the test on that package unless the code in that package changes. However, it doesn't realize that the result of that test is dependent on other code in the tree changing, so a cached test will give incorrect results. A way to solve this would be to stop using `go test` in this sort-of not intended way.

src/pkg/build/builder_sif.go
bauerm97: Sure, just forgot to delete!

mlocal/frags/Makefile.stub
tri-adam: OK, gotcha.

dtrudg: Is `GOCACHE=off go test ...` a better way of making that explicit?

tri-adam (approved):

mlocal/frags/Makefile.stub
tri-adam: From the help:

> The idiomatic way to disable test caching explicitly is to use -count=1.

Seems a bit weird to me, but who am I to argue with the `Go`ds?

PR #1615, opened 2018-05-31
Moving docs

GodloveD:
Moving the `/docs` directory so that it is within `/src` as requested by @yhcote.


tri-adam (dismissed):
LGTM

yhcote:
Note that the current content of docs/ should still contain : 
    2.x-slurm/
    2.x-tests/
    man/

for the time being. 2.x-slurm and 2.x-tests were elements of 2.x that we may wanna keep as reference for a while as we decide if we want some of those things into 3.x in the future. We can remove these 2 folders a any moment when we know we covered those things and don't need reminders anymore.

GodloveD:
@yhcote   I think we should just do away with `man/` altogether right?  afaik, the purpose of `/man` was to provide a place to save the lonely little man page that Greg originally wrote that would get installed when you installed Singularity.  Now we are generating man pages automagically (including the main Singularity man page) and installing them if the user runs `make man`.  So this directory is just cruft now right?

yhcote:
Well I think that with time, there will be more things that count as docs/ that we could shove in there. man is the only true doc we have in there now, but I expect anything related to info about singularity to eventually go there.

GodloveD:
@tri-adam your review was totally fresh and I did not mean to dismiss it. 😸

GodloveD:
@yhcote I deleted `docs/man`.  I guess if we get any other files that should go there we can recreate the directory again without too much trouble.  Other than that I think I've addressed the issues you brought up?

bauerm97 (approved):
Looks good to me. I agree with decision to remove `docs/man` since that was an artifact from days past :-)

PR #1618, opened 2018-06-04
Auth token

ArangoGutierrez:

Currently 
Pull/Push handles the AuthToken as:
```go
func init() {
	PushCmd.Flags().SetInterspersed(false)

	usr, err := user.Current()
	if err != nil {
		sylog.Fatalf("Couldn't determine user home directory: %v", err)
	}

	defaultTokenFile := path.Join(usr.HomeDir, ".singularity", "sylabs-token")
	PushCmd.Flags().StringVar(&PushLibraryURI, "libraryuri", "https://library.sylabs.io", "")
	PushCmd.Flags().StringVar(&PushTokenFile, "tokenfile", defaultTokenFile, "path to the file holding your sylabs authentication token")
	SingularityCmd.AddCommand(PushCmd)
}
```
And remote build handles as:
```go
func init() {
BuildCmd.Flags().StringVar(&authToken, "auth-token", "", "Specify the auth token for the remote builder")
}
```
Now is handled in a global way at the singularity/SingularityCmd init()
```go
	usr, err := user.Current()
	if err != nil {
		sylog.Fatalf("Couldn't determine user home directory: %v", err)
	}
	defaultTokenFile := path.Join(usr.HomeDir, ".singularity", "sylabs-token")
	// authToken priority default_file < env < file_flag
	authToken = auth.ReadToken(defaultTokenFile)
	SingularityCmd.Flags().StringVar(&tokenFile, "tokenfile", defaultTokenFile, "path to the file holding your sylabs authentication token")
	if val := os.Getenv("SYLABS_TOKEN"); val != "" {
		authToken = val
	}
	if i := strings.Compare(tokenFile, defaultTokenFile); i != 0 {
		authToken = auth.ReadToken(tokenFile)
	}
```
exposing `authToken` as a global var for the `cli` pkg

**Checkoff for all PRs:**

Attn: @singularityware-admin


src/pkg/util/auth/token.go
```diff
@@ -0,0 +1,47 @@
...
+	"strings"
+
+	"github.com/singularityware/singularity/src/pkg/sylog"
+)
+
+// ReadToken reads a sylabs JWT auth token from a file
+func ReadToken(tokenPath string) string {
+	buf, err := ioutil.ReadFile(tokenPath)
+	if err != nil {
+		sylog.Warningf("Couldn't read your Sylabs authentication token. Only pulls of public images will succeed.\n")
```
dtrudg: We probably need to rethink these warnings now that they are happening on all init of the CLI. I think we should not warn if there is no token... unless we are using build --remote / push / pull things that may need a token.

dtrudg (changes requested):
Main thing here, @ArangoGutierrez, is that we need to make the warning messages sane. We don't want to annoy people where they aren't using functions that need a token.

src/cmd/singularity/cli/singularity.go
```diff
@@ -38,6 +50,20 @@ func init() {
 	SingularityCmd.Flags().BoolVarP(&quiet, "quiet", "q", false, "Suppress all normal output")
 	SingularityCmd.Flags().BoolVarP(&verbose, "verbose", "v", false, "Increase verbosity +1")
 
+	usr, err := user.Current()
+	if err != nil {
+		sylog.Fatalf("Couldn't determine user home directory: %v", err)
+	}
+	defaultTokenFile := path.Join(usr.HomeDir, ".singularity", "sylabs-token")
+	// authToken priority default_file < env < file_flag
+	authToken = auth.ReadToken(defaultTokenFile)
```
dtrudg: I don't think we should read this before checking for an env / var / flag. The logic in here should stop uneccessary file reads.

src/cmd/singularity/cli/singularity.go
```diff
@@ -38,6 +50,20 @@ func init() {
...
+	SingularityCmd.Flags().StringVar(&tokenFile, "tokenfile", defaultTokenFile, "path to the file holding your sylabs authentication token")
+	if val := os.Getenv("SYLABS_TOKEN"); val != "" {
+		authToken = val
+	}
+	if i := strings.Compare(tokenFile, defaultTokenFile); i != 0 {
```
dtrudg: Is it possible to add tests for these env / var flag overrides?

dtrudg (dismissed):

src/cmd/singularity/cli/singularity.go
ArangoGutierrez: help appreciated :)

src/cmd/singularity/cli/singularity.go
ArangoGutierrez: agree

src/pkg/util/auth/token.go
ArangoGutierrez: agree

src/cmd/singularity/cli/singularity.go
```diff
@@ -56,13 +56,14 @@ func init() {
...
-	authToken = auth.ReadToken(defaultTokenFile)
 	SingularityCmd.Flags().StringVar(&tokenFile, "tokenfile", defaultTokenFile, "path to the file holding your sylabs authentication token")
 	if val := os.Getenv("SYLABS_TOKEN"); val != "" {
 		authToken = val
 	}
 	if i := strings.Compare(tokenFile, defaultTokenFile); i != 0 {
-		authToken = auth.ReadToken(tokenFile)
+		authToken, authWarning = auth.ReadToken(tokenFile)
+	} else {
+		authToken, authWarning = auth.ReadToken(defaultTokenFile)
```
dtrudg: Don't think this will work for the priority intended? The defaultTokenFile is now going to override something set by the env var if I'm reading this correctly.

src/pkg/util/auth/token.go
```diff
@@ -0,0 +1,48 @@
...
+package auth
+
+import (
+	"io/ioutil"
+	"strings"
+)
+
+var (
+	// WarningTokenTooShort Warning return for token shorter than 200 b
+	WarningTokenTooShort = "Token is too short to be valid. Only pulls of public images will succeed"
```
dtrudg: Not sure why it's useful to have these defined here - and if there is a reason they should be const.

src/cmd/singularity/cli/singularity.go
```diff
@@ -38,6 +50,21 @@ func init() {
...
+		sylog.Fatalf("Couldn't determine user home directory: %v", err)
+	}
+	defaultTokenFile := path.Join(usr.HomeDir, ".singularity", "sylabs-token")
+	// authToken priority default_file < env < file_flag
+	SingularityCmd.Flags().StringVar(&tokenFile, "tokenfile", defaultTokenFile, "path to the file holding your sylabs authentication token")
+	if val := os.Getenv("SYLABS_TOKEN"); val != "" {
+		authToken = val
+	}
+	if i := strings.Compare(tokenFile, defaultTokenFile); i != 0 {
+		authToken, authWarning = auth.ReadToken(tokenFile)
```
dtrudg: I can't see where anything is done with the authWarning?

I don't think it should really be handled here. Would it be better for something like:

- If there is a gross problem with reading an explicit token file then it's a fatal here.
- If no token exists the warning gets handled inside the builder / library code when that code sees authToken == "" ?

dtrudg (dismissed):

ArangoGutierrez:
@dctrud ready for a review

dtrudg:
Hi @ArangoGutierrez - Afraid this still does not work for the env var - it is still trying to read a token file

```
08:38 AM $ export SYLABS_TOKEN="....."
dave@piran:~/go/src/github.com/singularityware
08:38 AM $ singularity push --libraryuri https://library-test.sylabs.io ubuntu_bionic.sif library://tok/linux/ubuntu:latest
WARNING [U=1000,P=8357]    sylabsToken()                 Authentication token file not found : Only pulls of public images will succeed
FATAL   [U=1000,P=8357]    func12()    

```

dtrudg (approved):
This tests okay for me. Let's go with this and we can refine the implementation as needed as the CLI is fleshed out with more operations.

PR #1621, opened 2018-06-05
CentOS 6 Tests Rebase

dtrudg:

Support CentOS 6 for `make test` and travis rpmbuild tests. This is important now that we have integration of libs such as libarchive (used by docker-extract) that are quite different in older versions.

Changes:

- Let travis run test suite on CentOS 6
- Skip only python tests where python <2.7 (CentOS 6 has 2.6 and test code is not compatible)
- Skip arch linux container test if we are on a 2 series kernel (will get kernel too old on CentOS 6)
- Fix `SINGULARITY_OVERLAY_FS` check - this variable is always defined, 0 or 1 - so check for that, not `-n`
- Fix busybox non-overlay user bind test - `/opt` doesn't actually exist in the busybox container, so we have to user bind somewhere else to check functionality without overlay. Use `/var/tmp` which is available and empty.
- After a squashfs image is created, `chmod a+rx` (previously a+x) as on EL6 distros it will not have 'r' by default, and is unusable by other than root until manual chmod is performed.

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/.travis/rpmbuild_test
+++ b/.travis/rpmbuild_test
@@ -13,14 +13,13 @@ rpmbuild -ta *.tar.gz
 rpm -iv ~/rpmbuild/RPMS/*/*.rpm
 
 if [ "$OS_VERSION" = 6 ]; then
-    echo "the test suite has not yet been ported to centos6 python, skipping."
-    exit
+    echo "the python tests have not yet been ported to centos6 python, and will be skipped."
+else
+    # Install python 3 and pylint
+    yum install -y epel-release
+    yum install -y python34 python34-pip python34-setuptools docker
+    yum install -y pylint
 fi
-
-# Install python 3 and pylint
-yum install -y epel-release
-yum install -y python34 python34-pip python34-setuptools docker
-yum install -y pylint
 
 # run the test suite
 yum install -y libtool sudo which e2fsprogs
--- a/.travis/rpmbuild_test
+++ b/.travis/rpmbuild_test
@@ -27,7 +27,6 @@ sed -i 's,^prefix=.*,prefix=/usr,' test.sh
 sed -i 's,^sysconfdir=.*,sysconfdir=/etc,' test.sh
 sed -i 's,^localstatedir=.*,localstatedir=/var,' test.sh
 useradd -u 1000 testuser
-echo "Defaults:testuser env_keep=DOCKER_HOST" >>/etc/sudoers
-echo "Defaults umask=0777" >>/etc/sudoers
+echo "Defaults:testuser env_keep=DOCKER_HOST umask=0777 umask_override=1" >>/etc/sudoers
 echo "testuser ALL=(ALL) NOPASSWD: ALL" >>/etc/sudoers
-sudo -u testuser make test
+su testuser -c "make test"
--- a/.travis/rpmbuild_test
+++ b/.travis/rpmbuild_test
@@ -27,6 +27,6 @@ sed -i 's,^prefix=.*,prefix=/usr,' test.sh
 sed -i 's,^sysconfdir=.*,sysconfdir=/etc,' test.sh
 sed -i 's,^localstatedir=.*,localstatedir=/var,' test.sh
 useradd -u 1000 testuser
-echo "Defaults:testuser env_keep=DOCKER_HOST umask=0777 umask_override=1" >>/etc/sudoers
+echo "Defaults:testuser env_keep=DOCKER_HOST" >>/etc/sudoers
 echo "testuser ALL=(ALL) NOPASSWD: ALL" >>/etc/sudoers
 su testuser -c "make test"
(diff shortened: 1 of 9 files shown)


dtrudg:
@GodloveD - this is actually running the tests now and passing on CentOS6 - but there is a weirdness on the CentOS6 run that leads to confusing CI logs. It is not redirecting stdout to `/dev/null` properly within the Singularity `build.exec` so you see a bunch of spurious errors from `image-test` and `docker-extract` where `>/dev/null 2>/dev/null` is used in tests calling these binaries. Strangely the redirection on stderr is happening. This is a bit odd - and I can't find a fix yet :-(

dtrudg:
A bit more looking around on this - the issues with spurious messages in the CI logs (which should have gone to /dev/null) comes from the combination of bash <4.2, the `-x` option to bash (for tracing) and redirection inside a docker container. I don't think this can be fixed at all easily. The tests work correctly - just `docker-extract` and -`image-type` errors that should have gone to `/dev/null` appear in the CI logs.

tests/32-action_options.sh
```diff
@@ -54,7 +54,7 @@ stest 0 mkdir -p "$TESTDIR"
 stest 0 touch "$TESTDIR/testfile"
 stest 0 singularity exec --home "$TESTDIR" "$CONTAINER" test -f "$TESTDIR/testfile"
 stest 0 singularity exec --home "$TESTDIR:/home" "$CONTAINER" test -f "/home/testfile"
-if [ -n "${SINGULARITY_OVERLAY_FS:-}" ]; then
+if [  "$SINGULARITY_OVERLAY_FS" = "1" ]; then
```
jmstover: Can you change this to be:
`if [  "x${SINGULARITY_OVERLAY_FS}" = "x1" ]; then`

Not _that_ important... but if `SINGULARITY_OVERLAY_FS` gets undefined you'll have a parse error with no left side.

tests/35-userbinds.sh
```diff
@@ -33,9 +33,9 @@ CONTAINER="$SINGULARITY_TESTDIR/container.img"
 stest 0 sudo singularity build "$CONTAINER" "../examples/busybox/Singularity"
 
 stest 0 touch /tmp/hello_world_test
-stest 0 singularity exec -B /tmp:/opt "$CONTAINER" test -f /opt/hello_world_test
+stest 0 singularity exec -B /tmp:/var/tmp "$CONTAINER" test -f /var/tmp/hello_world_test
 
-if [ -n "$SINGULARITY_OVERLAY_FS" ]; then
+if [ "$SINGULARITY_OVERLAY_FS" = "1" ]; then
```
jmstover: Same here...

jmstover (approved):

GodloveD (changes requested):
Still getting the following on Ubuntu 14 (no overlay)
```
################################################################################
Checking escalation block (script: 40-privblock.sh, testdir: /tmp/stest.8zTLIU)

 + sudo singularity build /tmp/stest.8zTLIU/container.img docker://centos:7                      (retval=0) OK
 + singularity exec /tmp/stest.8zTLIU/container.img true                                         (retval=0) OK
 + singularity exec /tmp/stest.8zTLIU/container.img false                                        (retval=1) OK
 + sudo singularity exec /tmp/stest.8zTLIU/container.img chsh -s /bin/sh                         (retval=1) ERROR
chsh: user attribute not changed: error locking file: Read-only file system
Changing shell for root.
Full output in: /tmp/stest.8zTLIU
ERROR: Failed running test: 40-privblock.sh
make: *** [test] Error 1
```
Can you plz wrap those tests too?

GodloveD (approved):

GodloveD:
Thanks @dctrud and @jmstover !

.travis/rpmbuild_test
```diff
@@ -13,21 +13,20 @@ rpmbuild -ta *.tar.gz
 rpm -iv ~/rpmbuild/RPMS/*/*.rpm
 
 if [ "$OS_VERSION" = 6 ]; then
-    echo "the test suite has not yet been ported to centos6 python, skipping."
-    exit
+    echo "the python tests have not yet been ported to centos6 python, and will be skipped."
+else
```
DrDaveD, line 17: @dctrud did you know that python3 is available in EPEL6?  Just add the epel-release rpm and do 'yum install /usr/bin/python3'.

dtrudg: Hi @DrDaveD - yep, am aware - but under standard installs on RHEL6 / CentOS6 we'd not be using this - so decided it wouldn't really add anything to have it tested against here. Cheers.

DrDaveD: I see, but FYI EPEL now requires that all packages capable of supporting python3 be compiled to use it, so when I installed singularity-2.5.1 in EPEL I enabled it.  So installs of singularity from EPEL will indeed use python3, and that's probably the place people using EL6 are most likely to get it. 

Anyway, thank you very much for enabling this on the EL6 travis build.  It's a big step up from no tests.  Running python3 tests additionally there are probably not a very big value because they the tests are also run on EL7.

PR #1625, opened 2018-06-06
Development 3.0: Mount point management

cclerget:

This PR adds a mount point structure with associated methods allowing to create/import/manage a list of mount points based on OCI mount spec


src/pkg/util/fs/mount/mount.go
```diff
@@ -0,0 +1,366 @@
+/*
+  Copyright (c) 2018, Sylabs, Inc. All rights reserved.
+
+  This software is licensed under a 3-clause BSD license.  Please
+  consult LICENSE file distributed with the sources of this project regarding
+  your rights to use or distribute this software.
+*/
```
ArangoGutierrez:
```go
//  Copyright (c) 2018, Sylabs, Inc. All rights reserved.
//
//  This software is licensed under a 3-clause BSD license.  Please
//  consult LICENSE file distributed with the sources of this project regarding
//  your rights to use or distribute this software.
```

src/pkg/util/fs/mount/mount_test.go
```diff
@@ -0,0 +1,360 @@
+/*
+  Copyright (c) 2018, Sylabs, Inc. All rights reserved.
+
+  This software is licensed under a 3-clause BSD license.  Please
+  consult LICENSE file distributed with the sources of this project regarding
+  your rights to use or distribute this software.
+*/
```
ArangoGutierrez:
```go
//  Copyright (c) 2018, Sylabs, Inc. All rights reserved.
//
//  This software is licensed under a 3-clause BSD license.  Please
//  consult LICENSE file distributed with the sources of this project regarding
//  your rights to use or distribute this software.
```

ArangoGutierrez (dismissed):

src/pkg/util/fs/mount/mount_test.go
```diff
@@ -0,0 +1,450 @@
...
+	"testing"
+
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+func TestImage(t *testing.T) {
+	points := &Points{}
+
+	if err := points.AddImage("", "/fake", "ext3", 0, 0); err == nil {
+		t.Errorf("should failed with empty source")
```
DrDaveD: Do you mean "should have failed"?

cclerget: @DrDaveD Yes :), thanks !

src/pkg/util/fs/mount/mount.go
```diff
@@ -0,0 +1,370 @@
...
+			finalOptions = append(finalOptions, optionTrim)
+		}
+		isFlag = false
+	}
+	return flags, finalOptions
+}
+
+// Points defines and stores a set of mount points
+type Points struct {
+	Context string
```
bauerm97, line 99: Would it be better to have points be like this:

```
type Point struct {
    Context string
    point specs.Mount
}

var Points []Point
```

This way each point can have unique context? Or maybe I don't quite understand the use of `Context string` yet and it's really necessary only for the entire set of `Points`?

src/pkg/util/fs/mount/mount.go
```diff
@@ -0,0 +1,370 @@
...
+	return nil
+}
+
+// GetAll returns all registered mount points
+func (p *Points) GetAll() []specs.Mount {
+	return p.points
+}
+
+// Get returns registered mount points with the matched destination
+func (p *Points) Get(dest string) []specs.Mount {
```
bauerm97: Does it make more sense to get the point associated with the src rather than dest? Not sure the correct answer here, just a thought! Or maybe provide:

`func GetFromSource(src string)`
`func GetFromDest(dest string)`

So user of package can choose

src/pkg/util/fs/mount/mount.go
```diff
@@ -0,0 +1,370 @@
...
+		if point.Destination == dest {
+			mounts = append(mounts, point)
+			break
+		}
+	}
+	return mounts
+}
+
+// Remove removes mount points identified by destination
+func (p *Points) Remove(dest string) {
```
bauerm97: Same comment as from `Get()`

src/pkg/util/fs/mount/mount.go
```diff
@@ -0,0 +1,370 @@
...
+			}
+		}
+		p.RemoveAll()
+		return fmt.Errorf("can't import mount points list: %s", err)
+	}
+	return nil
+}
+
+// AddImage adds an image mount point
+func (p *Points) AddImage(source string, dest string, fstype string, flags uintptr, offset uint64) error {
```
bauerm97: Can we make this function additionally take `length uint64`. This way we can natively mount a partition from SIF

bauerm97 (changes requested):
Hi @cclerget looking good so far, there are a few things I think are very important to do. Once that gets done we can immediately make use of this in engines :-)

Also, we discussed on slack to provide helper function or interface to execute mounting behavior. Is this underway or do you want some extra manpower on this? Cheers man!

src/pkg/util/fs/mount/mount.go
cclerget: Context is for SELinux mount context, OCI allow to defines one mountlabel, I understood that as one mountlabel for all mount points. That's the meaning of Context here

src/pkg/util/fs/mount/mount.go
cclerget: Source can be empty (for remount point) or set to none for FS, dest ensure you get the right mount and unique mount point

src/pkg/util/fs/mount/mount.go
bauerm97: Ok cool, in which case makes sense to leave this as is

src/pkg/util/fs/mount/mount.go
bauerm97: I see! I think it works, but maybe nice to provide option to search for source as well?

src/pkg/util/fs/mount/mount.go
cclerget: There is offset argument ;)

bauerm97 (approved):
Changes look good. Thanks !

PR #1630, opened 2018-06-08
Expose libraryRef Validation Routines

tri-adam:
Expose `libraryRef` validation Routines, so they can be used by other packages.


dtrudg (approved):
Looks good to me.

PR #1636, opened 2018-06-11
RB Pull from Library

tri-adam:

Add `LibraryRef`/`LibraryURL` params to `RequestData` and `ResponseData`. Refactor authentication token handling. Add unit tests to exercise `LibraryRef`/`LibraryURL` handling.

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/src/cmd/singularity/cli/build.go
+++ b/src/cmd/singularity/cli/build.go
@@ -101,7 +101,7 @@ var BuildCmd = &cobra.Command{
 					authWarning != auth.WarningTokenToolong &&
 					authWarning != auth.WarningTokenTooShort {
 					if authToken != "" {
-						b = build.NewRemoteBuilder(args[0], def, false, remoteURL, authToken)
+						b = build.NewRemoteBuilder(args[0], "", def, false, remoteURL, authToken)
 					}
 				} else {
 					sylog.Fatalf("Unable to submit build job: %v", authWarning)
--- a/src/pkg/build/builder_remote.go
+++ b/src/pkg/build/builder_remote.go
@@ -10,7 +10,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"io"
 	"net/http"
 	"net/url"
 	"strings"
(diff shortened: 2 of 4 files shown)


src/cmd/singularity/cli/build.go
```diff
@@ -101,7 +101,7 @@ var BuildCmd = &cobra.Command{
 					authWarning != auth.WarningTokenToolong &&
 					authWarning != auth.WarningTokenTooShort {
 					if authToken != "" {
-						b = build.NewRemoteBuilder(args[0], def, false, remoteURL, authToken)
+						b = build.NewRemoteBuilder(args[0], "", def, false, remoteURL, authToken)
```
ArangoGutierrez, line 104: Should this be a flag? `--libraryurl` pointing to the default lib URL?

src/pkg/build/builder_remote.go
```diff
@@ -155,9 +168,17 @@ func (rb *RemoteBuilder) streamOutput(ctx context.Context, url string) (err erro
 }
 
 // doBuildRequest creates a new build on a Remote Build Service
-func (rb *RemoteBuilder) doBuildRequest(ctx context.Context, d Definition) (rd ResponseData, err error) {
+func (rb *RemoteBuilder) doBuildRequest(ctx context.Context, d Definition, libraryRef string) (rd ResponseData, err error) {
```
ArangoGutierrez, line 171: the `rb` var with method `doBuildRequest` has definition and library ref stored, we could just do `rb.definition` instead of passing then as an argument to the func

ArangoGutierrez (dismissed):

src/cmd/singularity/cli/build.go
tri-adam: Trying to keep it simple for now. We can expose that later when necessary.

src/pkg/build/builder_remote.go
tri-adam: They are different things. `rb.ImagePath` can refer to a local image path or a Library reference. The `libraryRef` passed to `doBuildRequest()` is always a Library reference (or `""`, in which case RB will auto-generate a Library reference.)

ArangoGutierrez (approved):

dtrudg (approved):
Thumbs up from me :+1:

One open question - do we need to thing about testing strategies as the interconnections between services grow?

tri-adam:
Yes, I think we are going to need to think about that. I was thinking it might be nice to have a small "mock" package for each service, so that any unit tests that interact with external services can easily fire up a mock service locally without duplicating code. Thoughts?

dtrudg:
@tri-adam a mock service definitely makes sense. Note that with swagger API definitions there is this.... 

https://github.com/stoplightio/prism

PR #1637, opened 2018-06-11
Fix read token workflow for build

ArangoGutierrez:

With HEAD of singularity-3.0 (e583a3b76ef917329fa0ce06163e4ece53ea78b3), running a remote Singualrity build without a token file present causes a SIGSEGV

**Checkoff for all PRs:**

Attn: @singularityware-admin


src/cmd/singularity/cli/build.go
```diff
@@ -102,6 +102,8 @@ var BuildCmd = &cobra.Command{
 					authWarning != auth.WarningTokenTooShort {
```
dtrudg: This should probably be refactored? Don't think you need to test the value of authWarning against all of these? Plus  the nested if / else are unnecessarily complex.

src/pkg/util/auth/token.go
```diff
@@ -28,10 +28,15 @@ const (
 // ReadToken reads a sylabs JWT auth token from a file
 func ReadToken(tokenPath string) (token, warning string) {
 	// check if token file exist
-	if _, err := os.Stat(tokenPath); os.IsNotExist(err) {
+	tokenFile, err := os.Stat(tokenPath)
+	if os.IsNotExist(err) {
 		return "", WarningTokenFileNotFound
 	}
 
+	if tokenFile.IsDir() {
```
dtrudg: Don't test for dir - it could also be a socket or something else, and we don't want to add tests for everything it shouldn't be. Try to test instead that it is a readable normal file (or a link that can be resolved to a readable normal file)

dtrudg (dismissed):

src/cmd/singularity/cli/build.go
```diff
@@ -102,6 +102,8 @@ var BuildCmd = &cobra.Command{
...
 					if authToken != "" {
 						b = build.NewRemoteBuilder(args[0], def, false, remoteURL, authToken)
+					} else {
+						return
```
tri-adam: I'm really not sure about the logic in this section. On the one hand, if `authWarning` is set to `auth.WarningEmptyToken`, it `Fatalf`s. But if the token is `""` (which is empty?) it returns? 

Shouldn't this just `Fatalf` if `authToken` is not valid for any reason? Why the divergent strategies? How can `authToken == ""` and `authWarning != auth.WarningEmptyToken` both be true (as seems to be the case right now?)

ArangoGutierrez: `auth.Warning` is set when reading the tokenFile, but we also leave room for reading the token from an env var. that's why the double check

src/pkg/util/auth/token.go
tri-adam: I think this check might be redundant, actually, since `ioutil.ReadFile()` will fail. The logic fix @ArangoGutierrez just applied in `build.go` should fix the issue there.  I'm not 100% sure about the Library side though.

For consistency, should the `"Couldn't read your Sylabs authentication token"` `string` below be a defined `const` like the other `warning` values?

src/cmd/singularity/cli/build.go
```diff
@@ -97,16 +96,11 @@ var BuildCmd = &cobra.Command{
 			}
 
 			if remote {
-				if authWarning != auth.WarningEmptyToken &&
-					authWarning != auth.WarningTokenToolong &&
-					authWarning != auth.WarningTokenTooShort {
-					if authToken != "" {
-						b = build.NewRemoteBuilder(args[0], def, false, remoteURL, authToken)
-					}
+				if authWarning == "" || authToken != "" {
```
dtrudg: When can authWarning be set but authToken is not?

src/pkg/util/auth/token.go
tri-adam: This `IsDir()` check should either be updated as @dctrud suggested, or removed completely (since `ioutil.ReadFile()` will catch the case where it isn't a file.)

src/cmd/singularity/cli/build.go
tri-adam: Should it ever proceed if `authWarning` is non-empty? For example, if `authWarning` is `auth.WarningTokenTooShort`? The logic here used to `Fatalf`, was this change intentional?

src/cmd/singularity/cli/push.go
```diff
@@ -32,15 +31,11 @@ var PushCmd = &cobra.Command{
...
 	PreRun: sylabsToken,
 	Run: func(cmd *cobra.Command, args []string) {
-		if authWarning != auth.WarningEmptyToken &&
-			authWarning != auth.WarningTokenToolong &&
-			authWarning != auth.WarningTokenTooShort {
-			if authToken != "" {
-				libexec.PushImage(args[0], args[1], PushLibraryURI, authToken)
-				return
-			}
+		if authWarning == "" || authToken != "" {
```
tri-adam: Same here, should it ever proceed if `authWarning` is non-empty? For example, if `authWarning` is `auth.WarningTokenTooShort`? The logic here used to `Fatalf` in this case, was this change intentional?

src/cmd/singularity/cli/push.go
```diff
@@ -32,15 +31,11 @@ var PushCmd = &cobra.Command{
...
+			libexec.PushImage(args[0], args[1], PushLibraryURI, authToken)
+		} else {
+			sylog.Fatalf("Couldn't push image to library: %v", authWarning)
```
dtrudg: This still reads weirdly to me. Are we certain we cannot we get a situation where the authWarning is "" and token is also ""?

tri-adam (changes requested):

src/pkg/util/auth/token.go
dctrud: ioutil.ReadFile does not fail if given a dir... see https://play.golang.org/p/EHGX4lymOKQ

dtrudg (dismissed):

src/cmd/singularity/cli/build.go
ArangoGutierrez: the logic I'm trying to get there is, if an env var is set, without a file, the warning is going to be not-null, but AuthToken is going to exist via the env var.
> will do more checks on the token set step.

dtrudg: @ArangoGutierrez - okay, in that case can we have more explicitly named vars so that it's intuitive reading the if?

src/cmd/singularity/cli/push.go
tri-adam: I'm still a little confused about whether `authWarning` is meant to prevent usage of `authToken` or just warn about it. If there are cases where we want to use the token when a warning is also present, I'd suggest naming those (rather than the failure cases.) Missing a warning that should have been fatal was what caused the segmentation fault in the build code.

src/cmd/singularity/cli/build.go
tri-adam: I'm still a little confused about whether `authWarning` is meant to prevent usage of `authToken` or just warn about it. If there are cases where we want to use the token when a warning is also present, I'd suggest naming those (rather than the failure cases.) Missing a warning that should have been fatal was what caused the segmentation fault in the build code.

src/pkg/util/auth/token.go
```diff
@@ -23,18 +23,23 @@ const (
 	WarningEmptyToken = "Token file is empty"
 	// WarningTokenFileNotFound token file not found
 	WarningTokenFileNotFound = "Authentication token file not found"
+	// WarningCouldntReadFile Warning return for issues when reading file
+	WarningCouldntReadFile = "Couldn't read your Sylabs authentication token"
+	// WarningPathIsDir return when given tokenFile path is a directory
+	WarningPathIsDir = "Given token file path points to a directory not a file"
```
dtrudg: I don't think this line is used ever now?

tri-adam: Can remove this now, I think it's unused?

tri-adam (dismissed):
Can confirm this fixes the original two issues. With a missing file:

```sh
WARNING [U=1000,P=109008]  sylabsToken()                 Authentication token file not found : Only pulls of public images will succeed
FATAL   [U=1000,P=109008]  func4()                       Unable to submit build job: Authentication token file not found
```

And with a directory:

```sh
FATAL   [U=1000,P=109018]  func4()                       Unable to submit build job: Couldn't read your Sylabs authentication token
```

ArangoGutierrez:
@tri-adam | @dctrud  (updated to upstream branch) and a happy travis .
ready for that 👍

dtrudg (approved):

tri-adam (approved):

PR #1645, opened 2018-06-13
Patch image-tools for extraction perms as non-root

dtrudg:

Patch `opencontainers/image-tools` so that if we are not running as root during build we ensure owner has write perm on every file and directory as they are extracted from layer files. This fixes a failure building e.g. `docker://centos` as non-root user.


jmstover (approved):
This looks good to me.

gmkurtzer (approved):
Working for me! Thanks!

jmstover:
@yhcote Since we're touching some of mlocal here in this patch you may want to glance over it

PR #1653, opened 2018-06-13
Small adjustements for PR 1639

cclerget:
- Fix function return.
- Fix file config parser bug.
- Remove redundant initialization of singularity.conf.
- Remove runtime parameter from SMaster and SContainer.
- Change smaster processus name.


bauerm97 (approved):
Changes look good Cedric. I like removing the sruntime arg from SContainer && SMaster

PR #1655, opened 2018-06-14
Point keysigning at a real URL

dtrudg:

Point the SIF signing code at a real, non example.com URL that will persist.


ArangoGutierrez (approved):
LGTM

PR #1663, opened 2018-06-17
Support additional docker and oci build sources

dtrudg:
This PR is working towards allowing `singularity build` to leverage the functionality in `containers/image` to  build from....

- Docker URIs (registy or hub docker://...)
- Docker archives (tar files from `docker save`)
- OCI archives
- OCI layouts (directories with an OCI specified structure

I'm following the `containers-image` reference style, so at the command line you currently use...

```
singularity build test.sif docker://alpine
singularity build test.sif docker-archive:alpine.tar
singularity build test.sif oci-archive:alpine.tar
singularity build test.sif oci:/tmp/alpine
```

Here's what it looks like building from a docker archive:

```
dave@piran:/tmp/oci
08:40 PM $ docker save alpine -o alpine.tar
dave@piran:/tmp/oci
08:40 PM $ singularity build alpine.sif docker-archive:alpine.tar
Copying blob sha256:cd7100a72410606589a54b932cabd804a17f9ae5b42a1882bd56d263e02b6215
 4.20 MiB / 4.20 MiB [======================================================] 0s
Copying config sha256:b1a7f144ece0194921befe57ab30ed1fd98c5950db7996719429020986092058
 585 B / 585 B [============================================================] 0s
Storing signatures
Parallel mksquashfs: Using 12 processors
Creating 4.0 filesystem on /tmp/squashfs-028201854.img, block size 131072.
[================================================================================================================|] 99/99 100%

Exportable Squashfs 4.0 filesystem, gzip compressed, data block size 131072
	compressed data, compressed metadata, compressed fragments, compressed xattrs
	duplicates are removed
Filesystem size 1957.78 Kbytes (1.91 Mbytes)
	47.90% of uncompressed filesystem size (4087.02 Kbytes)
Inode table size 3278 bytes (3.20 Kbytes)
	15.11% of uncompressed inode table size (21694 bytes)
Directory table size 4597 bytes (4.49 Kbytes)
	54.75% of uncompressed directory table size (8397 bytes)
Xattr table size 53 bytes (0.05 Kbytes)
	100.00% of uncompressed xattr table size (53 bytes)
Number of duplicate files found 6
Number of inodes 496
Number of files 75
Number of fragments 4
Number of symbolic links  332
Number of device nodes 0
Number of fifo nodes 0
Number of socket nodes 0
Number of directories 89
Number of ids (unique uids + gids) 1
Number of uids 1
	dave (1000)
Number of gids 1
	dave (1000)
```

This is WIP - needs tests etc, but I'd like to show it here as there are things to think about, e.g...

- Is the URI scheme okay, or do we want to use something slightly different than `containers/image`?
- Happy having this all implemented in a single `OCIConveyerPacker` - or split it out further? Since `containers/image` copy function brings everything to a temp oci location, there's no difference past the 'fetch' step.


bauerm97:
Hey DT looking good so far. On those two questions:

1. URI scheme looks great. I think it's good whenever possible to follow existing naming schemes if it doesn't matter to us.
2. Personally, I'm fine having this all within one data type as long as it makes sense. It looks like this definitely makes sense!

```
src/pkg/build/assembler_sif_test.go:25:10: undefined: DockerConveyorPacker
src/pkg/build/conveyorPacker_oci_test.go:23:9: undefined: DockerConveyor
```

Looks like a few test files need to be updated still! Quick question myself, since this is tagged WIP, what work is left to be done here?

dtrudg:
Hi @bauerm97 - thanks for the feedback.

With regard to WIP / what's to do.... mainly work out how to test the different sources sanely. For now, I may create archives of alpine and put them up at a public drive URL or something, which the tests can retrieve and then build. 

There's also a `docker-daemon` transport support by the library. This would allow import of an image directly from the docker daemon running on the same host as Singularity. It'd be nice to have, but I came across an issue with dependencies, so it's not in there yet - hope to have it.

bauerm97:
@dctrud Will you ping me when it's ready for review?

dtrudg:
@bauerm97 - yep. need to add tests and then it should be good. Will be before end of week.

dtrudg:
Tests have been added to cover the OCI sources. I placed some `docker save` and oci archives on S3 to allow this. There is a test that covers pulling directly from a docker daemon. It will run if `docker ps` succeeds, and is not a failure if docker is not available.

src/pkg/build/conveyorPacker_oci_test.go
```diff
@@ -0,0 +1,187 @@
...
+	if err != nil {
+		t.Fatalf("Could not download oci archive test file: %v", err)
+	}
+	defer os.Remove(archive)
+
+	// We need to extract the oci archive to a directory
+	// Don't want to implement untar routines here, so use system tar
+	dir, err := ioutil.TempDir("", "oci-test")
+	if err != nil {
+		t.Fatalf("Could not create remporary directory: %v", err)
```
ikaneshiro: Should probably read "temporary"

ikaneshiro (approved):
Looks good!

PR #1666, opened 2018-06-18
Add support for --cleanenv command

bauerm97:

This PR adds rudimentary support for `--cleanenv` flag in a very trivial manner. @cclerget Maybe have a smarter way to do this? For now this works.

For anybody, does this fully implement `--cleanev`? Let us know if it falls short somewhere!


cclerget (approved):
Looks good Michael ! At first look, I thought to put this into scontainer CheckConfig, but this is really not a critical option, so doing that on CLI side is fine.

PR #1667, opened 2018-06-19
Support read-only binds with user namespaces

DrDaveD:

This PR removes the check for not using a user namespace before remounting as read-only.  It works fine with user namespaces and there's not a reason to avoid the read-only mount in that case.  We have a case where we want to bind mount read-only, and want it to work also with user namespaces.

None

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/src/lib/runtime/mounts/userbinds/userbinds.c
+++ b/src/lib/runtime/mounts/userbinds/userbinds.c
@@ -139,18 +139,14 @@ int _singularity_runtime_mount_userbinds(void) {
                 ABORT(255);
             }
             if ( read_only ) {
-                if ( singularity_priv_userns_enabled() == 1 ) {
-                    singularity_message(WARNING, "Can not make bind mount read only within the user namespace: %s\n", dest);
-                } else {
-                    singularity_message(VERBOSE, "Remounting %s read-only\n", dest);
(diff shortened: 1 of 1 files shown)


DrDaveD:
Rebased on development-2.x at 2.6.0-rc2 tag.

I haven't heard any response to this one yet, what do core team people think about it?

cclerget:
I don't see any concerns about this change, in v3 all mount points are remounted even with user namespace excepting for `/proc` and `/sys`. Ping @GodloveD to see what he thinks about it and if that could be integrated in 2.6 release.

gmkurtzer:
The original intention was because the user namespace is not implemented equally across all kernels and on some kernels it was causing failures as MS_REMOUNT was not allowed unprivileged.

If this feature is necessary, I would suggest either a non-ABORT failure mode for the remount in read only if (and only if) running within a user namespace and without privs.

DrDaveD:
Hey @gmkurtzer , it's good to know that history, but if readonly mount is not supported, isn't it better to abort in that situation?  The user can always remove the ":ro" option from the bind command if they're OK with that and try again.  I think it is better to not continue if the user asks for readonly but singularity is not able to provide it.  Instead it would expose something to writing that the user does not want exposed.  In my opinion, that's a security flaw (although it is a situation that is not likely to affect very many users).

DrDaveD:
@godloved Please reopen this against release-2.6, I think it is still viable bug fix.  I have not rebased & force pushed so I think you should be able to do it.

PR #1692, opened 2018-06-27
Stop there when dep ensure fails.

yhcote:
Make sure that the build stops if "make dep" fails.


jmstover (approved):

yhcote:
Oh wow, that was prompt @jmstover !! Thanks.

PR #1694, opened 2018-06-27
Development 3.0: shared loop devices support

cclerget:

This PR allow multiple singularity processes to share loop device in order to optimize loop usage and kernel cache. Especially useful for MPI jobs.

**How it works ?**

An exclusive file lock is set on `/dev` primarily to avoid race condition while creating `/dev/loopX` blocks and secondly when a process try to acquire lock, the process will block until lock is released, once released process acquires the lock to block other processes, so it will open each loop devices and check if image to load is already loaded by checking inode/dev values and open mode are the same, if those conditions are met the process just return the loop device where image is already associated rather than associate image to a new loop device.


jmstover:
Re-open from `development-3.0` removal

jmstover (approved):

cclerget:
Close this one, done in #2499

PR #1700, opened 2018-06-28
Makeit dep

yhcote:
Seperate the patch and dep rules. And make sure dep runs whenever we touch go code.


cclerget (approved):
Looks good !

Issue #1680, opened 2018-06-22
Issue better error when there is not enough space on /tmp to build


GodloveD:
## Version of Singularity:

2.5.1

### Expected behavior

Get an error along the lines of:
```
There is not enough space on /tmp to stage building of a singularity image
```

### Actual behavior

The error looks more like:
```
Could not mount Singularity overlay: Invalid argument
```

### Steps to reproduce behavior

Try building, pulling, running, etc a big container from docker hub (like tensorflow) with a very limited amount of /tmp space.

carterpeel:
Hello,

This is a templated response that is being sent out to all open issues.  We are working hard on 'rebuilding' the Singularity community, and a major task on the agenda is finding out what issues are still outstanding.


**Please consider the following:**

1. Is this issue a duplicate, or has it been fixed/implemented since being added?
2. Is the issue still relevant to the current state of Singularity's functionality?
3. Would you like to continue discussing this issue or feature request?


Thanks,
Carter

pedroalvesbatista:
@GodloveD We're looking into the issue carefully, soon will bring to community and discuss ways to better solve as well address this. Thankyou for keeping the interest in the subject.

kmuriki:
Pending issues from the old repo are copied to the new repo (https://github.com/apptainer/apptainer/issues/1392) and cleaned from the old, retired repo.

PR #1703, opened 2018-06-28
POC of issue 1680

oddodaoddo:

This is a basic POC to fix the issue reported in #1680 on github. We try to estimate the size of the exploded archive before actually exploding the archive onto /tmp. This way we can prevent improper error messages. The PR is NOT READY to merge.

- Ref: #

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/src/docker-extract.c
+++ b/src/docker-extract.c
@@ -19,15 +19,6 @@
 #include "util/registry.h"
 #include "util/util.h"
 
-// get available space on a possible filesystem (any mounted path that supports call)
-long get_available_space_fs(char *path) {
-    struct statvfs stat;
-
-    if (statvfs(path, &stat) != 0)
-	return -1;
-    return stat.f_bsize * stat.f_bavail;
-}
-
 /* apply_opaque
  *  Given opq_marker as a path to a whiteout opaque marker
  *    e.g. usr/share/doc/test/.wh..wh..opq
@@ -258,7 +249,7 @@ int extract_tar(const char *tarfile, const char *rootfs_dir) {
     char *orig_dir;
     const char *pathname;
     int pathtype;
-
+    
     orig_dir = get_current_dir_name();
 
     /* Select which attributes we want to restore. */
@@ -385,7 +376,8 @@ int main(int argc, char **argv) {
     char *rootfs_dir = singularity_registry_get("ROOTFS");
     char *rootfs_realpath;
     char *tarfile = NULL;
-    long exploded_tar_size;
+    long exploded_tar_size, fs_avail_space;
+    struct statvfs stat;
 
     // Set UTF8 locale so that libarchive doesn't produce warnings for UTF8
     // names - en_US.UTF-8 is most likely to be available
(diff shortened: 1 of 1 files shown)


dtrudg:
Hi @oddodaoddo  - we already do 2 passes through the tar, 1 to handle whiteout files, 1 to extract it. Adding another whole pass isn't ideal here. I think the size check be implemented in the whiteout pass - since that doesn't extract anything.... so it can't hit out of space there.

oddodaoddo:
Hello, thanks! I have no issues with doing what you suggest - just wanted to know if the idea is correct. I will go back and re-implement and re-submit, if you agree. Thanks!

oddodaoddo:
Alright, I updated the PR to now get the estimated exploded archive size in apply_whiteouts. I changed the name of the function to reflect that fact since I hate functions that do multiple things as side-effects - if we are going to violate basic principles of programming, we might as well make it known ;)

New code passes all tests (one was failing previously due to initial PR).

Let me know what you think...

Thanks! Ognen

bauerm97:
> I hate functions that do multiple things as side-effects

I haven't yet had time to review the contents of this one, but I can tell you already that I really appreciate this ;) Naming things is hard business to get right!!

src/docker-extract.c
```diff
@@ -244,6 +259,7 @@ int extract_tar(const char *tarfile, const char *rootfs_dir) {
     const char *pathname;
     int pathtype;
 
+    printf("EXTRACT TAR NAME: %s\n", tarfile);
```
dtrudg: Anything that is a log message needs to be written out using the `singularity_message` function, so it can be silenced/shown by the set debug level.

dtrudg (changes requested):
Just one comment about changing a `printf` to a `singularity_message`

src/docker-extract.c
dtrudg: Probably this actually could be removed - since the outer wrappers for building etc. will log the tar which is going to be extracted.

oddodaoddo: So sorry, that was for my consumption only, I can't believe I left it. :)

oddodaoddo:
I noticed something while looking at some other code in the singularity code base - lines such as "char *tmpdir = soingularity_registry_get("TMPDIR") - does this change the contents of this PR then? Should I not assume that /tmp is what Singularity relies on for extracting layers into? Is that target configurable? Thanks!

dtrudg:
@oddodaoddo - ahh yes - sorry.... I'd just had a quick scan earlier when I saw the printf. docker-extract may not be extracting into `/tmp` so you should be checking the free space available for the filesystem that contains `*rootfs_dir = singularity_registry_get("ROOTFS");` - the docker layer won't always go into `/tmp` or even TMPDIR. If we are building a sandbox container it may go straight into a user specified final location.

oddodaoddo:
@dctrud : Thanks - but, are you sure about the last statement about ROOTFS vs TMPDIR? I have a VM I set up with 1G of /tmp space and if I fill it up with junk, the extraction will fail even though there is plenty of space on / (/tmp is set up as a separate partition/fs).

dtrudg:
@oddodaoddo - almost always the ROOTFS will be in TMPDIR. The situation where it is not is if you do a `singularity build -s mysandboxdir/ docker://ubuntu` - which creates a sandbox container. The docker layers are extracted directly into the `mysandboxdir/` wherever that is. 

Regardless, the ROOTFS is what `docker-extract` is using to get the destination for the extraction code - so that's where we should check.

oddodaoddo:
Ah, I understand! Perfect - I submitted the code; let me know what you
think. I removed the separate function to check fs space since it is a
one-off and will probably not be used anywhere else - I wrapped everything
into main for the sake of flow/clarity.

jmstover:
@oddodaoddo ... just noticed this PR is based on `master`. Is it possible to rebase it off `development-2.x` or `release-2.5` ?

oddodaoddo:
@jmstover : I opened a different PR against release-2.5. I can close the original PR. Would this work? Thanks.

jmstover:
It looks like they're linked:
`dctrud requested changes in #1703 `

So... I'd leave both open.

PR #1704, opened 2018-06-29
Use sylabs/sif new implementation for all SIF access in this tree.

yhcote:
Remove old references to sif C and use sylabs/sif to create and load containers.

This patch enabled the removal of a lot of unused C files.


src/pkg/build/assembler_sif.go
```diff
@@ -37,17 +93,14 @@ func (a *SIFAssembler) Assemble(b *Bundle, path string) (err error) {
 	mksquashfsCmd.Stderr = os.Stderr
 	err = mksquashfsCmd.Run()
 	if err != nil {
-		return err
+		return
 	}
+	defer os.Remove(squashfsPath)
```
cclerget: I think defer should be set before error check here to be sure squashfs temporary file is deleted, sif command could failed for some reasons and have already created temporary file

src/pkg/build/assembler_sif.go
```diff
@@ -6,23 +6,79 @@
...
+	return nil
+}
+
 // Assemble creates a SIF image from a Bundle
 func (a *SIFAssembler) Assemble(b *Bundle, path string) (err error) {
 	mksquashfs, err := exec.LookPath("mksquashfs")
 	if err != nil {
 		sylog.Errorf("mksquashfs is not installed on this system")
-		return err
+		return
```
cclerget, line 81: Is it intentional to not return err ?

src/pkg/build/assembler_sif.go
```diff
@@ -37,17 +93,14 @@ func (a *SIFAssembler) Assemble(b *Bundle, path string) (err error) {
 	mksquashfsCmd.Stderr = os.Stderr
 	err = mksquashfsCmd.Run()
 	if err != nil {
-		return err
+		return
```
cclerget: And here too ?

src/pkg/build/assembler_sif.go
```diff
@@ -37,17 +93,14 @@ func (a *SIFAssembler) Assemble(b *Bundle, path string) (err error) {
...
 
-	sifCmd := exec.Command("singularity", "sif", "create", "-P", squashfsPath, "-f", "SQUASHFS", "-p", "SYSTEM", "-c", "LINUX", path)
-	sifCmd.Stdin = os.Stdin
-	sifCmd.Stdout = os.Stdout
-	sifCmd.Stderr = os.Stderr
-	err = sifCmd.Run()
+	err = createSIFSinglePart(path, squashfsPath)
 	if err != nil {
-		return err
+		return
```
cclerget, line 102: same here ?

src/pkg/image/image.go
```diff
@@ -112,7 +112,9 @@ func Init(path string, writable bool) (*Image, error) {
 	if writable {
 		flags = os.O_RDWR
 	}
-	resolvedPath, err := filepath.EvalSymlinks(path)
+
+	abspath, _ := filepath.Abs(path)
```
cclerget: My bad ! I forgot but it may be useful to check error and return it

src/pkg/build/assembler_sif.go
```diff
@@ -6,23 +6,79 @@
...
+	}
+
+	return nil
+}
+
 // Assemble creates a SIF image from a Bundle
 func (a *SIFAssembler) Assemble(b *Bundle, path string) (err error) {
 	mksquashfs, err := exec.LookPath("mksquashfs")
 	if err != nil {
 		sylog.Errorf("mksquashfs is not installed on this system")
```
cclerget, line 80: why not have something like `return fmt.Errorf("mksquashfs is not installed on this system")` rather and let calling function display error ?

cclerget (changes requested):
Have some requests mainly for assembler_sif, otherwise looks good to me

src/pkg/build/assembler_sif.go
yhcote: it's a naked return. This is done when the return type is named. e.g., (err error)

src/pkg/build/assembler_sif.go
yhcote: same comment ;)

src/pkg/build/assembler_sif.go
yhcote: And same here too yes :)

src/pkg/build/assembler_sif.go
yhcote: good point but it's not part of my patch ? ;)

src/pkg/build/assembler_sif.go
cclerget: Oh I missed that ! Sorry !

src/pkg/build/assembler_sif.go
cclerget: Yes right ! :D

src/pkg/image/image.go
yhcote: ok no problem

src/pkg/build/assembler_sif.go
yhcote: true.

cclerget (approved):

PR #1705, opened 2018-06-29
Debootstrap ConveyorPacker

ikaneshiro:

fixing or enhancing.

- Ref: #

**Checkoff for all PRs:**

Attn: @singularityware-admin


src/pkg/build/conveyorPacker_debootstrap.go
```diff
@@ -0,0 +1,79 @@
...
+		return fmt.Errorf("debootstrap is not in PATH... Perhaps 'apt-get install' it?%v", debootstrapPath)
+	}
+
+	c.tmpfs, err = ioutil.TempDir("", "temp-debootstrap-")
+	if err != nil {
+		return
+	}
+
+	//Dont know what this does...
+	//REQUIRES=`echo "${INCLUDE:-}" | sed -e 's/\s/,/g'`
```
jmstover: This is to set an environment variable `INCLUDE` which is a space separated list of packages... i.e.
```
INCLUDE="sshd syslinux gcc"
```
And that variable, gets turning into a comma separated value that was included in the `debootstrap` command

src/pkg/build/conveyorPacker_debootstrap.go
```diff
@@ -0,0 +1,79 @@
...
+	requires := ""
+
+	//run debootstrap command
+	//$DEBOOTSTRAP_PATH --variant=minbase --exclude=openssl,udev,debconf-i18n,e2fsprogs --include=apt,$REQUIRES --arch=$ARCH '$OSVERSION' '$SINGULARITY_ROOTFS' '$MIRRORURL'
+	cmd := exec.Command(debootstrapPath, `--variant=minbase --exclude=openssl,udev,debconf-i18n,e2fsprogs`, `--include=apt,`+requires, `--arch=`+runtime.GOARCH, OSVersion, c.tmpfs, MirrorURL)
```
jmstover: What is the output of `runtime.GOARCH` ?? Is it what Debian expects? i.e. the `amd64`, `arm64`, etc...

src/pkg/build/conveyorPacker_debootstrap.go
ikaneshiro: Does include ever change or can we just hardcode these values?

src/pkg/build/conveyorPacker_debootstrap.go
ikaneshiro: I think so. Check out the list under "optional environment variables"
https://golang.org/doc/install/source#environment

src/pkg/build/conveyorPacker_debootstrap.go
jmstover: Yeah... it was a environment variable that the user could set when building with a `debootstrap` source. They could just export that variable and the build would include other packages. Or if not set, it `debootstrap` just installs the "bare" system.

Then usually you'd do a `apt-get install [pkg]` in the `%post` section

src/pkg/build/conveyorPacker_debootstrap.go
ikaneshiro: It outputs in that format and a list of the potential values can be found here https://github.com/golang/go/blob/master/src/go/build/syslist.go

src/pkg/build/conveyorPacker_debootstrap_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"os"
+	"testing"
+)
+
+func TestDebootstrapConveyor(t *testing.T) {
+
+	//test must be run as root
+	if os.Getuid() != 0 {
```
jmstover, line 16: Just a `return` here? Isn't not being root a fatal error for Debootstrap? Or just not on the "test" part?

jmstover (dismissed):
The structure of this looks good to me. Some comments added, but overall it looks good.

src/pkg/build/conveyorPacker_debootstrap.go
```diff
@@ -0,0 +1,155 @@
...
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"runtime"
+	"strings"
+
+	"github.com/singularityware/singularity/src/pkg/sylog"
+)
+
+//"github.com/singularityware/singularity/src/pkg/image"
```
yhcote: Cruft to remove

src/pkg/build/conveyorPacker_debootstrap.go
```diff
@@ -0,0 +1,155 @@
...
+
+// Get downloads container information from the specified source
+func (c *DebootstrapConveyor) Get(recipe Definition) (err error) {
+
+	c.recipe = recipe
+
+	//check for debootstrap on system(script using "singularity_which" not sure about its importance)
+	debootstrapPath, err := exec.LookPath("debootstrap")
+	if err != nil {
+		return fmt.Errorf("debootstrap is not in PATH... Perhaps 'apt-get install' it?%v", debootstrapPath)
```
yhcote: it's usually nice to add the original error to the error string.

example : return fmt.Errorf("....: %s", err)

src/pkg/build/conveyorPacker_debootstrap.go
```diff
@@ -0,0 +1,155 @@
...
+
+// Pack puts relevant objects in a Bundle!
+func (cp *DebootstrapConveyorPacker) Pack() (b *Bundle, err error) {
+	b, err = NewBundle("")
+	if err != nil {
+		return
+	}
+
+	//remove old rootfs
+	os.RemoveAll(b.Rootfs())
```
yhcote, line 95: Would there be a place where b.Rootfs/c.tmpfs could be removed when the action is done running. In other words, is that folder gone when singularity returns or is there more actions to perform on that directory before It can be cleared ?

yhcote (changes requested):
Just a few things and question.

src/pkg/build/conveyorPacker_debootstrap.go
jmstover: I believe that debootstrap requires an _empty_ rootfs to work.

*Edit:*
Wait... this is at a different point isn't it. Would this just to be sure that the `b.Rootfs()` given didn't already exist?

yhcote: Oh Yeah, when I saw the RemoveAll() I took the opportunity to say that there's no removal of the directory after the facts. Which makes me think that when this is done running, there will be a left over directory. Maybe I am wrong.

jmstover: Well... I wouldn't have the cleanup of that in the specific build files anyway. That would be back wherever it gets setup, and would be done by that code *after* the specific provisioner (Conveyor?) gets called and return successful.

yhcote: something to keep in mind...

ikaneshiro: So we need to have a tmpfs for the bundle after the conveyorPacker is done running. My intention is for the assembler to clean up the bundle after it if finished building from it

src/pkg/build/conveyorPacker_debootstrap_test.go
ikaneshiro: As far as i know our tests with travis dont run with sudo so I did this to pass CI. I think it would be better to test this at a higher level where we call "sudo singularity build..." in a script

src/pkg/build/conveyorPacker_debootstrap.go
yhcote: perfect.

src/pkg/build/conveyorPacker_debootstrap.go
ikaneshiro: Ah I used the wrong variable

src/pkg/build/conveyorPacker_debootstrap.go
```diff
@@ -97,32 +97,32 @@ func (cp *DebootstrapConveyorPacker) Pack() (b *Bundle, err error) {
...
 	return b, nil
 }
 
 func (cp *DebootstrapConveyorPacker) insertBaseEnv(b *Bundle) (err error) {
 	if err = makeBaseEnv(b.Rootfs()); err != nil {
-		sylog.Errorf("%v", err)
+		return
 	}
-	return
+	return nil
```
yhcote: This was fine with "return", since no error occurred err should still be nil going out.

yhcote (dismissed):
Looks good to me!

dtrudg (dismissed):
:+1: lgtm and can build a Debian container using debootstrap on RHEL7 and Fedora28 hosts.

jmstover (dismissed):

bauerm97 (approved):

GodloveD:
Yays!

PR #1668, opened 2018-06-19
3.0 build script support / image build engine- WIP

bauerm97:
**Description of the Pull Request (PR):**

Together with @ikaneshiro

PR #1710, opened 2018-06-30
3.0 - Build script support (image build engine)

bauerm97:

Updated and cleaned up version of #1668. 

- Rebased onto upstream
- Cleaned up git commit history 
- Removed code referencing C SIF and Image libs
- Fix issues regarding permissions of calling user
- Prevent SUID escalation for imgbuild engine


src/cmd/singularity/cli/sif.go
```diff
@@ -28,8 +28,8 @@ func init() {
 	SifCmd.Flags().SetInterspersed(false)
 	SingularityCmd.AddCommand(SifCmd)
 	// Create
-	SifCmd.AddCommand(SifCreate)
```
yhcote: I think we can graciously remove anything related to the create command here. I agree that it should not be there anymore. There should not be two different path from singularity to create a SIF image and "build" is the right command to accomplish that.

src/cmd/singularity/cli/sif.go
```diff
@@ -55,10 +55,25 @@ var SifCmd = &cobra.Command{
...
+		}
+		if err := sif.Wait(); err != nil {
+			sylog.Fatalf("sif failed: %v\n", err)
+		}
+
+	},
+	DisableFlagParsing: true,
 }
 
-// SifCreate sif create cmd
```
yhcote: This too should be completely removed as opposed to commented :)

yhcote (dismissed):
Nothing fancy but removing unused lines are satisfying ;)

I would also make sure there's no more references to "singularity sif create" from comments or docs for the cli.

mlocal/frags/Makefile.stub
```diff
@@ -67,7 +67,7 @@ $(wrapper_OBJ): $(libstartup) $(SOURCEDIR)/src/runtime/startup/c/wrapper.c
 $(wrapper): $(wrapper_OBJ) $(libstartup)
 	@echo " LD" $@
 	@mkdir -p $(@D)
-	$(V)$(CC) $(LDFLAGS) -o $@ $(wrapper_OBJ) -Wl,--start-group -L$(BUILDDIR)/lib -lruntime -lstartup -Wl,--end-group -luuid -lpthread $(WRAPPER_LIBS)
+	$(V)$(CC) $(LDFLAGS) -o $@ $(wrapper_OBJ) -Wl,--start-group -L$(BUILDDIR)/lib -lruntime -lstartup -Wl,--end-group -luuid -lpthread -lgpgme $(WRAPPER_LIBS)
```
yhcote, line 70: What is the gpgme for ? Do we import a package that requires it ?

bauerm97: Yes, this is an unfortunate side effect of using `build.Definition` as a member field of `src/runtime/engines/imgbuild/config.EngineConfig`. `build` package needs `-lgpgme`. Maybe this means the `Definition` type should be moved to a buildtypes subpackage?

src/runtime/engines/singularity/create.go
```diff
@@ -98,118 +187,127 @@ func (engine *EngineOperations) CreateContainer(pid int, rpcConn net.Conn) error
...
-		return fmt.Errorf("mount /dev failed: %s", err)
+	// rootfs is an image
+	return mountImage(rpcOps, mnt)
+}
+
+// mount any generic mount (not loop dev)
+func mountGeneric(rpcOps *client.RPC, mnt specs.Mount) error {
+	flags, opts, _ := mount.ConvertOptions(mnt.Options)
+	optsString := ""
+	for _, opt := range opts {
```
cclerget: you can use strings.Join here ;)

src/runtime/engines/singularity/create.go
```diff
@@ -98,118 +187,127 @@ func (engine *EngineOperations) CreateContainer(pid int, rpcConn net.Conn) error
...
+	_, err := rpcOps.Mount(mnt.Source, mnt.Destination, mnt.Type, flags, optsString)
+	return err
+}
+
+// mount image via loop
+func mountImage(rpcOps *client.RPC, mnt specs.Mount) error {
+	flags, opts, iopts := mount.ConvertOptions(mnt.Options)
+
+	optsString := ""
+	for _, opt := range opts {
```
cclerget: here too

cclerget (changes requested):
Looks good Michael ! Just suggested 2 minor changes

cclerget (approved):

PR #1713, opened 2018-07-01
travis: adding debian and opensuse CI tests

c-ney:

This is a follow up (or more like a replacement) to #1204. It includes some refactoring of the CI stuff and introduces debian 8/9 (by @olifre with some commit amendments by me, hope this is okay) and opensuse 42.3/15.0 tests. The latter are not enabled for travis yet and still somewhat of a WIP. Running locally, both opensuse versions currently fails at `39-deffile.sh` due to the issue from #1074.

The first commit for autoconf is identical to #1712 (didn't rebase yet). The debian changelog "fix" should likely go into a separate issue/pr.

In `centos_run` --now renamed to `docker_run`-- I did not fully understand the implications of `/usr/sbin/init` for centos 7. Debian and opensuse do not have it, so I just did some copy&paste from centos 6 and removed the `centos6_setup` line. Would be great if someone could elaborate.

Thanks for any feedback!

- Ref: #

**Checkoff for all PRs:**

Attn: @singularityware-admin


.travis/debuild_test
```diff
@@ -0,0 +1,48 @@
...
+mk-build-deps
+mv singularity-container-build-deps_*.deb ..
+gdebi -n ../singularity-container-build-deps_*.deb
+
+# Build package
+debuild -us -uc
+
+echo "Directory containing the build .deb-files:"
+ls -la ..
+gdebi -n ../singularity-container_*.deb
```
jmstover, line 26: Instead of adding `gdebi` ... could you just use `dpkg -i` to install the `.deb` file?

olifre: `dpkg` does not resolve dependencies, while `gdebi` does. So it's essential to use `gdebi` here to check that the dependencies expressed in the packaging scripts are actually complete.

olifre: > dpkg does not resolve dependencies, while gdebi does.

Or rather: `dpkg` _does_ resolve dependencies, of course, but does not install them (it only complains if things are not installed yet) :wink:.

.travis/debuild_test
```diff
@@ -0,0 +1,48 @@
...
+curl -fsSL https://download.docker.com/linux/$(. /etc/os-release; echo "$ID")/gpg | apt-key add -
+add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/$(. /etc/os-release; echo "$ID") $(lsb_release -cs) stable"
+apt-get update
+apt-get -y install docker-ce
+
+# run the test suite
+apt-get -y install libtool sudo e2fsprogs
+sed -i 's,^prefix=.*,prefix=/usr,' test.sh
+sed -i 's,^sysconfdir=.*,sysconfdir=/etc,' test.sh
+sed -i 's,^localstatedir=.*,localstatedir=/var,' test.sh
```
jmstover:
```
prefix=
exec_prefix=
libexecdir=
sysconfdir=
localstatedir=
bindir=
```

Should all be set to what was used on the `./configure` line (`test.sh` gets generated from `test.sh.in`). There's hard coded paths that need to match up here because of the SUID bits.

.travis/debuild_test
c-ney: I just found that you can apparently do `dpkg -i pkg.deb` followed by `apt-get -fy install` [1]. Will try this later.

[1] https://superuser.com/a/537371

jmstover: If you're going to do that, couldn't you do:

```
apt-get install ../singularity-container_*.deb
```

And then let `apt-get` pull the dependencies?

.travis/debuild_test
olifre: Oh yes! It seems this is not needed anymore. As of #1204, when I added those lines last year, `test.sh` was not correctly taking over what was passed to `configure`, but it seems that's corrected now. @c-ney : Just removing those `sed` lines should work :wink:.

.travis/debuild_test
olifre: Indeed, both of these should also work - of course `apt-get -fy install` may do more than wanted / expected, and `apt-get install ../singularity-container_*.deb` is an officially undocumented feature. The man-page says `Each package is a package name, not a fully qualified filename` when talking about parameters to `install`. So this unofficial feature may be broken in any future version of `apt-get`.

.travis/debuild_test
c-ney: Now that you mention this, I remember having to put in an opensuse specific line in the past, but forgot to add it here and it still works. Will remove them from ~~both~~ `debuild_test` ~~and `rpmbuild_test`~~

Edit: No, this doesn't work for the rpm test due to different `./configure` arguments in the spec file...

.travis/debuild_test
jmstover: Hrmm.... If that's the case, then I guess I would stick with installing `gdebi` and using it, since the `apt-get -fy install` may have unintended consequences, and `apt-get install package_name.deb` is a unsupported "feature."

I would have just thought apt-get could install from a local `.deb` package file as if it was pulling from sources.list, and not thought anything of it.

.travis/debuild_test
olifre: Ahh! I remember why this was needed! It works fine on Debian, since `debuild`runs "in-place" and executes `configure` with full parameters. 
`rpmbuild`, though, runs in a sandbox, and calls `configure` there, and not where we actually started it. 
So either we have to steal `test.sh` from `/root/rpmbuild/BUILD/singularity-2.5.1.999/`, or replicate the parameters `rpmbuild` is using further up (i.e. before `make dist`) when `configure` is called to make `make dist` actually work. 
I'm unsure which is nicer...

jmstover: Right... and this is on the debian one... I'm working on a comment on the rpmbuild one. ;)

.travis/rpmbuild_test
```diff
@@ -2,31 +2,41 @@
...
 
+install_test_deps
+
 # run the test suite
-yum install -y libtool sudo which e2fsprogs
 sed -i 's,^prefix=.*,prefix=/usr,' test.sh
 sed -i 's,^sysconfdir=.*,sysconfdir=/etc,' test.sh
 sed -i 's,^localstatedir=.*,localstatedir=/var,' test.sh
-useradd -u 1000 testuser
+useradd -u 1000 --create-home testuser
```
jmstover: On these sed lines, you should use the values from:
```
rpm -E '%{_prefix}'
rpm -E '%{_sysconfdir}'
rpm -E '%{_localstatedir}'
```

That will pull the expanded macro value out in case it ever changes... for some reason.

*Edit:*
So something like:
```
sed -i "s@^prefix=.*@prefix=$(rpm -E '%{_prefix}')@" test.sh
```

olifre: Good idea! I'm all for it :+1:

c-ney: I also stumbled upon this:

```sh
bash -c "$(rpm --eval '%configure')"
make dist
```

...but I don't think anyone would configure like this before a make dist... so, whatever you prefer!

jmstover: I've ... seen it. Usually I've needed to rip out the normal `%configure` macro, and use my own options, and/or override the normal `_libexecdir`, `_prefix`, etc... 

But, yeah... I've sadly seen that before. :/

jmstover (approved):
This all appears good...

GodloveD (changes requested):
This looks good to me.  Just a few asks.  Please add your name to the contributor's file @c-ney, and document these upgrades in the changelog under 2.6.0.  Thanks!

GodloveD:
@c-ney I'd like to include this in 2.6.0.  And I'd like to get a release candidate for 2.6.0 out pretty soon.  Do you think you can address these changes?  Thank you.

c-ney:
@GodloveD Yes, I'll try to do it this evening after work, but I'm 100% positive to get it done on Sunday.

Will also fix some minor stuff (like the empty file and a wrong comment). Should I rebase on another branch or is release-2.5 okay? There is also enabling the opensuse tests left, but those will fail atm.

GodloveD:
Thanks @c-ney development-2.x would actually be the correct branch to rebase on.  Those two branches are pretty similar so hopefully the rebase shouldn't be too difficult.  😄

c-ney:
@GodloveD I did some final finishing touches and, once the latest travis run succeeds, would consider this in a decent state.

I left two debian changelog commits in, because those are needed for the debian tests to finish. But there is also a separate PR (#1736) for them.

GodloveD (approved):
lgtm.  @jmstover do you like these new changes too?

debian/changelog
```diff
@@ -1,12 +1,14 @@
-singularity-container (2.5.1) unstable; urgency=high
+singularity-container (2.5.1-1) unstable; urgency=high
```
jmstover, line 1: Does the revision (i.e. the `-1`) *need* to exist? I have done a build on Debian 8/9 and without putting one in it defaulted to a `-1` revision without. I think...  Pretty sure I didn't make any changes my first few build attempts.

IIRC, there was also a comment made ages ago that the revision didn't need to be put there as the package maintainer would be modifying it...

c-ney: At least in the default configuration, building a debian package breaks at the `dpkg-source` step inside `debuild`. Using this container

```
bootstrap: docker
from: debian:8

%environment
    LC_ALL=C

%post
    apt-get update
    apt-get -y dist-upgrade
    apt-get -y install devscripts build-essential lintian gdebi-core git
```

as the environment gives me

```
Singularity deb8.simg:~/devel/extern/singularity> git status
On branch development-2.x
Your branch is up-to-date with 'origin/development-2.x'.
nothing to commit, working directory clean
Singularity deb8.simg:~/devel/extern/singularity> dpkg-source -b .
parsechangelog/debian: warning: singularity_clean/debian/changelog(l9): found start of entry where expected more change data or trailer
LINE: singularity-container (2.5.0) unstable; urgency=high
parsechangelog/debian: warning: singularity_clean/debian/changelog(l9): found eof where expected more change data or trailer
dpkg-source: error: can't build with source format '3.0 (quilt)': non-native package version does not contain a revision
```

This goes away when you add the revision and the maintainer.

```
Singularity deb8.simg:~/devel/extern/singularity> dpkg-source -b .
dpkg-source: info: using source format `3.0 (quilt)'
dpkg-source: info: building singularity-container using existing ./singularity-container_2.5.1.orig.tar.gz
dpkg-source: info: building singularity-container in singularity-container_2.5.1-1.debian.tar.xz
dpkg-source: info: building singularity-container in singularity-container_2.5.1-1.dsc
```

It all comes down to what source format is set in `debian/source/format`. `3.0 (quilt)` for non-native wants the revision, `3.0 (native)` does not.

I'm no expert in debian package building, so maybe this is a non-issue due to some convention as you mentioned.

Related PR: #1736

GodloveD: I'm fine with this.  Either way the maintainer has to edit the file.

GodloveD:
Thanks @c-ney, @olifre, and @jmstover!

PR #1723, opened 2018-07-04
Update the version to use for sylabs/sif.

yhcote:
Sif has an update format. Use this new version for V3 development.


cclerget (approved):
:+1:

PR #1728, opened 2018-07-05
Development 3.0: Engines operations

cclerget:

This PR adds `IsAllowSUID` engine operation, so each engines defines whether SUID workflow is allowed or not. Rename `CheckConfig` into `PrepareConfig`


src/runtime/engines/imgbuild/engine.go
```diff
@@ -30,8 +30,8 @@ func (e *EngineOperations) Config() config.EngineConfig {
 	return e.EngineConfig
 }
 
-// CheckConfig validates EngineConfig setup
-func (e *EngineOperations) CheckConfig() error {
+// PrepareConfig validates/prepares EngineConfig setup
+func (e *EngineOperations) PrepareConfig() error {
 	e.CommonConfig.OciConfig.SetProcessNoNewPrivileges(true)
 
 	if syscall.Getuid() != 0 {
```
bauerm97, line 37: Is this check for `syscall.Getuid() != 0` still necessary now that we have `IsAllowSUID()` function call? I added this check to error out if user is not root.

bauerm97 (changes requested):

src/runtime/engines/imgbuild/engine.go
cclerget: Not necessary but that could help to display an explicit error message rather than fail later at container creation step during first mount

bauerm97 (approved):

Issue #719, opened 2017-06-03
bad variable name with singularity import


jokedurnez:
## Version of Singularity:
2.3

### Expected behavior
I'm trying to create a singularity image from a docker container:
```
[jdurnez@sh-101-58]$ singularity create --size 20000 $SINGULARITY
[jdurnez@sh-101-58]$ singularity import $SINGULARITY docker://joke/cnp_task_analysis
Docker image path: index.docker.io/joke/cnp_task_analysis:latest
Cache folder set to /home/users/jdurnez/.singularity/docker
[32/32] |===================================| 100.0%
Importing: base Singularity environment
...
Importing: /home/users/jdurnez/.singularity/metadata/sha256:dbda4988c86376f77fb0ebfc4aadf6f5087acc58bf920df723ea935a197bfd21.tar.gz
```

### Actual behavior
It seems like the singularity container is generated, but when I try to access it:
```
[jdurnez@sh-101-58]$ singularity shell $SINGULARITY
Singularity: Invoking an interactive shell within container...

/.singularity.d/actions/shell: 35: export: /opt/mcr/v85/: bad variable name
```

However, with docker2singularity, it works.
```
[ec2-user@ip-172-31-3-3 CNP_task_analysis]$ docker run -v /var/run/docker.sock:/var/run/docker.sock -v /home/ec2-user/singularity:/output --privileged -t --rm singularityware/docker2singularity joke/cnp_task_analysis
...
(9/9) Moving the image to the output folder...
 11,954,815,008 100%   61.77MB/s    0:03:04 (xfr#1, to-chk=0/1)
[ec2-user@ip-172-31-3-3 singularity]$ singularity shell joke_cnp_task_analysis-2017-06-03-1246c19080c1.img
Singularity: Invoking an interactive shell within container...

ec2-user@ip-172-31-3-3:~/singularity$ python -c "import nipype"
ec2-user@ip-172-31-3-3:~/singularity$
```

vsoch:
I've been trying for a few hours to download the layers (albeit my internet is awful). How big of an image is this? I can go to campus later (with a better connection) to test. Based on the error message, here are my first gut feelings about what might be going on. 1) is that somewhere in the environment you might have something that looks like this:

```
export PATH=/bla/bla:$PATH
```
[...]

PR #1729, opened 2018-07-05
Escape Docker/OCI ENV, CMD, ENTRYPOINT

dtrudg:
This PR adds shell escape quoting for the ENV, CMD and ENTRYPOINT declarations in a Dockerfile, to make it work right once they reach our shell script driver/metadata.

**ENV**

This shouldn't be controversial - every ENV is wrapped in double quotes and escaped when written into `/.singularity.d/env/10-docker2singularity.sh` - e.g.

```
Singularity> cat /.singularity.d/env/10-docker2singularity.sh
#!/bin/sh
export PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
export bob="Hello\$n\"Bob"
```

This has long been broken in the 2.x series, see issues e.g. #719

**ENTRYPOINT CMD**

The current 2.x branch doesn't correctly combine ENTRYPOINT/CMD, or do shell escaping of these when creating runscript. 3.0 currently *does* combine ENTRYPOINT/CMD but needs this change to escape them for proper behavior in complex situations.

Consider the Dockerfile....

```
FROM alpine:latest
ENTRYPOINT ["/bin/sh", "-c", "echo EP && \"$0\" \"$@\""]
CMD ["echo", "Hello"]
```

The Entrypoint always runs and should start an sh instance which....

- should `echo EP` always
- should then `echo hello` if no args were supplied to run
- should run the command given by args to run, if they were supplied

Correct behaviour of `singularity run` is:

```
01:12 PM $ singularity run test.sif
EP
Hello
```

Correct behaviour of `singularity run test.sif ls` is:

```
﻿01:14 PM $ singularity run test.sif ls
EP
bin          environment  home         media        proc         run          singularity  sys          usr
dev          etc          lib          mnt          root         sbin         srv          tmp          var
```

This *maybe* slightly controversial as the `exec` as the final line of the runscript is changed to an `eval` to make it work. I have looked at this for about 4 hours, but can't see any way to use `exec` and have it working here under POSIX sh.... but I may be wrong :-)


jmstover (dismissed):

dtrudg:
Noting here as came up in conversation around this PR that with the `NAME=VAL` format we get from the OCI config struct for env vars, `VAL` is 'raw', representing the value that the env var must be set to. it will not be quoted already if there are spaces etc. Therefore our wrapping in double quotes when we export in the shell script that sets env vars is preserving the value, not modifying it.

here's an example... given the Dockerfile

```
FROM alpine:latest
ENV message Hello World
```

you get this in the oci json....

```
"config": {
    "Env": [
      "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
      "message=Hello World"
    ],
...
```
and confusingly you can also do in the Dockerfile...

```
FROM alpine:latest
ENV message="Hello World"

```
and your OCI config will still be....

```
"Env": [
      "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
      "message=Hello World"
    ],
```

(since the `ENV NAME=VAL` format for ENV in docker does consume quotes when the Dockerfile is processed, but the other `ENV NAME VAL` form doesn't)

ArangoGutierrez:
just as a comment 
on the copyright header
```
LICENSE => LICENSE.md
```

ArangoGutierrez:
DO NOT MERGE YET

dtrudg:
@ArangoGutierrez - can we not merge this and file a bug to get this okay with the other situations? I'm concerned this has been sitting here since July - and at that point the action handling didn't exist... so the fact it doesn't work on those is out of scope right now here.

ArangoGutierrez:
@jmstover and I are working on a fix to enable both `build/actions` so we can proceed with the merge

dtrudg:
okay - please be aware just found an issue with the `pull` part of things (well @GodloveD did really).

https://github.com/singularityware/singularity/issues/2018

If things are going to change fixing the above then I think getting the ENV on the working build portion would be a good step, to make a little progress here.

jmstover (approved):

ArangoGutierrez (approved):

bauerm97 (changes requested):
In car right now, but this needs a quick polish where it's touching the temp cached image handling. I'll write a review once I make it home in 45 min

src/cmd/singularity/cli/actions.go
```diff
@@ -91,10 +91,16 @@ func replaceURIWithImage(cmd *cobra.Command, args []string) {
 		sylog.Fatalf("That didn't work %v", err)
 	}
 
-	name := split[1]
-	imgabs := cache.OciTempImage(sum, name)
+	var ociRef string
```
bauerm97: This belongs in a different commit. I will fix properly and commit and make a PR

src/pkg/client/cache/oci.go
```diff
@@ -31,6 +32,12 @@ func OciTemp() string {
 func OciTempImage(sum, name string) string {
 	updateCacheSubdir(filepath.Join(OciTempDir, sum))
 
+	name = strings.Replace(strings.TrimPrefix(name, "//"), "/", "_", 3)
```
bauerm97: Again, this belongs in a separate commit & PR.

The images aren't being named for uniqueness, but rather for user readability. The unique identifier is the SHA name of the directory which the image lives in, and the name of the image should be in the most readable form possible. I've already committed a patch locally which I'll be submitting as a PR

bauerm97 (dismissed):
I think the contents of the original PR are good to merge, but I want us to break the habit of pushing unrelated changes into PRs together. It's becoming a big time sink to do code review when we have consistently disjoint PR's appearing. I think this echoes @dctrud 's sentiment about just making little steps at a time.

I'd like to see a separate PR created from `e63c901` (or change this PR's branch to point to `e63c901`) and I'll approve that commit from there.

dtrudg:
As @bauerm97 has stated above "I think the contents of the original PR are good to merge" I have reverted the previous commit dc5de5d and will dismiss the change request, and merge.

PR #1730, opened 2018-07-05
Development 3.0: selinux

cclerget:

This PR adds SELinux support and wrap all modules in package security with a `Configure` function.


jmstover:
Re-open from `development-3.0` removal

src/pkg/security/apparmor/apparmor_linux.go
```diff
@@ -13,19 +13,20 @@ import (
 	"os"
 )
 
-// LoadProfile write apparmor profile in /proc/self/attr/exec
-func LoadProfile(profile string) error {
+// Enabled returns if apparmor is whether enabled/supported or not
```
yhcote: // Enabled() returns whether apparmor is enabled/supported or not.

src/pkg/security/apparmor/apparmor_linux.go
```diff
@@ -13,19 +13,20 @@ import (
...
+func Enabled() bool {
 	data, err := ioutil.ReadFile("/sys/module/apparmor/parameters/enabled")
 	if err == nil {
```
yhcote: ...
```
        if err == nil && len(data) > 0 && data[0] == 'Y' {
                return true
        }
        return false
}
```

yhcote (approved):
looks good to me

PR #1735, opened 2018-07-07
3.0 wrapper fix and gpgme enhancement

bauerm97:

This PR addresses the issue of attempting to use Go's built in `plugin` package when compiling the Go code as a `c-archive` and using a standard C linker. Instead, we can compile the wrapper using `go build` and include the `wrapper.c` code as cgo. This serves as a precursor to the plugin system framework being introduced

Thanks to @cclerget for implementing this

This also removes the dependency on `-lgpgme` in the build process by adding `containers_image_openpgp` go build tag.


mlocal/frags/release_opts.mk
```diff
@@ -1,2 +1,2 @@
-CFLAGS += -O2
+CFLAGS += -O2 -fPIC
```
cclerget: My bad, common_opts.mk is the appropriate location for `-fPIC` flag, so debug/release build have same options

cclerget (changes requested):

bauerm97:
@cclerget Moved

cclerget (approved):

PR #1740, opened 2018-07-10
WiP: Port 2.x Tests

tri-adam:
**Work in progress PR - feedback welcome!**

Re-writing acceptance tests present in 2.x for 3.0, leveraging go's built-in framework. Summary of changes so far:

- `01-sanity.sh` replaced by `cli_test.go`/`TestMain()`. If sanity checks fail, the entire suite of acceptance tests will fail.
- `05-python-units.sh` removed (unit tests are now performed in Go).
- `10-help.sh` replaced by `help_test.go`.
- `15-selftest.sh` replaced by `cli_test.go`/`TestSelfTest()`.
- `60-legacy.sh` removed (all tests commented in 2.x).

This brings up some questions:

1. The 2.x tests call `sudo` directly from the tests. This works for interactive shells, when no-password sudo access is available for a user, or when the tests are run as root. In other cases this results in `sudo: no tty present and no askpass program specified`. How do we want to handle tests that require elevated privileges in 3.0?
a) Continue to use `sudo` within tests. If `sudo` doesn't work, do we fail the whole test suite? Disable tests that rely on `sudo`?
b) Require the entire test suite to be run with `sudo`?
2. There are several failures in the help/usage tests, which break down into a couple of categories:
a) The "bogus" tests don't return a failure. Should they?
b) Running a command followed by `help` doesn't print usage as it used to in 2.x. Should it?


dtrudg:
Hi @tri-adam - on having thought more about this, I don't think a requirement to have the whole test suite run as root is great - as we should be testing paths such as image builds and pulls as a normal user. The effective user there can affect behaviour, e.g. in the patch to containers/image where when running as sudo to root a build of `docker://centos` will succeed without the patch, but as a user would fail. I guess a happy medium might be to require sudo for the test suite, but drop privileges to test portions of the code?

tri-adam:
Yeah, that should be easy enough to do I think. At the moment I drop privilege in `TestMain()` using `seteuid()` and then re-elevate where necessary. I can port that to the other tests once I'm sure it's all working as expected in this suite.

tri-adam:
A proposal for assigning tests appropriate privilege is available in #1765.

tri-adam:
Closing this - work will be split up into multiple PRs, the first of which is #1772.

PR #1745, opened 2018-07-10
Development 3.0: runtime adjustements/fixes

cclerget:

- Fix instances bug
- Rename instance_socket by master_socket for multi-purpose (pass file descriptor)
- Add master socket connection as parameter for PrepareConfig and StartProcess.
- Clear environment variables in Go section because Go use its own duplicate environment variables stack.

### DIFF
--- a/src/runtime/startup/c/wrapper.c
+++ b/src/runtime/startup/c/wrapper.c
@@ -47,6 +47,14 @@
 #define PR_GET_NO_NEW_PRIVS 39
 #endif
 
+#ifndef CLONE_NEWUSER
+#define CLONE_NEWUSER       0x10000000
+#endif
+
+#ifndef CLONE_NEWCGROUP
+#define CLONE_NEWCGROUP     0x02000000
+#endif
+
 #include "c/lib/util/capability.h"
 #include "c/lib/util/message.h"
 
@@ -63,7 +71,7 @@ char *json_stdin;
 
 /* Socket process communication */
 int rpc_socket[2] = {-1, -1};
-int instance_socket[2] = {-1, -1};
+int master_socket[2] = {-1, -1};
 
 #define SCONTAINER_STAGE1   1
 #define SCONTAINER_STAGE2   2
@@ -472,7 +480,7 @@ __attribute__((constructor)) static void init(void) {
     char *pipe_fd_env;
     int output[2];
     int status;
-    struct pollfd fds;
+    struct pollfd fds[2];
     int syncfd = -1;
     int pipe_fd = -1;
 
@@ -519,6 +527,17 @@ __attribute__((constructor)) static void init(void) {
 
     config.isSuid = is_suid();
 
+    if ( config.isSuid || geteuid() == 0 ) {
+        /* force kernel to load overlay module to ease detection later */
+        if ( mount("none", "/", "overlay", MS_SILENT, "") < 0 ) {
+            if ( errno != EINVAL ) {
+                singularity_message(DEBUG, "Overlay seems not supported by kernel\n");
+            } else {
+                singularity_message(DEBUG, "Overlay seems supported by kernel\n");
+            }
+        }
+    }
+
     if ( config.isSuid ) {
         singularity_message(DEBUG, "Drop privileges\n");
         if ( setegid(gid) < 0 || seteuid(uid) < 0 ) {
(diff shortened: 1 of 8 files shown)


src/runtime/startup/c/wrapper.c
```diff
@@ -519,6 +527,17 @@ __attribute__((constructor)) static void init(void) {
 
     config.isSuid = is_suid();
 
+    if ( config.isSuid || geteuid() == 0 ) {
+        /* force kernel to load overlay module to ease detection later */
+        if ( mount("none", "/", "overlay", MS_SILENT, "") < 0 ) {
+            if ( errno != EINVAL ) {
+                singularity_message(DEBUG, "Overlay seems not supported by kernel\n");
```
bauerm97, line 534: Should we do something else besides just print a debug message if not supported? Or just continue on and later if we *need* overlay then we can fail

src/runtime/startup/c/wrapper.c
```diff
@@ -596,31 +622,39 @@ __attribute__((constructor)) static void init(void) {
...
             }
-            if ( (ret = read(output[0], json_stdin, config.jsonConfSize)) != config.jsonConfSize ) {
+            if ( (ret = read(fds[0].fd, json_stdin, config.jsonConfSize)) != config.jsonConfSize ) {
                 singularity_message(ERROR, "Failed to read JSON configuration from stdout pipe: %s\n", strerror(errno));
                 exit(1);
             }
             json_stdin[config.jsonConfSize] = '\0';
+            break;
+        }
+        /* TODO: pass file descriptors from stage 1 over unix socket */
```
bauerm97, line 654: I assume this is just put here temporarily until we need to use this socket?

src/runtime/startup/c/wrapper.c
```diff
@@ -760,13 +788,8 @@ __attribute__((constructor)) static void init(void) {
         stage_pid = fork();
     } else {
         if ( config.nsFlags & CLONE_NEWPID ) {
-#ifdef NS_CLONE_NEWPID
```
bauerm97, line 763: Why do we remove check for if PIDNS supported?

src/runtime/startup/c/wrapper.c
```diff
@@ -841,7 +862,6 @@ __attribute__((constructor)) static void init(void) {
                 }
             }
         }
-#ifdef NS_CLONE_NEWCGROUP
```
bauerm97, line 844: Similar question here, why remove check for if newcgroup supported? I assume has something to do with addition of `#ifndef` blocks at top of file?

src/runtime/startup/smaster.go
```diff
@@ -106,23 +114,33 @@ func SMaster(socket C.int, instanceSocket C.int, config *C.struct_cConfig, jsonC
 }
 
 func main() {
+	loglevel := os.Getenv("SINGULARITY_MESSAGELEVEL")
+
+	os.Clearenv()
```
bauerm97, line 119: Is necessary to have here too since it's added in C code, or it's just in case?

bauerm97 (changes requested):
Also just a few quick questions. Changes look sane. Seems to mostly just make `instance_socket` a more generic socket `master_socket` for more generic use. Couple questions so I can better understand the code before merge :)

src/runtime/startup/c/wrapper.c
cclerget: Nothing to do here, just a trick to force overlay module to be loaded for further detection in Go code with proc helper :)

src/runtime/startup/c/wrapper.c
cclerget: Yes

src/runtime/startup/c/wrapper.c
bauerm97: OH... I see what you've done there. That's sneaky and awesome trick :)

src/runtime/startup/c/wrapper.c
cclerget: I think CLONE_NEWPID is supported on system having NO_NEW_PRIVS :)

src/runtime/startup/c/wrapper.c
bauerm97: 👍

src/runtime/startup/c/wrapper.c
cclerget: The logic here is to fail rather than don't execute the code because not supported. More explicit for debug or user issues

src/runtime/startup/smaster.go
cclerget: Yes I was surprise too, Go code don't use the same environment variable stack as C, seems to create/copy its own :/

bauerm97: Go is a funny guy

bauerm97 (approved):

PR #1747, opened 2018-07-11
Development 3.0: Add /proc filesystem helpers

cclerget:

- Add `HasFilesystem` to check if a filesystem is supported or loaded by kernel
- Add `ParseMountInfo` to return a map of parent mount points with associated childs mount points

### DIFF
--- a/src/pkg/util/fs/proc/proc.go
+++ b/src/pkg/util/fs/proc/proc.go
@@ -1,0 +1,58 @@
+// Copyright (c) 2018, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package proc
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"strings"
+)
+
+// HasFilesystem returns whether kernel support filesystem or not
+func HasFilesystem(fs string) (bool, error) {
+	p, err := os.Open("/proc/filesystems")
+	if err != nil {
+		return false, fmt.Errorf("can't open /proc/filesystems: %s", err)
+	}
+	defer p.Close()
+
+	suffix := "\t" + fs
+	scanner := bufio.NewScanner(p)
+	for scanner.Scan() {
+		if strings.HasSuffix(scanner.Text(), suffix) {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+// ParseMountInfo parses mountinfo pointing to path and returns a map
+// of parent mount points with associated child mount points
+func ParseMountInfo(path string) (map[string][]string, error) {
+	mp := make(map[string][]string)
+	mountlist := make(map[string][]string)
+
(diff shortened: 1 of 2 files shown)


src/pkg/util/fs/proc/proc.go
```diff
@@ -0,0 +1,58 @@
...
+
+	p, err := os.Open(path)
+	if err != nil {
+		return mp, fmt.Errorf("can't open %s: %s", path, err)
+	}
+	defer p.Close()
+
+	scanner := bufio.NewScanner(p)
+	for scanner.Scan() {
+		splitted := strings.Split(scanner.Text(), " ")
```
bauerm97: Better is `fields := strings.Fields(scanner.Text())`, this will correctly split over any amount of tab/space whitespace.

src/pkg/util/fs/proc/proc.go
```diff
@@ -0,0 +1,58 @@
...
+		mountlist[splitted[0]] = splitted
+	}
+	for k := range mountlist {
```
bauerm97: Not sure I understand the logic going on in this `for` block. Could you clarify? Is this looking to see if the source of a mount is a mountpoint itself (which would be stored in the `mountlist` then)?

src/pkg/util/fs/proc/proc.go
```diff
@@ -0,0 +1,58 @@
...
+
+// HasFilesystem returns whether kernel support filesystem or not
+func HasFilesystem(fs string) (bool, error) {
+	p, err := os.Open("/proc/filesystems")
+	if err != nil {
+		return false, fmt.Errorf("can't open /proc/filesystems: %s", err)
+	}
+	defer p.Close()
+
+	suffix := "\t" + fs
```
bauerm97, line 23: Question, why do we split over tab + string and then use combination as suffix?

bauerm97 (changes requested):
Just few questions

src/pkg/util/fs/proc/proc.go
cclerget: The mountlist contains key / line of mountinfo, key corresponding to the first column which is a unique number identifying the mount point, in the second loop each keys of mountlist is searched for a match in the second column, if matched so it's a children of the mount point itself

src/pkg/util/fs/proc/proc.go
cclerget: Cool thanks for the trick Michael ;)

src/pkg/util/fs/proc/proc.go
cclerget: /proc/filesystems content is either :
- `nodev\tsysfs`
- `\toverlay`
Be sure to match the exact file system between with `\t` and EOL

src/pkg/util/fs/proc/proc.go
bauerm97: OK figured, this makes sense!

bauerm97 (approved):

PR #1748, opened 2018-07-11
3.0 Home Directory functionality (WIP ?)

bauerm97:

- Adds support for `-H/--home` CLI flag
- Adds support for `--pwd` CLI flag
- Properly sets environment variable in container when `--home` is set
- Properly changes to the appropriate working directory inside the container

TODO:
- Flesh out testing of `src/pkg/util/fs/mount/home` package


src/pkg/util/fs/helper_linux_test.go
```diff
@@ -45,3 +47,23 @@ func TestIsSuid(t *testing.T) {
...
+	var tests = []struct {
+		path     string
+		expected string
+	}{
+		{"/path/to/something", "/path"},
+		{"relative/path", "relative"},
+		{"/path/to/something/", "/path"},
+		{"relative/path/", "relative"},
+		{"/path", "/path"},
+		{"/path/", "/path"},
```
dtrudg, line 61: Is there a case to have tests here which include . and .. components?

bauerm97: Absolutely. So the reason this file is commented out currently is because I'm still pondering the best way to handle the testing, as well as what the interface to the package should be. Once I figure those things out, I'll definitely add some tests which include `.` and `..`. Thanks for the suggestion

bauerm97:
Resolved by #1781

PR #1752, opened 2018-07-11
Arch ConveyorPacker

ikaneshiro:
This contains a ConveyorPacker that can build a valid root fs from pacstrap. This isnt intended to properly handle definitions at this point.

Attn: @singularityware-admin


src/pkg/build/conveyorPacker_arch.go
```diff
@@ -0,0 +1,224 @@
+// Copyright (c) 2018, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE file distributed with the sources of this project regarding your
```
jmstover, line 3: `LICENSE` in these headers should be `LICENSE.md` ... :/

dtrudg:
Hi @ikaneshiro - I need this diff to add in `pacman-contrib` which contains `paccache` (recently moved there) and causes an error if it isn't included.

```diff
diff --git a/src/pkg/build/conveyorPacker_arch.go b/src/pkg/build/conveyorPacker_arch.go
index 303fd69e..58979dd7 100644
--- a/src/pkg/build/conveyorPacker_arch.go
+++ b/src/pkg/build/conveyorPacker_arch.go
@@ -177,6 +177,9 @@ func (c *ArchConveyor) getInstList() (instList []string, err error) {
                }
        }
 
+       // paccache is needed, and is in pacman-contrib not included in base
+       toInstall = append(toInstall, "pacman-contrib")
+
        return toInstall, nil
 }
```

I can the build an arch container on Fedora 28 with this setup...

 * `sudo dnf install pacman`
 * Uncomment a mirror in /etc/pacman.d/mirrorlist
 * Add to /etc/pacman.conf:
```
[extra]
SigLevel = Required DatabaseOptional
Include = /etc/pacman.d/mirrorlist
[community]
SigLevel = Required DatabaseOptional
Include = /etc/pacman.d/mirrorlist
```
 * `sudo pacman -Syuu`
 * `sudo pacman-key --populate archlinux`

then.... `sudo singularity build arch.def examples/arch/Singularity`

jmstover (dismissed):
Functionally looks correct to me with @dctrud changes added in.

src/pkg/build/conveyorPacker_arch.go
ikaneshiro: It looks like all files in 3.0 have this issue. May be better to handle this as a separate PR

dtrudg:
@ikaneshiro - noting here that .@GodloveD suggests we don't add pacman-contrib in the bootstrap, but manually install in the example def file instead. In which case this PR is okay for the conveyorPacker, but do you want to modify the example def within this, or separately?

ikaneshiro:
Ill modify the def file to what @GodloveD has for 2.x here once he makes that PR

dtrudg (dismissed):
This is working as expected for me :+1:

jmstover (dismissed):

dtrudg (approved):

PR #1753, opened 2018-07-12
Development 3.0: mount system

cclerget:

Add mount system to easily manipulate mount workflow by registering before/after tag hook functions, `System` allow to register a mount function which can be altered during mount phase via hook functions

### DIFF
--- a/src/pkg/util/fs/mount/system.go
+++ b/src/pkg/util/fs/mount/system.go
@@ -1,0 +1,84 @@
+// Copyright (c) 2018, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package mount
+
+import (
+	"fmt"
+)
+
+// hookFn describes function prototype for function
+// to be called before/after mounting a tag list
+type hookFn func(*System) error
+
+// mountFn describes function prototype for function responsible
+// of mount operation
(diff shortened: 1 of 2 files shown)


src/pkg/util/fs/mount/system.go
```diff
@@ -0,0 +1,84 @@
...
+	"fmt"
+)
+
+// hookFn describes function prototype for function
+// to be called before/after mounting a tag list
+type hookFn func(*System) error
+
+// mountFn describes function prototype for function responsible
+// of mount operation
+type mountFn func(*Point) error
```
bauerm97, line 18: I think this should be exported (`MountFn`). Reasoning is so that when someone is developing using this library, they can see exactly what the type they are supposed to use is (via GoDoc or VSCode, etc...)

src/pkg/util/fs/mount/system.go
```diff
@@ -0,0 +1,84 @@
...
+
+package mount
+
+import (
+	"fmt"
+)
+
+// hookFn describes function prototype for function
+// to be called before/after mounting a tag list
+type hookFn func(*System) error
```
bauerm97, line 14: Same here

bauerm97 (dismissed):

bauerm97 (approved):
Really nice. I definitely see the need for this after hacking around with it for 5 minutes ;) I'm going to hold off on #1748 until I get this and #1756 merged so I can utilize them. I think this is really in the right direction

PR #1755, opened 2018-07-12
Fix missing Arch package

jmstover:
Recently, Arch moved the paccache application (and others) to the
pacman-contrib package:
  https://bbs.archlinux.org/viewtopic.php?id=237507

paccache is one of these, and is needed ... but pacman-contrib is
not part of base.

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/libexec/bootstrap-scripts/deffile-driver-arch.sh
+++ b/libexec/bootstrap-scripts/deffile-driver-arch.sh
@@ -83,7 +83,7 @@ fi
 # having to wait for ages until entropy accumulates. See
 # https://wiki.archlinux.org/index.php/Install_from_Existing_Linux,
 # https://wiki.archlinux.org/index.php/Pacman/Package_signing.
-if ! eval "'$PACSTRAP' -C '$PACMAN_CONF' -c -d -G -M '$SINGULARITY_ROOTFS' haveged $BASE_TO_INST"; then
+if ! eval "'$PACSTRAP' -C '$PACMAN_CONF' -c -d -G -M '$SINGULARITY_ROOTFS' haveged pacman-contrib $BASE_TO_INST"; then
     rm -f "$PACMAN_CONF"
     message ERROR "\`$PACSTRAP' failed.\n"
     ABORT 255


GodloveD:
I'm not sure if this is needed or not.  If I try a build without the package on current development-2.x head (dc5ded53) I get this error, which seems pretty good. 

```
[vagrant@archlinux ~]$ sudo singularity build arch singularity/examples/arch/Singularity
Using container recipe deffile: singularity/examples/arch/Singularity
Sanitizing environment
ERROR: `pacstrap' is not in PATH. You can install it with `pacman -S arch-install-scripts'.
ABORT: Aborting with RETVAL=1
Cleaning up...
```

GodloveD:
I actually get the same error with or without this PR, so maybe I'm confused about what this PR does?

dtrudg:
@GodloveD - that error is not related to this one. You have to have the `arch-install-scripts` package on your host before you can bootstrap any arch container. This PR is addressing something later on, when we are bootstrapping the container itself.

dtrudg:
@GodloveD  - here's the error this addresses.... near the end of the build it will give...

```
(1/1) Updating the info directory file...
+ paccache -r -k0
/bin/sh: line 25: paccache: command not found
ABORT: Aborting with RETVAL=255
```

It looks like this is an issue since last month some time, from the arch forums.

dtrudg (approved):
Hi @jmstover - I can confirm the missing `paccache` error does occur in `development-2.x` and that this PR does resolve it.

With this PR I am building a container okay, with host versions:

```
pacman 5.0.2-5
arch-install-scripts 18-1
```



:+1:

GodloveD:
OK.  I'm on board with what this is supposed to do now.  derp.

@dctrud take a look at https://github.com/singularityware/singularity/pull/1757.  I think this should be the preferred fix.

dtrudg:
Closing this in favor of #1757

jmstover:
Handle a different way in #1757

Issue #1759, opened 2018-07-13
Add ability to specify inode limit in persistent overlay image


bbarker:
### Expected behavior

Allow user to specify inode limits with `singularity image.create`, or provide higher default.

### Actual behavior

Currently there is no way to specify inode limits according to `--help`:

```
$ singularity image.create --help
USAGE: singularity [...] image.create [create options...] <container path>
Create a new Singularity formatted blank image.

CREATE OPTIONS:
    -s|--size   Specify a size for an operation in MiB, i.e. 1024*1024B
                (default 768MiB)
    -F|--force  Overwrite an image file if it exists
```

Even though my host OS is only using 14% of the available inodes configured, I'm running out of inodes in my overlay image in Singularity:

```
Singularity> df -i
Filesystem              Inodes      Used Available Use% Mounted on
OverlayFS                49152     49149         3 100% /
OverlayFS                49152     49149         3 100% /
tmpfs                  4107507        46   4107461   0% /dev
````
### Steps to reproduce behavior

Create 50,000 files in a persistent overlay image

### Comments

Is this possible given our knowledge of OverlayFS? So far I haven't been able to find anything.

**Comment by bbarker** [comment]
I believe I can just add an `-N` parameter to the `mkfs.ext3` call at `https://raw.githubusercontent.com/singularityware/singularity/master/libexec/cli/image.create.exec`, and have this passed in if given as an arg to `image.create`.

**Comment by jmstover** [comment]
How... large is that image? A 4G sparse image (193MB usage) that I created has 262144 inodes.

**Comment by bbarker** [comment]
@jmstover It ran out of inodes around 600mb; I just used the default size from `image.create` (no `-s`  parameter given). Here's the output from within the container:

```
Singularity> df -h
Filesystem                Size      Used Available Use% Mounted on
OverlayFS               739.9M    602.9M     98.7M  86% /
OverlayFS               739.9M    602.9M     98.7M  86% /
```

**Comment by jmstover** [comment]
That looks like the right number of inodes based on mke2fs defaults... 

```
$ perl -le 'print 768*1024*1024;'
805306368
$ perl -le 'print 805306368/16384;'
49152
```

mke2fs makes 1 inode/16K ...  Are you storing a bunch of smaller files to where it makes sense to change the bytes-per-inode option? So, instead use something like:  `-i 8192`

Change it so it's 1 inode per 8K?

*Edit:*
You should be able to find your system defaults under:  `/etc/mke2fs.conf`

[...]

PR #1760, opened 2018-07-13
Specify bytes-per-inode when creating image

jmstover:
When using image.create to make an overlay image, this will allow
the specifying of the size of an inode.

mke2fs default is 16K per inode, which if you have a lot of small
files, means you would hit the inode limit before you would hit
the space limit.

This PR adds a `-i` option to specify the size in bytes of the
inodes to mkfs.ext3.

For example:
```
$ singularity image.create -i 8192 /tmp/overlay.img
```

Issue: #1759

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/etc/bash_completion.d/singularity
+++ b/etc/bash_completion.d/singularity
@@ -372,8 +372,10 @@ _singularity() {
                 create)
                     if [[ ${cur} == -s || ${cur} == --size ]]; then
                         return 0
+                    if [[ ${cur} == -i || ${cur} == --bytes-per-inode ]]; then
+                        return 0
                     elif [[ ${cur} == -* ]] ; then
-                        COMPREPLY=( $(compgen -W "--size --force --help" -- ${cur}) )
+                        COMPREPLY=( $(compgen -W "--size --force --help --bytes-per-inode" -- ${cur}) )
                     else
                         _filedir 
                     fi
--- a/etc/bash_completion.d/singularity
+++ b/etc/bash_completion.d/singularity
@@ -204,7 +204,7 @@ _singularity() {
             elif [[ ${cur} == -H || ${cur} == --hash ]]; then
                 _filedir
             elif [[ ${cur} == -s || ${cur} == --size ]]; then
-                _filedir
+                return 0
             elif [[ ${cur} == -* ]] ; then
                 COMPREPLY=( $(compgen -W "--name --commit --hash --size --force --help" -- ${cur}) )
             else
@@ -253,8 +253,8 @@ _singularity() {
 
         image.create)
             if [[ ${cur} == -s || ${cur} == --size ]]; then
-                _filedir
-            if [[ ${cur} == -i || ${cur} == --bytes-per-inode ]]; then
+                return 0
+            elif [[ ${cur} == -i || ${cur} == --bytes-per-inode ]]; then
                 return 0
             elif [[ ${cur} == -* ]] ; then
                 COMPREPLY=( $(compgen -W "--size --force --help --bytes-per-inode" -- ${cur}) )
@@ -275,7 +275,7 @@ _singularity() {
 
         image.expand)
             if [[ ${cur} == -s || ${cur} == --size ]]; then
-                _filedir
+                return 0
             elif [[ ${cur} == -* ]] ; then
                 COMPREPLY=( $(compgen -W "--size --help" -- ${cur}) )
             else
@@ -371,7 +371,7 @@ _singularity() {
 
                 create)
                     if [[ ${cur} == -s || ${cur} == --size ]]; then
-                        _filedir
+                        return 0
                     elif [[ ${cur} == -* ]] ; then
                         COMPREPLY=( $(compgen -W "--size --force --help" -- ${cur}) )
                     else
@@ -402,7 +402,7 @@ _singularity() {
 
                 expand)
                     if [[ ${cur} == -s || ${cur} == --size ]]; then
-                        _filedir
+                        return 0
                     elif [[ ${cur} == -* ]] ; then
                         COMPREPLY=( $(compgen -W "--size --help" -- ${cur}) )
                     else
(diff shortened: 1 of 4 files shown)


libexec/cli/image.create.exec
```diff
@@ -96,6 +102,19 @@ if [ -f "$SINGULARITY_IMAGE" ]; then
     fi
 fi
 
+# Get the inode option *before* creating the empty file
+# We want to exit before the 'dd' command
+INODE=""
+if [ -n "${SINGULARITY_INODESIZE:-}" ]; then
+    INODE="-i ${SINGULARITY_INODESIZE:-16384}"
```
GodloveD: Is the `:-16384}"` part of this really necessary?  We just checked in the previous line to make sure that `SINGULARITY_INODESIZE` is not emtpy.

libexec/cli/image.create.exec
```diff
@@ -96,6 +102,19 @@ if [ -f "$SINGULARITY_IMAGE" ]; then
...
+
+    if [ $(expr ${SINGULARITY_INODESIZE} % 1024) -ne 0 ]; then
+        message ERROR "Given inode size of ${SINGULARITY_INODESIZE} is not divisible by 1024\n"
+        exit 1
```
GodloveD: Should this be an `ABORT 255`?

GodloveD (changes requested):
I've got a few questions about the code inlined.  Also, can you add this to the `CHANGELOG.md` and to the `etc/bash_completion.d/singularity` file please?

libexec/cli/image.create.exec
jmstover: No ... Call it me being psychotic an putting a default value in case it's empty even though I just checked that it wasn't empty...

libexec/cli/image.create.exec
jmstover: Sure... I can change that. Call it habit for a `exit 1` failure ...

jmstover:
@GodloveD ... Changes and Additions done.

libexec/cli/image.create.info
```diff
@@ -11,11 +11,14 @@ CREATE OPTIONS:
     -s|--size   Specify a size for an operation in MiB, i.e. 1024*1024B
                 (default 768MiB)
     -F|--force  Overwrite an image file if it exists
+    -i          Specify a size for the inode ratio in Bytes. Default is
```
GodloveD: Oh sorry one more thing.  I think it's OK to have a long option without a short option, but I don't think we should have a short option without a long option.

GodloveD (dismissed):
Sorry one more thing.

GodloveD:
Just merged #1761 (which should fix travis).  So if you want to merge development-2.x back in here I'll take another look.

jmstover:
@GodloveD Whoo! It passed!

etc/bash_completion.d/singularity
```diff
@@ -369,7 +371,7 @@ _singularity() {
 
                 create)
```
GodloveD, line 372: Also needs the `-i` option to be added here:

```
11:13 AM
vagrant@ubuntu-xenial:~/singularity
$ singularity image create --
--force  --help   --size

11:13 AM
vagrant@ubuntu-xenial:~/singularity
$ singularity image.create --
--bytes-per-inode  --force            --help             --size
```

libexec/cli/image.create.exec
```diff
@@ -96,6 +102,19 @@ if [ -f "$SINGULARITY_IMAGE" ]; then
     fi
 fi
 
+# Get the inode option *before* creating the empty file
+# We want to exit before the 'dd' command
+INODE=""
+if [ -n "${SINGULARITY_INODESIZE:-}" ]; then
+    INODE="-i ${SINGULARITY_INODESIZE}"
+
+    if [ $(expr ${SINGULARITY_INODESIZE} % 1024) -ne 0 ]; then
```
GodloveD, line 111: This check is unnecessary.  We are not specifying the size of inodes, but the number of inodes in relation to the size of the disk space.  

On my system there is a minimum and maximum ratio that can be specified.  (min 1024/max 67108864), but I don't know how system specific so I'm hesitant to say we should check that the value is between that min and max.  

If you specify a value outside of that range, it will issue a warning and then create a broken file system for you.  I think the warning is pretty good though, and I'm inclined to just let it go at that.  

```
$ singularity image.create --bytes-per-inode 256 --size 512 ~/blah2
Creating empty 512MiB image file: /home/vagrant/blah2
Formatting image with ext3 file system
mkfs.ext3: invalid inode ratio 256 (min 1024/max 67108864)
```

etc/bash_completion.d/singularity
```diff
@@ -204,7 +204,7 @@ _singularity() {
             elif [[ ${cur} == -H || ${cur} == --hash ]]; then
                 _filedir
             elif [[ ${cur} == -s || ${cur} == --size ]]; then
-                _filedir
+                return 0
```
GodloveD: bummer.  This actually doesn't seem to change behavior at all.  wonder why?  🤔

GodloveD (changes requested):
Still needs a little more love I think.

etc/bash_completion.d/singularity
```diff
@@ -369,9 +371,11 @@ _singularity() {
 
                 create)
                     if [[ ${cur} == -s || ${cur} == --size ]]; then
-                        _filedir
+                        return 0
+                    if [[ ${cur} == -i || ${cur} == --bytes-per-inode ]]; then
```
GodloveD: Thars a syntax error here.  Needs to be an `elif`

GodloveD (changes requested):

etc/bash_completion.d/singularity
jmstover: Why you care about syntax errors? It should **know** what I _mean_ ...

GodloveD (approved):
lgtm!

ArangoGutierrez (approved):
LGTM

GodloveD:
Thanks @ArangoGutierrez and @jmstover !

PR #1761, opened 2018-07-13
Add option to not mount home directory

jmstover:

This PR adds a `--no-home` option that will *not* mount the users home directory even if `mount home = yes` is set in `singularity.conf`

This will still mount your $HOME if it is also your current working directory. There is not logic to circumvent that. So, normal usage would be something like:

```
$ cd /tmp
$ singularity exec --no-home ~/ubuntu.img ls -l $HOME
ls: cannot access '/home/jason': No such file or directory
```

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/.travis/debuild_test
+++ b/.travis/debuild_test
@@ -27,7 +27,7 @@ gdebi -n ../singularity-container_*.deb
 
 # Install python 3 and pylint
 apt-get -y install python3 python3-pip python3-setuptools
-pip3 install pylint
+pip3 install pylint==1.9.2
 
 # Install docker
 apt-get -y install apt-transport-https ca-certificates curl gnupg2 software-properties-common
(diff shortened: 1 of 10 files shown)


GodloveD (dismissed):
Looks sensible (though I haven't tested yet).  Can you add entry to changelog and bash completion script please?  (You knew I was going to ask!)

jmstover:
@cclerget @bauerm97 
If either of you can glance over this as well.

cclerget (approved):
Looks good :+1:

shahzebsiddiqui:
@jmstover 

Looks good with me, in terms of option name if `--no-home` makes sense then you can stick with this naming scheme but I think you will need to introduce option for other mount points. Since user could request disabling other paths you can do something like 
`--disable-mount=[/home, /var/, /tmp, /dev,] `. where you can pass option to disable specific mount points. I think you may need to discuss internally which mount can not be disabled.  Please note this option should override option defined in `singularity.conf`

GodloveD (approved):
lgtm, and works with my tests.

PR #1765, opened 2018-07-17
Privileged Test Framework

tri-adam:
Adding new package in `src/pkg/test`, which defines the following fuctions:
- `DropPrivilege()` and `ResetPrivilege()` - use in a test case that is to run without privileges. The former is used to drop privilege, the latter to reset to the original privilege. Note that tests using these two functions function as normal when run as an unprivileged user (for example, from VS Code.)
- `EnsurePrivilege()` - use in a test case that is to run with privilege. Calls `testing.Fatal` if not run with elevated privilege. Does nothing otherwise.
- `test.WithoutPrivilege()` - takes a test function and returns a test function that will run without elevated privilege. Useful for table-driven tests where `t.Run()` is called.
- `test.WithPrivilege()` - takes a test function and returns a test function that will run with elevated privilege. Useful for table-driven tests where `t.Run()` is called.

Existing tests have been updated to run with/without privilege as necessary.


dtrudg:
Hi @tri-adam - this looks pretty neat! One question - would it be good to call `t.Helper()` in the priv helper functions so that any error raised reports the location of the test that called them, rather than the line in the helper function?

tri-adam:
Oh, great idea. Will do!

dtrudg (dismissed):
I think this is about as elegant as it can get :+1:

dtrudg (approved):

yhcote (approved):
Looks good to me.

Issue #1622, opened 2018-06-06
Unable to download docker image from Azure registry using OAuth2 authentication


hakonenger:
## Version of Singularity:

2.5.1

### Expected behavior

Singularity should be able to download a docker image from an Azure registry.

### Actual behavior

Singularity is unable to log in, failing with the message

```
ERROR Error getting token for repository ***, please check your credentials.
'token'
Cleaning up...
```

I have attached the complete debug output here: 
[azure-registry-problem.txt](https://github.com/singularityware/singularity/files/2076221/azure-registry-problem.txt)

### Steps to reproduce behavior

`sudo singularity -d build test.simg docker://***.azurecr.io/***:latest`

hakonenger:
I found that the reason for the issue seems to be that the docker registry uses OAuth2 authentication and returns the access token with the key "access_token" instead of "token", which Singularity expects. This is supported by Docker "for compatibility with OAuth2", see https://docs.docker.com/registry/spec/auth/token/#token-response-fields

Here is a patch that fixes the issue: [oauth2-token-patch.txt](https://github.com/singularityware/singularity/files/2076307/oauth2-token-patch.txt)

hmeiland:
patch works great; should we make a pull request to get this into mainstream source?
thanks!

hmeiland:
I've made a small change to the patch to not use depricated has_key and committed a PR...

jakirkham:
Should this now be closed?

jmstover:
This was merged into 2.6.0 as well

JonShelley:
I just tried to do this and I am getting the same error. Not sure if I am doing something wrong or this is still not fixed.

Here is the error that I am seeing when I run the command
singularity -d build test.simg docker://***.azurecr.io/***/***:v1.3

DEBUG GET https://***.azurecr.io/v2/***/***/tags/list
DEBUG Http Error with code 401
DEBUG Requesting token from docker registry API
DEBUG GET https://***.azurecr.io/oauth2/token?scope=repository:***/***:pull&expires_in=300&service=***.azurecr.io
DEBUG Http Error with code 401
ERROR Error getting token for repository ***/***, please check your credentials.
No JSON object could be decoded
Cleaning up...

singularity --version
2.6.0-dist

JonShelley:
I finally got it to work. For those that are interested I did

export SINGULARITY_DOCKER_USERNAME=\<username\>
export SINGULARITY_DOCKER_PASSWORD=\<password\>
[...]

PR #1778, opened 2018-07-20
add token support for OAUTH2

hmeiland:

PR for issue #1622 , add token support for docker registries with OAUTH2
tested with Azure Container Registry.

- Ref: #1622 

**Checkoff for all PRs:**

- [ ] I have added changes to the [CHANGELOG](https://github.com/singularityware/singularity/blob/development/CHANGELOG.md) and and documentation updates to the [singularityware] (https://www.github.com/singularityware/singularityware.github.io) documentation base. -> to be done
- [ ] I have added myself as a contributor to the [contributors's file](https://github.com/singularityware/singularity/blob/master/CONTRIBUTORS.md) -> not my code, see real contributer in #1622

Attn: @singularityware-admin


hmeiland:
please verify 248:80; not sure if we can change that without creating other best-practise issues...

libexec/python/docker/api.py
```diff
@@ -225,7 +225,11 @@ def update_token(self, response=None, auth=None, expires_in=300):
 
         try:
             data = json.loads(response)
-            token = {"Authorization": "Bearer %s" % data["token"]}
+            if 'access_token' in data:
```
GodloveD, line 228: OK, so I'm going to show some of my ignorance about this stuff.  Please bear with me.  

In the old version we always default to `data["token"]`.  In this PR, we default to `data["access_token"]` if it exists and only fall back to `data["token"]` in the event that it doesn't.  Cool.  

Now here is where my ignorance comes in.  Will there ever be a situation in which: 
1) `data["access_token"]` exists and 
2) it is *not* the correct token to use

In other words could both `access_token` and `token` exist in the same hash table in a situation where OAUTH2 should not really be used.  And then would the presence of `access_token` block the correct usage of `token`.

If this might be the case, maybe you should add a `try`.  If `access_token` fails fall back to `token` and try again.  

Is this just nutty?

GodloveD (changes requested):
Can you please add this to the changelog under the enhancements section of 2.6?

Also my preference would be that both you and @hakonenger add yourselves to the authors file.

libexec/python/docker/api.py
dtrudg: This code will be fine as if `access_token` is present you are using the OAUTH2 path, which doesn't set a `token`. `token` will appear alone if you use the docker token authentication path.

Edit.... actually if an auth service is returning both (which would be very odd as is out of spec for the 2 types) we are still good by the client spec...

> At least one of these fields must be specified, but both may also appear (for compatibility with older clients). When both are specified, they should be equivalent; if they differ the client's choice is undefined.

dtrudg (approved):
This looks good to me. I'm not sure that it will handle all eventualities with the OAUTH2 auth flow. We don't handle e.g. refresh tokens at all, but if this allows basic usage of e.g. Azure I'm pretty confident it won't cause regression elsewhere.

dtrudg:
@hmeiland - for the PEP8 warning can you put a `# noqa` comment on the end of the line. I think this is one where making the linter happy is going to result in something less readable, which is not what we want.

libexec/python/docker/api.py
GodloveD: Thanks for the clarification @dctrud !

GodloveD:
So, just to summarize, add to changelog, add the `# noqa`, and add to contributors (optional) and I'll approve.  Thanks!

hmeiland:
Added # noqa, added contributers and added to changelog -> made bugfix entry for upcoming 2.6 release
Let me know if this ticks all the boxes...

GodloveD (approved):
lgtm

GodloveD:
Thanks @hmeiland and @dctrud !

PR #1781, opened 2018-07-20
Development 3.0: Bind stuff

cclerget:

Work in progress :
- Bind system
- Overlay/Underlay support
- Session management


src/pkg/util/fs/helper.go
```diff
@@ -63,3 +64,28 @@ func IsSuid(name string) bool {
...
+func MkdirAll(path string, mode os.FileMode) error {
+	oldmask := syscall.Umask(0)
+	defer syscall.Umask(oldmask)
+
+	return os.MkdirAll(path, mode)
+}
+
+// RootDir returns the root directory of path (rootdir of /my/path is /my).
+// Returns "." if path is empty
+func RootDir(path string) string {
```
bauerm97, line 79: I think we should add a few sanity test cases for this function in `helper_test.go`

src/runtime/engines/common/oci/config.go
```diff
@@ -30,6 +30,6 @@ func (c *Config) UnmarshalJSON(b []byte) error {
 	if err := json.Unmarshal(b, &c.Spec); err != nil {
 		return err
 	}
-	c.Generator = generate.NewFromSpec(&c.Spec)
+	c.Generator = generate.Generator{Config: &c.Spec}
```
bauerm97, line 33: Why is this change necessary? Just curious

bauerm97 (changes requested):
Can we make sure that Makefile.stub is only installing the necessary dir? (Since I believe now everything happens under `$(SESSIONDIR)`)

Also this isn't directly from this PR, but `src/pkg/util/fs/layout/manager.go`, can you change `func (m *Manager) GetPath(path string) (string, error)` to use `filepath.Join()`? Thanks!

This is just a first pass review looking at some code. I have yet to fully review `container.go`. And I have yet to do testing on functionality.

src/pkg/util/fs/helper.go
cclerget: Agreed

src/runtime/engines/common/oci/config.go
cclerget: Eduardo has pointed that NewFromSpec is marked as deprecated https://github.com/opencontainers/runtime-tools/blob/699dbced9a52c2d60213d8f22d6e170840d69c06/generate/generate.go#L234

bauerm97: Good catch @ArangoGutierrez

bauerm97:
Hey @cclerget I noticed a few issues so far in my testing:

- `--no-home` option leaves the `$HOME` environment variable set
- Config `allow pid ns = no` appears to not do anything
- In config file, at `LIMIT CONTAINER GROUPS` the default line says `DEFAULT: @LIMIT_CONTAINER_GROUPS_DEFAULT@` which should probably be `DEFAULT: NULL` same as limit container owners

bauerm97 (approved):

PR #1783, opened 2018-07-23
2.x->3.0 Test (21-build_docker.sh)

tri-adam:
Re-write 2.x tests in `21-build_docker.sh` for version 3.0.


src/cmd/singularity/cli_test.go
```diff
@@ -44,6 +46,17 @@ func run(m *testing.M) int {
 		log.Fatalf("singularity config is not a regular file")
 	}
 
+	// Make temp dir for tests
+	name, err := ioutil.TempDir("", "stest.")
+	if err != nil {
+		log.Fatalf("failed to create temporary directory: %v", err)
+	}
+	defer os.RemoveAll(name)
+	if err := os.Chmod(name, 0777); err != nil {
```
jmstover: Is the mode pre-umask or is umask ignored?

Did we have this conversation already? It seems like we might have had this conversation already...

tri-adam: I believe the `ioutil.TempDir()` will take `umask` into account... but the `os.Chmod()` will not.

dtrudg (approved):
The tests that are not excluded here look good to me. The local registry ones should be an easy fix - noted on issue. I'm not sure how we will manage the unpriv ones :-/

PR #1784, opened 2018-07-23
Localimage def fix

ikaneshiro:
Allow definitions to bootstrap with a localimage in the same way we can build normally from a local image, SIF still not supported. That will be handled in a separate PR.

### DIFF
--- a/src/cmd/singularity/cli/build.go
+++ b/src/cmd/singularity/cli/build.go
@@ -242,11 +242,11 @@ func makeDefinition(source string, isJSON bool) build.Definition {
 		}
 	} else if _, err := os.Stat(source); err == nil {
 		//local image or sandbox
-		//define source as local and follow uri format for defs
-		source = "local://" + source
-		def, err = build.NewDefinitionFromURI(source)
-		if err != nil {
-			sylog.Fatalf("unable to parse URI %s: %v\n", source, err)
+		def = build.Definition{
+			Header: map[string]string{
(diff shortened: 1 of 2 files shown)


src/cmd/singularity/cli/build.go
```diff
@@ -273,6 +273,8 @@ func makeBundle(def build.Definition) *build.Bundle {
 		cp = &build.DebootstrapConveyorPacker{}
 	case "arch":
 		cp = &build.ArchConveyorPacker{}
+	case "localimage":
+		cp = &build.LocalConveyorPacker{}
```
tri-adam, line 275: With 2.5.2, building with this def file:

```
Bootstrap: local
From: /tmp/image
```

Gives:

```
$ sudo singularity build /tmp/image2 recipe 
Using container recipe deffile: recipe
Sanitizing environment
ERROR: 'Bootstrap' type not supported: local
Cleaning up...
```

I think the case at line 266 is actually a typo... and we can just switch `local` to `localimage`.

tri-adam (changes requested):

src/cmd/singularity/cli/build.go
ikaneshiro: Agreed that option should have just been `localimage` in the first place. Made that change as well as a quick change on how definitions for local image builds are generated when building from the source file itself to make it cleaner, imo

tri-adam (approved):
Looks great!

PR #1774, opened 2018-07-19
2.x->3.0 Test (20-build.sh)

tri-adam:
Re-write 2.x tests in `20-build.sh` for version 3.0.

PR #1788, opened 2018-07-24
Sif packer

ikaneshiro:
Packer that uses the system partition of a SIF file as the rootfs for the bundle and subsequent build.

Attn: @singularityware-admin


src/pkg/build/packer_sif.go
```diff
@@ -0,0 +1,139 @@
...
+
+	if err := loopdev.Attach(src, os.O_RDONLY, &number); err != nil {
+		return err
+	}
+
+	if err := loopdev.SetStatus(info); err != nil {
+		return err
+	}
+
+	tmpmnt, err := ioutil.TempDir("/tmp", "tmpmnt-")
```
tri-adam: Is there a reason to explicitly use `/tmp`? I think we should probably leave that blank, that way users can override via `$TMPDIR` if necessary.

src/pkg/build/packer_sif.go
```diff
@@ -0,0 +1,139 @@
...
+	//copy filesystem into dest
+	sylog.Debugf("Copying filesystem from %s to %s\n", tmpmnt, dest)
+	cmd := exec.Command("cp", "-r", tmpmnt+`/.`, dest)
+	err = cmd.Run()
+	if err != nil {
+		sylog.Errorf("cp Failed", err.Error())
+		return err
+	}
+
+	return err
```
tri-adam: Nitpick: could be `return nil` here, just makes it a bit clearer that you're exiting without an error at this point.

tri-adam (changes requested):
Two small suggestions. Tested against the new tests in #1774 and I'm not seeing any SIF-related failures there.

tri-adam (approved):

src/pkg/build/packer_sif.go
```diff
@@ -0,0 +1,139 @@
...
+
+	return b, nil
+}
+
+// First pass just assumes a single system partition, later passes will handle more complex sif files
+// unpackSIF parses throught the sif file and places each component in the sandbox
+func (p *SIFPacker) unpackSIF(b *Bundle, rootfs string) (err error) {
+
+	// load the container
+	fimg, err := sif.LoadContainer(rootfs, false)
```
yhcote, line 48: The second parameter to LoadContainer() should be "true", because you do not modify the SIF file.

yhcote (commented):
The lookup for the 3 types of FS could be done with GetFromDescr() which would probably shorten the code a bit but I recalled that I did not complete the descriptor specific field lookup. That will come later.

PR #1799, opened 2018-07-29
Separate test rule into check and test.

yhcote:
Running make test, takes a while and during development there's interest
in checking the code formatting without running tests.

This change allows that to be done easily.


cclerget (approved):
Looks good ! I will add a shortcut to run check/test

PR #1802, opened 2018-07-29
Pgp enhancements and new "singularity keys" command

yhcote:
This PR implements the new "singularity keys" command and currently we have subcommands : 
- list : to list the keys from the public local store by default of private keys with "-s"
- newpair : to generate a new key pair and store it in the default store location ($HOME/.singularity/sypgp/*)
- search : to search a group of keys matching a search string by connecting to a key server
- pull : to fetch and store a key found at key server matching a specific fingerprint
- push : to upload a public key from local key store to a key server specified by fingerprint

Next PR will add at least one new command "update" to update a key and upload a copy to the key server if requested.

A few changes were made of the sypgp package to accomodate those new commands by adding a few new routine. Cleanup around has been done also at the same time.


cclerget (approved):
Looks good !

PR #1803, opened 2018-07-30
Adapt lookup function to new SIF lookup semantics

yhcote:
A few SIF lookup functions now may return multiple elements matching
search criterias. Get the singularity code ready for the new usage.


cclerget (approved):
Looks good

PR #1805, opened 2018-07-30
Build test skips for short runs

ikaneshiro:
Skips conveyor test for debootstrap, pacman, and busybox.
Adds tmpfs cleanup and a typo fix for debootstrap packer test

### DIFF
--- a/src/pkg/build/conveyorPacker_debootstrap_test.go
+++ b/src/pkg/build/conveyorPacker_debootstrap_test.go
@@ -37,13 +37,11 @@ func TestDebootstrapConveyor(t *testing.T) {
 	dc := DebootstrapConveyor{}
 
 	err := dc.Get(testDef)
+	//clean up tmpfs since assembler isnt called
+	defer os.RemoveAll(dc.tmpfs)
 	if err != nil {
-		//clean up tmpfs since assembler isnt called
-		os.RemoveAll(dc.tmpfs)
 		t.Fatalf("Debootstrap Get failed: %v", err)
 	}
-
-	os.RemoveAll(dc.tmpfs)
 }
 
 func TestDebootstrapPacker(t *testing.T) {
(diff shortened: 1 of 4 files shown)


src/pkg/build/conveyorPacker_debootstrap_test.go
```diff
@@ -33,8 +38,12 @@ func TestDebootstrapConveyor(t *testing.T) {
 
 	err := dc.Get(testDef)
 	if err != nil {
+		//clean up tmpfs since assembler isnt called
+		os.RemoveAll(dc.tmpfs)
 		t.Fatalf("Debootstrap Get failed: %v", err)
 	}
+
+	os.RemoveAll(dc.tmpfs)
```
tri-adam: Rather than these two `os.RemoveAll()`, you could `defer os.RemoveAll()` right after the `dc.Get()` line (before the conditional.)

src/pkg/build/conveyorPacker_debootstrap_test.go
```diff
@@ -48,19 +57,23 @@ func TestDebootstrapPacker(t *testing.T) {
...
 	dcp := DebootstrapConveyorPacker{}
 
 	err := dcp.Get(testDef)
 	if err != nil {
+		//clean up tmpfs since assembler isnt called
+		os.RemoveAll(dcp.tmpfs)
 		t.Fatalf("Debootstrap Get failed: %v", err)
 	}
 
+	defer os.RemoveAll(dcp.tmpfs)
```
tri-adam: Similar comment here, move this `defer` right after the `dc.Get()` and remove the duplicate `os.RemoveAll()`.

tri-adam (changes requested):
Two tiny changes, thanks Ian.

tri-adam (approved):

PR #1811, opened 2018-08-02
Development 3.0: Fix /proc/self/auxv permission denied during tests

cclerget:

Use setresuid/setresgid during uid/gid transition, required to execute binaries during tests with original user uid/gid


tri-adam (approved):
Brilliant, thanks Cedric!

PR #1825, opened 2018-08-08
3.0 buildapi and package refactoring

ikaneshiro:
Refactoring build package components into smaller packages in order to be better organized.

This also adds a nice api for build package in `build/build.go` that allows for building without having to deal with package internals like `ConveyorPackers`, `Bundles`, and `Assemblers`


src/pkg/build/build.go
```diff
@@ -0,0 +1,298 @@
...
+	"github.com/singularityware/singularity/src/runtime/engines/imgbuild"
+)
+
+// Build is an abstracted way to look at the entire build process.
+// For example calling NewBuild() will return this object.
+// From there we can call Full() on this build object, which will:
+// 		Call Bundle() to obtain all data needed to execute the specified build locally on the machine
+// 		Execute all of a definition using AllSections()
+// 		And finally call Assemble() to create our container image
+type Build struct {
```
dtrudg, line 34: Would it be possible to get some more per-field comments on a fundamental type like this? I think these are the types of things where we want a really nice godoc, or peek at definition experience for people who may be working in the build system, and need a grasp of the architecture.

If there's a build object a the godoc very briefly what a ConveyorPacker, Assmebler, Bundle, Definition are it would be easier to dive in.

ikaneshiro: Good point. Ill add some more info

src/pkg/build/build.go
```diff
@@ -0,0 +1,298 @@
...
+	dest        string           // Location for container after build is complete
```
bauerm97: Thanks for adding the comments. Can you change the formatting a bit? To work with `godoc` you should change:

`dest        string           // Location for container after build is complete`
to
```
// dest is the location for container after build is complete
dest string
```

ikaneshiro: Done

ArangoGutierrez:
+1

bauerm97 (approved):
looks good to me

PR #1834, opened 2018-08-11
cleaning up some cruft left when we were trying to include dot versio…

GodloveD:

Cleaning up some of the cruft that was left over from our attempt to support dot versions of the command groups.  It was implemented in a neat way, but we just decided not to support the dot versions for simplicity. 

**Checkoff for all PRs:**

Attn: @singularityware-admin


GodloveD:
Heya @bauerm97 could you have a peek at this when able?

src/cmd/singularity/cli/capability_add.go
```diff
@@ -30,14 +26,10 @@ func init() {
...
 
-	for _, cmd := range capabilityAddCmds {
-		cmd.Flags().AddFlag(capabilityAddFlags.Lookup("user"))
-		cmd.Flags().AddFlag(capabilityAddFlags.Lookup("group"))
-		cmd.Flags().AddFlag(capabilityAddFlags.Lookup("desc"))
-		cmd.Flags().SetInterspersed(false)
-	}
-
-	// SingularityCmd.AddCommand(capabilityDotAddCmd)
+	CapabilityAddCmd.Flags().AddFlag(capabilityAddFlags.Lookup("user"))
```
bauerm97: I think for the capability commands it's okay to directly add the flag to the `Cmd.Flags()` directly. We have the `actionFlags` variable because multiple commands needed to share the same set of flags.

So I would say we can simply do:

`CapabilityAddCmd.Flags().StringVarP(...)` etc...

Same for the other commands that don't share flags with other commands.

bauerm97 (changes requested):

src/cmd/singularity/cli/capability_add.go
GodloveD: I think these are shared between subcommands too.  So this avoids a bit of redundancy.  Right?

GodloveD: specifically the `add`, `drop`, and `list` subcommands afaik

GodloveD: Nevermind.  I'm dumb.  🤤

bauerm97 (approved):
Changes look good, cheers @GodloveD

PR #1837, opened 2018-08-13
Add SecurityOperations interface for greater engine flexibility

bauerm97:

Adds `SecurityOperations` interface to allow for greater engine flexibility. This will eventually be expanded to include most security features that an Engine can request. This also adds the `NamespaceFlags` function to the `SecurityOptions` interface, allowing the engine to return the `uint` bit field for the `cConf` to directly consume. Also includes a helper util package to provide default functionality.


bauerm97:
Handled by #1843

PR #1846, opened 2018-08-16
3.0: fix compilation on RHEL6 and RHEL7

DrDaveD:

This PR makes the master branch compile on RHEL6 and RHEL7.  Without it, compilation dies with these messages:
```
GO singularity
# github.com/singularityware/singularity/src/pkg/util/user
../src/pkg/util/user/identity.go:79: cannot use C.uid_t(uid) (type C.uid_t) as type C.__uid_t in argument to _C2func_getpwuid
../src/pkg/util/user/identity.go:114: cannot use C.gid_t(gid) (type C.gid_t) as type C.__gid_t in argument to _C2func_getgrgid
make: *** [singularity] Error 2
```

None

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/src/pkg/util/user/identity.go
+++ b/src/pkg/util/user/identity.go
@@ -15,11 +15,15 @@ package user
 #include <grp.h>
 
 #ifndef uid_t
+#ifndef __uid_t_defined
 typedef uid_t __uid_t;
+#endif
 #endif
 
 #ifndef gid_t
+#ifndef __gid_t_defined
 typedef gid_t __gid_t;
+#endif
 #endif
 */
 import "C"


jscook2345:
Thanks for the contribution @DrDaveD!

It appears the code here was originally meant to support the usage of musl-libc as an alternative to glibc. Here is the commit where the change went in: 

https://github.com/singularityware/singularity/pull/1846/commits/ca5fb96d8bfdd397030f6b5a5a28ef95cfe341bd

While I think the fix you provided *should* work for both libc and musl-libc (since the *_defined macro will never be defined for a musl-libc system) I think it would be a good idea to be more verbose here.

Perhaps it would be better to enclose the current section in an ifdef checking for GLIBC? Those typedefs do not appear to be needed so they can be ignored.

#ifndef __GLIBC__
...
#endif

What do you think?

DrDaveD:
That does seem to be less hacky than my approach, but I really don't know.  Let me know if you want me to update the PR.

bauerm97:
@DrDaveD Hey Dave, thanks for diving into 3.0 and helping out. It's greatly appreciated! Unfortunately, we're not sure that there's a great way to make a "non-hacky" cross-distro solution to wrapping the `C.getpwuid` function call. I think @cclerget is going to rework this package to instead wrap the functions implemented in the Go standard library `os/user` package. @cclerget has [submitted a pull request](https://github.com/singularityware/singularity/pull/1852) implementing this change, would you be able to test that and see if it resolves this issue?

DrDaveD:
I tested it on el6 and it got a different compilation error.  I will post the results in the PR.

bauerm97:
Fixed by #1852

PR #1848, opened 2018-08-17
Sif update

yhcote:
sylabs/sif has got a few new routines to deal with partition order and the setting of extra descriptor data. This PR takes advantage of that and cleans up the code accessing SIF.


bauerm97 (approved):
Looks like good changes to me! Thanks @yhcote

PR #1850, opened 2018-08-17
Development 3.0: Add check for duplicated mount point

cclerget:

This PR add checks for duplicated mount points, actually check for underlay is disabled because underlay add many false positive due to it's nature.


yhcote (approved):
looks good to me.

PR #1853, opened 2018-08-17
Building using Definitions and Metadata Reorganization

ikaneshiro:
#### This PR will implement builds respecting definitions minimal container FS metadata Currently:

- Insert hooks into build engine and assemblers in order to properly execute sections of definitions
- build sandboxes respecting definitions
- runtime handles exec, shell and run without relying on /.singularity.d

#### Work left to be done:

- build SIF files respecting definitions
- make runtime handle SIF metadata properly since that info won't be in the FS like sandboxes or legacy images
- build system needs to detect old metadata and handle building over it


ikaneshiro:
This metadata approach will not be pursued at the moment

PR #1860, opened 2018-08-19
Set Version Using Semver Git Tag

tri-adam:
Use `git describe` to set version in `PACKAGE_VERSION` and `PACKAGE_STRING` based on latest annotated git tag. With this change, the values in this file will not need to be updated manually on release.

When the current commit is tagged as a release, the tag name is used. So, for example, when `v3.0.0` is tagged, the behavior will be:

```sh
$ singularity version
3.0.0
```

When the current commit is not tagged as a release, the format is `<version_tag>-<commits_since_version_tag>-g<commit_hash>`. In this branch, the latest version tag is `v3.0.0-alpha.1`, I have `2` commits since that tag, and the current commit hash is `e8f1270b`. Thus the version strings:

```sh
$ singularity version
3.0.0-alpha.1-2-ge8f1270b
```

### DIFF
--- a/mlocal/checks/project.chk
+++ b/mlocal/checks/project.chk
@@ -10,7 +10,7 @@ echo "" > $makecond
 ########################
 # git version
 ########################
-git_version=`git describe --match 'v[0-9]*' --long --dirty --always`
+git_version=`git describe --match 'v[0-9]*' --dirty --always | sed -e "s/^v//"`
 
 config_add_def PACKAGE_NAME \"singularity\"
 config_add_def PACKAGE_TARNAME \"singularity\"
--- a/mlocal/checks/project.chk
+++ b/mlocal/checks/project.chk
@@ -7,10 +7,15 @@ set -e
 makecond="$builddir/Makefile.cond"
 echo "" > $makecond
 
+########################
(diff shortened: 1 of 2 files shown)


jmstover (dismissed):

mlocal/checks/project.chk
```diff
@@ -7,10 +7,15 @@ set -e
...
+########################
+git_version=`git describe --match 'v[0-9]*' --long --dirty --always`
+
 config_add_def PACKAGE_NAME \"singularity\"
 config_add_def PACKAGE_TARNAME \"singularity\"
-config_add_def PACKAGE_VERSION \"3.0.0\"
-config_add_def PACKAGE_STRING \"singularity 3.0.0\"
+config_add_def PACKAGE_VERSION \"$git_version\"
+config_add_def PACKAGE_STRING \"singularity $git_version\"
 config_add_def PACKAGE_BUGREPORT \"support@sylabs.io\"
```
jmstover, line 19: Question ... Should it be this email address for the OSS version? Or should for OSS it be the mailing list?

yhcote (approved):
looks good to me

PR #1861, opened 2018-08-20
Set HTTP User Agent (based on 1860)

tri-adam:
Note: review #1860 first... this  is based off the work in that PR.

Set `User-Agent` header in outgoing HTTP requests. On my workstation, this sets the value to:

```
Singularity/3.0.0 (Linux amd64) Go/1.10.3
```


yhcote (approved):
looks good to me, I like the idea.

PR #1862, opened 2018-08-20
Test PGP Key Fetch

tri-adam:
Add `FetchPubkey()` test. This takes code coverage for the `sypgp` package from `0%` -> `10.6%`.


yhcote (approved):
10 and 90 more in the way !

PR #1864, opened 2018-08-20
Building with definitions

ikaneshiro:
This PR performs builds that respect the various sections of definition files and executes them where appropriate. Builds are orchestrated through functions within `src/pkg/build/build.go`

Resulting container has a `/.singularity.d/` directory that contains all container metadata., in 
order to have metadata parity with 2.x for the time being.

Attn: @singularityware-admin


bauerm97 (changes requested):
Take care of exposing `SINGULARITY_ROOTFS` & `SINGULARITY_ENVIRONMENT` env vars as we discussed

bauerm97 (approved):
LGTM

PR #1865, opened 2018-08-21
Pgp flow

yhcote:
Better look and feel when using sign/verify and internal cleanup of code in signing and openpgp stack


src/pkg/signing/signing.go
```diff
@@ -22,23 +22,13 @@ import (
 )
 
 const (
-	keyserverURI = "https://keys.sylabs.io:11371"
+	defaultKeysServer = "https://keys.sylabs.io:11371"
```
tri-adam: Can you remove `:11371`? We can use https://keys.sylabs.io now, port 443 will work from everywhere (Starbucks with paranoid firewall etc.)

src/pkg/sypgp/sypgp.go
```diff
@@ -107,15 +108,28 @@ func printSignatures(entity *openpgp.Entity) error {
...
+}
+
 // DirPath returns a string describing the path to the sypgp home folder
 func DirPath() string {
 	user, err := user.GetPwUID(uint32(os.Getuid()))
 	if err != nil {
-		sylog.Errorf("could lookup user's real home folder %s\n", err)
-		return ""
+		sylog.Warningf("could not lookup user's real home folder %s\n", err)
+		sylog.Warningf("using current directory for %s\n", filepath.Join(".singularity", "sypgp"))
```
tri-adam: Nitpick: logs shouldn't end with `\n`.

tri-adam (changes requested):
Two tiny requests

src/pkg/sypgp/sypgp.go
bauerm97: Huh, so I just realized that sylog strips `\n` from the ending of any log messages, and then makes sure there is exactly one `\n` at the end. So there's a lot of code not doing this properly all over the rest of the codebase!

tri-adam: Yeah, it's an idiom change from v2->v3, I'm sure these will be pervasive for a while. If you like, we could start linting for that 😉

yhcote: Yes and this could have been a leftover from the previous sylog call. sypgp was an early sylog adopter ;-) This is the right way to do it though and I believed I lobbied (with support) for that to happen also.

A general code cleanup is a good idea @bauerm97 !

src/cmd/singularity/cli/keys_pull.go
```diff
@@ -18,7 +18,7 @@ import (
 
 func init() {
 	KeysPullCmd.Flags().SetInterspersed(false)
-	KeysPullCmd.Flags().StringVarP(&url, "url", "u", "", "overwrite the default remote url")
+	KeysPullCmd.Flags().StringVarP(&keyServerURL, "url", "u", "", "specify the key server URL (default: https://keys.sylabs.io)")
```
tri-adam: `StringVarP` fourth argument is the default, let's use that. And rather than duplicate the URL, let's use the variable. This:

```
KeysPullCmd.Flags().StringVarP(&keyServerURL, "url", "u", defaultKeysServer, "specify the key server URL")
```

Will give you the desired result:

```
  -u, --url string   specify the key server URL (default
                     "https://keys.sylabs.io")
```

src/cmd/singularity/cli/keys_push.go
```diff
@@ -18,7 +18,7 @@ import (
 
 func init() {
 	KeysPushCmd.Flags().SetInterspersed(false)
-	KeysPushCmd.Flags().StringVarP(&url, "url", "u", "", "overwrite the default remote url")
+	KeysPushCmd.Flags().StringVarP(&keyServerURL, "url", "u", "", "specify the key server URL (default: https://keys.sylabs.io)")
```
tri-adam: Same here:

```
KeysPushCmd.Flags().StringVarP(&keyServerURL, "url", "u", defaultKeysServer, "specify the key server URL")
```

src/cmd/singularity/cli/keys_search.go
```diff
@@ -18,7 +18,7 @@ import (
 
 func init() {
 	KeysSearchCmd.Flags().SetInterspersed(false)
-	KeysSearchCmd.Flags().StringVarP(&url, "url", "u", "", "overwrite the default remote url")
+	KeysSearchCmd.Flags().StringVarP(&keyServerURL, "url", "u", "", "overwrite the default remote url")
```
tri-adam: Missing the URL. Suggest same approach as the others here.

src/cmd/singularity/cli/sign.go
```diff
@@ -17,26 +17,43 @@ import (
 
 func init() {
 	SignCmd.Flags().SetInterspersed(false)
+	SignCmd.Flags().StringVarP(&keyServerURL, "url", "u", "", "specify the key server URL (default: https://keys.sylabs.io)")
```
tri-adam: Suggest same approach as the others here.

src/cmd/singularity/cli/sign.go
```diff
@@ -17,26 +17,43 @@ import (
...
 
 	Use:     docs.SignUse,
 	Short:   docs.SignShort,
 	Long:    docs.SignLong,
 	Example: docs.SignExample,
 }
+
+func doSignCmd(cpath, url string) error {
+	if url == "" {
+		// lookup key management server URL from singularity.conf
```
tri-adam: Overriding with the `singularity.conf` value here might be confusing to the user. Let's say I run `singularity sign -h`, and I see:

```
  -u, --url string   specify the key server URL (default:
                     https://keys.sylabs.io)
```

So now I think, OK, that default is not what I want. I want it to go to `https://otherkeys.sylabs.io`, so I do `singularity sign -u https://otherkeys.sylabs.io`. Now, unbeknownst to me, `singularity.conf` has `https://yetanotherkeys.sylabs.io` in it. I run that, and it sends my key to `https://yetanotherkeys.sylabs.io`.

I think the behaviour I would expect in the above scenario, is that the value in `singularity.conf` would affect the _default_, but then if I explicitly set `-u`, it overrides that default. Does that make sense?

src/cmd/singularity/cli/verify.go
```diff
@@ -17,19 +17,20 @@ import (
 
 func init() {
 	VerifyCmd.Flags().SetInterspersed(false)
+	VerifyCmd.Flags().StringVarP(&keyServerURL, "url", "u", "", "specify the key server URL (default: https://keys.sylabs.io)")
```
tri-adam: Suggest same approach here, use the built in default capability of `StringVarP`.

src/cmd/singularity/cli/verify.go
```diff
@@ -40,3 +41,18 @@ var VerifyCmd = &cobra.Command{
 	Long:    docs.VerifyLong,
 	Example: docs.VerifyExample,
 }
+
+func doVerifyCmd(cpath, url string) error {
+	if url == "" {
+		// lookup key management server URL from singularity.conf
+
+		// else use default builtin
+		url = defaultKeysServer
```
tri-adam: Same user confusion comment here. Suggest putting this logic in `init()`.

src/pkg/signing/signing.go
```diff
@@ -78,25 +64,49 @@ func sifAddSignature(fingerprint [20]byte, fimg *sif.FileImage, signature []byte
...
+		}
+		if resp == "" || resp == "y" || resp == "Y" {
+			entity, err = sypgp.GenKeyPair()
+			if err != nil {
+				return err
+			}
+		} else {
+			return fmt.Errorf("cannot sign without installed keys")
+		}
+		resp, err = sypgp.AskQuestion("Upload public key %X to key server? [Y/n] ", entity.PrimaryKey.Fingerprint)
```
tri-adam: It would be nice to see the URL that will be used in this message.

src/cmd/singularity/cli/keys_search.go
tri-adam: Should the `search` also read key server default from config file like `sign` and `verify`?

src/cmd/singularity/cli/keys_push.go
tri-adam: Should the `push` also read key server default from config file like `sign` and `verify`?

src/cmd/singularity/cli/keys_pull.go
tri-adam: Should `pull` also read key server default from config file like `sign` and `verify`?

tri-adam (changes requested):

src/cmd/singularity/cli/keys_pull.go
yhcote: This is removed completely see my comment above.

src/cmd/singularity/cli/keys_push.go
yhcote: This is removed completely see my comment above.

src/cmd/singularity/cli/keys_search.go
yhcote: This is removed completely see my comment above.

src/cmd/singularity/cli/verify.go
yhcote: This is removed completely see my comment above.

src/cmd/singularity/cli/sign.go
yhcote: Yeah, that logic is not done. Only a comment as a reminder that maybe we wanna setup a list of key servers in singularity.conf. I will remove that entire block for now since default + the usage of -u is functional until we decide on a format to store a list of keys in singularity.conf or a file similar.

src/cmd/singularity/cli/keys_pull.go
yhcote: Didn't even realize from looking at other commands help that the default output came from printing the string var with Cobra! Alright.

src/pkg/signing/signing.go
yhcote: Great now it's wrapping to the next output line ;-) Done.

src/cmd/singularity/cli/sign.go
```diff
@@ -17,26 +17,36 @@ import (
...
 	SignCmd.Flags().SetInterspersed(false)
+	SignCmd.Flags().StringVarP(&keyServerURL, "url", "u", defaultKeysServer, "specify the key server URL")
 	SingularityCmd.AddCommand(SignCmd)
 }
 
 // SignCmd singularity sign
 var SignCmd = &cobra.Command{
 	DisableFlagsInUseLine: true,
-	Args:   cobra.ExactArgs(1),
+	Args:   cobra.RangeArgs(1, 2),
```
tri-adam: I think this should be `cobra.ExactArgs(1)`, as `sign` only takes one positional argument.

src/cmd/singularity/cli/verify.go
```diff
@@ -17,19 +17,20 @@ import (
...
 	VerifyCmd.Flags().SetInterspersed(false)
+	VerifyCmd.Flags().StringVarP(&keyServerURL, "url", "u", defaultKeysServer, "specify the key server URL")
 	SingularityCmd.AddCommand(VerifyCmd)
 }
 
 // VerifyCmd singularity verify
 var VerifyCmd = &cobra.Command{
 	DisableFlagsInUseLine: true,
-	Args:   cobra.ExactArgs(1),
+	Args:   cobra.RangeArgs(1, 2),
```
tri-adam: I think this should be `cobra.ExactArgs(1)`, as `verify` only takes one positional argument.

src/cmd/singularity/cli/keys_search.go
```diff
@@ -27,7 +27,7 @@ var KeysSearchCmd = &cobra.Command{
 	DisableFlagsInUseLine: true,
```
tri-adam, line 27: Github won't let me comment on the correct line, but think this command should be `cobra.ExactArgs(1)` also.

src/cmd/singularity/cli/keys_push.go
```diff
@@ -27,7 +27,7 @@ var KeysPushCmd = &cobra.Command{
 	DisableFlagsInUseLine: true,
```
tri-adam, line 27: Same here, should be `cobra.ExactArgs(1)`.

src/cmd/singularity/cli/keys_pull.go
```diff
@@ -27,7 +27,7 @@ var KeysPullCmd = &cobra.Command{
 	DisableFlagsInUseLine: true,
```
tri-adam, line 27: Same here, should be `cobra.ExactArgs(1)`.

tri-adam (changes requested):
OK, last round I think, just fixing up `cobra.Command.Args`. Sorry I didn't catch these last round!

yhcote:
We should clean up the Cobra code across. I am pretty sure I was inspired by code we already committed ;)

tri-adam:
Agreed!

tri-adam (approved):

yhcote:
Yeah I found another one, keys_list should have been ExactArgs(0). Seems all good now.

PR #1869, opened 2018-08-22
Cleanup of code and prepare for id, groupid as argument

yhcote:
This PR is another small step to make the code clearer and put in place functions like descrToSign() and getSigsForSelection() to later implement features allowing argument or interactive id, groupid selection.

### DIFF
--- a/src/pkg/signing/signing.go
+++ b/src/pkg/signing/signing.go
@@ -198,6 +198,7 @@ func Verify(cpath, url, authToken string) error {
 	}
 
 	// compare freshly computed hash with hashes stored in signatures block(s)
+	var authok string
 	for _, v := range signatures {
 		// Extract hash string from signature block
 		data := v.GetData(&fimg)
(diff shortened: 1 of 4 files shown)


src/pkg/signing/signing.go
```diff
@@ -21,55 +21,63 @@ import (
...
 }
 
+// descrToSign determines via argument or interactively which descriptor to sign
+func descrToSign(fimg *sif.FileImage) (descr *sif.Descriptor, err error) {
+	descr, _, err = fimg.GetPartPrimSys()
+	if err != nil {
+		return
+	}
+
+	return
```
tri-adam, line 66: Good candidate for a naked return:

```
	descr, _, err = fimg.GetPartPrimSys()
	return
```

tri-adam (approved):
One minor suggestion, take it of leave it.

src/pkg/signing/signing.go
yhcote: Already in the next PR :)

PR #1870, opened 2018-08-22
Pull uris

ArangoGutierrez:

To make interacting with shub easier, implement a src/pkg/shub/client (analogous to `src/pkg/library/client). For now, this only should contain the Pull API handling.

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/src/pkg/shub/client/api.go
+++ b/src/pkg/shub/client/api.go
@@ -20,7 +20,7 @@ import (
 )
 
 const (
-	defaultRegistry string = `singularity-hub.org/api/container/`
+	defaultRegistry string = `singularity-hub.org`
 	shubAPIRoute    string = "/api/container/"
 	//URINotSupported if we are using a non default registry error out for now
 	URINotSupported string = "Only the default Singularity Hub registry is suported for now"
@@ -28,12 +28,11 @@ const (
 
 // ShubURI stores the various components of a singularityhub URI
 type ShubURI struct {
-	registry   string
-	user       string
-	container  string
-	tag        string
-	digest     string
-	defaultReg bool
+	registry  string
+	user      string
+	container string
+	tag       string
+	digest    string
 }
 
 func (s *ShubURI) String() string {
@@ -57,7 +56,7 @@ func getManifest(uri ShubURI) (manifest ShubAPIResponse, err error) {
 		Timeout: 30 * time.Second,
 	}
 
-	if !uri.defaultReg {
+	if uri.registry != fmt.Sprintf("%s%s", defaultRegistry, shubAPIRoute) {
 		return ShubAPIResponse{}, errors.New(URINotSupported)
 	}
 
--- a/src/pkg/shub/client/api.go
+++ b/src/pkg/shub/client/api.go
@@ -55,6 +55,7 @@ func getManifest(uri ShubURI) (manifest ShubAPIResponse, err error) {
 		Timeout: 30 * time.Second,
 	}
 
+	// TODO: support custom registry URI's
 	if uri.registry != fmt.Sprintf("%s%s", defaultRegistry, shubAPIRoute) {
 		return ShubAPIResponse{}, errors.New(URINotSupported)
 	}
@@ -77,7 +78,7 @@ func getManifest(uri ShubURI) (manifest ShubAPIResponse, err error) {
 
 	// Do the request, if status isn't success, return error
 	res, err := sc.Do(req)
-	sylog.Debugf("response: %v\n", res)
+	sylog.Debugf("%s response received, beginning body download\n", res.Status)
 
 	if err != nil {
 		return ShubAPIResponse{}, err
@@ -95,7 +96,7 @@ func getManifest(uri ShubURI) (manifest ShubAPIResponse, err error) {
 	}
 
 	err = json.Unmarshal(body, &manifest)
-	sylog.Debugf("manifest: %v\n", manifest.Image)
+	sylog.Debugf("manifest image name: %v\n", manifest.Name)
 	if err != nil {
 		return ShubAPIResponse{}, err
 	}
--- a/src/pkg/shub/client/api.go
+++ b/src/pkg/shub/client/api.go
@@ -21,6 +21,9 @@ import (
 
 const (
 	defaultRegistry string = `singularity-hub.org/api/container/`
+	shubAPIRoute    string = "/api/container/"
+	//URINotSupported if we are using a non default registry error out for now
+	URINotSupported string = "Only the default Singularity Hub registry is suported for now"
 )
 
 // ShubURI stores the various components of a singularityhub URI
@@ -33,6 +36,10 @@ type ShubURI struct {
 	defaultReg bool
 }
 
+func (s *ShubURI) String() string {
+	return s.registry + s.user + "/" + s.container + s.tag + s.digest
+}
+
 // ShubAPIResponse holds the information returned from the Shub API
 type ShubAPIResponse struct {
 	Image   string `json:"image"`
@@ -41,30 +48,21 @@ type ShubAPIResponse struct {
 	Version string `json:"version"`
 }
 
-// ShubClient holds the information for interacting with Singularity Hub API
-type ShubClient struct {
-	FilePath string
-	ShubURI
-	*ShubAPIResponse
-	ShubURL string
-}
-
 // getManifest will return the image manifest for a container uri
 // from Singularity Hub.
-func (s *ShubClient) getManifest() (err error) {
+func getManifest(uri ShubURI) (manifest ShubAPIResponse, err error) {
 
 	// Create a new Singularity Hub client
 	sc := http.Client{
 		Timeout: 30 * time.Second,
 	}
 
-	//if we are using a non default registry error out for now
-	if !s.ShubURI.defaultReg {
-		return errors.New("Only the default Singularity Hub registry is suported for now")
+	if !uri.defaultReg {
+		return ShubAPIResponse{}, errors.New(URINotSupported)
 	}
 
 	// Format the http address, coinciding with the image uri
-	httpAddr := fmt.Sprintf("www.%s", s.ShubURI.String())
+	httpAddr := fmt.Sprintf("www.%s", uri.String())
 
 	// Create the request, add headers context
 	url := url.URL{
@@ -75,7 +73,7 @@ func (s *ShubClient) getManifest() (err error) {
 
 	req, err := http.NewRequest(http.MethodGet, url.String(), nil)
 	if err != nil {
-		return err
+		return ShubAPIResponse{}, err
 	}
 	req.Header.Set("User-Agent", useragent.Value)
 
@@ -84,25 +82,25 @@ func (s *ShubClient) getManifest() (err error) {
 	sylog.Debugf("response: %v\n", res)
 
 	if err != nil {
-		return err
+		return ShubAPIResponse{}, err
 	}
 	defer res.Body.Close()
 
 	if res.StatusCode != http.StatusOK {
 		err = errors.New(res.Status)
-		return err
+		return ShubAPIResponse{}, err
 	}
 
 	body, err := ioutil.ReadAll(res.Body)
 	if err != nil {
-		return err
+		return ShubAPIResponse{}, err
 	}
 
-	err = json.Unmarshal(body, &s.ShubAPIResponse)
-	sylog.Debugf("manifest: %v\n", s.ShubAPIResponse.Image)
+	err = json.Unmarshal(body, &manifest)
+	sylog.Debugf("manifest: %v\n", manifest.Image)
 	if err != nil {
-		return err
+		return ShubAPIResponse{}, err
 	}
 
-	return nil
+	return
 }
--- a/src/pkg/shub/client/api.go
+++ b/src/pkg/shub/client/api.go
@@ -9,15 +9,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"net/http"
 	"net/url"
-	"os"
 	"strings"
 	"time"
 
-	sytypes "github.com/singularityware/singularity/src/pkg/build/types"
 	"github.com/singularityware/singularity/src/pkg/sylog"
 	"github.com/singularityware/singularity/src/pkg/util/user-agent"
 )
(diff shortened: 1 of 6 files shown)


src/pkg/shub/client/util_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+// TestShubPacker checks if we can create a Bundle from the pulled image
+func TestShubParser(t *testing.T) {
+	test.DropPrivilege(t)
+	defer test.ResetPrivilege(t)
+
+	validShubURIs := []string{
+		`//username/container`,
+		`//username/container:tag`,
+		`//username/container@00000000000000000000000000000000`,
```
bauerm97: What makes this URI valid but makes the URI on line 37 invalid? I can't tell the difference between the two

ikaneshiro (dismissed):
Hey @ArangoGutierrez, I think you might be better off making `DownloadImage()` a function that takes in a download destination as an additional argument and creates the `ShubClient` struct within that function in order to call the methods in `api.go`. From there I think it would be best to download directly to that location instead of making and downloading into `TmpFile` within a `Bundle`. I think this will help streamline this pulling process

src/pkg/shub/client/util_test.go
ikaneshiro: The digest on this one is 32 digits while the one on 37 is 31 digits and the one on 38 is 33 digits

ikaneshiro (dismissed):
LGTM

src/pkg/shub/client/util.go
```diff
@@ -0,0 +1,80 @@
...
+	found := shubRegex.FindString(src)
+
+	//sanity check
+	//if found string is not equal to the input, input isn't a valid URI
+	if strings.Compare(src, found) != 0 {
+		return uri, fmt.Errorf("Source string is not a valid URI: %s", src)
+	}
+
+	//strip `//` from start of src
+	src = src[2:]
```
dtrudg: Needing to do this kinda indicates to me that the wrong thing is being used as the ShubURI here. The `shub://` portion of the URI is the protocol.... then everything else is the path. 

If it's a URI then it should have `shub://` on the front. Or it should be a path that does not have any of that on the front. The `//` shouldn't be kept with the path, but all of `shub://` should be kept for a URI.

src/pkg/shub/client/api.go
```diff
@@ -0,0 +1,108 @@
...
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/singularityware/singularity/src/pkg/sylog"
+	"github.com/singularityware/singularity/src/pkg/util/user-agent"
+)
+
+const (
+	defaultRegistry string = `singularity-hub.org/api/container/`
```
dtrudg: Is `/api/container` always part of the URL to access the API endpoints? If so is the registry really `singularity-hub.org` and `/api/container` can be appended separately?

src/pkg/shub/client/api.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+// ShubURI stores the various components of a singularityhub URI
+type ShubURI struct {
+	registry   string
+	user       string
+	container  string
+	tag        string
+	digest     string
+	defaultReg bool
```
dtrudg: It seems a little odd having a bool for a default here. We know by virtue of `registry` `==` the `const defaultRegistry`?

src/pkg/shub/client/api.go
```diff
@@ -0,0 +1,108 @@
...
+// ShubAPIResponse holds the information returned from the Shub API
+type ShubAPIResponse struct {
+	Image   string `json:"image"`
+	Name    string `json:"name"`
+	Tag     string `json:"tag"`
+	Version string `json:"version"`
+}
+
+// ShubClient holds the information for interacting with Singularity Hub API
+type ShubClient struct {
```
dtrudg: The ShubClient struct here seems sort of in-between 2 approaches of providing the functionality needed to interact with shub.

As it is, `ShubClient` only implements `getManifest()` which returns an image manifest. Functionality for download of an actual image file is outside of `ShubClient` in `DownloadImage`. I think that it makes more sense if either:

- All the API / download calls against the shub service are written in functions on `ShubClient` so that e.g. `DownloadImage` is just a thin wrapper function. -or-
- You may not really need a ShubClient type.

In addition, when thinking about a client, I generally think about it being something that could be used to send multiple requests, so the `FilePath` might not really be a property of the client - but rather something passed in in a function call. Similarly, for a `getManifest()` call I'd expect to have to provide the URI. As is this client can only be used for a specific image. It's a more common pattern to have a client associated with a particular registry, and provide functions that can operate on anything in that registy.

src/pkg/shub/client/api.go
```diff
@@ -0,0 +1,108 @@
...
+	FilePath string
+	ShubURI
```
dtrudg: Having ShubURI and ShubURL here is a bit confusing - can the distinction be made clearer in the naming?

src/pkg/shub/client/pull.go
```diff
@@ -0,0 +1,108 @@
...
+	"net/http"
+	"os"
+
+	util "github.com/singularityware/singularity/src/pkg/library/client"
+	"github.com/singularityware/singularity/src/pkg/sylog"
+	pb "gopkg.in/cheggaaa/pb.v1"
+)
+
+// Timeout for an image pull in seconds
+const pullTimeout = 1800
```
dtrudg: Maybe 30 mins is not always be enough, in unusual circumstances. What are the maximum sizes on shub - do you know if there is a server-side timeout at all to match?

src/pkg/shub/client/util.go
```diff
@@ -0,0 +1,80 @@
...
+	"strings"
+)
+
+// ShubParseReference accepts a Shub reference string and parses its content
+// It will return an error if the given URI is not valid,
+// otherwise it will parse the contents into a ShubURI struct
+func ShubParseReference(src string) (uri ShubURI, err error) {
+
+	//define regex for each URI component
+	registryRegexp := `([-a-zA-Z0-9/]{1,64}\/)?` //target is very open, outside registry
```
dtrudg: should registry be able to be a dns name here? so it needs to allow periods?

i.e. `shub://myprivateregistry.sylabs.io/dtrudg/alpine:latest`

src/pkg/shub/client/util.go
```diff
@@ -0,0 +1,80 @@
...
+	nameRegexp := `([-a-zA-Z0-9]{1,39}\/)`       //target valid github usernames
+	containerRegexp := `([-_.a-zA-Z0-9]{1,64})`  //target valid github repo names
+	tagRegexp := `(:[-_.a-zA-Z0-9]{1,64})?`      //target is very open, file extensions or branch names
+	digestRegexp := `(\@[a-f0-9]{32})?`          //target md5 sum hash
```
dtrudg: After an '@' may not be an md5 hash:

https://github.com/singularityhub/singularityhub.github.io/wiki/Deploy#order-of-operations

> shub://vsoch/hello-world:@ is where it get's interesting. A digest is flexible to being the following, in the order of preference:
> hash the most specific kind of digest is the hash of the file. This is checked for first.
> commit the next most specific is the commit
> branch finally, if a matching hash and commit are not found, we fall back to the branch.

dtrudg (dismissed):

src/pkg/shub/client/pull.go
ArangoGutierrez: well, if you also think in low bandwidths like mine, yeah, you could need more than an hour.

src/pkg/shub/client/util.go
ArangoGutierrez: you mean just to adjust the comment right?

src/pkg/shub/client/pull.go
ArangoGutierrez: my proud 15 Mb WIFI ;P

src/pkg/shub/client/util.go
dtrudg: I mean currently this doesn't support all shub URIs... you can only pull `xxxx@<hash>` not `xxx@<commit>` or `xxxx@<branch>`

ArangoGutierrez:
@dctrud & @ikaneshiro  Ready for a next round

src/pkg/shub/client/api.go
```diff
@@ -0,0 +1,105 @@
...
+
+	if uri.registry != fmt.Sprintf("%s%s", defaultRegistry, shubAPIRoute) {
+		return ShubAPIResponse{}, errors.New(URINotSupported)
+	}
+
+	// Format the http address, coinciding with the image uri
+	httpAddr := fmt.Sprintf("www.%s", uri.String())
+
+	// Create the request, add headers context
+	url := url.URL{
```
dtrudg: Probably better to use URL.Parse here instead... it will catch weird stuff this manual work may not.. https://golang.org/pkg/net/url/#URL.Parse

src/pkg/shub/client/api.go
```diff
@@ -0,0 +1,105 @@
...
+		Scheme: "https",
+		Host:   strings.Split(httpAddr, `/`)[0],     //split url to match format, first half
+		Path:   strings.SplitN(httpAddr, `/`, 2)[1], //second half
+	}
+
+	req, err := http.NewRequest(http.MethodGet, url.String(), nil)
```
dtrudg: ... but it's a little odd to parse the string into a URL then immediately use it as a string. Is this error checking?

src/pkg/shub/client/api.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		return ShubAPIResponse{}, err
+	}
+	req.Header.Set("User-Agent", useragent.Value)
+
+	// Do the request, if status isn't success, return error
+	res, err := sc.Do(req)
+	sylog.Debugf("response: %v\n", res)
```
dtrudg: Dumping the response straight out can be a bit ugly.... like

```
response: &{200 OK 200 HTTP/1.1 1 1 map[Server:[nginx/1.13.5] Date:[Fri, 24 Aug 2018 15:27:43 GMT] Content-Type:[application/json] Content-Length:[480] Connection:[keep-alive] Vary:[Accept, Cookie] Allow:[GET, HEAD, OPTIONS] X-Frame-Options:[SAMEORIGIN]] 0xc420260780 480 [] false false map[] 0xc42014c400 0xc420379ad0}
```

What's the important information for debugging here? Maybe only show those bits?

src/pkg/shub/client/api.go
```diff
@@ -0,0 +1,105 @@
...
+		return ShubAPIResponse{}, err
+	}
+
+	body, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		return ShubAPIResponse{}, err
+	}
+
+	err = json.Unmarshal(body, &manifest)
+	sylog.Debugf("manifest: %v\n", manifest.Image)
```
dtrudg: in the debug message might want to notate this as `manifest image` if that's what is being shown

src/pkg/shub/client/pull.go
```diff
@@ -0,0 +1,111 @@
...
+
+	// Get the image manifest
+	manifest, err := getManifest(ShubURI)
+	if err != nil {
+		sylog.Fatalf("Failed to get manifest from Shub: %v", err)
+		return
+	}
+
+	// Get the image based on the manifest
+	resp, err := http.Get(manifest.Image)
```
dtrudg: We should probably use the user-agent for this request too? It is being used when getting the manifest.

src/pkg/shub/client/pull.go
```diff
@@ -0,0 +1,111 @@
...
+	sylog.Debugf("Downloading container from Shub")
+
+	//use custom parser to make sure we have a valid shub URI
+	if ok := isShubPullRef(shubRef); !ok {
+		sylog.Fatalf("Invalid shub URI: %v", err)
+	}
+
+	ShubURI, err := shubParseReference(shubRef)
+	if err != nil {
+		sylog.Fatalf("Failed to parse shub URI: %v", err)
```
dtrudg: This is a `sylog.Fatalf`but further down we are doing `return fmt.Errorf` - should this be consistent?

src/pkg/shub/client/pull.go
```diff
@@ -0,0 +1,111 @@
...
+	if !force {
+		if _, err := os.Stat(filePath); err == nil {
+			return fmt.Errorf("image file already exists - will not overwrite")
+		}
+	}
+
+	// Get the image manifest
+	manifest, err := getManifest(ShubURI)
+	if err != nil {
+		sylog.Fatalf("Failed to get manifest from Shub: %v", err)
```
dtrudg: This is a sylog.Fatalf but further elsewhere we are doing return fmt.Errorf - should this be consistent?

src/pkg/shub/client/pull_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"os"
+	"testing"
+
+	"github.com/singularityware/singularity/src/pkg/test"
+)
+
+const (
+	shubURI     = "shub://ikaneshiro/singularityhub:latest"
```
dtrudg, line 16: Can we test retrieval using the `@digest` for as well, since we are supporting that in the URI scheme?

src/pkg/shub/client/util.go
```diff
@@ -0,0 +1,80 @@
...
+
+// isShubPullRef returns true if the provided string is a valid Shub
+// reference for a pull operation.
+func isShubPullRef(shubRef string) bool {
+	//define regex for each URI component
+	registryRegexp := `([-.a-zA-Z0-9/]{1,64}\/)?` //target is very open, outside registry
+	nameRegexp := `([-a-zA-Z0-9]{1,39}\/)`        //target valid github usernames
+	containerRegexp := `([-_.a-zA-Z0-9]{1,64})`   //target valid github repo names
+	tagRegexp := `(:[-_.a-zA-Z0-9]{1,64})?`       //target is very open, file extensions or branch names
+	digestRegexp := `(\@[a-f0-9]{32})?`           //target file md5 has, git commit hash, git branch
```
dtrudg: The branch or commit hash will not match this regex. An unabbreviated SHA1 hash is `{40}` and a branch name is much more open.

src/pkg/shub/client/util_test.go
```diff
@@ -0,0 +1,83 @@
...
+	test.DropPrivilege(t)
+	defer test.ResetPrivilege(t)
+
+	invalidShubURIs := []string{
+		`shub://username/`,
+		`shub://username/container:`,
+		`shub://username/container@`,
+		`shub://username/container@0000000000000000000000000000000`,
+		`shub://username/container@000000000000000000000000000000000`,
+		`shub://username/container@abcdefghijklmnopqrstuvwxyz123456`,
```
dtrudg, line 42: This one would be a valid `@<branch>`

dtrudg (changes requested):
Hi @ArangoGutierrez - think the structure here now is a lot cleaner to read through. A few minor things highlighted now.

src/pkg/shub/client/api.go
ArangoGutierrez: Agree

src/pkg/shub/client/api.go
ArangoGutierrez: agree

src/pkg/shub/client/util_test.go
ArangoGutierrez: is not valid due you need a `:` before the `@`

src/pkg/shub/client/util.go
ArangoGutierrez: I'll set it for SHA1 for now

src/pkg/shub/client/pull_test.go
ArangoGutierrez: Will issue it as a TODO, it will add more download steps to the test process

src/pkg/shub/client/util_test.go
dtrudg: gotcha!

src/pkg/shub/client/util.go
```diff
@@ -19,7 +19,7 @@ func isShubPullRef(shubRef string) bool {
 	nameRegexp := `([-a-zA-Z0-9]{1,39}\/)`        //target valid github usernames
 	containerRegexp := `([-_.a-zA-Z0-9]{1,64})`   //target valid github repo names
 	tagRegexp := `(:[-_.a-zA-Z0-9]{1,64})?`       //target is very open, file extensions or branch names
-	digestRegexp := `(\@[a-f0-9]{32})?`           //target file md5 has, git commit hash, git branch
+	digestRegexp := `(\@[a-f0-9]{40})?`           //target file md5 has, git commit hash, git branch
```
dtrudg: Hey ed... this would need to allow {32} for the md5 *or* {40} for a sha1 commit

dtrudg (dismissed):

bauerm97:
@ArangoGutierrez Travis checks are failing here, can you fix that and then I'll go ahead and merge!

bauerm97 (approved):

Issue #1883, opened 2018-08-28
Need ability to transpose env vars into container from host


GodloveD:
## Version of Singularity:

```
$ singularity --version
singularity version 3.0.0-alpha.1-117-gc297513f
```

### Expected behavior

```
$ export SINGULARITYENV_FOO=bar

$ singularity exec my_container env | grep FOO
FOO=bar
```

### Actual behavior

```
$ export SINGULARITYENV_FOO=bar

$ singularity exec my_container env | grep FOO
SINGULARITYENV_FOO=bar
```

### Steps to reproduce behavior

Try to transpose variables into the container.

bauerm97:
#1898 Should have resolved this particular env var concern.

PR #1898, opened 2018-08-29
Issue 1883

ArangoGutierrez:

Fix: 
> Need ability to transpose env vars into container from host #1883 

**Checkoff for all PRs:**

Attn: @singularityware-admin


src/cmd/singularity/cli/actions.go
```diff
@@ -274,6 +282,14 @@ func execWrapper(cobraCmd *cobra.Command, image string, args []string, name stri
 		}
 	}
 
+	// Transpose environment
+	for _, e := range environment {
```
bauerm97: Not sure why this is being done in two different loops, this adds unnecessary complexity

src/cmd/singularity/cli/actions.go
```diff
@@ -255,13 +255,21 @@ func execWrapper(cobraCmd *cobra.Command, image string, args []string, name stri
...
-		for _, env := range os.Environ() {
+		for _, env := range environment {
 			e := strings.SplitN(env, "=", 2)
 			if len(e) != 2 {
 				sylog.Verbosef("can't process environment variable %s", env)
 				continue
 			}
+			// skip any variables prefixed with "SINGULARITYENV_"
+			if strings.HasPrefix(e[0], "SINGULARITYENV_") {
+				continue
```
bauerm97: I don't think we want to continue when we find the `SINGULARITYENV_` prefixed variables. IIRC, we *only* want to pass through variables which are prefixed with `SINGULARITYENV_` (and also any special cases such as `HOME`).

bauerm97 (changes requested):

src/cmd/singularity/cli/actions.go
ArangoGutierrez: the skip is to avoid something like
```
[eduardo@rhel7 singularity]$ export SINGULARITYENV_ISSUE=1883
[eduardo@rhel7 singularity]$ ./builddir/singularity exec /tmp/alpine_3.7.sif env | grep ISSUE
SINGULARITYENV_ISSUE=1883
ISSUE=1883
```

src/cmd/singularity/cli/actions.go
```diff
@@ -255,22 +255,34 @@ func execWrapper(cobraCmd *cobra.Command, image string, args []string, name stri
...
+	for _, env := range environment {
+		e := strings.SplitN(env, "=", 2)
+		if len(e) != 2 {
+			sylog.Verbosef("can't process environment variable %s", env)
+			continue
+		}
+
+		// Transpose environment
+		if strings.HasPrefix(env, "SINGULARITYENV_") {
+			te := strings.Split(strings.TrimPrefix(env, "SINGULARITYENV_"), "=")
```
bauerm97: Here we should handle checking the len of the split string. Also, we should just do:

```
e[0] = te[0]
```

and not write `continue`. This way the line 285 code still handles this. We're just simply changing the key

bauerm97 (changes requested):

ArangoGutierrez:
@bauerm97 RFR next round

src/cmd/singularity/cli/actions.go
```diff
@@ -255,22 +255,34 @@ func execWrapper(cobraCmd *cobra.Command, image string, args []string, name stri
...
+	// Clean environment
+	for _, env := range environment {
+		e := strings.SplitN(env, "=", 2)
+		if len(e) != 2 {
+			sylog.Verbosef("can't process environment variable %s", env)
+			continue
+		}
+
+		// Transpose environment
+		if strings.HasPrefix(env, "SINGULARITYENV_") {
```
bauerm97: Actually, still seems like we're doing more work than we need to here. 

```
if strings.HasPrefix(e[0], "SINGULARITYENV_") {
    e[0] = strings.TrimPrefix(e[0], "SINGULARITYENV_")
} else if IsCleanEnv {
...
```

Should be easy enough

bauerm97 (changes requested):

bauerm97 (approved):

PR #1900, opened 2018-08-29
Attempt directory creation without OverlayFS

jmstover:
Try creating directories with `--contain` even if OverlayFS isn't
supported. The reason for this is because /dev at this point is
a tmpfs, and we can write to it.

So say, for example, we want /dev/infiniband on a system that
doesn't support OverlayFS, we can do:

  `singularity exec -c -B /dev/infiniband [image] [a b c]`

And it will work.

fixing or enhancing.

**Checkoff for all PRs:**

- [] This PR is against the project's master branch

Attn: @singularityware-admin

### DIFF
--- a/src/lib/runtime/mounts/userbinds/userbinds.c
+++ b/src/lib/runtime/mounts/userbinds/userbinds.c
@@ -118,10 +118,10 @@ int _singularity_runtime_mount_userbinds(void) {
                     continue;
                 }
             } else if ( ( is_dir(source) == 0 ) && ( is_dir(joinpath(container_dir, dest)) < 0 ) ) {
-                if ( singularity_registry_get("OVERLAYFS_ENABLED") != NULL ) {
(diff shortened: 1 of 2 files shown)


src/lib/runtime/mounts/userbinds/userbinds.c
```diff
@@ -118,10 +118,10 @@ int _singularity_runtime_mount_userbinds(void) {
                     continue;
                 }
             } else if ( ( is_dir(source) == 0 ) && ( is_dir(joinpath(container_dir, dest)) < 0 ) ) {
-                if ( singularity_registry_get("OVERLAYFS_ENABLED") != NULL ) {
-                    singularity_message(VERBOSE3, "Creating bind directory on overlay file system: %s\n", dest);
+                if ( singularity_registry_get("OVERLAYFS_ENABLED") != NULL || singularity_registry_get("CONTAIN") != NULL ) {
```
cclerget, line 121: The problem with this check is when used with directory images, those images could be altered by users with new directories even if they don't own images

Issue #1891, opened 2018-08-28
the --home option is not working as expected (doesn't cd to $HOME when started)


GodloveD:
## Version of Singularity:

```
$ singularity --version
singularity version 3.0.0-alpha.1-117-gc297513f
```

### Expected behavior

```
$ ls /data
foo

$ singularity shell --home /data my_container
Singularity: Invoking an interactive shell within container...

Singularity my_container:~> pwd
/data
Singularity my_container:~> echo $HOME
/data
Singularity my_container:~> ls .
foo
```

### Actual behavior

```
$ singularity shell --home /data my_container 
Singularity :/home/david/sandbox> pwd
/home/david/sandbox
Singularity :/home/david/sandbox> echo $HOME
/data
Singularity :/home/david/sandbox> ls .
blah  lolcow  my_container  sections.def  ubuntu.def
```

### Steps to reproduce behavior

Use the home option and check your current working directory.

cclerget:
Fixed by #1905

Issue #1894, opened 2018-08-28
the --contain option is not containing home


GodloveD:
## Version of Singularity:
```
$ singularity --version
singularity version 3.0.0-alpha.1-117-gc297513f
```
### Expected behavior
This is how 2.6 works
```
$ singularity exec --contain my_container ls -l ~
total 0
```
### Actual behavior
But this is how 3. 0 works
```
$ singularity exec --contain my_container ls -l ~
total 116
drwxr-xr-x  2 david david  4096 May  9 15:16 Desktop
drwxr-xr-x  2 david david  4096 May  9 15:16 Documents
drwxr-xr-x  2 david david  4096 May  9 15:16 Downloads
drwxr-xr-x  2 david david  4096 May  9 15:16 Music
drwxr-xr-x  2 david david  4096 May  9 15:16 Pictures
drwxr-xr-x  2 david david  4096 May  9 15:16 Public
drwxr-xr-x  2 david david  4096 May  9 15:16 Templates
drwxr-xr-x  2 david david  4096 May  9 15:16 Videos
[...snip...]
```

### Steps to reproduce behavior

```
$ singularity exec --contain my_container ls -l ~
```

cclerget:
Fixed by #1905

PR #1905, opened 2018-08-30
Development 3.0: Fix home mount with contain

cclerget:

Fix issue #1891 and #1894


src/runtime/engines/singularity/container.go
```diff
@@ -838,43 +838,45 @@ func (c *container) addBindsMount(system *mount.System) error {
...
 	}
 
-	return nil
+	return source, dest, err
 }
 
-// getHomePaths returns the source and destination path of the requested home mount
-func (c *container) getHomePaths() (source string, dest string, err error) {
-	homeSlice := strings.Split(c.engine.EngineConfig.GetHomeDir(), ":")
+// addHomeStagingDir adds the home staging directory in session directory and
```
ArangoGutierrez: is this comment incomplete?

ArangoGutierrez (dismissed):
tested

src/runtime/engines/singularity/container.go
cclerget: fixed

GodloveD (changes requested):
This actually doesn't seem to fix #1891 

```
10:01 PM 
ubuntu ~/sandbox
$ singularity --version
singularity version 20758c4d

10:01 PM 
ubuntu ~/sandbox
$ singularity shell --home /data my_container
Singularity :/home/david/sandbox> pwd
/home/david/sandbox
```
In Singularity 2.6 that `pwd` command returns `/data`

cclerget:
@GodloveD Actually the code mount custom home but doesn't chdir into it since the current working directory is mounted with :
`cd $HOME/test && singularity shell --home /tmp/home my_container`

For the behaviour you described :

`cd $HOME/test && singularity shell --home /tmp/home:$HOME my_container`

I'm wondering if this is a bug to mount cwd with `--home` as long as cwd doesn't conflict with container home ?

GodloveD:
Oh I see.  Maybe this is just a bug in 2.6 then and we should _not_ `cd` to `$HOME` when we startup.  After all, we don't `cd` to `$HOME` when we start singularity without the `--home` option.  So that makes sense.  

But now I think the behavior is confusing the other way around.  Why do we `cd` to `$HOME` with this option argument pair?

```
--home /tmp/home:$HOME
```

Shouldn't we just stay in the `cwd` in that case too?  To be clear, the difference between these two commands doesn't make sense to me.

```
07:48 AM 
ubuntu ~/go/src/github.com/singularityware/singularity
$ cd ~/test

07:48 AM 
ubuntu ~/test
$ singularity exec --home /data:/mnt ~/sandbox/my_container pwd
/home/david/test

07:48 AM 
ubuntu ~/test
$ singularity exec --home /data:$HOME ~/sandbox/my_container pwd
/home/david
```

cclerget:
It depends where you are when you run Singularity, in your example you have `CWD=/home/david/test`, if you run with `--home /data:/home/david`, Singularity couldn't chdir to cwd because `/data` doesn't contain `test` directory path in order to bind it, that's why Singularity chdir goes into `/home/david`, if you create `/data/test` it should work with `CWD=/home/david/test`.

I agree this is confusing, and that may require to define a default behaviour after discussion

jmstover:
I thought current behavior was this:

1) `cd` to `$CWD` on host if possible  (i.e. we try mounting `$CWD`)
2) `cd` to `$HOME`
3) `cd` to `/`

So, if the host `$CWD` location is available, it wins; Then we fall back to `$HOME`; and finally `/`

GodloveD:
@jmstover I think that _should_ be the behavior.  But it doesn't look like it is the behavior?

jmstover:
This is what I get with 2.6.0

```
$ cd /tmp
$ /usr/local/singularity/2.6.0/bin/singularity exec ~/tmp/sing/centos7.img pwd
/tmp
$ /usr/local/singularity/2.6.0/bin/singularity exec -c ~/tmp/sing/centos7.img pwd
/home/jason
$ /usr/local/singularity/2.6.0/bin/singularity exec -c --no-home ~/tmp/sing/centos7.img pwd
WARNING: Could not chdir to home: /home/jason
/
```

So, $CWD, $HOME, then /

bauerm97:
Just want to chime in and say that I agree with @jmstover here. I think that behavior should be:

1. `cd` to `$CWD`
2. `cd` to `$HOME`
3. `cd` to `/`

jmstover:
@GodloveD  ... This looks like it's working correctly to me:

```
$ cd /bin2
$ /usr/local/singularity/3.0.0-dev/bin/singularity exec ~/tmp/sing/alpine.sif pwd
/bin2
$ cd /tmp/
$ /usr/local/singularity/3.0.0-dev/bin/singularity exec -c ~/tmp/sing/alpine.sif pwd
/home/jason
$ /usr/local/singularity/3.0.0-dev/bin/singularity exec -c --no-home ~/tmp/sing/alpine.sif pwd
/
```

GodloveD:
I think you are right @jmstover.  In my second test above:

```
07:48 AM 
ubuntu ~/test
$ singularity exec --home /data:$HOME ~/sandbox/my_container pwd
/home/david
```

the `~/test` directory is no longer available and so it defaults to `$HOME`.  Thanks for clearing that up.

GodloveD (dismissed):

GodloveD:
@cclerget you've got two 👍 reviews now.  If you would like to resolve the conflicts and then merge, please feel free.  😸

jmstover (approved):

GodloveD (approved):
lgtm.  Thanks @ArangoGutierrez!

EDIT:  Whoops!  Put this approval and comment on the wrong PR I think.  This one is already good to go.

PR #1912, opened 2018-08-31
use shub client package to download image

ikaneshiro:
removes duplicate shub code that exists in the build package and uses the shub client package to perform image downloads


bauerm97 (approved):
Looks good

ArangoGutierrez (approved):

Issue #1893, opened 2018-08-28
--cleanenv seems to be making the environment TOO clean


GodloveD:
## Version of Singularity:

```
$ singularity --version
singularity version 3.0.0-alpha.1-117-gc297513f
```

### Expected behavior
This is 2.6
```
$ singularity exec --cleanenv my_container env
LD_LIBRARY_PATH=/.singularity.d/libs
HOME=/home/vagrant
PS1=Singularity>
TERM=xterm-256color
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
LANG=C
SINGULARITY_CONTAINER=my_container
PWD=/home/vagrant/sandbox
SINGULARITY_NAME=my_container
```

### Actual behavior
This is 3.0 alpha
```
$ singularity exec --cleanenv my_container env
LD_LIBRARY_PATH=/.singularity.d/libs
PS1=Singularity> 
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
PWD=/home/david/sandbox
```

### Steps to reproduce behavior

```
$ singularity exec --cleanenv my_container env
```

PR #1920, opened 2018-09-04
Fix cleanenv

ArangoGutierrez:

- Ref: 
--cleanenv seems to be making the environment TOO clean #1893

**Checkoff for all PRs:**

Attn: @singularityware-admin


src/pkg/util/env/clean.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/singularityware/singularity/src/pkg/sylog"
+)
+
+// CleanEnv cleans environment variables before running the container
+func CleanEnv(g *generate.Generator, IsCleanEnv bool, environment []string) {
+
+	for _, env := range environment {
+		e := strings.SplitN(env, "=", 2)
+		if len(e) != 2 {
+			sylog.Verbosef("can't process environment variable %s", env)
```
jmstover: Capitalize `Can't` here...

jmstover (dismissed):
This works. Only difference between this and 2.x is the 3.0 version doesn't set `LANG=C`, and also the variables `SINGULARITY_CONTAINER` and `SINGULARITY_NAME` aren't set.

jmstover:
This PR:
```
$ /usr/local/singularity/3.0.0-dev/bin/singularity exec --cleanenv alpine.sif env
SHLVL=1
HOME=/home/jason
PS1=Singularity> 
TERM=xterm
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
LANG=en_US
PWD=/home/jason/tmp/sing
```

2.6.0
```
$ /usr/local/singularity/2.6.0/bin/singularity exec --cleanenv alpine.img env
SHLVL=1
HOME=/home/jason
PS1=Singularity> 
TERM=xterm
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
LANG=C
SINGULARITY_CONTAINER=alpine.img
PWD=/home/jason/tmp/sing
SINGULARITY_NAME=alpine.img
```

yhcote:
Just to understand, we are transforming SINGULARITYENV_ envvars but let SINGULARITY_ envvars pass through?

jmstover:
@yhcote 
No. Any `SINGULARITY_` are removed, and `SINGULARITYENV_` are transformed.

In 2.x, `SINGULARITY_CONTAINER`/`SINGULARITY_NAME` got set differently, and so were passed through when `--cleanenv` was used.

yhcote:
Me again, 3.X doesn't depend on the same behavior? Those 2 vars being exposed that it.

jmstover:
Sort of... The once place I know it's used is in `/.singularity.d/actions/shell`  ... that's still in there where it's setting the PS1 value when the shell is Bash. 

But that behavior can be modified to something else.... or if we want to keep the same behavior then we can explicitly set the variables after the `env.CleanEnv()` call.

That behavior, to me, was something that could be hashed out after this Env cleaning was in place.

yhcote:
I think we should be careful to break workflows if it's not clear that it will be broken.

jmstover:
@GodloveD ... do you know of anywhere beyond the `shell` action where SINGULARITY_CONTAINER, or SINGULARITY_NAME are used?

I think SINGULARITY_NAME was _possibly_ used by one of the SSH hacks to see if you were inside of a container... or maybe I'm thinking of one of the ways to implement one...

ArangoGutierrez:
PING -> @GodloveD

GodloveD:
Well, I think it is conceivable that either SINGULARITY_CONTAINER or SINGULARITY_NAME may be expected by the user to determine whether or not they are in a container. 

I def think we need to preserve at least one of them and maybe both.

src/pkg/util/env/clean.go
```diff
@@ -0,0 +1,41 @@
...
+			e[0] = strings.TrimPrefix(e[0], "SINGULARITYENV_")
+		} else if IsCleanEnv && (e[0] != "HOME" &&
+			e[0] != "TERM" &&
+			e[0] != "LANG" &&
+			e[0] != "http_proxy" &&
+			e[0] != "https_proxy" &&
+			e[0] != "no_proxy" &&
+			e[0] != "all_proxy" &&
+			e[0] != "ftp_proxy") {
+			continue
```
GodloveD: I would include the SINGULARITY_NAME and SINGULARITY_CONTAINER variables here.  And then we must set them somewhere before this code executes.

GodloveD (changes requested):

src/cmd/singularity/cli/actions.go
```diff
@@ -269,30 +270,7 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
...
-		}
-
-		// Transpose environment
-		if strings.HasPrefix(e[0], "SINGULARITYENV_") {
-			e[0] = strings.TrimPrefix(e[0], "SINGULARITYENV_")
-		} else if IsCleanEnv {
-			continue
-		}
-
-		if e[0] == "HOME" {
```
bauerm97, line 292: How does the home env variable get set if you're removing the handling from here?

src/pkg/util/env/clean.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"strings"
+
+	"github.com/opencontainers/runtime-tools/generate"
+	"github.com/singularityware/singularity/src/pkg/sylog"
+)
+
+// CleanEnv cleans environment variables before running the container
+func CleanEnv(g *generate.Generator, IsCleanEnv bool, environment []string) {
+
```
bauerm97: I would prefer if we don't leave lines of space here

src/pkg/util/env/clean.go
```diff
@@ -0,0 +1,41 @@
...
+		e := strings.SplitN(env, "=", 2)
+		if len(e) != 2 {
+			sylog.Verbosef("Can't process environment variable %s", env)
+			continue
+		}
+
+		// Transpose environment
+		if strings.HasPrefix(e[0], "SINGULARITYENV_") {
+			e[0] = strings.TrimPrefix(e[0], "SINGULARITYENV_")
+		} else if IsCleanEnv && (e[0] != "HOME" &&
```
bauerm97: This seems like we're doing things in a very naive way here. First of all, we're not guaranteed that these variables are set yet. For instance in 2.x, we always force the env var `LANG=C` to be set (however, this was to fix some bugs and I *don't* think that needs to happen in 3.0)

src/pkg/util/env/clean.go
```diff
@@ -0,0 +1,41 @@
...
+
+	for _, env := range environment {
+		e := strings.SplitN(env, "=", 2)
+		if len(e) != 2 {
+			sylog.Verbosef("Can't process environment variable %s", env)
+			continue
+		}
+
+		// Transpose environment
+		if strings.HasPrefix(e[0], "SINGULARITYENV_") {
```
bauerm97: Also, since this is in an if...else if block, this implies that `SINGULARITYENV_` variables are *always* transformed even when `--cleanenv` is set, is that what we want?

bauerm97 (changes requested):

src/pkg/util/env/clean.go
GodloveD: I spoke with Greg about the `LANG=C` thing and I think he just set it b/c it is critical for many apps (like Perl) and he was tired of getting issues raised about things being broken because of it.  IMHO we should leave it because we have already gone back and forth about it once.  It's a judgement call, a judgement has been made, and now it's important to just be consistent.

src/pkg/util/env/clean.go
GodloveD: Yes that is the behavior in 2.6.

src/pkg/util/env/clean.go
ArangoGutierrez: umm will check my VScode config on this one

src/pkg/util/env/clean.go
ArangoGutierrez: so... final call : 
 - force `LANG=C`?
 - or just pass the `LANG` env value from the host?

ArangoGutierrez:
@GodloveD , @bauerm97 RFR

src/pkg/util/env/clean.go
```diff
@@ -0,0 +1,50 @@
...
+			e[0] != "https_proxy" &&
+			e[0] != "no_proxy" &&
+			e[0] != "all_proxy" &&
+			e[0] != "ftp_proxy" &&
+			e[0] != "SINGULARITY_CONTAINER" &&
+			e[0] != "SINGULARITY_NAME") {
+			continue
+		}
+
+		if e[0] == "HOME" {
```
bauerm97: This conditional doesn't make a lot of sense to me. I think we should always set `HOME=HomeDest` inside the container, whether or not the `HOME` variable is already set on this host. This conditional implies that we're only setting `HOME` in the container if `HOME` is *already* set on the host.

In my opinion, it makes much more sense to set the unconditional env variables outside of the loop where we iterate over the environment:

```
for ... {
    // iterating over env
}

if !NoHome {
    g.AddProcessEnv("HOME", HomeDest)
} else {
    g.AddProcessEnv("HOME", "/")
}

g.AddProcessEnv("LANG", "C")
```

src/pkg/util/env/clean.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"strings"
+
+	"github.com/opencontainers/runtime-tools/generate"
+	"github.com/singularityware/singularity/src/pkg/sylog"
+)
+
+// CleanEnv cleans environment variables before running the container
+func CleanEnv(g *generate.Generator, NoHome bool, IsCleanEnv bool, HomeDest string, environment []string) {
```
bauerm97: I think this func should be called something like `SetContainerEnv()` instead of `CleanEnv()` to avoid confusion with the `--cleanenv` flag

src/pkg/util/env/clean.go
```diff
@@ -0,0 +1,50 @@
...
+			e[0] = strings.TrimPrefix(e[0], "SINGULARITYENV_")
+		} else if IsCleanEnv && (e[0] != "HOME" &&
+			e[0] != "TERM" &&
+			e[0] != "LANG" &&
+			e[0] != "http_proxy" &&
+			e[0] != "https_proxy" &&
+			e[0] != "no_proxy" &&
+			e[0] != "all_proxy" &&
+			e[0] != "ftp_proxy" &&
+			e[0] != "SINGULARITY_CONTAINER" &&
```
bauerm97: These two variables still aren't ever being set (`SINGULARITY_CONTAINER` && `SINGULARITY_NAME`). From what I remember, `SINGULARITY_CONTAINER` is image argument passed to the CLI and `SINGULARITY_NAME` is the name of the file. For 
example:

```
singularity run /my/test/image.sif

SINGULARITY_CONTAINER=/my/test/image.sif
SINGULARITY_NAME=image.sif
```

src/pkg/util/env/clean.go
bauerm97: Always set `LANG=C` I guess

bauerm97 (changes requested):

src/pkg/util/env/clean.go
ArangoGutierrez: and when the image comes from a URI?
e.g
```
singularity run library://alpine:latest
```

ArangoGutierrez: should we leave `SINGULARITY_NAME` when actually having a name, a.k.a when using instances?

jmstover: @bauerm97 ... look at `/.singularity.d/actions/shell` .... `$SINGULARITY_CONTAINER` is being used in that to set PS1 when bash exists as a shell in the container.

Here's a comparison of 2.6.0 and 3.0.0 on the same image:
```
$ /usr/local/singularity/2.6.0/bin/singularity shell ~/tmp/sing/centos7.img 
Singularity centos7.img:~/development/go/src/github.com/singularityware/singularity

$ /usr/local/singularity/3.0.0-dev/bin/singularity shell ~/tmp/sing/centos7.img
Singularity :~/development/go/src/github.com/singularityware/singularity>
```

jmstover: @ArangoGutierrez 

```
$ /usr/local/singularity/2.6.0/bin/singularity shell docker://centos:7
  [ ... ]
SINGULARITY_NAME=centos:7
SINGULARITY_CONTAINER=centos:7
```

ArangoGutierrez: they *Should* be different , at least `CONTAINER` should point the path/source, i kind of agree with the name. I would use `NAME` when running instances. but is just my opinion

ArangoGutierrez:
```
[eduardo@rhel7 singularity]$ singularity exec /tmp/alpine.sif env | grep SINGULARITY
SINGULARITY_CONTAINER=/tmp/alpine.sif
SINGULARITY_NAME=alpine.sif
```
and with instances
```
[eduardo@rhel7 singularity]$ singularity instance start image.sif namy
[eduardo@rhel7 singularity]$ singularity exec instance://namy env | grep SINGULARITY
SINGULARITY_CONTAINER=instance://namy
SINGULARITY_NAME=namy
```

@GodloveD , @bauerm97  LMK if this is the right behavior so I can go ahead and push the changes

src/pkg/util/env/clean.go
jmstover: That's opposite current behavior:
```
$ /usr/local/singularity/2.6.0/bin/singularity shell -c --no-home ~/tmp/sing/centos7.img 
WARNING: Could not chdir to home: /home/jason
Singularity centos7.img:/> pwd
/
Singularity centos7.img:/> echo $HOME
/home/jason
```

The above is the same behavior as if you SSH into a node, but your `$HOME` isn't availible. You get sent to `/` as your CWD, but `$HOME` is still set to what your HOME actually is.

What you are saying is when `--no-home` is specified, then `HOME=/` and that's just wrong from a system standpoint.

ArangoGutierrez:
```
singularity exec --help 
[...] 
      --no-home            Do NOT mount users home directory if home is
                           not the current working directory.
```
so @jmstover is right, `--no-home` should just NOT mount HOME

src/pkg/util/env/clean.go
ArangoGutierrez: From 2.6 
```
$ singularity exec ~/tmp/sing/centos7.img env | grep ^LANG
LANG=en_US
$ singularity exec --cleanenv ~/tmp/sing/centos7.img env | grep ^LANG 
LANG=C
```
so force `LANG=C` is just when using `--cleanenv` flag

ArangoGutierrez:
@bauerm97 , @GodloveD , @jmstover  RFR 
```
[eduardo@rhel7 tmp]$ singularity exec centos.sif env | grep LANG
LANG=en_US.UTF-8
GDM_LANG=en_US.UTF-8
[eduardo@rhel7 tmp]$ singularity exec --cleanenv centos.sif env | grep LANG
LANG=C
```
-/- 
```
[eduardo@rhel7 tmp]$ singularity shell --cleanenv centos.sif 
Singularity centos.sif:/tmp> env
TERM=xterm-256color
SHELL=/bin/bash
SINGULARITY_NAME=centos.sif
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
PWD=/tmp
LANG=C
PS1=Singularity centos.sif:\w> 
HOME=/home/eduardo
SHLVL=1
SINGULARITY_CONTAINER=/tmp/centos.sif
_=/usr/bin/env
```

Hope now `--cleanenv` is not TOO clean :P

ArangoGutierrez:
Sorry for the close==> open, hit double enter after the last comment and puff closed :man_shrugging:

ArangoGutierrez:
@jmstover RFR

GodloveD (approved):
lgtm

ArangoGutierrez:
@bauerm97 , @jmstover <- PING RFR 
Dave Cowlove approved :+1:

jmstover (approved):

bauerm97 (approved):

Issue #1914, opened 2018-08-31
Fatal messages should be warnings


jmstover:
Singularity 3.0 exits with a `FATAL` message when some files don't exist:

```
$ singularity shell lolcow.sif
FATAL:    container creation failed: hook function for tag layer returns error: stat failed for /etc/localtime: no such file or directory
```

These should be WARNINGS as in 2.x

```
WARNING: Non existent 'bind path' source '/etc/localtime'
```

All the files referenced are: `/etc/localtime`, `/etc/hosts`, `/etc/resolv.conf`

cclerget:
Fixed by #1922

PR #1922, opened 2018-09-05
Developement 3.0: Display warning rather than returning an error for non existent sources

cclerget:

This PR fix issue #1914


GodloveD:
This seems like it is a fix for #1914.  Not 1904.  @cclerget can you confirm?

cclerget:
@GodloveD Oops yes :). Fix that. Thanks Dave !

jmstover:
A non-existent `/etc/resolv.conf` causes this to still error with a FATAL error

src/pkg/util/fs/layout/layer/overlay/overlay.go
```diff
@@ -127,7 +128,8 @@ func (o *Overlay) createLayer(rootFsPath string, system *mount.System) error {
 			}
 			if point.Type == "" {
 				if err := syscall.Stat(point.Source, st); os.IsNotExist(err) {
-					return fmt.Errorf("stat failed for %s: %s", point.Source, err)
+					sylog.Warningf("skipping mount of %s: %s", point.Source, err)
+					continue
```
bauerm97: Should we still check `} else if err != nil {` to error out if there's some other form of error that's not `os.IsNotExist()`?

GodloveD (dismissed):
After @bauerm97 's suggested change I think this is good to go.

cclerget:
@GodloveD @bauerm97 Finally changes to catch all errors like it's done with underlay

GodloveD (approved):
lgtm

PR #1927, opened 2018-09-06
Add special handling for /dev in userbinds

### LINKED ISSUE #1907: Option to bind full /dev with contain
With 3.0 have an option to skip the minimal dev if `--contain`/`--containall` is used.

For example:

  `singularity exec --bind-dev -c [...]` 

Will do the normal `/dev` bind mounting instead of the minimal `/dev`. This _should not_ override the configuration setting of:  `mount dev = minimal`

cclerget:

Fixes #1907


jmstover:
@cclerget 
Per @gmkurtzer , this needs to do this:  If the `singularity.conf` has `mount dev = minimal` then a user _must not_ be able to do `-B /dev`

src/runtime/engines/singularity/container.go
```diff
@@ -989,6 +989,21 @@ func (c *container) addUserbindsMount(system *mount.System) error {
...
+			// by example with --contain option
+			if c.engine.EngineConfig.File.MountDev == "yes" {
+				system.Points.RemoveByTag(mount.DevTag)
+
+				err := system.Points.AddBind(mount.DevTag, "/dev", dst, syscall.MS_BIND|syscall.MS_REC)
+				if err != nil {
+					return fmt.Errorf("unable to add dev to mount list: %s", err)
+				}
+			} else {
+				sylog.Verbosef("Skipping /dev bind mount, disallowed by configuration")
```
GodloveD: I think this should be a `sylog.Warningf` not just a verbose message.

GodloveD (changes requested):
I think this looks great with the one small change I've suggested.

cclerget:
@jmstover Yes this line https://github.com/singularityware/singularity/blob/f604c3055e472ca2eb40c46da2aa552ff3cb4add/src/runtime/engines/singularity/container.go#L995 check that and display warning message as suggested by @GodloveD if `mount dev` is not equal to `yes`

GodloveD (dismissed):
lgtm

src/runtime/engines/singularity/container.go
```diff
@@ -1001,27 +1032,42 @@ func (c *container) addUserbindsMount(system *mount.System) error {
...
-				err := system.Points.AddBind(mount.DevTag, "/dev", dst, syscall.MS_BIND|syscall.MS_REC)
-				if err != nil {
-					return fmt.Errorf("unable to add dev to mount list: %s", err)
+		// special case for /dev mount to override default mount behaviour
+		// with --contain option or 'mount dev = minimal'
+		if strings.HasPrefix(src, "/dev") {
+			if c.engine.EngineConfig.File.MountDev == "minimal" || c.engine.EngineConfig.GetContain() {
+				if strings.HasPrefix(src, "/dev/shm") || strings.HasPrefix(src, "/dev/mqueue") {
+					sylog.Warningf("Skipping %s bind mount: not authorized", src)
+				}
```
DrDaveD: A fresh /dev/shm is already mounted so a request to bind that should get rejected as a duplicate bind, right?

As for /dev/mqueue, I don't see why someone would want to do that if they want to restrict the container, but if they ask for it, why should singularity prevent it?

cclerget: The main reason is that these mount points are world writable and don't want to deal with nasty symlinks for security reasons, so Singularity just prevent bind and display a `not authorized` message

DrDaveD: Oh I see, you're checking the prefix of the path, not only an exact match, I hadn't noticed.  Maybe those paths should end in slash though, so they don't prevent mounting the device directories themselves.

cclerget:
I added missing mount of /dev/mqueue when IPC namespace and `--contain` are specified. This should be ready for review @GodloveD

jmstover (approved):

GodloveD (approved):
This lgtm as soon as conflicts are resolved.

PR #1928, opened 2018-09-06
Implement an ECL mechanism

yhcote:
This PR implements a restriction list permitting to manage which containers may run on a host. A configuration file is used to setup rules about which containers to block on allow based on their location and signing entities.


src/pkg/signing/signing.go
```diff
@@ -266,3 +271,29 @@ func Verify(cpath, url, authToken string) error {
 
 	return nil
 }
+
+// GetSignEntities returns all signing entities for an ID/Groupid
+func GetSignEntities(cpath string) ([]string, error) {
```
cclerget, line 276: To keep trace: it's safest to use `sif.LoadContainerFp` with a `GetSignEntitiesFromFp`

cclerget (approved):
Looks good !

GodloveD:
I know this is already merged, but adding it to the proper releases, milestones, etc and putting story points on it will give us more data for estimating release date afaik.

PR #1949, opened 2018-09-10
Development 3.0: Honor --allow-setuid flag

cclerget:

This PR allows use of `--allow-setuid` command line option for root user in order to use setuid binaries inside container.

**Checkoff for all PRs:**


src/runtime/engines/singularity/container.go
```diff
@@ -77,6 +75,12 @@ func create(engine *EngineOperations, rpcOps *client.RPC, pid int) error {
 		}
 	}
 
+	if os.Geteuid() != 0 {
```
bauerm97, line 78: Should maybe be checking real uid instead of euid?

bauerm97 (changes requested):

bauerm97 (approved):
As discussed with @cclerget , prior to this we enforce `setresuid(uid,uid,uid)` of the user who called Singularity, so this thread is actually running with no privileges (effective OR real), thus this call is safe.

PR #1965, opened 2018-09-11
Fix unprivileged sandbox mount by removing unnecessary remount

DrDaveD:

This PR removes an unnecessary remount of the base image when it is a sandbox, which fails when running with -u with operation not permitted.   The remount did nothing useful anyway, it did not change any flags.

- None

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/src/runtime/engines/singularity/container.go
+++ b/src/runtime/engines/singularity/container.go
@@ -499,7 +499,6 @@ func (c *container) addRootfsMount(system *mount.System) error {
 		if err := system.Points.AddBind(mount.RootfsTag, rootfs, c.session.RootFsPath(), flags); err != nil {
 			return err
 		}
-		system.Points.AddRemount(mount.RootfsTag, c.session.RootFsPath(), flags)
 		return nil
 	}
 	flags |= syscall.MS_RDONLY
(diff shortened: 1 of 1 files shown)


DrDaveD:
In fact, I see now that container.go has lots of places where there's an AddRemount that does nothing.  @cclerget what is the reason for those?  Only a couple of them are surround by `if !c.userNS`.

src/runtime/engines/singularity/container.go
```diff
@@ -499,7 +499,6 @@ func (c *container) addRootfsMount(system *mount.System) error {
 		if err := system.Points.AddBind(mount.RootfsTag, rootfs, c.session.RootFsPath(), flags); err != nil {
 			return err
 		}
-		system.Points.AddRemount(mount.RootfsTag, c.session.RootFsPath(), flags)
```
cclerget, line 502: Instead of removing `AddRemount` it's better to add a check `if !c.userNS { AddRemount... }` to force remount with setuid workflow and correctly apply `MS_NOSUID | MS_NODEV ` option

cclerget (dismissed):

src/runtime/engines/singularity/container.go
DrDaveD: Oh, so that's what the remount is for.  Ok, I will update for that change.

bauerm97 (approved):
lgtm

PR #1975, opened 2018-09-13
Def parser fix for definitions with no header

ikaneshiro:
Def parser scanner returns the first section it runs into if there isn't a header, which was causing an index out of range panic. This detects if the first returned token is a header or not and parses it accordingly.

### DIFF
--- a/src/pkg/build/build.go
+++ b/src/pkg/build/build.go
@@ -320,11 +320,8 @@ func makeDef(spec string) (types.Definition, error) {
 		if err != nil {
 			return def, fmt.Errorf("unable to parse URI %s: %v", spec, err)
 		}
-	}
-
-	// Non-URI passed as spec
-	if _, err := os.Stat(spec); err == nil {
-
+	} else if _, err := os.Stat(spec); err == nil {
+		// Non-URI passed as spec
 		defFile, err := os.Open(spec)
 		if err != nil {
 			return def, fmt.Errorf("unable to open file %s: %v", spec, err)
--- a/src/pkg/build/build.go
+++ b/src/pkg/build/build.go
@@ -338,10 +338,8 @@ func makeDef(spec string) (types.Definition, error) {
 		if err != nil {
 			return def, fmt.Errorf("unable to parse URI %s: %v", spec, err)
 		}
-	}
-	// Non-URI passed as spec
-	if _, err := os.Stat(spec); err == nil {
-
+	} else if _, err := os.Stat(spec); err == nil {
+		// Non-URI passed as spec
 		defFile, err := os.Open(spec)
 		if err != nil {
 			return def, fmt.Errorf("unable to open file %s: %v", spec, err)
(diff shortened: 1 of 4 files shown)


src/pkg/build/build.go
```diff
@@ -243,23 +291,23 @@ func (b *Build) runBuildEngine() error {
 
 // Bundle creates the bundle using the ConveyorPacker and returns it. If this
 // function is called multiple times it will return the already created Bundle
-func (b *Build) Bundle() (*types.Bundle, error) {
-	if b.b != nil {
-		return b.b, nil
-	}
+// func (b *Build) Bundle() (*types.Bundle, error) {
```
jchavezlavalle: There's commented code in here @ikaneshiro

src/pkg/build/build.go
```diff
@@ -243,23 +291,23 @@ func (b *Build) runBuildEngine() error {
...
 
-	if err := b.c.Get(b.d); err != nil {
-		return nil, fmt.Errorf("conveyor failed to get: %v", err)
-	}
+// 	if err := b.c.Get(b.b); err != nil {
+// 		return nil, fmt.Errorf("conveyor failed to get: %v", err)
+// 	}
```
jchavezlavalle: and here also more commented code

src/pkg/build/build.go
```diff
@@ -243,23 +291,23 @@ func (b *Build) runBuildEngine() error {
...
 
-	bundle, err := b.c.Pack()
-	if err != nil {
-		return nil, fmt.Errorf("packer failed to pack: %v", err)
-	}
+// 	bundle, err := b.c.Pack()
+// 	if err != nil {
+// 		return nil, fmt.Errorf("packer failed to pack: %v", err)
+// 	}
```
jchavezlavalle: also here some commented code

src/pkg/build/build.go
```diff
@@ -243,23 +291,23 @@ func (b *Build) runBuildEngine() error {
...
 
-	b.b = bundle
-	return b.b, nil
-}
+// 	b.b = bundle
+
+// 	b.addOptions()
+
+// 	return b.b, nil
+// }
```
jchavezlavalle: as well as this part w/ some commented code

jchavezlavalle (commented):
just added some comments, there's some commented code, guess it should be removed

src/pkg/build/types/definition_parser_deffile_test.go
```diff
@@ -135,7 +136,8 @@ func TestParseDefinitionFileFailure(t *testing.T) {
 			}
 			defer defFile.Close()
 
-			if _, err = ParseDefinitionFile(defFile); err == nil {
+			if d, err := ParseDefinitionFile(defFile); err == nil {
+				fmt.Println(d)
```
bauerm97: Purpose of `fmt.Println()` here? Forgot to remove a debug statement? We should rather be using a `t.Log` statements to log here

bauerm97 (changes requested):

src/pkg/build/types/definition_parser_deffile_test.go
ikaneshiro: missed debug statement

bauerm97 (approved):

PR #1981, opened 2018-09-13
help cmd | man <image>

ArangoGutierrez:

Add support for help cmd
```bash
$ cat my_container.def
Bootstrap: docker
From: busybox

%help
    Some help for this container

$ sudo singularity build my_container.sif my_container.def
Using container recipe deffile: my_container.def
[...snip...]
Cleaning up...

$ singularity help my_container.sif


    Some help for this container
```

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/src/cmd/singularity/cli/actions.go
+++ b/src/cmd/singularity/cli/actions.go
@@ -31,7 +31,6 @@ func init() {
 		ExecCmd,
 		ShellCmd,
 		RunCmd,
-		HelpCmd,
 	}
 
 	// TODO : the next n lines of code are repeating too much but I don't
--- a/src/cmd/singularity/cli/actions.go
+++ b/src/cmd/singularity/cli/actions.go
@@ -31,6 +31,7 @@ func init() {
 		ExecCmd,
 		ShellCmd,
 		RunCmd,
+		HelpCmd,
 	}
 
 	// TODO : the next n lines of code are repeating too much but I don't
--- a/src/cmd/singularity/cli/help.go
+++ b/src/cmd/singularity/cli/help.go
@@ -22,9 +22,13 @@ func init() {
 // HelpCmd singularity help
 var HelpCmd = &cobra.Command{
 	DisableFlagsInUseLine: true,
-	Args:   cobra.ExactArgs(1),
-	PreRun: sylabsToken,
+	PreRun:                sylabsToken,
 	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) == 0 {
+			cmd.Root().Help()
+			return
+		}
+
 		c, _, e := cmd.Root().Find(args)
 		if _, err := os.Stat(args[0]); err == nil {
 			// Help prints (if set) the sourced %help section on the definition file
--- a/src/cmd/singularity/cli/help.go
+++ b/src/cmd/singularity/cli/help.go
@@ -1,0 +1,46 @@
+// Copyright (c) 2018, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package cli
+
+import (
+	"os"
+
+	"github.com/singularityware/singularity/src/docs"
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	HelpCmd.Flags().SetInterspersed(false)
+	SingularityCmd.SetHelpCommand(HelpCmd)
+
+	SingularityCmd.AddCommand(HelpCmd)
+}
+
+// HelpCmd singularity help
+var HelpCmd = &cobra.Command{
+	DisableFlagsInUseLine: true,
+	Args:   cobra.ExactArgs(1),
+	PreRun: sylabsToken,
+	Run: func(cmd *cobra.Command, args []string) {
+		c, _, e := cmd.Root().Find(args)
+		if _, err := os.Stat(args[0]); err == nil {
+			// Help prints (if set) the sourced %help section on the definition file
+			a := []string{"/bin/cat", "/.singularity.d/runscript.help"}
+			execStarter(cmd, args[0], a, "")
+		} else if c == nil || e != nil {
+			c.Printf("Unknown help topic %#q\n", args)
+			c.Root().Usage()
+		} else {
+			c.InitDefaultHelpFlag() // make possible 'help' flag to be shown
+			c.Help()
+		}
+	},
+
+	Use:     docs.HelpUse,
+	Short:   docs.HelpShort,
+	Long:    docs.HelpLong,
+	Example: docs.HelpExample,
+}
--- a/src/cmd/singularity/cli/help.go
+++ b/src/cmd/singularity/cli/help.go
@@ -6,9 +6,18 @@
 package cli
 
 import (
+	"encoding/json"
 	"os"
+	"path/filepath"
 
+	"github.com/opencontainers/runtime-tools/generate"
 	"github.com/singularityware/singularity/src/docs"
+	"github.com/singularityware/singularity/src/pkg/buildcfg"
+	"github.com/singularityware/singularity/src/pkg/sylog"
+	"github.com/singularityware/singularity/src/pkg/util/exec"
+	"github.com/singularityware/singularity/src/runtime/engines/config"
+	"github.com/singularityware/singularity/src/runtime/engines/config/oci"
+	"github.com/singularityware/singularity/src/runtime/engines/singularity"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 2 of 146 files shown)


src/docs/content.go
```diff
@@ -616,4 +616,29 @@ Enterprise Performance Computing (EPC)`
 	for all those blocks.`
 	VerifyExample string = `
   $ singularity verify container.sif`
+	// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+	// help
+	// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+	HelpUse   string = `help <image path>`
```
jmstover: This should probably be:
```
HelpUse   string = `help <command|image path>`
```
Which would also change the other text so it isn't so focused on the help of an *image* but help in general

jmstover (changes requested):

src/cmd/singularity/cli/help.go
```diff
@@ -0,0 +1,46 @@
...
+	HelpCmd.Flags().SetInterspersed(false)
+	SingularityCmd.SetHelpCommand(HelpCmd)
+
+	SingularityCmd.AddCommand(HelpCmd)
+}
+
+// HelpCmd singularity help
+var HelpCmd = &cobra.Command{
+	DisableFlagsInUseLine: true,
+	Args:   cobra.ExactArgs(1),
```
jmstover: This is wrong isn't is? I mean... what if you do:
```
singularity help instance start
```
It'd only look at the help for `instance` and not `instance start`

ArangoGutierrez: ummm I don't know how cobra can handle that ....

ArangoGutierrez: ha!  found!

src/docs/content.go
ArangoGutierrez: good catch, thanks!

src/docs/content.go
```diff
@@ -616,4 +613,29 @@ Enterprise Performance Computing (EPC)`
 	for all those blocks.`
 	VerifyExample string = `
   $ singularity verify container.sif`
+	// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+	// help
+	// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+	HelpUse   string = `help <command|image path>`
+	HelpShort string = `Display help for command||container if available`
```
jmstover: `command||container`
Change to either `command|container` _or_ `command or container image`

ArangoGutierrez:
@jmstover <- d9a1d2d

jmstover (approved):

src/cmd/singularity/cli/actions.go
```diff
@@ -31,6 +31,7 @@ func init() {
 		ExecCmd,
 		ShellCmd,
 		RunCmd,
+		HelpCmd,
```
bauerm97: Not sure I understand why we're adding a bunch of flags to the `HelpCmd` here. We don't want to do `singularity help --bind /path container` if we just want the help message in the container.

src/cmd/singularity/cli/help.go
```diff
@@ -0,0 +1,50 @@
...
+		if len(args) == 0 {
+			cmd.Root().Help()
+			return
+		}
+
+		c, _, e := cmd.Root().Find(args)
+		if _, err := os.Stat(args[0]); err == nil {
+			// Help prints (if set) the sourced %help section on the definition file
+			a := []string{"/bin/cat", "/.singularity.d/runscript.help"}
+			execStarter(cmd, args[0], a, "")
```
bauerm97: We should not be using `execStarter` here. We just need to set up a bare `EngineConfig`, put it into `CommonConfig`, and use `exec.Pipe`

bauerm97 (changes requested):

ArangoGutierrez:
Thanks for the heads up @bauerm97 
fix  addressed here 9525d45

ArangoGutierrez:
at request of the public
change `singularity help <image>` to  `singularity man <image>` to avoid ambiguity when seeking for help on a cmd `singularity help [command]` .
^
@bauerm97 , @jmstover , @GodloveD  , @gmkurtzer

src/cmd/singularity/cli/run_help.go
```diff
@@ -0,0 +1,76 @@
...
+	SingularityCmd.AddCommand(RunHelpCmd)
+}
+
+// RunHelpCmd singularity run-help <image>
+var RunHelpCmd = &cobra.Command{
+	DisableFlagsInUseLine: true,
+	PreRun:                sylabsToken,
+	Args:                  cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+
```
bauerm97: Remove blank line

src/docs/content.go
```diff
@@ -38,109 +35,109 @@ Enterprise Performance Computing (EPC)`
 	BuildLong  string = `
 
   IMAGE PATH:
-  
+
```
bauerm97, line 38: Are we sure we're not introducing too much whitespace here in all of these help messages? We want to make sure to be outputting in a very readable form. I'm indifferent on this, but I think @GodloveD should have good insight into if this is too much whitespace

src/cmd/singularity/cli/run_help.go
```diff
@@ -0,0 +1,76 @@
...
+		if _, err := os.Stat(args[0]); err == nil {
```
bauerm97: What happens when `err != nil`? We need to handle `err` after the conditional here. Some kind of error like `container not found` or something

bauerm97 (changes requested):

src/docs/content.go
ArangoGutierrez: I'm not adding a white line here, It seems, that my IDE, just removed a [tab/space] on that line, and replaced it with a clean line.

src/cmd/singularity/cli/run_help.go
ArangoGutierrez: ups, fixing

src/docs/content.go
bauerm97: Ahh okay, so it's not actually a change. No worries then it's already been approved by DG ;) carry on

src/cmd/singularity/cli/run_help.go
ArangoGutierrez: Addressed on f0e95e2

src/docs/content.go
```diff
@@ -616,4 +613,30 @@ Enterprise Performance Computing (EPC)`
 	for all those blocks.`
 	VerifyExample string = `
   $ singularity verify container.sif`
+	// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+	// Run-help
+	// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+	RunHelpUse   string = `run-help <image path>`
+	RunHelpShort string = `Display help for command|container if available`
```
bauerm97: This command does not display help for a command.

src/cmd/singularity/cli/run_help.go
```diff
@@ -0,0 +1,77 @@
...
+	SingularityCmd.AddCommand(RunHelpCmd)
+}
+
+// RunHelpCmd singularity run-help <image>
+var RunHelpCmd = &cobra.Command{
+	DisableFlagsInUseLine: true,
+	PreRun:                sylabsToken,
+	Args:                  cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		if _, err := os.Stat(args[0]); err == nil {
```
GodloveD: Instead of having the entire `Run` section wrapped in this `if` you could do something like this and then just move on with the rest of the code:
```
if _, err := os.Stat(args[0]); err != nil {
        sylog.Fatalf("container not found: %s", err)
}
<rest of code>
```

src/docs/content.go
```diff
@@ -616,4 +613,30 @@ Enterprise Performance Computing (EPC)`
...
+  %help
+      Some help for this container
+
+  $ sudo singularity build my_container.sif my_container.def
+  Using container recipe deffile: my_container.def
+  [...snip...]
+  Cleaning up...
+
+  $ singularity run-help my_container.sif
+
```
GodloveD, line 639: It looks like this implementation doesn't add a bunch of extra whitespace like the 2.x version did.  Should remove extra whitespace from example.

src/cmd/singularity/cli/run_help.go
```diff
@@ -0,0 +1,77 @@
...
+			// Help prints (if set) the sourced %help section on the definition file
+			abspath, err := filepath.Abs(args[0])
+			name := filepath.Base(abspath)
+			a := []string{"/bin/cat", "/.singularity.d/runscript.help"}
```
GodloveD: Is this how the old version of help used to work.  What if there is no `/bin/cat` in the container.  No need to change anything based on this comment.  Just curious.

src/cmd/singularity/cli/run_help.go
```diff
@@ -0,0 +1,77 @@
...
+			engineConfig.SetImage(abspath)
+
+			cfg := &config.Common{
+				EngineName:   singularity.Name,
+				ContainerID:  name,
+				EngineConfig: engineConfig,
+			}
+
+			configData, err := json.Marshal(cfg)
+			if err != nil {
```
GodloveD: For consistency, it'd be good to combine the two lines above like so:
```
if configData, err := json.Marshal(cfg); err != nil {
```

GodloveD (dismissed):
Just a bunch of style nitpicks.  

So I guess this is the basic set of things that must be done to mount a container?!  Seems like a lot!  

¯\_(ツ)_/¯

src/cmd/singularity/cli/run_help.go
bauerm97: That can't be done here. If we do that, then line 64 `configData` will be undefined. Variables which are declared/set in an if statement are scoped:

```
if test := 1; test == 1 {
    fmt.Println(test) // Prints "1"
}

fmt.Println(test) // Compiler error
```

The way Ed did it here is the idiomatic handling

src/cmd/singularity/cli/run_help.go
bauerm97: Agree with this comment

src/cmd/singularity/cli/run_help.go
GodloveD: I stand corrected.  Thanks @bauerm97.

src/cmd/singularity/cli/run_help.go
jmstover: @GodloveD basically yeah... 2.x runs:
`eval_abort cat "${SINGULARITY_MOUNTPOINT}/.singularity.d/runscript.help"`

In `libexec/helpers/help.sh`

ArangoGutierrez: Yeah I double checked with @jmstover this one

ArangoGutierrez:
@GodloveD , @bauerm97 RFR

bauerm97 (approved):

PR #1982, opened 2018-09-13
Remove useragent dependency on buildcfg package

bauerm97:
Users of `useragent` package should now reference `useragent.Value()` after setting value via `useragent.InitValue()`

### DIFF
--- a/src/pkg/util/user-agent/agent.go
+++ b/src/pkg/util/user-agent/agent.go
@@ -9,30 +9,37 @@ import (
 	"fmt"
 	"runtime"
 	"strings"
-
-	"github.com/singularityware/singularity/src/pkg/buildcfg"
 )
 
 // Value contains the Singularity user agent.
 //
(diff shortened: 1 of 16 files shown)


src/pkg/util/user-agent/agent.go
```diff
@@ -9,30 +9,37 @@ import (
...
-	"github.com/singularityware/singularity/src/pkg/buildcfg"
 )
 
 // Value contains the Singularity user agent.
 //
 // For example, "Singularity/3.0.0 (linux amd64) Go/1.10.3".
-var Value string
+func Value() string {
+	if value == "" {
+		panic("useragent.InitValue() must be called before calling useragent.Value()()")
```
tri-adam: Maybe you can fix this `()()` using emacs? :-)

tri-adam (changes requested):

tri-adam (approved):
LGTM

Issue #1990, opened 2018-09-14
Instance Stop stops all instances when supplied no arguments


bauerm97:
## Version of Singularity:

3.0.0-alpha.1-347-g3b2f9ac2

### Expected behavior

```
mibauer@ubuntu:~/testing$ singularity instance stop
Stop a named instance of a given container image

Usage:
  singularity instance stop [stop options...] [instance]

Description:
  The command singularity instance stop allows you to stop and clean up a named,
  running instance of a given container image.

Options:
  -a, --all             Stop all user's instances
  -f, --force           Force kill instance
  -h, --help            help for stop
  -s, --signal string   Signal sent to the instance
  -t, --timeout int     Force kill non stopped instances after X seconds
                        (default 10)
  -u, --user string     If running as root, list instances from "<username>"


Examples:
  $ singularity instance start my-sql.img mysql1
  $ singularity instance start my-sql.img mysql2
  $ singularity instance stop mysql*
  Stopping mysql1 instance of my-sql.img (PID=23845)
  Stopping mysql2 instance of my-sql.img (PID=23858)
  
  $ singularity instance start my-sql.img mysql1
  
  Force instance to shutdown
  $ singularity instance stop -f mysql1 (may corrupt data)
  
  Send SIGTERM to the instance
  $ singularity instance stop -s SIGTERM mysql1
  $ singularity instance stop -s TERM mysql1
  $ singularity instance stop -s 15 mysql1


For additional help or support, please visit https://docs.sylabs.io
```

### Actual behavior

```
mibauer@ubuntu:~/testing$ singularity instance stop
Stopping my_inst2 instance of /home/mibauer/testing/test.sif (PID=9100)
Stopping my_inst instance of /home/mibauer/testing/test.sif (PID=9134)
```

cclerget:
Fixed by #1993

PR #1993, opened 2018-09-14
Development 3.0: Fix instance stop

cclerget:

This PR fixes `instance stop` command from stopping all instances

- Ref: #1990 

**Checkoff for all PRs:**

### DIFF
--- a/src/cmd/singularity/cli/instance_stop.go
+++ b/src/cmd/singularity/cli/instance_stop.go
@@ -7,7 +7,6 @@ package cli
 
 import (
 	"github.com/singularityware/singularity/src/docs"
-	"github.com/singularityware/singularity/src/pkg/sylog"
 	"github.com/spf13/cobra"
 )
 
@@ -41,7 +40,7 @@ var InstanceStopCmd = &cobra.Command{
 		} else if stopAll {
 			stopInstance("*")
 		} else {
-			sylog.Fatalf("no arguments supplied")
+			cmd.Usage()
 		}
 	},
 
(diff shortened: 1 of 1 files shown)


jmstover (approved):

src/cmd/singularity/cli/instance_stop.go
```diff
@@ -37,8 +38,10 @@ var InstanceStopCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		if len(args) > 0 && !stopAll {
```
bauerm97, line 38: Should be using `Command.Args: ExactArgs(1)` here instead of manually checking argument length. https://godoc.org/github.com/spf13/cobra#ExactArgs

bauerm97 (changes requested):

src/cmd/singularity/cli/instance_stop.go
cclerget: The problem with ExactArgs or MinimumNArgs is that flag are not taken into account during processing, so `singularity instance stop -a` always display help with 1 arg missing. Will check if cobra support this without manual check.

bauerm97: Good point, actually think in this case is okay then. Will go ahead and merge :)

bauerm97 (approved):

src/cmd/singularity/cli/instance_stop.go
```diff
@@ -37,8 +38,10 @@ var InstanceStopCmd = &cobra.Command{
...
 			stopInstance(args[0])
-		} else {
+		} else if stopAll {
 			stopInstance("*")
+		} else {
+			sylog.Fatalf("no arguments supplied")
```
bauerm97: Actually, would like to see this:

`cmd.Usage()` instead of `sylog.Fatalf()`

This way we ensure we have consistent output when incorrectly using a command.

bauerm97 (changes requested):

bauerm97 (approved):

PR #1995, opened 2018-09-15
Remove flag "-vendor-only" dep ensure.

yhcote:
This helps with checking that our lock file is in sync with our usage of those vendor packages.


cclerget (approved):
Looks good !

PR #2002, opened 2018-09-17
Add patch to crypto/openpgp to enable private key encryption.

yhcote:
This implements as a patch, the ability to encrypt and serialize private keys.


cclerget (approved):
Looks good !

PR #1971, opened 2018-09-12
Support old images without /.singularity.d/actions/exec

DrDaveD:
**Description of the Pull Request (PR):**

This PR restores the old behavior of 2.x when /.singularity.d/actions/exec does not exist in an old image.  Depends on pr #1970 to work exactly the same so that should be merged first.

**This fixes or addresses the following GitHub issues:**

None

**Checkoff for all PRs:**

- [x] I have read the [Guidelines for Contributing](https://github.com/singularityware/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- [ ] I have added changes to the [CHANGELOG](https://github.com/singularityware/singularity/blob/development/CHANGELOG.md) and and documentation updates to the [singularityware](https://www.github.com/singularityware/singularityware.github.io) documentation base.
- [ ] I have tested this PR locally with a `make test`
- [x] This PR is against the project's master branch
- [x] I have added myself as a contributor to the [contributors's file](https://github.com/singularityware/singularity/blob/master/CONTRIBUTORS.md)
- [x] This PR is ready for review and/or merge

Attn: @singularityware-admin

PR #2007, opened 2018-09-17
3.0: fix go fmt check

DrDaveD:

This PR makes the go fmt check do what it is supposed to do: flag if there's a problem with the the go formatting.  Also prints the file names that have problems.

The current code had three bugs:
1. it used "$(...)" to attempt to run go fmt, which was interpreted by make and turned into an empty string.  I could have fixed with "$$(...)" but instead used backquotes like was used a few lines later in the Makefile.
2. It used "go fmt" which runs "gofmt -l -w".  That would have worked, but it would have corrected problems as a side effect, which would be very confusing.  I changed it to "gofmt -l".
3. It used "./..." to indicate everything recursively under the current directory, where it should have been just "."

- Mentioned in #1971

**Checkoff for all PRs:**

Attn: @singularityware-admin

### DIFF
--- a/mlocal/frags/Makefile.stub
+++ b/mlocal/frags/Makefile.stub
@@ -194,7 +194,7 @@ collect:
 check:
 	@echo " CHECK go fmt"
 	$(V)export CGO_CPPFLAGS="$(cgo_CPPFLAGS)" CGO_LDFLAGS="$(cgo_LDFLAGS)" && \
-		cd $(SOURCEDIR) && test -z $(go fmt ./...)
+		cd $(SOURCEDIR) && test -z `gofmt -l .|tee /dev/stderr`
 	@echo "       PASS"
 	@echo " CHECK go vet"
 	$(V)export CGO_CPPFLAGS="$(cgo_CPPFLAGS)" CGO_LDFLAGS="$(cgo_LDFLAGS)" && \
--- a/mlocal/frags/Makefile.stub
+++ b/mlocal/frags/Makefile.stub
@@ -194,7 +194,7 @@ collect:
 check:
 	@echo " CHECK go fmt"
 	$(V)export CGO_CPPFLAGS="$(cgo_CPPFLAGS)" CGO_LDFLAGS="$(cgo_LDFLAGS)" && \
-		cd $(SOURCEDIR) && test -z `gofmt -l src|tee /dev/stderr`
+		cd $(SOURCEDIR) && test -z "`gofmt -l src|tee /dev/stderr`"
 	@echo "       PASS"
 	@echo " CHECK go vet"
 	$(V)export CGO_CPPFLAGS="$(cgo_CPPFLAGS)" CGO_LDFLAGS="$(cgo_LDFLAGS)" && \
--- a/mlocal/frags/Makefile.stub
+++ b/mlocal/frags/Makefile.stub
@@ -196,7 +196,7 @@ check:
 	$(V)export CGO_CPPFLAGS="$(cgo_CPPFLAGS)" CGO_LDFLAGS="$(cgo_LDFLAGS)" && \
 		cd $(SOURCEDIR) && \
 		( test -z "`go fmt -n ./... | sed 's/ -w / /' | sh | tee /dev/stderr`" || \
-		   echo "Use 'go fmt' to update formatting"; false )
+		   ( echo "Use 'go fmt' to update formatting"; false ) )
 	@echo "       PASS"
 	@echo " CHECK go vet"
 	$(V)export CGO_CPPFLAGS="$(cgo_CPPFLAGS)" CGO_LDFLAGS="$(cgo_LDFLAGS)" && \
--- a/mlocal/frags/Makefile.stub
+++ b/mlocal/frags/Makefile.stub
@@ -196,7 +196,9 @@ check:
 	$(V)export CGO_CPPFLAGS="$(cgo_CPPFLAGS)" CGO_LDFLAGS="$(cgo_LDFLAGS)" && \
 		cd $(SOURCEDIR) && \
 		( test -z "`go fmt -n ./... | sed 's/ -w / /' | sh | tee /dev/stderr`" || \
-		   ( echo "Use 'go fmt' to update formatting"; false ) )
+		   ( echo "The above files have formatting errors."; \
+		     echo "Use 'go fmt github.com/singularityware/singularity/...' to correct the errors."; \
+		     false ) )
 	@echo "       PASS"
 	@echo " CHECK go vet"
 	$(V)export CGO_CPPFLAGS="$(cgo_CPPFLAGS)" CGO_LDFLAGS="$(cgo_LDFLAGS)" && \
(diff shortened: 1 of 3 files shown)


DrDaveD:
And I found a 4th problem: it was also covering the vendor directory.  I changed it to only look in "src".

DrDaveD:
And there was a fifth problem: the output of gofmt needed to be double-quoted.

DrDaveD:
The 3rd & 4th problems were only because I switched to gofmt from go fmt, which I still think is important because I don't think that "make check" should have any side effects.  Switching to look in just "src" left out .go files in etc and test, however, so I have updated it to look in all non-vendor subdirectories.

DrDaveD:
Added message to tell newbies what to do if gofmt fails.

DrDaveD:
Meanwhile somebody has updated src/cmd/singularity/cli/pull.go, src/cmd/singularity/cli/search.go, and src/pkg/syecl/syecl.go which fail the gofmt check in go version 1.10.4.  Worse, the changes to pull.go and search.go were in fact put in because of go version 1.11's gofmt.  The format complained about in syecl.go is not objected to in 1.11's gofmt.  I think the only reasonable thing to do is to switch travis to 1.11 first.

DrDaveD:
#2024 was merged. Now this is switched to using `go fmt -n ./...|sed 's/ -w / /'|sh` as suggested by @bauerm97 instead of `gofmt -l`.  Expect travis to fail this time because there are some formatting problems not yet fixed from #2027.

DrDaveD:
Now rebased on current master after #2027 merged, travis should pass.

mlocal/frags/Makefile.stub
```diff
@@ -194,7 +194,9 @@ collect:
 check:
 	@echo " CHECK go fmt"
 	$(V)export CGO_CPPFLAGS="$(cgo_CPPFLAGS)" CGO_LDFLAGS="$(cgo_LDFLAGS)" && \
-		cd $(SOURCEDIR) && test -z $(go fmt ./...)
+		cd $(SOURCEDIR) && \
+		( test -z "`go fmt -n ./... | sed 's/ -w / /' | sh | tee /dev/stderr`" || \
```
bauerm97, line 198:
```
test -z "`gofmt -l \`go list -f "{{.Dir}}" ./...\``"
```

mlocal/frags/Makefile.stub
```diff
@@ -194,7 +194,9 @@ collect:
...
+		   ( echo "Use 'go fmt' to update formatting"; false ) )
```
bauerm97:
```
( echo "The following files have formatting errors:" ) && \
( echo gofmt -l `go list -f "{{.Dir}}" ./...` ) && \
( echo "Use 'go fmt github.com/singularityware/singularity/...' to correct all formatting errors"; false ) )
```

bauerm97 (dismissed):
I think a few tweaks here would be good, just a slightly easier way to use `gofmt` tool with proper list (sorry if you had previously done it this way :P ).

Also we should make sure the statement at the end says that there are formatting errors, which files are problematic, and a command that works from anywhere to fix them

mlocal/frags/Makefile.stub
DrDaveD: I had tried `go list -f {{.Dir}}` but the problem is that it lists subdirectories of other directories so some files get checked twice.  I posted an alternative in a [comment on #2011](https://github.com/sylabs/singularity/pull/2011#issuecomment-422538900) that filters the output with awk but it isn't pretty.  I think the form I've got now is simpler.

mlocal/frags/Makefile.stub
DrDaveD: I updated the message to something similar.

mlocal/frags/Makefile.stub
bauerm97: I was able to reproduce this behavior, but I would really prefer to not use linux text processing to handle this so we can minimize the chance we make human error. I've come up with a workaround which sticks within the `go` to minimize error:

```
mibauer@ubuntu:~/go/src/github.com/sylabs/singularity$ gofmt -l `go list -f '{{$d := .Dir}}{{ range .GoFiles }}{{$d}}/{{.}} {{end}}' ./...`
/home/mibauer/go/src/github.com/sylabs/singularity/src/cmd/singularity/cli/build.go
/home/mibauer/go/src/github.com/sylabs/singularity/src/cmd/singularity/cli/capability.go
/home/mibauer/go/src/github.com/sylabs/singularity/src/cmd/singularity/cli/capability_add.go
/home/mibauer/go/src/github.com/sylabs/singularity/src/cmd/singularity/cli/capability_drop.go
/home/mibauer/go/src/github.com/sylabs/singularity/src/cmd/singularity/cli/capability_list.go
/home/mibauer/go/src/github.com/sylabs/singularity/src/cmd/singularity/cli/instance.go
/home/mibauer/go/src/github.com/sylabs/singularity/src/cmd/singularity/cli/instance_start.go
/home/mibauer/go/src/github.com/sylabs/singularity/src/cmd/singularity/cli/keys.go
/home/mibauer/go/src/github.com/sylabs/singularity/src/cmd/singularity/cli/keys_list.go
/home/mibauer/go/src/github.com/sylabs/singularity/src/cmd/singularity/cli/keys_newpair.go
/home/mibauer/go/src/github.com/sylabs/singularity/src/cmd/singularity/cli/keys_pull.go
/home/mibauer/go/src/github.com/sylabs/singularity/src/cmd/singularity/cli/keys_push.go
/home/mibauer/go/src/github.com/sylabs/singularity/src/cmd/singularity/cli/keys_search.go
/home/mibauer/go/src/github.com/sylabs/singularity/src/cmd/singularity/cli/push.go
/home/mibauer/go/src/github.com/sylabs/singularity/src/cmd/singularity/cli/sign.go
/home/mibauer/go/src/github.com/sylabs/singularity/src/cmd/singularity/cli/verify.go
/home/mibauer/go/src/github.com/sylabs/singularity/src/pkg/syecl/syecl.go
```

`go list [-f tmpl]` is a fully functional [template](https://golang.org/pkg/text/template/) meaning I was able to get `go list` to output a list of every single file's absolute path and `gofmt -l` can take those paths in. That seems to solve the recursion issue you caught and not forcing us to use `awk`!

DrDaveD: That's impressive, but are you sure you like that better than `go fmt -n ./... | sed 's/ -w / /' | sh`?  I think the go fmt version is a quite a bit easier to understand.  It's certainly a lot fewer characters.  If you're sure, I will make the change. 

Oh, the file list might get too long for the exec buffer on some OS versions.  Probably it would be better to pipe the output of go list into `xargs gofmt -l`.

DrDaveD: I see that the go list form also lists complete path names instead of relative, which I think is a disadvantage for readability.

DrDaveD:
Rebased on #2046, merge that first.

ArangoGutierrez (approved):

PR #2011, opened 2018-09-18
Fix gofmt check

bauerm97:

@DrDaveD 's PR didn't seem to be working correctly for me, so I went ahead and fixed the `go fmt` statement in the Makefile. Also fixed the formatting of files which are incorrectly formatted


DrDaveD:
This fixes only 1 of the 4 problems fixed in #2007, and it doesn't print to stderr the affected files.  As you can see, Travis didn't like it.  @bauerm97 What problem did you have with PR #2007?

I'm suspecting something about different versions of go fmt, because when I run it I don't see the other problems yours found.  Travis didn't find them either.

DrDaveD:
Yes if I run gofmt with go 1.11 I see it flagging the same files that you found.  Is that what you're running?  Travis still uses 1.10.4.

bauerm97:
@DrDaveD Well, #2007 didn't successfully fix the problem that go formatting wasn't being checked. There were formatting errors that are not being caught.

To your other points:
 
- Yes, I'm suspecting that different versions of `go fmt` are catching different things as well (I'm running `go1.11` version)
- `go fmt ./...` isn't a shorthand for recursively in the current directory, but rather it's interpreted by go as all go code in your repository (ignores `vendor`). This is point 3/4 you write under the other PR

I would prefer to use `go fmt` tool instead of `gofmt`. `gofmt` is the underlying implementation that `go fmt` (the format frontend) uses under the hood. Once we commit this change and we're properly testing formatting, running `make check` (or `make testall`) will only update your working tree if you've made changes which aren't properly formatted. It should never touch already committed code, as it will always be properly formatted. I would be happy to merge from your PR if you want to address these changes, and see if we can fix travis to respect the new formatting rules in 1.11.

DrDaveD:
I think I found what was the main thing causing you difficulty: the backquoted gofmt command had to be double-quoted as well.  I updated #2007 with that. 

#2007 did fix everything for the version of go currently used by travis.

Ok on ./... with go fmt, I see how that works now, but I think the problem with the side effect is very bad.  I really don't think we want that in a make check.  It could be very surprising to people if they don't first run go fmt, and that will happen with all newbies that run make check or make testall.  Unfortunately go fmt does not have a way to avoid the side effect as far as I can tell.  gofmt does not understand ./... though so that has to change to src if we use gofmt -l.

DrDaveD:
I see that using `gofmt -l src` leaves out .go files under etc and test.  So it could use `gofmt -l etc src test`.  If you worry about other directories being added in the future, this is the simplest alternative I was able to come up with using go list:
```
gofmt -l `go list -f '{{.Dir}}' ./...|awk '{if ((old=="") || (substr($0,1,length(old)+1) != old "/")) {old=$0; print}}`
```
The awk script removes subdirectories of previously listed directories, since gofmt always recurses.  What do you think?

DrDaveD:
Or `` gofmt -l `find * -maxdepth 0 -type d ! -name vendor` ``

bauerm97:
Addressed by @DrDaveD via #2007

PR #2013, opened 2018-09-18
Add the ability to test ECL from an opened SIF

yhcote:
Make sure there's no race conditions when using an already opened SIF.


src/runtime/engines/singularity/container.go
```diff
@@ -506,7 +506,7 @@ func (c *container) addRootfsMount(system *mount.System) error {
 			if err = ecl.ValidateConfig(); err != nil {
```
cclerget, line 506: Just one point we already discussed about moving ECL validation block in `prepare.go`

cclerget (changes requested):

PR #2020, opened 2018-09-19
Enable --nv option to bind in nvidia drivers

### LINKED ISSUE #1886: Need to implement --nv option
## Version of Singularity:

```
$ singularity --version
singularity version 3.0.0-alpha.1-117-gc297513f
```

### Expected behavior

The `--nv` option will expose the GPU like so:
```
$ singularity exec --nv my_container some command
```

### Actual behavior

The `--nv` option doesn't really do anything yet.

Try using the `--nv` option on a host system with a GPU.

ctmadison:

Add in support for automatically binding Nvidia libraries/drivers with a cmd line --nv argument.

- Fixes #1886

**Checkoff for all PRs:**

Attn: @singularityware-admin


src/pkg/util/nvidiautils/nvidiautils.go
```diff
@@ -0,0 +1,134 @@
...
+	var searchArray []string
+	var commentID = regexp.MustCompile(`#`)
+	var soID = regexp.MustCompile(".so")
+
+	// use nvidia-container-cli (if present)
+	command, err := exec.LookPath("nvidia-container-cli")
+	if err == nil {
+		cmd := exec.Command(command, "list", "--binaries", "--ipcs", "--libraries")
+		out, err := cmd.Output()
+		if err == nil {
```
bauerm97:
```
if err != nil {
    return nil, fmt.Errorf("error running nvidia-container-cli: %v", err)
}
```

src/pkg/util/nvidiautils/nvidiautils.go
```diff
@@ -0,0 +1,134 @@
...
+func GetNvidiaBindPath(abspath string) []string {
+	var strArray []string
+	var bindArray []string
+	var searchArray []string
+	var commentID = regexp.MustCompile(`#`)
+	var soID = regexp.MustCompile(".so")
+
+	// use nvidia-container-cli (if present)
+	command, err := exec.LookPath("nvidia-container-cli")
+	if err == nil {
```
bauerm97:
```
if err != nil {
    return nil, fmt.Errorf("nvidia-container-cli not found: %v", err)
}
```

src/pkg/util/nvidiautils/nvidiautils.go
```diff
@@ -0,0 +1,134 @@
...
+	"os"
+	"os/exec"
+	"path/filepath"
+	"regexp"
+	"strings"
+)
+
+// GetNvidiaBindPath returns a string array consisting of filepaths of nvidia
+// related files to be added to the BindPaths
+func GetNvidiaBindPath(abspath string) []string {
```
bauerm97: `func GetNvidiaBindPath(abspath string) ([]string, error) {`

src/pkg/util/nvidiautils/nvidiautils.go
```diff
@@ -0,0 +1,134 @@
...
+			}
+		}
+	}
+
+	cliEntries := strings.Join(strArray, " ") // save away for later comparison check (disallow duplicates)
+
+	// grab the entries in nvliblist.conf file
+	// use ldconfig to pattern match from ld.so.cache
+	newpath, err := filepath.Glob(abspath + "/nvliblist.conf")
+	if err == nil {
```
bauerm97:
```
if err != nil {
    return nil, fmt.Errorf("filepath.Glob() failed: %v", err)
}
```

src/pkg/util/nvidiautils/nvidiautils.go
```diff
@@ -0,0 +1,134 @@
...
+		for _, filename := range newpath {
+
+			file, err := os.Open(filename)
+			if err == nil {
```
bauerm97:
```
if err != nil {
    return nil, fmt.Errorf("failed to open file %v: %v", filename, err)
}
```

src/pkg/util/nvidiautils/nvidiautils.go
```diff
@@ -0,0 +1,134 @@
...
+				file.Close()
+			}
+		}
+	}
+
+	// walk thru the ldconfig output and add entries which contain the filenames located in
+	// the nvliblist.conf file (ldconfig filenames are full filepaths)
+	var searchFileName string
+	command, err = exec.LookPath("ldconfig")
+	if err == nil {
```
bauerm97:
```
if err != nil {
    return nil, fmt.Errorf("ldconfig not in path: %v", err)
}
```

src/pkg/util/nvidiautils/nvidiautils.go
```diff
@@ -0,0 +1,134 @@
...
+		cmd := exec.Command(command, "-p")
+		out, err := cmd.Output()
+		if err == nil {
```
bauerm97:
```
if err != nil {
    return nil, fmt.Errorf("failed running ldconfig -p: %v", err)
}
```

src/cmd/singularity/cli/actions.go
```diff
@@ -208,6 +209,11 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 		engineConfig.SetImage(abspath)
 	}
 
+	if Nvidia {
+		NvidiaBindPaths := nvidiautils.GetNvidiaBindPath(buildcfg.SINGULARITY_CONFDIR)
```
bauerm97:
```
NvidiaBindPaths, err := nvidiautils.GetNvidiaBindPath(buildcfg.SINGULARITY_CONFDIR)
if err != nil {
    sylog.Fatalf("Unable to capture nvidia bind points: %v", err)
}
```

bauerm97 (changes requested):
This needs some readability tweaks before I'll be able to dive in further, once you fix what I've pointed out I'll take a second glance

src/pkg/util/nvidiautils/nvidiautils.go
```diff
@@ -0,0 +1,134 @@
...
+	"strings"
+)
+
+// GetNvidiaBindPath returns a string array consisting of filepaths of nvidia
+// related files to be added to the BindPaths
+func GetNvidiaBindPath(abspath string) []string {
+	var strArray []string
+	var bindArray []string
+	var searchArray []string
+	var commentID = regexp.MustCompile(`#`)
```
GodloveD: I'm not expert on regular expressions, but this seems like it is saying "if there is a `#` anywhere in the line it's a comment".  

Shouldn't it be like "`^#`" or something?

src/cmd/singularity/cli/actions.go
GodloveD: Maybe it should be a warning instead of a failure though?  That is the way 2.x works.

src/pkg/util/nvidiautils/nvidiautils.go
GodloveD: That's what I was thinking at first too @bauerm97 but I don't think that will work in this instance.  If singularity cannot find `nvidia-container-cli` or if `nvidia-container-cli` returns an error it should just issue a warning and then move on to try to use the conf file instead.

src/pkg/util/nvidiautils/nvidiautils.go
GodloveD: Same comment as above.  As much as I dislike it, I think all of this code actually does need to be wrapped in the if statement b/c we don't want to bail if it doesn't work.

src/pkg/util/nvidiautils/nvidiautils.go
GodloveD: I agree with this though I think it should just spit a warning and then the function should return.

src/pkg/util/nvidiautils/nvidiautils.go
GodloveD: I agree with this though I think it should just spit a warning and then the function should return.

src/pkg/util/nvidiautils/nvidiautils.go
GodloveD: I agree with this though I think it should just spit a warning and then the function should return.

src/pkg/util/nvidiautils/nvidiautils.go
GodloveD: I agree with this though I think it should just spit a warning and then the function should return.

src/pkg/util/nvidiautils/nvidiautils.go
```diff
@@ -0,0 +1,134 @@
...
+			for _, line := range strings.Split(string(out), "\n") {
+				val := soID.FindString(line) // this will disallow binaries (non .so files)
+				if val != "" {               // contains .so
+					if line != "" {
+						// extract the filename from the path
+						fileNames := strings.SplitAfter(line, "/")
+						fileName := fileNames[len(fileNames)-1]
+
+						testString := line + ":/.singularity.d/libs/" + fileName
+						bindArray = append(bindArray, testString)
```
GodloveD: appending to a string array in a loop sounds like it will be very slow to me.  I think there are faster ways to do this?  But I don't want us spending time on premature optimization if this is already pretty quick.

src/pkg/util/nvidiautils/nvidiautils.go
```diff
@@ -0,0 +1,134 @@
...
+					}
+				} else { // binary executable
+					bindArray = append(bindArray, line)
+					strArray = append(strArray, line)
+				}
+			}
+		}
+	}
+
+	cliEntries := strings.Join(strArray, " ") // save away for later comparison check (disallow duplicates)
```
GodloveD: Is it necessary to have the second variable `strArray` here?  Can `cliEntries` be constructed from `bindArray`?

src/pkg/util/nvidiautils/nvidiautils.go
ctmadison: Hmmm... nvidia-container-cli is an optional component. Don't want to return if error.

src/pkg/util/nvidiautils/nvidiautils.go
ctmadison: Hmmm... nvidia-container-cli is an optional component. Don't want to return if error.

src/pkg/util/nvidiautils/nvidiautils.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// grab the entries in nvliblist.conf file
+	// use ldconfig to pattern match from ld.so.cache
+	newpath, err := filepath.Glob(abspath + "/nvliblist.conf")
```
GodloveD: Oh so we are now using the _union_ of what we find in the conf file and what `nvidia-container-cli` yields?  I'm not sure I like that.  Kindof think we should default to `nvidia-container-cli` output and only fall back to conf file if it fails.

ArangoGutierrez (changes requested):
`make install` should cp `etc/nvliblist.conf`  to `$PREFIX/etc/nvliblist.conf`

src/pkg/util/nvidiautils/nvidiautils.go
bauerm97: Definitely want to error if we're unable to execute `nvidia-container-cli` if it does exist in `$PATH`

GodloveD (changes requested):
I'm also not done looking at this yet, but in the interest of time I'm making a first round of comments.

src/pkg/util/nvidiautils/nvidiautils.go
bauerm97: Fair enough.

src/pkg/util/nvidiautils/nvidiautils.go
GodloveD: No I disagree. `nvidia-container-cli` is still in a pretty early state afaik.  They have already changed he cli on me once.  If the user has a newer version of the binary installed that doesn't do what we expect we still need the option of falling back to the conf file instead.

src/pkg/util/nvidiautils/nvidiautils.go
bauerm97: If you think it should spit a warning, then the person who calls this function can look at the error and determine if it should warn or error

src/pkg/util/nvidiautils/nvidiautils.go
bauerm97: In this case, I would actually say to break these into functions:

```
func nvidiaContainerCli() ([]string, error) {
    // use nvidia-container-cli (if present)
    command, err := exec.LookPath("nvidia-container-cli")
    if err =! nil {
        return nil, nil
    }
}      
```

Then within the main code can simply do this:

```
s, err := nvidiaContainerCli()
if err != nil {
    return nil, fmt.Println("Unable to do something...: %v", err)
}
...
```

src/pkg/util/nvidiautils/nvidiautils.go
```diff
@@ -0,0 +1,134 @@
...
+	if err == nil {
+		for _, filename := range newpath {
```
GodloveD: why is this in a for loop?  It should only be a single file right?

src/pkg/util/nvidiautils/nvidiautils.go
```diff
@@ -0,0 +1,134 @@
...
+
+			file, err := os.Open(filename)
```
GodloveD: I think you can use `defer` here to close when done?  Is that right?

src/pkg/util/nvidiautils/nvidiautils.go
```diff
@@ -0,0 +1,134 @@
...
+			if err == nil {
+				scanner := bufio.NewScanner(file)
+				for scanner.Scan() {
+					line := scanner.Text()
+					val := commentID.FindString(line)
```
GodloveD: For a possible simpler and more readable solution check this out (though I don't know if performance would suffer greatly).
https://golang.org/pkg/strings/#HasPrefix

src/pkg/util/nvidiautils/nvidiautils.go
ctmadison: Ok, I like this better...

src/pkg/util/nvidiautils/nvidiautils.go
jmstover: nvidia-container-cli didn't provide everything though right? I thought there was some stuff missing.

src/pkg/util/nvidiautils/nvidiautils.go
ctmadison: I don't think so. Because the bind array has entries like /dir/dir/file:/.singularity.d/file in it. I don't think it would be any faster and it would be more work. Are you thinking of a specific method?

src/pkg/util/nvidiautils/nvidiautils.go
ctmadison: I don't know another faster way (maybe other than building everything from scratch using buffers, etc.)

src/pkg/util/nvidiautils/nvidiautils.go
```diff
@@ -0,0 +1,134 @@
...
+	if err == nil {
+		cmd := exec.Command(command, "-p")
+		out, err := cmd.Output()
+		if err == nil {
+			lastadd := ""
+			for _, ldconfigOutputline := range strings.Split(strings.TrimSuffix(string(out), "\n"), "\n") {
+				if ldconfigOutputline != "" {
+
+					for _, nvidiaConfFileline := range searchArray {
+						if nvidiaConfFileline != "" {
```
GodloveD: This should already be handled above right?  nvidiaConfFileline should not include empty entries afaik.

https://github.com/singularityware/singularity/blob/672f0e753500b55b26a04f736fb72b3c47ba3216/src/pkg/util/nvidiautils/nvidiautils.go#L36

src/pkg/util/nvidiautils/nvidiautils.go
ctmadison: probably....

src/pkg/util/nvidiautils/nvidiautils.go
ctmadison: Well, the nvidia-container-cli function only returns the /dir/file.so.xxx.xx entries (where the xxx.xx is the driver version). I can't just strip off the .xxx.xx and use the so entries because some things (like nvidia-smi binary look for the file.so.1 entry). I could make it work if that is the way you think I should go...

src/pkg/util/nvidiautils/nvidiautils.go
ctmadison: agreed

GodloveD (changes requested):

src/pkg/util/nvidiautils/nvidiautils.go
GodloveD: I think what you want to do is use _either_ the output of `nvidia-container-cli` _or_ the output of the `nvidia.conf` file as a set of things to search the `ld.so.cache` for (through `ldconfig` output).  

Does that make sense?  In other words if you get usable output from `nvidia-container-cli` skip parsing the `nvidia.conf` file and just proceed directly to pattern matching between `nvidia-container-cli` output and `ldconfig` output.

src/pkg/util/nvidiautils/nvidiautils.go
ctmadison: yes, you are correct.

src/pkg/util/nvidiautils/nvidiautils.go
ctmadison: In my research, it was recommended to NOT use defer. I had it that way at first but changed it. But I will defer to others with more go experience...

src/pkg/util/nvidiautils/nvidiautils.go
GodloveD: If you want to just append `.1` to all of the `nvidia-container-cli` entries and search for them as well, I think it's reasonable?  

@jmstover yes there was a recent issue where a user was confused b/c `nvidia-container-cli` does not provide open GL libraries, but that behavior is correct and our behavior is wrong.  We should comment those libs out of the conf file imho because those libs are not strictly tied to the version of the driver as the other libs are.

src/pkg/util/nvidiautils/nvidiautils.go
ctmadison: agreed.

src/pkg/util/nvidiautils/nvidiautils.go
jmstover: @GodloveD 
Maybe not on _recent_ drivers with libglvnd. But my `375.39` driver version, the GL libs *are* tied to the driver version.

i.e.  libGLX_nvidia.so.375.39

And we had cards where we couldn't update past a specific driver version because they removed support. So... no recent driver to have any of the new fancy stuff.

src/pkg/util/nvidiautils/nvidiautils.go
ctmadison: Yes, I started using more string. functions after I wrote this code and didn't take time to go back and change it.

src/pkg/util/nvidiautils/nvidiautils.go
GodloveD: @ctmadison I just had a brain wave.  I think you need to strip driver version from the `nvidia-container-cli` output and then use that as a _prefix_ when searching the output of `ldconfig`.  Tht way you will get any stray `.so.1` files (and other file appends) too.

ctmadison: Yes, that should work. But, there is still the issue that @jmstover brought up about the GL libs that we have to agree on.

src/pkg/util/nvidiautils/nvidiautils.go
```diff
@@ -0,0 +1,142 @@
...
+
+	// use nvidia-container-cli (if present)
+	command, err := exec.LookPath("nvidia-container-cli")
+	if err != nil {
+		return nil, nil, nil
+	}
+	cmd := exec.Command(command, "list", "--binaries", "--ipcs", "--libraries")
+	out, err := cmd.Output()
+	if err != nil {
+		return nil, nil, nil
```
bauerm97: If we're unable to execute the `nvidia-container-cli` command, we *should* return an error:

```
if err != nil {
    return nil, nil, fmt.Errof("Unable to execute ldconfig: %v", err)
}
```

We should only be ignoring errors in the case where `nvidia-container-cli` doesn't exist in path

src/pkg/util/nvidiautils/nvidiautils.go
```diff
@@ -0,0 +1,142 @@
...
+// related files to be added to the BindPaths
+func GetNvidiaBindPath(abspath string) ([]string, error) {
+	var strArray []string
+	var bindArray []string
+	var searchArray []string
+	var commentID = regexp.MustCompile(`#`)
+
+	bindArray, strArray, err := nvidiaContainerCli()
+	if err != nil {
+		sylog.Warningf("no nvidia-container-cli returned: %v", err)
```
bauerm97: If we're only returning errors which *shouldn't* be ignored, then we should `sylog.Fataf(...)` instead.

src/pkg/util/nvidiautils/nvidiautils.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	cliEntries := strings.Join(strArray, " ") // save away for later comparison check (disallow duplicates)
+
+	// grab the entries in nvliblist.conf file
+	// use ldconfig to pattern match from ld.so.cache
+	file, err := os.Open(abspath + "/nvliblist.conf")
```
bauerm97: `defer file.Close()`

ctmadison: ok

src/pkg/util/nvidiautils/nvidiautils.go
```diff
@@ -0,0 +1,159 @@
...
+	cmd = exec.Command(command, "list", "--ipcs", "--libraries")
+	out, err = cmd.Output()
+	if err != nil {
+		return nil, nil, fmt.Errorf("Unable to execute nvidia-container-cli: %v\n", err)
+	}
+
+	for _, line := range strings.Split(string(out), "\n") {
+		if line != "" {
+
+			fileName := filepath.Base(line)
```
GodloveD, line 38: I'm not sure this will work when it gets input that does not include `/` or does include whitespace or whatever, but I cannot test b/c I don't have access to a GPU right now.

src/pkg/util/nvidiautils/nvidiautils.go
```diff
@@ -0,0 +1,159 @@
...
+
+					// ldconfigOutputSplitline[0] is the "libnvidia-ml.so[.1] (libc6,x86-64)"" (from the above example)
+					// ldconfigOutputSplitline[1] is the "/usr/lib64/nvidia/libnvidia-ml.so[.1]" (from the above example)
+					// these next 2 lines extract the "libnvdia-ml.so[.1]" (from the above example)
+
+					// ldconfigFileName is "libnvidia-ml.so[.1]" (from the above example)
+					ldconfigFileNames := strings.Split(ldconfigOutputSplitline[0], " ")
+					ldconfigFileName := strings.TrimSpace(string(ldconfigFileNames[0]))
+
+					if strings.HasPrefix(ldconfigFileName, nvidiaFileName) && ldconfigFileName != lastadd { // add if not duplicate
```
GodloveD: I think the addition of `strings.HasPrefix()` will pick up any `*.so.1` libs or others with different suffixes.  But needs to be tested to be sure.

GodloveD (dismissed):
I looked this over and added a few changes.  If they all work and are acceptable I approve.

jmstover:
`$(SRC)/etc/nvliblist.conf` is not being installed to $(PREFIX)/etc/singularity/nvliblist.conf on `make install`

src/cmd/singularity/cli/actions.go
```diff
@@ -206,6 +207,20 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
...
 	}
 
+	if Nvidia {
+		NvidiaBindPaths, err := nvidiautils.GetNvidiaBindPath(buildcfg.SINGULARITY_CONFDIR)
+		if err != nil {
+			sylog.Infof("Unable to capture nvidia bind points: %v", err)
+		} else {
+			if len(BindPaths) == 0 {
+				sylog.Warningf("Could not find any NVIDIA libraries on this host!")
+				sylog.Warningf("You may need to edit %v/singularity/nvliblist.conf", buildcfg.SINGULARITY_CONFDIR)
```
jmstover: This should be:  `%v/nvliblist.conf` ...   buldcfg.SINGULARITY_CONFDIR includes the singularity/ directory

src/cmd/singularity/cli/actions.go
```diff
@@ -206,6 +207,20 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 		engineConfig.SetImage(abspath)
 	}
 
+	if Nvidia {
+		NvidiaBindPaths, err := nvidiautils.GetNvidiaBindPath(buildcfg.SINGULARITY_CONFDIR)
+		if err != nil {
+			sylog.Infof("Unable to capture nvidia bind points: %v", err)
+		} else {
+			if len(BindPaths) == 0 {
```
jmstover: Should this be if `len(NvidiaBindPaths)` ?

jmstover (dismissed):

ArangoGutierrez:
Travis is not happy due
```
 CHECK go lint
/home/travis/gopath/src/github.com/sylabs/singularity/src/pkg/util/nvidiautils/nvidiautils.go:27:31: error strings should not be capitalized or end with punctuation or a newline
/home/travis/gopath/src/github.com/sylabs/singularity/src/pkg/util/nvidiautils/nvidiautils.go:34:31: error strings should not be capitalized or end with punctuation or a newline
/home/travis/gopath/src/github.com/sylabs/singularity/src/pkg/util/nvidiautils/nvidiautils.go:51:31: error strings should not be capitalized or end with punctuation or a newline
```

jmstover (approved):
This is fully working for me now.

ArangoGutierrez (approved):

ArangoGutierrez:
this img was rendered using the `--nv`

ArangoGutierrez:
```bash
[eduardo@rhel7 tmp]$ nvidia-container-[TAB]
nvidia-container-cli           nvidia-container-runtime       nvidia-container-runtime-hook  
[eduardo@rhel7 tmp]$ singularity  exec --nv docker://ubuntu nvidia-smi
Fri Sep 21 21:13:47 2018       
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 396.37                 Driver Version: 396.37                    |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  Quadro M620         Off  | 00000000:01:00.0  On |                  N/A |
| N/A   51C    P8    N/A /  N/A |   1055MiB /  1999MiB |      0%      Default |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                       GPU Memory |
|  GPU       PID   Type   Process name                             Usage      |
|=============================================================================|
|    0      2099      G   /usr/bin/X                                   229MiB |
|    0      2840      G   /usr/bin/gnome-shell                         354MiB |
|    0      3598      G   ...-token=06CB68BCB817E1EXXXXXXXXXXXXXXXXX    90MiB |
|    0      3788      G   ...uest-channel-token=17275785424XXXXXXXXX   134MiB |
|    0      4275      G   ...-token=5848D7EB3EABAXXXXXXXXXXXXXXXXXXX    74MiB |
|    0     30710      G   ...-type=gpu-process --disable-features=Co   166MiB |
+-----------------------------------------------------------------------------+
```

GodloveD (approved):
This looks good to me now.

bauerm97 (approved):

bauerm97:
@GodloveD Good to merge once Travis passes

GodloveD:
Thanks @ctmadison @jmstover @ArangoGutierrez and @bauerm97 !

PR #2048, opened 2018-09-21
3.0: Fix rpm build after introduction of cni

DrDaveD:

This PR fixes the rpm build which was broken by the introduction of cni.

None

**Checkoff for all PRs:**


mlocal/frags/Makefile.stub
```diff
@@ -69,7 +69,7 @@ cniplugins: $(BUILDDIR)/vendors-done
 	@for p in `go list -f '{{if eq .Name "main"}}{{.ImportPath}}{{end}}' ../vendor/$(cni_plugins_REPO)/...|grep -v sample`; do \
 		if [ ! -f $(cni_builddir)/`basename $$p` ]; then \
 			echo " CNI PLUGIN" `basename $$p`; \
-			go build -o $(cni_builddir)/`basename $$p` $$p; \
+			go build -ldflags="-s" -o $(cni_builddir)/`basename $$p` $$p; \
```
jmstover: If adding `-s` here, would we also want to add the `-w` flag?

DrDaveD: I had looked into that, but couldn't find what -w does in the ld man page.  Do you know?  -s is for stripping symbols, which is what is needed because otherwise rpm tried to extract the symbols and failed because it couldn't identify the "build id".

jmstover: IIRC, `-w` removes the DWARF debug info.

jmstover: Yeah... 
```
$ go tool link 2>&1 | egrep '\-s|\-w'
  -s	disable symbol table
  -w	disable DWARF generation
```

DrDaveD: Ok I added -w

jmstover (approved):
@yhcote ... Can you give a once over of the `Makefile.stub` changes

mlocal/frags/Makefile.stub
```diff
@@ -72,7 +72,7 @@ cniplugins: $(BUILDDIR)/vendors-done
 	@for p in `go list -f '{{if eq .Name "main"}}{{.ImportPath}}{{end}}' ../vendor/$(cni_plugins_REPO)/...|grep -v sample`; do \
 		if [ ! -f $(cni_builddir)/`basename $$p` ]; then \
 			echo " CNI PLUGIN" `basename $$p`; \
-			go build -o $(cni_builddir)/`basename $$p` $$p; \
+			go build -ldflags="-s -w" -o $(cni_builddir)/`basename $$p` $$p; \
```
yhcote, line 82: Why the stripping of symbols? In the advent that something goes wrong in those modules at runtime, it would become more difficult to figure out panics and what happened in the wild.

DrDaveD: It was already being done for the primary singularity executables.  If it is not done, then rpm build fails as it tries to extract debug info into a separate rpm.  It complains about not being able to find the build id.  I found a lot of discussion about this problem with go and rpm and the only advice I found that worked was to strip the symbols.

yhcote: We need to final binaries to be including debugging symbols. It's a must. I will review the build to make sure it's true for all components. Maybe it slipped without me noticing. If it's a problem with RPM building, I think we probably need a separate "rpm" build rule (make rpm), that builds all, and do all other manipulation on the results for packaging. What do you think?

jmstover: > We need to final binaries to be including debugging symbols.

Why on the _release_ binaries include the debugging symbols? Why not a separate `make debug` target that doesn't strip the symbols, or by setting a DEBUG=1 var, or similar?

DrDaveD: How do people do debugging of go programs?  Does gdb work well with it, for example, if symbols are available?

Rpm normally has a great way of handling symbols, by automatically putting them in a separate debuginfo package that people can install if they want to do debugging, but the mechanism doesn't work at least the way that the singularity Makefile is currently set up.  It would be good if we could figure out a way to make it work.

yhcote: I agree, I can help you creating Makefile anchors for that feature this week.

DrDaveD: So @yhcote can we merge this first to fix rpm builds again, then you later add a separate debug target?

DrDaveD: @yhcote can we merge this?

yhcote: We don't want debug / release targets. In go there's no optimized vs unoptimized binaries built per convention. The idea is weather you carry debug symbols or not. And we always should. If we don't and there's a panic, very little information will be available to debug.

What we need is a normal build of our binaries and apparently a "make rpm" target that is a special case where we strip symbols from the binaries.

I would really prefer we come up with a "make rpm" rule now that builds "all" but then add a stripping pass to the target. Then make rpm could also call the rpmbuild script you already included in the project. This also improve usability to easily build and package and rpm by simply typing "make rpm". 

What do you think?

DrDaveD: I think it is fine to have an rpm target that enables stripping of symbols.  It will still need to be done by setting a make variable like jmstover suggested, and it could be used for other purposes as well.

Please note however that this PR does not introduce symbol-stripping, it only extends it to these new CNI binaries.  The starter binary has been being stripped for quite some time without anybody doing anything about it, at least 3 months according to the git history; it must not be very urgent.  So please let's merge this PR to fix the broken functionality of rpm building, and then prepare another one for the enhancement of stripping optionally.  Enhancements are much less urgent than broken things.

DrDaveD: In fact the stripping of the helper binary has been there since its first successful build put in April 20 in this [commit](https://github.com/sylabs/singularity/commit/05a847739856ded2ebe9b9e000d5b2c473906dcd) from @bauerm97.   So it was definitely not put there to support rpms and there may be another important reason for the stripping to be there.

DrDaveD: Ok @yhcote?  rpm building has been broken for 2 weeks, let's get it working again.

yhcote: I am working on a patch to cleanup a lot of makeit before release an hopefully this also address that issue. I will have that ready tonight or tomorrow.

DrDaveD: Ok but most of this PR isn't related to this one line, it is related to the extra cni files and relocating them, so can we merge this PR first?

yhcote: Ok, updating the branch and waiting for CI...

PR #2055, opened 2018-09-21
Library Conveyor Packer

### LINKED ISSUE #2039: enable Bootstrap: library
## Version of Singularity:

enable building from library 

```
Bootstrap: library
From: sylabsed/linux/alpine:3.7

%runscript

%post
    echo "Install additional software here"
```

### Expected behavior

be able to use the above recipe

ikaneshiro:
Enables building from library.

This changes the `NewBuild()` function call to include a libraryurl and authtoken for the library conveyorPacker to use.

fixes #2039

### DIFF
--- a/src/pkg/build/sources/conveyorPacker_library_test.go
+++ b/src/pkg/build/sources/conveyorPacker_library_test.go
@@ -6,7 +6,6 @@
 package sources_test
 
 import (
-	"fmt"
 	"testing"
 
 	"github.com/sylabs/singularity/src/pkg/build/sources"
@@ -15,6 +14,7 @@ import (
 )
 
 const (
+	libraryURL = "https://cloud.sylabs.io/library"
 	libraryURI = "library://dtrudg/linux/alpine:latest"
 )
 
@@ -38,13 +38,13 @@ func TestLibraryConveyor(t *testing.T) {
 		t.Fatalf("unable to parse URI %s: %v\n", libraryURI, err)
 	}
 
-	fmt.Println(b.Recipe)
-
-	cp := &sources.LibraryConveyorPacker{}
+	cp := &sources.LibraryConveyorPacker{
+		LibraryURL: libraryURL,
+	}
 
 	err = cp.Get(b)
 	// clean up tmpfs since assembler isnt called
-	defer cp.CleanUp()
+	//defer cp.CleanUp()
 	if err != nil {
 		t.Fatalf("failed to Get from %s: %v\n", libraryURI, err)
 	}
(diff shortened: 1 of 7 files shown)


ArangoGutierrez:
This PR works for me, PING me when you fix the fmt issues and travis gets happy

src/pkg/build/sources/conveyorPacker_library_test.go
```diff
@@ -24,9 +23,6 @@ func TestLibraryConveyor(t *testing.T) {
 		t.SkipNow()
 	}
 
-	test.DropPrivilege(t)
```
ArangoGutierrez: use 
```go
// EnsurePrivilege ensures elevated privileges are available during a test.
func EnsurePrivilege(t *testing.T) {
	uid := os.Getuid()
	if uid != 0 {
		t.Fatal("test must be run with privilege")
	}
}
```

ArangoGutierrez (changes requested):

ArangoGutierrez (approved):

src/pkg/build/sources/conveyorPacker_library_test.go
```diff
@@ -0,0 +1,80 @@
...
+	libraryURI = "library://dtrudg/linux/alpine:latest"
+)
+
+// TestLibraryConveyor tests if we can pull an image from singularity hub
+func TestLibraryConveyor(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
+
+	test.EnsurePrivilege(t)
```
ArangoGutierrez, line 27: a last niptick, this line should be at the beginning of the func (not a blocker, already approved)

src/pkg/build/sources/conveyorPacker_library_test.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// TestLibraryPacker checks if we can create a Bundle from the pulled image
+func TestLibraryPacker(t *testing.T) {
+	b, err := types.NewBundle("sbuild-library")
+	if err != nil {
+		return
+	}
+
+	test.EnsurePrivilege(t)
```
ArangoGutierrez: a last niptick, this line should be at the beginning of the func (not a blocker, already approved)

src/pkg/build/sources/conveyorPacker_library_test.go
ikaneshiro: The way I'm thinking about it, I think it makes more sense to leave this after the skip check because if we are skipping i don't think we should be doing anything at all in the test case. Thoughts? I agree with the one below and will fix now.

ArangoGutierrez: agree

dtrudg (approved):
:+1: from me @ikaneshiro 

The NewBuild call is getting a bit long in the argument list... but tidying that up into a struct or something isn't a priority now.

Maybe worth checking with @tri-adam if change to the NewBuild call will need him to modify anything in builder?

bauerm97 (approved):
Looks good. Feel free to merge whenever travis passes after updating with master! Cheers @ikaneshiro @ArangoGutierrez

PR #2078, opened 2018-09-24
Sign and verify optional SIF IDs and Groups

yhcote:
This patch allows to selectively sign and verify specified SIF ids (-i) and groups (-g) instead of the default, which will sign and verify the primary partition.


cclerget (approved):
Looks good !

PR #2083, opened 2018-09-24
Add support for apps to Singularity 3.0

bauerm97:

As title says, this PR adds apps to Singularity v3.0.0 in a very "plugin-like" manner. This will give us a nice clean way of building apps as a plugin when plugin support is fully introduced in 3.1, this is very much a prototype of that functionality (except linked statically instead of dynamically)

### DIFF
--- a/src/pkg/syplugin/build.go
+++ b/src/pkg/syplugin/build.go
@@ -9,6 +9,8 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/sylabs/singularity/src/pkg/apps"
+	"github.com/sylabs/singularity/src/pkg/build/types"
 	"github.com/sylabs/singularity/src/pkg/sylog"
 )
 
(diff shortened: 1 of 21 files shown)


src/pkg/syplugin/build.go
```diff
@@ -0,0 +1,117 @@
...
+
+// BuildPlugin is the interface for plugins on the build system
+type BuildPlugin interface {
+	Name() string
+	HandleSection(string, string)
+	HandleBundle(*types.Bundle)
+	HandlePost() string
+}
+
+// func init() {
```
yhcote, line 114: Do you want to remove that ?

src/plugins/apps/apps.go
```diff
@@ -0,0 +1,318 @@
...
+
+func appMeta(b *types.Bundle, a *App) string {
+	return filepath.Join(appBase(b, a), "/scif/")
+}
+
+func appData(b *types.Bundle, a *App) string {
+	return filepath.Join(b.Rootfs(), "/scif/data/", a.Name)
+}
+
+func writeFile(path string, perm os.FileMode, s string) error {
```
yhcote, line 293: what about ioutil/WriteFile()?

src/pkg/syplugin/build.go
bauerm97: Yea, I'll make a PR removing the comments once we tag beta

yhcote (approved):
Looks OK to me.

PR #2091, opened 2018-09-25
Development 3.0: Add actions script check for old formats

### LINKED ISSUE #2014: singularity instance start does not work, provides no error
## Version of Singularity:

$ singularity --version
singularity version 3.0.0-alpha.1-468-gb034a40-dirty

### Expected behavior

$ singularity instance start Default.sif www
$ echo $?
0
$ singularity instance list
<instance info shown here>

### Actual behavior

$ singularity instance start Default.sif www
$ echo $?
1
$ singularity instance list
$ singularity --verbose instance start Default.sif www
$ echo $?
1

$ sudo singularity build Default.sif examples/busybox/Singularity
INFO:    Starting build...
INFO:    Adding runscript
INFO:    Creating SIF file...
INFO:    Build complete!
$ singularity instance start Default.sif www
$ echo $?
1
$ singularity instance list
$ singularity --verbose instance start Default.sif www
$ echo $?
1

cclerget:

This PR add actions script checks for old formats.

- Fixes #2014

**Checkoff for all PRs:**

### DIFF
--- a/src/runtime/engines/singularity/process.go
+++ b/src/runtime/engines/singularity/process.go
@@ -28,6 +28,12 @@ import (
 )
 
 func (engine *EngineOperations) checkExec() error {
+	shell := engine.EngineConfig.GetShell()
+
+	if shell == "" {
+		shell = "/bin/sh"
+	}
+
 	args := engine.EngineConfig.OciConfig.Process.Args
 	env := engine.EngineConfig.OciConfig.Process.Env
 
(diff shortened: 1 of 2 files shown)


src/runtime/engines/singularity/process.go
```diff
@@ -28,6 +27,80 @@ import (
...
+	}()
+
+	for _, keyval := range env {
+		if strings.HasPrefix(keyval, "PATH=") {
+			os.Setenv("PATH", keyval[5:])
+			break
+		}
+	}
+
+	if _, err := exec.LookPath(args[0]); err == nil {
```
bauerm97, line 57: Would add a comment here:

`// If args[0] is an absolute path, exec.LookPath() looks for this file directly instead of within PATH`

src/runtime/engines/singularity/process.go
```diff
@@ -28,6 +27,80 @@ import (
...
+		return nil
+	}
+
+	switch args[0] {
```
bauerm97, line 63: I would add a comment for clarity:
`// If args[0] isn't executable (either via PATH or absolute path), look for alternative approaches to handling it`

src/runtime/engines/singularity/process.go
```diff
@@ -99,7 +154,7 @@ func (engine *EngineOperations) StartProcess(masterConn net.Conn) error {
 	}
 
 	// Spawn and wait container process, signal handler
-	cmd := exec.Command(args[0], args...)
+	cmd := exec.Command(args[0], args[1:]...)
```
bauerm97, line 167: What's the purpose of this change here?

bauerm97 (changes requested):
Other than those two comments I think we should add for clarity, I really like this change.

src/runtime/engines/singularity/process.go
cclerget: To avoid redundant first argument, `Command` is not `syscall.Exec` :)

bauerm97 (approved):

PR #2100, opened 2018-09-26
Development 3.0: Provide debug output for instance and information status

### LINKED ISSUE #2014: singularity instance start does not work, provides no error

cclerget:

This PR provides full debug output when starting instance and provides information status in case of success/failure

- Fixes #2014

**Checkoff for all PRs:**

### DIFF
--- a/src/cmd/singularity/cli/actions.go
+++ b/src/cmd/singularity/cli/actions.go
@@ -9,7 +9,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-	osexec "os/exec"
 	"path/filepath"
 	"runtime"
 	"strconv"
@@ -533,21 +532,14 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 			sylog.Fatalf("failed to create instance log files: %s", err)
 		}
 
-		pipefd, err := exec.SetPipe(configData)
-		if err != nil {
-			sylog.Fatalf("failed to set pipe communication: %s", err)
-		}
-
-		cmd := osexec.Command(starter, "")
 		start, err := stderr.Seek(0, os.SEEK_END)
 		if err != nil {
 			sylog.Warningf("failed to get standard error stream offset: %s", err)
 		}
 
+		cmd, err := exec.PipeCommand(starter, []string{procname}, Env, configData)
 		cmd.Stdout = stdout
 		cmd.Stderr = stderr
-		cmd.Env = append(Env, pipefd)
-		cmd.Args[0] = procname
 
 		cmdErr := cmd.Run()
 
--- a/src/cmd/singularity/cli/actions.go
+++ b/src/cmd/singularity/cli/actions.go
@@ -7,12 +7,15 @@ package cli
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
+	osexec "os/exec"
 	"path/filepath"
 	"runtime"
 	"strconv"
 	"strings"
 	"syscall"
+	"time"
 
 	"github.com/opencontainers/runtime-tools/generate"
 	"github.com/sylabs/singularity/src/pkg/util/nvidiautils"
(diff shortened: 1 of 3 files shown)


src/cmd/singularity/cli/actions.go
```diff
@@ -466,7 +464,56 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
...
+		if err != nil {
+			sylog.Fatalf("failed to create instance log files: %s", err)
+		}
+
+		pipefd, err := exec.SetPipe(configData)
+		if err != nil {
+			sylog.Fatalf("failed to set pipe communication: %s", err)
+		}
+
+		cmd := osexec.Command(starter, "")
```
bauerm97: Maybe rebase on master, then check out the new function `src/pkg/util/exec/pipe.go:PipeCommand()`, and see if that fits your use case here

bauerm97 (changes requested):

src/cmd/singularity/cli/actions.go
cclerget: Unfortunately that won't fit here, need to redirect stdout/stderr for command file logging

bauerm97: Hmm, I think it will still work here. Look at function signature of `PipeCommand()`:

`func PipeCommand(command string, args []string, env []string, data []byte) (*exec.Cmd, error)`

`PipeCommand()` returns `*exec.Cmd` type, which is same type as `cmd` you're using here. So I think you can do this:

```
cmd, err := exec.PipeCommand(starter, "", Env, configData)
if err != nil {
    //fatal
}

// do stderr.Seek...

cmd.Stdout = stdout
cmd.Stderr = stderr

cmdErr := cmd.Run()
```

bauerm97: `PipeCommand()` func will automatically set environment variable `PIPE_EXEC_FD` to the fd of the pipe file for the `exec.Cmd` type, and will also ensure that the pipe file remains open and passed to `starter` binary.

src/cmd/singularity/cli/actions.go
```diff
@@ -466,7 +464,56 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
...
+		if err != nil {
+			sylog.Warningf("failed to get standard error stream offset: %s", err)
+		}
+
+		cmd.Stdout = stdout
+		cmd.Stderr = stderr
+		cmd.Env = append(Env, pipefd)
+		cmd.Args[0] = procname
+
+		cmdErr := cmd.Run()
```
bauerm97, line 544: Is intended to use `cmd.Run()` since `Run()` method blocks until `cmd` exits (either successfully or not-successfully)?

src/cmd/singularity/cli/actions.go
cclerget: Ok I look too quickly, that would work. Thanks Michael !

src/cmd/singularity/cli/actions.go
cclerget: Yes, instance always returns, so `cmd.Run()` will block until instance start or fail to provide information about success/failure

cclerget:
@bauerm97 Now use `PipeCommand`. I took the opportunity to simplify code and remove some code redundancies in `src/pkg/util/exec/pipe.go`. Also fixed the hardcoded PIPE_EXEC_FD value (always set to 3) used in `PipeCommand`.

bauerm97:
@cclerget Reason that I hardcoded `PIPE_EXEC_FD=3` is because I was passing the pipe `*os.File` to the `os/exec.Cmd` struct's `Files` field. When reading `os/exec` package I wasn't clear if open files are always passed to new process when not explicitly adding to `Files` of the `Cmd`. But it looks like this is working properly so it's ok this way

bauerm97 (approved):

PR #2110, opened 2018-09-27
Chown images after build with sudo

### LINKED ISSUE #2102: chown built images
## Version of Singularity:

v3.0.0-beta

when we do a `sudo singularity build...` I think we should chown to the user to avoid this:
```bash
ERROR:   signing container failed: failed to load SIF container file: opening(RDWR) container file: open demo/nginx.sif: permission denied
```
### Expected behavior

```bash
sudo singularity build my.sif DefFile
ls -lh my.sif
-rwxrwxr-x. 1 eduardo eduardo 187M Sep 21 16:54 my.sif
```
### Actual behavior
```bash
sudo singularity build my.sif DefFile
ls -lh my.sif
-rwxrwxr-x. 1 root root 187M Sep 21 16:54 my.sif
```

ArangoGutierrez:
- Fixes #2102

Signed-off-by: ArangoGutierrez <eduardo@sylabs.io>

> test on RHEL7 - Go 1.11
```bash
[eduardo@rhel7 singularity]$ sudo singularity build test.sif examples/busybox/Singularity
INFO:    Starting build...
<...>
[eduardo@rhel7 singularity]$ ls -lh test.sif
-rwxr-xr-x. 1 eduardo eduardo 640K Sep 27 10:46 test.sif
```
**Checkoff for all PRs:**


src/pkg/build/assemblers/assembler_sif.go
```diff
@@ -77,6 +79,14 @@ func createSIF(path string, definition []byte, squashfile string) (err error) {
 		return fmt.Errorf("while creating container: %s", err)
 	}
 
+	// chown the sif file to the calling user
+	if syscall.Getuid() == 0 {
+		uID, _ := strconv.Atoi(os.Getenv("SUDO_UID"))
```
bauerm97: We shouldn't be ignoring errors here, *especially* since this is touching security

bauerm97: Also, I think `uid` looks better than `uID`

src/pkg/build/assemblers/assembler_sif.go
```diff
@@ -77,6 +79,14 @@ func createSIF(path string, definition []byte, squashfile string) (err error) {
 		return fmt.Errorf("while creating container: %s", err)
 	}
 
+	// chown the sif file to the calling user
+	if syscall.Getuid() == 0 {
```
bauerm97: What about the case where the calling user is root. Not running `sudo ...` but they are logged in as the root user? In this case, `$SUDO_UID` may not be set, and since we're not checking for `Atoi` errors, we can't be guaranteed of the behavior

bauerm97 (changes requested):
In general, I'm not a fan of the approach of automatically calling `chown` on an image built with `uid=0`. This seems unexpected to me, especially because we're not handling the case where the user is actually logged in as root, and not just running as sudo. 

I think this feature would be better served by implementing a flag to optionally do this. Something like `--chown <uid>:<gid>` makes much more sense to me, as it means we won't have any unexpected default behavior. @gmkurtzer I'd like to hear your thoughts here since this was your idea IIRC

src/pkg/build/assemblers/assembler_sif.go
ArangoGutierrez: ummm lint on my IDE, recommend id always should be ID.
but I agree with you

ArangoGutierrez:
3519e2d behavior
```bash
[eduardo@rhel7 singularity]$ sudo singularity build user_with_sudo.sif examples/busybox/Singularity
INFO:    Starting build...
INFO:    Adding runscript
INFO:    Creating SIF file...
INFO:    Build complete!
[eduardo@rhel7 singularity]$ ls -lh user_with_sudo.sif
-rwxr-xr-x. 1 eduardo eduardo 640K Sep 27 14:17 user_with_sudo.sif
[eduardo@rhel7 singularity]$ sudo -i
[root@rhel7 ~]# cd /home/eduardo/go/src/github.com/sylabs/singularity/
[root@rhel7 singularity]# singularity build user_is_root.sif examples/busybox/Singularity
INFO:    Starting build...
INFO:    Adding runscript
INFO:    Creating SIF file...
INFO:    Build complete!
[root@rhel7 singularity]# ls -lh user_is_root.sif
-rwxr-xr-x. 1 root root 640K Sep 27 14:18 user_is_root.sif
```

src/pkg/build/assemblers/assembler_sif.go
```diff
@@ -77,6 +79,14 @@ func createSIF(path string, definition []byte, squashfile string) (err error) {
 		return fmt.Errorf("while creating container: %s", err)
 	}
 
+	// chown the sif file to the calling user
+	if syscall.Getuid() == 0 && os.Getenv("USERNAME") != "root" {
```
jmstover: If I run `sudo` ... `$USERNAME` is set to `root`. 

Check `SUDO_USER` instead

jmstover (dismissed):

jmstover (approved):
LGTM

gmkurtzer:
I would say, if calling `sudo singularity ...`, we should automatically chown. But if calling through `sudo` with any other command, I would suggest not. 

You can check the command being called with `sudo` with the environment variable `SUDO_COMMAND`. Pattern match that for the singularity command to trigger the chown.

ArangoGutierrez:
after af84fba
```bash
[eduardo@rhel7 sylabs]$ sudo singularity build user_call_sudo.sif singularity/examples/busybox/Singularity 
INFO:    Starting build...
INFO:    Adding runscript
INFO:    Creating SIF file...
INFO:    Build complete!
[eduardo@rhel7 sylabs]$ ls -lh user_call_sudo.sif 
-rwxr-xr-x. 1 eduardo eduardo 640K Oct  2 13:27 user_call_sudo.sif
[eduardo@rhel7 sylabs]$ sudo su -
Last login: Sat Sep 29 11:21:14 -05 2018 on pts/1
[root@rhel7 ~]# singularity build user_is_root.sif /home/eduardo/go/src/github.com/sylabs/singularity/examples/busybox/Singularity 
INFO:    Starting build...
INFO:    Adding runscript
INFO:    Creating SIF file...
INFO:    Build complete!
[root@rhel7 ~]# ls -lh user_is_root.sif 
-rwxr-xr-x. 1 root root 640K Oct  2 13:28 user_is_root.sif
```

src/pkg/build/assemblers/assembler_sif.go
```diff
@@ -129,3 +136,41 @@ func (a *SIFAssembler) Assemble(b *types.Bundle, path string) (err error) {
...
+
+			uid, err = strconv.Atoi(_uid)
+			if err != nil {
+				sylog.Errorf("Error while calling strconv: %v", err)
+
+				return 0, 0, false
+			}
+			gid, err = strconv.Atoi(_gid)
+			if err != nil {
+				sylog.Errorf("Error while calling strconv : %v", err)
```
bauerm97: I would say `Warningf` since these are not "errors" which will cause singularity to exit

bauerm97: Same as the other `Errorf` spots as well

src/pkg/build/assemblers/assembler_sif.go
```diff
@@ -129,3 +136,41 @@ func (a *SIFAssembler) Assemble(b *types.Bundle, path string) (err error) {
...
+// changeOwner check the command being called with sudo with the environment
+// variable SUDO_COMMAND. Pattern match that for the singularity bin
+func changeOwner() (int, int, bool) {
+	var uid, gid int
+	var chown bool
+	var err error
+
+	r := regexp.MustCompile("(singularity)")
+	sudoCmd := os.Getenv("SUDO_COMMAND")
+	if r.MatchString(sudoCmd) {
```
bauerm97:
```
if !r.MatchString(sudoCmd) {
    return 0, 0, false
}
```

src/pkg/build/assemblers/assembler_sif.go
```diff
@@ -129,3 +136,41 @@ func (a *SIFAssembler) Assemble(b *types.Bundle, path string) (err error) {
 
 	return
 }
+
+// changeOwner check the command being called with sudo with the environment
+// variable SUDO_COMMAND. Pattern match that for the singularity bin
+func changeOwner() (int, int, bool) {
+	var uid, gid int
```
bauerm97: Can probably omit this here and just use `:=` syntax to declare these vars at appropriate times

src/pkg/build/assemblers/assembler_sif.go
```diff
@@ -129,3 +136,41 @@ func (a *SIFAssembler) Assemble(b *types.Bundle, path string) (err error) {
...
+	var chown bool
+	var err error
+
+	r := regexp.MustCompile("(singularity)")
+	sudoCmd := os.Getenv("SUDO_COMMAND")
+	if r.MatchString(sudoCmd) {
+		if syscall.Getuid() == 0 && os.Getenv("SUDO_USER") != "root" {
```
bauerm97:
```
if !(syscall.Getuid() == 0 && os.Getenv("SUDO_USER") != "root") {
    return 0, 0, false
}
```

bauerm97 (changes requested):
In general pretty close to being ready. I think we should typically prefer the design pattern of avoiding lots of nested conditionals as it's harder to read. Also we should reserve `Errorf` for when we're actually returning an `error` type. If we're catching an error but are not worried about returning the error, we should be using `Warningf` instead.

src/pkg/build/assemblers/assembler_sif.go
```diff
@@ -129,3 +136,41 @@ func (a *SIFAssembler) Assemble(b *types.Bundle, path string) (err error) {
...
+
+		return 0, 0, false
+	}
+	gid, err := strconv.Atoi(_gid)
+	if err != nil {
+		sylog.Warningf("Error while calling strconv : %v", err)
+
+		return 0, 0, false
+	}
+	chown := true
```
bauerm97: This line doesn't serve a purpose. Just `return uid, gid, true` is okay.

ArangoGutierrez: agree

src/pkg/build/assemblers/assembler_sif.go
```diff
@@ -129,3 +136,40 @@ func (a *SIFAssembler) Assemble(b *types.Bundle, path string) (err error) {
...
+// changeOwner check the command being called with sudo with the environment
+// variable SUDO_COMMAND. Pattern match that for the singularity bin
+func changeOwner() (int, int, bool) {
+	r := regexp.MustCompile("(singularity)")
+	sudoCmd := os.Getenv("SUDO_COMMAND")
+	if !r.MatchString(sudoCmd) {
+		return 0, 0, false
+	}
+
+	if syscall.Getuid() == 0 && os.Getenv("SUDO_USER") == "root" {
```
bauerm97: This logic here is incorrect.

First of all, we want to continue only if the UID == 0. Second, if `SUDO_USER` is set to *anything* we should continue.

I think the logic should look like this:
`if os.Getenv("SUDO_USER") == "" || syscall.Getuid() != 0 {...}`

This means we *only* proceed with the function *if and only if* `SUDO_USER` is *set* and `uid==0`.

ArangoGutierrez: 066de5d

src/pkg/build/assemblers/assembler_sif.go
```diff
@@ -81,6 +83,11 @@ func createSIF(path string, definition []byte, squashfile string) (err error) {
 		return fmt.Errorf("while creating container: %s", err)
 	}
 
+	// chown the sif file to the calling user
+	if uid, gid, ok := changeOwner(); ok {
+		os.Chown(path, uid, gid)
```
bauerm97: We should always be checking for errors from functions which return errors

bauerm97 (changes requested):

src/pkg/build/assemblers/assembler_sif.go
ArangoGutierrez: agree.
done in 
2d22e57

ArangoGutierrez:
PING -> @bauerm97

cclerget (approved):
Looks good !

bauerm97 (approved):

PR #2119, opened 2018-09-27
Port 2.x 30-actions.sh to 3.0

ArangoGutierrez:

Port 2.x 30-actions.sh to 3.0 

**Checkoff for all PRs:**


jscook2345 (commented):
So some of this is duplicated in other tests. The actions test itself is non super specific. I would prefer breaking this out into each action. There already exists an 'exec_test'. Perhaps create a 'shell_test', and a 'run_test' to contain those parts.

ArangoGutierrez:
@jscook2345 
> So some of this is duplicated in other tests. The actions test itself is non super specific. I would prefer breaking this out into each action.
```go 
func TestSingularityActions(t *testing.T) {}
```
is calling 3 local funcs, (non Upercased) , so they can share a single SIF for all the test.
I don't see duplicated test for actions, could you point me to the duplicated one, so I can quick fix on a commit.
```go
	// singularity run
	t.Run("run", testSingularityRun)
	// singularity exec
	t.Run("exec", testSingularityExec)
	// singularity shell
	t.Run("shell", testSingularityShell)
```

> There already exists an 'exec_test'. Perhaps create a 'shell_test', and a 'run_test' to contain those parts.

has been changed for `actions_test.go` to reflect cli tree, on which `actions.go` stores `run/shell/exec`

src/cmd/singularity/actions_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+type execOpts struct {
+	binds   []string
+	contain bool
+	home    string
+	workdir string
+	pwd     string
+}
+
+func imageExec(opts execOpts, imagePath string, command []string) ([]byte, error) {
```
bauerm97: What is the purpose of this function if it never gets called?

src/cmd/singularity/actions_test.go
```diff
@@ -0,0 +1,185 @@
...
+		{cmdPath, "NoCommand", []string{"run", imagePath}, 0},
+		{cmdPath, "true", []string{"run", imagePath, "true"}, 0},
+		{cmdPath, "false", []string{"run", imagePath, "false"}, 1},
+		// Testing run command properly hands arguments
+		{"sh", "arguments", []string{"-c", fmt.Sprintf("singularity run %s foo | grep foo", imagePath)}, 0},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, test.WithoutPrivilege(func(t *testing.T) {
+			cmd := exec.Command(tt.binName, tt.argv...)
```
bauerm97: If we've created a function to run an image, then why are we not using it?

bauerm97 (changes requested):

src/cmd/singularity/actions_test.go
ArangoGutierrez: that func is used by `build_test.go` and `docker_test.go` , not in this file. as we need to catch the exit code

src/cmd/singularity/actions_test.go
ArangoGutierrez: `imageExec()` does not catch and return the exit code. should we modify that func to do it. or re allocate `imageExec()` to a diff test file

ArangoGutierrez:
@bauerm97  func `imageExec()` has been updated to be general used in all test.

bauerm97 (approved):

PR #2122, opened 2018-09-28
Add inspect tests

### LINKED ISSUE #2121: Add basic `singularity inspect` test
Add basic `singularity inspect` test

jscook2345:

Add a basic `singularity inspect` test

- Fixes #2121 

**Checkoff for all PRs:**

- [X] I have read the [Guidelines for Contributing](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- [X] I have tested this PR locally with a `make testall`
- [X] This PR is against the project's master branch

### DIFF
--- a/src/cmd/singularity/inspect_test.go
+++ b/src/cmd/singularity/inspect_test.go
@@ -1,0 +1,14 @@
+// Copyright (c) 2018, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package main
+
+import (
+	"testing"
+)
+
+func TestInspect(t *testing.T) {
+	t.Skip("Not yet implemented")
+}
--- a/src/cmd/singularity/inspect_test.go
+++ b/src/cmd/singularity/inspect_test.go
@@ -6,9 +6,30 @@
 package main
 
 import (
+	"os/exec"
 	"testing"
+
+	"github.com/sylabs/singularity/src/pkg/test"
 )
 
+func inspectImage(image string) ([]byte, error) {
+	args := []string{"inspect", image}
+	cmd := exec.Command(cmdPath, args...)
+
+	return cmd.CombinedOutput()
+}
+
 func TestInspect(t *testing.T) {
-	t.Skip("Not yet implemented")
+	var definition = "../../../examples/busybox/Singularity"
+	var image = "Default.sif"
+
+	if b, err := buildImage(definition, image); err != nil {
+		t.Fatalf("Error building an image from a definition: %v. Output follows.\n%s", err, string(b))
+	}
+
+	t.Run("InspectImage", test.WithoutPrivilege(func(t *testing.T) {
+		if b, err := inspectImage(image); err != nil {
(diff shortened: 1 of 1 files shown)


src/cmd/singularity/inspect_test.go
```diff
@@ -0,0 +1,35 @@
...
+	cmd := exec.Command(cmdPath, args...)
+
+	return cmd.CombinedOutput()
+}
+
+func TestInspect(t *testing.T) {
+	var definition = "../../../examples/busybox/Singularity"
+	var image = "Default.sif"
+
+	if b, err := buildImage(definition, image); err != nil {
```
ArangoGutierrez: I think we could add a labels section to that busybox definition
```
%labels

    Maintainer sylabs

    Version v3.0.0-beta
```

src/cmd/singularity/inspect_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"os/exec"
+	"testing"
+
+	"github.com/sylabs/singularity/src/pkg/test"
+)
+
+func inspectImage(image string) ([]byte, error) {
+	args := []string{"inspect", image}
```
ArangoGutierrez, line 16: and here test at least inspect over 2 sections
```
[eduardo@rhel7 singularity]$ singularity inspect --runscript labels.sif

#!/bin/sh

    echo "Running command: $*"
    exec "$@"

[eduardo@rhel7 singularity]$ singularity inspect --deffile labels.sif

mirrorurl: https://www.busybox.net/downloads/binaries/1.26.1-defconfig-multiarch/busybox-x86_64
bootstrap: busybox

%labels
        Version v3.0.0-beta
        Maintainer sylabs

%runscript
    echo "Running command: $*"
    exec "$@"

%post

```

ArangoGutierrez (changes requested):

src/cmd/singularity/inspect_test.go
GodloveD: well, I think there should already be some labels there by default right?  Like a timestamp of when the container was created.  So I don't know if we need to add any new ones.

GodloveD (changes requested):
So it looks like this is just checking the inspect command with no flags and then checking to see if the function returns error right?  Could it please do the following too?  

1. Run a new test with each of the inspect flags.  Maybe make the flags a string argument to the test function and then run through all of them.

2. Check the text output of the inspect command against some expectation.  I don't think this has to be exhaustive, like an exact match or something.   But the inspect output should contain keywords that can always be found.

src/cmd/singularity/inspect_test.go
bauerm97, line 26: Looks like this function doesn't exist

ArangoGutierrez: it was rm on https://github.com/sylabs/singularity/pull/2128 , to fix #2127
we should all use  `imageBuild()`
```go
	opts := buildOpts{
		force:    true,
		sandbox:  false,
		writable: false,
	}
	if b, err := imageBuild(opts, imagePath, definition); err != nil {
		t.Log(string(b))
		t.Fatalf("unexpected failure: %v", err)
	}
	imageVerify(t, imagePath, true)
```

jscook2345:
I'm going to delete this PR since it was pre-directory reorg. The code is simple enough to add back in later.

PR #2050, opened 2018-09-21
3.0: Make sure vendors are built before check, and show dep ensure errors

DrDaveD:
**Description of the Pull Request (PR):**

This PR makes $(BUILDDIR)/vendors-done a dependency of check, because otherwise go vet fails if `make check` is done before `make`.  Also show errors from 'dep ensure' by removing the redirect of stderr to /dev/null.

**This fixes or addresses the following GitHub issues:**

None

**Checkoff for all PRs:**

- [x] I have read the [Guidelines for Contributing](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- [ ] I have added changes to the [CHANGELOG](https://github.com/sylabs/singularity/blob/development/CHANGELOG.md) and and documentation updates to the [user](https://www.github.com/sylabs/singularity-userdocs) and [administrator](https://www.github.com/sylabs/singularity-admindocs) documentation bases.
- [ ] I have tested this PR locally with a `make testall`
- [x] This PR is against the project's master branch
- [x] I have added myself as a contributor to the [contributors's file](https://github.com/sylabs/singularity/blob/master/CONTRIBUTORS.md)
- [x] This PR is ready for review and/or merge

Attn: @singularity-maintainers

Issue #2120, opened 2018-09-28
version command and --version flag are not working in tag


GodloveD:
## Version of Singularity:

Installed from the [3.0 beta tarball.](https://github.com/sylabs/singularity/archive/v3.0.0-beta.1.tar.gz)

### Expected behavior

`singularity --version` and/or `singularity version` should give you the version number.

### Actual behavior

```
$ singularity version

$ singularity --version
Error: unknown flag: --version
Usage:
  singularity [global options...] <command>

Available Commands:
  build       Build a new Singularity container
  capability  Manage Linux capabilities on containers
  exec        Execute a command within container
  inspect     Display metadata for container if available
  instance    Manage containers running in the background
  keys        Manage OpenPGP key stores
  pull        Pull a container from a URI
  push        Push a container to a Library URI
  run         Launch a runscript within container
  run-help    Display help for container if available
  search      Search the library
  shell       Run a Bourne shell within container
  sign        Attach cryptographic signatures to container
  test        Run defined tests for this particular container
  verify      Verify cryptographic signatures on container
  version     Show application version
```

bauerm97:
So, it looks like the issue is that the source code tarballs that GitHub includes in the release don't include the `.git` directory, meaning there is no git repository present. I think these tarballs are mostly for archival purposes, so I don't think this is an issue. When you do `git checkout v3.0.0-beta.1`, this is the behavior that happens and I'm okay with that:

```
mibauer@bauer-desktop:~$ singularity version
v3.0.0-beta.1
mibauer@bauer-desktop:~$ singularity --version
singularity version v3.0.0-beta.1
```

GodloveD:
Well it's an issue because we are going to have a bunch of users installing from the tarballs in git b/c that is what we've encouraged users to do forever and then they will be reporting errors and we will have no way to tell what version they are running.

jscook2345:
Most apps (in C) have a header file with the version hard coded into it. Just before tagging you usually would run a script to bump the version (in various areas) to the correct one.

PR #2152, opened 2018-10-02
Enable 'make check' to be run before 'make'

DrDaveD:

This PR adds $(libruntime) as a dependency of check.  This makes 'go vet' work even if somebody has not done a 'make' first.  This seems to be a new requirement, because 'make check' worked after #2050 was integrated.

None

**Checkoff for all PRs:**


DrDaveD:
I found that check also needed a dependency of $(BUILDDIR)/config.h, and added that.

bauerm97:
So, I understand the motivation behind this PR, but I wonder if a better way to solve this problem is to fix the fact that `src/pkg/buildcfg` has nothing defined before being auto-generated. Perhaps a better solution would be:

1. Change the `buildcfg` values from `const` -> `var`
2. Add a new file `config_stub.go` which has stub declarations of each value
3. Change the `go generate` command to output in the form of `$(VAR_NAME) = $(VAR_VALUE)` instead of `const $(VAR_NAME) = $(VAR_VALUE)`. 
4. The `go generate` command could *also* generate the stub file as well, ensuring that if we make a change to the `config.h` header, that will be reflected in the commit as well.

This can also solve #2120, where the tarball generated by GitHub when tagging a release does not include the tag. I can take a stab at fixing it this way, and we can merge this as an intermediate solution

DrDaveD:
@bauerm97 If I understand correctly, your proposal should avoid the need to generate config.h first, but I think it doesn't affect the need for $(libruntime).  I agree it would be good to merge this PR first then you come with another one for changing the way buildcfg works.  I would have done something simpler for #2120 (by just putting a default value into mconfig if git describe doesn't find anything) but maybe your updated buildcfg idea is better.

bauerm97:
Ahh, yes maybe you are correct. Does `go vet` compile everything?

DrDaveD:
`go vet` doesn't compile everything but it does parse all the code including looking for type information.   It's a lot faster than a full compile, especially when building an rpm which redownloads all the needed go modules.

On the other hand, I am now trying to reproduce the problem that prompted me to create this PR, and it's not happening.  I don't know what changed.  Let's let this PR sit for awhile to see if the problem comes up again.

bauerm97:
@DrDaveD Try running these commands: 

```
git clean -xfd
./mconfig
make -C builddir -j6 testall
```

See if that reproduces? There will be a problem if `src/pkg/buildcfg/config.go` has not been generated yet, so by cleaning and then running testall with many jobs, it's probably more likely to trigger

DrDaveD:
Indeed, removing src/pkg/buildcfg/config.go showed the problems.  In fact it showed that the dependency should be $(go_OBJ) instead of the other three, and I have updated the PR for that.

I had already been removing builddir and vendor in my own build script, but I am afraid to routinely use git clean -xfd there because I sometimes leave around temporary files I don't want deleted.  I added removing of config.go.

I don't want to use testall because that requires compiling.  I just do make check, then dist/rpm/rpmbuild, and sometimes make test after that.

DrDaveD:
Travis failed here on something unrelated.  It worked on the [same commit in my personal repository](https://travis-ci.org/DrDaveD/singularity/builds/437246868?utm_source=github_status&utm_medium=notification).    Closing and reopening to trigger a travis rebuild.

DrDaveD:
Rebased on master after #2179 was merged to fix travis.

bauerm97 (approved):
This looks like a good first step towards completely fixing this problem. Approving

yhcote (approved):
Looks good to me!

PR #1620, opened 2018-06-05
Fix config parser malloc size calculation that resulted in writing off the end of the space

DrDaveD:
**Description of the Pull Request (PR):**

This PR fixes a miscalculation in malloc size in the config parser new_hash_entry() function.  I found this because I was using the valgrind malloc checker with these parameters:
```
valgrind --track-origins=yes --keep-stacktraces=alloc-and-free --tool=memcheck --trace-children=yes
```
and it had the following output
```
==7653== Invalid write of size 8
==7653==    at 0x4E3C5A6: new_hash_entry (config_parser.c:60)
==7653==    by 0x4E3C5A6: add_entry (config_parser.c:99)
==7653==    by 0x4E3C5A6: singularity_config_parse (config_parser.c:185)
==7653==    by 0x4E3C8E5: singularity_config_init (config_parser.c:212)
==7653==    by 0x10B6E4: main (action.c:70)
==7653==  Address 0x5639d50 is 512 bytes inside a block of size 513 alloc'd
==7653==    at 0x4C29BC3: malloc (vg_replace_malloc.c:299)
==7653==    by 0x4E3C586: new_hash_entry (config_parser.c:56)
==7653==    by 0x4E3C586: add_entry (config_parser.c:99)
==7653==    by 0x4E3C586: singularity_config_parse (config_parser.c:185)
==7653==    by 0x4E3C8E5: singularity_config_init (config_parser.c:212)
==7653==    by 0x10B6E4: main (action.c:70)
```

**This fixes or addresses the following GitHub issues:**

- None

**Checkoff for all PRs:**

- [x] I have read the [Guidelines for Contributing](https://github.com/singularityware/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- [ ] I have added changes to the [CHANGELOG](https://github.com/singularityware/singularity/blob/development/CHANGELOG.md) and and documentation updates to the [singularityware](https://www.github.com/singularityware/singularityware.github.io) documentation base.
- [x] I have tested this PR locally with a `make test`
- [x] This PR is NOT against the project's master branch
- [x] I have added myself as a contributor to the [contributors's file](https://github.com/singularityware/singularity/blob/master/CONTRIBUTORS.md)
- [x] This PR is ready for review and/or merge

Attn: @singularityware-admin

PR #2153, opened 2018-10-02
Tidy up repo documentation before v3.0.0 release

bauerm97:

Tidy up some of the documentation found in the repository prior to tagging `v3.0.0`


CHANGELOG.md
```diff
@@ -0,0 +1,66 @@
+# Changes Since v2.6.0
```
GodloveD, line 12: I understand that you want to make this a new changelog for the 3.x series, and I think that's fine.  But I think this should be in the same format as the changelog that is currently in 2.x.  When I say the same format I just mean the same markdown format with the same title and headers and stuff so the changelogs at least look like one another.

CHANGELOG.md
```diff
@@ -0,0 +1,66 @@
...
+
```
GodloveD, line 13: Should we state somewhere in the changelog that we are no longer supporting tar?

CONTRIBUTING.md
```diff
@@ -25,49 +25,46 @@ all your interactions with the project members and users.
...
+5. For any new functionality, please write appropriate go tests that will run
+   as part of the Continuous Integration (Circle CI and Travis) systems.
+6. Make sure that the project's default copyright and header have been included 
+   in any new source files.
+7. Make sure you have locally tested using `make test` and that all tests succeed
    before submitting the PR.
-8. Is the code human understandable? This can be accomplished via a clear code
+8. To conform to the Golang standards and idioms, make sure you have done the 
+   following:
+    - Ran `go fmt ./...` to format all `.go` files. We use `go1.11`'s
```
GodloveD: "Have run" instead of "ran"

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -12,11 +12,11 @@ fixing or enhancing.
 
 **Checkoff for all PRs:**
 
-- [ ] I have read the [Guidelines for Contributing](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
-- [ ] I have added changes to the [CHANGELOG](https://github.com/sylabs/singularity/blob/development/CHANGELOG.md) and and documentation updates to the [user](https://www.github.com/sylabs/singularity-userdocs) and [administrator](https://www.github.com/sylabs/singularity-admindocs) documentation bases.
+- [ ] I have read the [Guidelines for Contributing](../CONTRIBUTING.md), and this PR conforms to the stated requirements.
```
GodloveD: I would add to this list
```
- [ ] I have created the appropriate regression tests for any bugs that I have fixed and unit tests for any functions I have written.
```

CONTRIBUTING.md
```diff
@@ -25,49 +25,46 @@ all your interactions with the project members and users.
...
 
-### Process
-1. Essential bug fix PRs should be sent to both master and devopment branches.
-2. Small bug fix and feature enhancement PRs should be sent to development only.
-3. Follow the existing code style precedent. This does not need to be strictly
-   defined as there are many thousands of lines of examples. Note the lack
-   of tabs anywhere in the project, parentheses and spacing, curly bracket
-   locations, source code layout, variable scoping, etc. and follow the
-   project's standards.
+1. Essential bug fix PRs should be sent to both master and release branches.
```
GodloveD, line 28: It's nice if people do this, but I think it's actually fine to just send them to the release branch and then we will cherry pick them into master.  Whatever.

CHANGELOG.md
GodloveD, line 12: Above this line should be the explanation for what the CHANGELOG.md is as it appears in 2.x.  Maybe also a new bit about how this changelog is just for the 3.x series.

This line should read
```
## [v3.0.0](with link to release)
```

CHANGELOG.md
```diff
@@ -0,0 +1,66 @@
...
+    * `library://` - Supports the [Sylabs.io Cloud Library](https://cloud.sylabs.io/library)
+    * `docker-daemon:` - Supports images managed by the locally running docker daemon
+    * `docker-archive:` - Supports archived docker images
+    * `oci:` - Supports oci images
+    * `oci-archive:` - Supports archived oci images
+  * Handling of `docker` & `oci` URIs/images now utilizes [containers/image](https://github.com/containers/image) to parse and convert those image types in a supported way
+  * Replaced `singularity instance.*` command group with `singularity instance *`
+  * The command `singularity help` now only provides help regarding the usage of the `singularity` command. To display an image's `help` message, use `singularity run-help <image path>` instead
+ 
+## Removed Deprecated Commands
```
GodloveD, line 25: Three hashes instead of two on these headers for consistency with the style we've previously adopted.

CHANGELOG.md
```diff
@@ -0,0 +1,66 @@
+# Changes Since v2.6.0
+
+  * Singularity is now written primarily in Go to bring better integratiion with the existing container ecosystem
```
GodloveD: dashes (`-`) instead of bullets (`*`) for consistency with previous readme style.

CHANGELOG.md
```diff
@@ -0,0 +1,66 @@
...
+  * The command `singularity help` now only provides help regarding the usage of the `singularity` command. To display an image's `help` message, use `singularity run-help <image path>` instead
+ 
+## Removed Deprecated Commands
+  * Removed deprecated `singularity image.*` command group
+  * Removed deprecated `singularity create` command
+  * Removed deprecated `singularity bootstrap` command
+  * Removed deprecated `singularity mount` command
+
+## New Commands
+  * Added `singularity run-help <image path>` command to output an image's `help` message
```
GodloveD: Also added `push`.  Should maybe double check if there are others we have added.

CHANGELOG.md
```diff
@@ -0,0 +1,66 @@
...
+    * `--overlay <string>` now specifies a list of `ext3`/`sandbox` images which are set as the containers overlay [applicable to: all image formats] 
+
+## Build Command Changes:
+  * All images are now built as [Singularity Image Format (SIF)](https://www.sylabs.io/2018/03/sif-containing-your-containers/) images by default
+  * When building to a path that already exists, `singularity build` will now prompt the user if they wish to overwrite the file existing at the specified location
+  * The `-F|--force` flag now overrides the interactive prompt and will always attempt to overwrite the file existing at the specified location
+  * The `-u|--update` flag has been added to support the workflow of running a definition file on top of an existing container [implies `--sandbox`, only supports `sandbox` image types]
+  * The `singularity build` command now supports the following flags for integration with the [Sylabs.io Cloud Library](https://cloud.sylabs.io/library):
+    * `-r|--remote`: Build the image remotely on the Sylabs Remote Builder (currently unavailable)
+    * `-d|--detached`: Detach from the `stdout` of the remote build [requires `--remote`]
```
GodloveD: We should change this so that it simple assumes remote instead of requiring that a user passes both.

CONTRIBUTING.md
```diff
@@ -25,49 +25,46 @@ all your interactions with the project members and users.
...
+   "golint" tools for proper formatting.
 4. Ensure any install or build dependencies are removed before doing a build
    to test your PR locally.
-5. For any new functionality, please write a test to be added to Continuous
-   Integration (Travis) to test it (tests can be found in the `tests/`
-   directory).
-6. The project's default copyright and header have been included in any new
-   source files.
-7. Make sure you have implemented a local `make test` and all tests succeed
+5. For any new functionality, please write appropriate go tests that will run
```
GodloveD, line 35: And for any bug fixes write appropriate regression tests too.

CONTRIBUTING.md
```diff
@@ -25,49 +25,46 @@ all your interactions with the project members and users.
...
-
+10. The pull request will be reviewed by others, and finally merged when all
+   requirements are met.
+11. The `CHANGELOG.md` must be updated for any of the following changes:
+    - Renamed commands
+    - Deprecated / removed commands
+    - Changed defaults / behaviors
+    - Backwards incompatible changes
+    - Migration guidance
+    - New features / functionalities (*e.g. a new flag for a command*)
```
GodloveD: I liked having bug fixes in the CHANGELOG.md as well, but I'll leave that up to whomever.

GodloveD (changes requested):

CONTRIBUTING.md
bauerm97: The reason we don't want bug fixes included in the CHANGELOG is because we want to make an effort to reign in the scope of the CHANGELOG. We want the CHANGELOG to be mostly about functionality, *i.e. intended behavior*, of the product. I think it's okay to leave bug fixes as commit messages

CHANGELOG.md
bauerm97: Fair enough!

CONTRIBUTING.md
bauerm97: Good catch

CONTRIBUTING.md
bauerm97: So, reading through the old changelog and this is the type of message we want to avoid:

`Fix an error in malloc allocation #1620`

This is something that should be mentioned in a commit message rather than a changelog (this was listed in the old CHANGELOG under a Bug Fixes header). There might be some bug fixes (where there is really broken functionality) that necessitates an entry into the CHANGELOG but I think for most cases (like the example above) it's not useful

CHANGELOG.md
bauerm97: Yes, I have not included this line in yet.

CHANGELOG.md
bauerm97: Well, that's not the behavior of this flag right now and we most likely will not be changing that for the time being so it should stay this way for now

CHANGELOG.md
bauerm97: I believe `push` was a command in `2.x` for pushing to `shub`?

bauerm97: Nevermind, I was wrong. Good catch!

CHANGELOG.md
GodloveD: Right.  I mean we should change behaviour of the flag.

README.md
```diff
@@ -23,158 +10,31 @@ called development-3.0)._
...
-# Cite as:
+Singularity is an open source container platform designed to be simple, fast, and secure. Singularity is optimized for EPC and HPC workloads, allowing untrusted users to run untrusted containers in a trusted way.
+
+Check out [who is using Singularity](https://www.sylabs.io/as-seen-on/) and some [use cases of Singularity](https://www.sylabs.io/category/how-tos/) on our website.
+
+## Getting Started with Singularity
+
+To install Singularity from source, see the [installation instructions](INSTALL.md). For other installation options, see [our website](https://www.sylabs.io/singularity-pro/).
+
+For system administrators, see the [administrator documentation](https://www.sylabs.io/guides/2.6/admin-guide/).
```
GodloveD: @jmstover has a `latest` URL or something so that we do not have to continuously update this link.  Maybe he can share it.

README.md
```diff
@@ -23,158 +10,31 @@ called development-3.0)._
...
+
+For users, see the [user documentation](https://www.sylabs.io/guides/2.6/user-guide/).
```
GodloveD: Ditto on linking to latest documentation.

GodloveD (approved):

PR #2157, opened 2018-10-03
Admin Set Mksquashfs Path

### LINKED ISSUE #2144: 3.0: Admin set path for mksquashfs in singularity.conf
## Version of Singularity:
All 3.x

### Expected behavior
Administrator can specify the path to the mksquashfs binary  (in 2.x was done at compile time).

### Actual behavior
The location of `mksquashfs` is searched for with `exec.LookPath()`. If `mksquashfs` is *not* installed in a standard system location, an image build will fail.

```
ERROR:    mksquashfs is not installed on this system
FATAL:    While performing build: exec: "mksquashfs": executable file not found in $PATH
```

Move mksquashfs to a non-standard location and try to build an image.

ikaneshiro:
This PR adds a field to `singularity.conf` to allow an admin to specify the location of `mksquashfs` for the case that it is installed in a nonstandard location.

Fixes #2144 by looking for a custom `mksquashfs` location during image building and defaulting to looking at standard locations if not specified.


src/pkg/build/assemblers/assembler_sif.go
```diff
@@ -84,6 +88,33 @@ func createSIF(path string, definition []byte, squashfile string) (err error) {
...
+
+		if _, err := os.Stat(mksquashfs); os.IsNotExist(err) {
+			return "", fmt.Errorf("mksquashfs cannot be found at custom location: %v", c.MksquashfsPath)
+		}
+
+		return mksquashfs, nil
+	}
+
+	// look for mksquashfs in standard locations
+	mksquashfs, err := exec.LookPath("mksquashfs")
```
bauerm97: `return exec.LookPath("mksquashfs")`

bauerm97 (approved):

PR #2161, opened 2018-10-03
Clean build environment

### LINKED ISSUE #2103: cleaning/santizing the user's environment at build time
## Version of Singularity:

v3.0.0-beta

When doing a build, we need to clean the user's environment.
call  https://github.com/sylabs/singularity/blob/master/src/pkg/util/env/clean.go#L30
with `cleanEnv == true` 
when running a build

ikaneshiro:
Uses `SetContainerEnv()` to clean the environment that the build engine runs in. This requires the use of a `generator` to configure the `ociConfig`, so that was added.

Fixes #2103


src/pkg/build/build.go
```diff
@@ -290,22 +291,28 @@ func (b *Build) runBuildEngine() error {
...
 	sylog.Debugf("Starting build engine")
+
+	ociConfig := &oci.Config{}
+	generator := generate.Generator{Config: &ociConfig.Spec}
+
+	// Copy and cache environment
+	environment := os.Environ()
+
+	// Clean environment
+	env.SetContainerEnv(&generator, environment, true, "")
```
ArangoGutierrez: We need to move this from here, in order to respect `%setup` section, which should be for stuff that happens on the host before build.
test it with 
```
BootStrap: busybox
MirrorURL: https://www.busybox.net/downloads/binaries/1.26.1-defconfig-multiarch/busybox-x86_64

%setup
    env
    env > /tmp/pre_build_env

%post
    echo $PWD
    env > /tmp/build_env

%runscript
    echo "Running command: $*"
    exec "$@"
```

ArangoGutierrez (changes requested):

src/runtime/engines/imgbuild/process.go
```diff
@@ -11,14 +11,20 @@ import (
...
 
+	"github.com/opencontainers/runtime-tools/generate"
 	"github.com/sylabs/singularity/src/pkg/sylog"
+	"github.com/sylabs/singularity/src/pkg/util/env"
 )
 
 // StartProcess runs the %post script
 func (e *EngineOperations) StartProcess(masterConn net.Conn) error {
 
+	// clean environment that post script is run in
```
ArangoGutierrez: Niptick 
```go
// clean environment on which  %post section script is run in
```

ArangoGutierrez (approved):

ArangoGutierrez (approved):

bauerm97 (approved):

PR #2162, opened 2018-10-03
Add warning if more than 50 bind mounts required for underlay

DrDaveD:

This PR adds a warning message if more than 50 bind mounts were required to support a requested bind mount while using the underlay feature.  This is a feature included in my 2.6 underlay implementation.  I think it is important for users to know when their requests have a potential impact on system resources.  I'm not sure what will happen if hundreds of containers on the same system each use hundreds of bind mounts.

None

**Checkoff for all PRs:**


DrDaveD:
@cclerget please review

GodloveD:
How is this affected by @cclerget's proposed #1694 ?

DrDaveD:
@GodloveD I don't see any connection.

src/pkg/util/fs/layout/layer/underlay/underlay.go
```diff
@@ -136,7 +136,8 @@ func (u *Underlay) createLayer(rootFsPath string, system *mount.System) error {
 	return u.session.Update()
 }
 
-func (u *Underlay) duplicateDir(dir string, system *mount.System) error {
+func (u *Underlay) duplicateDir(dir string, system *mount.System, existingPath string) error {
```
cclerget, line 139: If I understand correctly, there is no need to add an argument here, instead of doing `if binds > 50 && existingPath != ""` you could have `if binds > 50 && dir != "/"`

cclerget (changes requested):

src/pkg/util/fs/layout/layer/underlay/underlay.go
DrDaveD: It's true you could check dir != "/" but the more important reason for the argument is to include it in the warning message.

cclerget: And `dir` is not enough for the warning message ?

DrDaveD: I don't think so because it helps people to know which bind caused the duplication, not just the name of the directory that was duplicated.

cclerget (approved):

PR #2175, opened 2018-10-03
Allow colon in image names

DrDaveD:

This PR allows colons to be in image names without being in transport:ref format.  Without this, if an image name or sandbox path has a colon, singularity dies with an error message "Unsupported transport type".  The PR also moves the "IsValidURI" function from a place where it makes little sense to be, build/conveyerPacker.go, to util/uri/uri.go.

None

**Checkoff for all PRs:**


src/pkg/build/build.go
```diff
@@ -357,7 +358,7 @@ func getcp(def types.Definition, libraryURL, authToken string) (ConveyorPacker,
 
 // makeDef gets a definition object from a spec
 func makeDef(spec string) (types.Definition, error) {
-	if ok, err := IsValidURI(spec); ok && err == nil {
+	if ok, err := uri.IsValidURI(spec); ok && err == nil {
```
ArangoGutierrez, line 361: If you are going to propose a `uri` pkg, `IsValidURI()` should be renamed `IsValid()`

jscook2345:
Looking at your changes I can see the reasoning behind this request. After reviewing it I think it looks OK.

Note, the colon character has to be escaped in the shell when used and thus leads to a few issues with usage. It also is an illegal character if you're sharing files with a Windows-based system.

ArangoGutierrez (dismissed):

src/pkg/build/build.go
DrDaveD: @ArangoGutierrez There already was a uri package with SplitURI and NameFromURI.  Must those change to drop the URI too?  I was just following the conventions I saw.

DrDaveD:
@jscook2345 Colons do not need to be escaped from any Linux shells that I am familiar with.

src/pkg/build/build.go
ArangoGutierrez: I will believe so, personally I don't like `pkg.FuncPkg` redundancy , I prefer `pkg.Func` .
But @bauerm97 is the one to call func names.

DrDaveD: In particular what would NameFromURI change to?  uri.NameFrom?   That's pretty awkward.

ArangoGutierrez: we accept suggestions ;P 
I guess `uri.GetName()` sounds nice.....

DrDaveD: Ok that sounds reasonable.  That's the name I thought of after I posted too.  

I think this PR is big enough however.  I suggest that we merge this as-is and then I'll submit another PR with the uri package name changes.

ArangoGutierrez: that sounds reasonable.
I'll leave @bauerm97 proceed with this PR

DrDaveD:
I rebased on the current master.  @bauerm97, can we please proceed with this PR?

src/pkg/build/build.go
yhcote: I am inclined to merge this one. @DrDaveD : Do you have a list of names you wanna refactor in this package?

DrDaveD: I'll rename all of them in a separate PR: NameFromURI, SplitURI, and IsValidURI

yhcote: Ok Sounds good

yhcote (approved):
Looks good to me.

Issue #2176, opened 2018-10-04
#2140 seems to have broken apt on debian somehow


GodloveD:
## Version of Singularity:

Current master.  But seems to have broken on  2c31aaaf166262b738cb350fe7fc390552845f4a.

### Expected behavior

One should be able to able to build a container with the following def file
```
Bootstrap: library
From: debian:9

%post
    apt-get -y update
```
### Actual behavior

Before 2c31aaaf166262b738cb350fe7fc390552845f4a, this worked just fine.  But after 2c31aaaf166262b738cb350fe7fc390552845f4a one gets the following results. 

```
$ sudo singularity build debian.sif debian.def 
WARNING: Authentication token file not found : Only pulls of public images will succeed
INFO:    Starting build...
 40.36 MiB / 40.36 MiB [=========================================] 100.00% 3.32 MiB/s 12s
INFO:    Running post scriptlet
+ apt-get -y update
Ign:1 http://cdn-fastly.deb.debian.org/debian stretch InRelease                    
Get:2 http://cdn-fastly.deb.debian.org/debian stretch-updates InRelease [91.0 kB]  
Get:4 http://security.debian.org/debian-security stretch/updates InRelease [94.3 kB]
Get:3 http://cdn-fastly.deb.debian.org/debian stretch Release [118 kB]               
Get:5 http://cdn-fastly.deb.debian.org/debian stretch Release.gpg [2434 B]                  
Get:6 http://cdn-fastly.deb.debian.org/debian stretch-updates/main amd64 Packages [5148 B]
Get:7 http://cdn-fastly.deb.debian.org/debian stretch/main amd64 Packages [7099 kB]        
Get:8 http://security.debian.org/debian-security stretch/updates/main amd64 Packages [447 kB]
Fetched 7856 kB in 4s (1582 kB/s)                               
Reading package lists... Done
E: Unable to change to (unreachable)/home/david/sandbox/ - chdir (2: No such file or directory)
FATAL:   post proc: exit status 100
FATAL:   While performing build: while running engine: exit status 255
```

### Steps to reproduce behavior

Debug output doesn't provide many clues.
```
[...snip...]
DEBUG   [U=0,P=86534]      CreateContainer()             Chdir into / to avoid errors
DEBUG   [U=0,P=86542]      set_parent_death_signal()                 Set parent death signal to 9
DEBUG   [U=0,P=86542]      prepare_scontainer_stage()                Entering in scontainer stage 2
[...]

PR #2158, opened 2018-10-03
Development 3.0: Instance fixes

cclerget:
**Description of the Pull Request (PR):**

This PR contains :
- fix for `--boot` regression
- replace SIGTERM with SIGINT by default to stop instance
- set correct log file owner when using target UID (some daemon could get permission denied)
- move some instance message as Infof to Verbosef
- add URI handler
- fix regression with RPC/stage 2 fork by adding CLONE_FS

**Checkoff for all PRs:**

- [x] I have read the [Guidelines for Contributing](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- [ ] I have added changes to the [CHANGELOG](https://github.com/sylabs/singularity/blob/development/CHANGELOG.md) and and documentation updates to the [user](https://www.github.com/sylabs/singularity-userdocs) and [administrator](https://www.github.com/sylabs/singularity-admindocs) documentation bases.
- [x] I have tested this PR locally with a `make testall`
- [x] This PR is against the project's master branch
- [ ] I have added myself as a contributor to the [contributors's file](https://github.com/sylabs/singularity/blob/master/CONTRIBUTORS.md)
- [x] This PR is ready for review and/or merge

Attn: @singularity-maintainers

PR #2178, opened 2018-10-04
Sandbox pseudo device copy fix

### LINKED ISSUE #2031: 3.0: copy.Copy() causing issues with sandbox build
## Version of Singularity:
v3.0.0-alpha.1-550-g73e3ad9

### Expected behavior
A sandbox image is built... 

```
$ sudo singularity build --sandbox nginx nginx.def
  [ ... ]
INFO:    Build complete!
```

### Actual behavior
Build hangs, and continuously copies `/dev/full` into the sandbox

```
$ ls -lh nginx/dev/full 
-rw-rw-rw- 1 root root 115G Sep 19 21:22 nginx/dev/full
```

Using below Def file, build a sandbox image cross-device, so the `copy.Copy()` call is triggered.

```
---- Start Cut ----
BootStrap: debootstrap
OSVersion: stable
MirrorURL: http://ftp.us.debian.org/debian/

%runscript

%post
    apt-get update
    apt-get -y install vim procps netcat iptables net-tools iputils-ping nginx systemd systemd-sysv
    apt-get clean
    systemctl enable nginx
---- End Cut ----
```

ikaneshiro:
This PR uses the 'mv' command instead of a combination of `rename()` and `copy()` in order to allow for building sandboxes across devices and to prevent the issue of attempting to read from pseudo devices such as `/dev/full` while moving the sandbox cross device.

This also removes the `github.com/otiai10/copy` dependency.

Fixes #2031


ArangoGutierrez (commented):
I am getting this ERROR while trying to build the definition file example on #2031 
```bash
Fetched 5396 kB in 3s (1469 kB/s)         
Reading package lists... Done
E: Unable to change to (unreachable)/tmp/ - chdir (2: No such file or directory)
FATAL:   post proc: exit status 100
```

jmstover:
@ArangoGutierrez See #2176 ... and #2158 for the PR with the fix for that issue.

src/pkg/build/assemblers/assembler_sandbox.go
```diff
@@ -28,16 +29,9 @@ func (a *SandboxAssembler) Assemble(b *types.Bundle, path string) (err error) {
...
-			return err
-		}
-
-		if err := os.RemoveAll(b.Rootfs()); err != nil {
-			sylog.Errorf("Unable to remove Bundle directory: %s", err)
-			return err
-		}
+	cmd := exec.Command("mv", b.Rootfs(), path)
+	if err = cmd.Run(); err != nil {
+		return fmt.Errorf("Sandbox Assemble Failed: %s", err)
```
jmstover, line 34: Use `fmt` ... or `sylog` that you're already importing?

bauerm97: Here is **returning** `fmt.Errorf()`, meaning we're using the format package to return an `error` type which contains formatted text. We can't return `sylog.Errorf()`, which is a function which prints output and does not return anything.

jmstover: :+1:

jmstover (approved):
LGTM

bauerm97 (approved):

PR #2184, opened 2018-10-05
 Addition of config profile, overall cleanup to the Makeit build system

yhcote:
This PR improves the build system in different ways but mostly by adding go configuration profiles and unifying how go tools are called on the project.


cclerget (approved):
Looks good !

PR #2185, opened 2018-10-05
Makeit: Allow configurable build/install paths.

yhcote:
This PR allows many build and install paths to be configured via mconfig + mlocal/frags/paths.stub.

This is useful mainly for RPM and other packaging.


cclerget (approved):
Looks good !

PR #2200, opened 2018-10-06
Development 3.0: Vendor commit

cclerget:

Since vendor commit will be very big, that will be done once approved in order to clearly identify current modification in build system.

**Checkoff for all PRs:**


mlocal/frags/Makefile.stub
```diff
@@ -145,19 +147,13 @@ $(dist_bin_SCRIPTS_INSTALL): $(dist_bin_SCRIPTS)
 	$(V)install -d $(@D)
 	$(V)install -m 0755 $(dist_bin_SCRIPTS) $(dist_bin_SCRIPTS_INSTALL)
 
-# generate the vendor folder with dep and patch when necessary
-$(BUILDDIR)/vendors-done:
-	@echo " VENDORS"
-	$(V)if [ ! -f $(BUILDDIR)/.dep-done ]; then \
+# generate the vendor folder with dep
+$(vendor_done):
```
jmstover: We need a check in here somewhere that is set as a dependency on the Build target (on `all`?) that does something like:

```
if ! dep check &>/dev/null; then echo "Vendor missmatch"; exit 1; fi 
```

We need to verify the files in `vendor/` match what we have stated in Gopkg.toml. We'll also need the "Dep Install" sections brought back into to CI config files.

@tri-adam

jmstover (dismissed):

mlocal/frags/Makefile.stub
tri-adam: Yes, I think that's a good approach. I'd suggest displaying the output of `dep check` instead of a generic `Vendor mismatch` message, since the command output is useful in diagnosing problems.

Note that this will also ensure that `Gopkg.lock` is in sync with `Gopkg.toml` and the `import`s in our Go files, in addition to making sure that the vendor directory is in sync with `Gopkg.lock`. IMO, this is a good thing, as these are all things we should be verifying.

cclerget: @jmstover @tri-adam Good it exists the subcommand doing that. Sounds good to introduce the command as a standalone `dep-check` target like `vendor` and call it during CI, but I'm reluctant to call it as part of build process during `make all` implying dep installation. IMO dep is a tool for the team/devs/contributors and should not be required by people who just want compile/try, this is also one of the reason we provide vendor. As long as this is covered by the CI, that should be good.

ArangoGutierrez: @cclerget could we add
```
if ! dep check &>/dev/null; then echo "Vendor missmatch"; exit 1; fi 
```
to `make check` ? that will check for `lint/vet/fmt` + `dep` 
`make check` used on testing/CI but not needed on build steps, so ti won't affect users that as you said 
> IMO dep is a tool for the team/devs/contributors and should not be required by people who just want compile/try

ArangoGutierrez (approved):
LGTM

yhcote (approved):
Looks good to me.

PR #2202, opened 2018-10-07
Add travis checks for centos 6 and 7 rpms

DrDaveD:
**Description of the Pull Request (PR):**

This PR adds travisci checks for centos 6 and 7, building rpms and testing them.   I considered circleci for just building, but couldn't find any examples for centos. Instead I ported forward the checks from singularity-2.6 and adapted them for singularity-3.  I included some of the infrastructure for debian and opensuse but will leave it to other people to complete them.

**This fixes or addresses the following GitHub issues:**

None

**Before submitting a PR, make sure you have done the following:**

- Read the [Guidelines for Contributing](../CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](../CHANGELOG.md) if necessary according to the [Contribution Guidelines](../CONTRIBUTING.md)
- Added tests to validate this PR and tested this PR locally with a `make testall`
- Based this PR against the appropriate branch according to the [Contribution Guidelines](../CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](../CONTRIBUTORS.md)

Attn: @singularity-maintainers

PR #2226, opened 2018-10-12
Fix RPM building with buildrpm script

jmstover:
This PR fixed a few of the RPM building issues with `dist/rpm/buildrpm`:
 * Fix `singularity.conf.in` 
 * Honor RPM macro values
 * Change install locations to appropriate Macro value (LOCALSTATEDIR, SYSCONFDIR, etc...)
 * Add DESTDIR for `make install`
 * Cleanup in `.spec.in` and makeit files


jmstover:
@DrDaveD
As you have an interest in this. This PR should make `buildrpm` work again, as well as honor RPM macros. It's a precursor patch to getting the build from dist tarball working.

DrDaveD:
@jmstover this conflicts with my pr #2202 which fixes the rpm builds and enables travis builds for centos6 & 7.  Could you please review that one first?

yhcote (approved):
Looks tedious to me.

yhcote:
@jmstover : I merged @DrDaveD PR #2202, Is there something that needs to be changed for this one to go in also?

Thanks.

jmstover:
@yhcote I don't think there was anything here that conflicts with #2202. That was more CI stuff than anything else.

*Edit:*
If the setting of the ldflags in `go_normal_opts.mk` works as advertised, then `-P release-stripped` can be removed here. I'd like to keep it as an option though.

DrDaveD:
@jmstover Please do remove the -P release-stripped from the singularity.spec.in.

jmstover:
@DrDaveD I changed it... Once quick question... It _appears_ to do as advertised. Though my `ld` shows using:
```
--build-id=0x[blah]
```
as the way to set it, not `-B 0x[blah]` ... using `-B` like that isn't mentioned in the man page (just -Bstatic, -Bgroup, etc...). `GNU ld version 2.26.20160125` (Slackware)

Is `-B` usage like this actually valid, and it's not documented? `objdump -sj .note.gnu.build-id singularity` _does_ return a section... but I can only get `-Wl,--build-id=[blah]` to work when doing it manually.

DrDaveD:
I thought at first it was ld too, but it isn't, it's go's own "link" program: https://golang.org/cmd/link/

jmstover:
Sounds good. I thought they offloaded to ld as well after doing some initial manipulation. :+1:

PR #2242, opened 2018-10-16
Polish runtime a little bit

sashayakovtseva:


src/runtime/starter/smaster.go
```diff
@@ -22,76 +22,76 @@ import (
...
-		conn, err := net.FileConn(comm)
+		comm := os.NewFile(uintptr(rpcSocket), "socket")
+		rpcConn, err := net.FileConn(comm)
+		comm.Close()
 		if err != nil {
 			fatalChan <- fmt.Errorf("failed to copy unix socket descriptor: %s", err)
 			return
 		}
-		comm.Close()
+		defer rpcConn.Close()
```
cclerget: Don't defer RPC connection close because `StartProcess` could be executed even if `CreateContainer` return an error

cclerget (changes requested):
Overall looks good Sasha ! Just need to address defer issue

src/runtime/starter/smaster.go
sashayakovtseva: My bad, thank you for pointing out.

cclerget (approved):

PR #2272, opened 2018-10-20
Build macOS from mconfig

mikegray:

This PR will allow building macOS/Darwin singularity CLI with same instructions as regular Linux binary - using `./mconfig && cd builddir && make` 

- Fixes #

### DIFF
--- a/mconfig
+++ b/mconfig
@@ -561,6 +561,8 @@ EOF
 ########################
 # modules
 ########################
+if [ "$host" = "unix" ]; then
+
 :>$genconfdir/module.lst
 
 found_modules=`(cat modules | awk '{ printf("%s ", $0) }') \
@@ -611,7 +613,7 @@ cat >> $genmk << EOF
 # make sure \`all' is the first rule
 all:
 EOF
-
+fi
 
 # call genmod.awk, generating Makefile components
 $topdir/genmod.awk modfile=$genconfdir/module.lst \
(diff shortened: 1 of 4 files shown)


mlocal/frags/Makefile.noruntime.stub
```diff
@@ -0,0 +1,111 @@
+singularity := $(BUILDDIR)/singularity
```
bauerm97, line 1: So, I haven't taken an in depth look into this PR yet so forgive me if I'm missing something obvious with this comment. However, I believe it shouldn't be necessary to have an entirely separate `Makefile.stub` for compiling without the runtime. Would it not be enough to just put build tags on the appropriate files which excludes those files from specific architectures?

mconfig
```diff
@@ -561,6 +561,8 @@ EOF
 ########################
 # modules
 ########################
+if [ "$host" = "unix" ]; then
```
bauerm97, line 564: I think it might be better to at the top of the file have one check for this:

```
if [ "$host" = "unix" ]; then
    compile="runtime"
else
    compile="noruntime"
fi
```

Or something along those lines. This way it's clear in each conditional why what we're checking is relevant:

```
if [ "$compile" = "runtime" ]; then
    ...
fi
```

Just a suggestion, I'll let @yhcote weigh in on if this is actually a good idea :)

mlocal/frags/Makefile.noruntime.stub
jscook2345: Build tags would be great because testing can also use them.

yhcote:
Hi Mike, thanks for the PR. I will keep it for when I merge those ideas with my next round of mconfig improvements. I will bring in other mechanisms to allow the system to check systems and oses a bit more easily. I will make sure I refer to this patch to remind me areas where things needs to merge so that platforms like MacOS and perhaps Windows can perform a restricted build.

mikegray:
Closing in favor of updates in PR https://github.com/sylabs/singularity/pull/2496

Issue #2269, opened 2018-10-19
make dist creates non-standard tarball


DrDaveD:
## Version of Singularity:

Current master branch

### Expected behavior

Expect make dist to create a tar.gz file where paths begin "singularity-3.0.0/" or whatever the current version number is.

### Actual behavior

Paths in the tar file begin "singularity/"

This is a problem because people and programs (e.g. rpmbuild) normally expect source tarballs to begin with package-version.  The github archive command creates source tarballs that way from tagged versions, and that is expected to be sufficient for passing directly to the rpmbuild command.

### Steps to reproduce behavior

make dist
tar tf singularity-*.tar.gz|head

yhcote:
We went back and forth with that one. The trade off is that when you untar singularity-3.0.0 in the GOPATH location. Going through the build process will fail. The expected folder is singularity/.

I don't have strong opinions but we should at least document one way or the other, depending on what pathname we choose to create.

DrDaveD:
I think that's a relatively minor consideration.  It's easy enough to put in a script to mv the directory from singularity-* to singularity.

DrDaveD:
I made a branch with this change, which is easy to do, but I am coming now to think that it's maybe not as important as I thought to have the version number in the top level.  We still wouldn't be able to use the github archive url because the VERSION and singularity.spec files are missing.   Instead, having a tarball made by `make -C builddir dist` in the github release area with each release will work better, and for that it doesn't matter what the path is inside.  With either path an `rpmbuild -ta singularity-3.X.Y.tar.gz` will work nice and easy.  And as @yhcote said it is slightly more convenient for compiling from go source to not have the version number.  So I don't know whether or not it is worth it to add the version just to look more standard.

zao:
As an end-user that downloads and extracts tarballs, it's extremely surprising to have a tarbomb that doesn't contain a version-dependent directory name.

It's quite possible that the build procedure for Go programs is different from anything else, but it's still _very_ surprising to a sysape that just wants to extract and see how to build it.

There's no mention of how to install from a release tarball in the instructions, it only talks about online builds where GitHub is available.

zao:
I would at least have a mention on the releases page that the tarball is not intended for normal use - that it needs to be extracted into some `GOPATH` or whatever.

Please think of the normal sysadmins that have been using 2.x series and never ever have touched Go.
I've personally postponed doing anything with 3.0 because of the installation situation.

GodloveD:
[...]

PR #2280, opened 2018-10-23
Fix developer rpmbuilds

DrDaveD:

Fixes two issues for developers of rpmbuilds: it cleans out an old build space if it exists (exactly the same as the %setup rule used to do) and prevents a git reset when there are modified files in a working space from killing a 'make dist'.

None


DrDaveD:
I also fixed another bad problem: the VERSION file created was always empty.  I also changed make-dist.sh to read the Version and Release from singularity.spec, which comes from mconfig, instead of recalculating defaults for them from git.  That way if there's a -V option to mconfig it will get carried through.

DrDaveD:
Also removed the need for temporarily adding the VERSION and singularity.spec file to git in order to build the tarball.  Those can cause problems like in issue #2284.

DrDaveD:
I also added back src and debuginfo rpms

scripts/make-dist.sh
```diff
@@ -4,23 +4,18 @@
 # in the LICENSE file.
 set -e
 
-package_name=singularity
-package_version=`(git describe --match 'v[0-9]*' --always 2>/dev/null || cat VERSION 2>/dev/null || echo "") | sed -e "s/^v//;s/-/_/g;s/_/-/;s/_/./g"`
-package_version_short=`(git describe --abbrev=0 --match 'v[0-9]*' --always 2>/dev/null || cat VERSION 2>/dev/null || echo "") | sed -e "s/^v//;s/-/_/g;s/_/-/;s/_/./g"`
+# spec file needs to be at the root of the project
```
yhcote: If spec file needs to be at the root of the project, why is it generated in the dist/rpm folder? I recommend that we generate it at the root of the project via mconfig right away and remove these 2 lines here.

scripts/make-dist.sh
```diff
@@ -4,23 +4,18 @@
...
-echo " DIST create tarball: $package_name-$package_version_short.tar.gz"
-git archive --format=tar --prefix=$package_name/ `git stash create` -o $package_name-$package_version_short.tar
-git reset VERSION singularity.spec
-gzip $package_name-$package_version_short.tar
-
+tarball="$package_name-$package_version_short.tar.gz"
+echo " DIST create tarball: $tarball"
+rm -f $tarball
+(echo VERSION; echo singularity.spec; git ls-files) | \
+        tar -T - --xform "s,^,$package_name/," -czf $tarball
```
yhcote: --xform is not supported on non GNU tar. That's why I preferred letting git generate the archive.

Can you achieve the same result with standard tools?

yhcote (changes requested):
Needs a little bit of work still.

scripts/make-dist.sh
DrDaveD: Yes I think I could use sed to add the prefix to every path and use tar -C .., because the source directory name happens to be singularity.  I will try it.

Separately in #2269 I am asking for getting the version name into the tarball.  That could be done too by temporarily creating a symlink, but since that's still controversial I will keep it out of this patch at least for now.

DrDaveD:
@yhcote I put in the requested changes, please re-review

mconfig
```diff
@@ -721,7 +721,7 @@ cat $makeit_fragsdir/Makefile.stub | awk \
 # Rpm spec for project package
 #######################################################################
 
-RPMSPEC=dist/rpm/singularity.spec
+RPMSPEC=singularity.spec
```
yhcote, line 724: This should be : RPMSPEC=$sourcedir/singularity.spec

mconfig
```diff
@@ -730,7 +730,7 @@ if [ "$VERSION" = "$RELEASE" ]; then
     # $package_version has no dash
     RELEASE=1
 fi
-sed "s/@PACKAGE_VERSION@/$VERSION/;s/@PACKAGE_RELEASE@/$RELEASE/;s/@GO_VERSION@/$go_version/" $RPMSPEC.in >$RPMSPEC
+sed "s/@PACKAGE_VERSION@/$VERSION/;s/@PACKAGE_RELEASE@/$RELEASE/;s/@GO_VERSION@/$go_version/" dist/rpm/$RPMSPEC.in >$RPMSPEC
```
yhcote: add $souredir to : dist/rpm/$RPMSPEC.in => $sourcedir/dist/rpm/$RPMSPEC.in

scripts/make-dist.sh
```diff
@@ -5,22 +5,21 @@
...
-gzip $package_name-$package_version_short.tar
-
+rmfiles="VERSION"
+tarball="$package_name-$package_version_short.tar.gz"
+echo " DIST create tarball: $tarball"
+rm -f $tarball
+pathtop="$package_name"
+thisdir="`basename $PWD`"
+if [ "$thisdir" != "$pathtop" ]; then
+    ln -fs $thisdir ../$pathtop
```
yhcote: Is that symlink the controversy you describe in your comment? Cause It breaks your build in this PR.

For the tarball directory name including the version number, I don't mind doing it. Unless @jmstover has reservations. Let's make sure we document how to prepare a source tree that is easy for users to unpack and build.

yhcote (changes requested):
A few asks still and make sure CI is happy. Thank you !

mconfig
DrDaveD: Hmm, the previous version didn't have it, but I see that it is needed.

I'll add $sourcedir where $RPMSPEC is used instead, since otherwise it won't work with dist/rpm.

scripts/make-dist.sh
DrDaveD: That's not the controversy, the controversy was over whether or not to include the version number in the top level directory name.

There were two problems with the travis build.  The problem that caused it to fail was that the parent directory is not writable and the current directory is not named "singularity".  Instead I will make the symlink in the current directory.  The second problem was that it was trying to read from dist/rpm/singularity.spec; that would have caused a failure later because of missing packages.

mconfig
yhcote: Yeah that was overlooked. Thanks!

scripts/make-dist.sh
yhcote: Ok Thanks for helping us wrapping that one up !

yhcote (approved):
LGTM

Issue #2285, opened 2018-10-24
Using cgo is painful, can we avoid it


DrDaveD:
## Version of Singularity:

master branch

### Expected behavior

Changing src/runtime/starter/c/starter.c affects the build

### Actual behavior

Changing starter.c has no effect on a development build output unless smaster.go is also edited or `go clean -cache` is run.  This is a [documented go bug](https://golang.org/cmd/go/#hdr-Build_and_test_caching) which I guess makes it a feature. 

### Steps to reproduce behavior

Build singularity, change starter.c, and build singularity again.

Can we avoid using the cgo feature?   I was frustrated for a couple of hours last night trying to figure out why changing starter.c had no effect.

cclerget:
@DrDaveD During development I recommend to set environment variable `GOCACHE=off`. I will submit a patch to force compilation of `starter` when `starter.c` is modified

DrDaveD:
Closed via #2288.

PR #2288, opened 2018-10-24
User capabilities and starter.c build fixes

### LINKED ISSUE #2287: ubuntu build fails on centos host
## Version of Singularity:
3.0.0-218.g093d5f6

### Expected behavior
build succeeds

### Actual behavior
build fails with
ERROR  : Failed to set ambient capability: Invalid argument
FATAL:   While performing build: while running engine: exit status 1

### Steps to reproduce behavior (executed on 20181023)
centos 7.2 virtualbox built via vagrant from bento/centos7.2 and configured thusly after logging in:

sudo yum -y groupinstall 'Development Tools'
sudo yum -y install openssl-devel git libuuid-devel golang squashfs-tools
echo 'export GOPATH=${HOME}/go' >> ~/.bashrc
echo 'export PATH=/usr/local/go/bin:${PATH}:${GOPATH}/bin' >> ~/.bashrc
source ~/.bashrc
mkdir -p $GOPATH/src/github.com/sylabs
cd $GOPATH/src/github.com/sylabs
git clone https://github.com/sylabs/singularity.git
cd singularity
go get -u -v github.com/golang/dep/cmd/dep
./mconfig
make -C builddir
sudo make -C builddir install

# created testing.def:
[vagrant@localhost ~]$ cat testing.def
BootStrap: library
From: ubuntu:16.04

%post
    apt-get -y update
    apt-get -y install fortune cowsay lolcat

%environment
    export LC_ALL=C
    export PATH=/usr/games:$PATH

%runscript
    fortune | cowsay | lolcat

%labels
    Author GodloveD

# attempted build
[vagrant@localhost ~]$ sudo /usr/local/bin/singularity build testing.sif testing.def
INFO:    Starting build...
 35.32 MiB / 35.32 MiB [================================================================================] 100.00% 10.65 MiB/s 3s
ERROR  : Failed to set ambient capability: Invalid argument
FATAL:   While performing build: while running engine: exit status 1

cclerget:

This PR fixes issues where user capabilities are falsely detected by just checking header file. Also fixes `starter.c` compilation when the file is modifed

- Fixes #2287 and #2285


mlocal/frags/Makefile.stub
```diff
@@ -133,7 +133,7 @@ $(libruntime): $(libutil_OBJ)
 	$(V)$(AR) rcs $@ $(libutil_OBJ)
 
 # starter
-$(starter): $(go_OBJ) $(starter_OBJ) $(libruntime)
+$(starter): $(go_OBJ) $(starter_SOURCE) $(starter_OBJ) $(libruntime)
```
bauerm97: If the idea here is to force `go build ...` to do a fresh build when the `starter.c` file has been changed, then this change won't work. The problem is that `go build` does not understand when a C file has changed, so the cache does not become invalidated. We'll still run the `go build` command, but it won't actually do a fresh compile with the new C code.

I think a fix for this should be submitted in a separate PR, and it can be addressed in a different way. Something like this:

```
starter_SOURCE := ...
starter_DUMMY := $(BUILDDIR)/src/runtime/starter/c/starter.c

$(starter_DUMMY): $(starter_SOURCE)
    go clean -cache
    touch $@

$(starter): $(go_OBJ) $(starter_DUMMY) $(starter_OBJ) $(libruntime)
    ...
```

Basically, create an empty target file in `BUILDDIR` which has the `starter.c` file as a prerequisite. If `starter.c` changes, we do `go clean -cache` (to force go to fully rebuild from scratch) and then touch the `starter_DUMMY` file to update it. Since `$(starter)` has that file as a prerequisite, if we ever change `starter.c`, the go build cache will get cleaned before we compile the `starter` binary.

cclerget: The idea is to recompile `starter` when `starter.c` is modified nothing more. Why not just simply disable go cache for the starter part by running `GOCACHE=off go build` ? That could limit bad surprises

cclerget: Or `go build -a` ?

bauerm97: Yea that would work as well. But that also means if any of `$(go_OBJ) $(starter_OBJ) $(libruntime)` are changed we always recompile the entire thing. Basically this means we lose caching at all, which isn't ideal. The ideal is that **if and only if** `starter.c` is modified, we run without cache

cclerget: Ok @bauerm97 changes applied

yhcote (changes requested):
The creation of C/bin files for test programs are now being done in builddir at a standard location managed by mconfig. This is NOT in the tree yet and still a work in progress, but let's update that PR once the mconfig PR is ready and merged.

yhcote (approved):
LGTM

PR #2311, opened 2018-10-27
Add dist/rpm/singularity.spec to .gitignore

bauerm97:


yhcote (changes requested):
This Cannot be done. The make-dist.sh script requires this file to be stage-able by git in order for archive inclusion.

I recommend to simply cancel that PR

yhcote:
I think that a form of this could be done soon as we reworked how we generate the release archive. Hold on...

yhcote (approved):
This works now. The file is actually generated in $sourcedir now.

jmstover (approved):

Issue #2294, opened 2018-10-24
Refactor of package locations for internal vs external usage


bauerm97:
It's now necessary that we make a small refactor of package locations to cleanly define which packages in Singularity are available for external usage as opposed to only being viable from inside the Singularity repository. I propose that we begin following the [internal package](https://docs.google.com/document/d/1e8kOo3r51b2BWtTs_1uADIA5djfXhPT36s6eHVRIvaU/edit) standard which Golang enforces. The new repository structure should look like this:

- `src/pkg`: Can contain code which is *always* able to be imported from another package, project, plugin, etc... In the context of the plugin system, all code which plugins will depend on **MUST** live here.
- `src/internal`: Can contain code which is only for usage by Singularity. Current `src/runtime/` will now live here, as well as parts of `src/pkg/build`, the entirety of `src/pkg/buildcfg`, and more packages which are only internally useful.

This will be a hassle to refactor certain packages to obey this new layout, but in order to maintain a very clean separation we must do this. Once we make this switch, we will no longer face issues with external projects (*e.g. remote builder*) being unable to import necessary packages due to compile-time constants

PR #2336, opened 2018-11-01
Initial pass of refactoring repository towards standard go layout

bauerm97:

This makes an initial pass at the refactoring of the repository we need to do. This makes the following relocations:

- `src/cmd` -> `cmd`
- `src/pkg` -> `internal/pkg`
- `src/runtime/starter` -> `cmd/starter`
- `src/runtime` -> `internal/pkg/runtime`

This is an initial first step towards #2294. The next step is to gradually move packages from `internal/pkg` -> `pkg` when they make sense to be available for external usage.


yhcote (approved):
looks good to me.

PR #2346, opened 2018-11-06
add support for http/https download of images

osallou:

This PR adds basic support for http/https in singularity pull/run commands allowing to pull an image file located on a remote http(s) server using singularity directly instead of downloading expected image manually.

Image is saved under latest part of uri, ex. https://a.b.c/d/e.img will be saved as e.img, no assumption is made of name or version.


internal/pkg/util/uri/uri.go
```diff
@@ -15,6 +15,10 @@ const (
 	Library = "library"
 	// Shub is the keyword for a shub ref
 	Shub = "shub"
+	// Http is the keyword for http ref
+	Http = "http"
+	// Https is the keyword for https ref
+	Https = "https"
```
ikaneshiro: Can you change these constants from `Http(s)` to `HTTP(S)` to make the go linter happy?

ikaneshiro (changes requested):
Hi @osallou! This looks pretty good to me. Just have one small thing that needs to be changed :+1:

internal/pkg/util/uri/uri.go
osallou: updated in new commit  58b0ac6

ikaneshiro (approved):

ikaneshiro:
Hey @osallou! Some of the paths for packages have changed recently on `master`, so I created a PR here: https://github.com/osallou/singularity/pull/1 that should fix those import issues. If that looks fine to you, or you just want to make the changes yourself, then we should be good to get this merged

ikaneshiro:
I _think_ the only thing that needs to be changed from here is `github.com/sylabs/singularity/internal/pkg/util/user-agent` needs to be changed to `github.com/sylabs/singularity/pkg/util/user-agent` in `client/net/pull.go`

osallou:
thanks, I just re-synced my branch with your latest  and updating it

PR #2358, opened 2018-11-11
Fix logic while creating loop devices

cclerget:

This PR fixes issue where loop devices are created in a way subject to race condition, typically when singularity is run with mpi. Also ensure created loop devices are owned by root user and root group.

### DIFF
--- a/internal/pkg/runtime/engines/singularity/rpc/server/server.go
+++ b/internal/pkg/runtime/engines/singularity/rpc/server/server.go
@@ -127,8 +127,7 @@ func (t *Methods) LoopDevice(arguments *args.LoopArgs, reply *int) error {
 		}
 	}
 
-	gr, err := user.GetGrNam("disk")
-	if err == nil {
+	if gr, err := user.GetGrNam("disk"); err == nil {
 		diskGID = int(gr.GID)
 	} else {
 		diskGID = 0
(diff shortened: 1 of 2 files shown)


jmstover:
`Also ensure created loop devices are owned by root user and root group.`

Shouldn't the loop device be owned by group `disk`?

internal/pkg/runtime/engines/singularity/rpc/server/server.go
```diff
@@ -127,8 +127,10 @@ func (t *Methods) LoopDevice(arguments *args.LoopArgs, reply *int) error {
 
 	runtime.LockOSThread()
 	syscall.Setfsuid(0)
+	syscall.Setfsgid(0)
```
jmstover: Should we be setting to `disk` group here? Or change when a new device is created with `Mknod()`?

internal/pkg/util/loop/loop.go
```diff
@@ -29,14 +29,15 @@ func (loop *Device) AttachFromFile(image *os.File, mode int, number *int) error
 		if fi, err := os.Stat(path); err != nil {
 			dev := int((7 << 8) | device)
 			esys := syscall.Mknod(path, syscall.S_IFBLK|0600, dev)
```
jmstover: To match other block devices, this should be:  `syscall.S_IFBLK|0660`   Ownership:  `root:disk`

jmstover (changes requested):

jmstover (dismissed):

jmstover (approved):

Issue #2320, opened 2018-10-30
Move minimum number of viable packages from internal/pkg to pkg


bauerm97:

bauerm97:
Currently I've found external references to these packages. Minimally these *must* be exported:

- [x] github.com/sylabs/singularity/src/pkg/jsonresp
- [ ] github.com/sylabs/singularity/src/pkg/build/types
- [x] github.com/sylabs/singularity/src/pkg/client/library
- [x] github.com/sylabs/singularity/src/pkg/util/user-agent

tri-adam:
The only one I can think of that isn't obvious is we are using the code from https://github.com/sylabs/singularity/tree/master/src/pkg/build/types/parser, but had to modify it due to `buildcfg` dependencies. That lives at https://github.com/sylabs/remote-build/tree/master/internal/pkg/parser at the moment.

bauerm97:
@yhcote has expressed interest in exposing these packages:

- [x] github.com/sylabs/singularity/internal/pkg/sypgp
- [x] github.com/sylabs/singularity/internal/pkg/signing

sashayakovtseva:
CRI needs those to work:

- [x] src/pkg/client/library
- [x] src/pkg/client/shub
- [x] src/pkg/signing
- [x] src/pkg/sypgp
- [x] src/pkg/util/loop
- [x] src/pkg/util/user-agent

PR #2364, opened 2018-11-14
Types Package Refactor

### LINKED ISSUE #2308: Incorrect sections in the definition file are silently ignored when building an image
## Version of Singularity:

3.0.0

### Expected behavior

Unknown section '%tests'

### Actual behavior

No output is given

Create a definition file with a `%tests` section (not that `%test` is the correct name for this).

Build an image from that definition file.

### LINKED ISSUE #2349: Only last %post section is processed when building an image
## Version of Singularity:

3.0.0

### Expected behavior

Given the following def file, successfully create an image file.  Note that the def files has multiple `%post` sections.

```
From: centos:7
Bootstrap: docker
%post
    . /.singularity.d/env/10-docker.sh

# GNU compiler
%post
    yum install -y \
        gcc \
        gcc-c++
    rm -rf /var/cache/yum/*

%files
    sources/hello.c /var/tmp/hello.c

%post
    cd /
    gcc -o /var/tmp/hello /var/tmp/hello.c
```

### Actual behavior

```
$ sudo singularity build hello.simg Singularity.def
INFO:   Starting build...
Copying blob sha256:aeb7866da422acc7e93dcf7323f38d7646f6269af33bcdb6647f2094fc4b3bf7
71.24 MiB / 71.24 MiB [====================================================] 7s
Copying config sha256:1e8ad5e15a62dea6c1a650a7ff14da420854aaac4f3b2b5e5da9f70fc6c22ec6
Storing signatures
INFO:   Copying sources/hello.c to /var/tmp/hello.c
INFO:   Running post scriptlet
+ cd /
+ gcc -o /var/tmp/hello /var/tmp/hello.c
/bin/sh: line 1: gcc: command not found
FATAL:   post proc: exit status 127
FATAL:   While performing build: while running engine: exit status 255\n",
FATAL:   failed to retrieved path for hello.simg: lstat hello.simg: no such file or directory
ERROR  : Child exit with status 255
```

Note that only the final `%post` section is actually evaluated.

See above.

The image builds correctly with 2.5.2 and 2.6.0.

### LINKED ISSUE #2404: [scif-apps]Files created at install step fall into an unexpected path
## Version of Singularity:

```
[eduardo@rhel7 singularity]$ singularity version
3.0.1-87.gaabae96
```

### Expected behavior

```
Bootstrap: library
From: alpine:3.7

%setup
    echo "SETUP"

%appinstall foo

    echo "INSTALLING FOO"
    touch filefoo.exec

%appinstall bar

    echo "INSTALLING BAR"
    touch filebar.exec
```

created files should end at ` "/scif/apps/foo/filefoo.exec"` 

### Actual behavior

Created files are at ` "/scif/apps/foo/scif/filefoo.exec"`

```
sudo singularity build issue.sif examples/apps/Singularity
singularity shell issue.sif
cd /scif/apps/foo
ls
<filefoo.exec> should be here...
```

ikaneshiro:

- Types and Parser package refactoring
- Support for repeated sections in definitions
- Parser now throws an error when encountering a non-standard section that can be handled at a higher level
- Parser supports app sections

- Fixes #2308 
- Fixes #2349 
- Related to #2320 
- Fixed #2404 


internal/pkg/build/sources/conveyorPacker_library.go
```diff
@@ -9,8 +9,8 @@ import (
 	"io/ioutil"
 	"os"
 
-	sytypes "github.com/sylabs/singularity/internal/pkg/build/types"
 	"github.com/sylabs/singularity/internal/pkg/sylog"
+	sytypes "github.com/sylabs/singularity/pkg/build/types"
```
bauerm97: Why is this the only place we're importing as `sytypes`?

internal/pkg/build/sources/conveyorPacker_shub.go
```diff
@@ -9,8 +9,8 @@ import (
 	"io/ioutil"
 	"os"
 
-	sytypes "github.com/sylabs/singularity/internal/pkg/build/types"
 	"github.com/sylabs/singularity/internal/pkg/sylog"
+	sytypes "github.com/sylabs/singularity/pkg/build/types"
```
bauerm97: Same question here as above, any reason to import as `sytypes`?

bauerm97 (changes requested):
Two small changes I point out here, then also we discussed adding a `Raw string` field to the definition file type to maintain all comments and formatting.

ikaneshiro:
@bauerm97 RFR

pkg/build/types/definition.go
```diff
@@ -90,5 +97,84 @@ func NewDefinitionFromJSON(r io.Reader) (d Definition, err error) {
...
+			w.Write([]byte(v))
+			w.Write([]byte("\n"))
+		}
+		w.Write([]byte("\n"))
+	}
+}
+
+// WriteDefinitionFile is a helper func to output a Definition struct
+// into a definition file.
+func WriteDefinitionFile(d *Definition, w io.Writer) {
```
bauerm97: Maybe this can just be replaced by writing out the new `Definition.Raw` field

pkg/build/types/definition.go
```diff
@@ -14,9 +15,11 @@ import (
...
 // Definition describes how to build an image.
 type Definition struct {
-	Header    map[string]string `json:"header"`
-	ImageData `json:"imageData"`
-	BuildData Data `json:"buildData"`
+	Header     map[string]string `json:"header"`
+	ImageData  `json:"imageData"`
+	BuildData  Data              `json:"buildData"`
+	CustomData map[string]string `json:"customData"`
+	RawDefData []byte            `json:"rawDefData"`
```
bauerm97: I think `Definition.RawDefData` is a bit too terse. Maybe just `Definition.Raw`

bauerm97 (changes requested):

pkg/build/types/definition.go
```diff
@@ -98,83 +98,22 @@ func NewDefinitionFromJSON(r io.Reader) (d Definition, err error) {
...
-		w.Write([]byte("\n\n"))
-	}
-}
-
-func writeFilesIfExists(w io.Writer, f []FileTransport) {
-
-	if len(f) > 0 {
+// WriteDefinitionFile is a helper func to output a Definition structs original definition file.
+func WriteDefinitionFile(d *Definition, w io.Writer) error {
+	n, err := w.Write(d.Raw)
```
bauerm97: Well, this causes a big problem it turns out. We use `WriteDefinitionFIle` on line `103` in this file to fill the `d.Raw` when the spec is not a def file, but rather a URI without a def file. This means that when doing `singularity shell docker://ubuntu`, for example, the `d.Raw` will be nil which breaks the creation of the SIF.

bauerm97 (approved):

PR #2365, opened 2018-11-14
Added output file in pull help

WestleyK:
# Updated pull help (added output file)

This PR, will update the `pull help` menu:

## Before:

```
[...]

Usage:
  singularity pull [pull options...] <URI>

[...]
```

## After:

```
[...]

Usage:
  singularity pull [pull options...] [output file] <URI>

[...]
```


src/docs/content.go
```diff
@@ -508,13 +508,13 @@ Enterprise Performance Computing (EPC)`
...
-  $ singularity pull library://alpine:latest
+  $ singularity pull alpine.sif library://alpine:latest
 
   From Docker
-  $ singularity pull docker://tensorflow/tensorflow:latest
+  $ singularity pull tensorflow.sif docker://tensorflow/tensorflow:latest
 
   From Shub
-  $ singularity pull shub://vsoch/singularity-images`
+  $ singularity pull singularity-images.sif shub://vsoch/singularity-images`
```
jscook2345, line 517: There is a '`' at the end that should not be there.

jscook2345 (changes requested):

src/docs/content.go
WestleyK: The '\`' is the end of a multiline string. Compiling singularity without the '\`' will fail.

jscook2345 (approved):

PR #2400, opened 2018-11-28
Fix network namespace usage for users

### LINKED ISSUE #2390: Non-root users can no longer use --net for isolation
## Semantics of `--net` flag have changed for Singularity 3.x.

Previously, in 2.x:

```
    -n|--net            Run container in a new network namespace (loopback is
                        only network device active)
```

And it was usable by non-root. Now in 3.x:

```
  -n, --net                    run container in a new network namespace
                               (sets up a bridge network interface by default)
      --network string         specify desired network type separated by
                               commas, each network will bring up a
                               dedicated interface inside container
                               (default "bridge")
      --network-args strings   specify network arguments to pass to CNI plugins
```

However, there's no way to specify "nothing but localhost" by default (presumably, possible by creating a custom SNI config) and no way to invoke that as non-root.

### Version of Singularity:

v.3.0.1

### Expected behavior

Ability to create a new network namespace as non-root.

```
$ singularity exec --net docker://alpine wget --spider -S example.com
wget: bad address 'example.com'
```

### Actual behavior

There's no way to unbundle the new CNI functionality.

```
$ singularity exec --net docker://alpine wget --spider -S example.com
FATAL:   container creation failed: Network requires root permissions
```

Use `--net` with Singularity 3.x as non-root

### Thoughts on fix

This is a clear behavior change from 2.x; probably the best would be to _not_ try to create networks by default (require explicit `--network` for this) and thus allow isolation use as before by non-root users.

If that's not acceptable, a new flag is needed that reproduces 2.x `--net` functionality (`--no-net` was [proposed on Slack](https://singularity-container.slack.com/archives/C0YKGK6D6/p1542215268404300)). I _personally_ don't like this approach as it makes the flag for `--net` differ in semantics, in a non-obvious way, from `--pid` / `--ipc` / `--userns` / `--uts`.

cclerget:

This PR fixes usage of network namespace for users since user can't enable network namespace just to provide network isolation, this PR will allow users to specify `--net --network=none` to enable namespace and bring up just a loopback interface.

- Fixes #2390

### DIFF
--- a/internal/pkg/network/network.go
+++ b/internal/pkg/network/network.go
@@ -110,24 +110,15 @@ func NewSetup(networks []string, containerID string, netNS string, cniPath *CNIP
 			finalCNIPath.Plugin = cniPath.Plugin
 		}
 	}
-	hasNone := false
+
 	for _, network := range networks {
-		if network == "none" {
-			hasNone = true
-			break
-		}
 		nlist = append(nlist, config{
 			name:    network,
 			portMap: make([]portMap, 0),
 			args:    make([][2]string, 0),
 		})
 	}
-	if hasNone {
-		if len(networks) > 1 {
-			return nil, fmt.Errorf("none network can't be specified with another network")
-		}
-		return &Setup{configs: []config{}}, nil
-	}
+
 	return &Setup{
 			configs:     nlist,
 			cniPath:     finalCNIPath,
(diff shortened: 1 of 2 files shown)


kav2k:
What's the behavior in case of, for example, `--network=none,bridge` specified?

cclerget:
@kav2k It fails with `container creation failed: no net configuration with name "none" in /usr/local/etc/singularity/network`

yhcote (approved):
LGTM

PR #2416, opened 2018-12-01
More thorough instance tests

### LINKED ISSUE #2129: More thorough instance testing
## Version of Singularity:

current master

### Expected behavior

I think that the instance tests should do the following:
1. Run all of the instance tests in a loop with the various flags that can be used with instances as input.  
2. Check for the presence and absence of the temp files that are created by the instance command at the appropriate times (check for cleanup after instance stop)
3. Look at the std out of the `instance list` command and make sure it returns what is expected.
4. call `instance list` when no instances are running and verify that the std out returns no running instances.

### Actual behavior

Looks like the tests just look at error codes of functions.

phpHavok:

This PR includes more thorough instance tests. Not all command-line options for the start, stop, and list commands have been tested, but all command-line options are included in the option structs to make testing in the future easier if desired. The testing framework is now tabular like the other test files.

Several operations are tested including verifying the number of instances running at various stages of the test suite. A real instance recipe has been built which launches a basic echo server to test network connectivity and basic background functionality. Additionally, here is the complete list of flags currently tested:

For start: --home, --hostname, --cleanenv, --contain, --workdir
For list: --json
For stop: single instance, --all

- Fixes #2129

### DIFF
--- a/cmd/singularity/instance_test.go
+++ b/cmd/singularity/instance_test.go
@@ -13,24 +13,172 @@ import (
 	"github.com/sylabs/singularity/internal/pkg/test"
 )
 
-func startInstance(image string, instance string) ([]byte, error) {
-	args := []string{"instance", "start", image, instance}
-	cmd := exec.Command(cmdPath, args...)
+type startOpts struct {
+	add_caps       string
+	allow_setuid   bool
+	apply_cgroups  string
+	bind           string
+	boot           bool
+	cleanenv       bool
+	contain        bool
+	containall     bool
+	dns            string
+	drop_caps      string
+	home           string
+	hostname       string
+	keep_privs     bool
+	net            bool
+	network        string
+	network_args   string
+	no_home        bool
+	no_privs       bool
+	nv             bool
+	overlay        string
+	scratch        string
+	security       string
+	userns         bool
+	uts            bool
+	workdir        string
+	writable       bool
+	writable_tmpfs bool
+}
 
+type listOpts struct {
+	json      bool
+	user      string
+	container string
+}
+
+type stopOpts struct {
+	all     bool
+	force   bool
+	signal  string
+	timeout string
+	user    string
+}
+
+func startInstance(image string, instance string, opts startOpts) ([]byte, error) {
+	args := []string{"instance", "start"}
+	if opts.add_caps != "" {
+		args = append(args, "--add-caps", opts.add_caps)
+	}
+	if opts.allow_setuid {
+		args = append(args, "--allow-setuid")
+	}
+	if opts.apply_cgroups != "" {
+		args = append(args, "--apply-cgroups", opts.apply_cgroups)
+	}
+	if opts.bind != "" {
+		args = append(args, "--bind", opts.bind)
+	}
+	if opts.boot {
+		args = append(args, "--boot")
+	}
+	if opts.cleanenv {
+		args = append(args, "--cleanenv")
+	}
+	if opts.contain {
+		args = append(args, "--contain")
+	}
+	if opts.containall {
+		args = append(args, "--containall")
+	}
+	if opts.dns != "" {
+		args = append(args, "--dns", opts.dns)
+	}
+	if opts.drop_caps != "" {
+		args = append(args, "--drop-caps", opts.drop_caps)
+	}
+	if opts.home != "" {
+		args = append(args, "--home", opts.home)
+	}
+	if opts.hostname != "" {
+		args = append(args, "--hostname", opts.hostname)
+	}
+	if opts.keep_privs {
+		args = append(args, "--keep-privs")
+	}
+	if opts.net {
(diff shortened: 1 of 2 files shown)


cmd/singularity/instance_test.go
```diff
@@ -6,67 +6,420 @@
...
 
 	"github.com/sylabs/singularity/internal/pkg/test"
 )
 
-func startInstance(image string, instance string) ([]byte, error) {
-	args := []string{"instance", "start", image, instance}
-	cmd := exec.Command(cmdPath, args...)
+const instanceStartPort = 11372
+const instanceDefinition = "../../examples/instances/Singularity"
+const instanceImagePath = "./instance_tests.sif"
```
ArangoGutierrez:
```go
const (
instanceStartPort = 11372 
instanceDefinition = "../../examples/instances/Singularity" 
instanceImagePath = "./instance_tests.sif"
)
```

cmd/singularity/instance_test.go
```diff
@@ -6,67 +6,420 @@
...
+		args = append(args, "--add-caps", opts.addCaps)
+	}
+	if opts.allowSetuid {
+		args = append(args, "--allow-setuid")
+	}
+	if opts.applyCgroups != "" {
+		args = append(args, "--apply-cgroups", opts.applyCgroups)
+	}
+	if opts.bind != "" {
+		args = append(args, "--bind", opts.bind)
```
ArangoGutierrez: Check https://github.com/sylabs/singularity/blob/master/cmd/singularity/actions_test.go#L44 ,so we can specify multiple binds on a test (not know but in a near future)

ArangoGutierrez (changes requested):
thanks for taking some time and write some boring tests @phpHavok !!! I really appreciate it! 
need some tweaks but is almost ready

cmd/singularity/instance_test.go
```diff
@@ -6,67 +6,420 @@
...
+	}{
+		{"InitialNoInstances", testNoInstances, false},
+		{"BasicEchoServer", testBasicEchoServer, false},
+		{"BasicOptions", testBasicOptions, false},
+		{"Contain", testContain, false},
+		{"CreateManyInstances", testCreateManyInstances, false},
+		{"StopAll", testStopAll, false},
+		{"FinalNoInstances", testNoInstances, false},
+	}
+	for _, currentTest := range tests {
```
ArangoGutierrez: on actions test we are using `tt` instead of `currentTest` , should be changed for consistency

cmd/singularity/instance_test.go
phpHavok: Changed!

cmd/singularity/instance_test.go
phpHavok: So I actually did look into that, but I'm a bit confused by something. The docs say that one bind accepts a single, comma-delimited string, but there's no indication that multiple `--bind`s can be supplied.
```
  -B, --bind strings           a user-bind path specification.  spec has
                               the format src[:dest[:opts]], where src and
                               dest are outside and inside paths.  If dest
                               is not given, it is set equal to src.
                               Mount options ('opts') may be specified as
                               'ro' (read-only) or 'rw' (read/write, which
                               is the default). Multiple bind paths can be
                               given by a comma separated list.
```
Edit: Is that why you said "in the future," i.e. multiple --binds will be an upcoming feature?

cmd/singularity/instance_test.go
phpHavok: Changed!

phpHavok:
@ArangoGutierrez Regarding the single test function you mentioned, I do have that I think, unless I misunderstood you.
```
func TestInstance(t *testing.T) {
```

cmd/singularity/instance_test.go
ArangoGutierrez: Look that the `for `loop appends a `--bind` along with the bind string
```go
argv = append(argv, "--bind", bind)
```

ArangoGutierrez:
> @ArangoGutierrez Regarding the single test function you mentioned, I do have that I think, unless I misunderstood you.
> 
> ```
> func TestInstance(t *testing.T) {
> ```

Yeah after reading that func 2 times I now agree with it.!! My bad

cmd/singularity/instance_test.go
phpHavok: Fixed; thanks!

phpHavok:
> > @ArangoGutierrez Regarding the single test function you mentioned, I do have that I think, unless I misunderstood you.
> > ```
> > func TestInstance(t *testing.T) {
> > ```
> 
> Yeah after reading that func 2 times I now agree with it.!! My bad

No problem :) New commits should address everything you've mentioned, pending CI!

ArangoGutierrez (approved):

jmstover (approved):
Looks fine from my side

PR #2419, opened 2018-12-03
Return status 1 on invalid/incomplete commands.

### LINKED ISSUE #2146: `singularity <unknown command>` returns 0 for an exit status
## Version of Singularity:

$ singularity --version
singularity version v3.0.0-alpha.2-310-g3812cec2-dirty

### Expected behavior

$ singularity bob
$ echo $?
1

### Actual behavior

$ singularity bob
$ echo $?
0

See above.

This is causing a few false positives with the integration tests. At the very least the `selftest` command, which is not implemented, is passing when it's test runs.

phpHavok:

This PR causes singularity to exit with status code 1 (rather than 0) when given invalid commands. This has been tested only for the commands which accept sub-commands: (singularity, singularity instance, singularity capability, singularity keys). The SilenceErrors directive prevents the error from being printed to the screen. I'm not sure if there are any unintended consequences of this for e.g. debugging. A couple tests had to be modified to expect this new failure.

- Fixes #2146

### DIFF
--- a/cmd/singularity/cli/capability.go
+++ b/cmd/singularity/cli/capability.go
@@ -8,6 +8,7 @@
 package cli
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"strings"
@@ -43,13 +44,16 @@ func init() {
 
 // CapabilityCmd is the capability command
 var CapabilityCmd = &cobra.Command{
-	Run:                   nil,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return errors.New("Invalid command")
+	},
 	DisableFlagsInUseLine: true,
 
-	Use:     docs.CapabilityUse,
-	Short:   docs.CapabilityShort,
-	Long:    docs.CapabilityLong,
-	Example: docs.CapabilityExample,
+	Use:           docs.CapabilityUse,
+	Short:         docs.CapabilityShort,
+	Long:          docs.CapabilityLong,
+	Example:       docs.CapabilityExample,
+	SilenceErrors: true,
 }
 
 func manageCap(capStr string, cmd int) {
--- a/cmd/singularity/cli_test.go
+++ b/cmd/singularity/cli_test.go
@@ -29,9 +29,9 @@ func TestSelfTest(t *testing.T) {
 	defer test.ResetPrivilege(t)
 
 	cmd := exec.Command(cmdPath, "selftest")
-	if b, err := cmd.CombinedOutput(); err != nil {
+	if b, err := cmd.CombinedOutput(); err == nil {
 		t.Log(string(b))
-		t.Fatalf("unexpected failure running selftest: %v", err)
+		t.Fatal("selftest passed, but it isn't implemented?")
 	}
 }
 
--- a/cmd/singularity/cli_test.go
+++ b/cmd/singularity/cli_test.go
@@ -31,7 +31,7 @@ func TestSelfTest(t *testing.T) {
 	cmd := exec.Command(cmdPath, "selftest")
 	if b, err := cmd.CombinedOutput(); err == nil {
 		t.Log(string(b))
-		t.Fatal("selftest passed, but it isn't implemented?")
+		t.Fatal("unexpected success running selftest")
 	}
 }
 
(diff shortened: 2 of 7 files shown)


cmd/singularity/cli/capability.go
```diff
@@ -43,13 +44,16 @@ func init() {
...
 
-	Use:     docs.CapabilityUse,
-	Short:   docs.CapabilityShort,
-	Long:    docs.CapabilityLong,
-	Example: docs.CapabilityExample,
+	Use:           docs.CapabilityUse,
+	Short:         docs.CapabilityShort,
+	Long:          docs.CapabilityLong,
+	Example:       docs.CapabilityExample,
+	SilenceErrors: true,
```
jscook2345, line 56: What is the purpose of SilenceErrors here? Should this be instead SilenceUsage?

phpHavok: I wasn't sure what the desired outcome was. I assumed the expected usage was to return 1 but still show usage, so I silenced the errors. I can swap that around if you prefer the other way.

Also, is it correct (as I've implemented it) to return an error on a blank "singularity" command?

cmd/singularity/cli_test.go
```diff
@@ -29,9 +29,9 @@ func TestSelfTest(t *testing.T) {
 	defer test.ResetPrivilege(t)
 
 	cmd := exec.Command(cmdPath, "selftest")
-	if b, err := cmd.CombinedOutput(); err != nil {
+	if b, err := cmd.CombinedOutput(); err == nil {
 		t.Log(string(b))
-		t.Fatalf("unexpected failure running selftest: %v", err)
+		t.Fatal("selftest passed, but it isn't implemented?")
```
ArangoGutierrez: I still prefer the message 
```
unexpected failure running <something>
```
for consistency across the testing suite

cmd/singularity/help_test.go
```diff
@@ -15,21 +15,26 @@ import (
...
 		t.Run(tt.name, test.WithoutPrivilege(func(t *testing.T) {
 			cmd := exec.Command(cmdPath, tt.argv...)
-			if b, err := cmd.CombinedOutput(); err != nil {
+			b, err := cmd.CombinedOutput()
+			if err != nil && tt.shouldPass {
 				t.Log(string(b))
 				t.Fatalf("unexpected failure running '%v': %v", strings.Join(tt.argv, " "), err)
+			} else if err == nil && !tt.shouldPass {
+				t.Log(string(b))
+				t.Fatalf("Test %s passed, but was expected to fail.", tt.name)
```
ArangoGutierrez: Same comment as above 
```
unexpected failure running <cmd>
```
or 
```
unexpected success running <cmd>
```
for consistency

ArangoGutierrez (changes requested):
Looks good for me, just some err messages changes for consistency across the test suite.

cmd/singularity/help_test.go
phpHavok: Fixed!

cmd/singularity/cli_test.go
phpHavok: Fixed!

ArangoGutierrez (approved):

bauerm97 (changes requested):
Hey @phpHavok 

First off, thanks for this contribution it's greatly appreciated. Everything here looks good on my end with your patch. In testing the return codes of all `singularity` commands, I did however find that `singularity instance stop` returns 0 even when run with no arguments. I looked into the `instance_stop.go` file briefly and I think the culprit is because that command can work with 0 arguments if a specific flag is set. Would you be willing to patch that command as well? I think the fix is as simple as changing it to a `RunE:` func, and returning an error after `cmd.Usage()`. That way all return codes should be consistent across the board.

Cheers!

phpHavok:
@bauerm97 Thanks for the comments, and no problem. I'm glad to be working on this project and getting to know you all.

I've made the change you requested and am running unit tests now. If they pass, I'll push so that CI can run here in a few minutes. The usage is actually automatically shown when RunE returns an error, so I removed the usage line as well. I think it works as intended now.

bauerm97 (approved):
LGTM, thanks for the contribution

PR #2426, opened 2018-12-05
Respect a read-only file mode.

### LINKED ISSUE #2403: Text file busy when running a container multiple times
## Version of Singularity:

singularity version 3.0.1-1.el7

### Expected behavior

No error when running a container concurrently multiple times.

### Actual behavior

This error appears when trying to run a container while another (from the same file) is already running.
```
$ ./test.sif
-bash: ./test.sif: /usr/bin/env: bad interpreter: Text file busy
```

1. Create a definition file `test.def`
```
Bootstrap: docker
From: ubuntu

%runscript
    read -p "waiting for input" a
```

2. `sudo singularity build test.sif test.def`

3. `./test.sif`

4. Now in another terminal try `./test.sif` again (while the other is still running)

I noticed that if I launch the second container with `singularity run test.sif`, it works.

phpHavok:

When executing an sif image as read-only, the image is still opened in read-write mode, disregarding the read-only request. This results in a strange bug when trying to execute the image like a binary (e.g. `./testimage` more than once. This PR fixes this by respecting the read-only request, and opening the image as read-only.

The previous code attempted to circumvent the request for a read-only image by running an access system call. The new code will always respect the mode.

- Fixes #2403 

### DIFF
--- a/internal/pkg/image/image.go
+++ b/internal/pkg/image/image.go
@@ -148,9 +148,9 @@ func Init(path string, writable bool) (*Image, error) {
 		if mode&os.O_RDWR != 0 {
 			if err := syscall.Access(resolvedPath, 2); err != nil {
 				sylog.Debugf("Opening %s in read-only mode: no write permissions", path)
-				mode = os.O_RDONLY
-				img.Writable = false
 			}
+			mode = os.O_RDONLY
+			img.Writable = false
 		}
 
 		img.File, err = os.OpenFile(resolvedPath, mode, 0)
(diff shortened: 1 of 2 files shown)


cclerget:
Hi @phpHavok and thank you for taking this one !

Actually your fix breaks `--writable` flag for ext3 images and writable overlay images, the mode must be set accordingly in `internal/pkg/image/sif.go` because SIF can contain writable sections too :

```
func (f *sifFormat) openMode(writable bool) int {
	if writable {
		return os.O_RDWR
	}
	return os.O_RDONLY
}
```

phpHavok:
Thanks @cclerget ! Sorry for breaking that, haha. New changes are building now. I did learn a ton trying to track this problem down though!

cclerget:
That's great @phpHavok ! No apologies, and as a man said : 

> The only man who never makes a mistake is the man who never does anything

cclerget (approved):
LGTM ! Thank you very much @phpHavok !

PR #2430, opened 2018-12-06
Fix misleading token warning

bauerm97:
Fixes a misleading warning about token not being found even when pulling from non-sylabs cloud locations. Also adds some instruction to the `WarningTokenFileNotFound` on how to generate a new token

### DIFF
--- a/cmd/singularity/cli/singularity.go
+++ b/cmd/singularity/cli/singularity.go
@@ -175,9 +175,6 @@ func sylabsToken(cmd *cobra.Command, args []string) {
 	if authToken == "" {
 		authToken, authWarning = auth.ReadToken(defaultTokenFile)
 	}
-	if authToken == "" && authWarning == auth.WarningTokenFileNotFound {
-		sylog.Warningf("%v : Only pulls of public images will succeed", authWarning)
-	}
 }
 
 // envAppend combines command line and environment var into a single argument
--- a/internal/pkg/util/auth/token.go
+++ b/internal/pkg/util/auth/token.go
@@ -22,7 +22,7 @@ const (
 	// WarningEmptyToken Warning return for empty token string
 	WarningEmptyToken = "Token file is empty"
 	// WarningTokenFileNotFound token file not found
-	WarningTokenFileNotFound = "Authentication token file not found"
+	WarningTokenFileNotFound = "Authentication token file not found. To generate a token, visit cloud.sylabs.io/auth"
 	// WarningCouldntReadFile Warning return for issues when reading file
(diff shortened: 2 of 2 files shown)


internal/pkg/util/auth/token.go
```diff
@@ -22,7 +22,7 @@ const (
 	// WarningEmptyToken Warning return for empty token string
 	WarningEmptyToken = "Token file is empty"
 	// WarningTokenFileNotFound token file not found
-	WarningTokenFileNotFound = "Authentication token file not found"
+	WarningTokenFileNotFound = "Authentication token file not found. To generate a token, visit cloud.sylabs.io/auth"
```
jmstover, line 25: Do you want to put in a leading  `https://`  here?

bauerm97: No, it gets filled in automatically by any modern browser afaik

jmstover: Default will be `http` ... The front end will redirect to `https` ... This is just to be explicit.

cmd/singularity/cli/singularity.go
```diff
@@ -175,9 +175,6 @@ func sylabsToken(cmd *cobra.Command, args []string) {
 	if authToken == "" {
 		authToken, authWarning = auth.ReadToken(defaultTokenFile)
 	}
-	if authToken == "" && authWarning == auth.WarningTokenFileNotFound {
-		sylog.Warningf("%v : Only pulls of public images will succeed", authWarning)
-	}
```
tri-adam, line 184: `authWarning` is set but never used, meaning `WarningTokenFileNotFound` (or any other `authWarning` value is not output to the user in any scenario.)

It would be better to move this function to `internal/pkg/util/auth`, and have it return the token and an `error`. Then call directly from the Library/Builder/Keystore commands that use it. That way, you could output the specific reason with the `FATAL` line.

tri-adam (changes requested):

internal/pkg/util/auth/token.go
gmkurtzer: Simplicity is good and but URLs in many terminals are clickable when they are prefixed properly. Let's prefix with https://.

cmd/singularity/cli/singularity.go
gmkurtzer: If this means that we can control the particular fatal error output from the server side,... I'd give it a huge +10.

GodloveD:
Where does this stand?  Sure would be nice to squash this.

GodloveD:
The issue fixed by this PR has been more thoroughly addressed by subsequent PRs afaik.

PR #2465, opened 2018-12-13
Build from scratch

ghuls:

This pull request mainly adds support for "Bootstrap: scratch" in Singularity definition file similar to Docker "FROM scratch" feature.

Examples:

  - Create a Busybox container from scratch (similar to "Bootstrap: busybox"):

```
Bootstrap: scratch
    
%setup
    busybox_url='https://www.busybox.net/downloads/binaries/1.28.1-defconfig-multiarch/busybox-x86_64'
    
    # Create necessary config files.
    printf 'root:!:0:0:root:/root:/bin/sh\n' > "${SINGULARITY_ROOTFS}/etc/passwd"
    printf 'root:x:0:\n' > "${SINGULARITY_ROOTFS}/etc/group"
    
    chmod 0644 "${SINGULARITY_ROOTFS}/etc/passwd" "${SINGULARITY_ROOTFS}/etc/group"
    
    # Create bin directory.
    mkdir "${SINGULARITY_ROOTFS}/bin"
    chmod 0755 "${SINGULARITY_ROOTFS}/bin"
    
    # Download busybox binary.
    curl -o "${SINGULARITY_ROOTFS}/bin/busybox" "${busybox_url}"
    chmod 0755 "${SINGULARITY_ROOTFS}/bin/busybox"
    
    # Install busybox.
    "${SINGULARITY_ROOTFS}/bin/busybox" --install "${SINGULARITY_ROOTFS}/bin"
```

  - Extract a rootfs archive in the minimal Singularity container:

```
$ cat alpine_from_scratch.def
Bootstrap: scratch

%setup
    alpine_minirootfs_url='http://dl-cdn.alpinelinux.org/alpine/v3.8/releases/x86_64/alpine-minirootfs-3.8.1-x86_64.tar.gz'

    # Download and extract alpine minirootfs.
    curl "${alpine_minirootfs_url}" \
       | tar xz -C "${SINGULARITY_ROOTFS}" --exclude=./etc/hosts
```

It also fixes some other small things:

  - Add support for "Bootstrap: scratch" in Singularity definition file.
  - Add example Singularity definition files for "Bootstrap: scratch".
  - Remove creation of "/etc/hosts" when using "Bootstrap: busybox".
  - Add "Bootstrap: busybox" to help of "singularity build".
  - Set shell to "/bin/sh" for all scripts in "/.singularity.d/env/".


ghuls:
@bauerm97 I fixed the formatting issue that was left in the previous pull request.

ghuls:
I don't get it why `circleci.com` failed for:
```
echo true | singularity shell docker://busybox
```

My code didn't touch anything docker related.

GodloveD:
Heya @ghuls I think you should move the changes in the shebangs in the env scripts to a seperate PR so we can properly discuss them.  There is a reason we made the change from `sh` to `bash`.  If we want to revisit that change and have a second discussion about it, then I would like to do so outside of the context of this PR. 

Thanks!

GodloveD (changes requested):
Heya @ghuls I think you should move the changes in the shebangs in the env scripts to a seperate PR so we can properly discuss them. There is a reason we made the change from sh to bash. If we want to revisit that change and have a second discussion about it, then I would like to do so outside of the context of this PR.

Thanks!

ghuls:
@GodloveD No problem, I can move them out. It is just that some of the scripts had /bin/sh and some /bin/bash. As far as I know, Busybox does not have bash (but ash), so the scripts should at least stay compatible with ash.

Any comments on the rest of the patches?

GodloveD:
Overall I think it's a great idea!  I don't know when I will have time to review though.  Maybe someone will get to it before me?  We'll see.

bauerm97:
While I agree with @GodloveD that maybe the shebang line changes should be their own PR, I do agree with the change.

>There is a reason we made the change from sh to bash. 
I actually don't think there was a reason that we changed from `sh` -> `bash`. If anything, that was just a goofy typo on my or Ian's part (probably mine ;) )... So I think it's important that we maintain `#!/bin/sh` across the board for maximum consistency and portability

GodloveD:
Right now, a user can do something like this:

```
%environment
    export FOO=bar
    export THIS=that
```
If we change all of the shebangs to `#!/bin/sh` they might have to rewrite def files as:
```
%environment
    FOO=bar
    THIS=that
    export FOO THIS
```
Because the former is not POSIX compliant while the latter is.  This changes behavior of existing def files and is likely to trip users up.  I don't really think that it should be necessary to require users to give up using bashisms when they may not even know they are using bashims and may not understand why their syntax no longer works.

ghuls:
@GodloveD 

https://www.unix.com/man-page/POSIX/1posix/export/
https://www.mankier.com/1p/export

As far as I can tell, both ways are valid by POSIX. The only thing that is not specified is just `export` (that is why they have `export -p`).
`checkbashisms` (Debian/Ubuntu) also does not complain about both syntaxes.

ghuls:
This pull request is not split in 3 separate pull requests.

GodloveD:
Thank you @ghuls.  Is it OK to close this PR now?

PR #2481, opened 2018-12-18
Fix unprivileged --contain on centos7 and move contained/minimal dev tty to /dev/console

### LINKED ISSUE #2406: Unprivileged --contain fails mounting devpts on centos 7.x
## Version of Singularity:

3.0.1

### Expected behavior

Running unprivileged with -c/--contain works

### Actual behavior

Fails with fatal error

On a centos 7.x system do:

```
$ singularity exec -u -c sandbox-image bash
FATAL:   container creation failed: mount error: can't mount devpts filesystem to /var/singularity/mnt/session/dev/pts: invalid argument
```

There has been extensive discussion on related issues on singularity-2.x.  See especially #1186 which refers to most of the rest of the discussion and was never fully resolved.  Creating /dev/pts on RHEL 7.x is apparently a privileged operation, although it isn't in later fedora kernels.

3.0.1 and the current master branch attempts to implement the same behavior as 2.6.0 but there's a flaw in the conversion to delayed mounts.  It checks the return value from system.Points.AddFS(..."devpts"...) for failure and attempts to ignore it, but that call doesn't fail because it just queues the request to be performed later.  It only fails later when it actually attempts the mount, and then it is treated as a fatal error.  In my 2.6.0 underlay implementation I created a "/dev" space for --contain and immediately mounted the devices inside it, and only delayed the mounting of "/dev".  3.0.x instead defers the mounting of each device separately. 

Perhaps the answer is to use my 2.6.0 approach in the new code, but that does not resolve the issue in #1186.  What do people think about when using --contain in unprivileged mode from a tty, on all operating systems bind mount /dev/pts/ptmx from the host, allocate a new pseudo tty, and bind mount the /dev/pts/N that is allocated?  I don't see a security problem with that.  [This go code](https://github.com/kr/pty/blob/master/pty_linux.go) would probably be helpful.  The docker run -t option allocates a new pseudo tty on the host and maps it to /dev/console in the container.

Oh, and I just noticed another problem with privileged mode --contain and tty: it does not allocate a new tty, it just makes a new /dev/pts/ptmx and keeps the original tty, but it does not map the original tty (e.g. /dev/pts/0) into the container.  That's another flaw, you're supposed to be able to re-open the device under the name given by tty.  Also if one shell spawns multiple singularity images from the same tty, they will share the same tty, which does not give the complete isolation desired.  I think a new pseudo-tty should be allocated with --contain whether privileged or not privileged.

I notify here the people who were involved in the #1186 discussion: @olifre @hollowec @cclerget

DrDaveD:

This PR prevents unprivileged --contain from failing on centos7.  Allocating a new devpts fails when unprivileged on the current centos7 kernel, and this PR makes it silently skip that as was done with singularity-2.6.x.  There's a [Red Hat bug]( https://bugzilla.redhat.com/show_bug.cgi?id=1522992) about this limitation and they will probably fix it at some point.  Meanwhile, allocating a new devpts only works with the --fakeroot option.

This PR also moves the tty for a minimal /dev (which is implied with --contain) to /dev/console.   This is needed with --contain whether or not there's a new devpts and whether or not it is unprivileged, because the host /dev/pts in all those cases is not mounted inside the container.

- Fixes #2406

Attn: @singularity-maintainers, @cclerget


cmd/starter/c/starter.c
```diff
@@ -265,6 +265,32 @@ static void prepare_scontainer_stage(int stage) {
         }
     }
 #endif
+
+    if ( config.isMinimalDev ) {
+        if ( isatty(STDIN_FILENO) || isatty(STDOUT_FILENO) || isatty(STDERR_FILENO) ) {
```
cclerget: Why doing those steps ? instead of simply bind /proc/self/fd/0 (or 1/2 depending if they point to a terminal) -> /dev/console from Go code ?

internal/pkg/runtime/engines/singularity/container.go
```diff
@@ -994,14 +1007,28 @@ func (c *container) addDevMount(system *mount.System) error {
...
+				return fmt.Errorf("failed to add devpts filesystem: %s", err)
+			}
+			// add additional PTY allocation symlink
+			if err := c.session.AddSymlink("/dev/ptmx", "/dev/pts/ptmx"); err != nil {
+				return fmt.Errorf("failed to create /dev/ptmx symlink: %s", err)
+			}
+
+		}
+		// add /dev/console mount pointing to original tty if there is one
+		for fd := 0; fd <= 2; fd++ {
```
cclerget, line 1015: To continue with the first comment, you could bind there instead, no need to call ttyname, just call `https://godoc.org/golang.org/x/crypto/ssh/terminal#IsTerminal` and bind accordingly with `/proc/self/fd/X` to `/dev/console`

internal/pkg/runtime/engines/singularity/container.go
```diff
@@ -994,14 +1007,28 @@ func (c *container) addDevMount(system *mount.System) error {
 			devptsPath, _ := c.session.GetPath("/dev/pts")
 			err = system.Points.AddFS(mount.DevTag, devptsPath, "devpts", syscall.MS_NOSUID|syscall.MS_NOEXEC, options)
 			if err != nil {
-				sylog.Verbosef("Couldn't mount devpts filesystem, continuing with PTY functionality disabled")
-			} else {
-				if err := c.addSessionDev("/dev/tty", system); err != nil {
+				return fmt.Errorf("failed to add devpts filesystem: %s", err)
+			}
+			// add additional PTY allocation symlink
+			if err := c.session.AddSymlink("/dev/ptmx", "/dev/pts/ptmx"); err != nil {
```
cclerget, line 1009: In case of failure with devpts / user namespace bug, we end with a dead link

cmd/starter/c/starter.h
```diff
@@ -50,6 +50,7 @@ struct cConfig {
     off_t pidNsPathOffset;
     unsigned char isSuid;
     unsigned char isInstance;
+    unsigned char isMinimalDev;
```
cclerget: Not required with suggestion above

internal/pkg/runtime/engines/config/starter/starter.go
```diff
@@ -69,6 +69,23 @@ func (c *Config) GetInstance() bool {
 	return false
 }
 
+// SetMinimalDev sets if there's a minimal /dev
+func (c *Config) SetMinimalDev(minimalDev bool) {
```
cclerget: Same, not required

internal/pkg/runtime/engines/singularity/container.go
```diff
@@ -5,6 +5,11 @@
 
 package singularity
 
+/*
```
cclerget: not required

internal/pkg/runtime/engines/singularity/container.go
```diff
@@ -399,9 +404,13 @@ func (c *container) mount(point *mount.Point) error {
 			if flags&syscall.MS_REMOUNT != 0 {
 				return fmt.Errorf("can't remount %s: %s", point.Destination, err)
 			}
-			// mount error for filesystems is considered fatal
 			if point.Type != "" {
-				return fmt.Errorf("can't mount %s filesystem to %s: %s", point.Type, point.Destination, err)
+				if point.Source == "devpts" {
```
cclerget, line 404: I would go for something more generic here, like the possibility to register for a mount point an "error" function executed in case of error. But it's good for now

internal/pkg/runtime/engines/singularity/prepare.go
```diff
@@ -279,6 +279,8 @@ func (e *EngineOperations) prepareContainerConfig(starterConfig *starter.Config)
 
 	starterConfig.SetInstance(e.EngineConfig.GetInstance())
 
+	starterConfig.SetMinimalDev(e.EngineConfig.File.MountDev == "minimal" || e.EngineConfig.GetContain())
```
cclerget: not required

cclerget (changes requested):
Code can be simplified by removing C part. Between thank you very much for taking this one @DrDaveD

cmd/starter/c/starter.c
DrDaveD: This is needed so that /proc/self/fd/[0-2] will point to /dev/console inside the container.  Newer ttyname() implementations give the right answer most of the time (not when a new copy of the /dev/pts/N device is allocated), but older implementation such as an el6 and other programs that bypass ttyname() and instead do readlink on /proc/self/fd/[0-2] would not get the right answer.  I will put in a big comment with this info.

internal/pkg/runtime/engines/singularity/container.go
DrDaveD: What you mean is to do a readlink() on /proc/self/fd/X and bind that.  The problem is that doesn't always work, for example inside docker where that mysteriously points to non-existent "/N" paths.   ttyname() has a better chance of working when the /proc/self/fd/X symlink targets are broken.

internal/pkg/runtime/engines/singularity/container.go
DrDaveD: Yes I know but I don't think it makes anything break that wouldn't already be broken so I thought it would be fine to leave it.

internal/pkg/runtime/engines/singularity/container.go
DrDaveD: In slack conversation with @cclerget on this, the real thing he'd like to avoid is calling a C function from go.  Unfortunately there's no go version of ttyname (other than one we found that was covered by GPL) so for now he asked me to replace this code with the less functional readlink().  He will later as part of the OCI implementation change it to invoke ttyname() in starter.c.

cmd/starter/c/starter.c
cclerget: @DrDaveD Could you rewrite this code in Go into `internal/pkg/runtime/engines/singularity/process.go` rather ? in order to make it specific to singularity engine because that could break other engines (eg: OCI engine). Like that there is no need for `isMinimalDev` and associated getter/setter methods for starter config

DrDaveD:
Close & reopen to retry transient circle ci error.

DrDaveD:
Close & reopen didn't do it.  I need somebody with privileges to go to the [circle ci test](https://circleci.com/gh/sylabs/singularity/3381) and press Rerun workflow.

cclerget (approved):

PR #2502, opened 2018-12-26
Library cache

### LINKED ISSUE #2474: Library Image Caching For Pull and Build Commands
## Version of Singularity:

3.0

### Expected behavior

Use cached image when building or pulling from the a library source multiple times.

### Actual behavior

Downloads a new image every time.

Build or pull the same image twice.

ikaneshiro:

Uses cache for builds and pulls of library images.

- Fixes #2474


internal/app/singularity/actions.go
```diff
@@ -144,22 +143,7 @@ func handleOCI(cmd *cobra.Command, u string) (string, error) {
...
-
-	if exists, err := cache.LibraryImageExists(libraryImage.Hash, imageName); err != nil {
-		return "", fmt.Errorf("unable to check if %v exists: %v", imagePath, err)
-	} else if !exists {
-		sylog.Infof("Downloading library image")
-		libexec.PullLibraryImage(imagePath, u, "https://library.sylabs.io", false, authToken)
-	}
-
-	return imagePath, nil
+	return cache.PullLibraryImage(u, "https://library.sylabs.io", authToken)
```
jmstover, line 146: Why is the URL hard coded here? Shouldn't this be like with later calls and using something similar to:  `cp.LibraryURL`

ikaneshiro: This is hardcoded because actions dont currently support the `LibraryURL` flag, probably a good thing to add

internal/pkg/client/cache/library.go
```diff
@@ -38,3 +43,29 @@ func LibraryImageExists(sum, name string) (bool, error) {
 
 	return true, nil
 }
+
+// PullLibraryImage is the function that is responsible for pulling an image from a Sylabs library into the cache.
+// Requires libraryRef to include library://
+func PullLibraryImage(libraryRef, libraryURL string, authToken string) (string, error) {
```
bauerm97, line 49: I don't think this function belongs here. The purpose of the `cache` package is just to return whether or not a image exists in that location in a repeatable way. Notice none of the other files (`oci.go`, `shub.go`, etc...) are responsible for downloading the actual image.

internal/pkg/libexec/pull.go
```diff
@@ -23,8 +26,33 @@ func PullNetImage(image, libraryURL string, force bool) {
 }
 
 // PullLibraryImage is the function that is responsible for pulling an image from a Sylabs library.
-func PullLibraryImage(image, libraryRef, libraryURL string, force bool, authToken string) {
-	err := library.DownloadImage(image, libraryRef, libraryURL, force, authToken)
+func PullLibraryImage(filePath, libraryRef, libraryURL string, force bool, authToken string) {
```
bauerm97, line 29: This function would be a good candidate to take on the functionality from the function I mentioned in the `cache` package. Here is where we could essentially check if the image exists in the cache (using functions from `pkg cache`) and then download it by calling into the library's API

bauerm97 (changes requested):

internal/pkg/libexec/pull.go
ikaneshiro: A problem with moving the function to this location is that the build package relies on this function in the library conveyorPacker and the libexec package imports the build package to build oci blobs into a sif

internal/pkg/client/cache/library.go
ikaneshiro: That's reasonable it doesn't really match the function of the rest of the package, though I don't know a better place to put it.

michaelmoore10:
@ikaneshiro Would this fix #2546 ?

ikaneshiro:
Hey @michaelmoore10! This won't fix #2546 since this will only apply to library souces. Eg. `library://alpine:latest`. #2546 will need to be handled in another PR

ikaneshiro:
Closing as #2551 will now handle this issue

PR #2505, opened 2018-12-27
Respect uppercase proxy env pass-through

### LINKED ISSUE #2504: Proxy env var passthrough does not match 2.6
## Version of Singularity:

3.0.x

### Expected behavior

In 2.6.x proxy environment vars are passed through into the container whether they are lower or uppercase. Most programs correctly look for lowercase these days, but some stuff (esp. crusty old bioinf scripts) may be expecting `HTTP_PROXY` etc. not `http_proxy`

In 2.6 the PR fixing the pass through did a case-insensitive check on the proxy vars:

https://github.com/sylabs/singularity/pull/1151/files

### Actual behavior

In 3.0.x the proxy var comparison is case-sensitive, so is not passing `HTTP_PROXY` etc.

dtrudg:

- Pass through upper case proxy env vars
- Update the `clean_test.go` so it checks for all pass-through proxy env vars.

- Fixes #2504 

### DIFF
--- a/internal/pkg/util/env/clean.go
+++ b/internal/pkg/util/env/clean.go
@@ -20,10 +20,15 @@ const (
 var alwaysPassKeys = map[string]bool{
 	"TERM":        true,
 	"http_proxy":  true,
+	"HTTP_PROXY":  true,
 	"https_proxy": true,
+	"HTTPS_PROXY": true,
 	"no_proxy":    true,
+	"NO_PROXY":    true,
 	"all_proxy":   true,
+	"ALL_PROXY":   true,
 	"ftp_proxy":   true,
+	"FTP_PROXY":   true,
 }
 
 // SetContainerEnv cleans environment variables before running the container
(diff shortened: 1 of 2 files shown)


sashayakovtseva:
Isn't it better to keep only upper or only lower case envs in `alwaysPassKeys`? I suggest translating key in `addIfReq` func with strings.ToUpper / ToLower, that is a more clean way to achieve the same IMO.

dtrudg:
Hi @sashayakovtseva - I was originally going to do a case conversion for the comparison in the loop - but decided against it as here the existing code is passing through `TERM` (which should *always* be an upper case var) and not `term` - I didn't want to change that particular behavior. Case conversion for the comparison would also pass through mistaken mixed caps  e.g. `hTtTp_PrOxY` or `Term`.

I have no strong opinion though - anything that will pass `XXX_PROXY` in upper case will address the issue, so please let me know if you do want me to change this to do a case conversion prior to comparison.

sashayakovtseva:
I wonder if lower/mixed case envs should be passed as it is to container, so conversion to upper case for each env var seems reasonable to me. So with conversion it won't matter whether you've passed `hTtTp_PrOxY` or `http_proxy`, they both will result in standard `HTTP_PROXY` env. 
Might want somebody else's opinion here.

dtrudg:
Lower case is the preferred/standard form that most software uses. Some like I think wget ignore upper case:

https://serverfault.com/questions/571887/why-wget-doesnt-read-http-proxy-in-uppercase-format

... so we can’t always force the proxy env vars to upper case in the container.

Some other software only reads upper case. This means we’d need to pass both forms in if set.




________________________________

sashayakovtseva:
Thanks for clarification, in this case I think your changes are perfectly fine.

sashayakovtseva (approved):

bauerm97 (approved):
I think the justification for only passing `ALL_CAPS` or `no_caps` forms here is solid. Looks good to me, many thanks @dctrud

PR #2511, opened 2019-01-02
Store new public key by default when prompts disabled

sashayakovtseva:
A small optimization made, so fetched public key is stored by default then SIF verification is done with prompts disabled.


cclerget (approved):
LGTM

PR #2533, opened 2019-01-11
Check ownership of config file

ArangoGutierrez:

Check for ownership of config file before running


jmstover (approved):
Works for me. I'll let @cclerget and/or @tri-adam speak up here.

internal/app/singularity/actions.go
```diff
@@ -310,6 +310,14 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 	engineConfig := singularity.NewConfig()
 
 	configurationFile := buildcfg.SYSCONFDIR + "/singularity/singularity.conf"
+	configFile, err := os.Stat(configurationFile)
```
cclerget: The whole check could be moved around this line : https://github.com/sylabs/singularity/blob/d61d61e2594c50ac3e1e0611c16a087c2f57d09f/internal/pkg/runtime/engines/singularity/prepare.go#L417

internal/app/singularity/actions.go
```diff
@@ -310,6 +310,14 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
...
+	if err != nil {
+		sylog.Fatalf("Error with Stat() on singularity.conf: %v", err)
+	}
+	if configFile.Sys().(*syscall.Stat_t).Uid != 0 {
```
cclerget: Should we check group too ?

cclerget (dismissed):
Need to move the code in `internal/pkg/runtime/engines/singularity/prepare.go` and execute check when `starterConfig.GetIsSUID()` returns true. Maybe consider additional checks on owner/group of `singularity` configuration directory, `capabilities` and `ecl.toml`

internal/app/singularity/actions.go
jmstover: Only ownership was checked in 2.x, and we're replicating that.  `src/util/suid.c` ~line 75

GodloveD: Well, just because we were only checking user ownership in 2.x doesn't mean we should only check the same here.  Seems sensible to check the group.  And probably not hard either right?

jmstover: As I told @ArangoGutierrez , hard-coding a group check for root will not make you fans with folks that change the perms like:
```
chmod 664 $sysconfdir/singularity/singularity.conf
chown root:singularity $sysconfdir/singularity/singularity.conf
```
So, the _group_ **singularity** can edit the `singularity.conf` file. Ownership of the file fine, but not group.

internal/pkg/runtime/engines/config/check.go
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) 2018, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package config
```
bauerm97: I don't think `internal/pkg/runtime/engines/config` is the right package for a utility function which is just checking the file ownership.

pkg/util/capabilities/file.go
```diff
@@ -11,6 +11,8 @@ import (
 	"io/ioutil"
 	"os"
 	"syscall"
+
+	"github.com/sylabs/singularity/internal/pkg/runtime/engines/config"
```
bauerm97: As I mentioned in the other comment, we shouldn't be importing `internal/pkg/runtime/...` from a package in `pkg/...`. This would (likely) break any external thing importing this package

bauerm97 (dismissed):

internal/pkg/runtime/engines/config/check.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"fmt"
+	"os"
+	"syscall"
+)
+
+// CheckUID for ownership of config file before reading it
+func CheckUID(filepath string) error {
+	fmt.Println(filepath)
```
jscook2345: Remove please.

bauerm97 (approved):

DrDaveD:
Will this break unprivileged installations?  It looks like it, I need to test.

DrDaveD:
It not only breaks unprivileged installations, but even running a root-installed installation with the -u option triggers the message
```
FATAL:   container creation failed: Unable to parse singularity.conf file: /etc/singularity/singularity.conf must be owned by root
```

PR #2575, opened 2019-01-20
Rework segmentation between darwin/linux

### LINKED ISSUE #2582: build fails for macos build
## Version of Singularity:

3.0.3

### Expected behavior

On MacOS, `make` OK

### Actual behavior

On MacOS, `make` fails:

 GO singularity
    [+] GO_TAGS "containers_image_openpgp"
../internal/pkg/runtime/engines/oci/process.go:29:2: build constraints exclude all Go files in /Users/justin/go/src/github.com/sylabs/singularity/pkg/util/rlimit
../internal/pkg/runtime/engines/oci/create.go:22:2: build constraints exclude all Go files in /Users/justin/go/src/github.com/sylabs/singularity/pkg/util/sysctl
make: *** [singularity] Error 1

Boot a MacOS system or image, run `./mconfig && make`

cclerget:

This PR segment build between darwin and linux, since starter is usable only on linux systems.

Also add two `mconfig` options :
- `--without-network` to not install/compile network plugins
- `--without-suid` to not install Singularity setuid binary

Also fixes #2582 


internal/app/singularity/capability.go
```diff
@@ -3,7 +3,7 @@
 // LICENSE.md file distributed with the sources of this project regarding your
 // rights to use or distribute this software.
 
-// +build singularity_runtime
+// +build linux
```
tri-adam: Would it be better to use file selection here (rename to `capability_linux.go`) here and in the other ~10 files with `+build linux`?

tri-adam (commented):
Just one idea...

tri-adam (approved):
This looks great to me

mlocal/frags/Makefile.stub
```diff
@@ -96,24 +100,34 @@ actions_LIST := $(SOURCEDIR)/etc/actions/exec $(SOURCEDIR)/etc/actions/run $(SOU
...
+CLEANFILES += $(starter)
+INSTALLFILES += $(starter_INSTALL) $(sessiondir) $(dist_bin_SCRIPTS_INSTALL) \
+                $(capability_JSON) $(syecl_config_INSTALL) $(actions_INSTALL) $(seccomp_profile_INSTALL) \
+                $(NVIDIA_liblist_INSTALL) $(cgroups_config_INSTALL)
+ALL += $(starter)
+
+endif # build_runtime
 
-all: cscope collect $(libruntime) $(go_BIN) $(config) $(starter) $(bash_completion) cniplugins
+ifeq ($(WITH_NETWORK),1)
```
bauerm97, line 116: I'm still questioning if this is the best approach. I don't think putting conditionals into our Makefile is necessarily a great idea. There is already a paradigm of `mconfig` conditionally adding contents to the final Makefile based on certain criteria. For example, `go_runtime_opts.mk` already exists to add specific options for when we are compiling with the runtime. Previously that file was being used to set go build tags for `runtime` build tag. I think it would be cleaner if we moved the code within these conditionals into their own file analogous to `go_runtime_opts.mk` and include those stubs in the final Makefile if appropriate.

For a naming scheme, I would suggest the following:

`Makefile_runtime.stub` - All rules for runtime
`Makefile_suid.stub` - All rules for including suid binary
`Makefile_network.stub` - All rules for CNI plugins

I think that this approach is more in line with the intention of `makeit`

bauerm97 (dismissed):

cmd/starter/c/starter.c
```diff
@@ -1,5 +1,5 @@
 /*
-  Copyright (c) 2018, Sylabs, Inc. All rights reserved.
+  Copyright (c) 2018-2019, Sylabs, Inc. All rights reserved.
```
WestleyK, line 2: Shouldn't this be: 

```suggestion
  Copyright (c) 2017-2019, Sylabs, Inc. All rights reserved.
```

PR #2578, opened 2019-01-21
Cache List, & Cache Clean

WestleyK:
# Description of the Pull Request (PR):

This PR will add a `singularity cache` command, this will be used to manage your local Singularity cache. (`${HOME}/.singularity/cache`)

### List Cache:

```bash
$ singularity cache list
ubuntu_latest.sif      2019-01-31 14:59:32    28.11 Mb         library
ubuntu_18.04.sif       2019-01-31 14:58:44    27.98 Mb         library
alpine_latest.sif      2019-01-31 14:58:24    2.18 Mb          library
centos_latest.sif      2019-01-31 14:59:07    72.96 Mb         library
centos_latest.sif      2019-01-31 14:59:26    73.45 Mb         oci
ubuntu_18.04.sif       2019-01-31 14:58:58    27.99 Mb         oci
ubuntu_latest.sif      2019-01-31 14:59:41    27.99 Mb         oci
alpine_latest.sif      2019-01-31 14:58:30    2.72 Mb          oci

There are 12 oci blob file(s) using 110.22 Mb of space. Use: '-T=blob' to list
```

<br>

### Clean only specific type of cache:

```bash
# clean only library cache
$ singularity cache clean --type=library

# clean only oci cache
$ singularity cache clean --type=oci

# clean only blob cache
$ singularity cache clean --type=blob

# clean only library, and oci cache
$ singularity cache clean --type=library,oci
```

*You can also replace the `clean`, with `list`, that will list the cache instead of cleaning it.*

<br>

### Clean cache with a specific name:

```bash
$ singularity cache clean --name alpine_latest.sif
```

And yes, the above command is compatible with `--type=library/oci`.

<br>

### Clean Cache:

```bash
# clean only blob cache (default)
$ singularity cache clean

# clean all cache
$ singularity cache clean --all
```

*There will be no output from the command.*

<br>


internal/pkg/cachecli/cache_list.go
```diff
@@ -0,0 +1,183 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"github.com/sylabs/singularity/internal/pkg/client/cache"
+	"github.com/sylabs/singularity/internal/pkg/sylog"
+)
+
+func join(strs ...string) string {
```
ikaneshiro: I'd recommend using `filepath.Join()` for the file path string concatenation and the `+` operator or `strings.Join()` for regular string concatenation instead of a custom function

cmd/internal/cli/cache.go
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2017-2019, Sylabs Inc. All rights reserved.
```
ikaneshiro: Do any of these files contain content for 2017? The only one that seems like it might is `docs/content.go`. The rest appear to be brand new or from 2018

internal/pkg/cachecli/cache_clean.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	return foundMatch
+}
+
+// CleanCacheName : clean a cache with a specific name (cacheName). if libraryCache == true; search only the
+// library. if ociCache == true; search the oci cache. returns false if no match found.
+func CleanCacheName(cacheName string, libraryCache, ociCache bool) bool {
+	foundMatch := false
+	if libraryCache == true {
```
ikaneshiro: If you want only one cache to be searched given one of the bools being true, id recommend a
```
if {}
else if {}
else {}
```

internal/pkg/cachecli/cache_clean.go
```diff
@@ -0,0 +1,182 @@
...
+		foundMatch = cleanLibraryCache(cacheName)
```
ikaneshiro: I'd recommend just returning at this point of nothing else is needed to be done. 
E.g. `return cleanLibraryCache(cacheName)`

internal/pkg/cachecli/cache_clean.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	if ociCache == true {
+		foundMatch = cleanOciCache(cacheName)
+	}
+	if libraryCache != true && ociCache != true {
+		foundMatch = cleanLibraryCache(cacheName)
+		foundMatch = cleanOciCache(cacheName)
```
ikaneshiro: If you have a case where `cleanLibraryCache()` returns `true` and then `cleanOciCache()` returns `false`, do you want the function to return `false`? If not I think using intermediate values and then `or`ing them together for the result is the way to go

internal/pkg/cachecli/cache_clean.go
```diff
@@ -0,0 +1,182 @@
...
+			os.Exit(0)
+		}
+		return nil
+	} else if len(cacheName) >= 1 && allClean == true || len(typeNameClean) >= 1 && allClean == true {
+		sylog.Fatalf("These flags are not compatible with each other")
+		os.Exit(2)
+	}
+
+	if allClean == true {
+		err = cache.Clean()
```
ikaneshiro: I think you should check for an error and return it within each of these if statements since the next statement could overwrite the error

internal/pkg/cachecli/cache_clean.go
```diff
@@ -0,0 +1,182 @@
...
+
+	return err
+
+}
+
+func cleanLibraryCache(cacheName string) bool {
+	foundMatch := false
+	libraryCacheFiles, err := ioutil.ReadDir(cache.Library())
+	if err != nil {
+		sylog.Fatalf("Failed while opening cache folder: %v", err)
```
ikaneshiro: Generally it's a better practice to return an `error` to be handled by the caller instead of calling `Sylog.Fatalf` since it gives the calling function the chance to handle the error how it wants. Eg. ignoring it, performing some kind of cleanup, etc.

ikaneshiro (dismissed):
Hey @WestleyK! This looks like it's coming along nicely. Left some comments to address.

internal/pkg/cachecli/cache_clean.go
WestleyK: Ahh, That makes more sense now that you explained it, I first thought it was weird passing error from function to function, but it makes sense now :+1:

internal/pkg/cachecli/cache_clean.go
WestleyK: Nice catch! Fixing that now :+1:

cmd/internal/cli/cache.go
WestleyK: Yeah, about that... I'm pretty sure @gmkurtzer wanted like this for simplicity, and consistency: `<year_of_project_started>-<current_year>` (2017-2019). But I'll will happily change it...

internal/pkg/cachecli/cache_list.go
```diff
@@ -0,0 +1,197 @@
...
+	ociList := false
+	blobList := false
+	listBlobSum := false
+
+	// split the string for each `,` then loop throught it and find what flags are there.
+	// then see whats true/false later. heres the benefit of doing it like this; if the user
+	// specified `library` twice, it will still only be printed once.
+	if len(typeNameList) >= 1 {
+		for _, nameType := range strings.Split(typeNameList, ",") {
+			if nameType == "library" {
```
bauerm97: I would replace these if-else blocks with a switch-case statement:

```
switch nameType {
    case "library":
    case "oci":
    [...]
    // `default:` signals what to do if none of the above were found, here it would be filling the function of your final else block.
    default: 
}
```

internal/pkg/client/cache/dir.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, Sylabs Inc. All rights reserved.
+// Copyright (c) 2017-2019, Sylabs Inc. All rights reserved.
```
bauerm97: I *think* this file should be `2018-2019`, since it didn't exist in 2017. @jmstover can you make sure that's correct

internal/pkg/cachecli/cache_clean.go
```diff
@@ -0,0 +1,221 @@
...
+	libraryClean := false
+	ociClean := false
+	blobClean := false
+
+	// split the string for each `,` then loop throught it and find what flags are there.
+	// then see whats true/false later. heres the benefit of doing it like this; if the user
+	// specified `library` twice, it will still only be printed once.
+	if len(typeNameClean) >= 1 {
+		for _, nameType := range strings.Split(typeNameClean, ",") {
+			if nameType == "library" {
```
bauerm97: Same comment about using switch-case statements as in the list code

internal/pkg/cachecli/cache_clean.go
```diff
@@ -0,0 +1,221 @@
...
+	for _, f := range blobs {
+		blob, err := ioutil.ReadDir(filepath.Join(cache.OciTemp(), f.Name()))
+		if err != nil {
+			sylog.Warningf("Unable to look in cache folder: %v", err)
+			return false, err
+		}
+		for _, b := range blob {
+			if b.Name() == cacheName {
+				sylog.Debugf("Removing: %v", filepath.Join(cache.OciTemp(), f.Name(), b.Name()))
+				err = os.RemoveAll(filepath.Join(cache.OciTemp(), f.Name(), b.Name()))
```
bauerm97: Here I think we need to make sure to remove the actual folder, not just the contents of it. So for example, I think we need to do `err = os.RemoveAll(filepath.Join(cache.OciTemp(), f.Name()))`, rather than only removing the block from within `f`.

internal/pkg/cachecli/cache_clean.go
```diff
@@ -0,0 +1,221 @@
...
+	for _, f := range libraryCacheFiles {
+		cont, err := ioutil.ReadDir(filepath.Join(cache.Library(), f.Name()))
+		if err != nil {
+			sylog.Warningf("Unable to look in cache folder: %v", err)
+			return false, err
+		}
+		for _, c := range cont {
+			if c.Name() == cacheName {
+				sylog.Debugf("Removing: %v", filepath.Join(cache.Library(), f.Name(), c.Name()))
+				err = os.RemoveAll(filepath.Join(cache.Library(), f.Name(), c.Name()))
```
bauerm97: Same comment here, it's necessary to do `err = os.RemoveAll(filepath.Join(cache.Library(), f.Name()))`

internal/pkg/cachecli/cache_clean.go
```diff
@@ -0,0 +1,221 @@
...
+			}
+		}
+	}
+
+	return foundMatch, nil
+}
+
+// CleanCacheName : clean a cache with a specific name (cacheName). if libraryCache == true; search only the
+// library. if ociCache == true; search the oci cache. returns false if no match found. Will return error if any occurs
+func CleanCacheName(cacheName string, libraryCache, ociCache bool) (bool, error) {
```
bauerm97: So, I think the name of this function is maybe slightly unclear. It should be clarified that this `cacheName` is actually the name of an image within the cache, not the name of which cache to go through.

docs/content.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, Sylabs Inc. All rights reserved.
+// Copyright (c) 2017-2019, Sylabs Inc. All rights reserved.
```
bauerm97: Same comment about copyright header here

internal/pkg/cachecli/cache_list.go
```diff
@@ -0,0 +1,197 @@
...
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/sylabs/singularity/internal/pkg/client/cache"
+	"github.com/sylabs/singularity/internal/pkg/sylog"
+)
+
+func findSize(size int64) string {
```
bauerm97: `func findSize(size int64) (string, error)`

internal/pkg/cachecli/cache_list.go
```diff
@@ -0,0 +1,197 @@
...
+		sizeF = float64(size) / 1000
+		return strings.Join([]string{fmt.Sprintf("%.2f", sizeF), " Kb"}, "")
+	} else if size <= 1000000000 {
+		sizeF = float64(size) / 1000000
+		return strings.Join([]string{fmt.Sprintf("%.2f", sizeF), " Mb"}, "")
+	} else if size >= 1000000000 {
+		sizeF = float64(size) / 1000000000
+		return strings.Join([]string{fmt.Sprintf("%.2f", sizeF), " Gb"}, "")
+	}
+	return "ERROR: failed to detect file size."
```
bauerm97: We shouldn't ever return error information encoded in a string return value. Once you change the function signature here according to the above comment, you can just do this:

`return "", fmt.Errorf("failed to detect file size")` And then we can determine how to handle a file whose size we cannot compute from wherever we call this function

bauerm97 (changes requested):
First round of feedback. I've left target comments at some sections of code here, but in this comment I'd like to suggest a bit of a larger change. First, I think that all of these files need to end with `_linux.go`. For now the cache isn't pertinent on MacOS, so we should add `_linux.go` to the ends of the file names to tell Go not to include those files on MacOS. Second, I think that `internal/pkg/cachecli/` isn't the appropriate place for those files. I think they belong in `internal/app/singularity/` package instead. So for example, `internal/pkg/cachecli/cache_clean.go` should end up here: `internal/app/singularity/cache_clean_linux.go`. Once we make that change and address the requested changes here, there will be a few more small changes to make before we can merge!

jscook2345:
@devprincess @WestleyK 

Probably can add to the user-docs now. I think the CLI interface looks solid.

@WestleyK 

Needs tests, but those can be added after the PR is merged. Ping me for that when this PR is merged. We can create an issue for the tests and work on them as a Pair or with feedback from me / singularity devs. Nice work on this feature!

internal/app/singularity/cache_list_linux.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	return nil
+}
+
+func listOciCache() error {
+	// loop thrught oci-tmp cache
+	ociTmp, err := ioutil.ReadDir(cache.OciTemp())
+	if err != nil {
+		sylog.Warningf("Unable to opening oci-tmp folder: %v", err)
+		return err
```
bauerm97: In general, we should follow the pattern of returning the information with the error. So instead of printing the Warning message here (imagine a scenario where the user is running --quiet), we should instead do something like:

`return fmt.Errorf("while opening oci-tmp folder: %v", err)`

I think there are a few other places in this PR where you should make that change!

internal/app/singularity/cache_list_linux.go
```diff
@@ -0,0 +1,222 @@
...
+				return err
+			}
+			if printList == true {
+				printFileSize, err := findSize(fileInfo.Size())
+				if err != nil {
+					// no need to describe the error, since it is already
+					sylog.Warningf("%v", err)
+				}
+				fmt.Printf("%-22.20s %-22s %-16s %s\n", b.Name(), fileInfo.ModTime().Format("2006-01-02 15:04:05"), printFileSize, "blob")
+				printFileSize = ""
```
bauerm97: I'm not quite sure why we're ding this assignment here. In Go, variable declaration/assignments are namespaced in a way. So, by doing `printFileSize, err := findSize(...)` inside of this `if printList` block, we don't have to re-assign it to an empty value. The `:=` assignment/declaration operator will assure that every time we enter into that if block, it will be a blank copy

internal/app/singularity/cache_list_linux.go
```diff
@@ -0,0 +1,222 @@
...
+			case "blob", "blobs":
+				blobList = true
+			case "all":
+				allList = true
+			default:
+				sylog.Fatalf("Not a valid type: %v", nameType)
+				os.Exit(2)
+			}
+		}
+	} else {
```
bauerm97: I'm confused at the purpose here. Are we saying that if `typeNameList == ""` it should list all?

bauerm97 (changes requested):

internal/app/singularity/cache_list_linux.go
WestleyK: Opps! I meant to delete this line :man_facepalming:

internal/app/singularity/cache_list_linux.go
WestleyK: Yes, if `typeNameList` is empty (`""`), then list `library`, and `oci` cache. Blobs are not listed by default (unless type blob is specified)

bauerm97: OK, so in this case I think there's a better approach for handling defaults. I will leave a few additional review comments for how we can more cleanly do this

cmd/internal/cli/cache_list_linux.go
```diff
@@ -0,0 +1,56 @@
...
+	"os"
+
+	"github.com/spf13/cobra"
+	"github.com/sylabs/singularity/docs"
+	"github.com/sylabs/singularity/internal/app/singularity"
+	"github.com/sylabs/singularity/internal/pkg/sylog"
+)
+
+var (
+	typeNameList string
```
bauerm97: `cacheListTypes []string`

cmd/internal/cli/cache_list_linux.go
```diff
@@ -0,0 +1,56 @@
...
+	allList      bool
+)
+
+func init() {
+	CacheListCmd.Flags().SetInterspersed(false)
+
+	CacheListCmd.Flags().StringVarP(&typeNameList, "type", "T", "", "list cache type, choose between: library, and oci")
```
bauerm97: `CacheListCmd.Flags().StringSliceVarP(&cacheListTypes, "type", "T", []string{"library", "oci"}, "list cache type, choose between: library, oci, and blob")`

There are two things that we're doing here which improve this code:

1. We're using `pflag` to convert whatever argument was passed to `--type <...>` and convert it into a `[]string`. It's a convenient way to avoid parsing the arguments to a flag
2. We're using `pflag` to set a default value to the flag. This way, we don't have to manually check that the `len(typeNameList) > 0`, we simply just assume that it's been set. If the user *doesn't* specify `--type <...>`, then `cacheListTypes` will be set to `[]string{"library", "oci"}`. Again, this is both a convenience thing but also it will clean up unnecessary conditionals from our code

internal/app/singularity/cache_list_linux.go
```diff
@@ -0,0 +1,209 @@
...
+			sylog.Warningf("%v", err)
+		}
+		fmt.Printf("\nThere are %d oci blob file(s) using %v of space. Use: '-T=blob' to list\n", count, printFileSize)
+	}
+	return nil
+}
+
+// ListSingularityCache : list local singularity cache, typeNameList : is a string of what cache
+// to list (seprate each type with a comma; like this: library,oci,blob) allList : force list all cache.
+func ListSingularityCache(typeNameList string, allList bool) error {
```
bauerm97: `func ListSingularityCache(cacheListTypes []string, listAll bool) error {`

internal/app/singularity/cache_list_linux.go
```diff
@@ -0,0 +1,209 @@
...
+	libraryList := false
+	ociList := false
+	blobList := false
+	listBlobSum := false
+
+	// split the string for each `,` then loop throught it and find what flags are there.
+	// then see whats true/false later. heres the benefit of doing it like this; if the user
+	// specified `library` twice, it will still only be printed once.
+	if len(typeNameList) >= 1 {
```
bauerm97: `for _, t := range typeNameList {`

internal/app/singularity/cache_list_linux.go
bauerm97: We can now just remove this `} else {...}`

bauerm97 (changes requested):
So, I only left specific suggestions this review about the `cache list` command, but I think you should apply the same idea for `cache clean` command as well.

internal/app/singularity/cache_list_linux.go
```diff
@@ -0,0 +1,207 @@
...
+func ListSingularityCache(cacheListTypes []string, listAll bool) error {
+	libraryList := false
+	ociList := false
+	blobList := false
+	listBlobSum := false
+
+	// split the string for each `,` then loop throught it and find what flags are there.
+	// then see whats true/false later. heres the benefit of doing it like this; if the user
+	// specified `library` twice, it will still only be printed once.
+	if len(cacheListTypes) >= 1 {
```
bauerm97: I don't think we need this conditional block here. `cacheListTypes []string` will always have the default which we set in the flag variable.

internal/pkg/client/cache/dir.go
```diff
@@ -41,9 +41,16 @@ func Root() string {
...
+// Clean : wipes all files in the cache directory, will return a error if one occurs
+func Clean() error {
+	sylog.Debugf("Removing: %v", Root())
+
+	err := os.RemoveAll(Root())
+	if err != nil {
+		sylog.Warningf("Unable to clean cache: %v", err)
+	}
+
+	return err
```
bauerm97: Maybe just:

```
if err := os.RemoveAll(Root()); err != nil {
    return fmt.Errorf("while cleaning cache: %s", err)
}

return nil
```

internal/app/singularity/cache_clean_linux.go
```diff
@@ -0,0 +1,212 @@
...
+// cache with that name.
+func CleanSingularityCache(cleanAll bool, cacheCleanTypes []string, cacheName string) error {
+	libraryClean := false
+	ociClean := false
+	blobClean := false
+
+	// split the string for each `,` then loop throught it and find what flags are there.
+	// then see whats true/false later. heres the benefit of doing it like this; if the user
+	// specified `library` twice, it will still only be printed once.
+	if len(cacheCleanTypes) >= 1 {
```
bauerm97: Same comment here about not needing to check the length of `cacheCleanTypes`

internal/app/singularity/cache_clean_linux.go
```diff
@@ -0,0 +1,212 @@
...
+		match, err = cleanOciCache(cacheName)
+		if err != nil {
+			return false, err
+		}
+		return match, nil
+	}
+	return false, nil
+}
+
+var err error
```
bauerm97: Not quite sure why this is here but it probably shouldn't be

internal/app/singularity/cache_list_linux.go
```diff
@@ -0,0 +1,207 @@
...
+			default:
+				sylog.Fatalf("Not a valid type: %v", t)
+				os.Exit(2)
+			}
+		}
+	}
+
+	fmt.Printf("%-22s %-22s %-16s %s\n", "NAME", "DATE CREATED", "SIZE", "TYPE")
+
+	if listAll == true {
```
bauerm97: I think we need to remove a bit of the complexity from the code here. There's no reason we need to have multiple calls here. I would propose to delete the entire `if listAll {...}` block. We can handle the `listAll` case in the next set of conditionals.

internal/app/singularity/cache_list_linux.go
```diff
@@ -0,0 +1,207 @@
...
+		if err != nil {
+			return err
+		}
+		err = listBlobCache(true)
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+	if libraryList == true {
```
bauerm97: `if libraryList || listAll {`

internal/app/singularity/cache_list_linux.go
```diff
@@ -0,0 +1,207 @@
...
+		err = listLibraryCache()
+		if err != nil {
+			return err
+		}
+	}
+	if ociList == true {
```
bauerm97: `if ociList || listAll {`

internal/app/singularity/cache_list_linux.go
```diff
@@ -0,0 +1,207 @@
...
+		err = listOciCache()
+		if err != nil {
+			return err
+		}
+	}
+	if blobList == true {
```
bauerm97: `if blobList || listAll {`

internal/app/singularity/cache_list_linux.go
```diff
@@ -0,0 +1,207 @@
...
+		err = listBlobCache(true)
+		if err != nil {
+			return err
+		}
+	}
+	if listBlobSum == true {
```
bauerm97: `if listBlobSum {`

bauerm97 (changes requested):

internal/app/singularity/cache_list_linux.go
WestleyK: `listAll` Is suppose to list all cache types (library, oci, & blob), and also override any other option.
I'm not sure how `if libraryList || listAll {...}` will list all cache if `listAll` is `true`... Maybe I'm missing something...

internal/app/singularity/cache_list_linux.go
WestleyK: Duh! Thanks!

internal/app/singularity/cache_list_linux.go
WestleyK: Wait! nevermined... I get it now, (sorry I'm a little slow sometimes)

internal/app/singularity/cache_clean_linux.go
```diff
@@ -0,0 +1,201 @@
...
+func CleanOciCache() error {
+	sylog.Debugf("Removing: %v", cache.OciTemp())
+
+	err := os.RemoveAll(cache.OciTemp())
+	if err != nil {
+		return fmt.Errorf("Unable to clean cache: %v", err)
+	}
+
+	return nil
+}
```
jchavezlavalle, line 38: So CleanLibraryCache, CleanOciCache and CleanBlobCache functions are pretty much the same code just differ on the type, why not having a cleanCache general function and removing it by type there? @WestleyK

internal/app/singularity/cache_clean_linux.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	return foundMatch, nil
+}
+
+func cleanOciCache(cacheName string) (bool, error) {
+	foundMatch := false
+	blobs, err := ioutil.ReadDir(cache.OciTemp())
+	if err != nil {
+		return false, fmt.Errorf("Unable to opening cache folder: %v", err)
```
jchavezlavalle: Would be nice to get error logs with the type in there as well: "Unable to open oci cache folder" for example, would make debugging easier @WestleyK

internal/app/singularity/cache_clean_linux.go
```diff
@@ -0,0 +1,201 @@
...
+		}
+		return nil
+	}
+
+	if cleanAll == true {
+		if err := cache.Clean(); err != nil {
+			return err
+		}
+	}
+	if libraryClean {
```
jchavezlavalle, line 202: @WestleyK If you create a general function "CleanCache" sending the type of the cache there, you would just need to verify which type of cache to delete inside that function and here the call would be just something like:

 ``` 
if err := CleanCache(cacheType); err != nil {
			return err
		}
```

File: internal/app/singularity/cache_list_linux.go [hunk 36]
```diff
@@ -0,0 +1,185 @@
...
+				sylog.Warningf("%v", err)
+			}
+			fmt.Printf("%-22s %-22s %-16s %s\n", c.Name(), fileInfo.ModTime().Format("2006-01-02 15:04:05"), printFileSize, "library")
+		}
+	}
+	return nil
+}
+
+func listOciCache() error {
+	// loop thrught oci-tmp cache
```
jchavezlavalle: thrught -> through

jchavezlavalle (changes requested):

internal/app/singularity/cache_clean_linux.go
bauerm97: Also, `fmt.Errorf(...)` should not start with an uppercase letter. [`unable ...` vs `Unable ...`]

jchavezlavalle (approved):

WestleyK:
Okay! I think I finally fixed all the requested changes! It should be ready for anther look when your ready!

internal/app/singularity/cache_clean_linux.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+	for _, f := range libraryCacheFiles {
+		cont, err := ioutil.ReadDir(filepath.Join(cache.Library(), f.Name()))
+		if err != nil {
+			return false, fmt.Errorf("unable to look in library cache folder: %v", err)
+		}
+		for _, c := range cont {
+			if c.Name() == cacheName {
+				sylog.Debugf("Removing: %v", filepath.Join(cache.Library(), f.Name(), c.Name()))
+				err = os.RemoveAll(filepath.Join(cache.Library(), f.Name(), c.Name()))
```
WestleyK, line 88: @bauerm97, I added back `c.Name()` here, because if you run: `singularity cache clean -T=library -N=ubuntu_latest.sif`, it would remove all `ubuntu_*.sif`.

But now it works fine, this may not be perfect, but it should be good enough...

bauerm97 (approved):

PR #2571, opened 2019-01-18
Help warning on run/exec -u should be on sandbox only

jchavezlavalle:
**Description of the Pull Request (PR):**

Warning on run and exec should be on sandbox only.

**This fixes or addresses the following GitHub issues:**

- Fixes #2557 

**Before submitting a PR, make sure you have done the following:**

- Read the [Guidelines for Contributing](../CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](../CHANGELOG.md) if necessary according to the [Contribution Guidelines](../CONTRIBUTING.md)
- Added tests to validate this PR and tested this PR locally with a `make testall`
- Based this PR against the appropriate branch according to the [Contribution Guidelines](../CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](../CONTRIBUTORS.md)

Attn: @singularity-maintainers

PR #2592, opened 2019-01-23
Update help for user namespace, non-setuid mode

DrDaveD:

This PR makes the explanation of the --userns option related to sandbox mode more clear, and adds a similar explanation by singularity.conf's allow setuid mode.

- Follow on to #2571 

### DIFF
--- a/internal/pkg/runtime/engines/singularity/data/singularity.conf
+++ b/internal/pkg/runtime/engines/singularity/data/singularity.conf
@@ -7,7 +7,8 @@
 # DEFAULT: yes
 # Should we allow users to utilize the setuid program flow within Singularity?
 # note1: This is the default mode, and to utilize all features, this option
-# will need to be enabled.
+# will need to be enabled.  For example, without this option loop mounts of
+# image files will not work; only sandbox images can work.
 # note2: If this option is disabled, it will rely on the user namespace
 # exclusively which has not been integrated equally between the different
 # Linux distributions.
--- a/internal/pkg/runtime/engines/singularity/data/singularity.conf
+++ b/internal/pkg/runtime/engines/singularity/data/singularity.conf
@@ -7,11 +7,12 @@
 # DEFAULT: yes
 # Should we allow users to utilize the setuid program flow within Singularity?
 # note1: This is the default mode, and to utilize all features, this option
-# will need to be enabled.  For example, without this option loop mounts of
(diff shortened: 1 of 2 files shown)


internal/pkg/runtime/engines/singularity/data/singularity.conf
```diff
@@ -7,7 +7,8 @@
 # DEFAULT: yes
 # Should we allow users to utilize the setuid program flow within Singularity?
 # note1: This is the default mode, and to utilize all features, this option
-# will need to be enabled.
+# will need to be enabled.  For example, without this option loop mounts of
+# image files will not work; only sandbox images can work.
```
WestleyK: Maybe change it a bit:

```
# must be enabled. For example, without this option enabled, loop mounts of
# containers will not work; only will work in sandbox containers.
```

GwendolynK-zz: must be enabled. For example, without this option enabled, loop mounts of
containers will not work; **loop mounts will** only ~will~ work in sandbox containers.

*What will work only in sandbox containers needs to be specified. Another alternative can be:*
"...;they will only work in sandbox containers." *or* "... loop mounts of containers will not work, unless they are in sandbox containers."

DrDaveD: @WestleyK I have no problem with changing "will need to be enabled" to "must be enabled" (although that wasn't something I had changed) but the second part doesn't make sense.  @GwendolynK's two alternate wordings make sense, but they are incorrect: loop mounts don't work non-setuid, period.  Sandbox images work without loopmounts.  I also deliberately said "can" instead of "will", because it depends on note2, that is, whether user namespaces are enabled on the system.

DrDaveD: I could say "only sandbox images can work (subject to note 2)."

DrDaveD: Or "only sandbox images, which do not use loop mounts, will work (subject to note 2)."

GwendolynK-zz: Both of those are great alternatives and will fix the issues I saw with the original sentence.

ArangoGutierrez (approved):
LGTM
Thanks @GwendolynK

PR #2606, opened 2019-01-26
Support passing mconfig prefix to rpmbuild

tcooper:

This PR propagates `PREFIX` supplied to mconfig through to the make rpm target with the translation of Makefile variables into `rpmbuild --define ...` statements and updates INSTALL.md to describe the usage with a specific example.

- None found

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ _The old changelog can be found in the `release-2.6` branch_
   - Add http/https protocols for singularity run/pull commands
   - Update to SIF 1.0.2
   - Add _noPrompt_ parameter to `pkg/signing/Verify` function to enable silent verification
+  - Minor change in RPM generation by `mconfig` allowing install to paths other than `/usr/local`
 
 # v3.0.2 - [2019.01.04]
 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -61,6 +61,7 @@
     - Sasha Yakovtseva <sasha@sylabs.io>, <sashayakovtseva@gmail.com>
     - Tarcisio Fedrizzi <tarcisio.fedrizzi@gmail.com>
     - Thomas Hamel <hmlth@t-hamel.fr>
+    - Trevor Cooper <tcooper@sdsc.edu>
     - Vanessa Sochat <vsochat@stanford.edu>
     - Westley Kurtzer <westley@sylabs.io>
     - Yannick Cote <y@sylabs.io>, <yhcote@gmail.com>
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -92,3 +92,12 @@ To build in a different folder and to set the install prefix to a different path
 ```
 $ ./mconfig -p /usr/local -b ./buildtree
 ```
+
+To set the install prefix to a different path and build an rpm on CentOS/RHEL use the following commands:
+
+```
+$ sudo yum install -y rpm-build
+$ cd $GOPATH/src/github.com/sylabs/singularity
+$ ./mconfig -p /usr/local
+$ make -C builddir rpm
+```
--- a/mlocal/frags/Makefile.stub
+++ b/mlocal/frags/Makefile.stub
@@ -289,7 +289,30 @@ testall: vendor-check check test
 .PHONY: rpm
 rpm: dist
 	@echo " BUILD RPM"
-	$(V)(cd $(SOURCEDIR) && rpmbuild $(RPMCLEAN) -ta $(SOURCEDIR)/singularity-$(SHORT_VERSION).tar.gz)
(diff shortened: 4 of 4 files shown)


bauerm97:
Hey @tcooper , thanks for the contribution. It appears that this PR is failing CI because of this:

```
2019/01/26 00:58:23 singularity is not installed on this system: exec: "singularity": executable file not found in $PATH
FAIL	github.com/sylabs/singularity/cmd/singularity	0.074s
```

Unfortunately I'm not entirely sure what could have broken this here. @jmstover or @DrDaveD could one of you comment on what behavior this change could have introduced which broke this?

DrDaveD:
This PR makes the default be to build the rpm so it installs everything into /usr/local, and /usr/local/bin is probably not in the PATH.  Not good.  The default behavior needs to be to install into the standard paths that rpms install into.

jmstover:
I agree with @DrDaveD . The default RPM macro values should be getting used in the `make rpm` target.

tcooper:
The Singularity instructions to create an rpm for CentOS/RHEL state that `./mconfig` should be run before `make rpm`.

Those instructions are followed within the same section of `INSTALL.md` with instructions for building a stable version of Singularity and for changing the install folder by using an install prefix with `./mconfig`.

This placement leads to confusion, specifically, that these options can be used together to create an rpm that installs to a location, specified by prefix, other than the default provided by `rpmbuild`.

Currently, `./mconfig` defines the default prefix to be `/usr/local`...
https://github.com/sylabs/singularity/blob/6cb5afbbcad959892f5bb8e507681f64d17c6e9c/mconfig#L377-L381
...while the default prefix used by `rpmbuild`, provided in `/usr/lib/rpm/macros` in CentOS, is `/usr`...
```
$ cat -n /usr/lib/rpm/macros | grep %_prefix -B4 -A4
   967	#==============================================================================
   968	# ---- configure macros.
   969	#	Macro(s) slavishly copied from autoconf's config.status.
   970	#
   971	%_prefix		/usr
   972	%_exec_prefix		%{_prefix}
   973	%_bindir		%{_exec_prefix}/bin
   974	%_sbindir		%{_exec_prefix}/sbin
   975	%_libexecdir		%{_exec_prefix}/libexec
```

Since the PR simply translates the Makefile variables created by `./mconfig` into `rpmbuild --define` statements for the rpm target of the Makefile the difference above is what is causing the diverging behavior when building an rpm with the default prefix provided by `./mconfig`.

I think changing `./mconfig` to use the same default prefix as `rpmbuild` (ie. `/usr` instead of `/usr/local`) would fix the discrepancy ***and*** make this PR function as expected when ***NOT*** supplying a new prefix.

**NOTE:** This may also require changes to the circleci and travis configuration as some builds specify `./mconfig -p /usr/local` while others **do not** specify a prefix explicitly.

If you do not want to change the default prefix for `./mconfig` I would suggest splitting `INSTALL.md` documentation for building rpms into a separate sub-section and explicitly mentioning the current inability to create an rpm that installs to a prefix other than the default specified.

DrDaveD:
I think too many people depend on the default prefix in mconfig and that it's not an option to change that.  

Instead, it could be done as a parameter to make -C builddir rpm, like PREFIX=/usr/local.  The problem is that mconfig currently not only sets PREFIX to /usr/local, but it also sets about 20 other variables based on the prefix set in mconfig, so setting just PREFIX is not currently enough.  This could be fixed by letting the default values of those other variables be based on make variables, so the evaluation of them is deferred until make time.  If they are given values explicitly in mconfig then pass those on to the Makefile, but let them default to be based on $(PREFIX).

DrDaveD:
Nope, I tried it, and it is still not enough.  The problem is that when we install in /usr/local, we want everything to be under there (e.g. /usr/local/etc/singularity) but when we install from an rpm we want some things to be under '/' (e.g. /etc/singularity) and some things under '/usr' (e.g. /usr/lib/singularity).  When you pass all the parameters through all the way from mconfig to rpm, there's no good place to introduce that distinction.

jmstover:
> there's no good place to introduce that distinction.

You would need to specify everything when you run `mconfig` ... i.e.
```
./mconfig --prefix=/usr --sysconfdir=/etc --localstatedir=/var
```

I see the `make rpm` target as just making a RPM based on what the default system RPM macros are. If you want different paths, then do:
```
./mconfig
make -C builddir/ dist
rpmbuild -tb --define '_prefix /somewhere/else' singularity-[ver].tar.gz
```

Or create a `${HOME}/.rpmmacros` to override the values. But not that we should pass the options from `mconfig` to `rpmbuild`.

tcooper:
We can (and will) build to get the behavior we want from an rpm instead of using the provided rpm target in the generated Makefile.

tcooper:
Capability suggested by this PR is provided by #2621. Closing.

PR #2613, opened 2019-01-28
Split Makefile.stub into smaller components based on function

bauerm97:

This splits `mlocal/frags/Makefile.stub` into smaller component pieces based on what they build. The idea behind this segmentation is to start forming a plan for letting `makeit` generate the Go build rules automatically, so eventually we will no longer need these `build_*.mk` files. The files added are:

- `build_cli.mk` - Rules for building the CLI binary, it's always included in the final Makefile
- `build_runtime.mk` - Rules for building `starter` binary, and various configs which starter needs
- `build_runtime_suid.mk` - Rules for installing `starter-suid`
- `build_network.mk` - Rules for building CNI networking plugins

@mikegray or @tri-adam can one of you test that it's working properly compiling on Darwin still? It *should* be, but just to be sure


cclerget (approved):
Tested. LGTM

PR #2642, opened 2019-02-01
Added check to verify cached file hash

SDkie:
Any interrupt in pull command will create an invalid cached file
Added check to verify cached file hash


internal/pkg/client/cache/library.go
```diff
@@ -29,12 +32,25 @@ func LibraryImage(sum, name string) string {
...
+	_, err := os.Stat(imagePath)
 	if os.IsNotExist(err) {
 		return false, nil
 	} else if err != nil {
 		return false, err
 	}
 
+	cacheSum, err := client.ImageHash(imagePath)
+	if err != nil {
+		sylog.Debugf("Error getting ImageHash - Deleting Cached File: %v", err)
```
ikaneshiro: This seems like an error that should be returned to the caller instead of just being part of debug output

internal/pkg/client/cache/library.go
```diff
@@ -29,12 +32,25 @@ func LibraryImage(sum, name string) string {
...
+		os.Remove(imagePath)
+		return false, nil
+	}
+	if cacheSum != sum {
+		sylog.Debugf("Cached File Sum(%s) and Expected Sum(%s) does not match - Deleting Cached File", cacheSum, sum)
+		os.Remove(imagePath)
```
ikaneshiro: I don't think we should be removing files from the cache in this function, just checking if something exists or not with the correct checksum. I think `cmd/internal/cli/pull.go` is a good place to remove/overwrite an invalid cache entry and it could be done simply by changing the force option of `client.DownloadImage()` to `true`

internal/pkg/client/cache/library.go
```diff
@@ -29,12 +32,25 @@ func LibraryImage(sum, name string) string {
...
 	if os.IsNotExist(err) {
 		return false, nil
 	} else if err != nil {
 		return false, err
 	}
 
+	cacheSum, err := client.ImageHash(imagePath)
+	if err != nil {
+		sylog.Debugf("Error getting ImageHash - Deleting Cached File: %v", err)
+		os.Remove(imagePath)
```
ikaneshiro: Probably shouldn't delete the image if we run into an error checking the hash

ikaneshiro (changes requested):
Hey @SDkie this is looking good. Just some comments about error handling and modifying the cache in the `LibraryImageExists()` function

internal/pkg/client/cache/library.go
SDkie: @ikaneshiro : whether failing to get ImageHash will fail the pull command itself?

SDkie:
@ikaneshiro: I made the changes you suggested. 
Also, the failure of the integration test is because of some network issue, I tested that locally.

internal/pkg/client/cache/library.go
```diff
@@ -28,13 +31,23 @@ func LibraryImage(sum, name string) string {
 }
 
 // LibraryImageExists returns whether the image with the SHA sum exists in the LibraryImage cache
-func LibraryImageExists(sum, name string) (bool, error) {
-	_, err := os.Stat(LibraryImage(sum, name))
+func LibraryImageExists(sum, name string) (bool, bool, error) {
```
ikaneshiro: I see how returning another bool to indicate if the file has the correct checksum, but I don't think its necessary for this function. I think we can just return `false` in the case that there is an image at the correct location and the hash doesn't match. This indicates that the image with the correct checksum doesn't exist in the cache regardless of if something with the correct filename exists in the cache. This will require the caller to know that there could be a file at that location in the cache even if this function returns `false`. This can be handled by having the force flag set when downloading an image into the cache, since it will either create a new file or overwrite an existing file with an incorrect checksum.

cmd/internal/cli/actions.go
```diff
@@ -132,11 +132,22 @@ func handleLibrary(u string) (string, error) {
...
+	if exists, validFile, err := cache.LibraryImageExists(libraryImage.Hash, imageName); err != nil {
 		return "", fmt.Errorf("unable to check if %v exists: %v", imagePath, err)
-	} else if !exists {
+	} else if !validFile {
 		sylog.Infof("Downloading library image")
-		libexec.PullLibraryImage(imagePath, u, "https://library.sylabs.io", false, authToken)
+
+		// Force Download if File already exists
+		force := exists
+		if err = library.DownloadImage(imagePath, u, PullLibraryURI, force, authToken); err != nil {
```
ikaneshiro: Here we can just hard code the `force` option to `true` since we always want to download an image in this location regardless of if its empty or there's an image there with an incorrect checksum. The same should apply to this function call in `pull.go` and `conveyorPacker_library.go`

ikaneshiro (changes requested):
@SDkie One more tweak and I think it will be good to go! With respect to having another function that supports cache downloads explicitly, we're currently trying to restructure the code base so that we can have something like that. With the code base as is, doing that in the the `libexec` package would cause a cyclic import, so for the moment I think its best to leave it in this state while we move code around.

SDkie:
@ikaneshiro : I have done the changes you suggested

cmd/internal/cli/actions.go
```diff
@@ -136,7 +136,15 @@ func handleLibrary(u string) (string, error) {
 		return "", fmt.Errorf("unable to check if %v exists: %v", imagePath, err)
 	} else if !exists {
 		sylog.Infof("Downloading library image")
-		libexec.PullLibraryImage(imagePath, u, "https://library.sylabs.io", false, authToken)
+		if err = library.DownloadImage(imagePath, u, PullLibraryURI, true, authToken); err != nil {
```
ikaneshiro: Sorry, one last nitpick. Lets leave libraryURL hardcoded as `https://library.sylabs.io` since action commands don't currently support a custom library url. Another option is to change line 127 to use `PullLIbraryURI`, but I think the former is clearer to a reader

ikaneshiro (changes requested):
Missed one small nitpick in my last review. Looks good to merge after this

ikaneshiro (approved):
LGTM. Will update with master once it passes CI and then merge

PR #2671, opened 2019-02-07
Fix /tmp and /var/tmp permissions when used with contain flag

cclerget:

See title


ikaneshiro (approved):
LGTM

internal/pkg/runtime/engines/singularity/container.go
```diff
@@ -1418,32 +1418,38 @@ func (c *container) addTmpMount(system *mount.System) error {
...
 				sylog.Warningf("Can't determine absolute path of workdir %s", workdir)
 			}
 
-			tmpSource = workdir + tmpSource
-			vartmpSource = workdir + vartmpSource
+			tmpSource = filepath.Join(workdir, tmpSource)
+			vartmpSource = filepath.Join(workdir, vartmpSource)
 
-			if err := fs.MkdirAll(tmpSource, 0755); err != nil {
+			if err := fs.Mkdir(tmpSource, os.ModeSticky|0777); err != nil {
```
bauerm97, line 1434: Why the change from `Mkdirall` -> `Mkdir`?

DrDaveD:
Looks good, but a question though: what difference does it make, since a singularity container is limited to a single user, the tmp and var/tmp directories are private, and the space isn't accessible outside the container?

internal/pkg/runtime/engines/singularity/container.go
cclerget: To avoid sticky bit on eventually create parent directory and additionally in the context if user specify a workdir we can safely assume the directory exists, so I changed for `Mkdir`

cclerget:
@DrDaveD I encountered an issue with an application complaining about `/tmp` and `/var/tmp` not containing sticky bit and not world writable

PR #2672, opened 2019-02-07
Create OCI bundle from SIF

### LINKED ISSUE #2637: Re-impliment mount command (root only)
This will allow sif files to be mounted via singularity exposing the file system as a directory.

cclerget:

This PR adds `ocibundle` public package to create OCI bundle from SIF image

- Fixes #2637 


pkg/ocibundle/sif_driver.go
```diff
@@ -0,0 +1,109 @@
...
+	// Umount rootfs
+	if err := syscall.Unmount(tools.RootFs(s.bundlePath).Path(), syscall.MNT_DETACH); err != nil {
+		return err
+	}
+	// delete bundle directory
+	return tools.DeleteBundle(s.bundlePath)
+}
+
+// FromSif ...
+func FromSif(image, bundle string, writable bool) (Driver, error) {
```
sashayakovtseva: Maybe NewDriver?

pkg/ocibundle/tools/loop.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/sylabs/singularity/pkg/util/loop"
+)
+
+// CreateLoop ...
+func CreateLoop(file *os.File, offset, size uint64) (string, error) {
```
sashayakovtseva, line 17: Do you think all SIF specific part should also go here? Here is not much point in having this function that already accepts everything. Instead it can accept `sif *os.File` and then load it and do all necessary stuff to find offset & size that is done in `Create()` now.

pkg/ocibundle/tools/oci.go
```diff
@@ -0,0 +1,55 @@
...
+	return filepath.Join(string(r), "rootfs")
+}
+
+// CreateBundle ...
+func CreateBundle(bundlePath string, config *specs.Spec) error {
+	var err error
+	var g generate.Generator
+
+	oldumask := syscall.Umask(0)
+	syscall.Umask(oldumask)
```
sashayakovtseva: defer?

pkg/ocibundle/tools/oci.go
```diff
@@ -0,0 +1,55 @@
...
+	"runtime"
+	"syscall"
+
+	"github.com/opencontainers/runtime-spec/specs-go"
+
+	"github.com/opencontainers/runtime-tools/generate"
+)
+
+// RootFs is the default root path for OCI bundle
+type RootFs string
```
sashayakovtseva, line 21: why need a separate type?

pkg/ocibundle/sif_driver.go
cclerget: Because SIF is one driver, we may want to generate OCI bundle from other source

pkg/ocibundle/tools/loop.go
cclerget: This function is generic all related stuff to SIF goes into sif_driver

pkg/ocibundle/tools/oci.go
cclerget: Oops :)

pkg/ocibundle/tools/oci.go
cclerget: A convenient way to append `rootfs` for a bundle

pkg/ocibundle/sif_driver.go
sashayakovtseva: we can define Driver interface somewhere and then add subdirectories with implementation which is a common design. E.g
```
ocibundle/
    driver/
        driver.go
        sif/
            sif_driver.go <- NewDriver() here, so it is clear we want sif
        other/
            other_driver.go
```

cclerget: I'm a bit concerned by `NewDriver` name because each driver may not require same arguments and I find that rather confusing to have same function name taking different arguments or type, additionally this is more explicit to have `ocibundle.FromSif(image, bundle)` instead of `ocibundle.NewDriver(image, bundle)` or `ocibundleSif.NewDriver(image, bundle)` depending of import name conflicts. Overall `Driver` is not well named and not descriptive, `Bundle` looks better

pkg/ocibundle/tools/overlay.go
```diff
@@ -0,0 +1,79 @@
...
+		return err
+	}
+	// best effort to cleanup mount
+	defer func() {
+		if err != nil {
+			syscall.Unmount(overlayDir, syscall.MNT_DETACH)
+		}
+	}()
+
+	if syscall.Mount("", overlayDir, "", syscall.MS_REMOUNT|syscall.MS_BIND, ""); err != nil {
```
bauerm97: `if err = syscall.Mount(...)`

pkg/ocibundle/tools/overlay.go
```diff
@@ -0,0 +1,79 @@
...
+		return err
+	}
+
+	upperDir := filepath.Join(overlayDir, "upper")
+	if err = os.Mkdir(upperDir, 0700); err != nil {
+		return
+	}
+	workDir := filepath.Join(overlayDir, "work")
+	if err = os.Mkdir(workDir, 0700); err != nil {
+		return
```
bauerm97: I am not a fan of these empty `if err ! = nil { return err }` type error passthroughs. If these ever error and we need to debug I have a feeling it will be incredibly hard to debug this because we don't know where the error is definitely coming from. I would much prefer: `return fmt.Errorf("while making overlay workDir (%s): %s", workDir, err)`

returning errors in this way should be much easier too debug should that need arise

pkg/ocibundle/tools/loop.go
```diff
@@ -0,0 +1,32 @@
...
+		Shared:         true,
+		Info: &loop.Info64{
+			SizeLimit: size,
+			Offset:    offset,
+			Flags:     loop.FlagsAutoClear,
+		},
+	}
+	idx := 0
+	if err := loopDev.AttachFromFile(file, os.O_RDONLY, &idx); err != nil {
+		return "", err
```
bauerm97: Same comment about error return

bauerm97 (changes requested):
Just some small tweaks to the SIF -> OciBundle portion of the code and I think we're good. I also appreciate moving the rest of the OCI code into `internal/app/singularity`, it's looking cleaner by the day :-)

bauerm97 (approved):

Issue #109, opened 2016-07-04
Support Singularity on Mac?


gmkurtzer:
Is this possible?

https://github.com/mist64/xhyve

nl384:
Are you still interested in this? I got xhyve working on my mac, and singularity seems to work fine on it too.

aculich:
@nl384 +1

gmkurtzer:
Whaaaaahhhhh? Ok, yes. Very interested! Do you have an idea on how to incorporate Xhyve directly into Singularity?

nl384:
Well I ran an Ubuntu image on my mac using Xhyve by following the instructions I found here : http://www.pagetable.com/?p=831 . And then I installed Singularity. I was able to create, bootstrap and shell into an Ubuntu image, so it seems to work fine. However, I'm still not very familiar with Xhyve and not exactly sure on how it would be directly incorporated.

gmkurtzer:
@nl384 We should circle back on this when you have time and see if we can spin out a SingularityWare project based on Xhyve and native Mac support.

fmorency:
+1!! Would love to see Singularity on xhyve for macOS and BSD.

EricEisaman:
Will Singularity Desktop for MacOS require root access for users running Docker containers after the desktop application has already been installed on the system?  If so, I would love to use Singularity Desktop with students on school issued laptops.

jmstover:
@EricEisaman ... No it will not require root access. The initial root requirement is currently because we're installing to /usr/local, and the `singularity.conf` file needs to be root owned (that core check isn't modified).

EricEisaman:
@jmstover Thank you.

PR #2675, opened 2019-02-08
Enable execution on Mac OS X

jmstover:
This enables us to execute Singularity containers on Mac OS X.

This provides the needed Singularity code to run a container on Mac OS X. We are still limited in functionality, but with this we can accomplish:

```
singularity run lolcow.sif
singularity exec lolcow.sif pwd
singularity shell lolcow.sif
```

The users $HOME is _automatically_ made available to the container.

Finally...
Fixes: #109 


bilke:
@jmstover I tried this but when I run `singularity shell alpine.sif` I get the following error that `xhyve` is not available:

```
➜ singularity run alpine_latest.sif
FATAL:   Failed to find xhyve executable at /usr/local/libexec/singularity/vm/xhyve
```

How to install it? Via Homebrew? Thanks!

mem:
In order to get rid of the merges and make the changes easier to review, I rebased the ctmadison:testpr branch on top of current master. Did I nuke existing comments by doing that? I cannot find them anywhere :cry:

jmstover:
That ... I don't know what happened. And was the branch off? Earlier it was clean against current master. I had been pulling it in.

gmkurtzer:
@bilke Even after building Singularity on the Mac, you will need 3 additional pieces:

1. mksqhashfs (you can get this through brew)
2. Xhyve installed to the predefined place
3. a micro-OS, custom built to run Singularity images through the hypervisor also installed to the predefined location

At Sylabs, we just so happen to have these things, and will be giving them away as a packaged solution called "Singularity Desktop" which will be given away freely shortly. Stay tuned for the announcement (which just may happen at our Singularity User's Group event in San Diego in 2 weeks).

cmd/internal/cli/actions.go
```diff
@@ -185,6 +184,33 @@ func replaceURIWithImage(cmd *cobra.Command, args []string) {
 	return
 }
 
+// TODO: Let's stick this in another file so that that CLI is just CLI
```
jmstover: You wouldn't have access to the `IsSyOS` or `VM` variables if you moved it outside the CLI package would you?

bauerm97: Well two things here:

1. Technically you're correct here, you will lose access to package-wide variables when moving a function outside of that package. However, in its current state, `vmStarter` takes `cmd *cobra.Command` as an argument which *does* give access to that flag's value.
2. We probably want to move this function to `internal/app/singularity`, and remove `cmd *cobra.Command` as an argument. Ideally, we want to parse the CLI options which have been set in `cmd/internal/cli` into some well-defined set of arguments (see the capability code for an example of this design pattern).

mem: What I was thinking was creating a struct with the configuration from the command line, and passing that to a command that needs access to to that information. So, instead of having a bunch of global variables, there would be a single one that could be passed around as needed.

cmd/internal/cli/startvm_darwin.go
```diff
@@ -0,0 +1,129 @@
...
+	defArgs := []string{"-A", "-m", VMRAM, "-c", VMCPU, "-s", "0:0,hostbridge", "-s", hdString, "-s", "31,lpc", "-l", "com1,stdio"}
+
+	if len(BindPaths) > 8 {
+		sylog.Fatalf("Maximum of 8 bind mounts")
+	}
+
+	// Bind mounts
+	singBinds := []string{""}
+
+	slot := 5
```
mem, line 40: I can't find documentation for this. Is there something special about PCI slot 5 on xhyve?

cmd/internal/cli/startvm_darwin.go
```diff
@@ -0,0 +1,129 @@
...
+
+	usr, err := user.Current()
+	if err != nil {
+		sylog.Fatalf("Failed to get current user")
+	}
+
+	// Force $HOME to be mounted
+	// TODO: engineConfig.GetHomeSource() / GetHomeDest() -- should probably be used
+	homeSrc := usr.HomeDir
+	pciArgs := fmt.Sprintf("4:0,virtio-9p,home=%s", homeSrc)
```
mem, line 73: same thing here, is there something special about PCI slot 4?

cmd/internal/cli/startvm_darwin.go
```diff
@@ -0,0 +1,129 @@
...
+	homeBind := fmt.Sprintf("home:%s", homeSrc)
+	singBinds = append(singBinds, homeBind)
+
+	sylog.Debugf("PCI: %s", pciArgs)
+	defArgs = append(defArgs, "-s", pciArgs)
+
+	userInfo := fmt.Sprintf("%s:%s:%s", usr.Username, usr.Uid, usr.Gid)
+
+	if IsSyOS {
+		cliExtra = "syos"
```
mem, line 87: I want to make sure this is intentional: this is discarding cliExtra passed as an argument to startVM and replacing it with syos. If that's the case, a comment would help.

cmd/internal/cli/actions.go
bauerm97: @mem I think that's a great goal to address in another PR at some point down the line

cmd/internal/cli/actions.go
```diff
@@ -217,6 +184,33 @@ func replaceURIWithImage(cmd *cobra.Command, args []string) {
...
+func vmStarter(cmd *cobra.Command, image string, args []string) bool {
```
bauerm97: I'm not a huge fan of this design pattern of having a bool function which will sometimes return, but will other times never return.

cmd/internal/cli/actions.go
```diff
@@ -217,6 +184,33 @@ func replaceURIWithImage(cmd *cobra.Command, args []string) {
...
+	}
+
+	// since --syos is a boolean, it cannot be added to the above list
+	if IsSyOS && !VM {
+		sylog.Warningf("The --syos option requires a virtual machine, automatically enabling --vm option.")
+		cmd.Flag("vm").Value.Set("true")
+	}
+
+	if VM {
+		prepareVM(cmd, image, args)
```
bauerm97: Let's remove the call to `prepareVM` from this function

cmd/internal/cli/startvm_darwin.go
jmstover: No. I just chose `5` as the starting slot for the bind points

cmd/internal/cli/startvm_darwin.go
jmstover: Nope. It was me choosing `4` for the `$HOME` bind coming in.

cmd/internal/cli/startvm_darwin.go
jmstover: Yup ... You can replicate the `--syos` flag with:  `singularity exec [image] syos` ...  We're ignoring anything passed since we want a SyOS shell ... We aren't going into the image automatically here.

cmd/internal/cli/actions.go
```diff
@@ -217,6 +184,33 @@ func replaceURIWithImage(cmd *cobra.Command, args []string) {
 	return
 }
 
+// TODO: Let's stick this in another file so that that CLI is just CLI
+func vmStarter(cmd *cobra.Command, image string, args []string) bool {
+	// check if --vm-ram or --vm-cpu changed from default value
```
bauerm97, line 196: I think all of this code which is simply responsible for setting the value of `VM` should just live in it's own function. We should just be able to set the value of the flag. Then, back outside the function we'll be able to just use `is VM {...}`

cmd/internal/cli/startvm.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/spf13/cobra"
+	"github.com/sylabs/singularity/internal/pkg/sylog"
+)
+
+func prepareVM(cmd *cobra.Command, image string, args []string) {
```
bauerm97: I think this function can be renamed `execVM()` to better align with `execStarter()`. Also, let's move this into `actions.go`, since it's common across architectures.

cmd/internal/cli/actions_linux.go
```diff
@@ -34,6 +34,10 @@ import (
 
 // TODO: Let's stick this in another file so that that CLI is just CLI
 func execStarter(cobraCmd *cobra.Command, image string, args []string, name string) {
+	if vmStarter(cobraCmd, image, args) {
```
bauerm97: Let's remove this code. Leave `execStarter()` function as a function which is solely responsible for running the `starter` binary on linux platform. Likewise, `execVM()` can be responsible for starting the VM.

cmd/internal/cli/actions_darwin.go
```diff
@@ -11,4 +11,5 @@ import (
 
 // TODO: Let's stick this in another file so that that CLI is just CLI
 func execStarter(cobraCmd *cobra.Command, image string, args []string, name string) {
```
bauerm97, line 15: I would say, move the function `execStarter` here into `actions_unsupported.go` and have it be a `panic`.

cmd/internal/cli/actions.go
```diff
@@ -217,6 +184,33 @@ func replaceURIWithImage(cmd *cobra.Command, args []string) {
...
+
+	if VM {
+		prepareVM(cmd, image, args)
+	}
+
+	return VM
+}
+
 // ExecCmd represents the exec command
 var ExecCmd = &cobra.Command{
```
bauerm97, line 214: Following from the other comments, I think that the run/exec/shell Command variables should all have their `Run` fields be a function like this:

```
func ActionRun(cmd *cobra.Command, args []string) {
	a := append([]string{"/.singularity.d/actions/exec"}, args[1:]...)
	for _, flagName := range []string{"vm-ram", "vm-cpu"} {
		if flag := cmd.Flag(flagName); flag != nil && flag.Changed {
			// the user passed an option related to the VM...
			if !VM {
				// ... but didn't enable the VM itself
				cmd.Flag("vm").Value.Set("true")
			}
			break
		}
	}

	// since --syos is a boolean, it cannot be added to the above list
	if IsSyOS && !VM {
		sylog.Warningf("The --syos option requires a virtual machine, automatically enabling --vm option.")
		cmd.Flag("vm").Value.Set("true")
	}

	if VM {
		execVM()
	}

	execStarter()
}
```

On both `darwin` && `linux`, `execVM` will be defined (in fact they will both be defined here in `actions.go`). However, on all other platforms `execStarter` will be defined as `panic()`.

cmd/internal/cli/startvm_darwin.go
```diff
@@ -0,0 +1,129 @@
...
+	"runtime"
+	"strconv"
+	"strings"
+	"syscall"
+
+	"github.com/sylabs/singularity/internal/pkg/buildcfg"
+	"github.com/sylabs/singularity/internal/pkg/sylog"
+)
+
+func startVM(sifImage, singAction, cliExtra string, isInternal bool) error {
```
bauerm97: Eventually `startVM` [for all archs] should be moved to `internal/app/singularity`, but for now this is fine.

bauerm97 (changes requested):

cmd/internal/cli/actions_linux.go
mem: Let me give this a try

cmd/internal/cli/actions.go
```diff
@@ -225,6 +219,10 @@ var ExecCmd = &cobra.Command{
 	PreRun:                replaceURIWithImage,
 	Run: func(cmd *cobra.Command, args []string) {
 		a := append([]string{"/.singularity.d/actions/exec"}, args[1:]...)
+		if wantsVM(cmd) {
```
bauerm97: Let's move this into its own function. As noted previously, just create `func ActionRun` and then here do: 

```
Run: ActionRun
```

Same comment applies for all of these related commands

cmd/internal/cli/actions_darwin.go
```diff
@@ -11,4 +11,5 @@ import (
...
+	panic("starter is unsupported on this platform")
```
bauerm97, line 16: Change this file name to `actions_unsupported`, to handle windows/etc.. as well

cmd/internal/cli/actions.go
```diff
@@ -35,48 +35,15 @@ func init() {
 	// TODO : the next n lines of code are repeating too much but I don't
 	// know how to shorten them tonight
```
WestleyK: I think this `TODO :` is solved, it can probably be removed.

bauerm97 (changes requested):

cmd/internal/cli/actions.go
bauerm97: Actually, this is kinda dumb, I don't think that makes sense here.

cmd/internal/cli/actions.go
mem: Got it, done

bauerm97 (approved):

PR #2684, opened 2019-02-11
Fix terminal issue with oci commands

cclerget:

Pass terminal file descriptors only when terminal is false. And fix stream duplication when used with terminal is true.

### DIFF
--- a/internal/app/singularity/oci_run_linux.go
+++ b/internal/app/singularity/oci_run_linux.go
@@ -95,6 +95,9 @@ func OciRun(containerID string, args *OciArgs) error {
 	}
 
 	if err := attach(engineConfig, true); err != nil {
+		// kill container before deletion
+		sylog.Errorf("%s", err)
+		OciKill(containerID, "SIGKILL", 1)
 		return err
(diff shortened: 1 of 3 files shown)


internal/app/singularity/oci_run_linux.go
```diff
@@ -95,6 +95,9 @@ func OciRun(containerID string, args *OciArgs) error {
 	}
 
 	if err := attach(engineConfig, true); err != nil {
+		// kill container before deletion
+		sylog.Errorf("%s", err)
```
bauerm97, line 99: `sylog.Errof("Unable to attach to container: %s", err)`

bauerm97 (changes requested):
I think just a small tweak to error message then it's good from here

internal/app/singularity/oci_run_linux.go
cclerget: Error message become a bit redundant and overly long, there is an explicit error message coming from attach

bauerm97 (approved):

Issue #2268, opened 2018-10-19
Depending on git tags for the version number is problematic


DrDaveD:
## Version of Singularity:

master branch

I think that depending on the git tag for mconfig to find a version number is more trouble than it is worth.  First, it doesn't work when there is no git.  Even with git, tags are not normally synchronized when people do git fetch or git pull, you have to explicitly do 'git fetch upstream --tags'.  I was testing travis in my own repository and it failed in obscure ways until I realized that I had never done 'git push origin --tags'.  This will continue to be a pain for people because people are not used to it.

I think it would be better to just check in a VERSION file into git and manually update the version as part of the release process.  That would also significantly simplify the make-dist.sh script and probably other things as well.  The default version set by mconfig can still add in the git describe info if it is available.

0xaf1f:
> Even with git, tags are not normally synchronized when people do git fetch or git pull, you have to explicitly do 'git fetch upstream --tags'

That's not been my experience. The man page also says so:

> By default, any tag that points into the histories being fetched is also fetched; the effect is to fetch tags that point at branches that you are interested in. This default behavior can be changed by using the --tags or --no-tags options or by configuring remote.<name>.tagOpt. By using a refspec that fetches tags explicitly, you can fetch tags that do not point into branches you are interested in as well.

Tags aren't *pushed* by default though, which explains what you experienced here:

> I was testing travis in my own repository and it failed in obscure ways until I realized that I had never done 'git push origin --tags'.

DrDaveD:
That's interesting, I wasn't aware of that.  When I look at git fetch --help on el6 and el7 gits they both say essentially the same thing but in different ways that are easier to understand, for example on el6 the explanation of --tags says:

> Most of the tags are fetched automatically as branch heads are
> downloaded, but tags that do not point at objects reachable from
> the branch heads that are being tracked will not be fetched by this
> mechanism. This flag lets all tags and their associated objects be
> downloaded.

So say for example that there becomes a release v3.0.1 which is tagged on the release-3.0 branch, and somebody is only tracking the master branch, they will not see get the v3.0.1 tag.  That's still going to trip people up.

DrDaveD:
Also I have been in the habit lately of doing "git pull upstream master" to both fetch and merge in one step, and that does not fetch the tags from upstream.  This is not obvious from "git pull --help", but that's been my experience, and it is reproducible.  On master branch do:
```
$ git tag -d v3.0.0
$ git pull upstream master
$ git fetch upstream
[...]

PR #2692, opened 2019-02-11
Fix RPMPREFIX on el6 and el7

DrDaveD:

This fixes RPMPREFIX for el6 and el7

- Addresses post-merge comments on the PR that enabled the feature, #2621 

### DIFF
--- a/mlocal/frags/Makefile.stub
+++ b/mlocal/frags/Makefile.stub
@@ -55,6 +55,8 @@ rpm: dist
 	    rpmbuild $(RPMCLEAN) -ta \
 	      --define '_prefix $(RPMPREFIX)' \
 	      --define '_sysconfdir $(RPMPREFIX)/etc' \
+	      --define '_localstatedir $(RPMPREFIX)/var' \
+	      --define '_mandir $(RPMPREFIX)/share/man' \
 	      $(SOURCEDIR)/singularity-$(SHORT_VERSION).tar.gz; \
 	  else \
 	    rpmbuild $(RPMCLEAN) -ta \


hpcdevops:
@DrDaveD ...

I've confirmed that while this change does drop man pages in the `RPMPREFIX` on CentOS 6 a side effect is that man pages are no longer compressed by `/usr/lib/rpm/redhat/brp-compress` which expects them to be in a path starting with `./usr/...`.

The appears to be a quirk of CentOS 6 (inherited from Redhat) and I'm not sure it's worth quibbling over.

Another thing that *is* fairly consistent in the macros provided by the rpm RPM though is that `%_mandir` is almost always defined as `%{_datarootdir}/man` with `%_datarootdir` itself defined as `%{_prefix}/man`.

Perhaps it's better to define and pass down `%_datarootdir` to get more consistent behavior across OS'es that build RPMs?

As an aside...

I notice that you don't appear to have sync'd the 3.1.0-rc# tags to your fork. This causes auto-configuration of RPM `VERSION` (using `git describe...`) to be inconsistent with upstream.

Thanks for getting this updated... again.

DrDaveD:
@hpcdevops thanks for testing, and for noticing the bug in the first place.   Makefile.stub does still define _prefix, which as you pointed out is used by default in _datarootdir (although I think you meant %{_prefix}/share), so _datarootdir is covered for the general case.  In the specific case of el6, /usr/lib/rpm/redhat/macros defines _mandir explicitly not in terms of %{_datarootdir}, so that's why this PR needs to explicitly set _mandir.  Even on el7 that file defines _mandir directly in terms of _prefix, not _datarootdir.

Also, this branch was made after I updated to 3.1.0-rc1 yesterday, so for me git describe says v3.1.0-rc1-3-gaae37fb in the branch.  Do you have the 3.1.0-rc1 tag?  I learned the hard way that it's not sufficient to do "git pull upstream master" in your master branch, you have to do "git fetch upstream" to make sure you get all the latest tags.  I made issue #2268 because I thought syncing tags was even harder than that.

hpcdevops:
@DrDaveD ... you are, of course, correct about my incorrect definition of `%{_datarootdir}` in my previous comment. Bad copy/paste from test machine... 😉 

With regard to which tag is on the repo clone I'm using to build/test...

I'm using a Singularity image to build Singularity and within the image my runscript is cloning your repository in a clean environment and checking out the specified branch.

When I did the build test on my system in that manner my generated RPM had the following name `singularity-3.0.0-1352.gaae37fb.el6.x86_64.rpm`.

DrDaveD:
Ah, that's what you meant.  You didn't [pull in my PR](https://help.github.com/articles/checking-out-pull-requests-locally/) into a clone of the sylabs repository, which is what I usually do when I test a PR.  So the issue is that I didn't do 'git push --tags', which I don't usually do.  I did that now, so a rebuild should show up as 3.1.0-rc1.  But next time it will probably have the same issue because I don't normally keep tags up to date in my github repository.  You could probably work around it by pulling the tags from the sylabs upstream repository into your copy of my repository.

DrDaveD:
@GodloveD @jmstover this is important for 3.1.0 milestone if the release notes mention RPMPREFIX.

jmstover (approved):

PR #2704, opened 2019-02-13
Escape '%' with double '%%'

### LINKED ISSUE #2646: Errors on any incorrect `%`
## Version of Singularity:

```
3.0.1-752.g456a1a62
```

### Expected behavior:

Container builds without any errors.
Or at least warns you about invalid sections.

### Actual behavior:

The build fails, complaining about `invalid section(s)`:

```
FATAL:   Unable to create build: unable to parse spec test.def: While parsing definition: test.def: invalid section(s) specified: {hello, hello
```

*(The def file)*
```bootstrap: library
bootstrap: library
from: alpine:3.8

%runscript
cat /hello-world

%post
cat << EOF > /hello-world

USAGE:
    ./somthing.sh %{option}... %bar

%{hello world}
%hello world

EOF
```

### Steps to reproduce this behavior:

Try building a container with the provided def file.

WestleyK:

This PR will add a escape for `%`, by using two `%%`.

## Example:

```
$ cat test.def
bootstrap: library
from: alpine:3.8

%runscript
cat /hello-world

%post
cat << EOF > /hello-world

USAGE:
    ./somthing.sh %{option}... %bar

%%{hello world}
%%hello world

%%post

EOF
```

*Notice we are only escaping if the word start with `%`.*

Build the container:

```
# build succeeds
```

Testing it:

```
$ ./test.sif 

USAGE:
    ./somthing.sh %{option}... %bar

%{hello world}
%hello world

%post

```

<br>

If you tried to build the container before this PR, this will happen:

```
FATAL:   Unable to create build: unable to parse spec bar.def: While parsing definition: bar.def: invalid section(s) specified: {hello, hello
```

<br>

- Fixes #2646 

<br>

### DIFF
--- a/pkg/build/types/parser/deffile.go
+++ b/pkg/build/types/parser/deffile.go
@@ -82,8 +82,8 @@ func scanDefinitionFile(data []byte, atEOF bool) (advance int, token []byte, err
 			return 0, nil, err
 		}
 
-		// Check if the first word starts with % sign
-		if word != nil && word[0] == '%' && word[1] != '%' {
+		// Check if the first word starts with % sign. But if theres two ('%%'); then escape
+		if word != nil && word[0] == '%' && len(word) >= 2 && word[1] != '%' {
 			// If the word starts with %, it's a section identifier
 
 			// We no longer check if the word is a valid section identifier here, since we want to move to
@@ -105,12 +105,15 @@ func scanDefinitionFile(data []byte, atEOF bool) (advance int, token []byte, err
 			}
 		} else {
 			// This line is not a section identifier
-			if word != nil {
-				if line[1] == '%' {
+			// Check if theres two '%%', if there is; remove the first one
+			if len(line) >= 2 {
+				if line[0] == '%' && line[1] == '%' {
 					retbuf.Write([]byte(strings.TrimPrefix(string(line), "%")))
 				} else {
 					retbuf.Write(line)
 				}
+			} else {
+				retbuf.Write(line)
 			}
 			retbuf.WriteString("\n")			
 		}
--- a/pkg/build/types/parser/deffile.go
+++ b/pkg/build/types/parser/deffile.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 
 	"github.com/sylabs/singularity/pkg/build/types"
+	"github.com/sylabs/singularity/internal/pkg/sylog"
 )
 
 var (
@@ -82,7 +83,7 @@ func scanDefinitionFile(data []byte, atEOF bool) (advance int, token []byte, err
 		}
 
 		// Check if the first word starts with % sign
-		if word != nil && word[0] == '%' {
+		if word != nil && word[0] == '%' && word[1] != '%' {
 			// If the word starts with %, it's a section identifier
 
 			// We no longer check if the word is a valid section identifier here, since we want to move to
@@ -104,8 +105,14 @@ func scanDefinitionFile(data []byte, atEOF bool) (advance int, token []byte, err
 			}
 		} else {
 			// This line is not a section identifier
-			retbuf.Write(line)
-			retbuf.WriteString("\n")
+			if word != nil {
+				if line[1] == '%' {
+					retbuf.Write([]byte(strings.TrimPrefix(string(line), "%")))
+				} else {
+					retbuf.Write(line)
+				}
+			}
+			retbuf.WriteString("\n")			
 		}
 
 		// Shift the advance retval to the next line
@@ -147,7 +154,6 @@ func splitToken(tok string) (ident string, content string) {
 
 // parseTokenSection splits the token into maximum 2 strings separated by a newline,
 // and then inserts the section into the sections map
-//
 func parseTokenSection(tok string, sections map[string]string) error {
 	split := strings.SplitN(tok, "\n", 2)
 	if len(split) != 2 {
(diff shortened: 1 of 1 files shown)


ikaneshiro:
@WestleyK I think this is a good aproach, but I wonder if we can find a generic way to escape a line or block of text as well. For multistage builds I'd like to make `Boostrap` a reserved keyword so that we can easily break a definition into appropriate stages to be executed. So having a general way to escape a line will become necessary. Any thoughts on this? This may be out of scope of the topic of this PR so feel free to respond to this back on the original issue!

WestleyK:
Agree, we should be able to escape `bootstrap:`, `from:` etc... But check this example:

```bash
$ cat test.def 
bootstrap: library
from: westleyk/official/busybox:1.28

%help
hello world

%runscript
cat /hello-world

%post
cat << EOF > /hello-world

bootstrap: hello-world
from: fooball

EOF

INFO:    Starting build...
+ cat
INFO:    Adding help info
INFO:    Adding runscript
INFO:    Creating SIF file...

$ ./test.sif 

bootstrap: hello-world
from: fooball
```

The container builds fine with no escape for `bootstrap:`, and `from:`... Is this what you were considered about?

ikaneshiro:
@WestleyK I agree this isnt a problem with how we are parsing currently. At the moment words like `Bootstrap` and `From` arent reserved and we just look for a colon separated list to interpret as the header or a `%` whichever comes first.

My thoughts are looking ahead to a point when we could have multiple `bootstrap` headers in the file. Consider a case where we have multistage build implemented and have a definition like this:
```
bootstrap: library
from: westleyk/official/busybox:1.28
stage: one

%help
hello world

%runscript
cat /hello-world

%post
cat << EOF > /hello-world

bootstrap: hello-world
from: fooball

EOF

bootstrap:scratch

# copy file from first stage to final stage
%files from one
    /hello-world

%runscript
    cat /hello-world
```

In this case I'd like to use `bootstrap` as a reserved word so that we can easily split the second stage of building from the first. But that would also interpret the `bootstrap: hello-world` as a new definition too so escaping that as well would be nice. It's not a hard requirement as it would be simple to have something like `#!STAGE:` to break up the various stages but I think definitions will look nicer if we can use `bootstrap`

pkg/build/types/parser/deffile.go
```diff
@@ -104,7 +105,16 @@ func scanDefinitionFile(data []byte, atEOF bool) (advance int, token []byte, err
 			}
 		} else {
 			// This line is not a section identifier
-			retbuf.Write(line)
+			// Check if theres two '%%', if there is; remove the first one
+			if len(line) >= 2 {
+				if line[0] == '%' && line[1] == '%' {
+					retbuf.Write([]byte(strings.TrimPrefix(string(line), "%")))
```
mem, line 111: If you make the other change, this doesn't matter, but just for reference, you can save yourself some casting by using bytes.TrimPrefix instead:

    retbuf.Write(bytes.TrimPrefix(line, []byte("%")))

In general the strings and bytes packages mirror each other, one operates on strings and the other on []byte.

pkg/build/types/parser/deffile.go
```diff
@@ -104,7 +105,16 @@ func scanDefinitionFile(data []byte, atEOF bool) (advance int, token []byte, err
 			}
 		} else {
 			// This line is not a section identifier
-			retbuf.Write(line)
+			// Check if theres two '%%', if there is; remove the first one
+			if len(line) >= 2 {
```
mem, line 109: I think this block can be simplified a little:

```
if bytes.HasPrefix(line, []byte("%%") {
    line = line[1:]
}
retbuf.Write(line)
```

since line is a slice of byte, you can simply modify the slice to skip the first %.

GodloveD:
I have to say, I'm not crazy about this.  It feels to me like a hack just to cover over a problem with the parsing.  What if we actually parsed for expected keywords that begin with a `%` rather than just assuming that anything that starts with a `%` is automatically a keyword?

GodloveD:
@ikaneshiro this has gotten pretty stale.  Is it possible to just refine the parser so that it does not just look for percentage signs but actually parses for keywords?  There are only a handful of keywords right?  Would this approach fix the issue?

ikaneshiro:
@GodloveD That wont work because if someone did something like try to write a definition within the post section it would still try to parse the sections since they use the correct keywords.
```
%post
cat << EOF > /hello-world

bootstrap: hello-world
from: fooball

%post
echo hello

EOF
```

We need some sort of escape sequence. I think defining a generic way to do this is best, but `%%` is a very reasonable choice

sashayakovtseva:
Should we decide something about this one already?

GodloveD:
Haha.  Yeah we probably should.  Here's my take. 

So, when I use bash, I can't help but think that the way that strings and escape sequences and such are handled is just a series of hacks.  Like they started with quotes, and then they added single quotes because sometimes you want double quotes in your string, and then they added `/` for escapes, and then they thought "oh shoot!  I need an / in my string", so they added `//` and on and and on and on.  I'm not saying for certain this is what happened, but it just feels like a poor design decision led to another that mopped up 90% of corner cases and that led to another that cleaned up another 9% and another that cleaned up another 0.9%, etc. etc. etc.

And that is what this kindof feels like to me too.   What happens when we have a legit need for two %% and we don't want to use them as an escape sequence?  Is this going to continue to get weirder and weirder?  Just seems like maybe there is an underlying design decision we could make to fix this. 

Don't know what it is.  But just feels like there should be one.

sashayakovtseva:
Agreed. Maybe that build error in PR description can be handled on def file parsing side?

WestleyK:
This PR is very old, we/I will make another PR when requested/needed.

PR #2712, opened 2019-02-15
Warning When push/pull a unsigned container

### LINKED ISSUE #2584: Add Feedback on Unsigned Containers
The new SIF format in 3.0 allows for signed containers. In order to encourage good security practices, and make this functionality simpler to use, and, we could add feedback to certain commands:

* When `run`/`exec`/`shell` commands are run with an unsigned SIF, print a message/warning indicating the image is unsigned (and thus, it is not possible to verify its authenticity.)
* When `push` is run with an unsigned SIF, print a message/warning that the image is unsigned (and thus, other users will not be able to verify its authenticity.)
* When `pull` is run and an unsigned SIF is returned, print a message/warning that the image is unsigned (and thus, it is not possible to verify its authenticity.)

There are likely other commands that could provide this type of feedback. Suggestions?

WestleyK:

This PR will add a warning message when pushing/pulling a container thats not signed. This PR also updates the tests for the pull command.

This will only check if a container is signed when pulling from the library. Not `shub`, or `docker`.

### Example:

```bash
# pulling a unsigned container:
$ singularity pull library://ubuntu:latest
WARNING: sylabs-token files are deprecated. Use 'singularity remote' to manage remote endpoints and tokens.
WARNING: No default remote in use, falling back to: https://library.sylabs.io
 26.81 MiB / 26.81 MiB [=====================================] 100.00% 5.21 TiB/s 0s
WARNING: unable to verify container: error while searching for signature blocks: no signatures found for system partition
This image is not signed, and thus its contents cannot be verified.
Do you with to proceed? [N/y] n
Aborting.

# pulling a signed container:
$ singularity pull library://alpine:latest
WARNING: sylabs-token files are deprecated. Use 'singularity remote' to manage remote endpoints and tokens.
WARNING: No default remote in use, falling back to: https://library.sylabs.io
 2.59 MiB / 2.59 MiB [=======================================] 100.00% 7.39 TiB/s 0s
INFO:    Container is signed
	Sylabs Admin <support@sylabs.io>, Fingerprint 8883491F4268F173C6E5DC49EDECE4F3F38D871E

# now with '--allow-unauthenticated':
$ singularity pull --allow-unauthenticated library://ubuntu:latest
WARNING: sylabs-token files are deprecated. Use 'singularity remote' to manage remote endpoints and tokens.
WARNING: No default remote in use, falling back to: https://library.sylabs.io
 26.81 MiB / 26.81 MiB [=====================================] 100.00% 6.71 TiB/s 0s
WARNING: Skipping container verification
```

Now with the push command:

```bash
# unsigned container:
$ singularity push test.sif library://westleyk/foobar/bar:latest
WARNING: sylabs-token files are deprecated. Use 'singularity remote' to manage remote endpoints and tokens.
WARNING: No default remote in use, falling back to: https://library.sylabs.io
WARNING: unable to verify container: error while searching for signature blocks: no signatures found for system partition
INFO:    TIP: Learn how to sign your own containers here : https://www.sylabs.io/docs/

Your container is **NOT** signed! You REALLY should sign your container before pushing!
Do you really want to continue? [N/y] y
INFO:    Now uploading test.sif to the library
 2.59 MiB / 2.59 MiB [=======================================] 100.00% 5.30 TiB/s 0s
INFO:    Setting tag latest

# signed container:
$ singularity push test.sif library://westleyk/foobar/barfoo:latest
WARNING: sylabs-token files are deprecated. Use 'singularity remote' to manage remote endpoints and tokens.
WARNING: No default remote in use, falling back to: https://library.sylabs.io
INFO:    Container is signed
	Westley (test) <westley@sylabs.io>, Fingerprint 93EE8637AF8AF616AD54EFFF25C5D19789677B67
INFO:    Now uploading test.sif to the library
 2.59 MiB / 2.59 MiB [=======================================] 100.00% 4.92 TiB/s 0s
INFO:    Setting tag latest

# and with '--allow-unauthenticated':
$ singularity push --allow-unauthenticated test.sif library://westleyk/foobar/foofoo:latest
WARNING: sylabs-token files are deprecated. Use 'singularity remote' to manage remote endpoints and tokens.
WARNING: No default remote in use, falling back to: https://library.sylabs.io
WARNING: Skipping container verifying
INFO:    Now uploading test.sif to the library
 2.59 MiB / 2.59 ===========================================] 100.00% 6.05 TiB/s 0s
INFO:    Setting tag latest
```

<br>

- Fixes #2584 

<br>

**This PR must be merged after https://github.com/sylabs/singularity/pull/3167.**

<br>


tri-adam:
Removed PR from Key Service GA milestone. Milestones should be estimated *issues* only. Otherwise we will be duplicating estimates, and the release timing cannot be accurately predicted.

gmkurtzer:
This is fantastic!

I'd love to see a `Are you sure [Y/n]: ` interaction when trying to push an unsigned container to the library and a documentation URL to how to sign images.

tri-adam:
This is looking great. I'd suggest tweaking the `pull` flow slightly.

Pulling an unsigned image:

```bash
$ singularity pull library://westleyk/test/unsigned-sif:latest
WARNING: This image is not signed, and thus its contents cannot be verified.
Do you wish to proceed? Y/n: n
```

Pulling a signed container without public key in keyring:

```bash
$ singularity pull library://westleyk/test/unsigned-sif:latest
Verifying image: unsigned-sif_latest.sif
WARNING: GPG error: the following signatures couldn't be verified because the public key is not available: 51BE5020C508C7E9
Attempt to retrieve the public key? Y/n: y
INFO:    key missing, searching key server for KeyID: 51BE5020C508C7E9...
INFO:    key retrieved successfully!
Store new public key 1CEC42121A994A3FBFAAF6CC51BE5020C508C7E9? [Y/n] y
	Sylabs Admin <support@sylabs.io>, KeyID 51BE5020C508C7E9
```

Pulling a signed container with public key in keyring:

```bash
$ singularity pull library://westleyk/test/unsigned-sif:latest
 660.00 KiB / 660.00 KiB [====================================] 100.00% 1.59 MiB/s 0s
Verifying image: unsigned-sif_latest.sif
	Sylabs Admin <support@sylabs.io>, KeyID 51BE5020C508C7E9
```

Unsigned _or_ signed image, with `--allow-unauthenticated`:

```bash
$ singularity pull library://westleyk/test/signed-sif:latest --allow-unauthenticated
 660.00 KiB / 660.00 KiB [====================================] 100.00% 1.59 MiB/s 0s
WARNING: Skipping data integrity check.
```

jscook2345:
The changes being introduced here look like they are going to break various forms of automation and cause problems for current users. I'd like to ask that you consider this and create a plan for addressing the following items before merging the PR.

Adding interactive input during pull will break automation. If you believe it is still needed than we will need a way to bypass it. If not, major changes to our tests will be required.

Automation often depends completely on text output, so a way to suppress warnings might be needed. If not, major changes to our tests may be required.

Command-line options that are not optional should be considered very carefully. When they are changed, any automation used around them must change. Users that currently automate around singularity will likely not be as happy to change their code to accommodate these features as I will. 😄 

I think it's OK to make command-line / api changes on a major release (3.2), but users should be notified in the changelog with specific information, not just a note that these features were added.

Please let me know if you have any further questions. Thanks!

WestleyK:
@jscook2345,

There is a way to avoid user input by using the `--allow-unauthenticated`, `-U` flag. And yes, we defiantly should add this to the change log.

WestleyK:
Github wont let me edit the description for this PR :angry: so heres the updated UI:

*(Edit: GitHub changed its mind, I am now able to edit by PR description! New UI in PR description)*

```bash
# pulling a unsigned container:
$ singularity pull library://ubuntu
 26.81 MiB / 26.81 MiB [=================================] 100.00% 6.48 MiB/s 4s
WARNING: This image is not signed, and thus its contents cannot be verified.
Pulled container (ubuntu_latest.sif) is **NOT** signed! Do you wish to proceed? [N/y] y

# now with '--allow-unauthenticated' flag:
$ singularity pull --allow-unauthenticated library://ubuntu
 26.81 MiB / 26.81 MiB [=================================] 100.00% 8.55 MiB/s 3s
WARNING: Skipping container verification

# pulling a signed container:
$ singularity pull library://alpine
 2.08 MiB / 2.08 MiB [===================================] 100.00% 2.16 MiB/s 0s
INFO:    Container is signed
	Sylabs Admin <support@sylabs.io>, KeyID 51BE5020C508C7E9


# pushing a unsigned container:
$ singularity push test.sif library://westleyk/test/unsigned:latest
Visit here for instructions on how to sign a container : https://www.sylabs.io/guides/3.0/user-guide/signNverify.html#verifying-containers-from-the-container-library
WARNING: Your container is **NOT** signed! You REALLY should sign your container before pushing!
Do you really want to continue? [N/y] y
INFO:    Now uploading test.sif to the library
 26.68 MiB / 26.68 MiB [=================================] 100.00% 6.96 MiB/s 3s
INFO:    Setting tag latest

# now with '--allow-unauthenticated' flag:
$ singularity push --allow-unauthenticated test.sif library://westleyk/test/unsigned:latest
WARNING: Skipping container verifying
INFO:    Now uploading test.sif to the library
 26.68 MiB / 26.68 MiB [=================================] 100.00% 7.25 MiB/s 3s
INFO:    Setting tag latest

# pushing a signed container:
$ singularity push test.sif library://westleyk/test/signed:latest
INFO:    Container is signed
	westleyk (examples) <westley@sylabs.io>, KeyID F1E47D55A7F3A56C
INFO:    Now uploading test.sif to the library
 26.68 MiB / 26.68 MiB [=================================] 100.00% 6.93 MiB/s 3s
INFO:    Setting tag latest
```

WestleyK:
OMG!!! Either Travis, or CircleCI fails for no reason! :man_facepalming:

gmkurtzer:
This part of the flow seems a bit cumbersome:

```
$ singularity pull library://westleyk/test/unsigned-sif:latest
Verifying image: unsigned-sif_latest.sif
WARNING: GPG error: the following signatures couldn't be verified because the public key is not available: 51BE5020C508C7E9
Attempt to retrieve the public key? Y/n: y
INFO:    key missing, searching key server for KeyID: 51BE5020C508C7E9...
INFO:    key retrieved successfully!
Store new public key 1CEC42121A994A3FBFAAF6CC51BE5020C508C7E9? [Y/n] y
	Sylabs Admin <support@sylabs.io>, KeyID 51BE5020C508C7E9
```

And to a "novice" the wording and warnings might seem concerning. I'd suggest we tune this messaging and workflow considerably. Open to suggestions.

tri-adam:
When doing a pull:

```
$ singularity pull library://ubuntu:latest
```

These three phrases seem to be saying similar things:

```
WARNING: unable to verify the container (ubuntu_latest.sif): error while searching for signature blocks: no signatures found for system partition
WARNING: This image is not signed, and thus its contents cannot be verified.
Pulled container (ubuntu_latest.sif) is **NOT** signed! Do you wish to proceed? [N/y] 
```

I think we could condense those into one message? I like the `This image is not signed, and thus its contents cannot be verified.` personally.

I agree with Greg about the flow after that, it is confusing. This part:

```
INFO:    key missing, searching key server for KeyID: 51BE5020C508C7E9...
INFO:    key retrieved successfully!
Store new public key 1CEC42121A994A3FBFAAF6CC51BE5020C508C7E9? [Y/n]
```

It isn't obvious why its looking for a key when I told it to pull an image. Maybe an INFO would help?

```
INFO:    Image 'library://alpine' is signed with key(s) that are not in your keyring.
INFO:    Searching 'https://keys.sylabs.io' for key (0x51BE5020C508C7E9).
Key found:
 Sylabs Admin () <support@sylabs.io>
 Created: 2018-10-08 13:48:34 -0400 EDT
 Fingerprint: 1CEC42121A994A3FBFAAF6CC51BE5020C508C7E9
Would you like to add this key to your local keyring? [N/y]
```

cmd/internal/cli/pull.go
```diff
@@ -50,6 +55,9 @@ func init() {
 	PullCmd.Flags().BoolVarP(&force, "force", "F", false, "overwrite an image file if it exists")
 	PullCmd.Flags().SetAnnotation("force", "envkey", []string{"FORCE"})
 
+	PullCmd.Flags().BoolVarP(&unauthenticatedPull, "allow-unauthenticated", "U", false, "dont check if the container is signed")
+	PullCmd.Flags().SetAnnotation("allow-unauthenticated", "envkey", []string{"ALLOW-UNAUTHENTICATED"})
```
mem: Environment variables cannot use dashes. Please use an underscore (_) instead.

cmd/internal/cli/pull.go
```diff
@@ -38,7 +42,8 @@ var (
 	// PullLibraryURI holds the base URI to a Sylabs library API instance
 	PullLibraryURI string
 	// PullImageName holds the name to be given to the pulled image
-	PullImageName string
+	PullImageName       string
+	unauthenticatedPull bool
```
mem, line 50: can you please add documentation even if the variable is not exported?

pkg/signing/signing.go
```diff
@@ -260,6 +260,17 @@ func getSigsForSelection(fimg *sif.FileImage, id uint32, isGroup bool) (sigs []*
 	return getSigsDescr(fimg, id)
 }
 
+// IsSigned : will return false if the givin container (cpath) is not signed.
+// Likewise, will return true if the container is signed and print who signed
+// it. will return a error if one occures.
+func IsSigned(cpath, url string, id uint32, isGroup bool, authToken string, noPrompt bool) (bool, error) {
+	err := Verify(cpath, url, id, isGroup, authToken, noPrompt)
+	if err != nil {
+		return false, fmt.Errorf("%v", err)
```
mem: You don't need to format the error, you can just pass it up

pkg/signing/signing.go
```diff
@@ -260,6 +260,17 @@ func getSigsForSelection(fimg *sif.FileImage, id uint32, isGroup bool) (sigs []*
 	return getSigsDescr(fimg, id)
 }
 
+// IsSigned : will return false if the givin container (cpath) is not signed.
+// Likewise, will return true if the container is signed and print who signed
+// it. will return a error if one occures.
+func IsSigned(cpath, url string, id uint32, isGroup bool, authToken string, noPrompt bool) (bool, error) {
```
mem: "url" is probably better named "keyserverURI"

cmd/internal/cli/push.go
```diff
@@ -34,6 +45,32 @@ var PushCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		// Push to library requires a valid authToken
 		if authToken != "" {
+			if !unauthenticatedPush {
+				// check if the container is signed
+				imageSigned, err := signing.IsSigned(args[0], "https://keys.sylabs.io", 0, false, authToken, true)
```
mem: We seem to have a growing collection of URIs that we are embedding in the code. It's probably a good idea to collect them in a single package. I have opened #2824 for this.

cmd/internal/cli/push.go
```diff
@@ -34,6 +45,32 @@ var PushCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
```
mem: this is becoming a bit large to have embedded as an unnamed function in the command declaration.

cmd/internal/cli/push.go
```diff
@@ -34,6 +45,32 @@ var PushCmd = &cobra.Command{
...
 		// Push to library requires a valid authToken
 		if authToken != "" {
+			if !unauthenticatedPush {
+				// check if the container is signed
+				imageSigned, err := signing.IsSigned(args[0], "https://keys.sylabs.io", 0, false, authToken, true)
+				if err != nil {
+					sylog.Warningf("%v", err)
```
mem, line 58: if signing.IsSigned returned an error we probably cannot continue. Also, I don't know what the error contains, but I would expect %v is not generally useful from the point of view of the user. %s (plus some context like "error while verifying signature for such and such") is probably more helpful for the user.

cmd/internal/cli/push.go
```diff
@@ -34,6 +45,32 @@ var PushCmd = &cobra.Command{
...
+				}
+				// if its not signed, print a warning
+				if !imageSigned {
+					sylog.Infof("TIP: Learn how to sign your own containers here : https://www.sylabs.io/guides/3.0/user-guide/signNverify.html")
```
mem: More URLs. This has the additional problem of hardcoding the singularity version.

cmd/internal/cli/push.go
```diff
@@ -34,6 +45,32 @@ var PushCmd = &cobra.Command{
...
+					fmt.Fprintf(os.Stderr, "\n")
+					sylog.Warningf("Your container is **NOT** signed! You REALLY should sign your container before pushing!")
+					fmt.Fprintf(os.Stderr, "Do you really want to continue? [N/y] ")
```
mem: this bit here until line 68 should probably go into is own function, in a separate package. I see there's already a sypgp.AskQuestion. It should be strict about getting a 'y', 'n' (or enter for default) from the user.

cmd/internal/cli/push.go
```diff
@@ -23,6 +31,9 @@ func init() {
 	PushCmd.Flags().StringVar(&PushLibraryURI, "library", "https://library.sylabs.io", "the library to push to")
 	PushCmd.Flags().SetAnnotation("library", "envkey", []string{"LIBRARY"})
 
+	PushCmd.Flags().BoolVarP(&unauthenticatedPush, "allow-unauthenticated", "U", false, "dont check if the container is signed")
+	PushCmd.Flags().SetAnnotation("allow-unauthenticated", "envkey", []string{"ALLOW-UNAUTHENTICATED"})
```
mem: Please change the dash to underscore.

cmd/singularity/pull_test.go
```diff
@@ -36,24 +39,26 @@ func TestPull(t *testing.T) {
...
 	}{
-		{"Pull_Library", "library://alpine:3.7", false, "", imagePath, true}, // https://cloud.sylabs.io/library
-		{"Force", "library://alpine:3.7", true, "", imagePath, true},
-		{"Pull_Docker", "docker://alpine:3.7", true, "", imagePath, true},   // https://hub.docker.com/
-		{"Pull_Shub", "shub://GodloveD/busybox", true, "", imagePath, true}, // https://singularity-hub.org/
-		{"PullWithHash", "library://alpine:sha256.af5e6b93dcfd08b7b46d6158c31ae5b6c2b4241c169b65595659ad9feac6f761", true, "", imagePath, true},
-		{"PullWithoutTransportProtocol", "alpine:3.7", true, "", imagePath, true},
+		{"Pull_Library", "library://alpine:3.8", false, false, "", imagePath, true}, // https://cloud.sylabs.io/library
+		{"Force", "library://alpine:3.8", true, false, "", imagePath, true},
+		{"Unsigned_image", "library://alpine:3.7", true, true, "", imagePath, true},
```
mem: I assumed this image does not have a signature. Shouldn't we test that trying to pull it when requesting authentication does fail?

```
{"Unsigned_image", "library://alpine:3.7", false, false, imagePath, false}
{"Unsigned_image", "library://alpine:3.7", true, false, imagePath, false}
```

The second case... should it succeed?

cmd/singularity/pull_test.go
```diff
@@ -36,24 +39,26 @@ func TestPull(t *testing.T) {
...
+		{"Pull_Docker", "docker://alpine:3.8", true, false, "", imagePath, true},   // https://hub.docker.com/
+		{"Pull_Shub", "shub://GodloveD/busybox", true, false, "", imagePath, true}, // https://singularity-hub.org/
+		{"PullWithHash", "library://alpine:sha256.69ce2a3dcc6d3e559e20ced0df251046ee6ecff390a945d856fe0dcb3bcb3ce8", true, false, "", imagePath, true},
+		{"PullWithoutTransportProtocol", "alpine:3.8", true, false, "", imagePath, true},
 	}
 	defer os.Remove(imagePath)
 	for _, tt := range tests {
 		t.Run(tt.name, test.WithoutPrivilege(func(t *testing.T) {
-			if b, err := imagePull(tt.library, tt.imagePath, tt.sourceSpec, tt.force); err != nil {
+			if b, err := imagePull(tt.library, tt.imagePath, tt.sourceSpec, tt.force, tt.unauthenticated); err != nil {
```
mem: this should probably look at tt.success in order to figure out if the operation should succeed or not.

cmd/singularity/pull_test.go
```diff
@@ -65,8 +70,25 @@ func TestPull(t *testing.T) {
...
 	test.DropPrivilege(t)
 
-	if b, err := imagePull("", "", "library://this_should_not/exist", false); err == nil {
+	if b, err := imagePull("", "", "library://this_should_not/exist", false, false); err == nil {
 		t.Log(string(b))
 		t.Fatalf("unexpected success")
 	}
 }
+
+func TestUnauthenticatedPull(t *testing.T) {
```
mem: It looks like this could be implemented with a table, just as `TestPull`

docs/content.go
```diff
@@ -574,8 +574,8 @@ Enterprise Performance Computing (EPC)`
 	PushUse   string = `push [push options...] <container image> library://[user[collection/[container[:tag]]]]`
 	PushShort string = `Push a container to a Library URI`
 	PushLong  string = `
-  The Singularity push command allows you to upload your sif image to a library
-  of your choosing`
+  The Singularity push command allows you to upload your container to a library.
+  Its always good practice to sign your container before pushing them to the library.`
```
mem: "It's" and either "your containers" or "pushing it"

pkg/signing/signing.go
```diff
@@ -315,12 +328,13 @@ func Verify(cpath, url string, id uint32, isGroup bool, authToken string, noProm
 		signer, err := openpgp.CheckDetachedSignature(elist, bytes.NewBuffer(block.Bytes), block.ArmoredSignature.Body)
 		if err != nil {
 			// verification with local keyring failed, try to fetch from key server
-			sylog.Infof("key missing, searching key server for KeyID: %s...", fingerprint[24:])
-			netlist, err := sypgp.FetchPubkey(fingerprint, url, authToken, noPrompt)
+			sylog.Infof("Image '%v' is signed with key(s) that are not in your keyring.", cpath)
+			sylog.Infof("Searching 'https://keys.sylabs.io' for key (0x%X)...", fingerprint[24:])
```
mem: fingerprint[24:] is showing up in multiple places. It's probably a good idea to have a keyID function that extracts the relevant part of the fingerprint.

pkg/signing/signing.go
```diff
@@ -360,8 +357,30 @@ func Verify(cpath, url string, id uint32, isGroup bool, authToken string, noProm
...
+				return fmt.Errorf("could not store public key: %s", err)
+			}
+		} else {
+			// Ask to store new public key
+			resp, err := sypgp.AskQuestion("Would you like to add this key to your local keyring? [Y/n] ")
+			if err != nil {
+				return err
+			}
+			if resp == "" || resp == "y" || resp == "Y" {
+				if err = sypgp.StorePubKey(netlist[0]); err != nil {
```
mem: it looks like this bit of code (storing the public key) is common to both the interactive and the non-interactive paths. Maybe you can write this so that you have something like this:

    if prompt {
        // prompt here
        store = askQuestion(...)
    } else {
        store = true
    }
    if store {
        sypgp.StorePubKey(...)
    }

mem (changes requested):

cmd/internal/cli/pull.go
WestleyK: Oh, I didn't even know thoughts are environment variables :man_facepalming:  Awesome!

pkg/signing/signing.go
WestleyK: I'm actually going to try to un-change this function `Verify()`, then do all the changes in #2821, To minimize the conflict.

mem (commented):
changes look good to me, thanks!

pkg/signing/signing.go
mem: Ok

cmd/singularity/pull_test.go
WestleyK: In this case, `alpine:3.7` is **NOT** signed, and `alpine:3.8` is signed. We should probably use a controlled unsigned container like `sylabs/tests/unsigned:1.0.0`. Eventually all default containers will be signed...

cmd/internal/cli/push.go
WestleyK: The `sylog.Warningf("%v", err)` will be a error like:

```
WARNING: unable to verify container: error while searching for signature blocks: no signatures found for system partition
```

After the error we ask the user if they want to continue using this unsigned container. Thats why its a `Warningf()` instead of a `Fatalf()`.

cmd/internal/cli/push.go
WestleyK: Agree, but should we import `sypgp` just to ask a question? Should there be a different package for that?

cmd/singularity/pull_test.go
WestleyK: Can you explain a bit more please?

cmd/internal/cli/push.go
mem: Hmm... I would expect `signing.IsSigned` to return `true, nil`, `false, nil` or `false, non-nil`. The first two are cases where the code was able to do whatever it needs to do to figure out if the image is signed or not. The third option means it was not able to do its job because there was an error, so I would expect that at that point it's not possible to continue.

In other words, "this thing does not have signatures" should return `false, nil`, because that's what it means not being signed. If it _is_ signed, but we don't recognize the signatures, that's a different situation.

Maybe what this needs is _two_ functions: `IsSigned` and `IsSignatureValid`?

cmd/internal/cli/push.go
mem: Since we seem to be asking questions in several places, I believe we should add a package for that. No need to do that in the context of this commit, though. This code can temporarily import sypgp and a future PR can refactor that into a separate package and have everyone use that code.

cmd/singularity/pull_test.go
mem: Oh, sorry for being terse, by bad.

What I mean is that each test case contains this `succeeds` field, which I expect specifies whether the test case is expect to succeed or not (succeeds == true means the test case should succeed, succeeds == false means the test case should fail).

By inspecting `err` you know whether the test case succeeded or not. Now you need to compare that against the expected result.

cmd/internal/cli/push.go
WestleyK: :thinking: ... I originally did it like this is to provide a some debugging, like *`there was a error downloading the public key`* or something like that, the container *could* still be signed, the user will see that it was *`unable to verify the container...`*, with the possibility it *could* still be signed with a valid key.

Thats just my thoughts, I'm open to changing it like how you said...

cmd/singularity/pull_test.go
```diff
@@ -36,37 +39,40 @@ func TestPull(t *testing.T) {
...
+		{"PullNonExistent", "library://this_should_not/exist/not_exist", true, false, "", imagePath, false}, // pull a non-existent container
 	}
 	defer os.Remove(imagePath)
 	for _, tt := range tests {
 		t.Run(tt.name, test.WithoutPrivilege(func(t *testing.T) {
-			if b, err := imagePull(tt.library, tt.imagePath, tt.sourceSpec, tt.force); err != nil {
-				t.Log(string(b))
-				t.Fatalf("unexpected failure: %v", err)
+			b, err := imagePull(tt.library, tt.imagePath, tt.sourceSpec, tt.force, tt.unauthenticated)
+			if tt.success {
```
WestleyK, line 68: @mem, is this what you mean by checking `tt.success`?

WestleyK:
Wake up Travis!!! :zzz:

cmd/internal/cli/pull.go
```diff
@@ -169,6 +178,36 @@ func pullRun(cmd *cobra.Command, args []string) {
...
+		if len(PullLibraryURI) >= 1 && !unauthenticatedPull {
+			imageSigned, err := signing.IsSigned(name, "https://keys.sylabs.io", 0, false, authToken, force)
+			if err != nil {
+				// err will be: "unable to verify container: %v", err
+				sylog.Warningf("%v", err)
+			}
+			// if container is not signed, print a warning
+			if !imageSigned {
+				fmt.Printf("This image is not signed, and thus its contents cannot be verified.")
+				fmt.Fprintf(os.Stderr, "Do you wish to proceed? [N/y] ")
```
ikaneshiro: When this is printed to a terminal there is no space between the two sentences. Also are `Do you wish to... [N/y]` messages typically over `stderr`?

cmd/internal/cli/pull.go
```diff
@@ -51,6 +57,9 @@ func init() {
 	PullCmd.Flags().BoolVarP(&force, "force", "F", false, "overwrite an image file if it exists")
 	PullCmd.Flags().SetAnnotation("force", "envkey", []string{"FORCE"})
 
+	PullCmd.Flags().BoolVarP(&unauthenticatedPull, "allow-unauthenticated", "U", false, "dont check if the container is signed")
+	PullCmd.Flags().SetAnnotation("allow-unauthenticated", "envkey", []string{"ALLOW_UNAUTHENTICATED"})
```
ikaneshiro, line 63: Add an entry to the map in `singularity.go` called `flagEnvFuncs` if you want this to be properly parsed, right now having an envvar like `SINGULARITY_ALLOW_UNAUTHENTICATED=...` will cause a panic

cmd/singularity/pull_test.go
ikaneshiro: @WestleyK looks correct to me

cmd/internal/cli/pull.go
```diff
@@ -169,6 +178,36 @@ func pullRun(cmd *cobra.Command, args []string) {
 		if err != nil {
 			sylog.Fatalf("%v\n", err)
 		}
+
+		// check if we pulled from the library, if so; is it signed?
+		if len(PullLibraryURI) >= 1 && !unauthenticatedPull {
+			imageSigned, err := signing.IsSigned(name, "https://keys.sylabs.io", 0, false, authToken, force)
```
ikaneshiro: Instead of having `handlePullFlags()` set what this should be, defaulting to `https://keys.sylabs.io` if no endpoints are configured.

cmd/internal/cli/push.go
```diff
@@ -37,6 +48,31 @@ var PushCmd = &cobra.Command{
 
 		// Push to library requires a valid authToken
 		if authToken != "" {
+			if !unauthenticatedPush {
+				// check if the container is signed
+				imageSigned, err := signing.IsSigned(args[0], "https://keys.sylabs.io", 0, false, authToken, true)
```
ikaneshiro: Same idea as in `pull.go` but this time with `handlePushFlags()`

ikaneshiro (dismissed):
Looking really good. Just have some nitpicks consider/work out.

cmd/internal/cli/pull.go
WestleyK: Opps, I think theres needs to be a newline there... :+1:

cmd/internal/cli/push.go
```diff
@@ -24,6 +32,9 @@ func init() {
 	PushCmd.Flags().StringVar(&PushLibraryURI, "library", "https://library.sylabs.io", "the library to push to")
 	PushCmd.Flags().SetAnnotation("library", "envkey", []string{"LIBRARY"})
 
+	PushCmd.Flags().BoolVarP(&unauthenticatedPush, "allow-unauthenticated", "U", false, "dont check if the container is signed")
+	PushCmd.Flags().SetAnnotation("allow-unauthenticated", "envkey", []string{"ALLOW_UNAUTHENTICATED"})
```
WestleyK: Should this be `ALLOW_UNAUTHENTICATED`? or `SINGULARITY_ALLOW_UNAUTHENTICATED`?

ikaneshiro: `ALLOW_UNAUTHENTICATED`, this is just specifying the suffix of the environment variable that will need to be set

cmd/internal/cli/pull.go
```diff
@@ -52,6 +59,9 @@ func init() {
 	PullCmd.Flags().BoolVarP(&force, "force", "F", false, "overwrite an image file if it exists")
 	PullCmd.Flags().SetAnnotation("force", "envkey", []string{"FORCE"})
 
+	PullCmd.Flags().BoolVarP(&unauthenticatedPull, "allow-unauthenticated", "U", false, "dont check if the container is signed")
```
mem:
```suggestion
	PullCmd.Flags().BoolVarP(&unauthenticatedPull, "allow-unauthenticated", "U", false, "do not require a signed container")
```

cmd/internal/cli/pull.go
```diff
@@ -170,6 +181,38 @@ func pullRun(cmd *cobra.Command, args []string) {
 		if err != nil {
 			sylog.Fatalf("%v\n", err)
 		}
+
+		// check if we pulled from the library, if so; is it signed?
+		if PullLibraryURI != "" && !unauthenticatedPull {
```
mem, line 186: I think I might have asked this in the past, but I can't find it.

Why is this check for `PullLibraryURI != ""` here? This variable is not used (that I can see) in the body of the `if`.

Shouldn't it be enough to check `!unauthenticatedPull`?

cmd/internal/cli/pull.go
```diff
@@ -170,6 +181,38 @@ func pullRun(cmd *cobra.Command, args []string) {
...
+				}
+				if resp == "" || resp != "y" && resp != "Y" {
+					fmt.Fprintf(os.Stderr, "Aborting.\n")
+					// not ideal to delete the container on the spot...
+					err := os.Remove(name)
+					if err != nil {
+						sylog.Fatalf("Unabel to delete the container: %v", err)
+					}
+					// exit status 10 after replying no
+					os.Exit(10)
```
mem: it looks like you could set exitStat to 10 and let the code exit below.

Or you remove exitStat, and just call exit on line 192.

cmd/internal/cli/push.go
```diff
@@ -24,6 +32,9 @@ func init() {
 	PushCmd.Flags().StringVar(&PushLibraryURI, "library", "https://library.sylabs.io", "the library to push to")
 	PushCmd.Flags().SetAnnotation("library", "envkey", []string{"LIBRARY"})
 
+	PushCmd.Flags().BoolVarP(&unauthenticatedPush, "allow-unauthenticated", "U", false, "dont check if the container is signed")
```
mem:
```suggestion
	PushCmd.Flags().BoolVarP(&unauthenticatedPush, "allow-unauthenticated", "U", false, "do not require a signed container")
```

cmd/internal/cli/push.go
```diff
@@ -37,6 +48,34 @@ var PushCmd = &cobra.Command{
 
 		// Push to library requires a valid authToken
 		if authToken != "" {
+			if _, err := os.Stat(args[0]); os.IsNotExist(err) {
+				sylog.Fatalf("Unable to open: %v: %v", args[0], err)
+			}
+			if !unauthenticatedPush {
+				// check if the container is signed
+				imageSigned, err := signing.IsSigned(args[0], "https://keys.sylabs.io", 0, false, authToken, true)
```
mem: can you please move the url (`https://keys.sylabs.io/`) to a package level constant? That makes it easier to collect it later and put it in a different location containing all the default URLs.

cmd/internal/cli/push.go
mem: My take is that if the user is asking for strict signature verification (unauthenticatedPush == false) errors during signature verification are fatal, not just warnings.

Like I said before, I cannot imagine a situation where `signing.IsSigned` should return `true, non-nil`.

The possibilities are:

- true, nil: image is properly signed → push
- false, nil: image is not signed → don't push
- false, non-nil: signature could not be verified → don't push, bail out
- true, non-nil: ??? → don't push (and possibly freak out)

cmd/internal/cli/push.go
```diff
@@ -37,6 +48,34 @@ var PushCmd = &cobra.Command{
...
+				if err != nil {
+					// err will be: "unable to verify container: %v", err
+					sylog.Warningf("%v", err)
+				}
+				// if its not signed, print a warning
+				if !imageSigned {
+					sylog.Infof("TIP: Learn how to sign your own containers here : https://www.sylabs.io/docs/")
+					fmt.Fprintf(os.Stderr, "\n")
+					fmt.Printf("Your container is **NOT** signed! You REALLY should sign your container before pushing!\n")
+					reps, err := sypgp.AskQuestion("Do you really want to continue? [N/y] ")
```
mem: no questions, the code got here because the user did not explicitly request to skip verification. If they want to push unauthenticated containers, they MUST pass --allow-unauthenticated (which also makes me think that in this specific case "unauthenticated" might be the wrong term, as there's no authentication going on, as it's only checking that the image has a signature and that that signature can be verified, maybe the option should be called "--allow-unsigned")

pkg/signing/signing.go
```diff
@@ -260,22 +260,41 @@ func getSigsForSelection(fimg *sif.FileImage, id uint32, isGroup bool) (sigs []*
...
+	if err != nil {
+		return false, fmt.Errorf("unable to verify container: %v", err)
+	}
+	if noLocalKey {
+		return true, fmt.Errorf("no local key matching entity")
+	}
+	return true, nil
+}
+
+// Verify takes a container path (cpath), and look for a verification block
```
mem, line 279:
```suggestion
// Verify takes a container path (cpath), and looks for a signature block
// for the specified descriptor id. If found, the signature block is used to verify
// the partition hash against the signer's version. Verify looks for keys in the
// local keyring. If keys are not found locally and localVerify is false, it queries
// the remote server specified by keyServiceURI.
//
// Verify returns a non-nil error if the signature block is not found, or if the necessary
// key cannot be found or the signature is invalid. If there's a signature present and the
// key is available, it returns true only if the signature block is present, it's properly
// signed and it matches the specified descriptor id.
```

pkg/signing/signing.go
```diff
@@ -350,9 +368,10 @@ func Verify(cpath, keyServiceURI string, id uint32, isGroup bool, authToken stri
 		}
 		author += fmt.Sprintf("\t%s, Fingerprint %X\n", name, signer.PrimaryKey.Fingerprint)
 	}
+	sylog.Infof("Container is signed")
 	fmt.Printf("Data integrity checked, authentic and signed by:\n%v", author)
 
-	return nil
+	return notLocalKey, nil
```
mem, line 374: shouldn't this be `return true, nil`?

mem (changes requested):
Sorry about the late comments :-(

cmd/internal/cli/pull.go
WestleyK: The `PullLibraryURI` is to check if we pulled from the library, as opposed to docker or shub. Because container pulled from docker/shub this way will not be signed, so we might as well skip the verifying.

cmd/internal/cli/push.go
WestleyK: This can return `true`, and an `error` if the container is signed, but you dont have a local key matching the signed container. Heres that that will look like:

```bash
$ singularity pull library://alpine:latest
 2.59 MiB / 2.59 MiB [============================================] 100.00% 440.96 KiB/s 6s
INFO:    Key with ID EDECE4F3F38D871E not found in local keyring, downloading from keystore...
INFO:    Container is signed
	Sylabs Admin <support@sylabs.io>, Fingerprint 8883491F4268F173C6E5DC49EDECE4F3F38D871E
WARNING: no local key matching entity

$ echo $?
1
```

Although, if you try to push a broken container (I deleted some lines in the sif file) it will not bail out, _it probably should_:

```bash
$ singularity shell test.sif 
FATAL:   container creation failed: mount /proc/self/fd/3->/usr/local/var/singularity/mnt/session/rootfs error: can't mount image /proc/self/fd/3: failed to mount squashfs filesystem: invalid argument

$ singularity push test.sif library://westleyk/foo/bar:latest
WARNING: sylabs-token files are deprecated. Use 'singularity remote' to manage remote endpoints and tokens.
WARNING: No default remote in use, falling back to: https://library.sylabs.io
WARNING: unable to verify container: failed to parse signature block                   <- should bail out after this error message
INFO:    TIP: Learn how to sign your own containers here : https://www.sylabs.io/docs/

Your container is **NOT** signed! You REALLY should sign your container before pushing!
Do you really want to continue? [N/y] 
```

pkg/signing/signing.go
WestleyK: No, this will return `true` if theres no local key matching the signed container. and will return `false` if you have a local key. Line `340` will decided if we return `true`, or `false`.

cmd/internal/cli/push.go
ikaneshiro: @WestleyK this package level constant should also be potentially set by `handlePushFlags()` just like it is in `pull.go`

ikaneshiro: Nevermind looks like that was already done :+1:

WestleyK: Yup, it is declared [here](https://github.com/sylabs/singularity/pull/2712/files#diff-1b3e82c021788b1761d186c7b249d2bdR48) and still *should* be set by [`handlePushFlags()`/`handlePullFlags()`](https://github.com/sylabs/singularity/pull/2712/files#diff-1b3e82c021788b1761d186c7b249d2bdR247). *If I did everything correctly...*

mem (approved):

WestleyK:
Travis is not happy about the *exit 1 if you dont have a local key matching the entity*. *Looking at that now...*

PR #2718, opened 2019-02-16
Fakeroot feature

### LINKED ISSUE #1924: --fakeroot is failing
The `--fakeroot` option is failing when using an image file.

```
DEBUG   [U=65534,P=2469]   setup_userns()                            Write line '0 1000 1' to gid_map
DEBUG   [U=65534,P=2469]   setup_userns()                            Write to UID map
VERBOSE [U=0,P=2469]       create_namespace()                        Create mount namespace
DEBUG   [U=0,P=2469]       init()                                    Create RPC socketpair for communication between scontainer and RPC server
VERBOSE [U=0,P=2469]       init()                                    Spawn smaster process
   [ ... ]
DEBUG   [U=0,P=2469]       create()                      Mount all
DEBUG   [U=0,P=2469]       mountGeneric()                Mounting tmpfs to /usr/local/singularity/3.0.0-dev/var/singularity/mnt/session
DEBUG   [U=0,P=2469]       CleanupContainer()            Cleanup container
FATAL   [U=0,P=2469]       SMaster()                     container creation failed: mount error: can't mount image /home/jason/tmp/sing/alpine.sif: failed to find loop device: permission denied
```

When using `shell --fakeroot` with a sandbox, it works as expected.

### LINKED ISSUE #3207: Allow admins to restrict usage of --fakeroot option
`--fakeroot` will require suid workflow to map uid gid info within the container.  Because ranges will be mapped from the host system to the container there could be some overlap.  Admins may want to restrict usage, and so there should be a way (in the conf file?) for admins to do this restriction.  

Should think about whether this should be enabled or disabled by default.  

To discuss with @DrDaveD.

cclerget:

This PR introduces fakeroot feature unlocking the possibility :
- to build an image as user, work only with the following bootstrap methods :
  - docker
  - library (if stored image format is SIF or squashfs)
  - shub (if stored image format is SIF or squashfs)
- execute container process as a fake root user able to switch to any UID/GID inside container (UID/GID are mapped to non-existent user/group outside of container)

Support :
- sandbox/squashfs/SIF images
- underlay
- network support with `--net` or `--net --network=fakeroot`
- usage is restricted by a whitelist, so user need to be added to `fakeroot allowed users` in `singularity.conf` in order to use fakeroot

As fakeroot use user namespace it has some user namespace limitations :
- no overlay support (`--overlay`, `--writable-tmpfs`)
- no cgroups support (`--apply-crgoups`)
- no `--security=uid` and `--security=gid` support (would make sense once filesystem issues adressed)
- `--drop-caps`, `--add-caps`, `--keep-privs`, `--no-privs` and `--allow-setuid` have no effect, fakeroot user has all capabilities by default and could manipulate those capabilities inside container
- **fakeroot users can't manipulate file owned by real root or system user**
- as real user UID/GID are mapped to root/root inside container, don't expect to change UID/GID to those you have outside container, they are mapped to non-existent user/group

Demo :
- build : https://asciinema.org/a/pSBeCEUNzoNCNhIIKHkO3sIkt
- exec : https://asciinema.org/a/ONi0fLRDFyDt71xq5IFrGnWJP

**Fakeroot requires:**
- **a privileged installation (with setuid binary)**
- **a kernel supporting user namespace (3.8 or later)**

**Issues related:**
- Fixes #1924
- Fixes #3207

### DIFF
--- a/cmd/internal/cli/action_flags.go
+++ b/cmd/internal/cli/action_flags.go
@@ -311,8 +311,7 @@ var actionFakerootFlag = cmdline.Flag{
 	DefaultValue: false,
 	Name:         "fakeroot",
 	ShortHand:    "f",
-	Hidden:       true,
-	Usage:        "run container in new user namespace as uid 0 (experimental)",
+	Usage:        "run container in new user namespace as uid 0",
 	EnvKeys:      []string{"FAKEROOT"},
 	ExcludedOS:   []string{cmdline.Darwin},
 }
--- a/cmd/internal/cli/actions_linux.go
+++ b/cmd/internal/cli/actions_linux.go
@@ -253,6 +253,7 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 	engineConfig.SetSecurity(Security)
 	engineConfig.SetShell(ShellPath)
 	engineConfig.SetLibrariesPath(ContainLibsPath)
+	engineConfig.SetFakeroot(IsFakeroot)
 
 	if ShellPath != "" {
 		generator.AddProcessEnv("SINGULARITY_SHELL", ShellPath)
@@ -365,6 +366,9 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 	}
 
 	if NetNamespace {
+		if IsFakeroot && Network != "none" {
+			engineConfig.SetNetwork("fakeroot")
+		}
 		generator.AddOrReplaceLinuxNamespace("network", "")
 	}
 	if UtsNamespace {
@@ -386,12 +390,9 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 
 	if UserNamespace {
 		generator.AddOrReplaceLinuxNamespace("user", "")
-		starter = buildcfg.LIBEXECDIR + "/singularity/bin/starter"
 
-		if IsFakeroot {
-			generator.AddLinuxUIDMapping(uid, 0, 1)
-			generator.AddLinuxGIDMapping(gid, 0, 1)
-		} else {
+		if !IsFakeroot {
+			starter = filepath.Join(buildcfg.LIBEXECDIR, "/singularity/bin/starter")
 			generator.AddLinuxUIDMapping(uid, uid, 1)
 			generator.AddLinuxGIDMapping(gid, gid, 1)
 		}
--- a/internal/pkg/build/build.go
+++ b/internal/pkg/build/build.go
@@ -60,7 +60,7 @@ type Config struct {
 
 // NewBuild creates a new Build struct from a spec (URI, definition file, etc...)
 func NewBuild(spec string, conf Config) (*Build, error) {
-	def, err := makeDef(spec, false)
+	def, err := MakeDef(spec)
 	if err != nil {
 		return nil, fmt.Errorf("unable to parse spec %v: %v", spec, err)
 	}
@@ -248,13 +248,19 @@ func engineRequired(def types.Definition) bool {
 
 // runBuildEngine creates an imgbuild engine and creates a container out of our bundle in order to execute %post %setup scripts in the bundle
 func runBuildEngine(b *types.Bundle) error {
-	if syscall.Getuid() != 0 {
-		return fmt.Errorf("Attempted to build with scripts as non-root user")
+	if syscall.Getuid() != 0 && !b.Opts.Fakeroot {
+		return fmt.Errorf("Attempted to build with scripts as non-root user or without --fakeroot")
 	}
 
 	sylog.Debugf("Starting build engine")
 	env := []string{sylog.GetEnvVar()}
 	starter := filepath.Join(buildcfg.LIBEXECDIR, "/singularity/bin/starter")
+	if b.Opts.Fakeroot {
+		starter = filepath.Join(buildcfg.LIBEXECDIR, "/singularity/bin/starter-suid")
+		if _, err := os.Stat(starter); os.IsNotExist(err) {
+			return fmt.Errorf("fakeroot feature requires to install Singularity as root")
+		}
+	}
 	progname := []string{"singularity image-build"}
 	ociConfig := &oci.Config{}
 
@@ -324,12 +330,7 @@ func getcp(def types.Definition, libraryURL, authToken string) (ConveyorPacker, 
 }
 
 // MakeDef gets a definition object from a spec
-func MakeDef(spec string, remote bool) (types.Definition, error) {
-	return makeDef(spec, remote)
-}
-
-// makeDef gets a definition object from a spec
-func makeDef(spec string, remote bool) (types.Definition, error) {
+func MakeDef(spec string) (types.Definition, error) {
 	if ok, err := uri.IsValid(spec); ok && err == nil {
 		// URI passed as spec
 		return types.NewDefinitionFromURI(spec)
@@ -347,11 +348,6 @@ func makeDef(spec string, remote bool) (types.Definition, error) {
 	}
 	defer defFile.Close()
 
-	// must be root to build from a definition
-	if os.Getuid() != 0 && !remote {
-		return types.Definition{}, fmt.Errorf("you must be the root user to build from a definition file")
-	}
-
 	d, err := parser.ParseDefinitionFile(defFile)
 	if err != nil {
 		return types.Definition{}, fmt.Errorf("while parsing definition: %s: %v", spec, err)
@@ -360,13 +356,8 @@ func makeDef(spec string, remote bool) (types.Definition, error) {
 	return d, nil
 }
 
-// MakeAllDefs gets a definition slice from a spec
-func MakeAllDefs(spec string, remote bool) ([]types.Definition, error) {
-	return makeAllDefs(spec, remote)
-}
-
-// makeAllDef gets a definition object from a spec
-func makeAllDefs(spec string, remote bool) ([]types.Definition, error) {
+// MakeAllDefs gets a definition object from a spec
+func MakeAllDefs(spec string) ([]types.Definition, error) {
 	if ok, err := uri.IsValid(spec); ok && err == nil {
 		// URI passed as spec
 		d, err := types.NewDefinitionFromURI(spec)
@@ -385,11 +376,6 @@ func makeAllDefs(spec string, remote bool) ([]types.Definition, error) {
 		return nil, fmt.Errorf("unable to open file %s: %v", spec, err)
 	}
 	defer defFile.Close()
-
-	// must be root to build from a definition
-	if os.Getuid() != 0 && !remote {
-		return nil, fmt.Errorf("you must be the root user to build from a definition file")
-	}
 
 	d, err := parser.All(defFile)
 	if err != nil {
--- a/internal/pkg/runtime/engines/imgbuild/engine_linux.go
+++ b/internal/pkg/runtime/engines/imgbuild/engine_linux.go
@@ -7,13 +7,17 @@ package imgbuild
 
 import (
 	"fmt"
-	"syscall"
+	"os"
+	"path/filepath"
 
 	specs "github.com/opencontainers/runtime-spec/specs-go"
-
+	"github.com/sylabs/singularity/internal/pkg/buildcfg"
+	"github.com/sylabs/singularity/internal/pkg/fakeroot"
 	"github.com/sylabs/singularity/internal/pkg/runtime/engines/config"
 	"github.com/sylabs/singularity/internal/pkg/runtime/engines/config/starter"
 	imgbuildConfig "github.com/sylabs/singularity/internal/pkg/runtime/engines/imgbuild/config"
+	"github.com/sylabs/singularity/internal/pkg/util/fs"
+	singularity "github.com/sylabs/singularity/pkg/runtime/engines/singularity/config"
 	"github.com/sylabs/singularity/pkg/util/capabilities"
 )
 
@@ -40,18 +44,56 @@ func (e *EngineOperations) PrepareConfig(starterConfig *starter.Config) error {
 		return fmt.Errorf("bad engine configuration provided")
 	}
 
+	configurationFile := filepath.Join(buildcfg.SYSCONFDIR, "/singularity/singularity.conf")
+
+	// check for ownership of singularity.conf
+	if starterConfig.GetIsSUID() && !fs.IsOwner(configurationFile, 0) {
+		return fmt.Errorf("%s must be owned by root", configurationFile)
+	}
+
+	fileConfig := &singularity.FileConfig{}
+	if err := config.Parser(configurationFile, fileConfig); err != nil {
+		return fmt.Errorf("unable to parse singularity.conf file: %s", err)
+	}
+
+	if !fileConfig.AllowSetuid && e.EngineConfig.Bundle.Opts.Fakeroot {
+		return fmt.Errorf("fakeroot requires to set 'allow setuid = yes' in %s", configurationFile)
+	}
+	if !starterConfig.GetIsSUID() && os.Getuid() != 0 {
+		return fmt.Errorf("unable to run imgbuild engine as non-root user or without --fakeroot")
+	}
+	if starterConfig.GetIsSUID() && !e.EngineConfig.Bundle.Opts.Fakeroot {
+		return fmt.Errorf("unable to run imgbuild engine as non-root user or without --fakeroot")
+	}
+
 	e.EngineConfig.OciConfig.SetProcessNoNewPrivileges(true)
 	starterConfig.SetNoNewPrivs(e.EngineConfig.OciConfig.Process.NoNewPrivileges)
 
-	if syscall.Getuid() != 0 {
-		return fmt.Errorf("unable to run imgbuild engine as non-root user")
-	}
-
-	if starterConfig.GetIsSUID() {
-		return fmt.Errorf("%s don't allow SUID workflow", e.CommonConfig.EngineName)
-	}
-
 	e.EngineConfig.OciConfig.SetupPrivileged(true)
+
+	if e.EngineConfig.Bundle.Opts.Fakeroot {
+		baseID := fileConfig.FakerootBaseID
+		allowedUsers := fileConfig.FakerootAllowedUsers
+		idRange, err := fakeroot.GetIDRange(baseID, allowedUsers)
+		if err != nil {
+			return err
+		}
+		e.EngineConfig.OciConfig.AddOrReplaceLinuxNamespace(specs.UserNamespace, "")
+
+		e.EngineConfig.OciConfig.AddLinuxUIDMapping(uint32(os.Getuid()), 0, 1)
+		e.EngineConfig.OciConfig.AddLinuxUIDMapping(idRange.HostID, idRange.ContainerID, idRange.Size)
+		starterConfig.AddUIDMappings(e.EngineConfig.OciConfig.Linux.UIDMappings)
+
+		e.EngineConfig.OciConfig.AddLinuxGIDMapping(uint32(os.Getgid()), 0, 1)
+		e.EngineConfig.OciConfig.AddLinuxGIDMapping(idRange.HostID, idRange.ContainerID, idRange.Size)
+		starterConfig.AddGIDMappings(e.EngineConfig.OciConfig.Linux.GIDMappings)
+
+		starterConfig.SetHybridWorkflow(true)
+		starterConfig.SetAllowSetgroups(true)
+
+		starterConfig.SetTargetUID(0)
+		starterConfig.SetTargetGID([]int{0})
+	}
 
 	e.EngineConfig.OciConfig.AddOrReplaceLinuxNamespace(specs.MountNamespace, "")
 
--- a/internal/pkg/runtime/engines/singularity/container_linux.go
+++ b/internal/pkg/runtime/engines/singularity/container_linux.go
@@ -14,6 +14,8 @@ import (
 	"strings"
 	"syscall"
 
+	"github.com/sylabs/singularity/internal/pkg/util/priv"
+
 	"github.com/sylabs/singularity/internal/pkg/util/mainthread"
 
 	specs "github.com/opencontainers/runtime-spec/specs-go"
@@ -239,24 +241,48 @@ func (c *container) setupSIFOverlay(img *image.Image, writable bool) error {
 	return nil
 }
 
+// checkOverlay will test if overlay is supported/allowed.
+func (c *container) checkOverlay() bool {
+	// NEED FIX: on ubuntu until 4.15 kernel it was possible to mount overlay
+	// with the current workflow, since 4.18 we get an operation not permitted
+	if c.userNS {
+		return false
+	}
+
+	if _, err := c.rpcOps.Mount("none", "/", "overlay", syscall.MS_SILENT, ""); err != nil {
+		// if an invalid argument error is returned, overlay is supported and is allowed
+		if err.Error() != os.ErrInvalid.Error() {
+			sylog.Debugf("Overlay seems not supported and/or not allowed by kernel")
+			return false
+		}
+	}
+
+	// previous mount forced overlay module to be loaded, check if we find
+	// it in /proc/filesystems
+	if has, _ := proc.HasFilesystem("overlay"); has {
+		sylog.Debugf("Overlay seems supported and allowed by kernel")
+		switch c.engine.EngineConfig.File.EnableOverlay {
+		case "yes", "try":
+			return true
+		default:
+			sylog.Debugf("Could not use overlay, disabled by configuration")
+		}
+	}
+
+	return false
+}
+
 // setupSessionLayout will create the session layout according to the capabilities of Singularity
 // on the system. It will first attempt to use "overlay", followed by "underlay", and if neither
 // are available it will not use either. If neither are used, we will not be able to bind mount
 // to non-existent paths within the container
 func (c *container) setupSessionLayout(system *mount.System) error {
 	writableTmpfs := c.engine.EngineConfig.GetWritableTmpfs()
-	overlayEnabled := false
+	overlayEnabled := c.checkOverlay()
 
 	sessionPath, err := filepath.EvalSymlinks(buildcfg.SESSIONDIR)
 	if err != nil {
 		return fmt.Errorf("failed to resolved session directory %s: %s", buildcfg.SESSIONDIR, err)
-	}
-
-	if enabled, _ := proc.HasFilesystem("overlay"); enabled && !c.userNS {
-		switch c.engine.EngineConfig.File.EnableOverlay {
-		case "yes", "try":
-			overlayEnabled = true
-		}
 	}
 
 	imgObject, err := c.loadImage(c.engine.EngineConfig.GetImage(), true)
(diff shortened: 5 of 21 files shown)


DrDaveD:
This looks quite unsafe to me.  I will discuss offline with @cclerget

DrDaveD:
My initial impression was due to a misunderstanding.  Now with a better understanding, it does seem mostly safe.

However, having looked into it more, I see that [another tool](https://github.com/rootless-containers/rootlesskit) that provides similar functionality does it via setuid root programs called 'newuidmap' and 'newgidmap', and those require the system administrator to whitelist allowed users in /etc/subuid and /etc/subgid.  Therefore I urge that this not be a default feature in setuid singularity, but only enabled by a setting in singularity.conf, possibly including a list of whitelisted users there.

In addition, I think that this would be much more useful and have equivalent safety if this namespace was *not* created with PR_SET_NO_NEW_PRIVS, so it could run setuid programs.  This could then also be an alternative to #2729.

One more thing that I think would also be useful and probably just as safe is, in addition to mapping id 0 to the original user id, to also map the original user id to the original user id.  That way within the container things could run "unprivileged" but still have write access to the original user's files.

Maybe this should be called "rootless" rather than "fakeroot".  That seems to be the more common term for this mode.

gmkurtzer:
How did your discussions go? Inquiring minds want to know! ;-)

DrDaveD:
@gmkurtzer I put a pretty good summary in the comments, perhaps you saw it after you asked.

DrDaveD:
I was wrong about mapping the original user id to the original user id; if that was done, there would be no way to distinguish its files from root's files.  What is needed is some uid/gid mapping separately inside the container, perhaps to some new separate filesystem mounted only inside the container, and I think maybe that's what @cclerget meant in "current limitations" above when he talked about using FUSE to fake uid/gid.  Or maybe what's needed is to support the user.rootlesscontainers xattr attribute mentioned on this [website about rootless containers](http://rootlesscontaine.rs) (which also mentions singularity as having made "odd design decisions", by the way).

DrDaveD:
Hey, this is way cooler than I thought.  All of the unprivileged users are not mapped to the same uid, they are each mapped to a unique id.  The number appears to be uid_in_container+2^32-2^16-2.  It must be because of the mapping set up in /proc/$$/uid_map, which is this:
```
         1 4294901759      65536
```
That second column is 2^32-2^16-1, and it probably subtracts the first column, so uid 1 becomes 4294901759, etc, for the next 65536 ids.  So one can switch user ids and create files on the host operating system that are different for the different ids.  Very cool.

I have this PR pulled in a branch in my workspace.  I rebased it on master to get recent changes, and added this patch:
```
diff --git a/cmd/starter/c/starter.c b/cmd/starter/c/starter.c
index b432886..d27ce66 100644
--- a/cmd/starter/c/starter.c
+++ b/cmd/starter/c/starter.c
@@ -216,7 +216,7 @@ static int prepare_stage(int stage, struct cConfig *config) 
         set_parent_death_signal(SIGKILL);
     }
 
-    if ( config->container.noNewPrivs ) {
+    if ( config->container.noNewPrivs && !config->container.fakeroot ) {
         if ( prctl(PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0) < 0 ) {
             fatalf("Failed to set no new privs flag: %s\n", strerror(errno));
         }
```
So then I was able to run setuid programs inside the container.

I ran into two problems, which may be able to be overcome.  First, any -B options for directories that did not exist in the container ahead of time did not get mounted; it did neither the overlay nor underlay feature when I did a "singularity exec --fakeroot --writable".  That may be a function of the --writable option, I'm not sure because I don't usually use it.   Second, if I attempt to run singularity-3.1.0 (installed with yum inside a --fakeroot --writable docker://fedora container) I get this error:
```
ERROR  : Cant' open /proc/self/auxv: Permission denied
```
I'm not sure why, because it does look like it is owned by the account I created inside the fakeroot container:
```
$ ls -l /proc/self/auxv
-r--------. 1 dwd docs 0 Mar 18 22:33 /proc/self/auxv
```

DrDaveD:
This leads to a strange situation where the original file owner can't remove files created inside the container when running on the host.  They need to go back into the container to do that.

Also, a significant problem with the current implementation is that all users map to the same 64k uids, so they can interfere with each other's files.  This is another reason to whitelist the users that may use it; it could subtract 64k from the range for each user in the list.

DrDaveD:
singularity-2.6 does not try to read /proc/self/auxv, so I tried running that inside a centos7 container.  With the -u option, it works!  Without -u, it gets an error 
```
ERROR  : There was an error remounting the path proc: Operation not permitted
```

DrDaveD:
singularity-3.1.0's -u also works inside the container.

DrDaveD:
Since the -u option only works on relatively modern host operating systems (e.g. centos 7.6) I'm still trying to see if I can get it to work with setuid singularity.  It would be great for my user community if we could run singularity inside of singularity on HPC systems.

If I modify singularity-3's starter.c to also treat geteuid() == 0 as setuid
```
diff --git a/cmd/starter/c/starter.c b/cmd/starter/c/starter.c
index d27ce66..e2ab142 100644
--- a/cmd/starter/c/starter.c
+++ b/cmd/starter/c/starter.c
@@ -684,6 +684,11 @@ static unsigned char is_suid(void) {
 
     verbosef("Check if we are running as setuid\n");
 
+    if ( geteuid() == 0 ) {
+        verbosef("Effective uid is 0\n");
+        return 1;
+    }
+
     if ( proc_auxv < 0 ) {
         fatalf("Cant' open /proc/self/auxv: %s\n", strerror(errno));
     }
```
then it fails on trying to set up overlay:
```
FATAL:   container creation failed: mount error: can't mount overlay filesystem to /var/singularity/mnt/session/final: operation not permitted
```
Since singularity.conf sets overlay = try I'm not sure why it doesn't fall back to underlay, but it's easily worked around by setting overlay = no.  Then it also fails as singularity-2.6 did:
```
FATAL:   container creation failed: mount error: can't remount /proc: operation not permitted
```
If I instead use -p it gets a little further but dies on
```
FATAL:   container creation failed: mount error: can't mount sysfs filesystem to /sys: operation not permitted
```

DrDaveD:
I discovered that running unprivileged singularity on centos 7.6 always works inside a singularity-3 container, even without --fakeroot, when started setuid-root or not.  I don't think I have tried that since singularity-2.  Inside a container started by singularity-2, it does not work, it gets `ERROR  : Failed invoking the NEWUSER namespace runtime: Operation not permitted`.  There must be something different about the way singularity-3 starts containers.  So that turns out to be no advantage for --fakeroot.  I was even able to nest it a third time, as long as I didn't try to use --contain more than once (that complained about being Multiple devpts instances).

cclerget:
@DrDaveD Great to hear it works for you without this feature, `--fakeroot` is really more oriented for micro-services

GodloveD:
Heya @cclerget, I know this is still a WIP, and you also said it was more oriented to working with the OCI command group, but I started playing with it and I have a few questions.  Maybe it would be helpful to provide some expected use cases to give users an idea of its intended use.

Anyway, here are a whole bunch of questions and comments in no particular order:

- Seems like the `--fakeroot` feature is of limited utility for `build` because you cannot execute scripts like `%post`.  Is this right?  

- How come the `--fakeroot` flag is conflicting with the `--writeable-tmpfs` flag?  Is that intentional?

- This makes sense (kindof) and is actually very cool:
```
dave@ubuntu:~/containers$ singularity exec --fakeroot alpine.sif ls -l foo
INFO:    Convert SIF file to sandbox...
-rw-r--r--    1 root     root             0 May 24 15:36 foo
INFO:    Cleaning up image...
dave@ubuntu:~/containers$ ls -l foo
-rw-r--r-- 1 dave dave 0 May 24 15:36 foo
```
But I'm not really sure what to make of this:
```
dave@ubuntu:~/containers$ singularity shell --fakeroot alpine.sif 
INFO:    Convert SIF file to sandbox...
Singularity> su - dave
ubuntu:~$ pwd
/home/dave
ubuntu:~$ touch bar
touch: bar: Permission denied
ubuntu:~$ id
uid=1000(dave) gid=1000(dave) groups=4(adm),24(cdrom),27(video),30(readproc),46(plugdev),116(lpadmin),126(sambashare),1000(dave),1000(dave)
ubuntu:~$ ls -ld /home/dave
drwxr-xr-x   24 root     root          4096 May 24 12:06 /home/dave
```
I guess it's a product of the UID=>UID mappings from host to inside container?

- I think you could add the short `-f` option to the `build --fakeroot` command.  `-F` is for `--force`.  

- Is this an error?
```
dave@ubuntu:~/containers$ singularity exec --fakeroot library://centos true
INFO:    Convert SIF file to sandbox...
WARNING: underlay of /etc/localtime required more than 50 (75) bind mounts
INFO:    Cleaning up image...
ERROR:   failed to delete container image /tmp/rootfs-857501020: permission denied
```

- This seems to work:
```
dave@ubuntu:~/containers$ sudo singularity build --sandbox ubuntu library://ubuntu
INFO:    Starting build...
INFO:    Build complete: ubuntu
dave@ubuntu:~/containers$ singularity shell --fakeroot --writable ubuntu
Singularity ubuntu:~/containers> apt-get update
[snip...]
```
But this fails:
```
dave@ubuntu:~/containers$ sudo singularity build --sandbox ubuntu2 docker://ubuntu
INFO:    Starting build...
Copying blob sha256:6abc03819f3e00a67ed5adc1132cfec041d5f7ec3c29d5416ba0433877547b6f
 27.52 MiB / 27.52 MiB [====================================================] 3s
Copying blob sha256:05731e63f21105725a5c062a725b33a54ad8c697f9c810870c6aa3e3cd9fb6a2
 844 B / 844 B [============================================================] 0s
Copying blob sha256:0bd67c50d6beeb55108476f72bea3b4b29a9f48832d6e045ec66b7ac4bf712a0
 164 B / 164 B [============================================================] 0s
Copying config sha256:68eb5e93296fbcd70feb84182a3121664ec2613435bd82f2e1205136352ae031
 2.36 KiB / 2.36 KiB [======================================================] 0s
Storing signatures
INFO:    Build complete: ubuntu2
dave@ubuntu:~/containers$ singularity shell --writable --fakeroot ubuntu2
Singularity ubuntu2:~/containers> apt-get update
Reading package lists... Done
E: List directory /var/lib/apt/lists/partial is missing. - Acquire (13: Permission denied)
```
Even though this works:
```
dave@ubuntu:~/containers$ sudo singularity shell --writable ubuntu2/
Singularity ubuntu2:~> apt-get update
[snip...]
```
Why?  Does it have to do with the incorrect permissions on the `library://ubuntu` container I wonder?  

- Also, should I expect to be able to use things with capabilities via `--fakeroot`?
```
dave@ubuntu:~/containers$ singularity exec centos/ ping 8.8.8.8
ping: socket: Operation not permitted
dave@ubuntu:~/containers$ sudo singularity exec centos/ ping 8.8.8.8
[sudo] password for dave: 
PING 8.8.8.8 (8.8.8.8) 56(84) bytes of data.
64 bytes from 8.8.8.8: icmp_seq=1 ttl=128 time=35.3 ms
64 bytes from 8.8.8.8: icmp_seq=2 ttl=128 time=35.3 ms
^C
--- 8.8.8.8 ping statistics ---
2 packets transmitted, 2 received, 0% packet loss, time 1003ms
rtt min/avg/max/mdev = 35.359/35.364/35.370/0.188 ms
dave@ubuntu:~/containers$ singularity exec --fakeroot centos/ ping 8.8.8.8
ping: socket: Operation not permitted
```
Maybe that is expected?  Don't know.

Thanks @cclerget.  I look forward to getting some more info from you and to playing more with this cool feature in the future.

cclerget:
>     * Seems like the `--fakeroot` feature is of limited utility for `build` because you cannot execute scripts like `%post`.  Is this right?
> 

You can execute `%post` section, it should work for docker/library bootstrap but not with debootstrap/yum/zypper

>     * How come the `--fakeroot` flag is conflicting with the `--writeable-tmpfs` flag?  Is that intentional?

yes as overlayfs requires privileges (or ubuntu to mount from user namespace), but as we are in a hybrid workflow, we could mount overlayfs for the user namespace

> ```
> dave@ubuntu:~/containers$ singularity shell --fakeroot alpine.sif 
> INFO:    Convert SIF file to sandbox...
> Singularity> su - dave
> ubuntu:~$ pwd
> /home/dave
> ubuntu:~$ touch bar
> touch: bar: Permission denied
> ubuntu:~$ id
> uid=1000(dave) gid=1000(dave) groups=4(adm),24(cdrom),27(video),30(readproc),46(plugdev),116(lpadmin),126(sambashare),1000(dave),1000(dave)
> ubuntu:~$ ls -ld /home/dave
> drwxr-xr-x   24 root     root          4096 May 24 12:06 /home/dave
> ```
> 
> I guess it's a product of the UID=>UID mappings from host to inside container?

Right, `root` in container is mapped to `dave` on host, and `dave` in container is mapped to a very large non-existent UID on host which doesn't correspond to the real user `dave` explaining the permission denied


> 
>     * I think you could add the short `-f` option to the `build --fakeroot` command.  `-F` is for `--force`.
> 

Yep will do

>     * Is this an error?
> ```
> dave@ubuntu:~/containers$ singularity exec --fakeroot library://centos true
> INFO:    Convert SIF file to sandbox...
> WARNING: underlay of /etc/localtime required more than 50 (75) bind mounts
> INFO:    Cleaning up image...
> ERROR:   failed to delete container image /tmp/rootfs-857501020: permission denied
> ```

Will check

>     * This seems to work:
> ```
> dave@ubuntu:~/containers$ sudo singularity build --sandbox ubuntu library://ubuntu
> INFO:    Starting build...
> INFO:    Creating sandbox directory...
> INFO:    Build complete: ubuntu
> dave@ubuntu:~/containers$ singularity shell --fakeroot --writable ubuntu
> Singularity ubuntu:~/containers> apt-get update
> [snip...]
> ```

A typo error `sudo singularity` ? because if you run it with sudo, the resulting sandbox files should be owned by the real `root`, so `apt-get update` command should have failed like with `sudo singularity build --sandbox ubuntu2 docker://ubuntu`

> Even though this works:
> 
> ```
> dave@ubuntu:~/containers$ sudo singularity shell --writable ubuntu2/
> Singularity ubuntu2:~> apt-get update
> [snip...]
> ```

Yes this is expected behaviour

> 
> Why? Does it have to do with the incorrect permissions on the `library://ubuntu` container I wonder?
> 
>     * Also, should I expect to be able to use things with capabilities via `--fakeroot`?

You can use capabilities as long you created namespaces (fakeroot make sense when creating full isolation with all namespaces enabled), by example you could not change hostname with just `--fakeroot` from container without passing `--fakeroot --uts`

> 
> 
> ```
> dave@ubuntu:~/containers$ singularity exec centos/ ping 8.8.8.8
> ping: socket: Operation not permitted
> dave@ubuntu:~/containers$ sudo singularity exec centos/ ping 8.8.8.8
> [sudo] password for dave: 
> PING 8.8.8.8 (8.8.8.8) 56(84) bytes of data.
> 64 bytes from 8.8.8.8: icmp_seq=1 ttl=128 time=35.3 ms
> 64 bytes from 8.8.8.8: icmp_seq=2 ttl=128 time=35.3 ms
> ^C
> --- 8.8.8.8 ping statistics ---
> 2 packets transmitted, 2 received, 0% packet loss, time 1003ms
> rtt min/avg/max/mdev = 35.359/35.364/35.370/0.188 ms
> dave@ubuntu:~/containers$ singularity exec --fakeroot centos/ ping 8.8.8.8
> ping: socket: Operation not permitted
> ```

as fakeroot use user namespace, it has only full privileges in namespaces it created, above you run ping in the host network namespace, you need to create one with `--net --network=fakeroot` to be able to ping

internal/pkg/build/build.go
```diff
@@ -248,13 +248,19 @@ func engineRequired(def types.Definition) bool {
 
 // runBuildEngine creates an imgbuild engine and creates a container out of our bundle in order to execute %post %setup scripts in the bundle
 func runBuildEngine(b *types.Bundle) error {
-	if syscall.Getuid() != 0 {
+	if syscall.Getuid() != 0 && !b.Opts.Fakeroot {
```
ikaneshiro, line 251: Can you add a change like this to `stage.go` as well. Currently the `%pre` script will cause the build to fail if it is defined

ikaneshiro (changes requested):
Overall this is looking good to me. 

We discussed the issue with overlays with a 4.18 kernel that needs resolution.

I also think it's important to document somewhere that building with definitions will not necessarily work as expected because of the sections outside of the container running without priv.

internal/pkg/build/build.go
```diff
@@ -322,12 +328,12 @@ func getcp(def types.Definition, libraryURL, authToken string) (ConveyorPacker,
...
 // MakeDef gets a definition object from a spec
-func MakeDef(spec string, remote bool) (types.Definition, error) {
-	return makeDef(spec, remote)
+func MakeDef(spec string) (types.Definition, error) {
+	return makeDef(spec)
 }
 
 // makeDef gets a definition object from a spec
-func makeDef(spec string, remote bool) (types.Definition, error) {
+func makeDef(spec string) (types.Definition, error) {
```
gvallee: I knew it is outside the scope of your PR but why have both `MakeDef` and `makeDef` since `MakeDef` simply directly call `makedef`?

internal/pkg/build/build.go
```diff
@@ -359,12 +360,12 @@ func makeDef(spec string, remote bool) (types.Definition, error) {
 }
 
 // MakeAllDefs gets a definition slice from a spec
-func MakeAllDefs(spec string, remote bool) ([]types.Definition, error) {
-	return makeAllDefs(spec, remote)
+func MakeAllDefs(spec string) ([]types.Definition, error) {
```
gvallee, line 360: Same here, why have both `MakeAllDefs` and `makeAllDefs`?

internal/pkg/runtime/engines/imgbuild/engine_linux.go
```diff
@@ -40,19 +44,56 @@ func (e *EngineOperations) PrepareConfig(starterConfig *starter.Config) error {
 		return fmt.Errorf("bad engine configuration provided")
 	}
 
-	e.EngineConfig.OciConfig.SetProcessNoNewPrivileges(true)
-	starterConfig.SetNoNewPrivs(e.EngineConfig.OciConfig.Process.NoNewPrivileges)
+	configurationFile := buildcfg.SYSCONFDIR + "/singularity/singularity.conf"
```
gvallee: Should we use file `filepath.Join` here?

internal/pkg/runtime/engines/imgbuild/engine_linux.go
```diff
@@ -40,19 +44,56 @@ func (e *EngineOperations) PrepareConfig(starterConfig *starter.Config) error {
...
+	// check for ownership of singularity.conf
+	if starterConfig.GetIsSUID() && !fs.IsOwner(configurationFile, 0) {
+		return fmt.Errorf("%s must be owned by root", configurationFile)
 	}
 
-	if starterConfig.GetIsSUID() {
-		return fmt.Errorf("%s don't allow SUID workflow", e.CommonConfig.EngineName)
+	fileConfig := &singularity.FileConfig{}
+	if err := config.Parser(configurationFile, fileConfig); err != nil {
+		return fmt.Errorf("Unable to parse singularity.conf file: %s", err)
```
gvallee:
```suggestion
		return fmt.Errorf("unable to parse singularity.conf file: %s", err)
```

internal/pkg/runtime/engines/singularity/container_linux.go
```diff
@@ -239,26 +241,49 @@ func (c *container) setupSIFOverlay(img *image.Image, writable bool) error {
 	return nil
 }
 
+// checkOverlay will test if overlay is supported/allowed. On recent ubuntu
+// distribution user namespace can use overlay filesystem, check mount error
+// to determine if we can use it
+func (c *container) checkOverlay() bool {
+	if _, err := c.rpcOps.Mount("none", "/", "overlay", syscall.MS_SILENT, ""); err != nil {
+		// invalid argument error means overlay is supported and is allowed,
```
gvallee:
```suggestion
		// if an invalid argument error is returned, overlay is supported and allowed,
```

internal/pkg/runtime/engines/singularity/container_linux.go
```diff
@@ -239,26 +241,49 @@ func (c *container) setupSIFOverlay(img *image.Image, writable bool) error {
...
+		// if RPC server is in a user namespace, we should get a permission
```
gvallee:
```suggestion
		// if the RPC server is in a user namespace, we should get a permission
```

internal/pkg/runtime/engines/singularity/container_linux.go
```diff
@@ -239,26 +241,49 @@ func (c *container) setupSIFOverlay(img *image.Image, writable bool) error {
...
+		// denied error on systems not allowing use of overlay from user
```
gvallee:
```suggestion
		// denied error on systems not allowing the use of overlay from user
```

internal/pkg/runtime/engines/singularity/container_linux.go
```diff
@@ -1715,55 +1746,79 @@ func (c *container) addActionsMount(system *mount.System) error {
...
-		if err := c.session.AddFile("/netns", nil); err != nil {
-			return nil, err
-		}
-		nspath, _ := c.session.GetPath("/netns")
-		if err := system.Points.AddBind(mount.SharedTag, "/proc/self/ns/net", nspath, 0); err != nil {
-			return nil, fmt.Errorf("could not hold network namespace reference: %s", err)
-		}
-		networks := strings.Split(c.engine.EngineConfig.GetNetwork(), ",")
+	// we hold a reference to container network namespace
+	// by binding /proc/self/ns/net (from RPC server) inside
```
gvallee:
```suggestion
	// by binding /proc/self/ns/net (from the RPC server) inside the
```

pkg/build/types/bundle.go
```diff
@@ -62,6 +62,8 @@ type Options struct {
 	// NoCleanUp allows a user to prevent a bundle from being cleaned up after a failed build
 	// useful for debugging
 	NoCleanUp bool `json:"noCleanUp"`
+	// fakeroot indicates if build engine use fakeroot feature
```
gvallee:
```suggestion
	// fakeroot indicates if the build engine uses the fakeroot feature
```

pkg/runtime/engines/singularity/config/data/singularity.conf
```diff
@@ -242,3 +242,34 @@ memory fs type = {{ .MemoryFSType }}
 # Allow to share same images associated with loop devices to minimize loop
 # usage and optimize kernel cache (useful for MPI)
 shared loop devices = {{ if eq .SharedLoopDevices true }}yes{{ else }}no{{ end }}
+
+# FAKEROOT BASE ID: [INT] 65536 <= BASE <= 4294901760
+# DEFAULT: 4227858432
+# This is the start of allocatable UID/GID in container for fakeroot users.
```
gvallee:
```suggestion
# This is the start of allocatable UID/GID in containers for fakeroot users.
```

pkg/runtime/engines/singularity/config/data/singularity.conf
```diff
@@ -242,3 +242,34 @@ memory fs type = {{ .MemoryFSType }}
...
+# Each user will be allocated a range of 65535 UID/GID based on this base,
+# eg:
+# the first user in allowed users list will get a UID/GID range starting at:
```
gvallee:
```suggestion
# the first user in the list of allowed users will get a UID/GID range starting at:
```

pkg/runtime/engines/singularity/config/data/singularity.conf
```diff
@@ -242,3 +242,34 @@ memory fs type = {{ .MemoryFSType }}
...
+# (BASE_ID)+(65536*0) -> 4227858432:4227923967
+# the second user in allowed users list will get a UID/GID range starting at:
```
gvallee:
```suggestion
# the second user in the list of allowed users will get a UID/GID range starting at:
```

pkg/runtime/engines/singularity/config/data/singularity.conf
```diff
@@ -242,3 +242,34 @@ memory fs type = {{ .MemoryFSType }}
...
+# (BASE_ID)+(65536*1) -> 4227923968:4227989503
+#
+# The default value of 4227858432 should be fine with most environment and
+# only changed if some users UID/GID on your system may overlap with the
+# allocated ranges. DON'T CHANGE IT IF USERS ALREADY USES THE FEATURE.
```
gvallee:
```suggestion
# allocated ranges. DON'T CHANGE IT IF USERS ALREADY USE THE FEATURE.
```

pkg/runtime/engines/singularity/config/data/singularity.conf
```diff
@@ -242,3 +242,34 @@ memory fs type = {{ .MemoryFSType }}
...
+# The default value allows to configure up to 1024 users, if you need to
```
gvallee:
```suggestion
# The default value enables the configuration of up to 1024 users; if you need to
```

pkg/runtime/engines/singularity/config/data/singularity.conf
```diff
@@ -242,3 +242,34 @@ memory fs type = {{ .MemoryFSType }}
...
+# configure more users, you must decrease this value, basically by 65536 to
+# add one user and so on.
+# Additionally this value must be a multiple of 65536.
+fakeroot base id = {{ .FakerootBaseID }}
+
+# FAKEROOT ALLOWED USERS: [STRING]
+# DEFAULT: Undefined
+# This is the list of users allowed to use the fakeroot feature.
+# User position will determine the allocated ID range based on
```
gvallee:
```suggestion
# User position will determine the allocated ID range based on the
```

gvallee (changes requested):
A few very minor comments

internal/pkg/build/build.go
cclerget: will make them public

cmd/internal/cli/action_flags.go
```diff
@@ -312,7 +312,7 @@ var actionFakerootFlag = cmdline.Flag{
 	Name:         "fakeroot",
 	ShortHand:    "f",
 	Hidden:       true,
```
GodloveD: probably also want to set `Hidden: false,`

cmd/internal/cli/actions_linux.go
```diff
@@ -386,12 +390,9 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
...
 	if UserNamespace {
 		generator.AddOrReplaceLinuxNamespace("user", "")
-		starter = buildcfg.LIBEXECDIR + "/singularity/bin/starter"
 
-		if IsFakeroot {
-			generator.AddLinuxUIDMapping(uid, 0, 1)
-			generator.AddLinuxGIDMapping(gid, 0, 1)
-		} else {
+		if !IsFakeroot {
+			starter = buildcfg.LIBEXECDIR + "/singularity/bin/starter"
```
ikaneshiro: Another place to use `filepath.Join()`

ikaneshiro (approved):
lgtm, apart from these requested changes that don't affect this PR's functionality.

GodloveD (changes requested):
As we discussed offline, I'm still seeing this error:
```
$ singularity exec --fakeroot library://centos true
INFO:    Convert SIF file to sandbox...
WARNING: underlay of /etc/localtime required more than 50 (75) bind mounts
INFO:    Cleaning up image...
ERROR:   failed to delete container image /tmp/rootfs-695506666: permission denied
```
And then as I commented, it would be good to remove the `Hidden: true` option from the `--fakeroot` cobra config.  Other than those 2 changes, this seems good to go.

cclerget:
> As we discussed offline, I'm still seeing this error:
> 
> ```
> $ singularity exec --fakeroot library://centos true
> INFO:    Convert SIF file to sandbox...
> WARNING: underlay of /etc/localtime required more than 50 (75) bind mounts
> INFO:    Cleaning up image...
> ERROR:   failed to delete container image /tmp/rootfs-695506666: permission denied
> ```
> 
> And then as I commented, it would be good to remove the `Hidden: true` option from the `--fakeroot` cobra config. Other than those 2 changes, this seems good to go.

The problem need to be solved at image level, `library://centos` have weird permissions, I don't know why but yum set 555 on most directory it created during bootstrap, this image could cause other permission denied in other situation if used as a base image for another build.

GodloveD:
>The problem need to be solved at image level, library://centos have weird permissions
OK, I'll buy that.  I know there have been a few weird images in the library that we've had to rebuild because of issues like that.

GodloveD (approved):

internal/pkg/build/build.go
GodloveD: Looks like these changes have been made, so I'm resolving the convo.  Holler at me if I'm wrong.

internal/pkg/runtime/engines/singularity/container_linux.go
GodloveD: Guess this is rendered moot b/c the whole comment is different now.

internal/pkg/runtime/engines/singularity/container_linux.go
GodloveD: Seems like this whole comment is deleted now, so this is moot I think.

gvallee (approved):
👍

PR #2728, opened 2019-02-18
Do not forward 'SINGULARITY_'-variables from user to container environment

obreitwi:
**Description of the Pull Request (PR):**

If environmental variables starting with `SINGULARITY_` are present, they will overwrite the ones set on container start. This makes it hard, for instance, to identify the current container if - for some reason - `SINGULARITY_NAME`/`SINGULARITY_CONTAINER` is already set:

```
$ singularity exec ./container.img env | grep SINGULARITY_
SINGULARITY_APPNAME=
SINGULARITY_CONTAINER=/home/obreitwi/containers/container.img
SINGULARITY_NAME=container.img

$ SINGULARITY_NAME="foobar" SINGULARITY_CONTAINER="deadbeef" singularity exec ./container.img env | grep SINGULARITY_
SINGULARITY_APPNAME=
SINGULARITY_CONTAINER=deadbeef
SINGULARITY_NAME=foobar
```

With this fix:
```
$ SINGULARITY_NAME="foobar" SINGULARITY_CONTAINER="deadbeef" singularity exec ./container.img env | grep SINGULARITY_                                                                                                                                                                                            (0)[20499] 
SINGULARITY_APPNAME=
SINGULARITY_CONTAINER=/home/obreitwi/containers/container.img
SINGULARITY_NAME=container.img
```

**This fixes or addresses the following GitHub issues:**

_None opened._

**Before submitting a PR, make sure you have done the following:**

- [x] Read the [Guidelines for Contributing](../CONTRIBUTING.md), and this PR conforms to the stated requirements.
- [ ] Added changes to the [CHANGELOG](../CHANGELOG.md) if necessary according to the [Contribution Guidelines](../CONTRIBUTING.md)
 → small bugfix, none needed
- [ ] Added tests to validate this PR and tested this PR locally with a `make testall`
 → small bugfix, ~none~ apparently some needed - is being able to set `SINGULARITY_CONTAINER` from the outside a desired feature?
- [x] Based this PR against the appropriate branch according to the [Contribution Guidelines](../CONTRIBUTING.md)
- [x] Added myself as a contributor to the [Contributors File](../CONTRIBUTORS.md)

Attn: @singularity-maintainers

PR #2729, opened 2019-02-18
Introduce trusted containers to allow nested execution of singularity containers

obreitwi:

In our group we are currently in the process of moving the complete user workflow into singularity containers which are rebuilt daily. Because we have several software environments which are incompatible with each others, we deploy several apps per container image. Since users are meant to do their complete development workflow inside some of the apps, being able to switch apps (and containers) within one session (or compute job) is a very desired feature. Sure, one can always leave the container and spawn a different one, but that would mean setting up the control infrastructure for every compute job in the minimal environment outside of containers.

This PR tries to solve the issues surrounding nested container execution.

It does so by introducing the concept of _trusted container images_ which have to reside under admin-chosen locations in the host's filesystem (i.e., per default users cannot provide their own images). Security measures on these images are less strict since they can only be deployed by a root-controlling user:
* The root filesystem is not mounted with `NOSUID`, hence execution of any setuid binary inside the container is possible (including ther `starter-suid` from `singularity`).
*  `PR_SET_NO_NEW_PRIVS` is not set which would prevent `singularity` inside the container from spawning another one.
* The bounding capabilities set is relaxed to a set of admin-chosen capabilities (for nested execution of containers these have to be at least `CAP_SETPCAP` and `CAP_SYS_ADMIN`).

With these measures in place, a singularity installation within a container is able to spawn another one, allowing for nested execution of containers as long as all of them reside under a trusted path.

Since we still want to allow non-root users to perform bind mounts on trusted images, some paths within the container have to be forbidden. Examples would be the path to `singularity.conf` or `/etc/{sudoers,group,passwd,shadow}` in a container that has `sudo` installed . Obviously, prior to enabling trusted containers, admins should check their container images for possible security vulnerabilities.

If users still want to bind mount a forbidden path, they have the option to execute with `--run-untrusted`. In this case the trusted image is executed in the default secure way without any user bind restrictions.

While the primary motivation for this PR is nested singularity execution, it might be useful in other situations where trusted container images should retain a different set of capabilities (for example to perform networking related tasks).

The changes in detail:
* Add `trusted container path` setting to singularity.conf:
  If the image to be loaded is under a path specified in trusted
  container path, then it is loaded in 'trusted' mode.
  This means that it is mounted without the `NOSUID` flag and that the
  process spawned inside the container does not have `PR_SET_NO_NEW_PRIVS`
  set. Hence non-root users are capable of executing setuid-binaries
  inside the trusted container (which is why extra care should be placed
  on ensuring they can do no harm).

* Add `forbidden bind trusted` setting to `singularity.conf`:
  If a container is run in 'trusted' mode by a non-root user, all paths
  specified as 'forbidden bind trusted' will not be available as bind
  mount target in order to prohibit exploits (for instance, by replacing
  the /etc/suoders file in a container with sudo installed).

* Add `--run-untrusted` flag to action commands:
  When `--run-untrusted` is specified, a trusted container is forcibly run
  in 'untrusted' mode, re-allowing all user binds while re-enabling
  PR_SET_NO_NEW_PRIVS.

* Add `bounding capability trusted` setting to `singularity.conf` that allows admins to
  specify a list of capabilities the bounding capabilities set should
  retain for processes spawned in trusted containers. This is useful to
  allow nested execution of singularity containers.

_None_

- [x] Read the [Guidelines for Contributing](../CONTRIBUTING.md), and this PR conforms to the stated requirements.
- [x] Added changes to the [CHANGELOG](../CHANGELOG.md) if necessary according to the [Contribution Guidelines](../CONTRIBUTING.md)
- [ ] Added tests to validate this PR and tested this PR locally with a `make testall`
  - [x] Added one test for `IsUnder` - probably is not enough..
- [x] Based this PR against the appropriate branch according to the [Contribution Guidelines](../CONTRIBUTING.md)
- [x] Added myself as a contributor to the [Contributors File](../CONTRIBUTORS.md) (in #2728)

Any feedback is greatly appreciated… :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ _With the release of `v3.0.0`, we're introducing a new changelog format in an at
 _The old changelog can be found in the `release-2.6` branch_
 
 # Changes Since v3.1.0
+  - Added `trusted container path` setting in `singularity.conf` to allow setuid binaries in designated containers to be run by non-root users
+  - Added `forbidden bind trusted` setting in `singularity.conf` to disallow some paths from user binds in trusted container to prevent exploits
+  - Added `--run-untrusted` flag to run trusted container in untrusted mode (disallowing setuid binaries but removing bind restrictions)
+  - Added `bounding capability trusted` setting in `singularity.conf` to retain some bounding capabilities in trusted container execution.
 
 # v3.1.0 - [2019.02.08]
 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -52,6 +52,7 @@
     - Michael Milton <ttmigueltt@gmail.com>
     - Nathan Lin <nathan.lin@yale.edu>
     - Oleksandr Moskalenko <om@rc.ufl.edu>
+    - Oliver Breitwieser <obreitwi@kip.uni-heidelberg.de>, <oliver@breitwieser.eu>
     - Oliver Freyermuth <freyermuth@physik.uni-bonn.de>
     - Olivier Sallou <olivier.sallou@irisa.fr>
     - Peter Steinbach <steinbach@scionics.de>
--- a/cmd/internal/cli/action_flags.go
+++ b/cmd/internal/cli/action_flags.go
@@ -49,11 +49,12 @@ var (
 	PidNamespace  bool
 	IpcNamespace  bool
 
-	AllowSUID bool
-	KeepPrivs bool
-	NoPrivs   bool
-	AddCaps   string
-	DropCaps  string
+	AllowSUID    bool
+	KeepPrivs    bool
+	NoPrivs      bool
+	AddCaps      string
+	DropCaps     string
+	RunUntrusted bool
 )
 
 var actionFlags = pflag.NewFlagSet("ActionFlags", pflag.ExitOnError)
@@ -267,4 +268,8 @@ func initPrivilegeVars() {
 	// --allow-setuid
 	actionFlags.BoolVar(&AllowSUID, "allow-setuid", false, "allow setuid binaries in container (root only)")
 	actionFlags.SetAnnotation("allow-setuid", "envkey", []string{"ALLOW_SETUID"})
+
+	// --run-untrusted
+	actionFlags.BoolVar(&RunUntrusted, "run-untrusted", false, "run container in untrusted mode even if it is in a trusted location (allows for more user binds)")
+	actionFlags.SetAnnotation("run-untrusted", "envkey", []string{"RUN_UNTRUSTED"})
 }
--- a/cmd/internal/cli/actions.go
+++ b/cmd/internal/cli/actions.go
@@ -72,6 +72,7 @@ func init() {
 		cmd.Flags().AddFlag(actionFlags.Lookup("docker-username"))
 		cmd.Flags().AddFlag(actionFlags.Lookup("docker-password"))
 		cmd.Flags().AddFlag(actionFlags.Lookup("docker-login"))
+		cmd.Flags().AddFlag(actionFlags.Lookup("run-untrusted"))
 		if cmd == ShellCmd {
 			cmd.Flags().AddFlag(actionFlags.Lookup("shell"))
 		}
--- a/cmd/internal/cli/actions_linux.go
+++ b/cmd/internal/cli/actions_linux.go
@@ -165,6 +165,7 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 	engineConfig.SetSecurity(Security)
 	engineConfig.SetShell(ShellPath)
 	engineConfig.SetLibrariesPath(ContainLibsPath)
+	engineConfig.SetRunUntrusted(RunUntrusted)
 
 	if ShellPath != "" {
 		generator.AddProcessEnv("SINGULARITY_SHELL", ShellPath)
--- a/internal/pkg/runtime/engines/singularity/config/config.go
+++ b/internal/pkg/runtime/engines/singularity/config/config.go
@@ -8,6 +8,7 @@ package singularity
 import (
 	"github.com/sylabs/singularity/internal/pkg/image"
 	"github.com/sylabs/singularity/internal/pkg/runtime/engines/config/oci"
+	"github.com/sylabs/singularity/internal/pkg/util/fs"
 )
 
 // Name is the name of the runtime.
@@ -48,6 +49,9 @@ type FileConfig struct {
 	CniPluginPath           string   `directive:"cni plugin path"`
 	MksquashfsPath          string   `directive:"mksquashfs path"`
 	SharedLoopDevices       bool     `default:"no" authorized:"yes,no" directive:"shared loop devices"`
+	TrustedContainerPaths   []string `directive:"trusted container path"`
+	ForbiddenBindsTrusted   []string `default:"/etc/passwd,/etc/group,/etc/sudoers,/etc/sudoers.d" directive:"forbidden bind trusted"`
+	BoundingCapsTrusted     []string `directive:"bounding capability trusted"`
 }
 
 // JSONConfig stores engine specific confguration that is allowed to be set by the user
@@ -90,6 +94,7 @@ type JSONConfig struct {
 	TargetUID     int           `json:"targetUID,omitempty"`
 	TargetGID     []int         `json:"targetGID,omitempty"`
 	LibrariesPath []string      `json:"librariesPath,omitempty"`
+	RunUntrusted  bool          `json:"runUntrusted,omitempty"`
 }
 
 // NewConfig returns singularity.EngineConfig with a parsed FileConfig
@@ -473,4 +478,30 @@ func (e *EngineConfig) SetLibrariesPath(libraries []string) {
 // /.singularity.d/libs directory
 func (e *EngineConfig) GetLibrariesPath() []string {
 	return e.JSON.LibrariesPath
+}
+
+// SetRunUntrusted returns true if run-untrusted is set and false if not.
+func (e *EngineConfig) SetRunUntrusted(runUntrusted bool) {
+	e.JSON.RunUntrusted = runUntrusted
+}
+
+// GetRunUntrusted returns true if run-untrusted is set and false if not.
+func (e *EngineConfig) GetRunUntrusted() bool {
+	return e.JSON.RunUntrusted
+}
+
+// GetImageTrusted returns true if the image is in a trusted paths and the user did
+// not override and false otherwise.
+func (e *EngineConfig) GetImageTrusted() (bool, error) {
+	if !e.GetRunUntrusted() {
+		isTrusted, err := fs.IsUnder(
+			e.GetImage(),
+			e.File.TrustedContainerPaths,
+			true)
+		if err != nil {
+			return false, err
+		}
+		return isTrusted, nil
+	}
+	return false, nil
 }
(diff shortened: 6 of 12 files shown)


bauerm97:
Hey @obreitwi . First of all, cheers for the contribution. Just to clarify and make sure we're on the same page as we discuss this proposal:

The general idea behind this proposal is to allow administrators to mark a container image (or set of container images) which can be executed **by a user** with more elevated permissions than usual?

obreitwi:
> The general idea behind this proposal is to allow administrators to mark a container image (or set of container images) which can be executed **by a user** with more elevated permissions than usual?

Yes, but the permissions are only potentially more elevated. The effective capabilities set remains empty. But, for instance via executing setuid-binaries, non-root users can elevate their permissions to as much as the bounding capabilities set permits.

The bounding capabilities set as well as which setuid-binaries are in the container/on the host system is again controlled by the administrators.

cclerget:
Hi @obreitwi , I'm afraid it will be very hard to do that securely as long as PR_SET_NO_NEW_PRIVS is disabled, actually nothing prevent user to inject its own library in trusted image so `forbidden bind trusted` is bypassed easily allowing a user to hijack libc inside container or any other library which could be loaded by a setuid binary like `starter-suid`, even with an empty bounding set a user could elevate his privileges to root outside of the container pretty easily

obreitwi:
> I'm afraid it will be very hard to do that securely as long as PR_SET_NO_NEW_PRIVS 
is disabled,

Oh, I fully agree about the "being hard" part. Granted, in our research environment it is more about preventing people from accidentally shooting themselves or others in the foot rather than preventing malicious attacks from the unauthorized strangers. Nevertheless, having nested container execution is a very much requested convenience feature we'd really like to have (even if it comes with a big red "I UNDERSTAND I MIGHT COMPROMISE MY SECURITY"-ribbon)..

> actually nothing prevent user to inject its own library in trusted image so `forbidden bind trusted` is bypassed easily allowing a user to hijack libc inside container or any other library which could be loaded by a setuid binary like `starter-suid`, even with an empty bounding set a user could elevate his privileges to root outside of the container pretty easily

This point I fail to understand: If the non-root user is forbidden from bind mounting anything into the folders containing `libc` and he cannot modify the trusted images, how could he hijack the setuid binaries any easier than he could on the host-system itself? Of course, any other support library would have to be linked via `RPATH` into all setuid binaries as well (while excluding those paths from user binds - again manually).


As I see it, using trusted container images vastly increases the attack surface for a malicious attacker, but with sufficient container vetting (and an accompanying long list of paths forbidden to user bind) it should be possible to do in principle, don't you agree?

bauerm97:
@obreitwi Sorry for the delay in response, just want to talk to a few of your points.

> As I see it, using trusted container images vastly increases the attack surface for a malicious attacker, but with sufficient container vetting (and an accompanying long list of paths forbidden to user bind) it should be possible to do in principle, don't you agree?

I agree with the first part of your assessment here, that a trusted container with elevated privileges does vastly increase the attack surface. However, even with sufficient container vetting, *any container* run with increased privilege which allows *any user bind mounts* will be inherently insecure. It's an all or nothing scenario. If a user can bind mount any host directory to an arbitrary spot in the container, they can potentially elevate privileges.

> This point I fail to understand: If the non-root user is forbidden from bind mounting anything into the folders containing `libc` and he cannot modify the trusted images, how could he hijack the setuid binaries any easier than he could on the host-system itself? Of course, any other support library would have to be linked via `RPATH` into all setuid binaries as well (while excluding those paths from user binds - again manually).

I think the problem with this approach is that we can't guarantee it can ever be secure. As discovered with the [recent container runtime CVE](https://www.sylabs.io/2019/02/cve-2019-5736-and-its-impact-on-singularity-containers/), there is always the potential to compromise the host when running a container as privileged. The _potential_ exploit vector here is giving the user the ability to bind-mount from any location on the host, where a clever user may be able to find a strange way to bind mount specific libraries or SUID binaries into the container at unexpected locations.

> Nevertheless, having nested container execution is a very much requested convenience feature we'd really like to have (even if it comes with a big red "I UNDERSTAND I MIGHT COMPROMISE MY SECURITY"-ribbon)..

If nested container execution is the goal, I do think this is something we can achieve with _minimal_ (not non-existent) security implications. I would propose talking a step back, and thinking about what exactly is needed for your use case + nested containers.

I think we want to explore this option by considering the expansion of the ECL feature of Singularity. We haven't (yet) publicly talked much about this, but if you look in `$(PREFIX)/etc/singularity`, there is a file `ecl.toml`. This file contains a set of restrictions (whitelist or blacklist) on whether users can run containers which are signed by certain entities and which live in certain directories. So, I think what I took away from your post is that we need to fulfill this:

> With these measures in place, a singularity installation within a container is able to spawn another one, allowing for nested execution of containers as long as all of them reside under a trusted path.

Giving users the ability to run nested Singularity containers. We could introduce some keyword to the `ecl.toml` file why allows a user to run a container with a special flag, enabling such nested execution. It could also specify a list of bind mounts that are always used. This guarantees two things:

1. You as the administrator HAVE to select the image, sign it with your image, and identify it in the `ecl.toml`
2. The user can run in "nested execution mode", where they cannot do potentially insecure actions (such as bind mounting); or they can run in regular mode, where they can use the image as normal

In my mind, the use case I see here is specifying a set of "base" images for users to choose from, and once the user is inside they can execute whatever container they like. You (as the administrator) could specify an automatic bind mount, potentially the users home directory, a shared filesystem with all of those "apps" you describe, etc... I *think* this would satisfy your requirement for:

> Sure, one can always leave the container and spawn a different one, but that would mean setting up the control infrastructure for every compute job in the minimal environment outside of containers.

Let me know if this idea is something that would solve your sites problems

DrDaveD:
I think the fakeroot feature proposed in #2718 could avoid PR_SET_NO_NEW_PRIVS and allow nested singularity.  It would be a lot simpler.

obreitwi:
Hey, thank you for your reply!

> I think the fakeroot feature proposed in #2718 could avoid PR_SET_NO_NEW_PRIVS and allow nested singularity. It would be a lot simpler.

@DrDaveD, in what way can the following "current limitations" in #2718 be mitigated?

As we are dealing with ~10GB SIF/squashfs images, we want to avoid unpacking them. Plus we need network access inside the nested containers.

> I agree with the first part of your assessment here, that a trusted container with elevated privileges does vastly increase the attack surface. However, even with sufficient container vetting, _any container_ run with increased privilege which allows _any user bind mounts_ will be inherently insecure. It's an all or nothing scenario. If a user can bind mount any host directory to _an arbitrary spot in the container_, they can potentially elevate privileges.

This is precisely the reason why certain user bind mounts have to be forbidden. I do see the principle problem though of having the process inside the container run without stripping all of its capabilities, especially `PR_NO_NEW_PRIVS`.

> I think the problem with this approach is that we can't guarantee it can ever be secure. As discovered with the [recent container runtime CVE](https://www.sylabs.io/2019/02/cve-2019-5736-and-its-impact-on-singularity-containers/), there is always the potential to compromise the host when running a container as privileged. The _potential_ exploit vector here is giving the user the ability to bind-mount from any location on the host, where a clever user may be able to find a strange way to bind mount specific libraries or SUID binaries into the container at unexpected locations.

Thanks for the link, was not aware of that exploit. I am beginning to think switching from a set of forbidden bind paths to a set of allowed ones (thereby not allowing user binds by default) would mitigate the problem quite a lot. Probably not sufficient in situations where you have possibly malicious parties executing containers, but still quite a lot.

> If nested container execution is the goal, I do think this is something we can achieve with _minimal_ (not non-existent) security implications. I would propose talking a step back, and thinking about what exactly is needed for your use case + nested containers.

Nested container execution is the main goal, but within those containers some binaries will also have to be run with non-zero capabilities (i.e., capturing raw network data).The ideal (i.e., non-realistic) situation would be if it made absolutely no difference to the user whether he was currently in a container or just ssh'ed into the machine when launching a singularity container.

> I think we want to explore this option by considering the expansion of the ECL feature of Singularity. We haven't (yet) publicly talked much about this, but if you look in `$(PREFIX)/etc/singularity`, there is a file `ecl.toml`. This file contains a set of restrictions (whitelist or blacklist) on whether users can run containers which are signed by certain entities and which live in certain directories.

> So, I think what I took away from your post is that we need to fulfill this:
> 
> Giving users the ability to run nested Singularity containers. We could introduce some keyword to the `ecl.toml` file why allows a user to run a container with a special flag, enabling such nested execution. It could also specify a list of bind mounts that are always used. This guarantees two things:
> 
> 1. You as the administrator HAVE to select the image, sign it with your image, and identify it in the `ecl.toml`

This would effectively be like the trusted container paths. Could the signing be optional? I dabbled around with signed container images and verifying the signature of a ~10GB image took on the order of 10s. Maybe verifying signatures could be cached though..

> 2. The user can run in "nested execution mode", where they cannot do potentially insecure actions (such as bind mounting); or they can run in regular mode, where they can use the image as normal

As mentioned above I think bind mounting should still be allowed in some admin-chosen locations, if need be. Plus, I would prefer it if "nested execution mode" could be set to be the default if the image is trusted (and that regular mode has to be enabled with a flag similar to `--run-untrusted`)

Just for clarification regarding implementation: Container images run in "nested execution mode" would still not have `PR_NO_NEW_PRIVS` set, right? Or is there a feature in regards to namespaces I missed that allows nested container execution regardless (e.g., #2718 but with network access/container images).

> In my mind, the use case I see here is specifying a set of "base" images for users to choose from, and once the user is inside they can execute whatever container they like. You (as the administrator) could specify an automatic bind mount, potentially the users home directory, a shared filesystem with all of those "apps" you describe, etc...

Yes, exactly. I think integrating `trusted container paths` into `ecl.toml` is much more reasonable than the current solution. Some remarks:

* The set of bounding capabilities with which each trusted container image is executed should be adjustable (`bounding capabilities trusted` in this PR). Our current specific use case is, as I mentioned above, the need of certain applications in verified containers to capture network traffic of the host they run on.
* The verification of large images should be fast (maybe via caching) or not mandatory (as admins control the host filesystem).
* Where should admin-approved locations to user bind mount to be specified?

Thank you for taking the time to respond!

DrDaveD:
> @DrDaveD, in what way can the following "current limitations" in #2718 be mitigated?
> As we are dealing with ~10GB SIF/squashfs images, we want to avoid unpacking them.
> Plus we need network access inside the nested containers.

I guess I'm assuming these will get solved and that we're talking about something a little longer term.  The most recent kernels have the ability mount FUSE filesystems in unprivileged user namespaces, and perhaps meanwhile singularity can do something clever with a setuid assist.  A [website on rootless containers](http://rootlesscontaine.rs) says that lxc solves the networking issue with a setuid assist.

GodloveD:
So the conversation has stalled on this it seems.  What is the final consensus view?  @bauerm97, @cclerget, @obreitwi, @DrDaveD ?

obreitwi:
> So the conversation has stalled on this it seems. What is the final consensus view? @bauerm97, @cclerget, @obreitwi, @DrDaveD ?

The final consensus was that the way this feature is implemented in the PR is still far too insecure to be deployed anywhere but a closed system. However, as we are already using the feature at @electronicvisions, a more restrictive version of these changes (e.g., allowing user-bind mounts under some paths only instead of forbidding) could be made available as a plugin.

A first starting point would be a rewrite in this direction in order to identify needed API injection points. I had hoped to finish it weeks ago, but my TODO list did not allow for its completion, yet…

GodloveD:
Thanks for the summary @obreitwi!  Sounds like you still have interest in this and have some new ideas to try.  🤘

dtrudg:
Hi @obreitwi - as this has gone commented for a while I'm going to close the PR now. If you do take a look at this again we'd be very glad to see what you have!

obreitwi:
No problem, since the current implementation does what we need with acceptable security implications, the rewrite, while still planned, is not as high up on my to-do list as I would like. I'll just open a new PR then.

PR #2755, opened 2019-02-21
Sylog dummy functions

cclerget:

This PR adds `sylog` dummy functions to allow `sylog` to be used in public package without interfering with logging of third-party applications using those public packages.


internal/pkg/sylog/sylog.go
```diff
@@ -5,6 +5,9 @@
 
 // Package sylog implements a basic logger for Singularity Go code to log
```
bauerm97: I think it would be smart here to create a new file `doc.go`, which is just an empty file with the package comment. This way we don't have to worry about the package comment not displaying in `go doc` when `sylog` build tag isn't set in the environment. Something like this [`doc.go`]:

```
// Package sylog implements a basic logger for Singularity Go code to log
// messages in the same format as singularity_message() from C code
package sylog
```

bauerm97 (changes requested):

internal/pkg/sylog/sylog.go
bauerm97: Then we can also remove the package comment from this `sylog.go` file

mem (approved):

bauerm97 (approved):

PR #2770, opened 2019-02-22
RPMPREFIX fix

jmstover:
Having RPMPREFIX defined where it was in mconfig, was overwriting the
value given on the command line. Remove the mconfig define, and just
accept if from the CLI if given.

### DIFF
--- a/mconfig
+++ b/mconfig
@@ -602,7 +602,6 @@ PSDIR := $psdir
 LIBDIR := $libdir
 LOCALEDIR := $localedir
 MANDIR := $mandir
-RPMPREFIX:=
 
 NAME := $package_name
 SHORT_VERSION := $short_version


jmstover:
@DrDaveD ... Can you scan over this?

DrDaveD:
What command line are you talking about, @jmstover?  A variable setting on the make command line overrides a setting inside a Makefile.  Can you give an example that reproduces the problem?

jmstover:
> What command line are you talking about, @jmstover? A variable setting on the make command line overrides a setting inside a Makefile. Can you give an example that reproduces the problem?

I ran this like:
```
./mconfig 
RPMPREFIX=/some/path/here make -C builddir/ rpm
```
This would give an `rpm` target of:
```
 BUILD RPM
(set -x; cd /root/development/go/src/github.com/sylabs/singularity && \
  if [ -n "" ]; then \
    rpmbuild  -ta \
      --define '_prefix ' \
      --define '_sysconfdir /etc' \
      --define '_localstatedir /var' \
      --define '_mandir /share/man' \
      /root/development/go/src/github.com/sylabs/singularity/singularity-3.1.0.tar.gz; \
  else \
    rpmbuild  -ta \
      /root/development/go/src/github.com/sylabs/singularity/singularity-3.1.0.tar.gz; \
  fi)
```

What was happening is the `RPMPREFIX:=` in the Makefile was clearing the value of `RPMPREFIX` given on the CLI.

DrDaveD:
That is correct and intended behavior.  By default make only accepts environment variables if they are not defined inside the Makefile.  That is in fact why I included initializing RPMPREFIX in the Makefile, to prevent unintended leakage from the environment.  The Usage is intended to be to set RPMPREFIX in the make command itself, after "make" on the command line, such as
```
make -C builddir/ rpm RPMPREFIX=/some/path/here
```
or
```
make -C builddir/ RPMPREFIX=/some/path/here rpm
```

jmstover:
Perfect. I'll make sure actual usage is noted somewhere. I'm so used to just defining the variables _somewhere_ on the CLI, and not needing it anywhere specific.

DrDaveD:
The correct usage is already in INSTALL.md, which is probably the right place for it.

PR #2776, opened 2019-02-25
Add Shub Image Caching

### LINKED ISSUE #2697: Shub Image Caching
## Version of Singularity:

```
$ singularity version
3.0.3-728.g2fb5f20e
```

### Expected behavior

Repeated pull to the same Shub image should use a cached image.

### Actual behavior

Repeated pull to the same Shub image download the same image multiple times.

run `singularity pull shub://vsoch/singularity-hello-world:master` twice

Since the shub API doesn't give an image hash in its manifest, we can't use that to uniquely identify the image in out cache. So, we need another method know if an image in a shub registry is currently cached.

michaelmoore10:

Image downloads using the Shub client (SingularityHub and sregistry) do not take advantage of the local cache, leading to repeated downloads of the same image.  This PR adds the logic to check for the existence of the image in the cache and uses it if it matches.

- Fixes #2697


cmd/internal/cli/actions.go
```diff
@@ -154,7 +154,16 @@ func handleShub(u string) (string, error) {
 	imageName := uri.GetName(u)
 	imagePath := cache.ShubImage("hash", imageName)
 
-	libexec.PullShubImage(imagePath, u, true, noHTTPS)
+    exists, err := cache.ShubImageExists("hash", imageName)
```
bauerm97: Hey @michaelmoore10 , the code looks good to me. Can you make sure you run `go fmt ./...` with the `go1.11` version of `go` installed and then push that as a new commit? Your code looks to be incorrectly formatted, and `go` is very particular about formatting. Cheers, looking forward to reviewing once that's fixed :)

bauerm97 (changes requested):

cmd/internal/cli/actions.go
michaelmoore10: Hi @bauerm97,

  Thanks!  I pushed a new version of this fix with the correct formatting.  

-Mike

cmd/internal/cli/actions.go
```diff
@@ -154,7 +154,16 @@ func handleShub(u string) (string, error) {
 	imageName := uri.GetName(u)
 	imagePath := cache.ShubImage("hash", imageName)
 
-	libexec.PullShubImage(imagePath, u, true, noHTTPS)
+	exists, err := cache.ShubImageExists("hash", imageName)
+	if err != nil {
+		return "", fmt.Errorf("unable to check if %v exists: %v", imagePath, err)
+	}
+	if !exists {
+		sylog.Infof("Downloading shub image")
```
bauerm97, line 166: I think this should be `Verbosef`.

cmd/internal/cli/actions.go
```diff
@@ -154,7 +154,16 @@ func handleShub(u string) (string, error) {
...
+		libexec.PullShubImage(imagePath, u, true, noHTTPS)
+	} else {
+		sylog.Infof("Use image from cache")
```
bauerm97: Same comment here, probably should be `Verbosef`. Does a user expect these to print when running without any logging level commands? What's your opinion here?

michaelmoore10: For both of these changes, the behavior is consistent with the handleLibrary() and handleNet() routines.  I agree, at least for the "Use image from cache" log message that Verbosef() is better.  No need to have that spit out for normal use.

Should the other routines be changed as well for using the image in the cache? And for the "downloading image" message in all of the routines? Is it ok to change the existing default behavior for logging?

bauerm97: >I agree, at least for the "Use image from cache" log message that Verbosef() is better. No need to have that spit out for normal use.

Why don't we go ahead and make that change then. I think it's not so useful to write that out

>Should the other routines be changed as well for using the image in the cache? And for the "downloading image" message in all of the routines? Is it ok to change the existing default behavior for logging?

I think it's okay if we change that. Maybe we will first merge this Pull Request as is, then you can follow up with a second pull request changing all of the logging messages? Are you up for that @michaelmoore10 ?

michaelmoore10: Yep.  I'm up for that.

bauerm97 (approved):
Looks good to me then!

PR #2784, opened 2019-02-27
Verify a Base Container from a def file build

### LINKED ISSUE #2748: Check if base container is signed when building from def file
## Version of Singularity:

```
3.1.0-rc3.28.gb3365013
```

### Expected behavior

There should be a warning message telling you that the container is not signed. otherwise if it is signed, verify it and show who signed it.

```bash
INFO:    Starting build...
 26.68 MiB / 26.68 MiB [====================================] 100.00% 9.20 MiB/s 2s
WARNING: Base container (ubuntu_latest.sif) is **NOT** signed!
Are you sure you want to continue? [N/y] y
INFO:    Creating SIF file...
```

### Actual behavior

There is no information about the base container, if its signed or not.

```bash
INFO:    Starting build...
 26.68 MiB / 26.68 MiB [====================================] 100.00% 9.20 MiB/s 2s
INFO:    Creating SIF file...
```

```bash
$ cat test.def
bootstrap: library
from: ubuntu:latest

%post
echo hello-world

[...]
```

WestleyK:

This PR will add: verifying the base container when building from a def file.

## Example:

**Unsigned base container:**

```bash
$ cat test.def
bootstrap: library
from: ubuntu:latest

INFO:    Starting build...
WARNING: The base container is **NOT** signed thus, its content cant be verified!
Do you really want to continue? [N/y] n
Stoping build.
FATAL:   While performing build: conveyor failed to get: user said not to build from unsigned container, good choice

$ sudo singularity build -U test.sif test.def 
INFO:    Starting build...
WARNING: Skipping verifction check.
INFO:    Creating SIF file...
```

**Signed base container:**

```bash
$ cat test.def 
bootstrap: library
from: alpine:latest

# With no local key matching the container
$ sudo singularity key list 
Public key listing (/root/.singularity/sypgp/pgp-public):

INFO:    Starting build...
INFO:    Key with ID EDECE4F3F38D871E not found in local keyring, downloading from keystore...
INFO:    Container is signed
	Sylabs Admin <support@sylabs.io>, Fingerprint 8883491F4268F173C6E5DC49EDECE4F3F38D871E
WARNING: no local key matching entity
INFO:    Creating SIF file...

# with '-l' flag
$ sudo singularity build -l test.sif test.def 
INFO:    Starting build...
FATAL:   While performing build: conveyor failed to get: unable to build container: no local key matching entity

# with a local key, and the '-l' flag
$ sudo singularity build -l test.sif test.def 
INFO:    Starting build...
INFO:    Container is signed
	Sylabs Admin <support@sylabs.io>, Fingerprint 8883491F4268F173C6E5DC49EDECE4F3F38D871E
INFO:    Creating SIF file...
```

- Fixes #2748 

<br>


cmd/internal/cli/pull.go
```diff
@@ -50,6 +55,9 @@ func init() {
 	PullCmd.Flags().BoolVarP(&force, "force", "F", false, "overwrite an image file if it exists")
 	PullCmd.Flags().SetAnnotation("force", "envkey", []string{"FORCE"})
 
+	PullCmd.Flags().BoolVarP(&unauthenticatedPull, "allow-unauthenticated", "U", false, "dont check if the container is signed")
+	PullCmd.Flags().SetAnnotation("allow-unauthenticated", "envkey", []string{"ALLOW-UNAUTHENTICATED"})
```
mem: `ALLOW_UNAUNTHENTICATED` please

cmd/internal/cli/pull.go
```diff
@@ -164,6 +172,38 @@ func pullRun(cmd *cobra.Command, args []string) {
...
+			imageSigned, err := signing.IsSigned(name, "https://keys.sylabs.io", 0, false, authToken, false)
+			if err != nil {
+				// warning message will be: ("Unable to verify the container (test.sif): %v", err)
+				sylog.Warningf("%v", err)
+			}
+			// if container is not signed, print a warning
+			if !imageSigned {
+				sylog.Warningf("This image is not signed, and thus its contents cannot be verified.")
+				fmt.Fprintf(os.Stderr, "Pulled container (%v) is **NOT** signed! Do you wish to proceed? [N/y] ", name)
+				reader := bufio.NewReader(os.Stdin)
```
mem: Can you please move the logic in lines 187-192 to a separate function?

cmd/internal/cli/pull.go
```diff
@@ -164,6 +172,38 @@ func pullRun(cmd *cobra.Command, args []string) {
...
+				if err != nil {
+					sylog.Fatalf("Error parsing input: %s", err)
+				}
+				if val := strings.Compare(strings.ToLower(input), "y\n"); val != 0 {
+					fmt.Fprintf(os.Stderr, "Aborting.\n")
+					err := os.Remove(name)
+					if err != nil {
+						// not ideal to delete the container on the spot...
+						sylog.Fatalf("Unable to delete container: %v", err)
+						os.Exit(255)
```
mem: please do not call os.Exit, as it prevents defer blocks from running.

cmd/internal/cli/push.go
```diff
@@ -23,6 +31,9 @@ func init() {
 	PushCmd.Flags().StringVar(&PushLibraryURI, "library", "https://library.sylabs.io", "the library to push to")
 	PushCmd.Flags().SetAnnotation("library", "envkey", []string{"LIBRARY"})
 
+	PushCmd.Flags().BoolVarP(&unauthenticatedPush, "allow-unauthenticated", "U", false, "dont check if the container is signed")
+	PushCmd.Flags().SetAnnotation("allow-unauthenticated", "envkey", []string{"ALLOW-UNAUTHENTICATED"})
```
mem: `ALLOW_UNAUTHENTICATED` please (underscore instead of dash)

cmd/internal/cli/push.go
```diff
@@ -34,6 +45,33 @@ var PushCmd = &cobra.Command{
...
+				if err != nil {
+					// warning message will be: ("Unable to verify the container (test.sif): %v", err)
+					sylog.Warningf("%v", err)
+				}
+				// if its not signed, print a warning
+				if !imageSigned {
+					sylog.Infof("TIP: Learn how to sign your own containers here : https://www.sylabs.io/guides/3.0/user-guide/signNverify.html")
+					fmt.Fprintf(os.Stderr, "\n")
+					sylog.Warningf("Your container is **NOT** signed! You REALLY should sign your container before pushing!")
+					fmt.Fprintf(os.Stderr, "Do you really want to continue? [N/y] ")
```
mem: can you please move lines 60-66 to a separate function? It should be the same as the other case, so you can reuse that code.

docs/content.go
```diff
@@ -574,8 +574,8 @@ Enterprise Performance Computing (EPC)`
 	PushUse   string = `push [push options...] <container image> library://[user[collection/[container[:tag]]]]`
 	PushShort string = `Push a container to a Library URI`
 	PushLong  string = `
-  The Singularity push command allows you to upload your sif image to a library
-  of your choosing`
+  The Singularity push command allows you to upload your container to a library.
+  Its always good practice to sign your container before pushing them to the library.`
```
mem: "It's" and either "your containers" or "pushing it"

internal/pkg/build/sources/conveyorPacker_library.go
```diff
@@ -69,6 +72,27 @@ func (cp *LibraryConveyorPacker) Get(b *types.Bundle) (err error) {
...
+	// check if the base container is signed
+	imageSigned, err := signing.IsSigned(imagePath, "https://keys.sylabs.io", 0, false, cp.AuthToken, true)
+	if err != nil {
+		// warning message will be: ("unable to verify the container (test.sif): %v", err)
+		sylog.Warningf("%v", err)
+	}
+	// if its not signed, print a warning
+	if !imageSigned {
+		sylog.Warningf("The base container is **NOT** signed thus, its content cant be verified!")
+		fmt.Fprintf(os.Stderr, "Do you really want to continue? [N/y] ")
```
mem: lines 84-90 are the same question-asking logic from before

mem (changes requested):

WestleyK:
Ummm, we may want to hold off on this and discuss the workflow... It dose _work_, its better than nothing, but we should discuss it.

internal/pkg/build/build.go
```diff
@@ -59,13 +59,13 @@ type Config struct {
 }
 
 // NewBuild creates a new Build struct from a spec (URI, definition file, etc...)
-func NewBuild(spec string, conf Config) (*Build, error) {
+func NewBuild(spec string, conf Config, allowUnauthenticatedBuild, localVerifyBuild bool) (*Build, error) {
```
ikaneshiro: Why not pass these as parameters of `Config`?

ikaneshiro (commented):
Question about the flow, do we want these redundant warnings?
```
$ singularity build test.sif library://ubuntu
INFO:    Starting build...
WARNING: unable to verify container: error while searching for signature blocks: no signatures found for system partition
WARNING: The base container is **NOT** signed thus, its content cant be verified!
Do you really want to continue? [N/y] y
You can avoid this by using the '-U' flag
INFO:    Creating SIF file...
```

internal/pkg/build/build.go
WestleyK: Good idea, I'll fix that up. *Back then when I made this PR my go knowledge was less then is is now.*

WestleyK: And yes, I know the warning messages need to be fixed. If you have suggestions that will help :+1:

internal/pkg/build/sources/conveyorPacker_library.go
```diff
@@ -70,11 +76,36 @@ func (cp *LibraryConveyorPacker) Get(b *types.Bundle) (err error) {
 		}
 	}
 
+	if !cp.AllowUnauthenticated {
+		// check if the base container is signed
+		imageSigned, err := signing.IsSigned(imagePath, "https://keys.sylabs.io", 0, false, cp.AuthToken, cp.LocalVerify, true)
+		if err != nil {
+			sylog.Warningf("%v", err)
```
ikaneshiro, line 88: @WestleyK What are the possible errors here? If its only that it can't be verified, then i think the flow on line 89 covers this case

WestleyK: There are two common errors:

```bash
WARNING: unable to verify container: error while searching for signature blocks: no signatures found for system partition

WARNING: unable to verify container: unable to verify container: openpgp: signature made by unknown entity
```

ikaneshiro: So it looks like the first case(referring to `IsSigned()`) is the only case where you return `false` indicating its not signed, why have an error when the boolean value already gives you the same information and this is expected if it is indeed not signed?

GodloveD:
I'll add my $0.02 to this.  I think the workflow is fine with one exception.  In the case that the user tries to build using a signed base container for which they do not have a local key, after Singularity displays the fingerprint, I think that the user should be given an `[y/N]` prompt where they can look at the fingerprint and see if they trust it.  If they answer `y` the build should proceed and they should get an exit code of 0.  If they answer `N`, the build should fail with a fatal error and they should get an error code of 1.  The user should be able to pass the `-U` option to override this behavior.

GodloveD:
Things have changed since this was opened.  `pull` will now simply error out if you pull an unverified container.  Need to add a flag so that it will accept.  We probably don't want to add a flag to the `build` command though.  Sounds like #3873 might be a better way to address this issue.

cmd/internal/cli/build.go
```diff
@@ -40,8 +40,37 @@ var (
 	dockerPassword string
 	dockerLogin    bool
 	noCleanUp      bool
+
+	// AllowUnauthenticatedBuild When true, will never ask if you want to build
```
mem, line 44: go documentation strings should be written as complete sentences as in `AllowUnauthenticatedBuild determines whether ...`. Also, I don't think I understand the sentence "this only works when your bootstrap is the library".

And this variable should not be exported.

cmd/internal/cli/build.go
```diff
@@ -40,8 +40,37 @@ var (
...
+	// a container from a unsigned base container. This only works when your bootstrap
+	// is the library
+	AllowUnauthenticatedBuild bool
+	// LocalVerifyBuild if true, ...
```
mem, line 48: Same here

cmd/internal/cli/build.go
WestleyK: This PR will likely be closed soon (no need to review anymore), and rewritten in a new cleaner PR (with less commits). Thanks for the suggestion!

WestleyK: The "this only works when your bootstrap is the library" means that this feature will only work if you are building from the library, as opposed to docker, shub, etc...

cmd/internal/cli/build.go
```diff
@@ -40,8 +40,37 @@ var (
...
 
+// --allow-unsigned, -U
+var buildAllowUnsignedFlag = cmdline.Flag{
+	ID:           "buildAllowUnsignedFlag",
+	Value:        &AllowUnauthenticatedBuild,
+	DefaultValue: false,
+	Name:         "allow-unsigned",
+	ShortHand:    "U",
+	Usage:        "never prompt when building from a unsigned container, only when bootstrap is library",
+	EnvKeys:      []string{"ALLOW_UNSIGNED"},
```
mem, line 60: there's something interesting going on here: this environment variable is already defined for push and pull. The name of the corresponding flag is the name. To me it seems it's the same flag in all cases, but they all associate different Go variables with the flags. The end effect in the UI is that you can set a single environment variable `SINGULARITY_ALLOW_UNSIGNED=true` and this will set all three flags.

This is either intentional, in which case I would suggest adding a comment to document that it is intentional, or an accident, in which case I would suggest changing the name of the environment variable to something like SINGULARITY_BUILD_ALLOW_UNSIGNED.

cmd/internal/cli/build.go
```diff
@@ -40,8 +40,37 @@ var (
...
+
+// --local, -l
+var buildLocalVerifyFlag = cmdline.Flag{
+	ID:           "buildLocalVerifyFlag",
+	Value:        &LocalVerifyBuild,
+	DefaultValue: false,
+	Name:         "local",
+	ShortHand:    "l",
+	Usage:        "nothing here... TODO:",
+	EnvKeys:      []string{"LOCAL_VERIFY_FLAG"}, // <-- TODO:
```
mem, line 71: what's the TODO?

What does this flag do? LocalVerifyBuild does not tell me much, nor does the `--local` flag itself.

cmd/internal/cli/pull.go
```diff
@@ -270,7 +270,7 @@ func pullRun(cmd *cobra.Command, args []string) {
 
 		// check if we pulled from the library, if so; is it signed?
 		if !unauthenticatedPull {
-			imageSigned, err := signing.IsSigned(name, KeyServerURL, 0, false, authToken, true)
+			imageSigned, err := signing.IsSigned(name, KeyServerURL, 0, false, authToken, false, true)
```
mem, line 273: boolean arguments are usually a hint of a function doing multiple things, as in:

```
func something(a, b, c bool) {
    if a {
        // do "a"
    }
    if b {
        // do "b"
    }
    if c {
        // do "c"
    }
}
```

I haven't looked at the change in the function, but would it be possible to add another function with a specific purpose?

cmd/internal/cli/pull.go
```diff
@@ -293,6 +293,7 @@ func pullRun(cmd *cobra.Command, args []string) {
 					// exit status 10 after replying no
 					exitStat = 10
 				}
+				fmt.Fprintf(os.Stderr, "You can avoid this by using the '-U' flag\n")
```
mem, line 296: what is "this"? Does it refer to the message above on line 282? If yes, can the message be moved there?

cmd/internal/cli/push.go
```diff
@@ -78,6 +78,7 @@ var PushCmd = &cobra.Command{
 				if !imageSigned {
 					sylog.Infof("TIP: Learn how to sign your own containers here : https://www.sylabs.io/docs/")
 					fmt.Fprintf(os.Stderr, "\nUnable to verify your container! You REALLY should sign your container before pushing!\n")
+					fmt.Fprintf(os.Stderr, "You can avoid this by using the '-U' flag\n")
```
mem, line 81: please use the long form of the flag in the message.

cmd/singularity/build_test.go
```diff
@@ -90,16 +97,19 @@ func TestBuild(t *testing.T) {
...
-		{"BusyBoxSandbox", "", "../../examples/busybox/Singularity", true},
-		{"Debootstrap", "debootstrap", "../../examples/debian/Singularity", true},
-		{"DockerURI", "", "docker://busybox", true},
-		{"DockerDefFile", "", "../../examples/docker/Singularity", true},
-		{"SHubURI", "", "shub://GodloveD/busybox", true},
-		{"SHubDefFile", "", "../../examples/shub/Singularity", true},
-		{"LibraryDefFile", "", "../../examples/library/Singularity", true},
-		{"Yum", "yum", "../../examples/centos/Singularity", true},
-		{"Zypper", "zypper", "../../examples/opensuse/Singularity", true},
+		{"BusyBox", "", "../../examples/busybox/Singularity", false, false, true},
```
mem, line 100: the type above (line 94) does not have three booleans, only one.

Also, is it necessary to put all of this in a single test? The three booleans at the end suggest there are three different tests here.

And this kind of test is more appropriate for E2E as this is not a unit test but a whole system test (it runs singularity with specific options).

internal/pkg/build/build.go
```diff
@@ -43,17 +43,28 @@ type Build struct {
...
 	Dest string
+
 	// Format is the format of built container, e.g., SIF, sandbox
 	Format string
+
 	// NoCleanUp allows a user to prevent a bundle from being cleaned up after a failed build
 	// useful for debugging
 	NoCleanUp bool
+
+	// AllowUnsigned will allow unsigned containers...
```
mem, line 62: as input or output?

internal/pkg/build/build.go
```diff
@@ -43,17 +43,28 @@ type Build struct {
...
+	AllowUnsigned bool
+
+	// LocalVerify will verify the container localy...
```
mem, line 65: "locally" as opposed to ... ???

internal/pkg/build/build.go
```diff
@@ -159,8 +170,9 @@ func (b Build) cleanUp() {
 }
 
 // Full runs a standard build from start to finish
-func (b *Build) Full() error {
+func (b *Build) Full() (int, error) {
 	sylog.Infof("Starting build...")
+	exitCode := 0
```
mem, line 175: exitCode is never set, is it?

internal/pkg/build/build.go
```diff
@@ -292,10 +308,13 @@ func runBuildEngine(b *types.Bundle) error {
 	return starterCmd.Run()
 }
 
-func getcp(def types.Definition, libraryURL, authToken string) (ConveyorPacker, error) {
+func getcp(def types.Definition, libraryURL, authToken string, unauthenticatedBuild, localVerifyBuild bool) (ConveyorPacker, error) {
```
mem, line 311: I'm going to guess `ConveyorPacker` is an interface.

The interface should be extended so that it has two new methods (say `AllowUnauthenticated()` and something else for which I don't have a good suggestion). The Conveyors (or Packers) that don't support unauthenticated operations should return an error and the ones that do should default to authenticated and go back to unauthenticated when this function is called.

So the code would call this "factory" function `getcp` to obtain a ConveyorPacker and then the client code should call `AllowUnauthenticated()` as appropriate.

internal/pkg/build/sources/conveyorPacker_library.go
```diff
@@ -6,6 +6,7 @@
 package sources
 
 import (
+	"errors"
```
mem, line 9: I would recommend replacing this by `github.com/pkg/errors` instead.

internal/pkg/build/sources/conveyorPacker_library.go
```diff
@@ -70,14 +81,47 @@ func (cp *LibraryConveyorPacker) Get(b *types.Bundle) (err error) {
 		}
 	}
 
+	if !cp.AllowUnauthenticated {
```
mem, line 84: this entire branch should be a different function, say `func (cp *LibraryConveyorPacker) authenticate(...) error`

internal/pkg/build/sources/conveyorPacker_library.go
```diff
@@ -70,14 +81,47 @@ func (cp *LibraryConveyorPacker) Get(b *types.Bundle) (err error) {
...
+		// check if the base container is signed
+		imageSigned, err := signing.IsSigned(imagePath, "https://keys.sylabs.io", 0, false, cp.AuthToken, cp.LocalVerify, true)
```
mem, line 86: the URL should not be hardcoded here. Pull has an option to specify this and there are at least another two places where this URL already exists.

internal/pkg/build/sources/conveyorPacker_library.go
```diff
@@ -70,14 +81,47 @@ func (cp *LibraryConveyorPacker) Get(b *types.Bundle) (err error) {
...
+			sylog.Warningf("The base container is **NOT** signed thus, its content cant be verified!")
+			resp, err := sypgp.AskQuestion("Do you really want to continue? [N/y] ")
+			if err != nil {
+				return fmt.Errorf("unable to parsing input: %s", err)
+			}
+			if resp == "" || resp != "y" && resp != "Y" {
+				fmt.Fprintf(os.Stderr, "Stoping build.\n")
+				return fmt.Errorf("user said not to build from unsigned container, good choice")
+			}
+			fmt.Fprintf(os.Stderr, "You can avoid this by using the '-U' flag\n")
```
mem, line 105: long flag form.

internal/pkg/build/sources/conveyorPacker_library.go
```diff
@@ -70,14 +81,47 @@ func (cp *LibraryConveyorPacker) Get(b *types.Bundle) (err error) {
...
+		if imageSigned && err != nil {
+			sylog.Warningf("%v", err)
+			noEntity = true
+		}
+		if cp.LocalVerify && !imageSigned {
+			return fmt.Errorf("unable to build container: no local key matching entity")
+		}
+		// if its not signed, print a warning
+		if !imageSigned {
+			sylog.Warningf("The base container is **NOT** signed thus, its content cant be verified!")
```
mem, line 96: either "can't" or "cannot" or "can not", preferably the second.

internal/pkg/build/sources/conveyorPacker_library.go
```diff
@@ -70,14 +81,47 @@ func (cp *LibraryConveyorPacker) Get(b *types.Bundle) (err error) {
...
+			resp, err := sypgp.AskQuestion("Do you really want to continue? [N/y] ")
+			if err != nil {
+				return fmt.Errorf("unable to parsing input: %s", err)
+			}
+			if resp == "" || resp != "y" && resp != "Y" {
+				fmt.Fprintf(os.Stderr, "Stoping build.\n")
+				return fmt.Errorf("user said not to build from unsigned container, good choice")
```
mem, line 103: this does not look like an error.

pkg/signing/signing.go
```diff
@@ -241,8 +241,8 @@ func getSigsForSelection(fimg *sif.FileImage, id uint32, isGroup bool) (sigs []*
 // will return true if the container is signed. Also returns a error
 // if one occures, eg. "the container is not signed", or "container is
 // signed by a unknown signer".
-func IsSigned(cpath, keyServerURI string, id uint32, isGroup bool, authToken string, noPrompt bool) (bool, error) {
-	noLocalKey, err := Verify(cpath, keyServerURI, id, isGroup, authToken, false, noPrompt)
+func IsSigned(cpath, keyServerURI string, id uint32, isGroup bool, authToken string, localVerify, noPrompt bool) (bool, error) {
+	noLocalKey, err := Verify(cpath, keyServerURI, id, isGroup, authToken, localVerify, noPrompt)
```
mem, line 245: I still don't understand what "local verify" means.

mem (changes requested):

pkg/signing/signing.go
WestleyK: LocalVerify is when your are verifying a container with your local keys (keys in your `~/.singularity/sypgp/pgp-public`), as opposed to keys in the keystore. During the `verify`, if you specify the `--local` flag, it will fail if the container is signed with a key thats not in your local keyring. So keys in your local keyring should only be keys that you trust.

_I hope all of that makes some sense..._

WestleyK:
Closing, this PR is outdated. Will be making a cleaner followup PR.

PR #2801, opened 2019-03-04
OCI Config Embedding In SIF

ikaneshiro:

Adds a generic `JSON` partition to `SIF` image containing OCI configuration when building from an OCI source.

### DIFF
--- a/internal/pkg/build/assemblers/assembler_sif.go
+++ b/internal/pkg/build/assemblers/assembler_sif.go
@@ -59,6 +59,7 @@ func createSIF(path string, definition, ociConf []byte, squashfile string) (err 
 			Groupid:  sif.DescrDefaultGroup,
 			Link:     sif.DescrUnusedLink,
 			Data:     ociConf,
+			Fname:    "oci-config.json",
 		}
 		ociInput.Size = int64(binary.Size(ociInput.Data))
 
--- a/internal/pkg/build/assemblers/assembler_sif.go
+++ b/internal/pkg/build/assemblers/assembler_sif.go
@@ -31,7 +31,7 @@ import (
 type SIFAssembler struct {
 }
 
-func createSIF(path string, definition []byte, squashfile string) (err error) {
+func createSIF(path string, definition, ociConf []byte, squashfile string) (err error) {
 	// general info for the new SIF file creation
 	cinfo := sif.CreateInfo{
 		Pathname:   path,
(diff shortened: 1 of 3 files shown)


sashayakovtseva:
How do I access OCI json config ?

sashayakovtseva:
Using siftool source code?

internal/pkg/build/assemblers/assembler_sif.go
```diff
@@ -52,6 +52,20 @@ func createSIF(path string, definition []byte, squashfile string) (err error) {
 	// add this descriptor input element to creation descriptor slice
 	cinfo.InputDescr = append(cinfo.InputDescr, definput)
 
+	if len(ociConf) > 0 {
+		// data we need to create a definition file descriptor
+		ociInput := sif.DescriptorInput{
+			Datatype: sif.DataGenericJSON,
+			Groupid:  sif.DescrDefaultGroup,
+			Link:     sif.DescrUnusedLink,
```
cclerget, line 60: To quickly identify descriptor, maybe we could use `Name: "oci-config"` here

internal/pkg/build/sources/conveyorPacker_oci.go
```diff
@@ -318,6 +334,18 @@ func (cp *OCIConveyorPacker) insertRunScript() (err error) {
 		}
 	}
 
+	if len(cp.imgConfig.WorkingDir) > 0 {
```
cclerget: That may be an issue because OCI engine or Singularity couldn't override current working directory with the one user provided, may be better to let engines decides by parsing image config

cclerget (changes requested):

ikaneshiro:
@sashayakovtseva yep, I think the code here is a good starting point https://github.com/sylabs/sif/blob/master/cmd/siftool/info.go#L81

internal/pkg/build/assemblers/assembler_sif.go
ikaneshiro: I agree. There isn't a name parameter for input descriptors so ill look to see if there's another way to do it

internal/pkg/build/sources/conveyorPacker_oci.go
ikaneshiro: Good point, parsing the image config would be a better way to implement `CMD` and `ENTRYPOINT` as well. I'll remove this from this PR for now

internal/pkg/build/assemblers/assembler_sif.go
cclerget: Looks like `Fname: "image-config.json"` will do the job for input descriptors

ikaneshiro: You are correct. What do you think is the best name for it? I'm thinking `oci-config.json` but maybe adding a version to it like `oci-config-v1.json` would be better. @sashayakovtseva thoughts?

sashayakovtseva: I think can keep it simple, whenever  you need new version `-v2` can be added

cclerget (approved):

gmkurtzer:
We need to move SIFtool features into Singularity....

ikaneshiro:
@gmkurtzer Do you mean having it integrated into the singularity CLI, potentially as part of the 'inspect' command? Or installing it automatically as a part of singularity's installation process?

gmkurtzer:
Good question. Both would be reasonable, but integrating it into `singularity` CLI I think would be preferable.

PR #2816, opened 2019-03-05
Handle root requirements more consistently

### LINKED ISSUE #2725: Commands that don't work should error
## Version of Singularity:

3.0.3

### Expected behavior

Commands should not ever silently fail.

### Actual behavior

For example, `--keep-privs` is supposed to be a root only option and doesn't error when being called from as a user (others like `--net` does).

```
$ singularity shell --keep-privs Containers/centos7-devel.sif 
Singularity centos7-devel.sif:~>
```

### LINKED ISSUE #2735: Incorrect help for oci mount and umount
## Version of Singularity:

3.1.0-rc3

### Expected behavior

Accurate help messages

### Actual behavior

singularity oci mount -h and singularity oci umount -h do not show the full Usage.  They're missing the parameters.  Also, the description of oci umount  is incorrect, it should probably be "Unmount bundle".   

First mentioned in #2689.

See above.

mem:
Some flags require root privileges to work (--keep-privs,
--apply-cgroups, --allow-setuid, --boot). Since lower layers should not
know about the names of the flags (or environment variables) used to set
these options, handle the requirement close to the CLI, before
translating the flags into engine configuration.

- Fixes #2725


cmd/internal/cli/actions_linux.go
```diff
@@ -73,10 +87,9 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 		uid = uint32(targetUID)
 
 		engineConfig.SetTargetUID(targetUID)
-	} else if uidParam != "" {
-		sylog.Warningf("uid security feature requires root privileges")
-	}
-	if os.Getuid() == 0 && gidParam != "" {
+	})
+
+	checkPrivileges(uidParam != "", "gid security feature", func() {
```
bauerm97: I think this should be:

`checkPrivileges(gidParam != "", "gid security feature", func() {`

bauerm97 (changes requested):

cmd/internal/cli/actions_linux.go
mem: Oh, right, thanks!

bauerm97 (approved):

Issue #2534, opened 2019-01-11
Silent Exit with Invalid Flag


tri-adam:
Issue reported by @renefritze - thank you René!

## Version of Singularity:

`3.0.2`

### Expected behavior

Running Singularity with `-f <value>` option before the `push` command should print the usage and exit, since `-f` is not listed as a valid flag.

```sh
$ singularity -f blah push
Usage:
  singularity [global options...] <command>

Available Commands:
  build       Build a new Singularity container
  capability  Manage Linux capabilities on containers
  exec        Execute a command within container
  inspect     Display metadata for container if available
  instance    Manage containers running in the background
  keys        Manage OpenPGP key stores
  pull        Pull a container from a URI
  push        Push a container to a Library URI
  run         Launch a runscript within container
  run-help    Display help for container if available
  search      Search the library
  shell       Run a Bourne shell within container
  sign        Attach cryptographic signatures to container
  test        Run defined tests for this particular container
  verify      Verify cryptographic signatures on container
  version     Show application version
```

### Actual behavior

Singularity exits silently:

```sh
$ singularity -f blah push
$
```

### Steps to reproduce behavior

Run `singularity -f blah push`.

mem:
This one is interesting... I'm following cobra code to figure out what's going on...

```
$ ./singularity exec ./alpine_latest.sif id
uid=1001(mem) gid=1001(mem) groups=27(video),1001(mem)

$ ./singularity -f exec ./alpine_latest.sif id
unknown shorthand flag: 'f' in -f
Usage:
  singularity [global options...] <command>

Available Commands:
  apps        apps <image path>
  build       Build a new Singularity container
  cache       Manage your local singularity cache
  capability  Manage Linux capabilities on containers
  exec        Execute a command within container
  inspect     Display metadata for container if available
  instance    Manage containers running in the background
  key         Manage OpenPGP key stores
  oci         Manage OCI containers
  plugin      Manage singularity plugins
  pull        Pull a container from a URI
  push        Push a container to a Library URI
  run         Launch a runscript within container
  run-help    Display help for container if available
  search      Search the library
  shell       Run a Bourne shell within container
  sign        Attach cryptographic signatures to container
  test        Run defined tests for this particular container
  verify      Verify cryptographic signatures on container
  version     Show application version

$ ./singularity -f xxx exec ./alpine_latest.sif id
$
```

[...]

PR #2817, opened 2019-03-05
Print usage information if an invalid flag is passed to singularity

mem:
In cobra, rootCmd.Execute() processes the command line and executes the
appropriate command or subcommand. If an invalid flag is passed to the
subcommand, the usage information for the subcommand is printed. If an
invalid subcommand or flag is passed to the root command, the usage
information for the root command is printed. If in invalid subcommand is
passed along with an invalid flag (`singularity -f xxx shell`), cobra
prints an error. Since we are suppressing cobra errors (SilenceErrors),
nothing is printed and singularity returns silently.

Fix this by passing SilenceUsage as well, and take care of printing the
usage string ourselves for all errors.

- Fixes: #2534


jscook2345:
A couple notes from playing around with things.

* It would be nice if the error indicated what option was invalid in case you are using many options. A good example would be to look at what `git` does.

<pre>
[vagrant@localhost singularity]$ git status -r
error: unknown switch `r'
usage: git status [options] [--] <pathspec>...

    -v, --verbose         be verbose
    -s, --short           show status concisely
... more options ...
</pre>

* The usage information does not show the options available, it shows all the sub-commands only. To me that is confusing because it does not show what the proper options are. It may be a better idea to print the output of the help. Note that you can see the difference with `singularity shell` (usage) vs `singularity help shell` (help).

* Sub-commands have the same issue in that printing the usage does not print the available options

<pre>
[vagrant@localhost singularity]$ singularity shell -r
Usage:
  singularity [global options...] shell [shell options...] <container>

Run 'singularity shell --help' for more detailed usage information.
</pre>

mem:
> A couple notes from playing around with things.
> 
> * It would be nice if the error indicated what option was invalid in case you are using many options. A good example would be to look at what `git` does.
> 
> [vagrant@localhost singularity]$ git status -r
> error: unknown switch `r'
> usage: git status [options] [--] ...

I think I can grab that from cobra; in fact when I was looking at this, I had some debug output that looked very much like that, and it was simply printing the error it gets from cobra.

> * The usage information does not show the options available, it shows all the sub-commands only. To me that is confusing because it does not show what the proper options are. It may be a better idea to print the output of the help. Note that you can see the difference with `singularity shell` (usage) vs `singularity help shell` (help).

I think I understand what you are saying, but let's double check.

In this case:

```
singularity exec --invalid
```

cobra processes the `exec` command and notes that `--invalid` is an invalid flag, so I could print out the help for `exec`.

In this case:

```
singularity --invalid exec
```

cobra notes that `--invalid` is an invalid global flag, so I could print the help for `singularity` (the one listing all the command and flags, etc).

```
singularity --valid foo exec
```

This is the case that the issue referred to originally. Here cobra is treating `foo` as an invalid command, therefore the error is associated with the `singularity` command, not the `exec` command. In this case I would be printing the help for singularity.

> * Sub-commands have the same issue in that printing the usage does not print the available options

Ok, I think we can fix this by modifying the usage. Cobra has "help" (the detailed description about the command and the flags, etc) and "usage" (the short description).

Does that make sense?

jscook2345:
@mem Sounds good to me!

GodloveD:
@jscook2345 want to take a second look at this and sign off if it is good to go?

jscook2345 (approved):
💛

PR #2821, opened 2019-03-06
Dont Store Public Key When Verifying a Container

### LINKED ISSUE #2579: Verify Should Check for Updates to PGP Public Key
## Version of Singularity:

3.0.2

### Expected behavior

When `singularity verify` is done on a signed container, and the public key used to sign the image is not present in the local key store, it is fetched from the Key Service. A subsequent `singularity verify` that uses the same public key should check with the Key Service that the key has not been updated (or revoked), and act accordingly.

### Actual behavior

Subsequent `singularity verify` commands do not check that the key has not been updated (or revoked.)

Generate a PGP key pair, and sign a container with the private key. Store the public key, and push it to the Key Service. Verify `singularity verify` works. Revoke the key and push an update to the Key Service. Observe that `singularity verify` still works, and that the public key stored in `~/.singularity/sypgp/pgp-public` has not been updated.

WestleyK:

When you verify a container, Singularity will **not** store that public key, (if you want to store it, you need to download it `singularity key pull ...`)

<br>

- Fixes #2579 

<br>


pkg/signing/signing.go
```diff
@@ -311,29 +312,81 @@ func Verify(cpath, url string, id uint32, isGroup bool, authToken string, noProm
 			return fmt.Errorf("could not get the signing entity fingerprint: %s", err)
 		}
 
-		// try to verify with local OpenPGP store first
-		signer, err := openpgp.CheckDetachedSignature(elist, bytes.NewBuffer(block.Bytes), block.ArmoredSignature.Body)
+		//fmt.Printf("fingerptinyOOOOOOOOOO: %v\n", fingerprint)
+
+		// remove the local key that signed the container
+		if err := sypgp.RemovePupKey(fingerprint); err != nil {
```
mem: RemovePubKey? (instead of pup)

pkg/signing/signing.go
```diff
@@ -311,29 +312,81 @@ func Verify(cpath, url string, id uint32, isGroup bool, authToken string, noProm
...
+			return fmt.Errorf("could not fetch public key from server: %s", err)
+		}
+		//sylog.Infof("key retrieved successfully!")
+		//fmt.Printf("netList type: %T\n", netlist)
 
-			block, _ := clearsign.Decode(data)
-			if block == nil {
-				return fmt.Errorf("failed to parse signature block")
-			}
+		block, _ = clearsign.Decode(data)
```
mem: I might be missing something here. Unless the operation on line 324 (or this one) has a side-effect, why does this need to be called in the middle of fetching and storing the public key? Is the fetch making the new public key available to openpgp somehow? If yes, I think a comment in the code would help. If not, can this be moved after the call to sypgp.StorePubKey?

pkg/signing/signing.go
```diff
@@ -311,29 +312,81 @@ func Verify(cpath, url string, id uint32, isGroup bool, authToken string, noProm
...
+		}
+
+		// verify the container
+		signer, err := openpgp.CheckDetachedSignature(netlist, bytes.NewBuffer(block.Bytes), block.ArmoredSignature.Body)
+		if err != nil {
+			return fmt.Errorf("signature verification failed: %s", err)
+		}
 
-			if noPrompt {
+//		sylog.Infof("key missing, searching key server for KeyID: %s...", fingerprint[24:])
```
mem: Is the commented-out code needed? Can it be removed?

pkg/sypgp/sypgp.go
```diff
@@ -349,6 +350,68 @@ func StorePubKey(e *openpgp.Entity) (err error) {
 	return
 }
 
+// compareLocalPubKey : compares a key ID with a string. return true if the
```
mem: Usually godoc strings should read like a sentence (so that when formatting them using go doc it comes out nicely).

For example:

    // compareLocalPubKey compares a key ID with a string, returning true if the
    // key and oldToken do not match.

Also, because of the name of the function, I would expect it to return true if they do match. Please consider changing the name of the function (or inverting the sense of the return value so that it matches the name of the function).

pkg/sypgp/sypgp.go
```diff
@@ -349,6 +350,68 @@ func StorePubKey(e *openpgp.Entity) (err error) {
...
+// key / oldToken *dont* match.
+func compareLocalPubKey(e *openpgp.Entity, oldToken string) bool {
+	if fmt.Sprintf("%X", e.PrimaryKey.Fingerprint) != oldToken {
```
mem: note that you are evaluating a boolean to return a boolean. In other words, you have this structure:

    if condition {
        return true
    }
    return false

that's equivalent to:

    return cond

pkg/sypgp/sypgp.go
```diff
@@ -349,6 +350,68 @@ func StorePubKey(e *openpgp.Entity) (err error) {
...
+		return true
+	}
+	return false
+}
+
+// RemovePupKey : will delete all keys with the ID of (toDelete string)
```
mem: Same comment regarding godoc strings.

pkg/sypgp/sypgp.go
```diff
@@ -349,6 +350,68 @@ func StorePubKey(e *openpgp.Entity) (err error) {
...
+func RemovePupKey(toDelete string) error {
```
mem: So... it's intentional :-)

What's a "pup key"?

pkg/sypgp/sypgp.go
```diff
@@ -349,6 +350,68 @@ func StorePubKey(e *openpgp.Entity) (err error) {
...
+	if err != nil {
+		return fmt.Errorf("unable to read keyring: %v", err)
+	}
+
+	var newKeyList []openpgp.Entity
+
+	// sort throught them, and remove any that match toDelete
+	for i := range elist {
+		// if the elist[i] dose not match toDelete, then add it to newKeyList
+		if compareLocalPubKey(elist[i], toDelete) {
```
mem: this is what I meant before: only looking at this code, I would flag a bug, as it suggests that it's removing everything _but_ whatever matches `toDelete`.

pkg/sypgp/sypgp.go
```diff
@@ -349,6 +350,68 @@ func StorePubKey(e *openpgp.Entity) (err error) {
...
+			newKeyList = append(newKeyList, *elist[i])
+		}
+	}
+
+	// delete all the local public keys... scary :O
+	// to my knowledge, there is no other way to do this...
+	sylog.Infof("Updating local keyring: %v", PublicPath())
+	os.Remove(PublicPath())
```
mem: Uh...

I guess this works, on Linux, at least. On Windows it definitely doesn't. I would expect it to work on Mac OS, too.

Instead of this, since you have all the keys in memory, I would close the file and open it for writing passing O_TRUNC. That saves you from having to try to preserve the permissions.

Because of line 365, I think there's an expectation that the file already exists.

mem (changes requested):

pkg/signing/signing.go
WestleyK: Yes, Theres a lot of commented code that can/will be removed! *(wip)*

pkg/sypgp/sypgp.go
WestleyK: On it! 
And yes, I know the true/false is backwards... fixing that.

pkg/sypgp/sypgp.go
WestleyK: Short for `public key`, as oppose `private key`... I think :thinking:

pkg/signing/signing.go
WestleyK: Typo, good find! :+1:

pkg/signing/signing.go
```diff
@@ -281,14 +281,12 @@ func Verify(cpath, url string, id uint32, isGroup bool, authToken string, noProm
 	// the selected data object is hashed for comparison against signature block's
 	sifhash := computeHashStr(&fimg, descr)
 
-	// load the public keys available locally from the cache
-	elist, err := sypgp.LoadPubKeyring()
-	if err != nil {
-		return fmt.Errorf("could not load public keyring: %s", err)
-	}
+	var authok string
```
jchavezlavalle: typo: authok -> author

pkg/sypgp/sypgp.go
```diff
@@ -349,6 +349,83 @@ func StorePubKey(e *openpgp.Entity) (err error) {
...
+// key and oldToken match.
+func compareLocalPubKey(e *openpgp.Entity, oldToken string) bool {
+	if fmt.Sprintf("%X", e.PrimaryKey.Fingerprint) == oldToken {
+		return true
+	}
+	return false
+}
+
+// CheckLocalPubKey : will check if we have a local public key matching ckey stirng
+// returns true if theres a match.
```
jchavezlavalle: theres -> there's

pkg/sypgp/sypgp.go
```diff
@@ -349,6 +349,83 @@ func StorePubKey(e *openpgp.Entity) (err error) {
...
+func CheckLocalPubKey(ckey string) (bool, error) {
+	//f, err := os.OpenFile(PublicPath(), os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0600)
```
jchavezlavalle: remove commented code

pkg/sypgp/sypgp.go
```diff
@@ -349,6 +349,83 @@ func StorePubKey(e *openpgp.Entity) (err error) {
...
+	// read all the local public keys
+	elist, err := openpgp.ReadKeyRing(f)
+	if err != nil {
+		return false, fmt.Errorf("unable to list local keyring: %v", err)
+	}
+
+	for i := range elist {
+		if compareLocalPubKey(elist[i], ckey) {
+			return true, nil
+		}
```
jchavezlavalle, line 385: lines 378 to 382 could be changed to:

```
return compareLocalPubKey(elist[i], ckey), nil
```

pkg/sypgp/sypgp.go
```diff
@@ -349,6 +349,83 @@ func StorePubKey(e *openpgp.Entity) (err error) {
...
+	for i := range elist {
+		// if the elist[i] dose not match toDelete, then add it to newKeyList
+		if !compareLocalPubKey(elist[i], toDelete) {
+			newKeyList = append(newKeyList, *elist[i])
+		}
+	}
+
+	sylog.Infof("Updating local keyring: %v", PublicPath())
+
+	// truncate all the public keys... scary :O
```
jchavezlavalle: remove this comment :smile:

pkg/sypgp/sypgp.go
```diff
@@ -349,6 +349,83 @@ func StorePubKey(e *openpgp.Entity) (err error) {
...
+	// to my knowledge, there is no other way to do this...
+	nf, err := os.OpenFile(PublicPath(), os.O_TRUNC|os.O_WRONLY, 0600)
+	if err != nil {
+		return fmt.Errorf("unable to clear, and open the file: %v", err)
+	}
+	defer nf.Close()
+
+	// loop throught a write all the other keys back
```
jchavezlavalle: throught -> through

jchavezlavalle (changes requested):
Added some comments @WestleyK thanks!

pkg/signing/signing.go
WestleyK: I'm not sure thats a typo... but I'll change it anyway.

mem: Westley is right, authok is meant to be read as "authorization OK". Maybe "authOk" or "authOK" to avoid confusion?

WestleyK: `authOk` sounds good! I'll change it!

mem: Nope, we are both wrong and @devprincess is right, it's author.

WestleyK: Okay, I'll hold off :+1:

mem (approved):

jchavezlavalle (approved):
looks good thanks @WestleyK

pkg/signing/signing.go
```diff
@@ -358,10 +348,10 @@ func Verify(cpath, url string, id uint32, isGroup bool, authToken string, noProm
 			name = i.Name
 			break
 		}
-		authok += fmt.Sprintf("\t%s, KeyID %X\n", name, signer.PrimaryKey.KeyId)
+		author += fmt.Sprintf("\t%s, KeyID %X\n", name, signer.PrimaryKey.KeyId)
 	}
 	fmt.Printf("Data integrity checked, authentic and signed by:\n")
-	fmt.Print(authok)
+	fmt.Print(author)
```
bauerm97: Let's move this to previous line.

`fmt.Printf("Data integrity checked, authentic and signed by:\n%v", author)`

bauerm97 (dismissed):

pkg/sypgp/sypgp.go
```diff
@@ -351,6 +351,76 @@ func StorePubKey(e *openpgp.Entity) (err error) {
...
+	defer f.Close()
+
+	// read all the local public keys
+	elist, err := openpgp.ReadKeyRing(f)
+	if err != nil {
+		return false, fmt.Errorf("unable to list local keyring: %v", err)
+	}
+
+	for i := range elist {
+		return compareLocalPubKey(elist[i], ckey), nil
```
ikaneshiro: Could there be more than one key in the keyring? It looks like this loop only checks the first one and returns.

ikaneshiro (changes requested):

pkg/sypgp/sypgp.go
WestleyK: Cuckoo -> @WestleyK

ikaneshiro (approved):

PR #2822, opened 2019-03-06
Compile plugins using -trimpath flag

### LINKED ISSUE #2798: Remove plugin compile command container magic
A fix has been found for compilation of plugins with different values of `$GOPATH`. We should update `mconfig` to pass the appropriate `-trimpath` flag to the Go compiler:

```
mibauer@ubuntu:~/modules-plugins/main$ pwd
/home/mibauer/modules-plugins/main
mibauer@ubuntu:~/modules-plugins/main$ go clean -cache
mibauer@ubuntu:~/modules-plugins/main$ go build -buildmode=plugin -gcflags=all=-trimpath=/home/mibauer/modules-plugins -asmflags=all=-trimpath=/home/mibauer/modules-plugins p1
mibauer@ubuntu:~/modules-plugins/main$ go build -buildmode=plugin -gcflags=all=-trimpath=/home/mibauer/modules-plugins -asmflags=all=-trimpath=/home/mibauer/modules-plugins p2
mibauer@ubuntu:~/modules-plugins/main$ ls
build-vx  dumpdep  go.mod  go.sum  main.go  p1.so  p2.so  pkg  tools  vendor
mibauer@ubuntu:~/modules-plugins/main$ cp p1.so ~/modules-plugins-path2/main/
mibauer@ubuntu:~/modules-plugins/main$ cp p2.so ~/modules-plugins-path2/main/
mibauer@ubuntu:~/modules-plugins/main$ cd ~/modules-plugins-path2/main/
mibauer@ubuntu:~/modules-plugins-path2/main$ go build .
mibauer@ubuntu:~/modules-plugins-path2/main$ ./main 
panic: plugin.Open("p1"): plugin was built with a different version of package gopkg.in/yaml.v2

goroutine 1 [running]:
main.main()
    /home/mibauer/modules-plugins-path2/main/main.go:13 +0x229
mibauer@ubuntu:~/modules-plugins-path2/main$ rm main
mibauer@ubuntu:~/modules-plugins-path2/main$ go build -gcflags=all=-trimpath=/home/mibauer/modules-plugins-path2 -asmflags=all=-trimpath=/home/mibauer/modules-plugins-path2 .
mibauer@ubuntu:~/modules-plugins-path2/main$ ./main 
p1: Type: yaml.Encoder
p2: Type: yaml.Encoder
yaml.Encoder
mibauer@ubuntu:~/modules-plugins-path2/main$
```

Specifically, `mconfig` should pass flags `-gcflags=all=-trimpath=$(SOURCEDIR) -asmflags=all=-trimpath=$(SOURCEDIR)`. This means we no longer need to compile plugins AND singularity in a container from the same absolute `$GOPATH`. As long as the appropriate `-trimpath` flags are specified (will be handled by `go plugin compile` and `mconfig`), plugins will be portable.

### LINKED ISSUE #2803: Update plugin compile command to no longer compile plugin in container
This is 1/2 of a fix for #2798

Currently `singularity plugin compile` is running a `go build [...]` command inside of a container that has been carefully crafted to ensure a portable value of `$(GOPATH)`. The plugin compile command should be updated to no longer need to compile in a container, instead passing the appropriate flags to `go build`:

`-gcflags=all=-trimpath=$(PLSOURCEDIR) -asmflags=all=-trimpath=$(PLSOURCEDIR)`

bauerm97:

This patch updates the plugin compile command to use `-trimpath` appropriately. After we merge this, plugin compiling will be portable.

- Fixes #2803 
- Fixes #2798 


internal/app/singularity/plugin_compile_linux.go
```diff
@@ -7,18 +7,27 @@ package singularity
...
 	"runtime"
+	"strings"
 
 	uuid "github.com/satori/go.uuid"
 	"github.com/sylabs/sif/pkg/sif"
 )
 
+var (
+	workpath   = filepath.Join(build.Default.GOPATH, repo)
+	trimpath   = filepath.Dir(workpath)
```
mem, line 23: Isn't this too much? If build.Default.GOPATH = "/home/user/go", workpath = "/home/user/go/src/github.com/sylabs/singularity" and trimpath is "/home/user/go/src/github.com/sylabs".

I would have expected that trimming "/home/user/go" or "/home/user/go/src" would have been enough. What's the supported use case?

Also note that GOPATH is a PATH, meaning it can contain multiple directories separated by colons.

internal/app/singularity/plugin_compile_linux.go
```diff
@@ -53,20 +69,38 @@ func CompilePlugin(sourceDir, destSif string) error {
...
-
-	scmd.Stderr = os.Stderr
-	scmd.Stdout = os.Stdout
-	scmd.Stdin = os.Stdin
-	return scmd.Run()
+func buildPlugin(sourceDir string) (string, error) {
+	// assuming that sourceDir is within trimpath for now
+	out := pluginObjPath(sourceDir)
+	c := fmt.Sprintf(buildtmpl, trimpath, out, sourceDir)
+	buildcmd := exec.Command("go", strings.Split(c, " ")...)
```
mem, line 79: I haven't had a chance to go thru this, but I have the hunch that this is missing build tags. I think when building singularity we are passing a series of build tags, and I assume it's possible to tweak those. I'm guessing the plugin needs to be built with the same set of tags.

internal/app/singularity/plugin_compile_linux.go
```diff
@@ -53,20 +69,38 @@ func CompilePlugin(sourceDir, destSif string) error {
...
+	return out, buildcmd.Run()
+}
+
+// generateManifest takes the path to the plugin source, sourceDir, and generates
+// its corresponding manifest file.
+func generateManifest(sourceDir string) (string, error) {
+	in := pluginObjPath(sourceDir)
+	out := pluginManifestPath(sourceDir)
+
+	c := fmt.Sprintf(manifesttmpl, trimpath, mangenpath, in+" "+out)
```
mem: why not pass in+" "+out as two separate arguments?

mem (commented):
It's more questions than requests for changes. I think it's ok as it is, but there are some strange corner cases to consider.

internal/app/singularity/plugin_compile_linux.go
bauerm97: >Isn't this too much? If build.Default.GOPATH = "/home/user/go", workpath = "/home/user/go/src/github.com/sylabs/singularity" and trimpath is "/home/user/go/src/github.com/sylabs".
>I would have expected that trimming "/home/user/go" or "/home/user/go/src" would have been enough. What's the supported use case?

Yes, I think that trimming `$(GOPATH)/src` is actually probably enough. However, currently in the upstream code we are trimming `dirname $(SOURCEDIR)` from the singularity builds. I would vote we merge this as is to get something up and running, and revisit that in a follow-up PR.

> Also note that GOPATH is a PATH, meaning it can contain multiple directories separated by colons.

Huh, I guess I technically knew that but figured that `build.Default.GOPATH` would properly return one path rather than a list, but it looks like that's not the case.

internal/app/singularity/plugin_compile_linux.go
bauerm97: Actually, the plugins don't need to be built with the same tags. You can verify this by compiling and installing Singularity, then compiling the example plugin using `singularity plugin compile $(GOPATH)/src/github.com/sylabs/singularity/examples/plugins/test-plugin` and then doing a `sudo cp test-plugin.so $(PREFIX)/libexecdir/plugin`. You'll see that in the current state, singularity is compiled with up to a few tags, and even though the plugin isn't it will still work

internal/app/singularity/plugin_compile_linux.go
bauerm97: I have no reasonable answer to this question

internal/app/singularity/plugin_compile_linux.go
mem: Ok, that makes sense.

internal/app/singularity/plugin_compile_linux.go
mem: That's surprising. I'm guessing the build flags we are using are not changing the exposed API but instead they only changing the implementation of some functions. Either we make that a promise (build tags won't change the API) or we make sure plugins are built with the same tags.

We can punt this for now.

mem (approved):

WestleyK (dismissed):
The copyright year should be `2019`. We should be careful when copy-paste-ing copyrights :wink:

PR #2826, opened 2019-03-07
Make NVIDIA utils public to reuse in device plugin

sashayakovtseva:


pkg/util/nvidia/nvidia.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	// get elf machine to match correct libraries during ldconfig lookup
+	self, err := elf.Open("/proc/self/exe")
+	if err != nil {
+		return nil, nil, fmt.Errorf("could not open /proc/self/exe: %v", err)
+	}
+
+	machine := self.Machine
+	if err := self.Close(); err != nil {
```
cclerget, line 124: if there is no write involved there is no need to check close error

pkg/util/nvidia/nvidia.go
```diff
@@ -0,0 +1,185 @@
...
+						sylog.Debugf("ignore library %s: %s", libName, err)
+						continue
+					}
+
+					if elib.Machine == machine {
+						libs[libName] = struct{}{}
+						libraries = append(libraries, libPath)
+					}
+
+					if err := elib.Close(); err != nil {
```
cclerget, line 165: same here

cclerget (approved):
LGTM

Issue #2774, opened 2019-02-25
Rearrange auth-token handling


bauerm97:
Currently auth-tokens are handled in a `cobra.Cmd` `PreRun` function. This causes auth tokens to always be searched for during any command which could *ever* need an auth token. We should rearrange auth-token management into a package `pkg/auth` (?) so that in `internal/app/singularity` we can grab an auth token at a specific time.

bauerm97:
Related to #2687

ikaneshiro:
A lot of the `authtoken` behavior is tightly coupled with the CLI and changing it would require a CLI change which should be done over the course of a couple minor releases.

I propose the following steps:
	- we integrate the remote endpoint management into the CLI leaving much of the `authtoken` behavior as is
	- Depriciate `tokenfile`, `tokenfile` flag and `SYLABS_TOKEN` in 3.2
	- Remove these flags completely in 3.3 and have remote be the sole manager of endpoints and tokens apart from subcommand specific CLI flags

@tri-adam @bauerm97 Thoughts?

tri-adam:
I like that approach!

ikaneshiro:
After second thought I think leaving `SYLABS_TOKEN` envvar is ok. I can see a case where having it stored in the environment would be easier to work with

PR #2829, opened 2019-03-07
SCS Remote Management

### LINKED ISSUE #2687: SCS Endpoint Management
Currently, working with an instance of the Singularity Cloud Services (SCS) other than the one at `cloud.sylabs.io` is difficult and tedious. Many of the associated commands require multiple options to be passed in order to set URLs and the associated API token.

For example, to do a build against `build.staging.sylabs.io`:

```sh
$ singularity --tokenfile staging-token build --builder https://build.staging.sylabs.io --library https://library.staging.sylabs.io image.sif recipe.txt
``` 

As a way to simplify this, I propose we introduce a `singularity remote` set of commands. These commands will allow a user to assign a single, human-readable name that maps to a set of API endpoints and an associated API token.

In this way, one could add a new set of SCS endpoints like so:

```sh
$ singularity remote add staging cloud.staging.sylabs.io
```

We should have an easy way of generating the API key (command TBD) something like:

```sh
$ singularity remote login staging
Generate an API Key at https://cloud.staging.sylabs.io/api-key, and paste here:

API Key: <PASTE>

Verifying API Key… OK!
```

It'd be nice to be able to see service and API key information:

```sh
$ singularity remote status staging
Status report for staging (cloud.staging.sylabs.io):
SERVICE            STATUS   VERSION
Auth Service:      OK	   [v1.0.0-alpha.5-0-gefb1e63]
Build Service:     N/A
Container Library: N/A
Key Service:       OK       [v1.3.0-0-gaff1db5]

API Key: valid!
```

Finally, you could have more than one `remote` configured:

```sh
$ singularity remote list
NAME        URI
[cloud]     cloud.sylabs.io
staging     cloud.staging.sylabs.io
```

And switch between them:

```sh
$ singularity remote use staging
$ singularity remote list
NAME        URI
cloud       cloud.sylabs.io
[staging]   cloud.staging.sylabs.io
```

ikaneshiro:

This PR introduces the `remote` command and package which are used to manage multiple SCS remote endpoints and tokens.

Having this work in conjunction with current `authtoken` behavior has been really tricky. I have opted for using `remotes` taking precedence if there is one set as default. Otherwise normal `authtoken` and CLI flag behavior remains. This will mean that there will be no change in CLI behavior(other than potentially some debug or warning messages) if remotes are never set. I would like to move away from current `authtoken` in the future and have expressed more thoughts on #2774  

Includes:
- Unit testing for `remote` package
- e2e testing for `remote` command 
- Documentation of `remote` command

Note: Does not have tests for `login` command or `ValidateToken()` function because of trouble using valid tokens in CI.

- Fixes #2687 


internal/app/singularity/remote_list.go
```diff
@@ -0,0 +1,55 @@
...
+	"text/tabwriter"
+
+	"github.com/sylabs/singularity/internal/pkg/remote"
+)
+
+const listLine = "%s\t%s\n"
+
+// RemoteList prints information about remote configurations
+func RemoteList(configFile string) (err error) {
+
```
bauerm97: Remove empty line

docs/content.go
```diff
@@ -901,4 +901,15 @@ found at:
   Umount will umount an OCI bundle previously mounted with singularity oci mount.`
 	OciUmountExample string = `
   $ singularity oci umount /var/lib/singularity/bundles/example`
+
```
bauerm97: Move into `docs/remote.go` and add a block for each remote subcommand

cmd/internal/cli/push.go
```diff
@@ -32,6 +33,25 @@ var PushCmd = &cobra.Command{
 	Args:                  cobra.ExactArgs(2),
 	PreRun:                sylabsToken,
 	Run: func(cmd *cobra.Command, args []string) {
+
```
bauerm97: Remove empty line

cmd/internal/cli/key_push.go
```diff
@@ -29,6 +30,25 @@ var KeyPushCmd = &cobra.Command{
 	DisableFlagsInUseLine: true,
 	PreRun:                sylabsToken,
 	Run: func(cmd *cobra.Command, args []string) {
+
```
bauerm97: Remove empty line

cmd/internal/cli/key_pull.go
```diff
@@ -28,6 +29,23 @@ var KeyPullCmd = &cobra.Command{
 	DisableFlagsInUseLine: true,
 	PreRun:                sylabsToken,
 	Run: func(cmd *cobra.Command, args []string) {
+
```
bauerm97: Remove empty line

cmd/internal/cli/push.go
```diff
@@ -32,6 +33,25 @@ var PushCmd = &cobra.Command{
...
+				uri, err := e.GetServiceURI("library")
+				if err != nil {
+					sylog.Fatalf("Unable to get library service URI: %v", err)
+				}
+				PushLibraryURI = uri
+			}
+		} else if err == scs.ErrNoDefault {
+			sylog.Warningf("No default remote in use, falling back to: %v", PushLibraryURI)
+		} else {
+			sylog.Debugf("Unable to load remote configuration: %v", err)
```
bauerm97: If we're unable to load the remote configuration in some way (other than `ErrNoDefault`), I think we should fail rather than a debug message and continuing. If a user runs into this, they'll typically not be running with `--debug` and might not even realize there's something going on. I think this should apply to all new commands which are using this

internal/pkg/remote/remote.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	c.Remotes[newName] = c.Remotes[name]
+	delete(c.Remotes, name)
+	return nil
+}
+
+// returns an error if a token is not valid
+func (e *EndPoint) VerifyToken() error {
+
```
bauerm97: Remove empty line

bauerm97 (changes requested):
Honestly looking really great so far. I was only able to find a few minor nit-picks. Also there appears to be no testing

cmd/internal/cli/build.go
```diff
@@ -224,3 +225,53 @@ func makeDockerCredentials(cmd *cobra.Command) (authConf *ocitypes.DockerAuthCon
 
 	return
 }
+
+// remote builds need to fail if we cannot resolve remote URLS
+func handleRemoteBuildFlags(cmd *cobra.Command) {
+	// if we can load config and if default endpoint is set, use that
+	// otherwise fall back on regular authtoken and URI behavior
+	e, err := sylabsRemote(remoteConfig)
+	if err == nil {
```
bauerm97: Can we do this instead:

```
if err == scs.ErrNoDefault {
    sylog.Warningf(...)
    return
} else if err != nil {
    sylog.Fatalf(...)
}

[...]
```

This way we avoid having things living too deep inside of conditional blocks

cmd/internal/cli/singularity.go
```diff
@@ -207,6 +208,22 @@ func sylabsToken(cmd *cobra.Command, args []string) {
 	}
 }
 
+// sylabsRemote returns the remote in use or an error
+func sylabsRemote(filepath string) (*scs.EndPoint, error) {
+	file, err := os.OpenFile(filepath, os.O_RDONLY|os.O_CREATE, 0644)
+	if err != nil {
+		return nil, err
```
bauerm97: Should we add some context to this error using `fmt.Errorf(...)`?

cmd/internal/cli/singularity.go
```diff
@@ -207,6 +208,22 @@ func sylabsToken(cmd *cobra.Command, args []string) {
...
+	}
+	defer file.Close()
+
+	c, err := scs.ReadFrom(file)
+	if err != nil {
+		return nil, err
```
bauerm97: Same comment here, I think in general we should try to add context to error messages whenever possible

cmd/internal/cli/singularity.go
```diff
@@ -207,6 +208,22 @@ func sylabsToken(cmd *cobra.Command, args []string) {
...
+	}
+
+	return c.GetDefault()
```
bauerm97, line 224: Here I understand that we don't want to do that because the `GetDefault()` function will at times return `scs.ErrNoDefault`

cmd/internal/cli/build.go
```diff
@@ -224,3 +225,53 @@ func makeDockerCredentials(cmd *cobra.Command) (authConf *ocitypes.DockerAuthCon
...
+		sylog.Fatalf("Unable to load remote configuration: %v", err)
+	}
+}
+
+// standard builds should just warn and fall back to CLI default if we cannot resolve library URL
+func handleBuildFlags(cmd *cobra.Command) {
+	// if we can load config and if default endpoint is set, use that
+	// otherwise fall back on regular authtoken and URI behavior
+	e, err := sylabsRemote(remoteConfig)
+	if err == nil {
```
bauerm97: Here same comment as in `handleRemoteBuildFlags`. I think in general the fewer and shallower conditionals we use the better off we will be in terms of readability of code

cmd/internal/cli/push.go
```diff
@@ -32,6 +33,24 @@ var PushCmd = &cobra.Command{
 	Args:                  cobra.ExactArgs(2),
 	PreRun:                sylabsToken,
 	Run: func(cmd *cobra.Command, args []string) {
+		// if we can load config and if default endpoint is set, use that
```
bauerm97: Not sure how to fix this, but it really looks like there's a lot of code that's duplicated here in a few spots, is it possible for us to de-duplicate somehow?

cmd/internal/cli/pull.go
```diff
@@ -121,6 +122,24 @@ func pullRun(cmd *cobra.Command, args []string) {
 			}
 		}
 
+		// if we can load config and if default endpoint is set, use that
```
bauerm97: Notice how similar this is here

cmd/internal/cli/key_push.go
```diff
@@ -29,6 +30,24 @@ var KeyPushCmd = &cobra.Command{
 	DisableFlagsInUseLine: true,
 	PreRun:                sylabsToken,
 	Run: func(cmd *cobra.Command, args []string) {
+		// if we can load config and if default endpoint is set, use that
```
bauerm97: And here

cmd/internal/cli/key_pull.go
```diff
@@ -28,6 +29,22 @@ var KeyPullCmd = &cobra.Command{
 	DisableFlagsInUseLine: true,
 	PreRun:                sylabsToken,
 	Run: func(cmd *cobra.Command, args []string) {
+		// If flag did not change, look for a remote configuration, otherwise fall back to default
```
bauerm97: And here. It might be possible to move this into a shared function, possibly taking a function argument to handle the different needs in each area.

Maybe something like:

```
func handleFlag(name string, fn func(bool)) {
    ...
}
```

That function could take the bool given by `cmd.Flags().Lookup("url").Changed`, or could take no arguments and just be run inside of that block. This function definitely needs to be fleshed out, but the idea is there maybe. Just something to keep in mind to maybe reduce duplication of efforts

internal/app/singularity/remote_login.go
```diff
@@ -0,0 +1,67 @@
...
+	r.Token, err = sypgp.AskQuestionNoEcho("API Key: ")
+	if err != nil {
+		return err
+	}
+
+	if err := r.VerifyToken(); err != nil {
+		return fmt.Errorf("while verifying token: %v", err)
+	}
+
+	sylog.Infof("API Key Verified!")
```
bauerm97, line 49: This is too excited

bauerm97 (changes requested):

internal/app/singularity/remote_login.go
ikaneshiro: Boo!

internal/app/singularity/remote_list.go
```diff
@@ -0,0 +1,54 @@
...
+const listLine = "%s\t%s\n"
+
+// RemoteList prints information about remote configurations
+func RemoteList(configFile string) (err error) {
+	c := &remote.Config{}
+	file, err := os.OpenFile(configFile, os.O_RDONLY|os.O_CREATE, 0600)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return fmt.Errorf("No Remote configurations")
```
WestleyK: I think `error` massages are not suppose to have caps, so:

```suggestion
			return fmt.Errorf("no remote configurations")
```

internal/app/singularity/remote_login.go
```diff
@@ -0,0 +1,67 @@
...
+// rights to use or distribute this software.
+
+package singularity
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/sylabs/singularity/internal/pkg/sylog"
+
```
WestleyK: Extra `\n`

internal/app/singularity/remote_status.go
```diff
@@ -0,0 +1,132 @@
...
+	req, err := http.NewRequest(http.MethodGet, url+"/version", nil)
+	if err != nil {
+		return "", err
+	}
+
+	req.Header.Set("User-Agent", useragent.Value())
+
+	res, err := client.Do(req)
+	if err != nil {
+		return "", fmt.Errorf("error making request to server:\n\t%v", err)
```
WestleyK: I think error messages are not supposed to have newlines, or tabs, etc... But I could be wrong...

internal/app/singularity/remote_use.go
```diff
@@ -0,0 +1,51 @@
...
+	c, err = remote.ReadFrom(file)
+	if err != nil {
+		return fmt.Errorf("while parsing remote config data: %s", err)
+	}
+
+	if err := c.SetDefault(name); err != nil {
+		return err
+	}
+
+	if err := file.Truncate(0); err != nil {
```
WestleyK, line 37: I would add some comments here, like:

```
// truncate the file...
```
...
```
// open file config file...
```

Just to help people (like me) understand whats going on here.

internal/pkg/remote/remote.go
```diff
@@ -0,0 +1,262 @@
...
+	c.Remotes[newName] = c.Remotes[name]
+	delete(c.Remotes, name)
+	return nil
+}
+
+// VerifyToken returns an error if a token is not valid
+func (e *EndPoint) VerifyToken() error {
+	baseURL, err := e.GetServiceURI("token")
+	if err != nil {
+		return fmt.Errorf("While getting token service uri: %v", err)
```
WestleyK:
```suggestion
		return fmt.Errorf("while getting token service uri: %v", err)
```

WestleyK (dismissed):
Looks amazing!!! Have a couple suggestions :+1:

cmd/internal/cli/key_push.go
```diff
@@ -92,3 +76,23 @@ func doKeyPushCmd(fingerprint string, url string) error {
 
 	return nil
 }
+
+func handleKeyFlags(cmd *cobra.Command) {
+	// if we can load config and if default endpoint is set, use that
+	// otherwise fall back on regular authtoken and URI behavior
+	e, err := sylabsRemote(remoteConfig)
+	if err == scs.ErrNoDefault {
+		sylog.Warningf("No default remote in use, falling back to: %v", keyServerURL)
```
bauerm97, line 85: Should we be returning here?

bauerm97 (changes requested):

internal/app/singularity/remote_remove.go
```diff
@@ -0,0 +1,51 @@
...
+package singularity
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/sylabs/singularity/internal/pkg/remote"
+)
+
+// RemoteRemove deletes a remote remote from the configuration
```
bauerm97: `deletes a remote remote`

hmm

internal/app/singularity/remote_use.go
bauerm97: Actually, in this case I would say that the intention of the code is clear enough from the function names.

cmd/internal/cli/key_push.go
ikaneshiro: yes

internal/app/singularity/remote_remove.go
ikaneshiro: what what?

internal/app/singularity/remote_status.go
ikaneshiro: Youre correct :+1:

bauerm97 (approved):

WestleyK (approved):
Looks Good! :+1:

PR #2834, opened 2019-03-07
adding argument --nocolor to disable sylog printing of color

### LINKED ISSUE #2832: Add flag to disable color output
Reported on the google group, the user wants to write to a file without the color formatting:

```
Hi Singularities,

I piped my --debug to a file as in:
/xyz/singularity/3.1.0/bin/singularity --debug run my_container.sif > & debug_singularity.log

It includes the ANSI characters that are responsible for the pretty colors to my terminal.  However, in the file they are annoying.

May I request that singularity --debug distinguish and change its behavior accordingly?
If that makes sense to do, you might consider an option like what `ls` provides '--color=always'

Reference:
https://unix.stackexchange.com/questions/10823/where-do-my-ansi-escape-codes-go-when-i-pipe-to-another-process-can-i-keep-them

If this is something worth considering I'd be happy to create a github issue but I wanted to check in with you all first.

Robert 
```
I'll give this a shot.

vsoch:
Signed-off-by: Vanessa Sochat <vsochat@stanford.edu>

This pull request will give the user the option to add `--nocolor` to the singularity command to disable color formatting of output. For example, here we can pull a container (without the flag):

```bash
$ singularity --debug pull docker://busybox >& container-color.log
```
The container-color.log is a mess: 

```
^[[0mDEBUG   ^[[0m[U=1000,P=355]     NewBundle()                   Created temporary directory for bundle /tmp/sbuild-878156988
^[[34mINFO    ^[[0m[U=1000,P=355]     Full()                        Starting build...
^[[0mDEBUG   ^[[0m[U=1000,P=355]     Get()                         Reference: busybox
^[[0mDEBUG   ^[[0m[U=1000,P=355]     updateCacheSubdir()           Caching directory set to /home/vanessa/.singularity/cache/oci
Skipping fetch of repeat blob sha256:697743189b6d255069caf6c455be10c7f8cae8076c6f94d224ae15cd41420e87
Copying config sha256:85dcb8c21e86b3e01a3a45fc932b31f5194c174d0a5d8a993381cc1fca749f40
^M 0 B / 575 B [-----------------------------------------------------------------]^M 575 B / 575 B [============================================================] 0s
Storing signatures
^[[0mDEBUG   ^[[0m[U=1000,P=355]     Full()                        Inserting Metadata
^[[0mDEBUG   ^[[0m[U=1000,P=355]     Full()                        Calling assembler
^[[34mINFO    ^[[0m[U=1000,P=355]     Assemble()                    Creating SIF file...
^[[34mINFO    ^[[0m[U=1000,P=355]     Full()                        Build complete: busybox_latest.sif
^[[0mDEBUG   ^[[0m[U=1000,P=355]     cleanUp()                     Build bundle cleanup: /tmp/sbuild-878156988
```
When we add the flag

```bash
$ singularity --debug --nocolor pull docker://busybox >& container.log
```
It's much cleaner.

```
WARNING [U=1000,P=316]     sylabsToken()                 Authentication token file not found : Only pulls of public images will succeed
DEBUG   [U=1000,P=316]     NewBundle()                   Created temporary directory for bundle /tmp/sbuild-751271701
INFO    [U=1000,P=316]     Full()                        Starting build...
DEBUG   [U=1000,P=316]     Get()                         Reference: busybox
DEBUG   [U=1000,P=316]     updateCacheSubdir()           Caching directory set to /home/vanessa/.singularity/cache/oci
Skipping fetch of repeat blob sha256:697743189b6d255069caf6c455be10c7f8cae8076c6f94d224ae15cd41420e87
Copying config sha256:85dcb8c21e86b3e01a3a45fc932b31f5194c174d0a5d8a993381cc1fca749f40
^M 0 B / 575 B [-----------------------------------------------------------------]^M 575 B / 575 B [============================================================] 0s
Storing signatures
DEBUG   [U=1000,P=316]     Full()                        Inserting Metadata
DEBUG   [U=1000,P=316]     Full()                        Calling assembler
INFO    [U=1000,P=316]     Assemble()                    Creating SIF file...
INFO    [U=1000,P=316]     Full()                        Build complete: busybox_latest.sif
DEBUG   [U=1000,P=316]     cleanUp()                     Build bundle cleanup: /tmp/sbuild-751271701

```

## Points for discussion:

 - Naming of variables
 - If a boolean is appropriate (or something more similar to color=always)
 - If there are other spots in the code that color output that I didn't see?

- Fixes #2832 

### DIFF
--- a/cmd/internal/cli/singularity.go
+++ b/cmd/internal/cli/singularity.go
@@ -104,7 +104,6 @@ func setSylogMessageLevel(cmd *cobra.Command, args []string) {
 }
 
 func setSylogColorOn(cmd *cobra.Command, args []string) {
-
 	if nocolor {
 		sylog.DisableColor()
 	}
--- a/cmd/internal/cli/singularity.go
+++ b/cmd/internal/cli/singularity.go
@@ -27,6 +27,7 @@ import (
 // Global variables for singularity CLI
 var (
 	debug   bool
+	nocolor bool
 	silent  bool
 	verbose bool
 	quiet   bool
@@ -71,6 +72,7 @@ func init() {
 	defaultTokenFile = path.Join(usr.HomeDir, ".singularity", "sylabs-token")
 
 	SingularityCmd.Flags().BoolVarP(&debug, "debug", "d", false, "print debugging information (highest verbosity)")
+	SingularityCmd.Flags().BoolVar(&nocolor, "nocolor", false, "print without color output (default False)")
 	SingularityCmd.Flags().BoolVarP(&silent, "silent", "s", false, "only print errors")
 	SingularityCmd.Flags().BoolVarP(&quiet, "quiet", "q", false, "suppress normal output")
 	SingularityCmd.Flags().BoolVarP(&verbose, "verbose", "v", false, "print additional information")
@@ -99,6 +101,13 @@ func setSylogMessageLevel(cmd *cobra.Command, args []string) {
 	}
 
 	sylog.SetLevel(level)
+}
+
+func setSylogColorOn(cmd *cobra.Command, args []string) {
+
+	if nocolor {
+		sylog.DisableColor()
+	}
 }
 
 // SingularityCmd is the base command when called without any subcommands
@@ -178,6 +187,7 @@ func handleEnv(flag *pflag.Flag) {
 
 func persistentPreRun(cmd *cobra.Command, args []string) {
 	setSylogMessageLevel(cmd, args)
+	setSylogColorOn(cmd, args)
 	updateFlagsFromEnv(cmd)
 }
 
--- a/cmd/internal/cli/singularity.go
+++ b/cmd/internal/cli/singularity.go
@@ -103,7 +103,7 @@ func setSylogMessageLevel(cmd *cobra.Command, args []string) {
 	sylog.SetLevel(level)
 }
 
-func setSylogColorOn(cmd *cobra.Command, args []string) {
+func setSylogColor(cmd *cobra.Command, args []string) {
 	if nocolor {
 		sylog.DisableColor()
 	}
@@ -186,7 +186,7 @@ func handleEnv(flag *pflag.Flag) {
 
 func persistentPreRun(cmd *cobra.Command, args []string) {
 	setSylogMessageLevel(cmd, args)
-	setSylogColorOn(cmd, args)
+	setSylogColor(cmd, args)
 	updateFlagsFromEnv(cmd)
 }
 
--- a/internal/pkg/sylog/sylog.go
+++ b/internal/pkg/sylog/sylog.go
@@ -61,12 +61,9 @@ var messageColors = map[messageLevel]string{
 	info:  "\x1b[34m",
 }
 
-const colorReset string = "\x1b[0m"
+var colorReset string = "\x1b[0m"
 
 var loggerLevel messageLevel
-
-// By default, color is enabled
-var disableColor = false
 
 func init() {
 	_level, ok := os.LookupEnv("SINGULARITY_MESSAGELEVEL")
@@ -90,11 +87,7 @@ func prefix(level messageLevel) string {
 
 	// This section builds and returns the prefix for levels < debug
 	if loggerLevel < debug {
-		if disableColor {
-			return fmt.Sprintf("%-8s ", level.String()+":")
-		}
 		return fmt.Sprintf("%s%-8s%s ", messageColor, level.String()+":", colorReset)
-
 	}
 
 	pc, _, _, ok := runtime.Caller(3)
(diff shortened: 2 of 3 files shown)


cmd/internal/cli/singularity.go
```diff
@@ -101,6 +103,13 @@ func setSylogMessageLevel(cmd *cobra.Command, args []string) {
 	sylog.SetLevel(level)
 }
 
+func setSylogColorOn(cmd *cobra.Command, args []string) {
+
+	if nocolor {
```
vsoch, line 107: Should I remove extra space here?

cmd/internal/cli/singularity.go
```diff
@@ -178,6 +187,7 @@ func handleEnv(flag *pflag.Flag) {
 
 func persistentPreRun(cmd *cobra.Command, args []string) {
 	setSylogMessageLevel(cmd, args)
+	setSylogColorOn(cmd, args)
```
vsoch: Is there a better name for this function?

internal/pkg/sylog/sylog.go
```diff
@@ -159,6 +170,11 @@ func SetLevel(l int) {
 	loggerLevel = messageLevel(l)
 }
 
+// DisableColor for the logger
+func DisableColor() {
```
bauerm97, line 164: It's probably easier in this function to just do this:

```
messageColors = map[messageLevel]string{
	fatal: "",
	error: "",
	warn: "",
	info: "",
}

colorReset = ""
```

This way we don't need all of the extra conditionals scattered in the code.

cmd/internal/cli/singularity.go
```diff
@@ -101,6 +103,13 @@ func setSylogMessageLevel(cmd *cobra.Command, args []string) {
 	sylog.SetLevel(level)
 }
 
+func setSylogColorOn(cmd *cobra.Command, args []string) {
+
```
bauerm97: Remove empty line

bauerm97 (changes requested):
I think just a few small tweaks and it'll probably be good to go after that

vsoch (commented):
Removed empty line, and simplified function to edit messageColors instead of adding a new boolean.

cmd/internal/cli/singularity.go
vsoch: Done.

internal/pkg/sylog/sylog.go
vsoch: I like this better too.

internal/pkg/sylog/sylog.go
```diff
@@ -61,7 +61,7 @@ var messageColors = map[messageLevel]string{
 	info:  "\x1b[34m",
 }
 
-const colorReset string = "\x1b[0m"
+var colorReset string = "\x1b[0m"
```
vsoch: Note that the colorReset needed to be changed to a var from constant here.

cmd/internal/cli/singularity.go
bauerm97: I would just remove the `On` personally: `setSylogColor`

vsoch: Done.

bauerm97:
Hey @vsoch , the linter is still complaining at you :(

`/go/src/github.com/sylabs/singularity/internal/pkg/sylog/sylog.go:64:16: should omit type string from declaration of var colorReset; it will be inferred from the right-hand side`

To make sure you're catching all of these errors locally, you'll need to be running Go 1.11. Go 1.10 has slightly different linting/vetting rules, but we enforce Go 1.11's.

vsoch:
@bauerm97 how do I run the linter locally?

vsoch:
That's weird, I had only changed that line to "var" from const, I guess the linter doesn't mind having the type declaration for a constant! Let me fix this up.

vsoch:
I'll work on updating my version of go and installing the linter.

bauerm97:
@vsoch No worries, I think everyone who has submitted a Golang patch to the repo has had this happen like 15 times :P

bauerm97:
@vsoch 

```
./mconfig
make -C builddir/ test-all
```

vsoch:
Ah, that's the trick! I needed to remove the slash, is that correct (or is this my old version of golang?)

```
$ make -C builddir/ test all
```

bauerm97:
Sorry, I gave you a slightly incorrect command.

You want `make -C builddir/ testall`

bauerm97 (approved):

DrDaveD:
Why not make this the default when the log output is to a file as a opposed to a tty?  You can use !terminal.IsTerminal() like on [this line of code](https://github.com/sylabs/singularity/blob/master/internal/pkg/runtime/engines/singularity/container.go#L1000).

vsoch:
Oh cool idea! Yeah let me mess around with that.

vsoch:
There also seems to be a bug with the way that we changed it - not all the levels are caught and the default goes to a particular color (without the ending string so its not colored, but still wrong in the text file).

vsoch:
Just fixed. Will open PR soon.

PR #2836, opened 2019-03-07
Implement plugin install and uninstall functionality

### LINKED ISSUE #2317: singularity plugin command for administration of plugins
In order to provide an easy way for system administrators to manage the plugins installed on their system, a `singularity plugin` command will be implemented. This command will contain several subcommands related to plugin management:

- `enable`/`disable`
- `install`/`uninistall`
- `configure`
- `list`
- `info`

etc...

### LINKED ISSUE #2321: Plugin install and uninstall commands
Plugin administration commands:

- `singularity plugin install <plugin path>`
- `singularity plugin uninstall <name>`

Still some questions to be answered: 

- Should install take a file and name as arguments, or just a file and have the plugin's name be stored inside of the plugin SIF file?
- Should we reference the SIF file name or plugin name when doing `singularity plugin uninstall`?

### LINKED ISSUE #2324: Plugin list command
`singularity plugin list`

Provides a list of names of installed plugins

mem:
Implement plugin install / uninstall and list subcommands

- Fixes #2321 
- Fixes #2324


cmd/internal/cli/plugin_install_linux.go
```diff
@@ -28,9 +28,11 @@ func init() {
 // singularity plugin install <path> [-n name]
 var PluginInstallCmd = &cobra.Command{
 	RunE: func(cmd *cobra.Command, args []string) error {
-		fmt.Println("Installing plugin")
-
-		return singularity.InstallPlugin(args[0], buildcfg.LIBEXECDIR)
+		err := singularity.InstallPlugin(args[0], buildcfg.LIBEXECDIR)
+		if err != nil {
+			fmt.Println(err)
```
bauerm97: Similar to what I changed in `PluginCompileCmd`, I think we should just use `Run` here, and do `sylog.Fatalf` instead of printing the error.

examples/plugins/test-plugin/main.go
```diff
@@ -18,7 +18,7 @@ import (
 // by the plugin framework to initialize the plugin
 var Plugin = pluginapi.Plugin{
 	Manifest: pluginapi.Manifest{
-		Name:        "TestPlugin",
+		Name:        "sylabs.io/test-plugin",
```
bauerm97, line 21: +1

internal/app/singularity/plugin_install_linux.go
```diff
@@ -5,23 +5,31 @@
...
+	}
+
+	defer fimg.UnloadContainer()
+
+	m, err := plugin.InstallFromSIF(&fimg, libexecdir)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("%#v\n", m)
```
bauerm97: Stray debug output left behind here?

internal/app/singularity/plugin_install_linux.go
```diff
@@ -5,23 +5,31 @@
...
-	// if !isPluginFile(&fimg) {
-	// 	return fmt.Errorf("sif file is not a plugin")
-	// }
-
-	// if err := copyFile(pluginPath, filepath.Join(libexecdir, ".")); err != nil {
-	// 	return fmt.Errorf("while copying plugin file to install location: %s", err)
-	// }
+	fimg, err := sif.LoadContainer(pluginPath, true)
+	if err != nil {
+		return err
```
bauerm97, line 20: I think in general we should add clarifying information to `return err` statements to make debugging easier

`return fmt.Errorf("while loading sif: %s", err)`

bauerm97 (changes requested):
So for the most part this is really awesome. I think you perfectly captured my intentions when I was writing the first draft of this, hopefully I left detailed enough comments throughout that it wasn't too nasty to take care of ;)

Just a few minor tweaks and tune ups before I think it's ready to merge. We can either wait and add testing to this PR; or (my preference) we can merge the PR as is and create a new issue detailing what testing is required before a `3.1.0` release. I think the second approach is best because there is actually a whole bunch of testing we need to add which spans more than just installation, but I leave it to you to consider

bauerm97:
p.s. I updated the PR message to reference the correct issue. #2317 is an epic which encompasses multiple commands such as `configure`, `compile`, etc... We still will have a few more to go after this PR

internal/app/singularity/plugin_install_linux.go
mem: Sure, I still need to figure out how to test some of the error paths.

bauerm97:
Also, I just noticed that we're missing `uninstall` command here

cmd/internal/cli/plugin_install_linux.go
mem: Yep, I need to go over error reporting for all these commands

cmd/internal/cli/plugin_uninstall_linux.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/sylabs/singularity/docs"
+	"github.com/sylabs/singularity/internal/app/singularity"
+	"github.com/sylabs/singularity/internal/pkg/buildcfg"
+)
+
+// PluginUninstall takes the name of a plugin and uninstalls it from the
+// plugin directory
+//
+// singularity plugin uninstall <name>
+var PluginUninstallCmd = &cobra.Command{
```
bauerm97, line 20: Comment incorrectly formed as pointed out by linter

internal/pkg/plugin/binary.go
```diff
@@ -72,65 +76,229 @@ func (m *Meta) Config() (*os.File, error) {
...
-	m.installTo(libexecdir)
-	return m, nil
+	err = m.install(plugindir)
+	return m, err
 }
 
-// installTo installs the plugin represented by m into libexecdir. This should
-// normally only be called in NewFromImage
-func (m *Meta) installTo(libexecdir string) {
+func Uninstall(name, libexecdir string) error {
```
bauerm97, line 117: Needs a comment as pointed out by linter. Pesky thing

internal/pkg/plugin/binary.go
```diff
@@ -162,10 +330,101 @@ func (m *Meta) configName() string {
...
+	if err := json.Unmarshal(data, &manifest); err != nil {
+		fmt.Println(err)
+	}
+
+	return manifest
+}
+
+// List returns all the singularity plugins installed in libexecdir in
+// the form of a list of Meta information
+func List(libexecdir string) ([]*Meta, error) {
```
bauerm97, line 388: I commented this somewhere else, but maybe we can return `map[string]*Meta`, unless we see a reason not to. Key can be the SHA name of the meta file, since it's guaranteed to be unique per name, we shouldn't have any hash map collisions. What're your thoughts here @mem ? I don't see any immediate drawbacks to using a map, can you think of any?

internal/pkg/plugin/binary.go
```diff
@@ -72,65 +76,229 @@ func (m *Meta) Config() (*os.File, error) {
...
+	return m, err
 }
 
-// installTo installs the plugin represented by m into libexecdir. This should
-// normally only be called in NewFromImage
-func (m *Meta) installTo(libexecdir string) {
+func Uninstall(name, libexecdir string) error {
+	sylog.Debugf("Uninstalling plugin %q from %q", name, libexecdir)
+
+	metas, err := List(libexecdir)
```
bauerm97, line 120: Is it possible to change `List()` to return a `map[string]*Meta` to avoid having to iterate over a slice every time? I know it will likely never be an issue because of how quick Go is and the relatively few amount of plugins we expect to be installed

bauerm97 (changes requested):

internal/pkg/plugin/binary.go
mem: oops, sorry, missed this one, done.

cmd/internal/cli/plugin_uninstall_linux.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+	"github.com/sylabs/singularity/docs"
+	"github.com/sylabs/singularity/internal/app/singularity"
+	"github.com/sylabs/singularity/internal/pkg/buildcfg"
+)
+
+// pluginUninstall takes the name of a plugin and uninstalls it from the
```
bauerm97: `PluginUninstall`

cmd/internal/cli/plugin_install_linux.go
```diff
@@ -27,10 +27,11 @@ func init() {
...
 // singularity plugin install <path> [-n name]
 var PluginInstallCmd = &cobra.Command{
-	RunE: func(cmd *cobra.Command, args []string) error {
-		fmt.Println("Installing plugin")
-
-		return singularity.InstallPlugin(args[0], buildcfg.LIBEXECDIR)
+	Run: func(cmd *cobra.Command, args []string) {
+		err := singularity.InstallPlugin(args[0], buildcfg.LIBEXECDIR)
+		if err != nil {
+			fmt.Printf("Failed to install plugin %q: %s.", args[0], err)
```
bauerm97: Let's use `sylog.Fatalf()` here instead of `fmt.Printf()`. This way we get automatic exit code handling

bauerm97: Same for the rest of the `Run` functions

bauerm97 (approved):
Assuming CI passes, this looks good to me.

PR #2838, opened 2019-03-08
Fix URI usage with user namespace

### LINKED ISSUE #2588: Running an image from docker in 3.0.x compared to 2.6.1: non-suid
## Version of Singularity

3.0.x
### Expected behavior

With singularity <3.0.x it was possible to exec an image from docker using without setuid using userNS both with overlay and with underlay it was also possible to use directories non existing in the image. For example using a standard alpine image and binding $PWD:/minnie where $PWD ≠ $HOME I can run a Hello World! script from the host node using overlay allegedly using a NON-SETUID workflow

-bash-4.2$ singularity --version
**2.6.1-dist**
-bash-4.2$ cd /pippo/aforti
-bash-4.2$ cat /proc/sys/user/max_user_namespaces 
15000
-bash-4.2$  egrep 'setuid|lay ' /etc/singularity/singularity.conf|grep -v '#'
**allow setuid = no
enable overlay = yes**
enable underlay = yes
-bash-4.2$ lsb_release -r
Release:	7.6.1810
-bash-4.2$ uname -r
3.10.0-957.1.3.el7.x86_64
-bash-4.2$  singularity --verbose exec -C -B $PWD:/minnie docker://alpine cat /minnie/pluto
Increasing verbosity level (2)
Singularity version: 2.6.1-dist
Exec'ing: /usr/libexec/singularity/cli/exec.exec
Evaluating args: '-C -B /pippo/aforti:/minnie docker://alpine cat /minnie/pluto'
VERBOSE: Set messagelevel to: 2
VERBOSE: Initialize configuration file: /etc/singularity/singularity.conf
VERBOSE Docker image: alpine
VERBOSE Namespace: library
VERBOSE Repo Name: alpine
VERBOSE Repo Tag: latest
VERBOSE Version: None
VERBOSE Obtaining tags: https://index.docker.io/v2/library/alpine/tags/list
Docker image path: index.docker.io/library/alpine:latest
VERBOSE Obtaining manifest: https://index.docker.io/v2/library/alpine/manifests/latest
VERBOSE Obtaining manifest: https://index.docker.io/v2/library/alpine/manifests/latest
VERBOSE Obtaining manifest: https://index.docker.io/v2/library/alpine/manifests/sha256:3d2e482b82608d153a374df3357c0291589a61cc194ec4a9ca2381073a17f58e
Cache folder set to /home/aforti/.singularity/docker
Creating container runtime...
Exploding layer: sha256:cd784148e3483c2c86c50a48e535302ab0288bebd587accf40b714fffd0646b3.tar.gz
VERBOSE: Set messagelevel to: 2
VERBOSE: Initializing Singularity Registry
Exploding layer: sha256:1f6d219d0077f6af480398f264117d2e74c03ec5f12d09b7b29b1eefbba872eb.tar.gz
VERBOSE: Set messagelevel to: 2
VERBOSE: Initializing Singularity Registry
VERBOSE: Set messagelevel to: 2
VERBOSE: Initialize configuration file: /etc/singularity/singularity.conf
VERBOSE: Initializing Singularity Registry
VERBOSE: Invoking non-SUID program flow: /usr/libexec/singularity/bin/action
VERBOSE: Set messagelevel to: 2
VERBOSE: Initialize configuration file: /etc/singularity/singularity.conf
VERBOSE: Initializing Singularity Registry
**VERBOSE: Running NON-SUID program workflow**
VERBOSE: Invoking the user namespace
VERBOSE: No autofs bug path in configuration, skipping
VERBOSE: Exec'ing cleanupd thread: /usr/libexec/singularity/bin/cleanupd
VERBOSE: Set messagelevel to: 2
VERBOSE: Daemonizing cleandir cleanup process
VERBOSE: Using session directory: /var/singularity/mnt/session
VERBOSE: Mounting overlay with options: lowerdir=/var/singularity/mnt/container,upperdir=/var/singularity/mnt/overlay/upper,workdir=/var/singularity/mnt/overlay/work
VERBOSE: Singularity overlay mount did not work (Operation not permitted), continuing without it
VERBOSE: Running all mount components
VERBOSE: Couldn't mount devpts filesystem, continuing with PTY functionality disabled.
VERBOSE: Queuing mount of new procfs
VERBOSE: Queuing bind mount of /sys
VERBOSE: Using sessiondir for home directory
VERBOSE: Queuing bind mount of staged home directory to container: /var/singularity/mnt/session//home/aforti -> /home/aforti
VERBOSE: Queuing bind mount of '/pippo/aforti' to '/minnie'
VERBOSE: Queuing bind mount of /tmp
VERBOSE: Queuing bind mount of /var/tmp
VERBOSE: Not mounting current directory: contain was requested
VERBOSE: Mounting '/var/singularity/mnt/session/dev' at '/var/singularity/mnt/final/dev'
VERBOSE: Mounting 'proc' at '/var/singularity/mnt/final/proc'
VERBOSE: Mounting '/sys' at '/var/singularity/mnt/final/sys'
VERBOSE: Mounting '/var/singularity/mnt/session/home/aforti' at '/var/singularity/mnt/final/home/aforti'
VERBOSE: Mounting '/pippo/aforti' at '/var/singularity/mnt/final/minnie'
VERBOSE: Mounting '/var/singularity/mnt/session/tmp' at '/var/singularity/mnt/final/tmp'
VERBOSE: Mounting '/var/singularity/mnt/session/var_tmp' at '/var/singularity/mnt/final/var/tmp'
VERBOSE: Running file components
VERBOSE: Creating template passwd file and appending user data: /var/singularity/mnt/session/passwd
VERBOSE: Binding file '/var/singularity/mnt/session/passwd' to '/var/singularity/mnt/final/etc/passwd'
VERBOSE: Updating group file with user info
VERBOSE: Binding file '/var/singularity/mnt/session/group' to '/var/singularity/mnt/final/etc/group'
VERBOSE: Binding file '/etc/resolv.conf' to '/var/singularity/mnt/final/etc/resolv.conf'
VERBOSE: Containing all rootfs components
VERBOSE: Entering container file system root: /var/singularity/mnt/final
LOG    : USER=aforti, IMAGE='alpine', COMMAND='exec'
VERBOSE: Exec'ing /.singularity.d/actions/exec
**Hello World!**
VERBOSE: Cleaning directory: /tmp/.singularity-runtime.19KgOs0H

### Actual behavior
If I use singularity 3.0.3 instead. With the same setup it fails

-bash-4.2$ singularity --version
**singularity version 3.0.3-rc2.el7**
-bash-4.2$ cd /pippo/aforti
-bash-4.2$ cat /proc/sys/user/max_user_namespaces 
15000
-bash-4.2$ egrep 'setuid|lay ' /etc/singularity/singularity.conf|grep -v '#'
allow setuid = no
enable overlay = yes
enable underlay = yes
-bash-4.2$ lsb_release -r
Release:	7.6.1810
-bash-4.2$ uname -r
3.10.0-957.1.3.el7.x86_64
-bash-4.2$ singularity --verbose exec -C -B $PWD:/minnie docker://alpine cat /minnie/pluto
VERBOSE: Set messagelevel to: 4
VERBOSE: Container runtime
VERBOSE: Check if we are running as setuid
VERBOSE: Spawn scontainer stage 1
VERBOSE: Execute scontainer stage 1
**VERBOSE: Create user namespace**
VERBOSE: Create mount namespace
VERBOSE: Create pid namespace
VERBOSE: Spawn smaster process
VERBOSE: Spawn scontainer stage 2
VERBOSE: Create ipc namespace
VERBOSE: Create mount namespace
VERBOSE: Spawn RPC server
VERBOSE: Execute smaster process
VERBOSE: Serve RPC requests
VERBOSE: Not mounting current directory: container was requested
**FATAL:   container creation failed: mount error: can't mount image /proc/self/fd/9: failed to find loop device: could not attach image file too loop device: permission denied**

Note that this is not only a overlay issue. Underlay doesn't work either

steps above.

Is this expected? or is it a bug in 3.0.x. Running user images from docker is an important workflow for us (and others). If we cannot run like we do in 2.6.1 we cannot use non-suid mode.

@lukasheinrich

cclerget:

This PR fixes a regression from v2 allowing user namespace to execute container directly from URI like :

```
singularity shell -u docker://busybox
```

- Fixes #2588

### DIFF
--- a/cmd/internal/cli/actions_linux.go
+++ b/cmd/internal/cli/actions_linux.go
@@ -45,9 +45,12 @@ func convertImage(filename string, unsquashfsPath string) (string, error) {
 	}
 	defer img.File.Close()
 
+	// squashfs only
 	if img.Partitions[0].Type != image.SQUASHFS {
 		return "", fmt.Errorf("not a squashfs root filesystem")
 	}
+
+	// create a reader for rootfs partition
 	reader, err := image.NewPartitionReader(img, "", 0)
 	if err != nil {
 		return "", fmt.Errorf("could not extract root filesystem: %s", err)
@@ -56,14 +59,34 @@ func convertImage(filename string, unsquashfsPath string) (string, error) {
 	if !s.HasUnsquashfs() && unsquashfsPath != "" {
 		s.UnsquashfsPath = unsquashfsPath
 	}
-	dir, err := ioutil.TempDir("", "rootfs-")
+
+	// keep compatibility with v2
+	tmpdir := os.Getenv("SINGULARITY_LOCALCACHEDIR")
+	if tmpdir == "" {
+		pw, err := user.GetPwUID(uint32(os.Getuid()))
+		if err != nil {
+			return "", fmt.Errorf("could not find current user information: %s", err)
+		}
+		tmpdir = filepath.Join(pw.Dir, ".singularity", "tmp")
+		if !fs.IsDir(tmpdir) {
+			if err := os.Mkdir(tmpdir, 0755); err != nil {
+				return "", fmt.Errorf("could not create directory %s: %s", tmpdir, err)
+			}
+		}
+	}
+
+	// create temporary sandbox
+	dir, err := ioutil.TempDir(tmpdir, "rootfs-")
 	if err != nil {
 		return "", fmt.Errorf("could not create temporary sandbox: %s", err)
 	}
+
+	// extract root filesystem
 	if err := s.ExtractAll(reader, dir); err != nil {
 		os.RemoveAll(dir)
 		return "", fmt.Errorf("root filesystem extraction failed: %s", err)
 	}
+
 	return dir, err
 }
 
@@ -402,6 +425,7 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 		}
 		engineConfig.SetImage(dir)
 		engineConfig.SetDeleteImage(true)
+		generator.AddProcessEnv("SINGULARITY_CONTAINER", dir)
 	}
 
 	plugin.FlagHookCallbacks(engineConfig)
--- a/cmd/internal/cli/actions_linux.go
+++ b/cmd/internal/cli/actions_linux.go
@@ -9,6 +9,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -16,7 +17,10 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/sylabs/singularity/internal/pkg/image/unpacker"
+
 	"github.com/opencontainers/runtime-tools/generate"
+	"github.com/sylabs/singularity/internal/pkg/image"
 	"github.com/sylabs/singularity/internal/pkg/plugin"
 	"github.com/sylabs/singularity/pkg/util/nvidia"
 
@@ -30,8 +34,38 @@ import (
 	"github.com/sylabs/singularity/internal/pkg/sylog"
 	"github.com/sylabs/singularity/internal/pkg/util/env"
 	"github.com/sylabs/singularity/internal/pkg/util/exec"
+	"github.com/sylabs/singularity/internal/pkg/util/fs"
 	"github.com/sylabs/singularity/internal/pkg/util/user"
 )
+
+func convertImage(filename string, unsquashfsPath string) (string, error) {
+	img, err := image.Init(filename, false)
+	if err != nil {
+		return "", fmt.Errorf("could not open image %s: %s", filename, err)
+	}
+	defer img.File.Close()
+
+	if img.Partitions[0].Type != image.SQUASHFS {
+		return "", fmt.Errorf("not a squashfs root filesystem")
+	}
+	reader, err := image.NewPartitionReader(img, "", 0)
+	if err != nil {
+		return "", fmt.Errorf("could not extract root filesystem: %s", err)
+	}
+	s := unpacker.NewSquashfs()
+	if !s.HasUnsquashfs() && unsquashfsPath != "" {
+		s.UnsquashfsPath = unsquashfsPath
+	}
+	dir, err := ioutil.TempDir("", "rootfs-")
+	if err != nil {
+		return "", fmt.Errorf("could not create temporary sandbox: %s", err)
+	}
+	if err := s.ExtractAll(reader, dir); err != nil {
+		os.RemoveAll(dir)
+		return "", fmt.Errorf("root filesystem extraction failed: %s", err)
+	}
+	return dir, err
+}
 
 // TODO: Let's stick this in another file so that that CLI is just CLI
 func execStarter(cobraCmd *cobra.Command, image string, args []string, name string) {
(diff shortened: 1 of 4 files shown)


internal/pkg/runtime/engines/singularity/cleanup.go
```diff
@@ -25,6 +25,13 @@ import (
 func (engine *EngineOperations) CleanupContainer(fatal error, status syscall.WaitStatus) error {
 	sylog.Debugf("Cleanup container")
 
+	if engine.EngineConfig.GetDeleteImage() {
+		image := engine.EngineConfig.GetImage()
+		if err := os.RemoveAll(image); err != nil {
+			sylog.Errorf("failed to delete container image %s: %s", image, err)
```
bauerm97, line 31: Are we intentionally calling `Errorf` and not exiting here? This makes it possible to fail the image deletion and still return a 0 exit code

internal/pkg/runtime/engines/singularity/config/config.go
```diff
@@ -90,6 +90,7 @@ type JSONConfig struct {
 	TargetUID     int           `json:"targetUID,omitempty"`
 	TargetGID     []int         `json:"targetGID,omitempty"`
 	LibrariesPath []string      `json:"librariesPath,omitempty"`
+	DeleteImage   bool          `json:"deleteImage,omitempty"`
```
bauerm97, line 93: With this option, doesn't this render it impossible for two executions of singularity to share the same temporary splatted fs?

cmd/internal/cli/actions_linux.go
```diff
@@ -30,9 +34,62 @@ import (
...
+		tmpdir = filepath.Join(pw.Dir, ".singularity", "tmp")
+		if !fs.IsDir(tmpdir) {
+			if err := os.Mkdir(tmpdir, 0755); err != nil {
+				return "", fmt.Errorf("could not create directory %s: %s", tmpdir, err)
+			}
+		}
+	}
+
+	// create temporary sandbox
+	dir, err := ioutil.TempDir(tmpdir, "rootfs-")
```
bauerm97, line 82: I'm questioning this approach. It should be possible to have a shared cache of splatted out images (I believe this was the behavior in 2.x, correct me if I'm wrong anyone). Even if this isn't the case with 2.x, I still think it should be possible:

1. Aquire shared lock on a file representing the hash of the fs we want to splat
2. If we're first to aquire this lock (i.e. create this file) then splat the image out to disk
3. After image extraction is complete, write a second file (or write some data to hash file) to signal that splatted image can be used
4. Aquire exclusive lock on file (block other files from opening/shared locking)
5. Check if any other shared locks on this file exist
6. Delete splatted filesystem if no other shared locks exist
7. Release both exclusive and shared lock

The approach outlined above does seem much more complex, so perhaps we should merge this Pull Request as is and discuss the possibility of a shared cache and revisit if applicable

bauerm97 (changes requested):
Just some general comments from a quick read through while on the airplane. Once I land in San Diego I'll try to revisit and see if there is anything which needs changing.

One thing I would like to see is the inclusion of tests here so we don't get another regression down the line

internal/pkg/runtime/engines/singularity/cleanup.go
cclerget: Yes it's best effort, we can't return error here and interrupt all other cleanup steps like cgroups, network and let garbage on node

internal/pkg/runtime/engines/singularity/config/config.go
cclerget: each execution has its own copy

cmd/internal/cli/actions_linux.go
cclerget: The PR approach is strictly identical to 2.x, create a temporary folder and delete it after container execution, those images were not cached in 2.x.

Now if we go this way, imo it should be handled by cache package instead and additionally deal with multi-node cases with the concern that some network filesystem doesn't support file locking

bauerm97 (approved):

PR #2841, opened 2019-03-08
fixing logger, adding default to no color given not the terminal

vsoch:
Signed-off-by: Vanessa Sochat <vsochat@stanford.edu>

This pull request will detect if the user is not directing output to the terminal, and if not, disable color automatically. It also fixes a bug that @bauerm97 and I overlooked of missing a few of the message levels! Specifically this part:

```go
// DisableColor for the logger
func DisableColor() {
	messageColors = map[messageLevel]string{
		fatal:    "",
		error:    "",
		warn:     "",
		info:     "",
	}
	colorReset = ""
}
```

was missing levels!

```go
// DisableColor for the logger
func DisableColor() {
	messageColors = map[messageLevel]string{
		fatal:    "",
		error:    "",
		warn:     "",
		info:     "",
		verbose:  "",
		verbose2: "",
		verbose3: "",
		debug:    "",
	}
	colorReset = ""
}
```

Thus the debug still had weird characters with the flag.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ _The old changelog can be found in the `release-2.6` branch_
 
   - Fixed usage docstrings for OCI - was missing "oci" in command examples.
   - Added --nocolor flag to Singularity client to disable color in logging
+  - writing to (non terminal) for stderr automatically disables color
 
 # v3.1.0 - [2019.02.22]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,6 @@ _The old changelog can be found in the `release-2.6` branch_
 
   - Fixed usage docstrings for OCI - was missing "oci" in command examples.
   - Added --nocolor flag to Singularity client to disable color in logging
-  - writing to (non terminal) for stderr automatically disables color
 
 # v3.1.0 - [2019.02.22]
 
--- a/cmd/internal/cli/singularity.go
+++ b/cmd/internal/cli/singularity.go
@@ -22,6 +22,7 @@ import (
 	"github.com/sylabs/singularity/internal/pkg/plugin"
 	"github.com/sylabs/singularity/internal/pkg/sylog"
 	"github.com/sylabs/singularity/internal/pkg/util/auth"
+	"golang.org/x/crypto/ssh/terminal"
 )
 
 // Global variables for singularity CLI
@@ -104,7 +105,9 @@ func setSylogMessageLevel(cmd *cobra.Command, args []string) {
 }
 
 func setSylogColor(cmd *cobra.Command, args []string) {
-	if nocolor {
+	// Disable color if nocolor flag, or NOT writing to terminal
+	fd := int(os.Stderr.Fd())
+	if nocolor || !terminal.IsTerminal(fd) {
 		sylog.DisableColor()
 	}
 }
--- a/internal/pkg/sylog/sylog_test.go
+++ b/internal/pkg/sylog/sylog_test.go
@@ -12,8 +12,8 @@ import (
 	"testing"
 )
 
-// TestSylogSetLevel to ensure the correct integer is returned
-func TestSylogSetLevel(t *testing.T) {
+// TestSetLevel to ensure the correct integer is returned
+func TestSetLevel(t *testing.T) {
 	var levelTests = []struct {
 		name  string
 		level int
@@ -32,9 +32,7 @@ func TestSylogSetLevel(t *testing.T) {
 	for _, tt := range levelTests {
 		t.Run(tt.name, func(t *testing.T) {
 			SetLevel(tt.level)
-
 			l := GetLevel()
-
 			if l != tt.level {
 				t.Errorf("got %d, want %d", l, tt.level)
 			}
@@ -42,8 +40,7 @@ func TestSylogSetLevel(t *testing.T) {
 	}
 }
 
-func TestSylogPrefix(t *testing.T) {
-
+func TestPrefix(t *testing.T) {
 	var logSuffix = "\x1b[0m "
 	var levelTests = []struct {
 		name  string
@@ -80,8 +77,7 @@ func TestSylogPrefix(t *testing.T) {
 	}
 }
 
-func TestSylogDisableColor(t *testing.T) {
-
+func TestDisableColor(t *testing.T) {
 	var logSuffix = "\x1b[0m "
 	var levelTests = []struct {
 		name  string
@@ -101,19 +97,19 @@ func TestSylogDisableColor(t *testing.T) {
 	// Disable all color output, removing prefix and off suffix
 	DisableColor()
 
-	// Test that prefix is colored
+	// Test that prefix is not colored
 	for _, tt := range levelTests {
 		t.Run(tt.name, func(t *testing.T) {
 
 			SetLevel(int(tt.level))
 			levelPrefix := prefix(tt.level)
 
-			// Check that we start with the color prefix
+			// Check that we don't start with the color prefix
 			if !strings.HasPrefix(levelPrefix, tt.name) {
 				t.Errorf("got prefix %s, want %s", levelPrefix, tt.name)
 			}
 
-			// The suffix should be consistently the "off" color string
+			// The suffix should not contain the the "off" color string
 			if strings.HasSuffix(levelPrefix, logSuffix) && tt.level != debug {
 				t.Errorf("%s does ends with %s", levelPrefix, logSuffix)
 			}
--- a/internal/pkg/sylog/sylog_test.go
+++ b/internal/pkg/sylog/sylog_test.go
@@ -1,0 +1,122 @@
+// Copyright (c) 2019, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+// +build sylog
+
+package sylog
+
+import (
+	"strings"
+	"testing"
+)
+
+// TestSylogSetLevel to ensure the correct integer is returned
+func TestSylogSetLevel(t *testing.T) {
+	var levelTests = []struct {
+		name  string
+		level int
+	}{
+		{"SetLevelFatal", int(fatal)},
+		{"SetLevelError", int(error)},
+		{"SetLevelWarn", int(warn)},
+		{"SetLevelLog", int(log)},
+		{"SetLevelInfo", int(info)},
+		{"SetLevelVerbose", int(verbose)},
+		{"SetLevelVerbose2", int(verbose2)},
+		{"SetLevelVerbose3", int(verbose3)},
+		{"SetLevelDebug", int(debug)},
+	}
+
+	for _, tt := range levelTests {
+		t.Run(tt.name, func(t *testing.T) {
+			SetLevel(tt.level)
+
+			l := GetLevel()
+
+			if l != tt.level {
+				t.Errorf("got %d, want %d", l, tt.level)
+			}
+		})
+	}
+}
+
+func TestSylogPrefix(t *testing.T) {
+
+	var logSuffix = "\x1b[0m "
+	var levelTests = []struct {
+		name  string
+		level messageLevel
+	}{
+		{"\x1b[31mFATAL", fatal},
+		{"\x1b[31mERROR", error},
+		{"\x1b[33mWARNING", warn},
+		{"\x1b[0mLOG", log},
+		{"\x1b[34mINFO", info},
+		{"\x1b[0mVERBOSE", verbose},
+		{"\x1b[0mVERBOSE", verbose2},
+		{"\x1b[0mVERBOSE", verbose3},
+		{"\x1b[0mDEBUG", debug},
+	}
+
+	// Test that prefix is colored
+	for _, tt := range levelTests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			SetLevel(int(tt.level))
+			levelPrefix := prefix(tt.level)
+
+			// Check that we start with the color prefix
+			if !strings.HasPrefix(levelPrefix, tt.name) {
+				t.Errorf("got prefix %s, want %s", levelPrefix, tt.name)
+			}
+
+			// The suffix should be consistently the "off" color string
+			if !strings.HasSuffix(levelPrefix, logSuffix) && tt.level != debug {
+				t.Errorf("%s does not end with %s", levelPrefix, logSuffix)
+			}
+		})
+	}
+}
+
+func TestSylogDisableColor(t *testing.T) {
+
+	var logSuffix = "\x1b[0m "
+	var levelTests = []struct {
+		name  string
+		level messageLevel
+	}{
+		{"FATAL", fatal},
+		{"ERROR", error},
+		{"WARNING", warn},
+		{"LOG", log},
+		{"INFO", info},
+		{"VERBOSE", verbose},
+		{"VERBOSE", verbose2},
+		{"VERBOSE", verbose3},
+		{"DEBUG", debug},
+	}
+
+	// Disable all color output, removing prefix and off suffix
+	DisableColor()
+
+	// Test that prefix is colored
+	for _, tt := range levelTests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			SetLevel(int(tt.level))
+			levelPrefix := prefix(tt.level)
+
+			// Check that we start with the color prefix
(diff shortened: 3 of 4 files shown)


bauerm97:
Hey @vsoch, maybe since we already identified this as potentially buggy (we both missed a bug already at first pass) there should be some unit tests to ensure that we don't have any more regressions. A simple `sylog_test.go` file could be added which does a few tests. Would you be willing to take a stab? 

A simple baseline test that does the following would suffice:
1. Call `prefix()` against values of `level` (input to the function) and `loggerLevel` (level of logger verbosity, use `SetLevel()` to change this). This is effectively an N^2 matrix of variables
2. Check returned prefixes contain the appropriate color strings
3. Call `DisableColor()`
4. Run `prefix()` against the matrix again
5. Check that all returned prefixes do *not* contain color strings this time

vsoch:
Good idea! I'm thinking also about the functionality - if we disable color automatically given output isn't to a terminal, do we need the --nocolor option? I think so, because there are terminals that don't support color where it would come out looking funky. On the flip side, what if someone is piping to a file and does want color? I can imagine someone wanting to pipe to a file that is printed later. In this case, given that we have a --nocolor argument, I think there should be a --color flag. Or we can consolidate those two into one thing and have something like `--color` or `--color=off` (does GoLang support having a flag that can serve as a boolean and a string argument?) We could also completely remove --nocolor, assuming that the default of disabling when we pipe to a file will fit most use cases. Thoughts?

vsoch:
@bauerm97 how do I call private functions? I tried adding -v and capitalizing prefix, for example, neither of those work.

vsoch:
I'm having quite a bit of trouble writing tests for private functions (it seems like it's not common to do) but I updated the PR with the bug fixes.

bauerm97:
@vsoch actually I think it's fairly common. I'm pretty sure you simply declare your test file as part of the same package. In `sylog_test.go`:

```
package sylog

...
```

Also you do need to ensure that you're using the `sylog` build tag, otherwise it will compile with the `sylog_dummy.go` file which will not have a `prefix()` function. I would hazard a guess that this is what's causing issues

vsoch:
I had the package sylog at the top, what is this build tag that you speak of? I saw this header in the sylog.go file:

```
// +build sylog
```
And I added it to my test file, but I got the same error.. Here is everything (a bit messy atm) that I have if you want to take a look. I'm running with `go test -v .`

```go

// +build sylog

package sylog

import (
        "fmt"
	"testing"

	"github.com/magiconair/properties/assert"
)

var nocolor bool
type messageLevel int

const (
	fatal    messageLevel = iota - 4 // fatal    : -4
	error                            // error    : -3
	warn                             // warn     : -2
	log                              // log      : -1
	_                                // SKIP     : 0
	info                             // info     : 1
	verbose                          // verbose  : 2
	verbose2                         // verbose2 : 3
	verbose3                         // verbose3 : 4
	debug                            // debug    : 5
)

var messageColors = map[messageLevel]string{
	fatal:    "\x1b[31m",
	error:    "\x1b[31m",
	warn:     "\x1b[33m",
	info:     "\x1b[34m",
	log:      "\x1b[0m",
	verbose:  "\x1b[0m",
	verbose2: "\x1b[0m",
	verbose3: "\x1b[0m",
	debug:    "\x1b[0m",
}

func TestSetLevel(t *testing.T) {
        var loggerLevel int

        // Loop through logging levels and check that they are set
        for level := -4; level <= 5; level++ {
                SetLevel(level)
                loggerLevel = GetLevel()
                assert.Equal(t, loggerLevel, level, "The level should be set.")
        }
}


//func TestSylogDisableColor(m *testing.T) {
//		sylog.DisableColor()
//}
func TestSylogPrefix(m *testing.T) {
	var messagePrefix string
        type level int

        // Call prefix() against values of level
        for level := -4; level <= 5; level++ {
                SetLevel(level)
                messagePrefix = prefix(level)
                fmt.Println(messagePrefix)                
                // Check returned prefixes contain the appropriate color strings
        	//assert.T(prefix, level, "true", "The level should be set.")
        }



//Call DisableColor()
//Run prefix() against the matrix again
//Check that all returned prefixes do not contain color strings this time
}
```
The functions don't seem to be defined:

```
$ go test -v sylog_test.go 
# command-line-arguments [command-line-arguments.test]
./sylog_test.go:50:17: undefined: SetLevel
./sylog_test.go:51:31: undefined: GetLevel
./sylog_test.go:66:17: undefined: SetLevel
./sylog_test.go:67:33: undefined: prefix
FAIL	command-line-arguments [build failed]
```
Then if I just use a . (I've had better luck with this in the past):

```
$ go test -v .
?   	github.com/sylabs/singularity/internal/pkg/sylog	[no test files]
```

bauerm97:
Hey @vsoch , sorry for the confusion here (wrote that post on my phone on a plane). What you need to do is run the `go test` command with the appropriate `sylog` tag set. To do this, run the command:

`go test -tags=sylog .`

By adding `-tags=sylog`, you tell Go to compile the test code with the `sylog` tag. This causes the proper `sylog.go` file to get included in the build.

vsoch:
thanks @bauerm97! I'll try that.

vsoch:
Hey @bauerm97 can you give me some kind of template to start with? I'm not making good use of time because I don't have a sense of how to do this. I think it will be better for me to learn from your example.

vsoch:
This looks like a Travis error:

```bash
ERROR:   unpackSIF Failed: While copying partition data to bundle: cp Failed: exit status 1: cp: reading `/tmp/tmpmnt-610282689/./.singularity.d/Singularity': Input/output error
                cp: reading `/tmp/tmpmnt-610282689/./.singularity.d/actions/exec': Input/output error
...
                FATAL:   While performing build: packer failed to pack: While copying partition data to bundle: cp Failed: exit status 1: cp: reading `/tmp/tmpmnt-610282689/./.singularity.d/Singularity': Input/output error
```

Have you guys seen this before? Rerun?

cmd/internal/cli/singularity.go
```diff
@@ -104,7 +105,9 @@ func setSylogMessageLevel(cmd *cobra.Command, args []string) {
 }
 
 func setSylogColor(cmd *cobra.Command, args []string) {
-	if nocolor {
+	// Disable color if nocolor flag, or NOT writing to terminal
+	fd := int(os.Stderr.Fd())
```
bauerm97, line 109: I don't think we should automatically be trying to do this here. There are too many variables at play for that to make sense in my opinion. I think just providing the flag is good enough

internal/pkg/sylog/sylog_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+package sylog
+
+import (
+	"strings"
+	"testing"
+)
+
+// TestSylogSetLevel to ensure the correct integer is returned
+func TestSylogSetLevel(t *testing.T) {
```
bauerm97: None of these function names should be including `Sylog`. All of these functions are already within `package sylog`, so it's redundant

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ _The old changelog can be found in the `release-2.6` branch_
 
   - Fixed usage docstrings for OCI - was missing "oci" in command examples.
   - Added --nocolor flag to Singularity client to disable color in logging
+  - writing to (non terminal) for stderr automatically disables color
```
bauerm97: I think this should be removed from here

internal/pkg/sylog/sylog_test.go
```diff
@@ -0,0 +1,122 @@
...
+		{"SetLevelVerbose2", int(verbose2)},
+		{"SetLevelVerbose3", int(verbose3)},
+		{"SetLevelDebug", int(debug)},
+	}
+
+	for _, tt := range levelTests {
+		t.Run(tt.name, func(t *testing.T) {
+			SetLevel(tt.level)
+
+			l := GetLevel()
```
bauerm97, line 35: This line is unnecessary, we can just do `if GetLevel() == tt.level`

internal/pkg/sylog/sylog_test.go
```diff
@@ -0,0 +1,122 @@
...
+			// The suffix should be consistently the "off" color string
+			if !strings.HasSuffix(levelPrefix, logSuffix) && tt.level != debug {
+				t.Errorf("%s does not end with %s", levelPrefix, logSuffix)
+			}
+		})
+	}
+}
+
+func TestSylogDisableColor(t *testing.T) {
+
```
bauerm97: Delete empty line

internal/pkg/sylog/sylog_test.go
```diff
@@ -0,0 +1,122 @@
...
+			if l != tt.level {
+				t.Errorf("got %d, want %d", l, tt.level)
+			}
+		})
+	}
+}
+
+func TestSylogPrefix(t *testing.T) {
+
```
bauerm97: Delete empty line

internal/pkg/sylog/sylog_test.go
```diff
@@ -0,0 +1,122 @@
...
+		{"VERBOSE", verbose},
+		{"VERBOSE", verbose2},
+		{"VERBOSE", verbose3},
+		{"DEBUG", debug},
+	}
+
+	// Disable all color output, removing prefix and off suffix
+	DisableColor()
+
+	// Test that prefix is colored
```
bauerm97: This comment is probably a copy paste error? Test that the prefix isn't colored, right?

internal/pkg/sylog/sylog_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+			SetLevel(int(tt.level))
+			levelPrefix := prefix(tt.level)
+
+			// Check that we start with the color prefix
+			if !strings.HasPrefix(levelPrefix, tt.name) {
+				t.Errorf("got prefix %s, want %s", levelPrefix, tt.name)
+			}
+
+			// The suffix should be consistently the "off" color string
```
bauerm97: Here, the suffix should _not_ contain a color reset string.

internal/pkg/sylog/sylog_test.go
```diff
@@ -0,0 +1,122 @@
...
+			if strings.HasSuffix(levelPrefix, logSuffix) && tt.level != debug {
```
bauerm97, line 113: I'm not sure why this isn't failing, if we have called `DisableColor`, then we should never be including any color codes in our return value from `prefix()`

bauerm97 (changes requested):
There are a few changes we still need to make to the testing package. Also an open question about this approach to testing: If we call `DisableColor` in one `TestXXX` function, are the subsequent tests running in that same context or is a new Go runtime spawned up. It might be possible that running these tests will result in different behavior depending on which test was ran first. 

cc @cclerget @tri-adam if you guys know the answer to that ^

cmd/internal/cli/singularity.go
vsoch: This was a suggestion / request of someone else (I can't remember now, @cclerget was it you?) that I added after the fact. It the maintainers want to decide on a consensus I'd be happy to update the functionality.

internal/pkg/sylog/sylog_test.go
vsoch: Affirmative.

internal/pkg/sylog/sylog_test.go
vsoch: That means it's correct - the error is only called if the return value (levelPrefix) has the suffix (which it doesn't because it's disabled).

internal/pkg/sylog/sylog_test.go
vsoch: The reason is because we want to save the variable and report the result back to the user - we use it twice.

cmd/internal/cli/singularity.go
mem: I actually think this is nice to make. If we want color for example when using a pager (e.g. "singularity ... | less") then we need an additional option `--color` or follow git's lead (and other tools' as well) and have `--color=on`, `--color=off`, `--color=auto`, with the last one being the default value. `--color=auto` in git and other tools means "color when the output is a terminal, no color otherwise".

internal/pkg/sylog/sylog_test.go
mem: That makes sense. Why do you need the extra tt.level != debug, though?

cmd/internal/cli/singularity.go
vsoch: Agree! This is similar to what I was bringing up a few days ago :point_up:  https://github.com/sylabs/singularity/pull/2841#issuecomment-471022566

internal/pkg/sylog/sylog_test.go
vsoch: Good question! When debug is the level, there is the added uid and process id to the output (something like `[uid=1000,pid=1234]`) and it's not colored, so it doesn't have the same suffix as the others.

cmd/internal/cli/singularity.go
bauerm97: Alright, so if this is a standard scheme for coloring output, then I'm happy to oblige here.

vsoch: @mem and @bauerm97 do you want to add the `--color` variable as @mem detailed above, or have two booleans `--nocolor` and `--color` in addition to the automatic detection of where is being written (which is overridden by specifying one of these flags).

internal/pkg/sylog/sylog_test.go
bauerm97: So, I think it's actually probably easier to just match a regex here instead of trying to handle every individual case. We can create a color regex, and a no-color regex:

To match for colored prefixes: `^(\\x1b\[\d+m)(\w{1,7}\s*)(\\x1b\[0m)`
To match for without prefixes: `^(FATAL|ERROR|WARNING|LOG|INFO|VERBOSE|(DEBUG\s*\[.*\]))\s*$`

I think we can just apply those regexps on the returned prefixes. The first should only match for prefixes returned in `TestPrefix`, and the second one should match only for prefixes returned within `TestDisableColor`. I think this will greatly simplify the possibility that we've gotten something wrong in the testing matrix

bauerm97: Actually that second one needs a small tweak, hold on

bauerm97: OK, second regex is correct now I believe

vsoch: A single regex would be a lot cleaner, but Isn't the idea of matching the messageLevels to strings to check for that? For example, we want to check that "info" (the integer / messageLevel) corresponds to the string "INFO." If we do one gigantic regular expression, we don't check for that.

cmd/internal/cli/singularity.go
vsoch: ping! What is the preference for the argument?

GodloveD:
@mem @bauerm97 @vsoch where does this stand?

cmd/internal/cli/singularity.go
vsoch: @GodloveD I was waiting for feedback here.

WestleyK: @vsoch, Imo, I like the variable flag better, just like `ls`, `git` (although git also has a `--no-color` flag).

vsoch: Thanks for the feedback! I had totally forgot about this PR. Let me know what changes would be requested, if any.

sashayakovtseva:
@vsoch Can you update your branch, please?

sashayakovtseva (approved):

vsoch:
I can do the first file, but I don't know how to do the second, it's just been too long. I don't have the branch locally anymore.

sashayakovtseva:
Can't you pull from your remote?

vsoch:
Yes I can, but it produces the same merge conflicts. As I mentioned, I can resolve the first file (singularity.go) but I don't have knowledge about the changes that were made subsequently to sylog_test.go so I wouldn't know how to address the conflicts. The resolution can be done in the interface by someone that has this knowledge. Thanks!

sashayakovtseva:
Ok, I will update this if the rest of the team approves PR.

vsoch:
Great thank you! I gave it a shot, but I cancelled the conflict resolution when I got to the second file. I don't want to mess it up. :P

WestleyK:
Ping @sashayakovtseva

sashayakovtseva:
There are lots of conflicts, honestly.
Looks like creating . a new one on top of the master is easier than to solve them, sorry @vsoch.

vsoch:
yeah, that unfortunately happens when a review takes 6 months for a very busy repository, no worries @sashayakovtseva.

PR #2846, opened 2019-03-10
fixing bind warning / fail

### LINKED ISSUE #1469: Do not exit/abort in case of overlapping directory binds
## Version of Singularity:
2.4.6

### Expected behavior

If two directory binds are overlapping (e.g. `singularity shell -B /data -B /data/test ...`) it would be nice to just get a warning instead of an error. 
(This was the behavior before commit 7ecb26f6f6a8dc3e2d21f0e0fadcbffdf3bcceb9)

Alternatively, it would be nice if the user could decide whether a warning or an error should be raised.

The reason for this "wish" is that some pipelining frame works / tools automatically generate Singularity commands based on some input, where it can happen that multiple directory binds are overlapping.

### Actual behavior

In the current version, overlapping directory binds lead to an error and abort:
`ERROR  : Not mounting requested bind point (already mounted in container)`

`singularity shell -B /data -B /data/test image.simg`

Thank you very much!

vsoch:

This pull request will address the issue that if a user defines the same bind point twice, the (current) implementation will fail. Instead, we can skip and give the user a warning of a repeated bind:

```bash
$ singularity shell --bind /tmp --bind /tmp docker://busybox
WARNING: destination /tmp is already in the mount point list
Singularity> 
```

- Fixes #1469 

### DIFF
--- a/internal/pkg/util/fs/mount/mount.go
+++ b/internal/pkg/util/fs/mount/mount.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"syscall"
 
+	"github.com/sylabs/singularity/internal/pkg/sylog"
 	"github.com/sylabs/singularity/pkg/util/fs/proc"
 
 	specs "github.com/opencontainers/runtime-spec/specs-go"
@@ -339,7 +340,9 @@ func (p *Points) add(tag AuthorizedTag, source string, dest string, fstype strin
 			}
 		}
 		if present {
-			return fmt.Errorf("destination %s is already in the mount point list", dest)
+			sylog.Warningf("destination %s is already in the mount point list", dest)
+			return nil
+
 		}
 		if len(p.points[tag]) == 1 && !authorizedTags[tag].multiPoint {
 			return fmt.Errorf("tag %s allow only one mount point", tag)
--- a/internal/pkg/util/fs/mount/mount.go
+++ b/internal/pkg/util/fs/mount/mount.go
@@ -349,8 +349,8 @@ func (p *Points) add(tag AuthorizedTag, source string, dest string, fstype strin
 		}
 		if present {
 			return ErrMountExists
-
 		}
+
 		if len(p.points[tag]) == 1 && !authorizedTags[tag].multiPoint {
 			return fmt.Errorf("tag %s allow only one mount point", tag)
 		}
--- a/internal/pkg/util/fs/mount/mount.go
+++ b/internal/pkg/util/fs/mount/mount.go
@@ -11,10 +11,18 @@ import (
 	"strings"
 	"syscall"
 
-	"github.com/sylabs/singularity/internal/pkg/sylog"
 	"github.com/sylabs/singularity/pkg/util/fs/proc"
 
 	specs "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+type mountError string
+
+func (e mountError) Error() string { return string(e) }
+
+const (
+	// ErrMountExists indicates a duplicated mount being asked for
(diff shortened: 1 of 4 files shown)


cclerget:
Hey @vsoch , thanks to tackle this one. A comment about not returning an error in mount package, it could potentially break other situations where it's intended to interrupt execution in case of duplicated destination. I would go to return a generic error in mount package like `ErrMountExist` (take a look at `pkg/network/network.go` about error constant) that could be easily compared in `addUserbindsMount` to ignore this error and display warning in `addUserbindsMount` instead

vsoch:
okay I got it working to return an error:

```bash
$ singularity shell --bind /tmp --bind /tmp docker://busybox
FATAL:   container creation failed: unable to add /tmp to mount list: destination is already in the mount point list
```
What's the best way to compare (the variable type? or message) in network.go?

cclerget:
Like this :
```
if err := s.AddBind(...); err != nil && err == mount.ErrMountExist {
    sylog.Warningf(...)
} else if err != nil {
    return fmt.Errorf(...)
} else {
    system.Points.AddRemount(mount.UserbindsTag, dst, flags)
    flags &^= syscall.MS_RDONLY
}
```

vsoch:
I think I got at!

internal/pkg/util/fs/mount/mount.go
```diff
@@ -339,7 +348,8 @@ func (p *Points) add(tag AuthorizedTag, source string, dest string, fstype strin
 			}
 		}
 		if present {
-			return fmt.Errorf("destination %s is already in the mount point list", dest)
+			return ErrMountExists
+
```
bauerm97: Remove this empty line here. I would suggest that actually a new blank line can be added after the next `}` character (add empty line after line 353)

bauerm97: Remove this empty line here. I would suggest that actually a new blank line can be added after the next } character (add empty line after line 353)

internal/pkg/util/fs/mount/mount_test.go
```diff
@@ -59,8 +59,8 @@ func TestImage(t *testing.T) {
 	if err := points.AddImage(RootfsTag, "/fake", "/", "squashfs", 0, 0, 0); err == nil {
 		t.Errorf("should have failed with 0 size limit")
 	}
-	if err := points.AddImage(RootfsTag, "/fake", "/squash", "squashfs", 0, 0, 10); err == nil {
-		t.Errorf("should have failed with destination already in list")
+	if err := points.AddImage(RootfsTag, "/fake", "/squash", "squashfs", 0, 0, 10); err != nil {
```
bauerm97: This line was updated incorrectly here. This function still needs a non-nil exit. I would say this:

`if err := [...]; err == nil || (err != nil && err != mount.ErrMountExists) {...}`

bauerm97 (changes requested):

internal/pkg/util/fs/mount/mount_test.go
bauerm97: Actually, I would go as far as doing this:

```
if err := [...]; err == nil {
    t.Errorf("nil error returned, should have returned non-nil mount.ErrMountExists")
} else if err != nil && err != mount.ErrMountExists {
    t.Errorf("non-nil error should have been mount.ErrMountExists")
}
```

This way if tests are failing we have information here about _why_ the error is happening

vsoch: okay I'll give that a try!

vsoch:
It's green! It's green!

bauerm97 (approved):

PR #2855, opened 2019-03-11
Hypervisor fix

jmstover:

Hypervisor fixes

* Add xhyve executable to the start of args list
* Return nil on hypervisor non-zero exit. For now.


cmd/internal/cli/startvm.go
```diff
@@ -85,6 +85,7 @@ func startVM(sifImage, singAction, cliExtra string, isInternal bool) error {
 			//Program exited with non-zero return code
 			if status, ok := exitErr.Sys().(syscall.WaitStatus); ok {
 				sylog.Debugf("Process exited with non-zero return code: %d\n", status.ExitStatus())
+                return nil
```
WestleyK: Gofmt will complain about this...

Fix: 
```
$ gofmt -w cmd/internal/cli/startvm.go
```

jmstover: Gah...  It *looked* right in the editor

WestleyK: I know!!! gofmt like tabs :man_facepalming:

cmd/internal/cli/startvm_darwin.go
```diff
@@ -21,6 +22,7 @@ func getHypervisorArgs(sifImage, bzImage, initramfs, singAction, cliExtra string
 
 	// Default xhyve Arguments
 	args := []string{
+		buildcfg.LIBEXECDIR + "/singularity/vm/xhyve",
```
bauerm97: `filepath.Join(buildcfg.LIBEXECDIR, "singularity/vm/xhyve")`

cmd/internal/cli/startvm.go
```diff
@@ -85,6 +85,7 @@ func startVM(sifImage, singAction, cliExtra string, isInternal bool) error {
 			//Program exited with non-zero return code
 			if status, ok := exitErr.Sys().(syscall.WaitStatus); ok {
 				sylog.Debugf("Process exited with non-zero return code: %d\n", status.ExitStatus())
+				return nil
```
bauerm97, line 88: Why returning nil here? If anything, if the VM is exiting with non-zero, we should _also_ be exiting with that same error code.

bauerm97 (changes requested):

cmd/internal/cli/startvm.go
jmstover: Because xhyve is *always* exiting with a non-zero exit code. Even running in the shell. I *think* it comes because we're triggering an ACPI RTC error, but it doesn't have any functional implications ... except xhyve is exiting with 1.  It's why I put the "For Now" in the message. If I can figure out a work around to make it stop, that will be used instead of this.

But us throwing a sylog.Errorf(), when there actually isn't an error every run isn't good.

bauerm97 (approved):

mem (approved):

PR #2857, opened 2019-03-11
Do not list all plugins when uninstalling

mem:
Since the filename of the meta file is computable from the plugin's
name, just compute the expected filename and use that to locate the
plugin to uninstall.


internal/pkg/plugin/binary.go
```diff
@@ -115,24 +115,38 @@ func InstallFromSIF(fimg *sif.FileImage, libexecdir string) (*Meta, error) {
...
+	defer fh.Close()
+
+	m, err := LoadFromJSON(fh)
+	if err != nil {
+		return nil, err
+	}
+
+	// make sure we loaded the right thing
+	if m.Name != name {
+		return nil, fmt.Errorf("Unexpected plugin name %q when loading plugin %q", m.Name, name)
```
bauerm97: `fmt.Errorf(...)` strings should not start with capital:

`fmt.Errorf("unexpected [...]")`

internal/pkg/plugin/binary.go
```diff
@@ -115,24 +115,38 @@ func InstallFromSIF(fimg *sif.FileImage, libexecdir string) (*Meta, error) {
...
 func Uninstall(name, libexecdir string) error {
-	sylog.Debugf("Uninstalling plugin %q from %q", name, libexecdir)
-
-	metas, err := List(libexecdir)
+	pluginDir := filepath.Join(libexecdir, DirRoot)
+	sylog.Debugf("Uninstalling plugin %q from %q", name, pluginDir)
 
+	meta, err := loadMetaByName(name, pluginDir)
 	if err != nil {
+		// figure out if this is a not found error?
```
bauerm97, line 123: +1 here, let's tackle this in a follow up PR. I think there is actually a decent chunk of useful work we can do with plugins error. In `pkg/plugin` we can add a `pkg/plugin/error.go` or similar where we define many `pluginapi` specific errors.

bauerm97 (changes requested):
Looks mostly good. One comment here is just a comment for follow-up work, one small change I saw otherwise

bauerm97 (approved):

PR #2861, opened 2019-03-12
Use SCS Key Service Client

### LINKED ISSUE #2786: Use SCS Key Service Client
The Singularity Container Services (SCS) team at Sylabs has recently released a client package for the Key Service (docs [here](https://godoc.org/github.com/sylabs/scs-key-client/client)). This client has support for pagination, which Singularity does not currently handle properly.

This client should replace some of the code in `pkg/sypgp`.

tri-adam:

Use `github.com/sylabs/scs-key-client` to interact with key server, which adds support for pagination and allows detailed error messages to be displayed (when applicable.) Update `github.com/sylabs/json-resp` to `v0.5.0`. Improve testing for `sypgp` package.

- Fixes #2786 


pkg/sypgp/sypgp_test.go
```diff
@@ -24,6 +29,168 @@ var (
...
+		uri       string
+		authToken string
+		wantErr   bool
+	}{
+		{"Success", http.StatusOK, openpgp.EntityList{testEntity}, "search", srv.URL, "", false},
+		{"SuccessToken", http.StatusOK, openpgp.EntityList{testEntity}, "search", srv.URL, "token", false},
+		{"Unauthorized", http.StatusUnauthorized, nil, "search", srv.URL, "", true},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, test.WithoutPrivilege(func(t *testing.T) {
```
bauerm97: I really would prefer that we don't use `test.WithoutPrivilege` or `test.WithPrivilege` functions unless we absolutely *have* to call those. That package causes problems with running tests on MacOS

bauerm97: I really would prefer that we don't use `test.WithoutPrivilege` or `test.WithPrivilege` functions unless we absolutely _have_ to call those. That package causes problems with running tests on MacOS

pkg/sypgp/sypgp.go
```diff
@@ -487,138 +489,118 @@ func helpAuthentication() (token string, err error) {
...
 }
 
-// doSearchRequest prepares an HKP search request
-func doSearchRequest(search, keyserverURI, authToken string) (*http.Request, error) {
-	v := url.Values{}
-	v.Set("search", search)
-	v.Set("op", "index")
-	v.Set("fingerprint", "on")
+// SearchPubkey connects to a key server and searches for a specific key
+func SearchPubkey(search, keyserverURI, authToken string) error {
```
bauerm97, line 493: A small question here, but these functions look like code which belongs in `internal/app/singularity`, which is what we will take care of in a follow up Pull Request, correct?

pkg/sypgp/sypgp.go
```diff
@@ -487,138 +489,118 @@ func helpAuthentication() (token string, err error) {
...
 }
 
-// doFetchRequest prepares an HKP get request
-func doFetchRequest(fingerprint, keyserverURI, authToken string) (*http.Request, error) {
-	v := url.Values{}
-	v.Set("op", "get")
-	v.Set("options", "mr")
-	v.Set("search", "0x"+fingerprint)
+// FetchPubkey pulls a public key from the Key Service.
+func FetchPubkey(fingerprint, keyserverURI, authToken string, noPrompt bool) (openpgp.EntityList, error) {
```
bauerm97, line 558: Is `Fetch` the common nomenclature for this type of function?

pkg/sypgp/sypgp.go
```diff
@@ -487,138 +489,118 @@ func helpAuthentication() (token string, err error) {
...
+	var fp [20]byte
+	b, err := hex.DecodeString(fingerprint)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to decode fingerprint: %v", err)
 	}
-	u.Path = "pks/lookup"
-	u.RawQuery = v.Encode()
+	if got, want := len(b), len(fp); got != want {
+		return nil, fmt.Errorf("unexpected fingerprint length: %v", len(b))
```
bauerm97: Since we're assigning `got, want := [...]`, I think we can just do `return nil, fmt.Errorf("unexpected fingerprint length: %v", got)`

Perhaps even better would be:

`fmt.Errorf("unexpected fingerprint length: %v; expected length: %v", got, want)`

pkg/sypgp/sypgp.go
```diff
@@ -628,79 +610,57 @@ func FetchPubkey(fingerprint, keyserverURI, authToken string, noPrompt bool) (op
...
-	u.Path = "pks/add"
 
-	r, err := http.NewRequest(http.MethodPost, u.String(), strings.NewReader(v.Encode()))
-	if err != nil {
-		return nil, err
-	}
-	if authToken != "" {
-		r.Header.Set("Authorization", fmt.Sprintf("BEARER %s", authToken))
+	if err = e.Serialize(wr); err != nil {
+		wr.Close()
```
bauerm97, line 625: Possible to just do a `defer wr.Close()` before line 624 since we're calling `wr.Close()` before we return in either execution pathway?

pkg/sypgp/sypgp.go
```diff
@@ -628,79 +610,57 @@ func FetchPubkey(fingerprint, keyserverURI, authToken string, noPrompt bool) (op
...
-			return fmt.Errorf("could not obtain or install authentication token: %s", err)
-		}
-		// try request again
-		r, err := doPushRequest(w, keyserverURI, token)
-		if err != nil {
-			return fmt.Errorf("error while preparing http request: %s", err)
+			// Try to push key to Key Service again with new auth token.
+			c.AuthToken = authToken
+			if err := c.PKSAdd(context.TODO(), keyText); err != nil {
+				return err
```
bauerm97: Should we add some context to this error return with `return fmt.Errorf(...)`?

pkg/sypgp/sypgp.go
```diff
@@ -628,79 +610,57 @@ func FetchPubkey(fingerprint, keyserverURI, authToken string, noPrompt bool) (op
...
+			}
+		} else {
+			return fmt.Errorf("Key server did not accept PGP key: %v", err)
```
bauerm97: `fmt.Errof(...)` should not begin with capital letter. Instead: `fmt.Errorf("key server [...]")`

bauerm97 (changes requested):
Looks good so far, I think we need a few small tweaks. Also had a few generic questions to try and understand better the context. Cheers

pkg/sypgp/sypgp.go
tri-adam: Yes, there is an issue open for that (#2786 )

pkg/sypgp/sypgp.go
tri-adam: Not sure, just re-using what was here already. The SCS Key Client equivalent is `GetKey`.

pkg/sypgp/sypgp.go
tri-adam: `wr` must be closed before `w.String()` will return the entire serialized entity below.

pkg/sypgp/sypgp_test.go
tri-adam: Fair enough, will do.

pkg/sypgp/sypgp.go
tri-adam: Will do.

pkg/sypgp/sypgp.go
tri-adam: Will do.

pkg/sypgp/sypgp.go
tri-adam: Will do.

bauerm97 (approved):

PR #2865, opened 2019-03-12
Add missing DisableColor when building without +sylog

mem:
When building without the sylog tag, the build is broken because the
sylog package is missing the new DisableColor color function. This
causes, among other things, guru to stop working.


jscook2345 (approved):
LGTM!

WestleyK (approved):
Looks good to me! :+1:

PR #2874, opened 2019-03-12
Implement plugin enable / disable

### LINKED ISSUE #2322: Plugin enable and disable commands
Plugin administration commands:

- `singularity plugin enable`
- `singularity plugin disable`

This command should be a relatively small task, as simple as moving the actual `.so` file in and out of the directory where the singularity looks for plugins

mem:
Implement all the plugin enable / disable functionality.

- Fixes #2322 

### DIFF
--- a/cmd/internal/cli/plugin_disable_linux.go
+++ b/cmd/internal/cli/plugin_disable_linux.go
@@ -1,0 +1,34 @@
+package cli
+
+import (
+	"os"
+
+	"github.com/spf13/cobra"
+	"github.com/sylabs/singularity/docs"
+	"github.com/sylabs/singularity/internal/app/singularity"
+	"github.com/sylabs/singularity/internal/pkg/buildcfg"
+	"github.com/sylabs/singularity/internal/pkg/sylog"
+)
+
+// PluginDisableCmd disables the named plugin
+//
+// singularity plugin disable <name>
+var PluginDisableCmd = &cobra.Command{
+	Run: func(cmd *cobra.Command, args []string) {
+		err := singularity.DisablePlugin(args[0], buildcfg.LIBEXECDIR)
+		if err != nil {
+			if os.IsNotExist(err) {
+				sylog.Fatalf("Failed to disable plugin %q: plugin not found.", args[0])
+			} else {
+				sylog.Fatalf("Failed to disable plugin %q: %s.", args[0], err)
+			}
+		}
+	},
+	DisableFlagsInUseLine: true,
+	Args:                  cobra.ExactArgs(1),
+
+	Use:     docs.PluginDisableUse,
+	Short:   docs.PluginDisableShort,
+	Long:    docs.PluginDisableLong,
+	Example: docs.PluginDisableExample,
+}
--- a/cmd/internal/cli/plugin_disable_linux.go
+++ b/cmd/internal/cli/plugin_disable_linux.go
@@ -18,9 +18,9 @@ var PluginDisableCmd = &cobra.Command{
 		err := singularity.DisablePlugin(args[0], buildcfg.LIBEXECDIR)
 		if err != nil {
 			if os.IsNotExist(err) {
-				sylog.Fatalf("Failed to disable plugin %q: plugin not found.", args[0])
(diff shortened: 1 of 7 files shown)


cmd/internal/cli/plugin_enable_linux.go
```diff
@@ -0,0 +1,34 @@
...
+// PluginEnableCmd enables the named plugin
+//
+// singularity plugin enable <name>
+var PluginEnableCmd = &cobra.Command{
+	Run: func(cmd *cobra.Command, args []string) {
+		err := singularity.EnablePlugin(args[0], buildcfg.LIBEXECDIR)
+		if err != nil {
+			if os.IsNotExist(err) {
+				sylog.Fatalf("Failed to enable plugin %q: plugin not found.", args[0])
+			} else {
```
bauerm97: `} else {` not necessary here. Might be simpler as:

```
if os.IsNotExist(err) {
    sylog.Fatalf([...])
}
sylog.Fatalf([...])
```

cmd/internal/cli/plugin_disable_linux.go
```diff
@@ -0,0 +1,34 @@
...
+// PluginDisableCmd disables the named plugin
+//
+// singularity plugin disable <name>
+var PluginDisableCmd = &cobra.Command{
+	Run: func(cmd *cobra.Command, args []string) {
+		err := singularity.DisablePlugin(args[0], buildcfg.LIBEXECDIR)
+		if err != nil {
+			if os.IsNotExist(err) {
+				sylog.Fatalf("Failed to disable plugin %q: plugin not found.", args[0])
+			} else {
```
bauerm97: Same comment

internal/pkg/plugin/binary.go
```diff
@@ -149,6 +182,17 @@ func loadMetaByName(name, plugindir string) (*Meta, error) {
 	return m, nil
 }
 
+func loadMetaByFilename(filename string) (*Meta, error) {
+	fh, err := os.Open(filename)
+	if err != nil {
+		sylog.Debugf("Error opening meta file %q: %s\n", filename, err)
```
bauerm97, line 188: `sylog.Errorf`

internal/pkg/plugin/binary.go
```diff
@@ -120,23 +120,56 @@ func Uninstall(name, libexecdir string) error {
...
 
 	sylog.Debugf("Found plugin %q, meta=%#v", name, meta)
+
 	return meta.uninstall()
 }
 
-func loadMetaByName(name, plugindir string) (*Meta, error) {
-	fh, err := os.Open(metaPath(plugindir, name))
+// Enable enables the plugin named "name" found under "libexecdir"
+func Enable(name, libexecdir string) error {
```
bauerm97, line 132: It does seem like these functions here are more like "application code" rather than the underlying backing package. I think that the `internal/app/singularity/remote_*` files are good examples of this. I think we revisit this in follow up PRs as we flesh out the API we're exposing here in `internal/pkg/plugin`

bauerm97 (changes requested):

internal/pkg/plugin/binary.go
mem: I don't actually want that. The debug entry is there just in case it's needed (you know, for debugging :-) but for the user I would like to have a different error. That's why I'm checking whether it's a not exist error elsewhere.

cmd/internal/cli/plugin_enable_linux.go
mem: I was trying to make it non-ambiguous that it's handling "all the other errors" without forcing the reader to keep in mind that "sylog.Fatalf" interrupts the program, but I can change this. Originally, I had a switch here, but since it's only one special case, it looks weird.

internal/pkg/plugin/binary.go
mem: Ok

cmd/internal/cli/plugin_enable_linux.go
mem: Changed it to what @bauerm97 is asking for.

bauerm97 (approved):

PR #224, opened 2016-09-23
reorganizing python scripts to close #10, and fixing bug to close #11

vsoch:
This PR will do two things:
## Python / Docker Organization
- reorganize the python folder to add specificity to the "submodules." They aren't official modules (all imports are relative and the scripts added via Makefiles) but we can now better organize python scripts using folders. If we need to make an "official" module at some point we can discuss, but I think it's overkill.
  - When we have more extensive docker functionality, I'm going to eventually break apart the docker/api.py script into more specific functions, for now it's all under "api.py"
## Runscript permissions bug

I was trying to change a file called "runscript" instead of "singularity." I don't even know how I put my shoes on the right feet in the morning.

I just tested this for the bootstrap and it looked good, we probably should have more testing for other stuffs! I can take a look at the continuous integration at some point and add more tests.

PR #2879, opened 2019-03-13
Enable Remotes for Action Commands

ikaneshiro:

Missed `library` uris in `action` commands in #2829. This fixes that by checking for a default remote and using that.


bauerm97 (approved):

bauerm97 (dismissed):
Actually wait, looks like you broke CI

cmd/internal/cli/actions.go
```diff
@@ -210,6 +211,32 @@ func setVM(cmd *cobra.Command) {
...
 }
 
+// returns url for library and sets auth token based on remote config
+// defaults to https://library.sylabs.io
+func handleActionRemote(cmd *cobra.Command) (remoteURL string) {
+	remoteURL = "https://library.sylabs.io"
+
+	// if we can load config and if default endpoint is set, use that
+	// otherwise fall back on regular authtoken and URI behavior
+	e, err := sylabsRemote(remoteConfig)
```
mem: could you please name the variable `endpoint` (or something else more descriptive if you prefer)? It's not immediately obvious from the function name what this value is supposed to be, which makes the rest of the code a little hard to read.

cmd/internal/cli/actions.go
```diff
@@ -210,6 +211,32 @@ func setVM(cmd *cobra.Command) {
 	}
 }
 
+// returns url for library and sets auth token based on remote config
+// defaults to https://library.sylabs.io
+func handleActionRemote(cmd *cobra.Command) (remoteURL string) {
```
mem: Please no named returns. See [here](https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters).

That takes care of the naked return on lines #224 and #237.

Also, because of the way this function works, it's probably a good idea to return both the remote URL and an error in case it cannot be obtained.

cmd/internal/cli/actions.go
```diff
@@ -210,6 +211,32 @@ func setVM(cmd *cobra.Command) {
...
+	e, err := sylabsRemote(remoteConfig)
+	if err == scs.ErrNoDefault {
+		sylog.Warningf("No default remote in use, falling back to %v", remoteURL)
+		return
+	} else if err != nil {
+		sylog.Fatalf("Unable to load remote configuration: %v", err)
+	}
+
+	authToken = e.Token
+	uri, err := e.GetServiceURI("library")
```
mem: This is probably better written as:

```
remoteURL, err := e.GetServiceURI("library")
if err != nil {
    sylog.Warning("Unable to get library service URI: %v", err)
}
return remoteURL // or return remoteURL, err if you modify the return value as suggested
```

mem (changes requested):

cmd/internal/cli/actions.go
ikaneshiro: Will do. I'll also change the other places in the CLI where I was just using `e`

cmd/internal/cli/actions.go
ikaneshiro: Refactored this function a bit to get rid of the named and naked returns. I think it's ok to just warn or exit on an error here. I don't see a need for the caller to handle the error.

cmd/internal/cli/actions.go
ikaneshiro: This doesn't quite work since `GetServiceURI()` returns `""` in the case of an error and we want to always return a correct `URI` from this function. During the refactoring for the previous comment I think I address this.

mem (approved):
Thanks!

PR #2884, opened 2019-03-13
Fix `go vet` for macos

### LINKED ISSUE #2629: `go vet` fails for darwin platform
## Version of Singularity:

master

### Expected behavior

`make -C builddir check` passes.

### Actual behavior

`make -C builddir check` fails with:

<pre>
 CHECK go fmt
       PASS
 CHECK go vet
internal/pkg/runtime/engines/oci/process.go:30:2: build constraints exclude all Go files in /Users/justin/go/src/github.com/sylabs/singularity/pkg/util/rlimit
internal/pkg/runtime/engines/oci/create.go:22:2: build constraints exclude all Go files in /Users/justin/go/src/github.com/sylabs/singularity/pkg/util/sysctl
make: *** [check] Error 1
</pre>

On a MacOS system:

    ./mconfig
    make -C builddir check

jscook2345:

Rename source files that require a `linux` build target so that `go vet` works on macos

- Fixes #2629 

[ ] Read the [Guidelines for Contributing](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
[ ] Added changes to the [CHANGELOG](https://github.com/sylabs/singularity/blob/master/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
[ ] Added tests to validate this PR and tested this PR locally with a `make testall`
[ ] Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
[ ] Added myself as a contributor to the [Contributors File](https://github.com/sylabs/singularity/blob/master/CONTRIBUTORS.md)


.circleci/config.yml
```diff
@@ -121,6 +139,9 @@ workflows:
       - go1.11-alpine:
           requires:
             - get_source
+      - go1.11-macos:
```
bauerm97: Are we able to add a MacOS job into a workflow that is usually running on Linux?

internal/pkg/test/privilege_darwin.go
```diff
@@ -0,0 +1,136 @@
+// Copyright (c) 2019, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package test
```
bauerm97, line 6: To be honest, I think all of these can just be no-op functions for Darwin. I can't (yet) think of a scenario where we need do testing in this way on Darwin

bauerm97, line 6: To be honest, I think all of these can just be no-op functions for Darwin. I can't (yet) think of a scenario where we need do testing in this way on Darwin

jscook2345:
@bauerm97 

Some of the functions in there are used in unit testing so I'd like to keep as many around as possible. Once I mess around with testing a bit and see how good / bad things are than I think we can revisit the no-ops.

internal/pkg/test/privilege_darwin.go
```diff
@@ -0,0 +1,136 @@
...
+	if uid != 0 {
+		t.Fatal("test must be run with privilege")
+	}
+}
+
+// DropPrivilege drops privilege. Use this at the start of a test that does
+// not require elevated privileges. A matching call to ResetPrivilege must
+// occur before the test completes (a defer statement is recommended.)
+func DropPrivilege(t *testing.T) {
+	t.Fatalf("Currently not supported for Darwin")
```
mem: Shouldn't this be t.Skip("Currently not supported on Mac OS")?

(It's about "Skip" vs "Fatal", not "Darwin" vs "Mac OS"; I wrote "Mac OS" because I have no idea whether Mac OS users in general know what "darwin" is)

internal/pkg/test/privilege_darwin.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// ResetPrivilege returns effective privilege to the original user.
+func ResetPrivilege(t *testing.T) {
+	t.Fatalf("Currently not supported for Darwin")
```
mem: same here

jscook2345:
I changed over to a Skip per @mem

mem (approved):

PR #2889, opened 2019-03-13
Plugin loading

mem:
Currently plugin loading is failing because:

- The code is trying to load the meta file instead of the ELF one
- It's trying to compile the plugin with a potentially different version of the Go compiler
- It's not building the code with the same Go build tags

Fix all of these to get loading to work.

### DIFF
--- a/internal/app/singularity/plugin_compile_linux.go
+++ b/internal/app/singularity/plugin_compile_linux.go
@@ -6,13 +6,13 @@
 package singularity
 
 import (
+	"errors"
 	"fmt"
 	"go/build"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"runtime"
-	"strings"
 
 	uuid "github.com/satori/go.uuid"
 	"github.com/sylabs/sif/pkg/sif"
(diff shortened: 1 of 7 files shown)


internal/pkg/plugin/load.go
```diff
@@ -26,24 +27,53 @@ func assertInitialized() {
...
+			// grabbing only the textual description of the
+			// error
+			wrappedErr := fmt.Errorf("while initializin plugin %q: %s", meta.Name, err)
+			errs = append(errs, wrappedErr)
 		}
 	}
 
-	initialized = true // set initialized to true
-	return nil
+	switch len(errs) {
```
bauerm97: I really think this switch statement is unnecessary. I think we can just do this:

```
if len(errs) == 0 {
    initialized = true 
    return nil
}

// Collect all the errors into a single one that can be
// returned. See similar comment in List(...) for more
// details.
var b strings.Builder

for i, err := range errs {
	if i > 0 {
		b.WriteString("; ")
	}
	b.WriteString(err.Error())
}

return errors.New(b.String())
```

bauerm97: The only thing the switch statement is bringing us is directly returning the `err` in the case where there is only 1 error. I don't think it's too important

internal/app/singularity/plugin_compile_linux.go
```diff
@@ -68,15 +68,75 @@ func CompilePlugin(sourceDir, destSif string) error {
 	return nil
 }
 
+// lookPath mostly copied from Go's exec.LookPath; modified to take a
+// path as an argument
+func lookPath(file, path string) (string, error) {
```
bauerm97: To be honest, I'm really skeptical about this. I don't understand why we can't just use PATH

bauerm97 (changes requested):

bauerm97 (approved):

PR #2890, opened 2019-03-14
Do not list ipcs in nvidia bind list

sashayakovtseva:


mem (approved):

GodloveD:
What's the reason for this change?  Will it break processes that want to use nvlink for instance?

sashayakovtseva:
@GodloveD As I have been testing device plugin on AWS, it appeared that `--ipcs` returns one additional file to mount - `/run/nvidia-persistenced/socket`. But Singularity engine mounts it to `/usr/bin/socket` so I assume it is not even used. Also after talking to @cclerget we decided to remove `--ipcs` b/c it should only be accessible by host. Correct me if I am wrong.

WestleyK:
... The copyright should be `2018-2019`?? 

https://github.com/sylabs/singularity/pull/2890/files#diff-40a3c8e47c91bcd7d9215c285a6e9062R1

PR #2891, opened 2019-03-14
Move Go version checks completely to build time

mem:
The current method of checking the Go version is implemented partly at
build time and partly at runtime. At build-time the compiler selects
between to versions of the same function, one that produces an error if
the Go version is not supported and one that does not produce an error.
At runtime, this function is called, and the error is checked. This
means that the user has to build, install and run the software in order
to figure out that the Go version is not supported.

With this change the check happens not only at build-time, but it
happens very quickly. Near the end of the output, there's a line that
reads:

    ../internal/pkg/util/goversion/version_check.go:19:9: undefined: __BUILD_REQUIRES_GO_VERSION_1_10_OR_LATER__

which should be clear enough for anyone trying to build singularity on
their own. It has the additional side-effect of causing CI to fail fast,
too.


sashayakovtseva (approved):

internal/pkg/util/goversion/goversion.go
```diff
@@ -7,7 +7,12 @@
...
-	return nil
-}
+// __BUILD_REQUIRES_GO_VERSION_1_10_OR_LATER__ provides a human-readable
+// error message when building this package with an unsupported version
+// of the Go compiler.
+//
+// Keep the name of this variable in sync with the minimum required
+// version specified in the build tag above.
+//
+// nolint:golint
```
mem, line 17: and this is the reason for wanting to switch to golangci-lint: to be able to add exceptions :-\

The ALL CAPS variable is fine in this case, as it works better for the intended purpose in that way.

sashayakovtseva: btw as I know such comments are not indented with space in order not to include them in docs, so better do `//nolint:golint` instead of `// nolint:golint`

PR #2897, opened 2019-03-14
Run gofmt -s on a few files that can be simplified

mem:
The simplifications are all of the form:

	[]T{T{...}, T{...}, ...}

to

	[]T{{...}, {...}, ...}

In other words, when you have an array of a particular type, you don't
need to name the type for every single element of the array.

The resulting code is less cluttered and easier to read.


jscook2345 (approved):
👍

PR #2898, opened 2019-03-14
Clean up imports using goimports

mem:
goimports will rewrite some imports to conform to recommended style. In
this cases a couple kicked in (std library first, followed by a blank
line; explicitly naming imports if the package name does not match the
last element of the import path).

### DIFF
--- a/cmd/docs/docs.go
+++ b/cmd/docs/docs.go
@@ -7,6 +7,7 @@ package main
 
 import (
 	"fmt"
+
 	"github.com/spf13/cobra"
 	"github.com/spf13/cobra/doc"
 	"github.com/sylabs/singularity/cmd/internal/cli"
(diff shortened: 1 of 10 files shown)


jscook2345 (commented):
One thing that is confusing to me.

`client "github.com/sylabs/singularity/pkg/client/library"`

vs

`shub "github.com/sylabs/singularity/pkg/client/shub"`

why do they both not use `client` as the short name?

mem:
> One thing that is confusing to me.
> 
> `client "github.com/sylabs/singularity/pkg/client/library"`
> 
> vs
> 
> `shub "github.com/sylabs/singularity/pkg/client/shub"`
> 
> why do they both not use `client` as the short name?

The goimports tool, which is the one that writes these entries based on the code, seems to be getting confused if I use "client" for the second case, I'm not sure why.

sashayakovtseva (approved):

PR #2899, opened 2019-03-14
Remove dead code as reported by linter

mem:


cclerget (approved):
LGTM

pkg/util/capabilities/config_test.go
```diff
@@ -75,25 +70,6 @@ func TestReadFromWriteTo(t *testing.T) {
 
 }
 
-func checkReadFrom(t *testing.T, confExpected Config, shouldFail bool) {
```
bauerm97, line 78: Why are we deleting this here?

internal/pkg/remote/remote_test.go
```diff
@@ -14,16 +14,6 @@ import (
 	useragent "github.com/sylabs/singularity/pkg/util/user-agent"
 )
 
-var remoteConfig = Config{
```
bauerm97, line 17: Why is this dead code?

bauerm97 (changes requested):

pkg/util/capabilities/config_test.go
mem: Nothing is using the function. :man_shrugging:

internal/pkg/remote/remote_test.go
mem: nothing is using the variable

bauerm97 (approved):

PR #2901, opened 2019-03-14
Fix issues reported by maligned

mem:
maligned reports cases of structs that might be larger than they need to
be because of padding necessary to align types to the natural alignment.

Most of the fixes are simply moving the smaller members of a struct to
the front of the struct so that only minimal padding is needed.


cclerget (approved):
LGTM

PR #2902, opened 2019-03-14
Fix ineffective assignments as reported by ineffassign

mem:
There are assignments without any effect, for example:

	i := getValue(0)
	i = getValue(0)
	if i > 0 {
		// ...
	}

the first assignment has no effect because it's immediately overwritten
by the second assignment. In some cases this is done because there's a
side-effect to calling a function (in which case the value should be
discarded explictly), and in some cases it's an oversight.


cclerget (approved):
LGTM

PR #2918, opened 2019-03-18
Print appropriate error when /bin/sh is missing.

### LINKED ISSUE #2386: Exec fails with no such file or directory
## Version of Singularity:

3.0.1-112.g5655f87.dirty

### Expected behavior

Command is executed inside a container successfully.

### Actual behavior

Command execution results in `exec /.singularity.d/actions/exec failed: no such file or directory`.

1. sudo singularity -d build echo.sif docker://hashicorp/http-echo
2. singularity exec echo.sif ls

Output: 
```
WARNING: passwd file doesn't exist in container, not updating
WARNING: group file doesn't exist in container, not updating
FATAL:   exec /.singularity.d/actions/exec failed: no such file or directory
```

phpHavok:

This PR prints a descriptive error when a container fails to exec due to `/bin/sh` missing.

- Related #2386


sashayakovtseva:
@phpHavok I think this PR is not fixing the problem, but improves error message only.

phpHavok:
@sashayakovtseva Ah, yes that is true. I may have misunderstood the thread on #2386 but it seemed like @jscook2345 just wanted the error message updated.

sashayakovtseva:
@phpHavok that is perfectly fine, you'll just need to remove `Fixes #...` so that issue will not be closed when your PR is merged. You can change `Fixes` to `Related` and then issue will not be affected

phpHavok:
@sashayakovtseva Ah, I see :smile: I updated it it. Thanks!

internal/pkg/runtime/engines/singularity/process_linux.go
```diff
@@ -193,6 +193,9 @@ func (engine *EngineOperations) StartProcess(masterConn net.Conn) error {
 
 	if (!isInstance && !shimProcess) || bootInstance || engine.EngineConfig.GetInstanceJoin() {
 		err := syscall.Exec(args[0], args, env)
+		if _, errNoShell := os.Stat("/bin/sh"); os.IsNotExist(errNoShell) {
```
mem: As I mentioned in the issue, I think this is happening too late. We should check for `/bin/sh` much earlier (when we decide which wrapper to launch, in `checkExec`) so that we can provide a better error message there?

phpHavok: Resolved, I think. Thanks!

internal/pkg/runtime/engines/singularity/process_linux.go
```diff
@@ -35,6 +38,24 @@ func (engine *EngineOperations) checkExec() error {
 		shell = "/bin/sh"
 	}
 
+	// Make sure the shell exists.
+	if _, err := os.Stat(shell); os.IsNotExist(err) {
+		return fmt.Errorf("shell %s doesn't exist in container", shell)
+	}
+
+	// Verify architecture of image matches architecture of Go.
```
cclerget: We can't assume image is always SIF, could be a sandbox, a single squashfs/ext3 image. After thinking, architecture check should be after `syscall.Exec` instead and if `err == ENOENT` because we need to be able to run things like `singularity shell -B /usr/bin/qemu-ppc64le-static docker://ppc64le/ubuntu`, and adding architecture check here will always fail for this use case

internal/pkg/runtime/engines/singularity/process_linux.go
```diff
@@ -192,8 +213,8 @@ func (engine *EngineOperations) StartProcess(masterConn net.Conn) error {
 	}
 
 	if (!isInstance && !shimProcess) || bootInstance || engine.EngineConfig.GetInstanceJoin() {
-		err := syscall.Exec(args[0], args, env)
-		return fmt.Errorf("exec %s failed: %s", args[0], err)
+		syscall.Exec(args[0], args, env)
```
cclerget: keep `err` and check if `err == syscall.ENOENT` for shared library

cclerget (changes requested):

GodloveD:
Heya @phpHavok looks like @cclerget asked for some changes on this PR some time ago.  What do you think?  Would you like to refine this PR for inclusion?  😄

phpHavok:
@GodloveD @cclerget My bad! I got sidetracked on some other tasks and totally neglected to come back and work on this. I'll get it updated ASAP :)

phpHavok:
@cclerget How's this latest code look?
```
[jacob@gideon temp]$ singularity exec docker://hashicorp/http-echo ls
[...snip...]
WARNING: passwd file doesn't exist in container, not updating
WARNING: group file doesn't exist in container, not updating
FATAL:   shell /bin/sh doesn't exist in container
[jacob@gideon temp]$ singularity shell docker://ppc64le/ubuntu
FATAL:   image targets ppc64le, cannot run on amd64
[jacob@gideon temp]$ singularity exec library://ubuntu:18.04 ls
echo.sif  ubuntu_18.04.sif
```

cclerget:
Thank you @phpHavok ! Also tested by removing a shared library and code behave as expected. Great work :+1:

cclerget (approved):

PR #2919, opened 2019-03-19
Improved Cache List Output Summery

### LINKED ISSUE #2720: 'cache list' should also list a summery for all containers types
## Version of Singularity:

```
3.1.0-rc2.28.ga72e4274
```

### Expected behavior

`singularity cache list` should also list a summery for all cache:

```
$ singularity cache list
alpine_3.8.sif         2019-02-09 20:26:57    2.18 Mb          library
alpine_latest.sif      2019-02-09 20:26:27    2.18 Mb          library
alpine_latest.sif      2019-02-09 20:26:38    2.72 Mb          oci
alpine_3.8.sif         2019-02-09 20:27:07    2.19 Mb          oci

There 4 containers using: 9.27 Mbs, 6 oci blob file(s) using 4.96 Mbs of space.
Total used space: 14.23Mbs
```

There should also be a `-s, --summery` flag, so you can only list a cache summery, like this:

```
$ singularity cache list -s
There 4 container(s) using: 9.27 Mbs, 6 oci blob file(s) using 4.96 Mbs of space.
Total used space: 14.23Mbs
```
### Actual behavior

```
$ singularity cache list
alpine_3.8.sif         2019-02-09 20:26:57    2.18 Mb          library
alpine_latest.sif      2019-02-09 20:26:27    2.18 Mb          library
alpine_latest.sif      2019-02-09 20:26:38    2.72 Mb          oci
alpine_3.8.sif         2019-02-09 20:27:07    2.19 Mb          oci

There are 6 oci blob file(s) using 4.96 Mb of space. Use: '-T=blob' to list
```

```
$ singularity cache list
```

WestleyK:

This PR will improve the `singularity cache list` summery, heres a example:

```bash
$ singularity cache list 
alpine_latest.sif      2019-03-18 17:56:22    2.72 Mb          library

There 1 containers using: 2.72 Mb, 3 oci blob file(s) using 2.76 Mb of space.
Total space used: 5.48 Mb

# you can also list the summery:
$ singularity cache list -s

There 1 containers using: 2.72 Mb, 3 oci blob file(s) using 2.76 Mb of space.
Total space used: 5.48 Mb
```

- Fixes #2720 

<br>


cmd/internal/cli/cache_list_linux.go
```diff
@@ -47,7 +49,7 @@ var CacheListCmd = &cobra.Command{
 
 func cacheListCmd() error {
 
-	err := singularity.ListSingularityCache(cacheListTypes, allList)
+	err := singularity.ListSingularityCache(cacheListTypes, allList, cacheListSummery)
```
jchavezlavalle: change `cacheListSummery` to `cacheListSummary`

cmd/internal/cli/cache_list_linux.go
```diff
@@ -47,7 +49,7 @@ var CacheListCmd = &cobra.Command{
...
 	if err != nil {
 		sylog.Fatalf("Not listing cache; an error occured: %v", err)
```
jchavezlavalle: an error **occurred**

internal/app/singularity/cache_list_linux.go
```diff
@@ -31,82 +34,98 @@ func findSize(size int64) (string, error) {
 	return "", fmt.Errorf("failed to detect file size")
 }
 
-func listLibraryCache() error {
-	// loop through library cache
+// listLibraryCache will loop throught and list all library cache (~/.singularity/cache/library).
```
jchavezlavalle: loop **through**

internal/app/singularity/cache_list_linux.go
```diff
@@ -31,82 +34,98 @@ func findSize(size int64) (string, error) {
...
+// will return the amount of library containers, the total space thoughts containers are using,
+// and an error if one occures.
```
jchavezlavalle: and an error if one **occurs**

internal/app/singularity/cache_list_linux.go
```diff
@@ -31,82 +34,98 @@ func findSize(size int64) (string, error) {
...
+			if !listFiles {
+				fmt.Printf("%-22s %-22s %-16s %s\n", b.Name(), fileInfo.ModTime().Format("2006-01-02 15:04:05"), printFileSize, "oci")
+			}
+			count++
+			totalSize += fileInfo.Size()
 		}
 	}
-	return nil
+	return count, totalSize, nil
 }
```
jchavezlavalle, line 93: looks good, although the listOcicache, listLibraryCache and listBlobCache functions are very similar, would be nice to have a single listCache function handling the logic by passing in there a param to define whether it's blob, library or oci , this may need additional changes since these functions are called on other places but might be good to have it on mind for later on :+1:

internal/app/singularity/cache_list_linux.go
```diff
@@ -157,29 +166,59 @@ func ListSingularityCache(cacheListTypes []string, listAll bool) error {
...
+
+	if !listAll || cacheListSummery {
+		totalSpace = containerSpace + blobSpace
+		realTotalSpace, err := findSize(totalSpace)
+		if err != nil {
 			return err
 		}
+		realContainerSpace, _ := findSize(containerSpace)
+		realBlobSpace, _ := findSize(blobSpace)
+		fmt.Printf("\nThere %v containers using: %v, %v oci blob file(s) using %v of space.\n", containerCount, realContainerSpace, blobCount, realBlobSpace)
```
jchavezlavalle: update this line to:

```
fmt.Printf("\nThere are %v container(s) using: %v, %v oci blob file(s) using %v of space.\n", containerCount, realContainerSpace, blobCount, realBlobSpace)
    
```

Let's enforce the "There are x container(s)" message to avoid issue when there's a single container, please do change where applicable on other functions :+1:

jchavezlavalle (changes requested):
left some comments around @WestleyK thanks !

cmd/internal/cli/cache_list_linux.go
```diff
@@ -15,21 +15,23 @@ import (
...
+	allList          bool
+	cacheListSummery bool
 )
 
 func init() {
 	CacheListCmd.Flags().SetInterspersed(false)
 
-	CacheListCmd.Flags().StringSliceVarP(&cacheListTypes, "type", "T", []string{"library", "oci", "blobSum"}, "list cache type, choose between: library, oci, and blob")
-	CacheListCmd.Flags().SetAnnotation("type", "envkey", []string{"TYPE"})
+	CacheListCmd.Flags().StringSliceVarP(&cacheListTypes, "type", "T", []string{"library", "oci"}, "list cache type, choose between: library, oci, and blob")
```
mem: Since the value of the option is a list, please specify that multiple types can be specified by separating them with a comma. Something like `list of cache types, choose between: library, oci, and blob. Multiple values can be specified separating them with a comma.`

cmd/internal/cli/cache_list_linux.go
```diff
@@ -15,21 +15,23 @@ import (
...
+	CacheListCmd.Flags().SetAnnotation("type", "envkey", []string{"CACHE_LIST_TYPE"})
+
+	CacheListCmd.Flags().BoolVarP(&cacheListSummery, "summery", "s", false, "list cache summery")
```
mem: please change the description to read `display a cache summary`.

internal/app/singularity/cache_list_linux.go
```diff
@@ -16,6 +16,9 @@ import (
 	"github.com/sylabs/singularity/internal/pkg/sylog"
 )
 
+// findSize will take a int64 (like: '4781321234') and convert it to human readable output
+// in Kb, Mb, and Gb. output will be a string.
```
mem: since these are bytes and not bits (I assume), the SI units are "kB", "MB" and "GB". It's probably not a bad idea to add "TB" to the list, too.

And to prevent other people from having to make an assumption, it's probably a good idea to document that the input is bytes, so you could change the text to something along the lines of "findSize takes a size in bytes and converts it to a human-readable string representation expressing kB, MB or GB (whatever is smaller, but still larger than one)."

Also note that the actual code is using "Kb", "Mb", "Gb", so that should be fixed as well.

internal/app/singularity/cache_list_linux.go
mem: "list all elements in the ..."? "list all entries in the ..."?

internal/app/singularity/cache_list_linux.go
```diff
@@ -31,82 +34,98 @@ func findSize(size int64) (string, error) {
...
+// will return the amount of library containers, the total space thoughts containers are using,
+// and an error if one occures.
+func listLibraryCache(listFiles bool) (int, int64, error) {
+	var totalSize int64
+	count := 0
+
 	libraryCacheFiles, err := ioutil.ReadDir(cache.Library())
 	if err != nil {
-		return fmt.Errorf("unable to open library cache folder: %v", err)
+		return 0, 0, fmt.Errorf("unable to open library cache folder: %v", err)
```
mem: In general Linux prefers to call these things "directories" and not "folders" (for example, the function above is called ReadDir and not ReadFolder)

internal/app/singularity/cache_list_linux.go
```diff
@@ -31,82 +34,98 @@ func findSize(size int64) (string, error) {
...
 	}
 	for _, f := range libraryCacheFiles {
 		cont, err := ioutil.ReadDir(filepath.Join(cache.Library(), f.Name()))
```
mem: is this code trying to recursive list the contents of the cache? There's a function in the `filepath` package called `Walk`, that takes a directory and a function and it calls the function for each entry in the directory.

If that's not desired for some reason, this code should at least try to verify that f.Name() is a directory as expected, because ioutil.ReadDir will fail if you pass it a regular filename. You can use `os.Stat` to get a `os.FileInfo` structure and then call `IsDir` on that result to figure out if the entry is a directory before calling `ioutil.ReadDir`. In that way you can produce a warning about an unexpected file.

internal/app/singularity/cache_list_linux.go
mem: "the number of containers in the library"?

"the total space these containers"?

internal/app/singularity/cache_list_linux.go
```diff
@@ -31,82 +34,98 @@ func findSize(size int64) (string, error) {
...
-				return fmt.Errorf("unable to get stat for library cache: %v", err)
+				return 0, 0, fmt.Errorf("unable to get stat for library cache: %v", err)
 			}
 			printFileSize, err := findSize(fileInfo.Size())
 			if err != nil {
 				// no need to describe the error, since it is already
 				sylog.Warningf("%v", err)
 			}
-			fmt.Printf("%-22s %-22s %-16s %s\n", c.Name(), fileInfo.ModTime().Format("2006-01-02 15:04:05"), printFileSize, "library")
+			if !listFiles {
```
mem: the name `listFiles` suggests that if it's true, then will list the files. The code is doing the opposite. Maybe call the argument `onlyDirectories`? Or reverse the sense of the argument, so that if it's true, then the file information is listed?

internal/app/singularity/cache_list_linux.go
```diff
@@ -31,82 +34,98 @@ func findSize(size int64) (string, error) {
...
+			totalSize += fileInfo.Size()
 		}
 	}
-	return nil
+	return count, totalSize, nil
 }
 
-func listOciCache() error {
-	// loop through oci-tmp cache
+// listOciCache will list all you oci-tmp cache.
```
mem: `listOciCache lists all the entries in the OCI cache` maybe? or "temporary OCI cache"?

internal/app/singularity/cache_list_linux.go
```diff
@@ -120,24 +139,16 @@ func listBlobCache(printList bool) error {
...
-		fmt.Printf("\nThere are %d oci blob file(s) using %v of space. Use: '-T=blob' to list\n", count, printFileSize)
-	}
-	return nil
+	return count, totalSize, nil
 }
 
 // ListSingularityCache : list local singularity cache, typeNameList : is a string of what cache
 // to list (seprate each type with a comma; like this: library,oci,blob) allList : force list all cache.
-func ListSingularityCache(cacheListTypes []string, listAll bool) error {
+func ListSingularityCache(cacheListTypes []string, listAll, cacheListSummery bool) error {
```
mem: `cacheListSummary`

internal/app/singularity/cache_list_linux.go
mem: There are %d container(s) using %s; %d blob file(s) using %s

internal/app/singularity/cache_list_linux.go
```diff
@@ -16,6 +16,9 @@ import (
 	"github.com/sylabs/singularity/internal/pkg/sylog"
 )
 
+// findSize will take a int64 (like: '4781321234') and convert it to human readable output
+// in Kb, Mb, and Gb. output will be a string.
+// TODO: I'm sure theres a default 'human readable' function
 func findSize(size int64) (string, error) {
 	var sizeF float64
 	if size <= 1000000 {
```
mem: I know this is existing code and not what you are changing, but ...

The code cannot produce an error, as all possible values are covered (anything to 1000000; 1000001 to 1000000000; 1000000001 and up), so returning an error does not serve a useful purpose.

Note that because of the way the ranges are checked, findSize(1000000) will return "1000 kB" when it's likely that "1 MB" is expected. Same thing with findSize(1000000000) which returns "1000 MB" instead of "1 GB".

To help with the readability of this code, maybe this could be written like this:

```
var factor float64
var unit string
switch {
case size < 1e6:
    factor = 1e3
    unit = "kB"
case size < 1e9:
    factor = 1e6
    unit = "MB"
case size < 1e12:
    factor = 1e9
    unit = "GB"
default:
    factor = 1e12
    unit = "TB"
}

return fmt.Sprintf("%.2f %s, float(size) / factor, unit)
```

mem (commented):
Sorry for commenting outside the changes...

WestleyK:
> Sorry for commenting outside the changes...

@mem, No worries. This was one of my earlier PR, so I'm sure its not perfect :smile:

internal/app/singularity/cache_list_linux.go
WestleyK: I combined `listLibraryCache` and `listOciCache` into one, but `listOciCache` is a little different, thats still its own function.

jchavezlavalle (approved):

PR #2942, opened 2019-03-19
Fix RPM spec building on SLES 11

jmstover:
Go 1.8.3 is the newest Go version availible in the repositories so
we remove Go as a BuildRequire and expect it to be in the PATH
when rpmbuild is ran

libseccomp1 is to old to be used. It actually cannot be installed
as currently makeit will pick it up as existing and try using it.
Remove is as a BuildRequire as well.


jscook2345 (approved):
🥇

PR #2955, opened 2019-03-20
Add info messages when converting from SIF and removing temporary converted image

### LINKED ISSUE #2880: Problems with new unprivileged docker image run
## Version of Singularity:

master branch including #2838

### Expected behavior

similar behavior to singularity-2.6, or at least information about what is happening

### Actual behavior

When running unprivileged, docker containers are first converted to sif and then to a "sandbox" unpacked directory.  The step converting to "sandbox" can take quite a long time, and there is no notice this is happening unless debugging is enabled.   Then when the container exits, it takes an additional long time to delete the sandbox, again quietly doing it without any explanation.  One time when I did this while on an NFS home directory and not running any other processes I also got this error message:
```
ERROR:   failed to delete container image /cloud/login/dwd/.singularity/tmp/rootfs-848023522: remove /cloud/login/dwd/.singularity/tmp/rootfs-848023522/.nfs00000000a4960a57000000b1: device or resource busy
```
So it appears that something was holding open a file descriptor into the sandbox area while it was attempting to delete it.  I haven't looked into what yet.

I think at least the sandbox should also be kept in the cache, and an INFO message printed while it is being created.  In singularity-2.6 it didn't create the intermediate sif image, it just created the sandbox and used it, which was considerably faster.  I think it's OK to create the intermediate sif image, even though it will be unlikely to be used when running unprivileged, but only if the sandbox version is also kept in the cache so the next time it will run considerably faster.

The command I ran which resulted in the above error was:
```
singularity exec -u docker://centos bash
```
and then I exited the bash shell without running anything else.

I think this will be of interest to @afortiorama, @cclerget, and @bauerm97

DrDaveD:

Adds info messages to indicate what's happening when converting from SIF to sandbox and removing the sandbox.   This behavior was added in #2838.

- Partially fixes #2880

### DIFF
--- a/cmd/internal/cli/actions_linux.go
+++ b/cmd/internal/cli/actions_linux.go
@@ -421,6 +421,7 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 			unsquashfsPath = filepath.Join(d, "unsquashfs")
 		}
 		sylog.Verbosef("User namespace requested, convert image %s to sandbox", image)
+		sylog.Infof("Convert SIF file to sandbox...")
 		dir, err := convertImage(image, unsquashfsPath)
 		if err != nil {
 			sylog.Fatalf("while extracting %s: %s", image, err)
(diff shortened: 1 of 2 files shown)


cclerget (approved):
Thanks @DrDaveD

PR #2973, opened 2019-03-20
Enable "gosimple" linter

mem:
gosimple checks for common constructs that can be written in simpler
forms (for example "boolVar" vs "boolVar == true").


cclerget (approved):
LGTM

PR #2974, opened 2019-03-20
Prealloc slices where possible

mem:
Code like this:

    var names []string
    for n := range c.Remotes {
        names = append(names, n)
    }

knows in advance the required capacity for the slice, so it's possible
to avoid trashing in the allocator by simply preallocating the slice,
like this:

    names := make([]string, 0, len(c.Remotes))
    for n := range c.Remotes {
        names = append(names, n)
    }

### DIFF
--- a/internal/app/singularity/remote_list.go
+++ b/internal/app/singularity/remote_list.go
@@ -42,7 +42,7 @@ func RemoteList(usrConfigFile, sysConfigFile string) (err error) {
 	}
 
 	// list in alphanumeric order
-	var names []string
+	names := make([]string, 0, len(c.Remotes))
 	for n := range c.Remotes {
 		names = append(names, n)
 	}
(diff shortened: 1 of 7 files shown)


pkg/build/types/parser/deffile.go
```diff
@@ -200,7 +200,7 @@ func populateDefinition(sections map[string]string, d *types.Definition) (err er
 	// Files are parsed as a map[string]string
 	filesSections := strings.TrimSpace(sections["files"])
 	subs := strings.Split(filesSections, "\n")
-	var files []types.FileTransport
+	files := make([]types.FileTransport, 0, len(subs))
```
cclerget: Is CI failure related to this change ?

mem: good question, looking

mem: I fixed this in the most recent change, thanks!

cclerget (approved):
LGTM

GodloveD:
@mem so I guess you are all good to resolve the conflict and then merge this when you are able.

PR #3028, opened 2019-03-21
Do remote build when builder flag is set

### LINKED ISSUE #2985: build flag builder should default flag remote
## Version of Singularity:

3.1

### Expected behavior

when the `--builder` flag is used, we should use the `--remote` under the hood, to avoid 
` --remote --builder <url>` 

### Actual behavior

```
-bash-3.2$ singularity build --builder https://cloud.staging.sylabs.io/library builder3_apline-3.6.sif docker://alpine:3.6
INFO:    Starting build...
Skipping fetch of repeat blob sha256:5a3ea8efae5d0abb93d2a04be0a4870087042b8ecab8001f613cdc2a9440616a
Copying config sha256:8186e170383f46da698e9a1dd3e1384e2c86a75c2ebdef62843a23cb8899955b
 585 B / 585 B [============================================================] 0s
Storing signatures
INFO:    Creating SIF file...
INFO:    Build complete: builder3_apline-3.6.sif
```
a local build

```
singularity build --builder https://cloud.staging.sylabs.io/library builder3_apline-3.6.sif docker://alpine:3.6
```

ikaneshiro:

Title

- Fixes #2985 

### DIFF
--- a/cmd/internal/cli/build_darwin.go
+++ b/cmd/internal/cli/build_darwin.go
@@ -14,10 +14,6 @@ import (
 	"github.com/sylabs/singularity/internal/pkg/sylog"
 )
 
-func preRun(cmd *cobra.Command, args []string) {
-	sylabsToken(cmd, args)
-}
-
 func run(cmd *cobra.Command, args []string) {
 	dest := args[0]
 	spec := args[1]
@@ -27,7 +23,7 @@ func run(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
-	if !remote && !cmd.Flags().Lookup("builder").Changed {
+	if !remote {
 		sylog.Fatalf("Only remote builds are supported on this platform")
 	}
 
--- a/cmd/internal/cli/build_darwin.go
+++ b/cmd/internal/cli/build_darwin.go
@@ -27,7 +27,7 @@ func run(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
-	if !remote {
+	if !remote && !cmd.Flags().Lookup("builder").Changed {
 		sylog.Fatalf("Only remote builds are supported on this platform")
 	}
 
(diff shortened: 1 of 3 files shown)


ArangoGutierrez (approved):
Keep it simple, I like it

cmd/internal/cli/build_darwin.go
```diff
@@ -27,7 +27,7 @@ func run(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
-	if !remote {
+	if !remote && !cmd.Flags().Lookup("builder").Changed {
```
mem: I see what you are doing here: either the use set remote explicitly or they specified builder and therefore set remote implicitly.

In order to reduce maintaining costs and avoid having to check the two things continuously, I think it might be easier to check for builder Changed once, and *set* remote if that's the case. Maybe take preRun out of build_linux.go and build_darwin.go (since it's the same in both) and move it to build.go, and perform this check there:

```
func preRun(...) {
    ...
    if cmd.Flags().Lookup("builder").Changed {
        cmd.Flags().Lookup("remote").Value.Set("true")
    }
}
```

in this way, in the rest of the code you can continue checking only remote, as it is now and you don't have to introduce the explicit handling of the dependency everywhere.

It's also a good idea to update the documentation for the builder flag and add something like "(implies --remote)".

cmd/internal/cli/build_linux.go
```diff
@@ -36,7 +36,7 @@ func run(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
-	if remote {
+	if remote || cmd.Flags().Lookup("builder").Changed {
```
mem: I can see that this is the change the issue is asking for: setting builder implies --remote.

mem (changes requested):

mem (approved):
Thanks!

PR #3044, opened 2019-03-21
Adds detection for invalid URI transports

### LINKED ISSUE #2989: Improve error text for `singularity pull` when misspelling transport URI
## Version of Singularity:

3.1.0

### Expected behavior

Write here.

### Actual behavior

<pre>
$ singularity pull docke://alpine:latest
FATAL:   While getting image info: get did not succeed: 400 Bad Request
        Bad image reference Provided collection ref is not valid: docke:
</pre>

The above sticks out to me, but it has about 5 errors in there that are not initially obvious to a new user.

Additionally, somehow I got a `400 Bad Request`. Does that mean it tried to download something from somewhere? But where from? That seems wrong as well

@ikaneshiro I'd fix this with two PRs. One for the error message and one for the `400 bad request`.

ikaneshiro:

Adds DockerProtocol definition, catching and handling of invalid transport types.
Fixes bug in URI package that did not properly check for '//' within a split URI.

- Fixes #2989 

### DIFF
--- a/cmd/internal/cli/pull.go
+++ b/cmd/internal/cli/pull.go
@@ -14,6 +14,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/sylabs/singularity/docs"
 	"github.com/sylabs/singularity/internal/pkg/client/cache"
+	ociclient "github.com/sylabs/singularity/internal/pkg/client/oci"
 	"github.com/sylabs/singularity/internal/pkg/libexec"
 	scs "github.com/sylabs/singularity/internal/pkg/remote"
 	"github.com/sylabs/singularity/internal/pkg/sylog"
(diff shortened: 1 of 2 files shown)


cmd/internal/cli/pull.go
```diff
@@ -173,7 +175,7 @@ func pullRun(cmd *cobra.Command, args []string) {
 		libexec.PullShubImage(name, args[i], force, noHTTPS)
 	case HTTPProtocol, HTTPSProtocol:
 		libexec.PullNetImage(name, args[i], force)
-	default:
+	case DockerProtocol:
```
bauerm97: This should _not_ be passing CI (I think). For instance, `singularity pull oci://[...]` will fail here. The reason we were using  `default` is because anything that's not `library`, `http`/`https`, or `shub`, are all still potentially valid URI's if they are handled by image transport code (this is also what handles `docker://`). If CI is passing, it means we're not testing `singularity pull oci:` or other non-`docker://` URIs

bauerm97 (dismissed):

WestleyK (changes requested):
I would update the copyright year on `internal/pkg/util/uri/uri.go`. Otherwise it looks good :+1:

WestleyK (approved):
:+1:

mem (approved):
Addressed @bauerm97 comments.

PR #3047, opened 2019-03-21
'key remove' Command

WestleyK:

This PR will add a `singularity key remove` command to remove a local public key.

- Fixes # *non that I can think of...*

<br>


pkg/sypgp/sypgp.go
```diff
@@ -285,7 +285,7 @@ func StorePubKey(e *openpgp.Entity) (err error) {
 // key and oldToken match.
 func compareLocalPubKey(e *openpgp.Entity, oldToken string) bool {
```
jchavezlavalle: uppercase this function name and update the reference where needed, should be called `CompareLocalPubKey` to have same format of all functions in sypgp

jchavezlavalle (changes requested):

pkg/sypgp/sypgp.go
ikaneshiro: Is this something that we plan on using outside of this package? Should only uppercase things we want to be part of the exported API of the package.

cmd/internal/cli/key_remove.go
```diff
@@ -0,0 +1,38 @@
...
+	KeyRemoveCmd.Flags().SetInterspersed(false)
+}
+
+// KeyRemoveCmd is `singularity key remove <fingerprint>' command
+var KeyRemoveCmd = &cobra.Command{
+	Args:                  cobra.ExactArgs(1),
+	DisableFlagsInUseLine: true,
+	Run: func(cmd *cobra.Command, args []string) {
+
+		err := sypgp.RemovePubKey(fmt.Sprintf("%s", args[0]))
```
mem: `args[0]` is already a string, no need to pass it thru fmt.Sprintf

cmd/internal/cli/key_remove.go
```diff
@@ -0,0 +1,38 @@
...
+		if err != nil {
+			sylog.Fatalf("Unable to remove public key: %v", err)
```
mem: in this case it's better to use "%s" instead of "%v" because "%v" might, in some cases, dump the structure of the variable (not all errors are strings underneath).

pkg/sypgp/sypgp.go
```diff
@@ -327,14 +327,22 @@ func RemovePubKey(toDelete string) error {
...
 		// if the elist[i] dose not match toDelete, then add it to newKeyList
 		if !compareLocalPubKey(elist[i], toDelete) {
 			newKeyList = append(newKeyList, *elist[i])
+		} else {
+			matchKey = true
 		}
 	}
 
+	if !matchKey {
+		return fmt.Errorf("no key matching given fingerprint")
```
mem: "no key matching given fingerprint found"

mem (dismissed):
looks good, just a couple minor fixes.

pkg/sypgp/sypgp.go
WestleyK: I changed the function name to `CompareKeyEntity()`, hopefully everyone is happy :smile:

pkg/sypgp/sypgp.go
```diff
@@ -327,14 +327,22 @@ func RemovePubKey(toDelete string) error {
...
+	matchKey := false
+
 	// sort through them, and remove any that match toDelete
 	for i := range elist {
 		// if the elist[i] dose not match toDelete, then add it to newKeyList
-		if !compareLocalPubKey(elist[i], toDelete) {
+		if !CompareKeyEntity(elist[i], toDelete) {
 			newKeyList = append(newKeyList, *elist[i])
+		} else {
+			matchKey = true
```
jchavezlavalle, line 338: if `matchKey` is true, shouldn't it stop looping arount the list of entities `elist`? if that's so I would add a `break` after `matchKey = true`

jchavezlavalle (changes requested):

pkg/sypgp/sypgp.go
jchavezlavalle: @ikaneshiro gotcha, well if it's only intended to be used on this package of course it makes no sense to uppercase it :+1:

pkg/sypgp/sypgp.go
WestleyK: No, this is only used to test if theres was a match with the *fingerprint_to_delete*, and the local fingerprint.

jchavezlavalle: Yea but if it matches shouldn't it stop looking for it?

WestleyK: In this case no, because we are building up `newKeyList` and dont want to lose it, this will result in missing public keys if the function should succeed.

ikaneshiro (approved):
lgtm

jchavezlavalle (approved):
LGTM :+1: thanks @WestleyK

PR #3072, opened 2019-03-22
Document blobSum type cache option

### LINKED ISSUE #3012: undocumented option in `singularity cache list` `--type blobSum`
## Version of Singularity:

3.0.1

<pre>
$ singularity cache list --type blobSum

There are 3 oci blob file(s) using 2.76 Mb of space. Use: '-T=blob' to list
</pre>

This option (and any other hidden ones), should be documented.

sashayakovtseva:
Also cache type listing removed from description so we won't need to update it in two places if cache type is added. All cache types are listed in --type option decsription and there is no need to duplicate it in command description as well.

Closes #3012

### DIFF
--- a/docs/content.go
+++ b/docs/content.go
@@ -162,8 +162,7 @@ Enterprise Performance Computing (EPC)`
 	CacheUse   string = `cache <subcommand>`
 	CacheShort string = `Manage the local cache`
 	CacheLong  string = `
-  Manage your local singularity cache. There are 3 types of cache; library, oci, and blob.
-  You can list/clean using the specific types.`
+  Manage your local singularity cache. You can list/clean using the specific types.`
 	CacheExample string = `
   All group commands have their own help output:
 
@@ -176,9 +175,8 @@ Enterprise Performance Computing (EPC)`
 	CacheCleanUse   string = `clean [clean options...]`
 	CacheCleanShort string = `Clean your local Singularity cache`
 	CacheCleanLong  string = `
-  This will clean you local cache: "${HOME}/.singularity/cache". The available cache
-  types are: library, oci, and blob. By default cache clean will only clean blob cache,
-  use: '--all' to clean all cache.`
+  This will clean you local cache stored at $HOME/.singularity/cache.
+  By default only blob cache is cleaned, use '--all' to clean all cache.`
 	CacheCleanExample string = `
   All group commands have their own help output:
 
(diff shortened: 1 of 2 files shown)


docs/content.go
```diff
@@ -192,8 +190,7 @@ Enterprise Performance Computing (EPC)`
 	CacheListUse   string = `list [list options...]`
 	CacheListShort string = `List your local Singularity cache`
 	CacheListLong  string = `
-  This will list you local cache: "${HOME}/.singularity/cache". The available cache
-  types are: library, oci, and blob.`
+  This will list you local cache stored at $HOME/.singularity/cache.`
```
WestleyK: By default the cache directory is `$HOME/.singularity/cache`, but if `SINGULARITY_CACHEDIR` is set, the cache directory will change. Maybe this can be updated like:

```
This will list your local Singularity cache.
```

*Also `you` => `your`.*

sashayakovtseva: Ok, updated

docs/content.go
```diff
@@ -176,9 +175,8 @@ Enterprise Performance Computing (EPC)`
 	CacheCleanUse   string = `clean [clean options...]`
 	CacheCleanShort string = `Clean your local Singularity cache`
 	CacheCleanLong  string = `
-  This will clean you local cache: "${HOME}/.singularity/cache". The available cache
-  types are: library, oci, and blob. By default cache clean will only clean blob cache,
-  use: '--all' to clean all cache.`
+  This will clean your local cache (stored at $HOME/.singularity/cache if SINGULARITY_CACHEDIR is not set).
+  By default only blob cache is cleaned, use '--all' to clean all cache.`
```
WestleyK: Maybe change it a bit more...

```suggestion
  By default only oci blobs cache is cleaned, use '--all' to clean all cache.`
```

mem: "the entire cache" maybe?

```suggestion
  By default only blob cache is cleaned, use '--all' to clean the entire cache.`
```

docs/content.go
WestleyK: Here too:

```suggestion
  This will list your local cache (stored at $HOME/.singularity/cache if SINGULARITY_CACHEDIR is not set).
```

WestleyK (commented):
One more small change, otherwise, looks good! Thanks for updating this :+1:

docs/content.go
sashayakovtseva: We already removed `oci blob` phrase [here](https://github.com/sylabs/singularity/pull/3073) so let's not add it here as well for consistency.

docs/content.go
sashayakovtseva: @WestleyK Are suggesting to write `cache clean` description where `cache list` should be?

WestleyK: Sorry, copy-paste error. Fixed now! :wink:

WestleyK (approved):
Looks good! :+1:

WestleyK:
Maybe run `gofmt -s -w cmd/internal/cli/cache_list_linux.go`, to fix CCI.

PR #3123, opened 2019-03-25
Implement unit tests for plugin helpers

mem:
Refactor to make testing possible.

Add tests for some internal functions:

-  isPluginFile
- getManifest

Fixes: #2965


cclerget:
@mem Checks like descriptor number may be subject to changes (eg: add signature descriptor for plugin signing), maybe we could use `Fname` (for manifest and plugin object) in `internal/app/singularity/plugin_compile_linux.go` to name descriptors instead with generic name like `plugin.manifest` and `plugin.so` instead of using path which seems not useful for plugin install. By naming descriptors we could use package `Init()` `pkg/image/image.go` and `NewSectionReader()` to quickly retrieve descriptors by name with an associated reader without dealing with descriptor indexes and SIF descriptors. Use of named descriptors instead of indexes is more flexible and involves less maintenance for the long term imo

mem:
> @mem Checks like descriptor number may be subject to changes (eg: add signature descriptor for plugin signing), maybe we could use `Fname` (for manifest and plugin object) in `internal/app/singularity/plugin_compile_linux.go` to name descriptors instead with generic name like `plugin.manifest` and `plugin.so` instead of using path which seems not useful for plugin install. By naming descriptors we could use package `Init()` `pkg/image/image.go` and `NewSectionReader()` to quickly retrieve descriptors by name with an associated reader without dealing with descriptor indexes and SIF descriptors. Use of named descriptors instead of indexes is more flexible and involves less maintenance for the long term imo

Thanks for looking at this!

I think I see what you mean:

```
$ siftool info 1 test-plugin.sif
Descr slot#: 0
  Datatype:  FS
  ID:        1
  Used:      true
  Groupid:   1
  Link:      NONE
  Fileoff:   32768
  Filelen:   20187416
  Ctime:     2019-03-25 10:31:01 -0600 CST
  Mtime:     2019-03-25 10:31:01 -0600 CST
  UID:       1001
  Gid:       1001
  Name:      test-plugin.so
  Fstype:    Raw
  Parttype:  Data
  Arch:      amd64
```

The idea is that can I build a `sif.Descriptor` with `Name` set to the correct idenfier for the file (plugin.so, ...) and pass that to FileImage.GetFromDescr to find the correct descriptor?

cclerget:
@mem Is threre any concerns to not use existing package handling image stuff in `pkg/image/image.go` and `pkg/image/reader.go` to avoid code duplication/maintenance ?

WestleyK (approved):
Looks good to me :+1:

GodloveD:
@cclerget looks like you had a question above.  Is this ready to go or does your question need to be addressed?

cclerget:
@GodloveD That would be great to address it in order to centralize usage of SIF helpers in `pkg/image/sif.go` public package (eg: arch support coming), @mem actually the only bug I identified is the lack of filtering on `Used` field (maybe I miss something else) in image package but it's easily fixable

EDIT: @mem requested your review for #3195 that fixes blocker avoiding use of `pkg/image` for this PR

PR #3177, opened 2019-03-31
Support Sandbox Image on macOS

### LINKED ISSUE #2871: MacOS alpha version does not work with sandbox images
## Version of Singularity:

First alpha release of MacOS version of singularity.  singularity --version says 3.1.0-196.g8d47ee66.dirty

### Expected behavior

Works with sandbox images.

### Actual behavior

FATAL:   image format not recognized

Copy a sandbox image from a Linux machine and do singularity shell sandbox_dirname

@jmstover

jmstover:
Check if we have a directory used for the image, and if so be sure to pass it instead of a image.

This will help support running a Sandbox image from a macOS host.

- Fixes #2871 

### DIFF
--- a/cmd/internal/cli/startvm_darwin.go
+++ b/cmd/internal/cli/startvm_darwin.go
@@ -28,7 +28,6 @@ func getHypervisorArgs(sifImage, bzImage, initramfs, singAction, cliExtra string
 		"-m", VMRAM,
 		"-c", VMCPU,
 		"-s", "0:0,hostbridge",
-		"-s", hdString,
 		"-s", "31,lpc",
 		"-l", "com1,stdio",
 	}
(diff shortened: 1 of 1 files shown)


cmd/internal/cli/startvm_darwin.go
```diff
@@ -70,16 +69,31 @@ func getHypervisorArgs(sifImage, bzImage, initramfs, singAction, cliExtra string
...
 	args = append(args, "-s", pciArgs)
 
+	// Check for Sandbox Image
+	sylog.Debugf("Check for sandbox image")
+	if f, err := os.Stat(sifImage); err == nil {
+		if f.IsDir() {
+			sylog.Debugf("Image is sandbox. Setting up share.")
+			pciArgs = fmt.Sprintf("4:1,virtio-9p,runimg=%s", sifImage)
+			args = append(args, "-s", pciArgs)
+			sboxImgBind := fmt.Sprintf("runimg:/runImage")
```
cclerget, line 89: `runImage` or sifImage value ?

jmstover: `runimg` is the 9p mount tag. `/runImage` is the mount point.

jmstover: To expand on this... line 87, it sets:
`virtio-9p,runimg=[sifImage]`

Where `runimg` is the 9p mount tag, and what we're passing to be mounted is the value of `sifImage`, which in this case is a sandbox directory.

cclerget (approved):
LGTM

PR #3178, opened 2019-03-31
Generate 9p Mount Tag on Bind Mount

jmstover:
Generate a random 6 alphabetic char 9p mount tag for all mounts that
are specified with the `-B` option. This means that a user can set
both `-B /tmp -B /var/tmp` and it will work as expected.

Previously `tmp` would have been used as the mount tag for both
of the bind mounts.

### DIFF
--- a/cmd/internal/cli/startvm_darwin.go
+++ b/cmd/internal/cli/startvm_darwin.go
@@ -7,17 +7,33 @@ package cli
 
 import (
 	"fmt"
+	"math/rand"
 	"os"
 	"os/user"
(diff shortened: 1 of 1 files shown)


jmstover:
@bauerm97 @mem ... I've added a `genString()` func here... There should _probably_ be a utility function that can be shared to do this ... but I'm not 100% sure exactly where/how to best implement that.

WestleyK (approved):
Looks good to me :+1:

Although it may be a good idea to change `genString`, to `genRandomString`.

PR #3181, opened 2019-04-01
Update Bind Mount Handling

jmstover:
Modify how bind mounts are getting passed. We were hard limiting the
number to 8, and that is now at 48 from the Mac host (assuming my
math is right... which I don't guarantee)

Start the bind mount slot at 26, and increase the slot number every
time we hit 8 mounts.

Error if we surpass the max slot.


cmd/internal/cli/startvm_darwin.go
```diff
@@ -33,16 +33,19 @@ func getHypervisorArgs(sifImage, bzImage, initramfs, singAction, cliExtra string
 		"-l", "com1,stdio",
 	}
 
-	if len(BindPaths) > 8 {
```
cclerget, line 51: Or replace by 
```
if len(BindPaths) >= 48 {
    sylog.Fatalf("Surpassed max amount of binds we can pass to virtual machine")
}
```

cmd/internal/cli/startvm_darwin.go
```diff
@@ -33,16 +33,19 @@ func getHypervisorArgs(sifImage, bzImage, initramfs, singAction, cliExtra string
...
 	singBinds := []string{""}
 
-	slot := 5
+	// Set slot to 26. slot has a max value of 31, so this will give us a max of 48 bind mounts from the Mac host.
+	slot := 26
+	function := 0
+
+	for _, bindpath := range BindPaths {
+		// If we somehow surpass slot 31 ... error. We can't do anything at this point.
+		if slot > 31 {
```
cclerget: and remove this one

cclerget (approved):
Just suggestion otherwise LGTM

Issue #3132, opened 2019-03-26
Plugin Compile Improvements


tri-adam:
The `plugin compile` usage is inaccurate:

```
The 'plugin compile' command allows a developer to compile a plugin in the
expected environment. The host directory specified is the location of the plugins
source code folder which will be bind mounted into the compile container. Compilation
of a container must happen in a container due to constraints in the Go plugin package.
```

AFAIK, we are not bind mounting into a container to do the compilation any more.

Also, the `plugin compile` command only works on paths _inside_ the Singularity tree. This needs to be documented as well.

The output of `plugin compile` is also confusing:

```
[/home/adam/go/src/github.com/sylabs/singularity/examples/plugins/test-plugin/test-plugin.so /home/adam/go/src/github.com/sylabs/singularity/examples/plugins/test-plugin/test-plugin.manifest]
DEBUG   [U=1000,P=15460]   RegisterStringFlag()          Registering string flag test-flag
DEBUG   [U=1000,P=15460]   RegisterStringFlag()          Successfully registered new string flag test-flag
```

The only things output aside from `DEBUG` statements are the path to a `.so` and a `.manifest`, neither of which are relevant to `plugin install`. I'd suggest clearly indicating where the `.sif` file has been output, and removing the intermediate files (perhaps adding a flag to keep the intermediates, if you believe these would be useful for some reason.

PR #3184, opened 2019-04-02
Switch to using go modules

mem:
- Add go.mod (and the corresponding go.sum)
- Fix makeit/gengodep so that it lists all the go files that are
  needed to build a particular import path
- Add internal/pkg/runtime/engines/singularity/plugins.go to have
  something that pulls the relevant plugin code.
- Remove Gopkg.toml and Gopkg.lock
- Updated vendor/ with go mod vendor
- Make sure we are calling the same go compiler in all cases by
  resolving the absolute path only once.
- Explicitly export GO111MODULE=on to make sure that everything is
  running with module support activated. In particular, this seems
  necessary to build the CNI plugins, as having a go.mod file and
  passing -mod=vendor in that case doesn't seem to be sufficient (go
  still fails to locate the relevant modules).
- Replace dep with go mod

Signed-off-by: Michael Bauer <bauerm@umich.edu>


mlocal/frags/build_network.mk
```diff
@@ -57,12 +58,12 @@ cni_config_INSTALL := $(DESTDIR)$(SYSCONFDIR)/singularity/network
...
 cniplugins:
 	$(V)install -d $(cni_builddir)
-	$(V)for p in $(cni_plugins_GOPATH); do \
+	$(V)for p in $(cni_plugins); do \
 		name=`basename $$p`; \
 		cniplugin=$(cni_builddir)/$$name; \
 		if [ ! -f $$cniplugin ]; then \
 			echo " CNI PLUGIN" $$name; \
-		go build $(GO_BUILDMODE) -tags "$(GO_TAGS)" $(GO_LDFLAGS) $(GO_GCFLAGS) $(GO_ASMFLAGS) \
+		go build $(GO_MODFLAGS) $(GO_BUILDMODE) -tags "$(GO_TAGS)" $(GO_LDFLAGS) $(GO_GCFLAGS) $(GO_ASMFLAGS) \
```
cclerget:
```suggestion
		GOPATH=$(SOURCEDIR)/vendor go build $(GO_MODFLAGS) $(GO_BUILDMODE) -tags "$(GO_TAGS)" $(GO_LDFLAGS) $(GO_GCFLAGS) $(GO_ASMFLAGS) \
```

internal/pkg/runtime/engines/singularity/plugins.go
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) 2019, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+// +build cni_plugins
```
cclerget, line 13: To avoid confusion, maybe add a comment mentioning this is a hack for `go mod` and `cni_plugins` build tag is not intended to be used during build

cclerget (changes requested):
Also tested by removing vendor and executing `go mod vendor`, everything goes back without any issues. Nice work !

mlocal/frags/build_network.mk
mem: Thanks! I'm going to guess you ran into the same issue I did, where sometimes the CNI plugins couldn't be found.

Experimenting around, setting GO111MODULE=on explicitly seems to take care of this in all cases.

internal/pkg/runtime/engines/singularity/plugins.go
mem: Good point, I'll update this.

cclerget (approved):

mem:
This is connected to issue #3132 because we need some way to ensure the cobra version used when building singularity and the version used when building plugins match. Go modules provide that.

Issue #2559, opened 2019-01-17
Packages test coverage


cclerget:
By order of priority :

**Not tested packages :**

- Public packages

github.com/sylabs/singularity/pkg/build/types (#3131)
github.com/sylabs/singularity/pkg/client/net (sylog call)
github.com/sylabs/singularity/pkg/network (#2760)
github.com/sylabs/singularity/pkg/signing
github.com/sylabs/singularity/pkg/util/copy (#2563)
github.com/sylabs/singularity/pkg/util/fs/lock (#2564)
github.com/sylabs/singularity/pkg/util/loop (#2565)
github.com/sylabs/singularity/pkg/util/unix (#2574)

- Private packages

github.com/sylabs/singularity/internal/pkg/build
github.com/sylabs/singularity/internal/pkg/build/apps
github.com/sylabs/singularity/internal/pkg/buildcfg
github.com/sylabs/singularity/internal/pkg/buildcfg/confgen
github.com/sylabs/singularity/internal/pkg/client/oci
github.com/sylabs/singularity/internal/pkg/image
github.com/sylabs/singularity/internal/pkg/instance (#3191)
github.com/sylabs/singularity/internal/pkg/libexec
github.com/sylabs/singularity/internal/pkg/runtime/engines
github.com/sylabs/singularity/internal/pkg/runtime/engines/config/oci
github.com/sylabs/singularity/internal/pkg/runtime/engines/config/starter
github.com/sylabs/singularity/internal/pkg/runtime/engines/imgbuild
github.com/sylabs/singularity/internal/pkg/runtime/engines/oci
github.com/sylabs/singularity/internal/pkg/runtime/engines/oci/rpc
github.com/sylabs/singularity/internal/pkg/runtime/engines/oci/rpc/client
github.com/sylabs/singularity/internal/pkg/runtime/engines/oci/rpc/server
github.com/sylabs/singularity/internal/pkg/runtime/engines/singularity
github.com/sylabs/singularity/internal/pkg/runtime/engines/singularity/rpc
github.com/sylabs/singularity/internal/pkg/runtime/engines/singularity/rpc/client
github.com/sylabs/singularity/internal/pkg/runtime/engines/singularity/rpc/server
github.com/sylabs/singularity/internal/pkg/security (#3196)
github.com/sylabs/singularity/internal/pkg/security/apparmor (#3196)
github.com/sylabs/singularity/internal/pkg/security/selinux (#3196)
[...]

PR #3191, opened 2019-04-02
Add tests for internal/pkg/instance

cclerget:

- #2559 


internal/pkg/instance/instance_linux_test.go
```diff
@@ -0,0 +1,279 @@
...
+	"github.com/sylabs/singularity/internal/pkg/test"
+)
+
+const testSubDir = "testing"
+
+func TestProcName(t *testing.T) {
+	test.DropPrivilege(t)
+	defer test.ResetPrivilege(t)
+
+	test := ProcName("test", "test")
```
mem: here `test` is a new variable; on line 20 `test` is a package selector. To avoid the ambiguity it's probably better to name the variable something else, like `actual` to reflect that this is the actual value obtained from calling the function under test.

Also, without looking at the internals of ProcName, I would expect that there are at least four interesting cases to test:

```
    ProcName("str1", "str2")
    ProcName("str1", "")
    ProcName("", "str1")
    ProcName("", "")
```

- two different strings (to test that the correct one goes in the correct place)
- second string empty
- first string empty
- both strings empty

internal/pkg/instance/instance_linux_test.go
```diff
@@ -0,0 +1,279 @@
...
+		},
+		{
+			input:  "instance:///test",
+			output: "/test",
+		},
+	}
+	for _, e := range tests {
+		o := ExtractName(e.input)
+		if o != e.output {
+			t.Errorf("unexpected match, got %s instead of %s", o, e.output)
```
mem: "unexpect result" maybe?

internal/pkg/instance/instance_linux_test.go
```diff
@@ -0,0 +1,279 @@
...
+			expectFailure: false,
+		},
+		{
+			name:          "test/123",
+			expectFailure: true,
+		},
+	}
+	for _, e := range tests {
+		path, err := GetDirUnprivileged(e.name, testSubDir)
+		if err != nil && !e.expectFailure {
```
mem, line 189: there's also the case of:

    if err == nil && e.expectFailure {
        t.Errorf("...")
    }

the complete set could be expressed as:

    switch {
    case err == nil && e.expectFailure:
        t.Errorf(...)
    case err != nil && !e.expectFailure:
        t.Errorf(...)
    case err != nil && e.expectFailure:
        // everything OK
    case path != e.path:
        t.Errorf(...)
     }
```

File: internal/pkg/instance/instance_linux_test.go [hunk 4]
```diff
@@ -0,0 +1,279 @@
...
+			path:          filepath.Join(instancePath, "test"),
+			expectFailure: false,
+		},
+		{
+			name:          "test/123",
+			expectFailure: true,
+		},
+	}
+	for _, e := range tests {
+		path, err := GetDirPrivileged(e.name, testSubDir)
```
mem, line 220: same here

internal/pkg/instance/instance_linux_test.go
```diff
@@ -0,0 +1,279 @@
...
+
+func TestAdd(t *testing.T) {
+	test.EnsurePrivilege(t)
+
+	for _, e := range instanceTests {
+		var err error
+		var file *File
+
+		file, err = Add(e.name, e.privileged, testSubDir)
+		if err != nil && !e.expectFailure {
```
mem, line 271: something similar here. You can also test for an unexpect file == nil, I guess.

internal/pkg/instance/instance_linux_test.go
```diff
@@ -0,0 +1,279 @@
...
+
+func TestGet(t *testing.T) {
+	test.EnsurePrivilege(t)
+
+	for _, e := range instanceTests {
+		var err error
+		var file *File
+
+		file, err = Get(e.name, testSubDir)
+		if err != nil && !e.expectFailure {
```
mem, line 317: and here, too.

mem (changes requested):

mem (approved):
thanks!

PR #3192, opened 2019-04-02
Wait for dpkg lock on CircleCI

mem:
It seems that when the VM boots, it runs `apt-get update` (or something
similar) and this prevents `apt-get install ...` from working.

Wait until that lock is released. Since the VM is using a somewhat old
version of Ubuntu, other (less racy) methods to handle this are not
available.


WestleyK (approved):
Looks good to me :+1:

ikaneshiro (approved):
lgtm

PR #3195, opened 2019-04-03
Consolidate image package and make SIF image code more generic

cclerget:

This PR consolidate image package and makes SIF image code more generic to handle use case like plugin embedded in SIF image in #3123


pkg/image/image.go
```diff
@@ -29,7 +29,7 @@ const (
 
 const (
 	// RootFs partition name
-	RootFs       = "rootfs"
+	RootFs       = "!__rootfs__!"
```
mem, line 32: Everything looks good.

The one thing I'm not sure about is this change.

Does this have an impact on the images that get created? In particular, what happens with images created before this change? I see there's a new check using HasRootFs... are older images rejected?

Related to this: does the change on line 21 (SQUASHFS) have an impact on images?

pkg/image/image.go
```diff
@@ -134,6 +134,16 @@ func (i *Image) AuthorizedGroup(groups []string) (bool, error) {
 	return authorized, nil
 }
 
+// HasRootfs returns if image contains a root filesystem partition
+func (i *Image) HasRootfs() bool {
```
mem: minor thing: maybe HasRootFs?

pkg/image/image.go
cclerget: No impact, those informations are used by runtime only to somewhat abstract image format

pkg/image/image.go
cclerget: yep will change that. Thanks !

mem (approved):
thanks!

ikaneshiro:
@cclerget seems like you should add a `HasRootFs()` check to this function as well
https://github.com/sylabs/singularity/blob/master/cmd/internal/cli/actions_linux.go#L40

Issue #3176, opened 2019-03-31
Errors when incorrect global flag is used with just one dash


15baraniana:
## Version of Singularity:

3.1-2

### Expected behavior

`sudo singularity -debug build lolcow.sif library://sylabs-jms/testing/lolcow`
Probably should give an error about bad flags

'sudo singularity -star build lolcow.sif library://sylabs-jms/testing/lolcow'
Also should give an error

### Actual behavior

`sudo singularity -debug build lolcow.sif library://sylabs-jms/testing/lolcow`
No error and no actions

`sudo singularity -star build lolcow.sif library://sylabs-jms/testing/lolcow`
Performs the build in silent mode

### Steps to reproduce behavior

Try some misspelled global flags with just 1 dash (`-star,-deb,-de`)

gvallee:
This has been fixed in PR #3230

PR #3230, opened 2019-04-04
CLI command/flag manager

cclerget:

Add a CLI command/flag register system to ease CLI command/flag management


pkg/cmdline/cmd.go
```diff
@@ -0,0 +1,124 @@
...
+
+// NewCommandManager instantiates a CommandManager
+func NewCommandManager(rootCmd *cobra.Command) *CommandManager {
+	if rootCmd == nil {
+		panic("nil root command passed")
+	}
+	cm := &CommandManager{
+		rootCmd:   rootCmd,
+		groupCmds: make(map[string][]*cobra.Command),
+		errPool:   make([]error, 0),
```
mem: you don't need this

pkg/cmdline/cmd.go
```diff
@@ -0,0 +1,124 @@
...
+		fm:        newFlagManager(),
+	}
+	return cm
+}
+
+func (m *CommandManager) pushError(f string, a ...interface{}) {
+	m.errPool = append(m.errPool, fmt.Errorf(f, a...))
```
mem: it would be better if `pushError` took error (`pushError(errs ...error)`) so that the client code can decide what to do with the errors.

If you feel formatting errors like this is necessary, you could have `pushErrorf(format string, args ...interface{})` (in addition to pushError, not instead of)

pkg/cmdline/cmd.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// GetError returns the error pool
```
mem, line 72: what's the intention behind this?

pkg/cmdline/cmd.go
```diff
@@ -0,0 +1,124 @@
...
+func (m *CommandManager) GetError() []error {
+	return m.errPool
+}
+
+// RegisterCmd registers a child command for the root command
+func (m *CommandManager) RegisterCmd(cmd *cobra.Command, interspersed bool) {
+	// panic here because it's a misuse of API and generally from
+	// global context or init() functions
+	if cmd == nil {
+		panic("nil command passed")
```
mem, line 84: please don't panic, please return errors.

pkg/cmdline/cmd.go
```diff
@@ -0,0 +1,124 @@
...
+	rootCmd   *cobra.Command
+	groupCmds map[string][]*cobra.Command
+	errPool   []error
+	fm        *flagManager
+}
+
+// NewCommandManager instantiates a CommandManager
+func NewCommandManager(rootCmd *cobra.Command) *CommandManager {
+	if rootCmd == nil {
+		panic("nil root command passed")
```
mem, line 46: please don't panic, please return errors

pkg/cmdline/cmd.go
```diff
@@ -0,0 +1,124 @@
...
+	} else if childCmd == nil {
+		panic("nil child command passed")
+	}
+	parentCmd.AddCommand(childCmd)
+	childCmd.Flags().SetInterspersed(interspersed)
+}
+
+// SetCmdGroup creates a group of commands identified by name
+func (m *CommandManager) SetCmdGroup(name string, cmds ...*cobra.Command) {
+	m.groupCmds[name] = make([]*cobra.Command, 0)
```
mem: you can do `m.groupCmds[name] = append([]*cobra.Command{}, cmds...)` and you can loop over cmds to check for `nil` if necessary. Or preallocate the slice to the correct size (`make(..., 0, len(cmds))`)

pkg/cmdline/cmd.go
```diff
@@ -0,0 +1,124 @@
...
+	for _, c := range m.rootCmd.Commands() {
+		if c.Name() == name {
+			return c
+		}
+	}
+	return nil
+}
+
+// GetSubCmd return the named command associated with parent command given as argument
+func (m *CommandManager) GetSubCmd(parentCmd *cobra.Command, name string) *cobra.Command {
```
mem: you can leave `m` out since you are not using it.

And that makes me question why you need to attach `GetSubCmd` to `CommandManager` (same with the add variant, but here it's more obvious)

pkg/cmdline/cmd.go
```diff
@@ -0,0 +1,124 @@
...
+	for _, c := range parentCmd.Commands() {
+		if c.Name() == name {
+			return c
+		}
+	}
+	return nil
+}
+
+// RegisterCmdFlag registers a flag for a command
+func (m *CommandManager) RegisterCmdFlag(flag *Flag, cmds ...*cobra.Command) {
```
mem: I was wondering why you had cmd as the second argument...

Since the common case is to register a flag for a single command, wouldn't it be more natural to have `RegisterCmdFlag(cmd *cobra.Command, flag *Flag)`?

pkg/cmdline/cmd.go
```diff
@@ -0,0 +1,124 @@
...
+	if err := m.fm.registerCmdFlag(flag, cmds...); err != nil {
+		m.pushError(err.Error())
```
mem: this is what I meant before: err.Error() discards all the information associated with err and only leaves the string representation.

pkg/cmdline/cmd_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+	if err := registerCmd(cm, nil); err == nil {
+		t.Errorf("unexpected success with nil command")
+	}
+
+	if err := registerCmd(cm, childCmd); err != nil {
+		t.Errorf("unexpected error while registering command: %s", err)
+	}
+
+	if cm.GetCmd("nochild") != nil {
```
mem: where do these strings come from?

pkg/cmdline/env.go
```diff
@@ -0,0 +1,61 @@
...
+	return nil
+}
+
+// EnvBool sets a bool flag if the CLI option is unset and env var is set
+func EnvBool(flag *pflag.Flag, envvar string) error {
+	if flag.Changed || envvar == "" {
+		return nil
+	}
+
+	if err := flag.Value.Set(envvar); err != nil {
```
mem: the name "envvar" suggests this is the name of the environment variable, but the use suggests this is the value.

mem: the reason why this can return an error is because the value `envvar` is not a boolean (1, t, T, true, TRUE, True, 0, f, F, false, FALSE, False). If the value the user is placing in the environment variable is not a boolean, I'm not sure setting the flag to true is a good idea, as this might lead to surprising behavior (e.g. "VAR=no ...")

pkg/cmdline/env.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	flag.Changed = true
+	sylog.Debugf("Updated flag '%s' value to: %s", flag.Name, flag.Value)
+	return nil
+}
+
+// EnvStringNSlice writes to a string or slice flag if CLI option/argument
+// string is unset and env var is set
+func EnvStringNSlice(flag *pflag.Flag, envvar string) error {
```
mem: I would stick to `EnvStringSlice` (I can't really figure out what "N" represents here)

pkg/cmdline/env_test.go
```diff
@@ -70,11 +70,11 @@ func TestEnvStringNSlice(t *testing.T) {
 	v = cmd.Flag("stringSlice").Value.String()
 	assert.Equal(t, v, "[]", "The flag should be unset.")
 
-	envStringNSlice(cmd.Flag("stringSlice"), "sliceval")
+	EnvStringNSlice(cmd.Flag("stringSlice"), "sliceval")
 	v = cmd.Flag("stringSlice").Value.String()
 	assert.Equal(t, v, "[sliceval]", "The flag should be set to value provided.")
 
-	envStringNSlice(cmd.Flag("stringSlice"), "newsliceval")
+	EnvStringNSlice(cmd.Flag("stringSlice"), "newsliceval")
```
mem: interesting... there should be tests that test the slice part of "EnvStringSlice", all these tests are looking at a single string.

mem (changes requested):

pkg/cmdline/cmd.go
cclerget: The intention is to check for error once CLI has been created and before executing cobra main command in https://github.com/sylabs/singularity/pull/3230/files#diff-8ad62a32a8a9f39c5f800c0dbb487d60R218

pkg/cmdline/cmd.go
cclerget: Right !

pkg/cmdline/cmd.go
cclerget: It's really intentional and decided to add `panic` because there is no interaction with user input here, those conditions are not supposed to be reached when used correctly, so if developer does something wrong, code just panic with an explicit message instead of simply ignoring this check like most lib

pkg/cmdline/cmd.go
cclerget: Well both `GetCmd` and `GetSubCmd` are not used actually, but this refactor is intended to also provide a convenient way to deal with CLI from plugin by just passing `CmdManager` instance to plugin interface. Yes it doesn't necessary to be a member of `CommandManager`

pkg/cmdline/cmd.go
cclerget: Because cmds is a variadic function

pkg/cmdline/cmd_test.go
cclerget: This is the check for non-existent command. Will add comment

pkg/cmdline/env.go
cclerget: I didn't originally write this code, just copied `env.go` as is and did few changes. Will address your comments. Thanks !

pkg/cmdline/env.go
cclerget: I guess it's for `And`

pkg/cmdline/env_test.go
cclerget: Will rewrite the whole tests for env

pkg/cmdline/cmd.go
cclerget: Looks like not necessary. Will change

gvallee:
I ran a test and I confirm that it fixes #3176

cmd/internal/cli/singularity.go
```diff
@@ -141,15 +208,27 @@ var SingularityCmd = &cobra.Command{
 // flags appropriately. This is called by main.main(). It only needs to happen
 // once to the root command (singularity).
 func ExecuteSingularity() {
+	setSylogMessageLevel()
+	setSylogColor()
+
+	cmdManager.UpdateCmdFlagFromEnv(envPrefix)
+
+	// error reported by command manager is considered as fatal
+	for _, e := range cmdManager.GetError() {
```
ikaneshiro, line 215: This loop will only print the first error, is that intentional?

ikaneshiro (commented):
Looking good to me, one question

cmd/internal/cli/singularity.go
cclerget: Will fix that. Thanks !

ikaneshiro (approved):

cmd/internal/cli/singularity.go
```diff
@@ -141,15 +208,31 @@ var SingularityCmd = &cobra.Command{
...
+	setSylogColor()
+
+	cmdManager.UpdateCmdFlagFromEnv(envPrefix)
+
+	cliErrors := 0
+	for _, e := range cmdManager.GetError() {
+		sylog.Errorf("%s", e)
+	}
+	// any error reported by command manager is considered as fatal
+	if cliErrors > 0 {
```
mem, line 220: why should cliErrors change between line 216 and here?

cmd/internal/cli/singularity.go
```diff
@@ -123,9 +191,8 @@ func setSylogColor(cmd *cobra.Command, args []string) {
 var SingularityCmd = &cobra.Command{
 	TraverseChildren:      true,
 	DisableFlagsInUseLine: true,
-	PersistentPreRun:      persistentPreRun,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return errors.New("Invalid command")
+		return errors.New("")
```
mem: I'm guessing this empty error message has to do with the code below...

cmd/internal/cli/singularity.go
```diff
@@ -141,15 +208,31 @@ var SingularityCmd = &cobra.Command{
...
-		if str := err.Error(); strings.Contains(str, "unknown flag: ") {
-			flag := strings.TrimPrefix(str, "unknown flag: ")
-			SingularityCmd.Printf("Invalid flag %q for command %q.\n\nOptions:\n\n%s\n",
-				flag,
-				cmd.Name(),
-				cmd.Flags().FlagUsagesWrapped(getColumns()))
-		} else {
+		name := cmd.Name()
+		if _, ok := err.(cmdline.FlagError); !ok {
+			if err.Error() != "" {
```
mem: ... this has to do with the code above, I guess?

The flow of the code is a little confusing. When I first read this, I was expecting to see "ok", so I thought this branch was entered when the type-assertion succeeded. After reading more carefully, I realize this branch is hit when the type-assertion _fails_ (and the asserted value is discarded).

Even after figuring that bit out, I'm still confused as to what's going on here.

The original code (which is ugly, I wrote that, I'm allowed to say it) was trying to deal with cobra's handling of unknown flags.

You figured how to clean that up: yay! (I completely missed FlagErrorFunc).

What I'm thinking is that instead of returning errors.New("") above, you can return either a sentinel value (package-level variable that can be compared against, as in `if err == foo.ErrSomeProblem`) or a specific type, so that you can type assert here:

```
switch err := err.(type) {
case cmdline.FlagError:
    // lines 233-235 to report invalid flags
case somethingElse:
    // whatever extra handling is needed, this should cover this special case
default:
   // generic error that we don't know how to handle
   SingularityCmd.Printf("Error for command %q: %s\n\n", name, err)
   SingularityCmd.Println(cmd.UsageString())
}
```

pkg/cmdline/cmd.go
mem: Not sure about this. In the specific case you describe, without this you would still get a panic from the next line (cmd.SetFlagErrorFunc) because that function tries to access a field in cmd (in Go it's valid to call a method with a nil receiver, so calling the function does not panic). From that point of view, the only thing this is doing is changing the message that gets displayed. You will still have to walk up the stack to figure out why cmd is nil.

I guess I'm saying this specific use is OK.

pkg/cmdline/cmd.go
```diff
@@ -0,0 +1,167 @@
...
+// RegisterSubCmd registers a child command for parent command given as argument
+func (m *CommandManager) RegisterSubCmd(parentCmd, childCmd *cobra.Command, interspersed bool) {
+	// panic here because it's a misuse of API and generally from
+	// global context or init() functions
+	if parentCmd == nil {
+		panic("nil parent command passed")
+	} else if childCmd == nil {
+		panic("nil child command passed")
+	} else if !m.isRegistered(parentCmd) {
+		panic("parent command not registered")
```
mem, line 104: this one I'm not so sure about. childCmd == nil would cause a panic around line 95 or 96 if not in line 94.

What exactly happens if parentCmd is not registered?

Also, since I'm looking at this code: is there _anywhere_ where we set Interspersed to a different value? I get the impression that we want to set it to something specific (true, maybe?) everywhere, so why bother making that an argument?

pkg/cmdline/cmd.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	parentCmd.AddCommand(childCmd)
+	childCmd.Flags().SetInterspersed(interspersed)
+	m.SetCmdGroup(m.GetCmdName(childCmd), childCmd)
```
mem, line 108: Looking at this (and only this), shouldn't this be `m.GetCmdName(parentCmd)`?

I'm probably misunderstanding what this does

pkg/cmdline/cmd.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// SetCmdGroup creates a group of commands identified by name
+func (m *CommandManager) SetCmdGroup(name string, cmds ...*cobra.Command) {
+	if m.groupCmds[name] != nil {
+		panic(fmt.Sprintf("group %s already exists", name))
```
mem, line 116: please do not panic.

In this particular instance, this is strange. I would expect this method to be called multiple times with the same value in name (it's a command group, you are trying to add multiple things to a single group, and I would expect "name" to be the name of that group).

Either documentation needs extending, or this check needs to be removed.

pkg/cmdline/cmd.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	tmp := make([]*cobra.Command, 0, len(cmds))
+	for _, c := range cmds {
+		if c != nil {
+			tmp = append(tmp, c)
+		}
+	}
+	if len(tmp) == 0 {
+		panic("creation of an empty group")
```
mem: Please do not panic.

I would say the proper way to handle this is to check len(cmds) at the top and return if it's zero.

pkg/cmdline/cmd.go
```diff
@@ -0,0 +1,167 @@
...
+
+// GetCmdGroup returns group of commands corresponding to name
+func (m *CommandManager) GetCmdGroup(name string) []*cobra.Command {
+	return m.groupCmds[name]
+}
+
+// GetCmd returns the named command associated with root command
+func (m *CommandManager) GetCmd(name string) *cobra.Command {
+	cmds := m.groupCmds[name]
+	if cmds == nil || len(cmds) > 1 {
```
mem, line 148: `len(cmds) == 0`?

pkg/cmdline/cmd.go
```diff
@@ -0,0 +1,167 @@
...
+		return nil
+	}
+	return cmds[0]
```
mem, line 151: wait... the first command added to the group is associated with the root command? And the others?

pkg/cmdline/cmd.go
```diff
@@ -0,0 +1,167 @@
...
+
+// GetCmdName returns name associated with the provided command.
+// If command is named child and has two parents named parent1 and parent2,
+// this function will return "parent1_parent2_child".
+// Passing the root command to this function returns an empty string.
+func (m *CommandManager) GetCmdName(cmd *cobra.Command) string {
+	var names []string
+
+	c := cmd
+	for c != nil {
```
mem: I would prefer

```
for c := cmd; c != nil; c = c.Parent() {
```

that should make it more evident what this is doing.

pkg/cmdline/cmd.go
```diff
@@ -0,0 +1,167 @@
...
+		if c == m.rootCmd {
+			break
+		}
+		names = append([]string{c.Name()}, names...)
```
mem: in the common case this won't matter, but this is a lot of work in terms of allocations.

    names = append(names, c.Name())

and then reverse the slice:

    for i, j := 0, len(names)-1; i < j; i, j = i+1, j-1 {
        names[i], names[j] = names[j], names[i]
    }

pkg/cmdline/cmd.go
mem: Yeah, that I can see.

I meant since the common case is to have a flag added to a single command, reverse the arguments and take a single command. For the non-common case, you can still use:

    RegisterCmdFlag(cmd1, flag)
    RegisterCmdFlag(cmd2, flag)
    // ...

why would that be a bad thing?

pkg/cmdline/cmd_test.go
```diff
@@ -0,0 +1,155 @@
...
+import (
+	"fmt"
+	"reflect"
+	"testing"
+
+	"github.com/spf13/cobra"
+	"github.com/sylabs/singularity/internal/pkg/test"
+)
+
+var rootCmd = &cobra.Command{Use: "root"}
```
mem, line 17: why do these need to be globals?

pkg/cmdline/env.go
```diff
@@ -0,0 +1,52 @@
...
+	return setValue(flag, value)
+}
+
+// EnvSetValue set flag value if CLI option/argument is unset and env var is set
+func EnvSetValue(flag *pflag.Flag, value string) error {
+	v := strings.TrimSpace(value)
+	if flag.Changed || v == "" {
+		return nil
+	}
+	if flag.Value.Type() == "stringSlice" {
```
mem, line 43: I'm probably missing something here (please add a comment) ...

isn't this what GetStringSlice does?

pkg/cmdline/env_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+var cmd cobra.Command
+
+func TestEnvAppendValue(t *testing.T) {
+	test.DropPrivilege(t)
+	defer test.ResetPrivilege(t)
+
+	cmd.Flags().BoolSlice("boolFlag", []bool{}, "")
+	if cmd.Flag("boolFlag").Value.String() != "[]" {
+		t.Errorf("The flag should be unset.")
```
mem: "empty" instead of "unset"?

pkg/cmdline/env_test.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	EnvAppendValue(cmd.Flag("boolFlag"), "1")
+	if cmd.Flag("boolFlag").Value.String() != "[true]" {
+		t.Errorf("The flag should be set to the value provided.")
+	}
+
+	EnvAppendValue(cmd.Flag("boolFlag"), "false")
+	if cmd.Flag("boolFlag").Value.String() != "[true,false]" {
+		t.Errorf("The flag should appended with the value provided.")
```
mem: "should be"

pkg/cmdline/env_test.go
```diff
@@ -0,0 +1,75 @@
...
+		t.Errorf("The flag should be set to value provided.")
+	}
+
+	cmd.Flags().StringSlice("stringSlice", []string{""}, "")
+	if cmd.Flag("stringSlice").Value.String() != "[]" {
+		t.Errorf("The flag should be unset.")
+	}
+
+	EnvSetValue(cmd.Flag("stringSlice"), "sliceval1,sliceval2")
+	if cmd.Flag("stringSlice").Value.String() != "[sliceval1,sliceval2]" {
```
mem, line 67: this one is tricky: how do you differentiate the value being "sliceval1,sliceval2" from two values joined by a comma because of the call to String()?

Shouldn't you inspect Value directly?

pkg/cmdline/env_test.go
```diff
@@ -0,0 +1,75 @@
...
+		t.Errorf("The flag should be set to value provided.")
+	}
+
+	EnvSetValue(cmd.Flag("stringSlice"), "newsliceval")
+	if cmd.Flag("stringSlice").Value.String() != "[sliceval1,sliceval2]" {
+		t.Errorf("Once set, the flag should not be appended or overwritten.")
```
mem: I'm confused. Calling EnvSetValue multiple times should not have an effect? Or what does the error mean?

mem (changes requested):

pkg/cmdline/cmd.go
cclerget: Look at `cmd/internal/cli/action_flags.go` and you will see it's useful. I can rename it to `RegisterFlagForCmd(flag, cmds...)` to be more natural, but I really want to keep variadic command argument

pkg/cmdline/cmd.go
cclerget: Those panic are not supposed to happen, if it occurs that means CLI manager wasn't correctly used and triggered by developer mistake

cmd/internal/cli/singularity.go
cclerget: Good catch

cmd/internal/cli/singularity.go
cclerget: Will do that

pkg/cmdline/cmd.go
cclerget: Those checks are here to catch potential dev mistakes, nothing else.

Yes will remove interspersed argument, it could be set outside of CLI manager if needed

pkg/cmdline/cmd.go
cclerget: The group is computed from child command, by example for a registered sub-command `list` of `instance` command it will create group with ID `instance_list`.
It could provide an easy way to retrieve command or a group of command by group ID from plugin, by example calling `manager.GetCmd("shell")` would return `shellCmd` instance

pkg/cmdline/cmd.go
cclerget: This function create a group in one shot manner and is not intended to be used twice for the same group, if shell command was already registered (a group "shell" was automatically created) and if someone want to create shell group it will panic.
Will move panic to a pushed error and add documentation

pkg/cmdline/cmd.go
cclerget: Ok I will manage that differently and use cobra function with root command instead

pkg/cmdline/cmd_test.go
cclerget: because it's also used in `flag_test.go`

pkg/cmdline/env.go
cclerget: will look

pkg/cmdline/env_test.go
cclerget: Yes, first call will set flag value and second call should not override current flag value. Will fix error message

pkg/cmdline/env_test.go
cclerget: No need, cobra will return a string flag value as `sliceval1,sliceval2` in this case without brackets

pkg/cmdline/cmd.go
cclerget: left a comment about why this is not done before

pkg/cmdline/cmd.go
cclerget: Sorry but I revert to panic because I think this is the thing to do in CLI context and those panic can only be triggered by dev mistake

pkg/cmdline/env.go
cclerget: Added comment, type detection purpose

pkg/cmdline/cmd.go
cclerget: I left a better function description

mem (approved):
Let's go with this version

Flamefire:
This introduced a serious bug: Uninitialized use of the variable `PullLibraryURI` due to a C&P error at https://github.com/sylabs/singularity/pull/3230/files#diff-1b3e82c021788b1761d186c7b249d2bdR62

PR #3242, opened 2019-04-05
Do not pass -mod=vendor when building singularity

mem:
This interferes with plugin loading. See:

- https://github.com/golang/go/issues/31278
- https://github.com/golang/go/issues/30531
- https://github.com/golang/go/issues/28983


mem:
No need for this one now, #3258 fixes the issue

Issue #1488, opened 2018-04-27
Containers with different architecture from host produce confusing error


markeganfuller:
## Version of Singularity:

2.4.6-HEAD.g9ca70f5

### Expected behavior

Sensible error message like:

```
ERROR : Container architecture is different from host.
```

### Actual behavior

```
$ singularity shell test.img 
Singularity: Invoking an interactive shell within container...

ERROR  : Failed to execv() /.singularity.d/actions/shell, continuing to /bin/sh: Exec format error
ERROR  : Failed to execv() /bin/sh: Exec format error
ABORT  : Retval = 255
```

### Steps to reproduce behavior

* Build container on one arch, e.g. `x86_64`
* Attempt to use container on another arch e.g. `ppc64le`
* Error occurs

denisovlev:
Same issue, right? https://github.com/singularityware/singularity/issues/1836

What is a possible workaround for that? Only building on the same architecture?

If Singularity image is built from a docker container, does it matter what architecture this docker container was built in?

jscook2345:
This is fixed in a newer version of singularity. Please let me know if you have any other comments or questions!

Vishal-Joshi:
I am facing this issue where I build a *docker* image on my Mac M1 chip, and try to use singularity to run it on HPC nodes.

```
vjoshi@<some slurm node>: singularity run --bind <file_system_to_mount>:/volume docker://<image_path>:<image_label>
INFO:    Converting OCI blobs to SIF format
INFO:    Starting build...
Getting image source signatures
Copying blob 7734efb8b826 done
Copying blob bdd1baa98274 done
Copying blob 305fd4841944 done
Copying blob 81e700ced0b7 done
Copying blob 9d4a9f6bf611 done
Copying blob cd0a2bf031cd done
Copying config bc4b9c1d07 done
Writing manifest to image destination
Storing signatures
2023/12/21 12:17:39  info unpack layer: sha256:7734efb8b826f955c6a3bb51d7c84cb262d0e600ea3980f5bce69fa7f4f92d24
2023/12/21 12:17:41  info unpack layer: sha256:bdd1baa98274beac658c72b432fb0b893afa24a12f9806133933597419a299c6
2023/12/21 12:17:42  info unpack layer: sha256:305fd4841944b872966b16ccbf4bd161e200ffe17eb7c610a02aa02c025f4012
[...]

PR #3254, opened 2019-04-08
Perform basic architecture check in the starter

gvallee:

Perform basic architecture check in the starter.

- Addresses #1488. 


gvallee:
If I understand all the code correctly, note that it does not resolve the entire problem since when building/pulling an image, the architecture of the SIF file is based on the architecture of the Go runtime. I believe (to be confirmed) that we want to support cases where the creation of a local SIF is based only on getting file, no architecture-dependent actions, on an architecture that is not the architecture of the image. For example, we may want to let users create a SIF targeting the PPC64LE architecture on a amd64 system when we know that no action that require the target architecture is required during the creation of the SIF.
At the moment, the SIF architecture will be amd64 (the Go runtime architecture when creating the image, not PPC64LE, even if we are getting a PPC64LE image. Then, it is impossible to execute the image on PPC64LE since the architecture is amd64.

All this needs to be confirmed by someone who has a better understanding of the big picture. We still want to merge this PR meanwhile, even if it is more restrictive than we intent, it is still better that the current situation where we are not checking the architecture at any point.

cclerget:
Yes, the pull step may not require architecture check, to allow to pull a ppc64 image, sign it and upload/run it on dedicated cluster.

@gvallee Typically for pull/build process :

1. pull image from docker/library/shub (pull/build command)
2. extract image filesystem to temporary folder (build command)
3. if definition file contains "executable" sections, the architecture check could occur at this step, but it could also be delegated to runtime which would fail during execution. (build command)

pkg/image/sif.go
```diff
@@ -38,6 +40,12 @@ func (f *sifFormat) initializer(img *Image, fileinfo os.FileInfo) error {
 		return err
 	}
 
+	// Check the compatibility of the image's target architecture
+	sifArch := string(fimg.Header.Arch[:sif.HdrArchLen-1])
+	if sifArch != sif.HdrArchUnknown && sifArch != sif.GetSIFArch(runtime.GOARCH) {
```
cclerget, line 51: like you said it's good for now. Comment for reminder, architecture like i386 can run on amd64, maybe some arm processor can run <= armv6 instructions on aarch64, however not 100% sure.

cclerget (approved):

pkg/image/sif.go
gvallee: Good idea and done. I tried to capture the entire situation and hopefully will work on better support during the next few weeks. Meanwhile I believe this PR is ready to go.

PR #3179, opened 2019-04-01
Remove early debug output in internal/pkg/plugin

mem:
Since the init functions for internal/pkg/plugin are running before
those for sylog, the default log level is still at "debug" and these
messages show up whenever a plugin is enabled (irrespective of the
actual log level set on the command line).

Since these are not critical, remove them. If they are deemed to be
useful later, they can be restored after the underlying issue is fixed.

Fixes: #3133

Signed-off-by: Marcelo E. Magallon <marcelo@sylabs.io>

PR #3258, opened 2019-04-09
Fix building plugins in combination with Go modules

### LINKED ISSUE #3132: Plugin Compile Improvements

mem:
There are a number of small fixes required to get plugins building and working again after the migration to Go modules in the singularity repository. Each one is described in the individual commits in this PR.

- Fixes: #3132, #3184, #3179, #2889


vsoch:
This is a cool post, wanted to share :) https://smalldata.tech/blog/2019/03/16/a-quick-introduction-to-golang-modules

tri-adam (approved):
LGTM. Looking forward, think we should consider moving away from checked in `vendor/`, as this approach seems to be falling out of favour in Go 1.13+ ([ref](https://blog.golang.org/modules2019#TOC_6.)).

bauerm97 (approved):
So, this works. If the source code of the plugin is _within_ the singularity repo this will compile and work just fine. But we can take it a step further. `singularity plugin compile` command _could_ add the plugin to the `go.mod` file within the singularity repo, and add a replace statement such as `replace $(PLUGIN_NAME) => $(PATH_TO_PLUGIN_SOURCE)`, at which point running `go mod vendor` will bring the plugin source code into `vendor/` (temporarily).  This is maybe something that doesn't need to be addressed immediately, but down the line (perhaps 3.2.1?). If we decide to pass on that (since I know it'll be a lot extra work to throw it together, that's probably best), then I think this is good to merge as is.

@tri-adam @mem

mem:
> LGTM. Looking forward, think we should consider moving away from checked in `vendor/`, as this approach seems to be falling out of favour in Go 1.13+ ([ref](https://blog.golang.org/modules2019#TOC_6.)).

The bit was mentioned as the reason for wanting to vendor modules is distribution packages (and perhaps closer to our users, heavily filtered internet access; in theory this can be solved with a Go proxy hosted locally, but right now that's a very theoretical possibility).

Packaging Go programs in general is difficult because Go favors static binaries, which means that the best you can hope for is distributing each package as a static library and have (distribution) packages have build dependencies on those packaged Go packages. This still leaves the problem of fixing a security issue in one package and having to rebuild all the dependencies. You could force the use of shared libraries, but without proper versining, shared libraries are a nightmare. Since Go modules encourage the use of semver, that part of the problem could be considered solved (in theory). Now you need upstreams to actually follow semver rules.

Which brings us to singularity and plugins again. Since the tag we are using is v3.x.y, Go considers that a "post v2 package", and those packages are expected to have a slightly different source code structure than what we have right now. Not sure how amicable we are towards moving files around to satisfy this. It would make the life of people writing plugins easier.

ArangoGutierrez (approved):

WestleyK (approved):
:+1:

PR #3267, opened 2019-04-09
Add the vendor-check target to our code checking steps in the circleci configuration

jscook2345:
Verify that the vendor libraries have not been modified outside of the `go mod` workflow (ie, manually edited). This is already done in travis.

Attn: @jmstover @singularity-maintainers

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ jobs:
       - run:
           name: Check code
           command: |-
-            make -j$(nproc) -C ./builddir check
+            make -j$(nproc) -C ./builddir vendor-check check
 
   go1.11-alpine:
     <<: *defaults
@@ -47,7 +47,7 @@ jobs:
       - run:
           name: Check code
           command: |-
-            make -j$(nproc) -C ./builddir check
+            make -j$(nproc) -C ./builddir vendor-check check
 
   go1.11-macos:
     macos:
@@ -71,7 +71,7 @@ jobs:
       - run:
           name: Check code
           command: |-
-            make -j$(sysctl -n hw.physicalcpu) -C ./builddir check
+            make -j$(sysctl -n hw.physicalcpu) -C ./builddir vendor-check check
 
   unit_tests:
     machine: true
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -74,7 +74,7 @@ jobs:
           name: Check code
           command: |-
             curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | TMPDIR=$(mktemp -d) sh -s -- -b $(go env GOPATH)/bin v1.15.0
-            make -j$(sysctl -n hw.physicalcpu) -C ./builddir check
+            make -j$(sysctl -n hw.physicalcpu) -C ./builddir vendor-check check
```
jmstover: Why is this `make` call using `hw.physicalcpu` instead of `nproc` ? Is there a reason for it?

jmstover (changes requested):

.circleci/config.yml
jscook2345: @jmstover `nproc` does not exist on the mac platform

mem (approved):

.circleci/config.yml
jscook2345: @jmstover @mem I see that `sysctl` has a `hw.logicalcpu` name. That may be more reasonable to use since it includes logical core count from hyperthreading. What do you think?

mem: that would match nproc, which gives you the CPU count from the point of view of the OS.

`getconf _NPROCESSORS_ONLN` gives you the number of online processors, which should be equal or larger than the number reported by nproc. That should work on OSX, too.

- hw.physicalcpu :-1:
- hw.logicalcpu :+1: 
- getconf _NPROCESSORS_ONLN :+1:

jmstover (approved):

GodloveD:
@jscook2345 if you can resolve the conflict I'm happy to merge.

jscook2345:
Hold off merging for a hot minute while I discuss some issues with @mem and @tri-adam

PR #3165, opened 2019-03-28
[WIP] Revisit sypgp to enable better testing coverage

gvallee:
**Description of the Pull Request (PR):**

Improve testing coverage for sypgp package. Addresses issue #2559

**This fixes or addresses the following GitHub issues:**

- Addresses #2559 

**Before submitting a PR, make sure you have done the following:**

- Read the [Guidelines for Contributing](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/sylabs/singularity/blob/master/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- Added tests to validate this PR and tested this PR locally with a `make testall`
- Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](https://github.com/sylabs/singularity/blob/master/CONTRIBUTORS.md)

Attn: @singularity-maintainers

PR #3290, opened 2019-04-12
Increase test coverage in pkg/sypgp

mem:
- Break up PathsCheck so that the individual pieces can be tested
- Add ensureDirPrivate and ensureFilePrivate which are testable using
  temporary directories
- Add tests for ensureDirPrivate and ensureFilePrivate
- Add PrintEntity test

### DIFF
--- a/pkg/sypgp/sypgp.go
+++ b/pkg/sypgp/sypgp.go
@@ -414,9 +414,9 @@ func storePubKeyring(keys openpgp.EntityList) error {
 	return nil
 }
 
-// CompareKeyEntity compares a key ID with a string, returning true if the
+// compareKeyEntity compares a key ID with a string, returning true if the
 // key and oldToken match.
-func CompareKeyEntity(e *openpgp.Entity, oldToken string) bool {
+func compareKeyEntity(e *openpgp.Entity, oldToken string) bool {
 	// TODO: there must be a better way to do this...
 	return fmt.Sprintf("%X", e.PrimaryKey.Fingerprint) == oldToken
 }
@@ -435,7 +435,7 @@ func CheckLocalPubKey(ckey string) (bool, error) {
 	}
 
 	for i := range elist {
-		if CompareKeyEntity(elist[i], ckey) {
+		if compareKeyEntity(elist[i], ckey) {
 			return true, nil
 		}
 	}
@@ -461,7 +461,7 @@ func RemovePubKey(toDelete string) error {
 	// sort through them, and remove any that match toDelete
 	for i := range elist {
 		// if the elist[i] dose not match toDelete, then add it to newKeyList
-		if !CompareKeyEntity(elist[i], toDelete) {
+		if !compareKeyEntity(elist[i], toDelete) {
 			newKeyList = append(newKeyList, elist[i])
 		} else {
 			matchKey = true
--- a/pkg/sypgp/sypgp.go
+++ b/pkg/sypgp/sypgp.go
@@ -832,20 +832,17 @@ func serializePrivateEntity(e *openpgp.Entity, blockType string) (string, error)
 
 // RecryptKey Will decrypt a entity, then recrypt it with the same password.
 // This function seems pritty usless, but its not!
-func RecryptKey(k *openpgp.Entity) error {
-	if k.PrivateKey.Encrypted {
-		pass, err := AskQuestionNoEcho("Enter key passphrase : ")
-		if err != nil {
-			return err
-		}
-		err = k.PrivateKey.Decrypt([]byte(pass))
-		if err != nil {
-			return err
-		}
-		err = k.PrivateKey.Encrypt([]byte(pass))
-		if err != nil {
-			return err
-		}
+func RecryptKey(k *openpgp.Entity, passphrase []byte) error {
+	if !k.PrivateKey.Encrypted {
+		return errNotEncrypted
+	}
+
+	if err := k.PrivateKey.Decrypt(passphrase); err != nil {
+		return err
+	}
+
+	if err := k.PrivateKey.Encrypt(passphrase); err != nil {
+		return err
 	}
 
 	return nil
@@ -864,7 +861,12 @@ func ExportPrivateKey(kpath string, armor bool) error {
 		return err
 	}
 
-	err = RecryptKey(entityToExport)
+	pass, err := AskQuestionNoEcho("Enter key passphrase : ")
+	if err != nil {
+		return err
+	}
+
+	err = RecryptKey(entityToExport, []byte(pass))
 	if err != nil {
 		return err
 	}
--- a/pkg/sypgp/sypgp.go
+++ b/pkg/sypgp/sypgp.go
@@ -477,6 +477,24 @@ func CheckLocalPubKey(ckey string) (bool, error) {
 	return findKeyByFingerprint(elist, ckey) != nil, nil
 }
 
+// removeKey removes one key identified by fingerprint from list.
+//
+// removeKey returns a new list with the key removed, or nil if the key
+// was not found. The elements of the new list are the _same_ pointers
+// found in the original list.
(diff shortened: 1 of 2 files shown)


pkg/sypgp/sypgp.go
```diff
@@ -98,61 +105,67 @@ func PublicPath() string {
...
 }
 
-// PathsCheck creates the sypgp home folder, secret and public keyring files
-func PathsCheck() error {
-	// create the sypgp base directory
-	if err := os.MkdirAll(DirPath(), 0700); err != nil {
+func ensureDirPrivate(dn string) error {
+	mode := os.FileMode(0700)
+
+	if err := os.MkdirAll(dn, mode); err != nil {
```
cclerget: There is almost no chance that a fancy umask alter those permissions here but just in case I would reset umask before creating directories/files by precaution

pkg/sypgp/sypgp.go
```diff
@@ -98,61 +105,67 @@ func PublicPath() string {
...
 
-	// create or open the secret OpenPGP key cache file
-	fs, err := os.OpenFile(SecretPath(), os.O_RDWR|os.O_CREATE, 0600)
+	return nil
+}
+
+func ensureFilePrivate(fn string) error {
+	mode := os.FileMode(0600)
+
+	fs, err := os.OpenFile(fn, os.O_RDWR|os.O_CREATE, mode)
```
cclerget: same here

cclerget (dismissed):

pkg/sypgp/sypgp.go
mem: yikes ... I tried to _only_ move code around, with minimal fixes, but given how what this code is doing is already not so nice, I'll do as suggested, thanks.

gvallee:
Just for reference, this will create major conflicts with the work done in #3165 but I think we should merge this PR first.

pkg/sypgp/sypgp.go
```diff
@@ -376,144 +479,136 @@ func GetPassphrase(message string, retries int) (string, error) {
...
 
 	// Ask to push the new key to the keystore
-	pushKeyQ, err := AskQuestion("Would you like to push it to the keystore? [Y,n] : ")
-	if err != nil {
-		return
-	}
+	ans, err := askYNQuestion("y", "Would you like to push it to the keystore? [Y,n] ")
+	switch {
+	case err != nil:
+		fmt.Printf("Not pushing newly created key to keystore: %s\n", err)
```
WestleyK: Should this print to stderr?

```suggestion
		fmt.Fprintf(os.Stderr, "Not pushing newly created key to keystore: %s\n", err)
```

pkg/sypgp/sypgp.go
```diff
@@ -376,144 +479,136 @@ func GetPassphrase(message string, retries int) (string, error) {
...
-	fmt.Printf("Done.\n")
 
-	return
+	return entity, nil
 }
 
 // DecryptKey decrypts a private key provided a pass phrase
 func DecryptKey(k *openpgp.Entity, message string) error {
+	if !k.PrivateKey.Encrypted {
+		return errNotEncrypted
```
WestleyK: Why return a error if not encrypted?

pkg/sypgp/sypgp_test.go
```diff
@@ -6,16 +6,20 @@
 package sypgp
```
WestleyK:
```
// Copyright (c) 2018-2019, Sylabs Inc. All rights reserved.
```

Update copyright year in this file.

WestleyK (commented):
Couple suggestions/questions, otherwise looks really good.

pkg/sypgp/sypgp.go
mem: my expectation is that if you are calling the DecryptKey function it's because you think the key is encrypted. If the key is not encrypted, why are you calling DecryptKey?

bauerm97 (approved):
I gave it a 10 minute look through, structure of the code looks good to me. As long as CI is passing, I don't see any glaring reasons why we shouldn't merge

pkg/sypgp/sypgp.go
WestleyK: What if you remove the `if !k.PrivateKey.Encrypted {...}`, then when we want to decrypt a key later we can do:

```go
if k.PrivateKey.Encrypted {
    err := sypgp.DecryptKey(entity, pass)
    if err != nil {
        sylog.Fatalf("Unable to decrypt key: %s", err)
    }
}
```

Of course you can check `k.PrivateKey.Encrypted` anyway, but then `k.PrivateKey.Encrypted` will be called twice.

pkg/sypgp/sypgp.go
```diff
@@ -376,144 +479,136 @@ func GetPassphrase(message string, retries int) (string, error) {
...
+	return el[n], nil
 }
 
 // SelectPrivKey prints a secret key list to user and returns the choice
 func SelectPrivKey(el openpgp.EntityList) (*openpgp.Entity, error) {
-	PrintPrivKeyring()
+	printEntities(os.Stdout, el)
 
-	index, err := AskQuestion("Enter # of signing key to use : ")
+	n, err := askNumberInRange(0, len(el)-1, "Enter # of signing key to use : ")
```
```suggestion
	n, err := askNumberInRange(0, len(el)-1, "Enter # of private key to use : ")
```

Because when you export a key.

pkg/sypgp/sypgp.go
```diff
@@ -376,144 +479,136 @@ func GetPassphrase(message string, retries int) (string, error) {
...
-		return nil, fmt.Errorf("invalid key choice")
-	}
-
-	return el[i], nil
+	return el[n], nil
 }
 
 // SelectPrivKey prints a secret key list to user and returns the choice
 func SelectPrivKey(el openpgp.EntityList) (*openpgp.Entity, error) {
-	PrintPrivKeyring()
```
WestleyK, line 651: Please add:

```go
if el == nil {
    return nil, fmt.Errorf("no private keys in local keyring")
}
```

To avoid a panic. _and same for `SelectPubKey`._

WestleyK (changes requested):
Left some suggestions, and this comment: https://github.com/sylabs/singularity/pull/3290#discussion_r278761615

pkg/sypgp/sypgp.go
mem: I see this with an empty keyring:

```
$ > /home/mem/.singularity/sypgp/pgp-public
$ builddir/singularity key list
Public key listing (/home/mem/.singularity/sypgp/pgp-public):

$ > /home/mem/.singularity/sypgp/pgp-secret
$ builddir/singularity sign alpine_3.9.sif
Signing image: alpine_3.9.sif
ERROR:   signing container failed: no private keys in keyring. use 'key newpair' to generate a key, or 'key import' to import a private key from gpg
```

pkg/sypgp/sypgp.go
mem: Done, thanks.

pkg/sypgp/sypgp.go
WestleyK: First, I made a PR for this change https://github.com/sylabs/singularity/pull/3857 (bug fix PR). Second, it will only panic if you try to export a key.

WestleyK: Probably best to just ignore this and it will be fixed in the bug-fix PR. Better to keep bug-fixes separate with testing PRs.

WestleyK (commented):
Left a comment: https://github.com/sylabs/singularity/pull/3290#discussion_r300819863, just ignore my suggestion (it will be fixed in a separate bug-fix PR). Otherwise looks good.

WestleyK (approved):

pkg/sypgp/sypgp.go
mem: this is resolved, dismissing in order to be able to move.

Issue #2722, opened 2019-02-18
library: uninformative error messages when referencing image without tag when latest tag does not exist


gmkurtzer:
## Version of Singularity:

```
$ singularity version
3.0.3-756.ga72e4274
```

### Expected behavior

"latest" tag isn't assumed and it is confusing when pulling a container that doesn't have latest tag.

### Actual behavior

I see why this behavior maybe deemed as appropriate, but from a user's perspective the error is confusing and unexpected.

### Steps to reproduce behavior

```
$ singularity search centos
No users found for 'centos'

No collections found for 'centos'

Found 7 containers for 'centos'
	library://dtrudg/linux/centos
		Tags: 6 7 centos6 centos7 latest
	library://library/default/centos
		Tags: 6 7 latest
	library://gmk/demo/centos-vim
		Tags: latest
	library://mroche/baseline/centos
		Tags: 7 7.5 7.5.1804 7.6 7.6.1810 latest
	library://gmk/default/centos7-devel
		Tags: latest
	library://emmeff/default/centos7-python36
		Tags: 1.0
	library://westleyk/official/centos
		Tags: 7

$ singularity pull library://emmeff/default/centos7-python36
FATAL:   While getting image info: the requested image was not found in the library
$ singularity pull library://emmeff/default/centos7-python36:1.0
INFO:    Downloading library image
 82.90 MiB / 82.90 MiB [==========================================================================================================] 100.00% 10.03 MiB/s 8s
$
```

We either need to change the error to include a tag, show the available tags, or assume latest.

bauerm97:
@gmkurtzer So to start, I _do_ think it's necessary for us to add the `latest` tag to an image which has no tag. If we don't add a tag, we have no way of pulling an image. You could argue that we should just grab the tag with the most recent timestamp, but I'm not sure the library supports this, and it would be confusing in the case of e.g. a nightly devel build being tagged, but a `latest` tag existing to mark the latest **stable** version of an image. 

I think this primarily boils down to a logging issue. Let's make sure to do the following:

1. When a tag is not specified, let's output a nice message saying something along the lines of "No tag specified, searching for tag: latest".
2. When no latest tag is then found, we print out a much more descriptive error message. The error should detail if an image collection exists, and if it does exist but without the tags it should also note that. Something like "The image [...] does not have the specified tag: latest", vs "The image [...] does not exist" (in cases where the image simply does not exist, regardless of tag)

[...]

PR #3329, opened 2019-04-15
More Verbose When Pulling a Container Without a Tag

WestleyK:

This PR will add more verbose messages when pulling a container without a tag.

#### Example:

```bash
$ singularity pull library://alpine
INFO:    No tag specifyed, using ':latest'   <- ***This line was not there before
[...]

$ singularity pull library://jialigithub/default/alpine
FATAL:   While getting image info: image 'alpine:latest' was not found in 'https://library.sylabs.io'
```

- Related #2722 

<br>


internal/pkg/libexec/pull.go
```diff
@@ -39,13 +42,18 @@ func PullShubImage(filePath, shubRef string, force, noHTTPS bool) {
 }
 
 // PullOciImage pulls an OCI image to a sif
-func PullOciImage(path, uri string, opts types.Options) {
-	b, err := build.NewBuild(uri, build.Config{Dest: path, Format: "sif", Opts: opts})
+func PullOciImage(path, imageURI string, opts types.Options) {
```
ikaneshiro: I don't think this is the behavior we want from an `oci` pull. My understanding of the issue was that this should only affect pulls from `library`

internal/pkg/libexec/pull.go
```diff
@@ -39,13 +42,18 @@ func PullShubImage(filePath, shubRef string, force, noHTTPS bool) {
...
+	_, imageName := uri.Split(imageURI)
+	if !strings.Contains(imageName, ":") {
```
ikaneshiro: I don't think this is the behavior we want from an `oci` pull. My understanding of the issue was that this should only affect pulls from `library`

ikaneshiro (changes requested):
Don't think the changes to the `oci` pull are needed. Rest looks good though.

pkg/client/library/pull.go
```diff
@@ -38,6 +38,7 @@ func DownloadImage(filePath string, libraryRef string, libraryURL string, Force
 	libraryRef = strings.TrimPrefix(libraryRef, "library://")
 
 	if !strings.Contains(libraryRef, ":") {
+		sylog.Infof("No tag specifyed, using ':latest'")
```
ikaneshiro: Whoops missed this on the first round, specifyed -> specified

ikaneshiro (changes requested):

ikaneshiro (approved):

PR #3343, opened 2019-04-17
Add scripts/check-vendor and update .circleci/config.yml to use it

mem:
scripts/check-vendor makes sure that the contents of the vendor/
directory match the modules specified in go.mod. This prevents changes
that accidentally modify go.mod without updating vendor from merging. It
also prevents accidental modifications to vendor.


jscook2345 (approved):
LGTM mmmm

PR #3360, opened 2019-04-17
Add initial support for pushing SIF to OCI registries (ORAS)

### LINKED ISSUE #2831: oras:// URI to support OCI registry [e.g. azurecr.io] as SIF storage endpoint
According to @vermagit it is possible to push arbitrary artifacts (including SIF files) up to OCI registries using a tool called [oras](https://github.com/deislabs/oras/) like so:

```
$ singularity pull library://verma/default/layer:centos

$ oras push hpccontainers.azurecr.io/singularity:centos ./layer_centos.sif -u hpccontainers -p $PASSWORD
```

oras is written in Go and has an MIT license.  Maybe we could integrate it with Singularity and use it in our push command so that SIF files can be stored on OCI registries.

bauerm97:
This work is currently leveraging github.com/deislabs/oras to
help facilitate this. However, it looks like we need to implement
a content.Provider since the oras package doesn't quite fulfill
what we need. In its current state, we either need to write the
config JSON into a file, or read the entirety of the SIF file into
memory to allow oras to push it. We can fix this before we merge
this PR into Singularity.

Signed-off-by: Michael Bauer <michael@bauer.dev>

**To Do:**

- Understand if we need to use a custom layer `mediaType` or should just use the standard oci image layer `mediaType`
- ~Properly pass the manifest config `mediaType` to `oras.Push`~
- Testing
- When pulling, we should pull to a `oras.FileStore` located in a temporary location, since we don't actually have the ability to rename the file before it's pulled. By moving the `FileStore` to a temporary disposable location, we can first pull the file there, then move it to it's final location with the correct name.

- Fixes #2831 
[Kind of, we decided that from a technical perspective it makes more sense to not implement this in a plugin]

cc: @SteveLasker @sajayantony


shizhMSFT:
Since the config file is not actually used, it is better not to add `SifConfigMediaType` to `allowedMediaTypes`. You might want to discard the log by `content.WithLoggerDiscarded()` or simply use `context.Background()` from `github.com/deislabs/oras/pkg/context`.

shizhMSFT:
If you know the file name in advance, you can invoke `FileStore.MapPath()` to relocate the path to store the file. Note: This method is thread-safe.

To know the filename, you might want to use the `oras.WithPullBaseHandler` option in the latest version of oras package (still developing, will be stabilized at v0.4.0) to get the descriptor before fetching. 

For example:
```go
_, _, err = oras.Pull(orascontext.Background(), resolver, ref, store,
	oras.WithAllowedMediaType(SifLayerMediaType),
	oras.WithPullBaseHandler(images.HandlerFunc(func(ctx context.Context, desc ocispec.Descriptor) ([]ocispec.Descriptor, error) {
		if desc.MediaType == SifLayerMediaType {
			name, _ := orascontent.ResolveName(desc)
			store.MapPath(name, desirePath)
		}
		return nil, nil
	})))
```

bauerm97:
Closing in favor of #3526

PR #3362, opened 2019-04-18
Make the testing makefile rules more consistent

jscook2345:

Make the testing makefile rules more consistent so they can be split up later.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -178,7 +178,7 @@ jobs:
           name: Run E2E tests
           command: |-
             cd $HOME/go/singularity
-            make -C ./builddir e2e-tests
+            make -C ./builddir e2e-test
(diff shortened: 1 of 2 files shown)


ArangoGutierrez (changes requested):
you also need to apply this change on travis and circlci file

jmstover (approved):

jscook2345:
I actually fixed the `.PHONY`.

ArangoGutierrez:
My point is not about the PHONY, but about the make target. circleci config will be calling a make target that doesn't exist

jscook2345:
It never did. `.PHONY` does not create a proper make target. I'm just changing things now to fix that.

ArangoGutierrez (approved):

PR #3382, opened 2019-04-22
Add tests for the oci commands help text

jscook2345:
Add tests for the oci commands help text

My main concern is that I did the `go mod` and `go vendor` stuff correctly. If @mem could double check that and @ArangoGutierrez could check the help code that would be great!

Thanks!


e2e/help/help.go
```diff
@@ -0,0 +1,74 @@
...
+	{[]string{"oci", "resume"}},
+	{[]string{"oci", "run"}},
+	{[]string{"oci", "start"}},
+	{[]string{"oci", "state"}},
+	{[]string{"oci", "umount"}},
+	{[]string{"oci", "update"}},
+}
+
+func TestHelp(t *testing.T) {
+	c := test.NewCmd("singularity")
```
mem: `testenv.CmdPath` contains the path to the singularity binary. Shouldn't this use that?

e2e/help/help.go
```diff
@@ -0,0 +1,74 @@
...
+
+	for _, tc := range helpTests {
+		name := fmt.Sprintf("%s.txt", strings.Join(tc.cmds, "-"))
+		path := filepath.Join("help", name)
+		args := append([]string{"help"}, tc.cmds...)
+
+		t.Run(name, func(t *testing.T) {
```
mem, line 56: matter of taste, but I would prefer if the subtest was named help-oci-start instead of oci-start.txt

e2e/help/help.go
```diff
@@ -0,0 +1,74 @@
...
+			got := c.Run(t, args...).Stdout()
```
mem: I would prefer if this used a local variable `c` (to the anonymous function) instead of having this declared in the scope of TestHelp.

e2e/help/help.go
```diff
@@ -0,0 +1,74 @@
...
+	{[]string{"oci", "pause"}},
+	{[]string{"oci", "resume"}},
+	{[]string{"oci", "run"}},
+	{[]string{"oci", "start"}},
+	{[]string{"oci", "state"}},
+	{[]string{"oci", "umount"}},
+	{[]string{"oci", "update"}},
+}
+
+func TestHelp(t *testing.T) {
```
mem: this does not need to be exported, you can use `testHelp`

e2e/help/help.go
```diff
@@ -0,0 +1,74 @@
...
+package help
+
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/kelseyhightower/envconfig"
+	"github.com/sylabs/singularity/internal/pkg/test"
```
mem, line 15: This is not a problem, but I'm also not so happy about the idea of having code that is only used in tests living in this path. You could move this to `e2e/internal/pkg/test` (or perhaps a more descriptive name than just `test`).

e2e/testdata/help/oci-attach.txt
```diff
@@ -0,0 +1,17 @@
+Attach console to a running container process (root user only)
```
mem: there should be a README.md here explaining:

- how to create these files (the output of `singularity help` is sensitive to some environment variables)
- how to name these files
- how / when to update these files

mem (dismissed):

mem:
module files look good

e2e/help/help.go
ArangoGutierrez: agree

e2e/help/help.go
ArangoGutierrez: +1

e2e/help/help.go
ArangoGutierrez: this can be a whole issue and added to the EPIC, I will do that

e2e/help/help.go
jscook2345: Fixed

e2e/help/help.go
jscook2345: Fixed

e2e/help/help.go
jscook2345: Fixed

e2e/help/help.go
jscook2345: Those packages are meant to be used by all (both e2e and unit) testing where needed. That is why I put it there. It is meant to replace the use of the `exec.Command` and add some traceability.

e2e/help/help.go
jscook2345: This seems like it would add a lot of extra instances (1 per loop?) of the struct to be garbage collected when just one instance seems like enough. What do you think @mem ?

e2e/testdata/help/oci-attach.txt
jscook2345: Done!

jscook2345:
Ready for re-review!

ArangoGutierrez (changes requested):
could we find a way to generate the `txt` files ? having them hard coded can lead to errors in the future, I think that with cobra we can generate this files.   maybe using https://github.com/sylabs/singularity/tree/master/docs

jscook2345:
@ArangoGutierrez 

Sure. That's not a problem. It's just a matter of further using the `golden` library. I think that can be another PR after this one though, since it will take some experimentation.

ArangoGutierrez:
> @ArangoGutierrez
> 
> Sure. That's not a problem. It's just a matter of further using the `golden` library. I think that can be another PR after this one though, since it will take some experimentation.

ok let's make an issue for that, and tag it here as a comment, I will approve and merge after that :) 
great work team!

jscook2345:
https://github.com/sylabs/singularity/issues/3476 !

I will fix the conflict and than re-request a review @ArangoGutierrez

ArangoGutierrez (approved):

ArangoGutierrez:
*PING* @mem 
LGTM

jscook2345:
The old `help_test.go` was ported over and took over my 'namespace' so there's a lot of git conflicts. I think it would be better to just start the PR over at this point.

Issue #3315, opened 2019-04-15
Improve test coverage of internal/app


gvallee:
## Version of Singularity:

Improve test coverage of internal/app.

### Expected behavior

Get close to 100% coverage.

### Actual behavior

`
$ go test -coverprofile=coverage.out && go tool cover -func=coverage.out
?   	github.com/sylabs/singularity/internal/app/starter	[no test files]
total:	(statements)	0.0%
`

### Steps to reproduce behavior

Write here.

gvallee:
Need to improve coverage for both `internal/app/singularity` and `internal/app/starter`

carterpeel:
Hello,

This is a templated response that is being sent out to all open issues.  We are working hard on 'rebuilding' the Singularity community, and a major task on the agenda is finding out what issues are still outstanding.


**Please consider the following:**

1. Is this issue a duplicate, or has it been fixed/implemented since being added?
2. Is the issue still relevant to the current state of Singularity's functionality?
3. Would you like to continue discussing this issue or feature request?


Thanks,
Carter

pedroalvesbatista:
@gvallee Still getting this problem ? Have this been solved already ? If yes, what work around have you followed and applied ?

We're looking into the issue carefully, soon will bring to community and discuss ways to better solve as well address this. Thankyou for keeping the interest in the subject.

kmuriki:
This repository is no longer active. Please open new issue at https://github.com/apptainer/apptainer.

PR #3396, opened 2019-04-23
Improve testing coverage of internal/app/singularity

gvallee:

Improve testing coverage of internal/app/singularity, focusing on `RemoteAdd()` and `RemoteRemove()`

- Addresses #3315 


gvallee:
I would like to have feedback about the cases that I am testing in the context of `RemoteAdd()` and `RemoteRemove()`. Specifically, I would like to have feedback on whether some of the corner cases that currently pass should in fact fail and whether some error paths should success. In other words, I would like to clarify the intent.

internal/app/singularity/remote_add_test.go
```diff
@@ -0,0 +1,300 @@
...
+			// RemoveRemove().
+			name:       "9: valid config file; valid remote name; valid URI; local",
+			cfgfile:    validCfgFile,
+			remoteName: "cloud_testing",
+			uri:        "cloud.random.io",
+			global:     false,
+			shallPass:  true,
+		},
+		{
+			name:       "10: valid but dummy file; empty remote name; empty URI; local",
```
ikaneshiro: I think this is a duplicate of 2?

internal/app/singularity/remote_add_test.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+	cfg := aDummyStruct{
+		NoneSenseRemote: "toto",
+	}
+
+	yaml, err := yaml.Marshal(cfg)
+	if err != nil {
+		f.Close()
+		os.Remove(path)
+		t.Fatalf("cannot marshal YANK: %s\n", err)
```
ikaneshiro: Typo: `YANK`->`YAML`

ikaneshiro (changes requested):
I think several of these corner cases should indeed fail. I think the `remove` behavior is ok but here are the cases I think `add` should catch and fail:
- Empty remote name
- Empty remote URI

With respect to `invalidCfgFile`. I think this should always cause a failure but this needs to be fixed in `ReadFrom()` in the `remote` package by having it use to use `UnmarshallStrict()`.

gvallee:
Thanks for the feedback, I will address of your comments and update the PR.

gvallee:
I create issue #3398 to track and fix the problem with ReadFrom().

internal/app/singularity/remote_add_test.go
gvallee: I made a pass over all the cases, removed the duplicates and added the missing cases.

internal/app/singularity/remote_add_test.go
gvallee: fixed

ikaneshiro: Looks like the typo is still there

internal/app/singularity/remote_add.go
```diff
@@ -10,12 +10,21 @@ import (
...
+	"strings"
 
 	"github.com/sylabs/singularity/internal/pkg/remote"
 )
 
 // RemoteAdd adds remote to configuration
 func RemoteAdd(configFile, name, uri string, global bool) (err error) {
+	// Explicit handling of corner cases: name and uri must be valid strings
+	if strings.TrimSpace(name) == "" {
+		return fmt.Errorf("invalid name: %s", name)
```
ikaneshiro: Printing an error with an empty string might be a bit confusing. How about something like `invalid name: cannot have empty name`

internal/app/singularity/remote_add.go
```diff
@@ -10,12 +10,21 @@ import (
...
+	}
+	if strings.TrimSpace(uri) == "" {
+		return fmt.Errorf("invalid URI: %s", uri)
```
ikaneshiro: Same here

ikaneshiro (changes requested):

internal/app/singularity/remote_add.go
gvallee: Good point, fixed

internal/app/singularity/remote_add.go
gvallee: Done

internal/app/singularity/remote_add_test.go
gvallee: Very sorry about that. I am getting tired and tried to go too fast. It is now fixed.

gvallee:
@ikaneshiro I would like now to have the confirmation that cases 12, 13, 29 and 30 are indeed supposed to succeed. My impression is that they should not but I am absolutely not sure.

ikaneshiro:
@gvallee I think those should pass.

Actually I think we can remove the tests in add that are only testing for `invalid remote name` or `invalid URI` since users should be allowed to add remotes with whatever names and locations they want. The only case to test for name and uri would be an empty entry.

Testing for invalid names makes sense for the remove command though since the only valid names are the ones currently configured

gvallee:
@ikaneshiro It makes sense. I was concerned about the URI since some cases clearly do not use a valid URI but I understand your point. I will update the code ASAP and maybe then we could merge that PR; I think this work clarify a few things. Thanks again for your time!

gvallee:
All done

ikaneshiro (approved):

PR #3400, opened 2019-04-23
Update ReadFrom() to properly handle invalid YAML files

### LINKED ISSUE #3398: Update ReadFrom() in the remote package to validate config files
## Version of Singularity:

At the moment, when using RemoteAdd(), the function will call ReadFrom() which would succeed even if the YAML file is totally invalid (wrong elements, dummy values, etc.). 

### Expected behavior

ReadFrom() should return an error when the YAML file is invalid. For that, it should use UnmarshallStrict().

### Actual behavior

It will successfully read the YAML file and return to the caller without an error.

Unit test for internal/app/singularity, especially PR #3396

gvallee:

Update the ReadFrom() code and the associated test to correctly handle invalid YAML files. It now returns an error, as expected. Modifications based on the suggestion from Ian Kaneshiro.

- Fixes #3398 

### DIFF
--- a/internal/pkg/remote/remote_test.go
+++ b/internal/pkg/remote/remote_test.go
@@ -8,7 +8,6 @@ package remote
 import (
 	"bytes"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"reflect"
 	"testing"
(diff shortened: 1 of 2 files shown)


internal/pkg/remote/remote_test.go
```diff
@@ -72,6 +88,39 @@ func TestWriteToReadFrom(t *testing.T) {
...
 
+	for _, test := range testsFail {
+		t.Run(test.name, func(t *testing.T) {
+			yaml, err := yaml.Marshal(test.data)
+			if err != nil {
+				t.Fatalf("cannot mashal YAML: %s\n", err)
+			}
+
+			// We manually create the file to test ReadFrom()
+			tempFile, err := ioutil.TempFile("", "")
```
ikaneshiro: I think its cleaner to implement this test using a `bytes.Buffer` like `testsPass` is since both types implement the same interfaces

ikaneshiro (changes requested):
Looks good. One suggestion

internal/pkg/remote/remote_test.go
gvallee: Good point and done.

ikaneshiro (approved):

PR #3405, opened 2019-04-24
Improve e2e driver error reporting

mem:
Use singularity's binary derived from compilation configuration to make
sure we are testing the binary that matches the source tree.

Use full paths when reporting errors so that the user knows what needs
to be fixed.


ArangoGutierrez (approved):
Awesome thanks

PR #3414, opened 2019-04-25
Return an error instead of calling os.Exit()

gvallee:

Return an error instead of calling os.Exit() in order to facilitate testing. The callers are already checking for errors and appropriately calling os.Exit().

- Addresses #3315 


mem (approved):

internal/app/singularity/cache_clean_linux.go
```diff
@@ -65,10 +65,9 @@ func CleanCache(cacheType string) error {
 		err := cache.Clean()
 		return err
 	default:
-		sylog.Fatalf("Not a valid type: %v", cacheType)
-		os.Exit(2)
+		// The caller checks the returned error and will exit as required
+		return fmt.Errorf("Not a valid type: %v", cacheType)
```
WestleyK: Uncap `N`, and use `%s`.

```suggestion
		return fmt.Errorf("not a valid type: %s", cacheType)
```

internal/app/singularity/cache_clean_linux.go
```diff
@@ -176,8 +175,8 @@ func CleanSingularityCache(cleanAll bool, cacheCleanTypes []string, cacheName st
 		case "all":
 			cleanAll = true
 		default:
-			sylog.Fatalf("Not a valid type: %v", t)
-			os.Exit(2)
+			// The caller checks the returned error and exit when appropriate
+			return fmt.Errorf("Not a valid type: %v", t)
```
WestleyK: Same here.

internal/app/singularity/cache_clean_linux.go
```diff
@@ -187,8 +186,7 @@ func CleanSingularityCache(cleanAll bool, cacheCleanTypes []string, cacheName st
 			return err
 		}
 		if !foundMatch {
-			sylog.Warningf("No cache found with given name: %v", cacheName)
-			os.Exit(0)
+			return fmt.Errorf("No cache found with given name: %v", cacheName)
```
WestleyK: Umm, this will exit with a non-0. I think it would be better if you do:

```go
sylog.Warningf("No cache found with given name: %s", cacheName)
return nil
```

Because if a user just wants to make sure that cache is cleared, without failing in there script.

WestleyK (changes requested):
Thanks for fixing this up! I have a couple suggestions...

internal/app/singularity/cache_clean_linux.go
gvallee: fixed

internal/app/singularity/cache_clean_linux.go
gvallee: fixed

internal/app/singularity/cache_clean_linux.go
gvallee: fixed

internal/app/singularity/cache_clean_linux.go
```diff
@@ -187,8 +186,8 @@ func CleanSingularityCache(cleanAll bool, cacheCleanTypes []string, cacheName st
 			return err
 		}
 		if !foundMatch {
-			sylog.Warningf("No cache found with given name: %v", cacheName)
-			os.Exit(0)
+			sylog.Warningf("No cache found with given name: %s", cacheName)
+			return nil
```
WestleyK: I dont think the `return nil` is necessary... Because it will `return nil` right after the `if !foundMatch {...}`.

WestleyK (commented):
One last comment/suggestion. Thanks :+1:

internal/app/singularity/cache_clean_linux.go
gvallee: Good catch. Fixed.

internal/app/singularity/cache_clean_linux.go
```diff
@@ -65,10 +65,9 @@ func CleanCache(cacheType string) error {
 		err := cache.Clean()
 		return err
 	default:
-		sylog.Fatalf("Not a valid type: %v", cacheType)
-		os.Exit(2)
+		// The caller checks the returned error and will exit as required
+		return fmt.Errorf("not a valid type: %s", cacheType)
 	}
-	return nil
```
WestleyK, line 71: Why did you remove this line?

gvallee: Yes the linter complained because it is a non-reachable line: all cases of the switch are calling return, including default.

WestleyK: Okay :+1:

WestleyK (approved):
Looks good to me! Thanks for fixing this up :shipit:

PR #3429, opened 2019-04-26
Add EnsureImage function to make sure a test image exists

### LINKED ISSUE #3409: [E2E]  Create a stand alone image retrieval func
Move https://github.com/sylabs/singularity/blob/master/e2e/suite.go#L85 
to a stand alone func that basically says "give me a test image", that will check for a temp folder, if the image exist, grab the full path, if not, build it and return the path.

mem:
Instead of simply creating an image for every run of e2e tests, move the
test image creation to a separate package and provide a function
(EnsureImage) that makes sure the image exists before the test that
needs it starts to run.

Add LoadEnv to avoid copy-pasting the same code to load the environment.

- Fixes #3409


mem:
Naming and organization can probably be improved, suggestions welcomed.

jscook2345 (approved):
LGTM

PR #3431, opened 2019-04-26
Check the Go compiler during mconfig

### LINKED ISSUE #3376: Building singularity 3.2+ requires Go 1.11+
## Version of Singularity:

3.2

### Expected behavior

A message indicating that the installed Go version is not recent enough.

### Actual behavior

Build fails with error about invalid option -deps

Install Go 1.10, run ./mconfig, build singularity.

The build will fail when trying to compute build dependencies (before it actually builds anything) because the build-dependency generation script uses `go list -deps`, which is not available before Go 1.11. If that's removed, the build will _still_ fail because we require 1.11+ for module support. So the minimum version of Go that we support is 1.11, and mconfig should emit an error about that.

mem:
- Check that there's a Go compiler
- Check that the version of that compiler is at least what we need (1.11
  as of this writing)

Fixes: #3376

### DIFF
--- a/mlocal/checks/basechecks.chk
+++ b/mlocal/checks/basechecks.chk
@@ -74,7 +74,7 @@ else
 	printf " checking: host Go compiler (at least version $hstgo_version)... "
 	if $hstgo run $makeit_checksdir/version.go $hstgo_version >/dev/null 2>&1; then
 		hstgo=`command -v $go`
-		echo $hostgo
+		echo $hstgo
 	else
 		echo "$hstgo cannot compile test code!"
 		usage
--- a/mlocal/checks/basechecks.chk
+++ b/mlocal/checks/basechecks.chk
@@ -55,6 +55,33 @@ else
 	fi
 fi
 
+if [ "$hstgo" = "" ]; then
+	printf " checking: host Go compiler (at least version $hstgo_version)... "
+	for go in $hstgo_opts; do
+		if $go run $makeit_checksdir/version.go $hstgo_version >/dev/null 2>&1; then
+			hstgo=`command -v $go`
+			break
+		fi
+	done
+	if [ "$hstgo" != "" ]; then
+		echo $hstgo
+	else
+		echo "not found!"
+		usage
+		exit 1
+	fi
+else
+	printf " checking: host Go compiler (at least version $hstgo_version)... "
+	if $hstgo run $makeit_checksdir/version.go $hstgo_version >/dev/null 2>&1; then
+		hstgo=`command -v $go`
+		echo $hostgo
+	else
+		echo "$hstgo cannot compile test code!"
+		usage
+		exit 1
+	fi
+fi
+
 ########################
 # host compilers options
 ########################
--- a/mlocal/checks/version.go
+++ b/mlocal/checks/version.go
@@ -1,0 +1,22 @@
+package main
+
+import (
+	"fmt"
+	"go/build"
+	"os"
+)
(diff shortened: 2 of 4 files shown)


mlocal/checks/version.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"fmt"
+	"go/build"
+	"os"
+)
+
+func main() {
+	if len(os.Args) != 2 {
+		fmt.Printf("E: missing target Go version\n")
```
WestleyK, line 11: Should this print to stderr?

mlocal/checks/basechecks.chk
```diff
@@ -55,6 +55,33 @@ else
...
+	else
+		echo "not found!"
+		usage
+		exit 1
+	fi
+else
+	printf " checking: host Go compiler (at least version $hstgo_version)... "
+	if $hstgo run $makeit_checksdir/version.go $hstgo_version >/dev/null 2>&1; then
+		hstgo=`command -v $go`
+		echo $hostgo
```
cclerget:
```suggestion
		echo $hstgo
```

mlocal/checks/basechecks.chk
```diff
@@ -55,6 +55,33 @@ else
 	fi
 fi
 
+if [ "$hstgo" = "" ]; then
+	printf " checking: host Go compiler (at least version $hstgo_version)... "
+	for go in $hstgo_opts; do
+		if $go run $makeit_checksdir/version.go $hstgo_version >/dev/null 2>&1; then
+			hstgo=`command -v $go`
```
cclerget: Use `which` instead, `command` may not be installed

mlocal/checks/basechecks.chk
```diff
@@ -55,6 +55,33 @@ else
...
+		echo $hstgo
+	else
+		echo "not found!"
+		usage
+		exit 1
+	fi
+else
+	printf " checking: host Go compiler (at least version $hstgo_version)... "
+	if $hstgo run $makeit_checksdir/version.go $hstgo_version >/dev/null 2>&1; then
+		hstgo=`command -v $go`
```
cclerget: same

cclerget (changes requested):

jscook2345 (approved):
LGTM. I'm not sure if using `which` or `command -v` is better. I prefer `command -v` because it's posix and part of the shell, removing a needed dependency. However, I will leave it to your discretion.

WestleyK:
^^^ I agree with @jscook2345 :+1:

mem:
> LGTM. I'm not sure if using `which` or `command -v` is better. I prefer `command -v` because it's posix and part of the shell, removing a needed dependency. However, I will leave it to your discretion.

`command -v` is part of the base specification (http://pubs.opengroup.org/onlinepubs/9699919799/utilities/command.html) and it's usually shell built-in; `which` is an optional program. The other option, `type -P`, is a bashism. The problem with `command -v` is that it will report shell functions, but it's still preferred over which and type.

mlocal/checks/basechecks.chk
mem: Ah, thanks!

cclerget (approved):

mlocal/checks/version.go
mem: I don't think it makes a difference, given how this code is used specifically to test the version of the compiler, but yes, that can be changed.

PR #3435, opened 2019-04-29
Migrate Definition Build Tests to E2E Framework

### LINKED ISSUE #3356: [E2E] port integration test "definition files" to the e2e framework
## E2E Task

Port: 
- https://github.com/sylabs/singularity/blob/master/cmd/singularity/deffile_test.go
to the e2e framework 
- https://github.com/sylabs/singularity/tree/master/e2e
under `e2e/imgbuild`

ikaneshiro:

- Moves definition based build tests to e2e framework
- Splits image related `e2e` functions into multiple files named after their purpose

- Fixes #3356 


e2e/imgbuild/imgbuild.go
```diff
@@ -154,6 +159,511 @@ func badPath(t *testing.T) {
 	}
 }
 
+func buildMultiStageDefinition(t *testing.T) {
+	tmpfile, err := ioutil.TempFile(testenv.TestDir, "testFile-")
```
jscook2345: This `tmpfile` stuff looks like it would be good to move to a function for reuse!

e2e/imgbuild/imgbuild.go
```diff
@@ -154,6 +159,511 @@ func badPath(t *testing.T) {
...
+	defer os.Remove(tmpfile.Name()) // clean up
+
+	if _, err := tmpfile.Write([]byte(testFileContent)); err != nil {
+		log.Fatal(err)
+	}
+	if err := tmpfile.Close(); err != nil {
+		log.Fatal(err)
+	}
+
+	tests := []struct {
```
jscook2345, line 166: These test cases are really difficult to discern what is going on. I know that you're trying to avoid repetition with templates but perhaps here you should use actual definition files that people can view under /examples or /e2e/testdata ? What do you think?

e2e/imgbuild/imgbuild.go
```diff
@@ -154,6 +159,511 @@ func badPath(t *testing.T) {
...
```
jscook2345, line 364: These are a bit easier to read in comparison to the multi-stage test cases, but I'm wondering if all the code to generate these templates is worth it over creating the definition files themselves and having them as test artifacts?

e2e/imgbuild/imgbuild.go
```diff
@@ -163,7 +673,11 @@ func RunE2ETests(t *testing.T) {
 	// builds from definition file and URI
 	t.Run("From", buildFrom)
 	// build and image from an existing image
-	t.Run("multistage", buildMultiStage)
+	t.Run("multistage", buildMultiple)
```
jscook2345: I'd prefer not naming these similarly. `multistage` and `MultiStage` below is a bit confusing.

e2e/imgbuild/imgbuild.go
```diff
@@ -154,6 +159,511 @@ func badPath(t *testing.T) {
...
+			opts := e2e.BuildOpts{
+				Sandbox: tt.sandbox,
+			}
+
+			imagePath := path.Join(testenv.TestDir, "container")
+			defer os.RemoveAll(imagePath)
+
+			if b, err := e2e.ImageBuild(testenv.CmdPath, opts, imagePath, defFile); err != nil {
+				t.Log(string(b))
+				t.Fatalf("unexpected failure: %v", err)
```
jscook2345, line 348: All of your `t.Fatalf` calls use `%v` for the format, but I think you can use `%s` instead.

e2e/internal/e2e/imageverify.go
```diff
@@ -0,0 +1,406 @@
...
+)
+
+// ImageVerify checks for an image integrity
+func ImageVerify(t *testing.T, cmdPath string, imagePath string, labels bool, runDisabled bool) {
+	type testSpec struct {
+		name          string
+		execArgs      []string
+		expectSuccess bool
+	}
+	tests := []testSpec{
```
jscook2345, line 28: Could these `/.singularity.d/` paths be put into `const` values somewhere? They seem to be repeating in a few places.

Perhaps a struct (`SingularityRoot`) could manage them?

```
sr = NewSingularityRoot('/.singularity.d')
test -f sr.GetRunScript()
test -f sr.GetEnvironmentFile()
```

What do you think?

jscook2345: This could be a feature added after your PR.

e2e/internal/e2e/imageverify.go
```diff
@@ -0,0 +1,406 @@
...
+
+			if err := verifyScript(t, scriptPath, app.Test); err != nil {
+				t.Fatalf("unexpected failure in app %v: test script: %v", app.Name, err)
+			}
+		}
+	}
+
+}
+
+func fileExists(t *testing.T, path string) bool {
```
jscook2345, line 227: This seems like it should go into a higher level pkg so it can be reused.

e2e/internal/e2e/imageverify.go
```diff
@@ -0,0 +1,406 @@
...
+		return false
+	} else if err != nil {
+		t.Fatalf("While stating file: %v", err)
+	}
+
+	return true
+}
+
+func verifyFile(t *testing.T, original, copy string) error {
+	ofi, err := os.Stat(original)
```
jscook2345, line 238: https://godoc.org/os#SameFile

This seems problematic for big files but I'm not sure we need to optimize this yet.

e2e/internal/e2e/imageverify.go
```diff
@@ -0,0 +1,406 @@
...
+		{"False", []string{"false"}, false},
+		{"RunScript", []string{"test", "-f", "/.singularity.d/runscript"}, true},
+		{"OneBase", []string{"test", "-f", "/.singularity.d/env/01-base.sh"}, true},
+		{"ActionsShell", []string{"test", "-f", "/.singularity.d/actions/shell"}, true},
+		{"ActionsExec", []string{"test", "-f", "/.singularity.d/actions/exec"}, true},
+		{"ActionsRun", []string{"test", "-f", "/.singularity.d/actions/run"}, true},
+		{"Environment", []string{"test", "-L", "/environment"}, true},
+		{"Singularity", []string{"test", "-L", "/singularity"}, true},
+	}
+	if labels && runDisabled { // TODO
```
jscook2345: What do you need to TODO here?

e2e/imgbuild/imgbuild.go
jscook2345: I wonder what the new error handling feature will want us to use? Maybe just leave this for now and we can raise the issue globally later :)

jscook2345 (dismissed):
I left some comments and some requested changes, but looks good so far!

ArangoGutierrez (changes requested):
could you pls add a header comment, like e.g on the file under cmd/singularity
```
// This file is been deprecated and will disappear on with version 3.3
// of singularity. The functionality has been moved to e2e/pull/pull.go
```

e2e/imgbuild/imgbuild.go
mem: > All of your `t.Fatalf` calls use `%v` for the format, but I think you can use `%s` instead.

I hope we can move to Go 1.13's error package (or [xerrors](https://godoc.org/golang.org/x/xerrors) in Go 1.12). With that pacakge, %v gets handled in such a way that it's possible to unpack the errors, so building a habit of using %v (or %+v) is a good thing.

e2e/imgbuild/imgbuild.go
ikaneshiro: I think it's worth it in this case because the amount of code we use to define what artifacts should be in the image is all we need to generate the definition itself from this struct. We could move to using definition files in a folder, but we would still need to have all these parameters defined in order to verify the final container was built properly.

e2e/imgbuild/imgbuild.go
ikaneshiro: I've been thinking about this for a bit and I'm a bit torn. Though I agree this isn't great to read, and for these tests using an actual definition file would cut down the amount of information needed in this test struct, it would be a significant change from `buildDefinition` and potentially be confusing for someone to understand the difference between the two at first glance. Maybe it might be easier for a reader if these tests appeared after `buildDefinition` so that they're familiar with what that is doing and seeing what this is doing will be more intuitive?

e2e/internal/e2e/imageverify.go
ikaneshiro: Seems like `SameFile()` is essentially checking if the files are stored in the same inodes and not checking the content? I think we want to check the content in our case. This is a potential alternative https://godoc.org/github.com/udhos/equalfile

e2e/internal/e2e/imageverify.go
ikaneshiro: What package are you thinking? I figured this `e2e` package is the best place to put helpers like this

e2e/internal/e2e/imageverify.go
ikaneshiro: Looks like this is no longer needed. Seems like an artifact of when singularity was still incomplete but adam was trying to write some basic tests

ArangoGutierrez:
*PING* @ikaneshiro

e2e/internal/e2e/imageverify.go
ikaneshiro: I think this would be best to add later on if we do extensive metadata related checks

ArangoGutierrez (approved):

e2e/internal/e2e/imageverify.go
jscook2345: Yeah, I just noticed that you have part of your function doing what os.SameFile does as well and thought you could substitute that in.

e2e/imgbuild/imgbuild.go
jscook2345: I just wanted to have a discussion here. As long as you've thought about it a bit and are open to more conversations I think it's something we can approach later.

jscook2345 (approved):
🏆

PR #3447, opened 2019-04-29
Wildcard support for %files section

### LINKED ISSUE #2957: Wildcard * not supported in %files section in version 3.0
Hi there!

I tried to build a recipe which works perfectly using singularity 2.6 but it fails in singularity 3.0 and 3.1. It fails when using the wildcard * while copying in the % files section. 
This was asked and solved in the singularity google group [here](https://groups.google.com/a/lbl.gov/forum/#!topic/singularity/DKyS8KRmVus) just removing the * and copying the entire folder inside /opt. However it will be useful to have the same functionality as in version 2.6.

## Version of Singularity:
version 3.0

### Expected behavior
The files in scif_app_recipes should be copied inside /opt

### Actual behavior
```
INFO:    Starting build... 
Skipping fetch of repeat blob sha256:8ba884070f611d31cb2c42eddb691319dc9facf5e0ec67672fcfa135181ab3df 
Copying config sha256:b9a1b1f0b2baaec83946a26d7045e4028f11eccc8b0e5b3514568e56a391beb2 
Storing signatures 
INFO:    Copying ./scif_app_recipes/* to /tmp/sbuild-006893969/fs/opt 
FATAL:   container creation failed: unable to copy files to container fs: While copying ./scif_app_recipes/* to /tmp/sbuild-006893969/fs/opt: exit status 1 
```

This is the recipe I am using.

```
Bootstrap: docker 
From: centos:latest 
 
%files 
        ./scif_app_recipes/* /opt/ 

%post 
        echo "Install basic development tools" 
        yum -y groupinstall "Development Tools" 
        yum -y update && yum -y install wget curl 
 
        echo "Install python2.7 setuptools and pip" 
        yum -y install python-setuptools 
        easy_install pip 
 
        echo "Installing SCI-F" 
    pip install scif 
 
    echo "Installing plasmidID app" 
    scif install /opt/plasmidid_v1.4.2_centos7.scif 
 
    # Executables must be exported for nextflow, if you use their singularity native integration. 
    # It would be cool to use $SCIF_APPBIN_bwa variable, but it must be set after PATH variable, because I tried to use it here and in %environment without success. 
    echo "PlasmidID Done" 
    echo "export LC_ALL=en_US.UTF-8" >> $SINGULARITY_ENVIRONMENT 
    find /scif/apps -maxdepth 2 -name "bin" | while read in; do echo "export PATH=\${PATH}:$in" >> $SINGULARITY_ENVIRONMENT;done 
 
 
%runscript 
    exec scif "$@"


```

ikaneshiro:

- Adds posix wildcard support for %files section using `sh` to interpret paths
- Changes `copy` package name to `files` as it's more descriptive
- Cleans up copy test dir that was previously left behind

- Fixes #2957 


internal/pkg/build/files/copy.go
```diff
@@ -17,21 +17,40 @@ import (
...
+	_, err = os.Stat(dst)
 	if os.IsNotExist(err) {
 		// if destination doesn't exist, create parent directories
 		if err := os.MkdirAll(filepath.Dir(dst), 0755); err != nil {
-			return fmt.Errorf("while creating parent directories: %v", err)
+			return fmt.Errorf("while creating parent directories: %s", err)
+		}
+		// if source expands to multiple elements, create destination as directory if it doesnt exist
+		if len(paths) > 1 {
+			if err := os.MkdirAll(dst, 0755); err != nil {
```
mem: if we hit this code, it should be creating exactly one directory, dst, as all the parent directories are already there because of line L29.

I would prefer to use os.Mkdir here.

internal/pkg/build/files/files.go
```diff
@@ -0,0 +1,39 @@
...
+func expandPath(path string) ([]string, error) {
+	var output, stderr bytes.Buffer
+	cmd := exec.Command("/bin/sh", "-c", "echo "+path)
+	cmd.Stdout = &output
+	cmd.Stderr = &stderr
+	if err := cmd.Run(); err != nil {
+		return nil, fmt.Errorf("%s: %s", err, stderr.String())
+	}
+
+	return strings.Fields(output.String()), nil
```
mem: since path can expand to filenames with spaces, I don't think this works.

In principle this works:

```
#!/bin/sh

for n in "$@" ; do
    printf "$n\0"
done
```

the arguments to that script get expanded by the calling shell, and that script prints each one separated by a NULL.

internal/pkg/build/files/files.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// Join wraps filepath.Join() and appends a trailing '/' if originally specified
+// in the last element since that has meaning to the cp command
```
mem: but then this is not a generic join function, but a special-purpose one. Should this be called Join?

internal/pkg/build/files/files_test.go
```diff
@@ -0,0 +1,232 @@
...
+
+*/
+
+func createTestDirLayout(t *testing.T) string {
+	testDirName, err := ioutil.TempDir("", "files-test-dir-")
+	if err != nil {
+		t.Fatalf("cannot create temporary dir for testing: %s\n", err)
+	}
+
+	dirList := []string{
```
mem, line 68: I think this needs entries with:

- blanks in them (spaces, tabs, new lines), both in the filenames and in the intermediate directories
- `*` in them
- `?` in them

mem (changes requested):

internal/pkg/build/files/files.go
```diff
@@ -0,0 +1,39 @@
...
+	"bytes"
+	"fmt"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+func expandPath(path string) ([]string, error) {
+	var output, stderr bytes.Buffer
+	cmd := exec.Command("/bin/sh", "-c", "echo "+path)
```
schebro: Will filepath.Glob() not work in this case? It resolves wildcards and returns files in a list of strings.

internal/pkg/build/files/copy.go
```diff
@@ -17,21 +17,40 @@ import (
...
 func Copy(src, dst string) error {
-	_, err := os.Stat(dst)
+	// resolve any bash globbing in filepath
+	paths, err := expandPath(src)
+	if err != nil {
+		return fmt.Errorf("while expanding source path with bash: %s: %s", src, err)
+	}
+
+	_, err = os.Stat(dst)
 	if os.IsNotExist(err) {
```
schebro: Is dst always going to be a directory? If dst is a file, os.Stat(dst) would return nil but copy can fail.

internal/pkg/build/files/files.go
ikaneshiro: Is a `Join` function something that _should_ clean the result in a particular way? If so, then I agree the name should be changed, but if not I feel the comment is enough to explain the particular behavior of this function since in all cases except those involving a trailing slash will behave the same way

internal/pkg/build/files/files.go
ikaneshiro: Unfortunately for this application it doesn't. All the globbing related go packages I have found do not handle hidden files in the same manner as most shells, which is how users will expect these wildcards to work.

internal/pkg/build/files/copy.go
ikaneshiro: No, dst can be a file. Can you describe the scenario where copy will fail if this is a file?

schebro: cp /dir/* file

ikaneshiro: I think this a reasonable failure as normally using `cp` in this manner should fail and I don't see a reason to do something like remove the file if it does exist and create a directory since it's likely that isn't the intention of the user

schebro: Agreed. As long as we send a failure it's fine. It's not right to delete the file and create a directory anyway.

schebro:
The main issue to address here is Marcelo's feedback regarding files with spaces. Otherwise, LGTM

internal/pkg/build/files/files.go
schebro: At least on Mac, I see that filepath.Glob() returns hidden files.

internal/pkg/build/files/files.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"bytes"
+	"fmt"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+const script = `for n in %[1]s ; do
```
WestleyK: Maybe you should use a more proper name for this?... Like:

```suggestion
const printFilesScript = `for n in %[1]s ; do
```
Or:
```suggestion
const filesScript = `for n in %[1]s ; do
```

Or whatever the script is used for...

WestleyK (approved):
Looks good to my knowledge. Just one comment... *optional*

internal/pkg/build/files/files.go
```diff
@@ -13,14 +13,14 @@ import (
 	"strings"
 )
 
-const script = `for n in %[1]s ; do
+const filenameExpansionScript = `for n in %[1]s ; do
```
WestleyK, line 16: :+1:

WestleyK (approved):

gvallee (approved):
I do not additional comments outside of the ones already made.

internal/pkg/build/files/files.go
mem: I'm thinking about the person reading the code: it's easy to see files.Join(...) and mistake it for filepath.Join(...). Also thinking of the case of someone writing new code and seeing that "file.Join" is used. They would think that for whatever reason, in this code that's how you join paths (instead of using filepath.Join).

I'm trying to think of a name for this operation and I cannot think of something descriptive enough. Basically when you ask it to join "/a" and "b/" it returns "/a/b/", but if you ask it to join "/a" and "b" it returns "/a/b". It's not quite the reverse of "dirname" because dirname removes the last element of a path, no matter what that element might be.

files.AddPrefix? And maybe change to `func AddPrefix(prefix, path string) string` (since all the uses use only two arguments)?

mem (approved):

internal/pkg/build/files/files.go
ikaneshiro: Makes more sense to me, will update

ikaneshiro:
Encountered a bug in this PR, will update when it is ready to merge

ikaneshiro:
Fixed

ikaneshiro:
Updated to address all comments

PR #3466, opened 2019-04-30
[E2E] Security E2E Tests

### LINKED ISSUE #3349: [E2E] port integration test  "security " to the e2e framework
##  E2E Task
Port 
- https://github.com/sylabs/singularity/blob/master/cmd/singularity/security_test.go
to the E2E framework
- https://github.com/sylabs/singularity/tree/master/e2e
under `e2e/security`

WestleyK:

This PR adds security to the new E2E testing framework.

- Fixes #3349 

<br>


ArangoGutierrez:
@WestleyK CircleCi is not happy :(

ArangoGutierrez:
*PING* @WestleyK

e2e/security/security.go
```diff
@@ -0,0 +1,292 @@
...
+		opts          e2e.ExecOpts
+		exit          int
+		expectSuccess bool
+	}{
+		// taget UID/GID
+		{
+			name:          "Set_uid",
+			image:         imagePath,
+			action:        "exec",
+			argv:          []string{"id", "-u", "|", "grep", "99"},
```
mem: can you please make this test tighter? This will match 99, 199, 991, etc.

Also, is it necessary to test this using the shell? Can you not just capture the output and compare it to the expected output using code in the test instead of relying on the shell?

e2e/security/security.go
```diff
@@ -0,0 +1,292 @@
...
+			opts:          e2e.ExecOpts{Security: []string{"uid:99"}},
+			exit:          1,
+			expectSuccess: false,
+		},
+		{
+			name:          "Set_gid",
+			image:         imagePath,
+			action:        "exec",
+			argv:          []string{"id", "-g", "|", "grep", "99"},
```
mem: same here

e2e/security/security.go
```diff
@@ -0,0 +1,292 @@
...
+			opts:          e2e.ExecOpts{Security: []string{"seccomp:./testdata/seccomp-profile.json"}},
+			exit:          0,
+			expectSuccess: true,
+		},
+		// capabilities
+		{
+			name:          "capabilities_keep_true",
+			image:         imagePath,
+			action:        "exec",
+			argv:          []string{"ping", "-c", "1", "8.8.8.8"},
```
mem, line 76: this is fine if there's no other alternative, but is there something else that can be used to test this? Going to the network and expecting it to work the same way 100% of the time is not realistic.

What are you testing here? that net_raw is not dropped? You should get the same by using the loopback address instead of Google's public DNS.

e2e/security/security.go
```diff
@@ -0,0 +1,292 @@
...
+			opts:          e2e.ExecOpts{KeepPrivs: true},
+			exit:          1,
```
mem, line 78: why do you expect 1 here?

e2e/security/security.go
```diff
@@ -0,0 +1,292 @@
...
+			expectSuccess: false,
+		},
+		{
+			name:          "capabilities_keep-false",
+			image:         imagePath,
+			action:        "exec",
+			argv:          []string{"ping", "-c", "1", "8.8.8.8"},
```
mem, line 83: same here

e2e/security/security.go
```diff
@@ -0,0 +1,292 @@
...
+			opts:          e2e.ExecOpts{KeepPrivs: false},
+			exit:          1,
+			expectSuccess: false,
+		},
+		{
+			name:          "capabilities_drop",
+			image:         imagePath,
+			action:        "exec",
+			argv:          []string{"ping", "-c", "1", "8.8.8.8"},
```
mem, line 90: and here

e2e/security/security.go
```diff
@@ -0,0 +1,292 @@
...
+			exit:          1,
+			expectSuccess: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run("unpriv "+tt.name, test.WithoutPrivilege(func(t *testing.T) {
+			_, stderr, exitCode, err := e2e.ImageExec(t, testenv.CmdPath, tt.action, tt.opts, tt.image, tt.argv)
+			if tt.expectSuccess && (exitCode != 0) {
+				t.Log(stderr, err, exitCode)
```
mem, line 114: this is the same in both branches

e2e/security/security.go
```diff
@@ -0,0 +1,292 @@
...
+		opts          e2e.ExecOpts
+		exit          int
+		expectSuccess bool
+	}{
+		// taget UID/GID
+		{
+			name:          "Set_uid",
+			image:         imagePath,
+			action:        "exec",
+			argv:          []string{"id", "-u", "|", "grep", "99"},
```
mem: same here

e2e/security/security.go
```diff
@@ -0,0 +1,292 @@
...
+			opts:          e2e.ExecOpts{Security: []string{"uid:99"}},
+			exit:          1,
+			expectSuccess: false,
+		},
+		{
+			name:          "Set_gid",
+			image:         imagePath,
+			action:        "exec",
+			argv:          []string{"id", "-g", "|", "grep", "99"},
```
mem: and here

e2e/security/security.go
```diff
@@ -0,0 +1,292 @@
...
+			opts:          e2e.ExecOpts{Security: []string{"seccomp:./testdata/seccomp-profile.json"}},
+			exit:          0,
+			expectSuccess: true,
+		},
+		// capabilities
+		{
+			name:          "capabilities_keep",
+			image:         imagePath,
+			action:        "exec",
+			argv:          []string{"ping", "-c", "1", "8.8.8.8"},
```
mem, line 171: and here

e2e/security/security.go
```diff
@@ -0,0 +1,292 @@
...
+			opts:          e2e.ExecOpts{KeepPrivs: true},
+			exit:          0,
+			expectSuccess: true,
+		},
+		{
+			name:          "capabilities_drop",
+			image:         imagePath,
+			action:        "exec",
+			argv:          []string{"ping", "-c", "1", "8.8.8.8"},
```
mem, line 178: and here

e2e/security/security.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+}
+
+func testSecurity(t *testing.T) {
+	test.EnsurePrivilege(t)
+	opts := e2e.BuildOpts{
+		Force:   true,
+		Sandbox: false,
+	}
+	if b, err := e2e.ImageBuild(testenv.CmdPath, opts, imagePath, "../../examples/busybox/Singularity"); err == nil {
```
mem: could this use the image provided by the e2e framework? You can make sure it exists by calling e2e.EnsureImage(t)

e2e/security/security.go
```diff
@@ -0,0 +1,292 @@
...
+
+	// Security
+	t.Run("Security_unpriv", testSecurityPriv)
+	t.Run("Security_priv", testSecurityUnpriv)
+	t.Run("Security_config_ownerships", testSecurityConfOwnership)
+
+}
+
+// pullTestContainer ...
+func pullTestContainer(t *testing.T) {
```
mem: there are pull tests? What's different about this?

e2e/security/security.go
```diff
@@ -0,0 +1,292 @@
...
+	b, err := cmd.CombinedOutput()
+
+	if err != nil {
+		t.Log(string(b))
+		t.Fatalf("Unable to pull test container: %s", err)
+	}
+}
+
+// MakeTmpDir will make a tmp dir and return a string of the path
+func makeTmpDir(t *testing.T) {
```
mem: this should probably go in a shared package?

mem (changes requested):

e2e/security/security.go
WestleyK: Agree, this should, but I was having trouble getting that to work...

e2e/security/security.go
WestleyK: This function just pulls a image to run all of these tests.

ArangoGutierrez:
@WestleyK *PING*

e2e/security/security.go
WestleyK: I figured it out! its in `e2e` package.

ArangoGutierrez:
@mem PR has been updated

e2e/internal/e2e/image.go
```diff
@@ -50,3 +53,28 @@ func EnsureImage(t *testing.T) {
...
+// path (with a container name 'test_container.sif').
+func MakeTmpDir(t *testing.T) (string, string) {
+	name, err := ioutil.TempDir("", "stest.")
+	if err != nil {
+		t.Fatalf("Failed to create temporary directory: %v", err)
+	}
+	if err := os.Chmod(name, 0777); err != nil {
+		t.Fatalf("Failed to chmod temporary directory: %v", err)
+	}
+	fullName := name
```
mem: why is this here? The function is creating a temporary directory. What's "test_container.sif"?

e2e/internal/e2e/image.go
```diff
@@ -50,3 +53,28 @@ func EnsureImage(t *testing.T) {
...
+	fullName += "test_container.sif"
+
+	return fullName, name
+}
+
+// PullTestAlpineContainer will pull the 'library://alpine:latest' container for tests.
+// This will pull to the pervided path ('imagePath'), and overide any image that was there.
+func PullTestAlpineContainer(cmdPath string, imagePath string) ([]byte, error) {
+	argv := []string{"pull", "-U", "--force", imagePath, "library://alpine:latest"}
```
mem: if you use "--allow-unsgined" the code is easier to read

e2e/security/security.go
```diff
@@ -3,14 +3,12 @@
 // LICENSE.md file distributed with the sources of this project regarding your
 // rights to use or distribute this software.
 
-// +build seccomp
+// TODO: --> +build seccomp
```
mem: what's the "TODO"?

e2e/internal/e2e/image.go
```diff
@@ -50,3 +53,28 @@ func EnsureImage(t *testing.T) {
 		t.Fatalf("Unexpected failure: %+v", err)
 	}
 }
+
+// MakeTmpDir will make a tmp dir and return a string of the path, and a full
+// path (with a container name 'test_container.sif').
+func MakeTmpDir(t *testing.T) (string, string) {
+	name, err := ioutil.TempDir("", "stest.")
```
mem: for e2e tests there's an environment variable that specifies which temporary directory to use for the test. All the temporary files should be created under that directory, so that it's easy to clean up.

e2e/security/security.go
mem: What I mean before is that when you call e2e.EnsuareImage() you are making sure that a specific test image exists.

Can you use that image instead of building another one?

If yes, please delete the call to e2e.ImageBuild.

If no, please add a comment as to why a specific image is required.

e2e/security/security.go
```diff
@@ -251,42 +250,22 @@ func testSecurity(t *testing.T) {
...
 func RunE2ETests(t *testing.T) {
 	err := envconfig.Process("E2E", &testenv)
 	if err != nil {
 		t.Fatal(err.Error())
 	}
 
-	t.Run("makeing_tmp_dir", makeTmpDir)
-	t.Run("pulling_test_contianer", pullTestContainer)
+	// Make a tmp directory
+	imagePath, _ = e2e.MakeTmpDir(t)
```
mem: A function called "MakeTmpDir" should not be returning the path to an image file. It should return the path to a temporary directory.

e2e/security/security.go
```diff
@@ -251,42 +250,22 @@ func testSecurity(t *testing.T) {
...
+
+	// pull a test image to that directory
+	b, err := e2e.PullTestAlpineContainer(testenv.CmdPath, imagePath)
+	if err != nil {
+		t.Log(string(b))
```
mem: It's probably a good idea to provide more context as to what happened here and what the test was trying to do.

mem (changes requested):

e2e/internal/e2e/image.go
WestleyK: Okay, I didn't know that. I'll fix this...

e2e/internal/e2e/image.go
WestleyK: This function did return a full container path eg. `/tmp/tests/test_container.sif`, and a tmp path eg. `/tmp/tests`. This defiantly needs to be improved.

e2e/security/security.go
WestleyK: This file originally had a comment:

```go
// +build seccomp
```

CCI did not like that, and when I removed that line, Circle CI was happy. So dose that comment do anything important?

e2e/security/security.go
```diff
@@ -0,0 +1,263 @@
...
+
+// RunE2ETests is the main func to trigger the test suite
+func RunE2ETests(t *testing.T) {
+	err := envconfig.Process("E2E", &testenv)
+	if err != nil {
+		t.Fatal(err.Error())
+	}
+
+	//	// Make a tmp directory
+	//	imagePath = e2e.MakeTmpDir(t)
```
ikaneshiro: These comments probably aren't needed anymore

e2e/security/security.go
```diff
@@ -0,0 +1,263 @@
...
+	//	imagePath += "test_container.sif"
+
+	imagePath = filepath.Join(testenv.TestDir, "test_container.sif")
```
ikaneshiro: I would recommend using `ioutil.TempFile()` to generate the file handle so that we do not risk a name collision with another container in `testenv.TestDir` made by another test.

e2e/suite.go
```diff
@@ -17,15 +17,27 @@ import (
 	"testing"
 
 	"github.com/sylabs/singularity/e2e/actions"
+
```
ikaneshiro, line 20: Why add all these whitespaces?

ikaneshiro (changes requested):
I also agree with @mem's comments about making the tests tighter so that it only accepts the value we are looking for. I think the points of not using the shell inside the container and using a loopback address instead of a real IP should be fixed but maybe we can can turn those into a tech debt issue if they are not fixed here

e2e/suite.go
WestleyK: Seeing if git will resolve the conflict automatically, theres always some conflict here. Same thing where the test are ran line 110++.

e2e/security/security.go
WestleyK: Not sure, I didn't write these tests. Maybe ask @ArangoGutierrez or @jscook2345.

e2e/internal/e2e/image.go
```diff
@@ -50,3 +52,12 @@ func EnsureImage(t *testing.T) {
 		t.Fatalf("Unexpected failure: %+v", err)
 	}
 }
+
+// PullTestAlpineContainer will pull the 'library://alpine:latest' container for tests.
+// This will pull to the pervided path ('imagePath'), and overide any image that was there.
+func PullTestAlpineContainer(cmdPath string, imagePath string) ([]byte, error) {
+	argv := []string{"pull", "--allow-unsigned", "--force", imagePath, "library://alpine:latest"}
```
mem, line 59: I don't think "latest" is good for a test. Prefer to use a hash (tags could be moved around).

Also, can you please document _why_ we need to use alpine and the image that is created for tests is not sufficient?

e2e/security/security.go
```diff
@@ -0,0 +1,292 @@
...
+		argv          []string
+		opts          e2e.ExecOpts
+		exit          int
+		expectSuccess bool
+	}{
+		// taget UID/GID
+		{
+			name:          "Set_uid",
+			image:         imagePath,
+			action:        "exec",
```
mem: it looks like "action" is always "exec". Maybe it's possible to remove this?

Same goes for "image", which seems to be set to "imagePath" in all cases.

e2e/security/security.go
mem: actually, it does not look like this is used. Am I missing something?

e2e/security/security.go
```diff
@@ -0,0 +1,292 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run("unpriv "+tt.name, test.WithoutPrivilege(func(t *testing.T) {
+			stdout, stderr, exitCode, err := e2e.ImageExec(t, testenv.CmdPath, tt.action, tt.opts, tt.image, tt.argv)
+
+			lines := bytes.Fields([]byte(stdout))
+
+			if tt.expectSuccess && (exitCode != 0) {
```
mem: Please consider [testing all cases](https://github.com/sylabs/singularity/wiki/End-to-end-testing#check-all-cases)

e2e/security/security.go
```diff
@@ -0,0 +1,292 @@
...
+				t.Log(stderr, err, exitCode)
+				t.Fatalf("unexpected failure running %q: %v", strings.Join(tt.argv, " "), err)
+			} else if !tt.expectSuccess && (exitCode != 1) {
+				t.Log(stderr, err, exitCode)
+				t.Fatalf("unexpected success running %q", strings.Join(tt.argv, " "))
+			}
+
+			//if len(lines) != 1 || string(lines[0]) != "99" {
```
mem: note how you are landing here if there was an expectation of failure and the code did in fact fail. In that case I would expect this code to fail. I'm guessing that the check len(lines) == 1 is guarding against two things: that the output does indeed have a line, and that in case of failure we don't try to perform the additional checks. That's two separate bits of logic bundled into a single check.

Maybe:

```
    if tt.expectSuccess && exitCode == 0 {
        if len(lines) != 1 {
            tt.Fatalf("Unexpected output ...")
        }
        if lines[0] != tt.expectID {
            // wrong ID?
            ...
        }
    } else if !tt.expectSuccess && exitCode != 0 {
        // expect failure, command failed
        // clean up, check output, ...
    }
```

e2e/security/security.go
```diff
@@ -0,0 +1,292 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run("priv "+tt.name, test.WithPrivilege(func(t *testing.T) {
+			stdout, stderr, exitCode, err := e2e.ImageExec(t, testenv.CmdPath, tt.action, tt.opts, tt.image, tt.argv)
+
+			lines := bytes.Fields([]byte(stdout))
+
+			if tt.expectSuccess && (exitCode != 0) {
```
mem: same here

e2e/security/security.go
```diff
@@ -0,0 +1,292 @@
...
+				t.Log(stderr, err, exitCode)
+				t.Fatalf("unexpected failure running '%v': %v", strings.Join(tt.argv, " "), err)
+			} else if !tt.expectSuccess && (exitCode != 1) {
+				t.Log(stderr, err, exitCode)
+				t.Fatalf("unexpected success running '%v'", strings.Join(tt.argv, " "))
+			}
+			//if len(lines) != 1 || string(lines[0]) != "99" {
+			if len(lines) == 1 && string(lines[0]) != tt.expectID {
```
mem: same here

e2e/security/security.go
```diff
@@ -0,0 +1,292 @@
...
+
+		}))
+	}
+}
+
+// testSecurityConfOwnership tests checks on config files ownerships
+func testSecurityConfOwnership(t *testing.T) {
+	configFile := buildcfg.SYSCONFDIR + "/singularity/singularity.conf"
+	// Change file ownership (do not try this at home)
+	err := os.Chown(configFile, 1001, 0)
```
mem, line 218: probably better to use "nobody" instead of 1001.

For example, in my specific case my UID is in fact 1001.

e2e/security/security.go
```diff
@@ -0,0 +1,292 @@
...
+
+// RunE2ETests is the main func to trigger the test suite
+func RunE2ETests(t *testing.T) {
+	err := envconfig.Process("E2E", &testenv)
+	if err != nil {
+		t.Fatal(err.Error())
+	}
+
+	// pull a test image to that directory
+	t.Run("non_root_config", test.WithoutPrivilege(func(t *testing.T) {
```
mem: this is not a test, this is set up, no need to run it as a subtest.

mem (changes requested):

e2e/internal/e2e/image.go
WestleyK: > image that is created for tests is not sufficient?

Where? That image should work instead.

sashayakovtseva: @WestleyK I think it is `TestEnv.ImagePath`

sashayakovtseva:
@WestleyK Will you update this PR?

WestleyK:
@sashayakovtseva, unlikely, this needs to be rewritten, because the e2e testing framework changed since I made this PR...

I left it open as a reminder to make a new one.

WestleyK:
Closing by: https://github.com/sylabs/singularity/pull/4233.

PR #3471, opened 2019-05-01
Slightly improved error message when pushing a image to the library that already exist

### LINKED ISSUE #3408: confusing output when pushing existing image to library
## Version of Singularity:

3.2.0rc2

```
[vagrant@localhost tmp]$ singularity push -U illolcow.sif library://jscook2345-double/default/illolcow.sif
WARNING: Skipping container verifying
INFO:    Image is already present in the library - not uploading.
INFO:    Setting tag latest
WARNING: latest replaces an existing tag
```

A few things:

* I would figure that this would upload the image no matter what, but it appears if the image exists it will not overwrite it.

* There is no `--force` option to overwrite, so perhaps there should be?

* Since the image is not being uploaded, why is there additional output after that message?

* I'm not sure INFO is right for saying 'we are not uploading your image'.  Perhaps just a message than exit?

WestleyK:

This PR will slightly improve the warning message when pushing a container to the library that already exist.

Example:

```bash
$ singularity push -U test.sif library://westleyk/foo/bar:latest,1.0
WARNING: Skipping container verifying
INFO:    Image already exist in the library
INFO:    Setting tag 1.0
```

- Fixes #3408 

<br>


pkg/client/library/push.go
```diff
@@ -95,7 +95,7 @@ func UploadImage(filePath string, libraryRef string, libraryURL string, authToke
 		}
 		sylog.Debugf("Upload completed OK\n")
 	} else {
-		sylog.Infof("Image is already present in the library - not uploading.\n")
+		return fmt.Errorf("image (%q) is already present in the library: not uploading", filePath)
```
jmstover: This isn't an error that should be called with `Errorf` should it? I mean, if you give it the same image, but a different tag, the image isn't uploaded, but the tag is created.

jmstover (changes requested):

pkg/client/library/push.go
WestleyK: Oh yeah, Let me see if I can fix this up... :+1:

WestleyK: @jmstover, Better now?

jmstover: Hrmm... This is where other folks need to chime in (@jscook2345 ??). I think a `Infof is perfectly acceptable here, and ... it isn't really a warning. It's just an informational message about the upload...

jmstover: And that was modified in between me starting to write this and actually finishing it... :P

jmstover (approved):

Issue #3478, opened 2019-05-02
go modules require a newer git than is available on centos6


jscook2345:
## Version of Singularity:

master

On centos6, `mconfig` currently fails because `go` is not able to find the `go modules` correctly for modules using `git tag`.

The solution seems to be to update the `git` version for the build environment. I don't think `git` is a hard requirement for the package itself.

Since our centos6 testing is for building an RPM, I think that this should be OK.

jscook2345:
@DrDaveD @mem @GodloveD @ikaneshiro @jmstover

DrDaveD:
How is this problem reproduced?  I just tried building the latest master branch on my centos6 VM and it built fine.

jscook2345:
@DrDaveD 

It happens in all of the travis builds, currently. All the PRs have the failing check I think.

What is your `git version` and `go version` output for centos6? Both myself and @GodloveD were able to reproduce this with `vagrant`, `virtualbox`, and a centos6.10 VM.

DrDaveD:
```
$ git version
git version 1.7.1
$ go version
go version go1.11.5 linux/amd64
$ rpm -q git
git-1.7.1-9.el6_9.x86_64
$ rpm -q golang
golang-1.11.5-1.el6.x86_64
```

jscook2345:
@mem made a good point that modules should not be downloaded at build-time and things should be using the vendor directory instead. His fix for this issues is here: #3481

PR #3481, opened 2019-05-02
Make sure Go modules are disabled when checking Go version

mem:
Since we are running a simple program to check that the Go version
meets the minimum requirements, we should not need any modules listed in
go.mod.

This takes care of a problem in CentOS 6 where, seemingly, the git
version that ships with it does not work well with the Go version that
ships with it _when modules are enabled_ (then the go tool tries to go
to the network to donwload modules).

The rest of the code should be fine as it's using the vendor directory
(by way of passing -mod=vendor to the go tool).


jscook2345:
Fixes #3478

jscook2345 (approved):
LGTM. Thanks @mem !

mem:
Merging this to unblock CI. The failures in Travis are unrelated to this.

PR #3482, opened 2019-05-02
Pass -mod=vendor to go list to avoid downloading modules

mem:
Since we have vendored copies of the Go modules' code, we need to pass
-mod=vendor almost everywhere where we call go to prevent go from trying
to download them to the module cache.

makeit/gengodep is missing that.


WestleyK (approved):
Looks good to me :shipit:

PR #3486, opened 2019-05-03
Script test framework

cclerget:

This framework basically used a custom shell interpreter running scripts from a Go test function, the shell interpreter can be customized with command/test builtins written in Go. As test builtins are tightly integrated with Go `testing` package, it allows to write e2e tests from shell script like this https://github.com/sylabs/singularity/blob/b60567929a16aa4efdf351aae3d8ad1838b64e71/e2e/scripts/actions/exec.test. Command builtin allow to add builtin command to enhance tests scripts to fit specific requirements (see https://github.com/sylabs/singularity/tree/b60567929a16aa4efdf351aae3d8ad1838b64e71/e2e/scripts_builtins/tools)


ArangoGutierrez:
+1

ArangoGutierrez (commented):
Iteration 1 sounds great.
let's move it to the e2e folder and trigger a review as iteration 2

jscook2345 (changes requested):
For the time being, please do not delete the tests that are currently in place (ex: `actions/actions.go`)

e2e/e2e_scripts_test.go
```diff
@@ -0,0 +1,135 @@
+// Copyright (c) 2019, Sylabs Inc. All rights reserved.
```
jscook2345: I think you can remove the `e2e` part of the filename since it's part of the `e2e` package?

e2e/e2e_scripts_test.go
```diff
@@ -0,0 +1,135 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/sylabs/singularity/internal/pkg/buildcfg"
```
jscook2345: The level of testing we're doing here should avoid importing any packages that are a part of `singularity` (especially internal ones). We really only want to test the executable and any other build artifacts that are installed. We want to avoid knowing anything about the internals of how `singularity` works, ie it should be considered a black box.

Examples: `buildcfg`, `sylog`, `user-agent`

e2e/e2e_scripts_test.go
```diff
@@ -0,0 +1,135 @@
...
+	// custom builtins
+	_ "github.com/sylabs/singularity/e2e/scripts_builtins/command"
+	_ "github.com/sylabs/singularity/e2e/scripts_builtins/test"
+)
+
+var testScripts = []struct {
+	name string
+	path string
+}{
+	{"BUILD/BASIC", "build/basic.test"},
```
jscook2345: As a part of our testing best practices (which we're working on), it would be good to be more verbose here with this struct.

Example: `{ name: "ACTIONS/ENV_PATH", path: "actions/env_path.test" },`

e2e/e2e_scripts_test.go
```diff
@@ -0,0 +1,135 @@
...
+	// maintain sudo session for use in test scripts without
+	// password ask
+	go func() {
+		time.Sleep(1 * time.Minute)
+		if err := sudoExec(sudo, []string{"true"}); err != nil {
+			sylog.Fatalf("%s", err)
+		}
+	}()
+
+	testDir, err := ioutil.TempDir("", "stest-")
```
jscook2345: Could we move the temporary directory handling to a function, and have the `stest-` in a `const` value?

e2e/e2e_scripts_test.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		sylog.Fatalf("%s", err)
+	}
+	// use sudo here to remove test directory in order to
+	// delete image/files/directories that could be created
+	// by privileged run/tests
+	defer sudoExec(sudo, []string{"rm", "-rf", testDir})
+
+	cacheDirPriv := filepath.Join(testDir, "priv")
```
jscook2345: Could we move the cache handling to a function and have the names in `const` values as well?

e2e/e2e_scripts_test.go
```diff
@@ -0,0 +1,135 @@
...
+		sylog.Fatalf("failed to create %s: %s", cacheDirPriv, err)
+	}
+	if err := os.Mkdir(cacheDirUnpriv, 0755); err != nil {
+		sylog.Fatalf("failed to create %s: %s", cacheDirUnpriv, err)
+	}
+
+	sourceDir := filepath.Dir(buildcfg.BUILDDIR)
+	envPath := os.Getenv("PATH")
+
+	fmt.Println("Available environment variable in test script:")
```
jscook2345: These `Println` calls about the environment look like they should be verbose output, rather than standard test output. What do you think?

e2e/e2e_scripts_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+	os.Setenv("CACHEDIR_PRIV", cacheDirPriv)
+	fmt.Printf("CACHEDIR_PRIV: %s\n", cacheDirPriv)
+
+	os.Setenv("SOURCEDIR", sourceDir)
+	fmt.Printf("SOURCEDIR: %s\n", sourceDir)
+
+	fmt.Printf("PATH: %s\n", envPath)
+
+	if err := os.Chdir("scripts"); err != nil {
```
jscook2345: Should this be a part of `stest.RunScripts` ? Maybe give the scripts base path as a dependency which could be injected?

e2e/scripts/actions/env_path.test
```diff
@@ -0,0 +1,30 @@
...
+
+default_image="docker://alpine:3.8"
+default_path="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
+
+custom_image="docker://godlovedc/lolcow"
+custom_path="/usr/games:${default_path}"
+partial_path="/foo"
+overwritten_path="/usr/bin:/bin"
+
+stdout-contains "DefaultPath"           "${default_path}"                   singularity exec ${default_image} printenv PATH
```
jscook2345, line 14: How do you test `stdout` and `stderr` at the same time? This is something that is not uncommon to do.

e2e/scripts/actions/env_path.test
```diff
@@ -0,0 +1,30 @@
+#!/singularity/test
+# env path tests
+
+source common.func
+
+default_image="docker://alpine:3.8"
```
jscook2345, line 6: I think these defaults and custom options should be inputs to the script/test so that we can test many different containers here. A table-driven approach with a loop would be how we would do it in `go`. How would we accomplish that here?

e2e/scripts/actions/exec.test
```diff
@@ -0,0 +1,54 @@
+#!/singularity/test
+# exec tests
+
+source common.func
+
+# workdir
```
jscook2345, line 6: Creating these temporary directories looks like they could be another built-in that does a bit more of this work for you.

e2e/scripts/actions/from_uri.test
```diff
@@ -0,0 +1,38 @@
...
+
+# Run from supported URI's and check the runscript call works
+expect-success "RunFromDockerOK"    singularity run ${option} docker://busybox:latest ${size}
+expect-success "RunFromLibraryOK"   singularity run ${option} library://busybox:latest ${size}
+expect-success "RunFromShubOK"      singularity run ${option} shub://singularityhub/busybox ${size}
+expect-failure "RunFromDockerKO"    singularity run ${option} docker://busybox:latest 0
+expect-failure "RunFromLibraryKO"   singularity run ${option} library://busybox:latest 0
+expect-failure "RunFromShubKO"      singularity run ${option} shub://singularityhub/busybox 0
+
+# exec from a supported URI's and check the exit code
```
jscook2345, line 20: It would be important to check the output and the success from some of these commands

e2e/scripts/actions/from_uri.test
```diff
@@ -0,0 +1,38 @@
...
+expect-success "TrueDocker"         singularity exec docker://busybox:latest true
+expect-success "TrueLibrary"    	singularity exec library://busybox:latest true
+expect-success "TrueShub"   		singularity exec shub://singularityhub/busybox true
+expect-failure "FalseDocker"    	singularity exec docker://busybox:latest false
+expect-failure "Falselibrary"   	singularity exec library://busybox:latest false
+expect-failure "FalseShub"      	singularity exec shub://singularityhub/busybox false
+
+# exec from URI with user namespace enabled
+if check-namespace "user"; then
```
jscook2345, line 29: I like this! Having functions to see if certain features are enabled is great! I was thinking of using build tags, but using this and either skipping a test or conditionally executing it are good ideas!

e2e/scripts/actions/persistent_overlay.test
```diff
@@ -0,0 +1,49 @@
+#!/singularity/test
+# persistent overlay tests
+
+source common.func
+
+# check overlay is supported, or skip those tests
+check-overlayfs
+
+# check command presence
```
jscook2345, line 9: Another great check for tests. Verify commands which we use are actually available instead of assuming they are! Great!

e2e/scripts/actions/persistent_overlay.test
```diff
@@ -0,0 +1,49 @@
...
+check-command mkfs.ext3
+
+# test staging files/directories
+ext3_overlay_image=`mktmpfile`
+overlaydir=`mktmpdir`
+squashdir=`mktmpdir`
+squash_overlay_img=`mktmpfile`
+echo "bogus" > ${squashdir}/squash_overlay
+
+# create overlay images
```
jscook2345, line 21: I'd really hope to see the output of these commands if we used a verbose flag

e2e/scripts/actions/env_path.test
jscook2345, line 14: what does a failing test look like? do these `testing` functions have verbose output?

e2e/scripts/actions/persistent_overlay.test
```diff
@@ -0,0 +1,49 @@
...
+run-command mksquashfs ${squashdir} ${squash_overlay_img} -noappend
+run-command dd if=/dev/zero of=${ext3_overlay_image} bs=1M count=64
+run-command mkfs.ext3 -q -F ${ext3_overlay_image}
+
+# get test image
+image=`get-test-image`
+
+# test directory overlay
+expect-success "TouchWithDirOverlay"        sudo singularity exec -o ${overlaydir} ${image} touch /dir_overlay
```
jscook2345, line 30: I think we should have a check to see if a user can actually elevate their privileges before running `sudo` and potentially skip if they cannot. This way we can easily test both privileged and non-privileged users.

It would also be good if we could run these as regular users to verify they fail without having to repeat the command without `sudo` attached.

e2e/scripts/build/basic.test
```diff
@@ -0,0 +1,41 @@
+#!/singularity/test
+# build tests
+
+source common.func
+
+buildImage() {
```
jscook2345: I think this would be good in a common library

e2e/scripts/common.func
```diff
@@ -0,0 +1,178 @@
+# generic functions
+
+fatal() {
+    test-error "$1"
+    exit 1
+}
+
+# test if singularity binary is present in $PATH
```
jscook2345: Could you use your `check-command` built-in here?

e2e/scripts/common.func
```diff
@@ -0,0 +1,178 @@
...
+mktmpdir() {
+    create-tmpdir ${TESTDIR}
+}
+
+mktmpfile() {
+    create-tmpfile ${TESTDIR}
+}
+
+# test image
+get-test-image() {
```
jscook2345, line 255: There are a lot of ways to get an image for testing, perhaps this should be something like `build-test-image-from-definition`?

e2e/scripts/common.func
```diff
@@ -0,0 +1,178 @@
...
+
+run-command() {
+    $@ >${TESTDIR}/run-command.out 2>&1
+    if [ $? != 0 ]; then
+        output=`cat ${TESTDIR}/run-command.out`
+        fatal "$@ failed: ${output}"
+    fi
+}
+
+check-kernel-version() {
```
jscook2345, line 377: Using the kernel version for a check on a feature usually doesn't work since a lot of distros add patches to the kernel out of order. I've learned the hard way relying on the kernel version is usually a problem. Best to check if something works instead.

e2e/e2e_scripts_test.go
cclerget: yep

e2e/e2e_scripts_test.go
cclerget: yep

e2e/e2e_scripts_test.go
cclerget: Agreed, will had `t.Verbose` check

e2e/scripts/actions/env_path.test
cclerget: Will enhance this test builtin to allow use of "combined" stream argument

e2e/e2e_scripts_test.go
cclerget: well as `stest` package is public, I don't want to force this, that's why it's done here

e2e/scripts/actions/env_path.test
cclerget: Well I just ported current tests as it were written in Go. Here we could use same approach with a bash array

e2e/scripts/actions/exec.test
cclerget: `mktmpdir/mktmpfile` are already function wrappers around `create-tmpdir/create-tmpfile` builtins

e2e/scripts/actions/from_uri.test
cclerget: Will look

e2e/scripts/actions/env_path.test
cclerget: When it fails : https://github.com/sylabs/singularity/pull/3486/files/e714d276a7150107570c967490ca95b1e8bca17c#diff-8281e19ba9c3129e27bece1c9222ad0bR84

e2e/scripts/actions/persistent_overlay.test
cclerget: This is not required because the whole test suite run sudo at the very beginning (and keep running a sudo true every minute to keep user sudo session in case of long running test), so if user doesn't have right it could not even run tests, like it's done for every `go test` we have with make.

e2e/scripts/common.func
cclerget: no because if singularity is not found, it's considered as fatal, `check-command` just skip the current test script

e2e/scripts/common.func
cclerget: this is the equivalent of `e2e.EnsureImage` from e2e package, I could rename to `ensure-test-image` ?

e2e/scripts/common.func
cclerget: Agreed

e2e/scripts/docker/definition.test
```diff
@@ -0,0 +1,30 @@
+#!/singularity/test
```
jscook2345, line 1: This test is a good example of where a loop / table-driven test would be great to have. Having to copy and paste all the test code just for different examples seems a little problematic.

jscook2345 (changes requested):
I think the test data should still follow the 'go' best practice of having a `testdata` directory. I would put things like definition files there.

e2e/scripts/docker/registry.test
```diff
@@ -0,0 +1,32 @@
+#!/singularity/test
+# test from docker registry
+
+source common.func
+
+check-command docker
+
+atexit() {
```
jscook2345: I didn't see this defined elsewhere, did I miss something?

jscook2345: Yes I did! just further down!

jscook2345 (commented):
One of the tasks we were given as a part of this framework was to easily identify and log all the singularity commands and options that were tested so that we can analyze what is and is not tested.

Given that there is no single point where execution happens and that a lot of the execution is done with variables, how would we accomplish this?

e2e/scripts/oci/attach.test
```diff
@@ -0,0 +1,31 @@
...
+
+expect-exit 0       "OciAttachCreate"           sudo singularity oci create -b ${bundle} ${containerID}
+check-container-state ${containerID} "created"
+
+expect-exit 0       "OciAttachStart"            sudo singularity oci start ${containerID}
+check-container-state ${containerID} "running"
+
+start-interactive   "OciAttachInteractive"      sudo singularity oci attach ${containerID}
+interactive-sendline    "hostname"
+interactive-expect      "runc"
```
jscook2345: Should this be something else besides `runc`?

e2e/scripts/oci/basic.test
```diff
@@ -0,0 +1,45 @@
...
+# override generated configuration with the default one, need terminal set to true
+sudo cp oci/testdata/config.json ${bundle}/config.json
+
+expect-exit 0   "OciCreate"                 sudo singularity oci create -b ${bundle} ${containerID}
+check-container-state ${containerID} "created"
+
+expect-exit 0   "OciStart"                  sudo singularity oci start ${containerID}
+check-container-state ${containerID} "running"
+
+if [ -d /sys/fs/cgroup/freezer ]; then
```
jscook2345, line 22: Looks like a good option for a built-in check similar to `check-overlay`

jscook2345 (commented):
We are losing a lot of the functionality that the basic `go test` framework provided us. Test discovery, filtering, running individual tests, running tests for individual packages, attempts at coverage, failfast, etc.

How are we going to get back some of those features that we were getting for free?

cclerget:
> One of the tasks we were given as a part of this framework was to easily identify and log all the singularity commands and options that were tested so that we can analyze what is and is not tested.
> 
> Given that there is no single point where execution happens and that a lot of the execution is done with variables, how would we accomplish this?

We can log everything if needed, I will apply few changes both log and command output/error logs.

> We are losing a lot of the functionality that the basic `go test` framework provided us. Test discovery, filtering, running individual tests, running tests for individual packages, attempts at coverage, failfast, etc.
> 
> How are we going to get back some of those features that we were getting for free?

We don't lose it, by example if you want to run all tests in `build/basic.test` based on this line : `{name: "BUILD/BASIC", path: "build/basic.test"},` you can run them with

```
go test -run TestE2EScripts/BUILD/BASIC -v -tags e2e_scripts ./e2e
```

And if you want to run Busybox test build only based on line
`buildImage "image"   "Busybox"        ""            "${SOURCEDIR}/examples/busybox/Singularity"` in `build/basic.test`

you can run it with `go test -run TestE2EScripts/BUILD/BASIC/Busybox -v -tags e2e_scripts ./e2e`

PR #3497, opened 2019-05-06
Add content tests for the oci help commands

jscook2345:
Redo of PR adding content tests for the oci help commands

### DIFF
--- a/e2e/help/help.go
+++ b/e2e/help/help.go
@@ -1,5 +1,4 @@
 // Copyright (c) 2019, Sylabs Inc. All rights reserved.
-
 // This software is licensed under a 3-clause BSD license. Please consult the
 // LICENSE.md file distributed with the sources of this project regarding your
 // rights to use or distribute this software.
--- a/e2e/help/help.go
+++ b/e2e/help/help.go
@@ -1,4 +1,5 @@
 // Copyright (c) 2019, Sylabs Inc. All rights reserved.
+
 // This software is licensed under a 3-clause BSD license. Please consult the
 // LICENSE.md file distributed with the sources of this project regarding your
 // rights to use or distribute this software.
@@ -9,12 +10,16 @@
 package help
 
 import (
+	"fmt"
 	"os/exec"
+	"path/filepath"
 	"strings"
 	"testing"
 
 	"github.com/sylabs/singularity/e2e/internal/e2e"
 	"github.com/sylabs/singularity/internal/pkg/test"
+	"gotest.tools/assert"
+	"gotest.tools/golden"
 )
 
 type testingEnv struct {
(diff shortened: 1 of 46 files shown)


internal/pkg/test/cmd.go
```diff
@@ -0,0 +1,36 @@
...
+type Cmd struct {
+	path string
+}
+
+type Result struct {
+	*icmd.Result
+}
+
+func NewCmd(path string) *Cmd {
+	c := &Cmd{
```
mem: this can be return `&Cmd{ path: path }`, no need to create the variable to return it on the next line.

internal/pkg/test/cmd.go
```diff
@@ -0,0 +1,36 @@
...
+		path: path,
+	}
+
+	return c
+}
+
+func (c *Cmd) Run(t *testing.T, args ...string) *Result {
+	t.Logf("Running cmd %s with args %s", c.path, args)
```
mem: We can use some quotes here, just to make it clear what's running and how it's running.

What I'm thinking is that if you are going to run something like `singularity foo --bar=baz xyz` you want to be able to differentiate flag `--bar=baz` with extra argument `xyz` from flag `--bar=baz xyz`.

The quotes around c.path are trivial, just add them to the format string.

The quotes around the arguments...

A very basic way of handling this is something like:

```
type stringQuoter []string

func (sq stringQuoter) Format(f State, c rune) {
    sep := []byte{}
    commaBlank := []byte{',', ' '}
    for _, s := range sq {
        f.Write(sep)
        f.Write([]byte(`"`))
        f.Write([]byte(s))
        f.Write([]byte(`"`))
        sep = commaBlank
    }
}
```

and then you wrap `args` in that type when passing it to `t.Logf`.

(and I'm sure that code can be much much better, I'm just providing it as a general idea of how this could be done)

mem (changes requested):
LGTM, just a couple of minor changes

e2e/help/help.go
```diff
@@ -1,4 +1,5 @@
 // Copyright (c) 2019, Sylabs Inc. All rights reserved.
+
```
ArangoGutierrez:
```suggestion
```

internal/pkg/test/cmd.go
ArangoGutierrez: +1

ArangoGutierrez (commented):
LGTM -> @mem

internal/pkg/test/cmd.go
jscook2345: Done. I thought I would add more arguments but perhaps that will be later.

e2e/help/help.go
jscook2345: Done!

internal/pkg/test/cmd.go
jscook2345: To quote arguments, I switched from `%s` to `%q`.

Does this seem OK to you?

`cmd.go:27: Running cmd "/usr/local/singularity/bin/singularity" with args ["help" "oci" "pause"]`

mem: That's good enough for this

mem (approved):

PR #3504, opened 2019-05-06
Fix printing of app labels

ikaneshiro:

- Fixes issue where app labels are not printed when `--app appname` is specified

### DIFF
--- a/cmd/internal/cli/inspect_linux.go
+++ b/cmd/internal/cli/inspect_linux.go
@@ -39,7 +39,7 @@ var (
 	environment bool
 	helpfile    bool
 	jsonfmt     bool
-	apps        bool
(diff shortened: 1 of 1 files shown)


cmd/internal/cli/inspect_linux.go
```diff
@@ -60,7 +60,7 @@ type inspectFormat struct {
 // -d|--deffile
 var inspectAppsListFlag = cmdline.Flag{
 	ID:           "inspectAppsListFlag",
-	Value:        &apps,
+	Value:        &listApps,
```
WestleyK, line 63: Thanks for fixing that!

WestleyK (approved):
Looks good to me :+1:

Issue #3308, opened 2019-04-15
Improve test coverage of internal/pkg/client


gvallee:
## Version of Singularity:

Improve test coverage of internal/pkg/client.

### Expected behavior

Get close to 100% coverage.

### Actual behavior

`
$ go test -coverprofile=coverage.out && go tool cover -func=coverage.out
PASS
coverage: 37.0% of statements
ok  	github.com/sylabs/singularity/internal/pkg/client/cache	0.006s
github.com/sylabs/singularity/internal/pkg/client/cache/dir.go:38:	Root			100.0%
github.com/sylabs/singularity/internal/pkg/client/cache/dir.go:45:	Clean			75.0%
github.com/sylabs/singularity/internal/pkg/client/cache/dir.go:55:	updateCacheRoot		75.0%
github.com/sylabs/singularity/internal/pkg/client/cache/dir.go:72:	updateCacheSubdir	75.0%
github.com/sylabs/singularity/internal/pkg/client/cache/dir.go:88:	initCacheDir		71.4%
github.com/sylabs/singularity/internal/pkg/client/cache/library.go:22:	Library			100.0%
github.com/sylabs/singularity/internal/pkg/client/cache/library.go:27:	LibraryImage		0.0%
github.com/sylabs/singularity/internal/pkg/client/cache/library.go:34:	LibraryImageExists	0.0%
github.com/sylabs/singularity/internal/pkg/client/cache/net.go:19:	Net			100.0%
github.com/sylabs/singularity/internal/pkg/client/cache/net.go:24:	NetImage		0.0%
github.com/sylabs/singularity/internal/pkg/client/cache/net.go:31:	NetImageExists		0.0%
github.com/sylabs/singularity/internal/pkg/client/cache/oci.go:21:	OciBlob			100.0%
github.com/sylabs/singularity/internal/pkg/client/cache/oci.go:26:	OciTemp			100.0%
github.com/sylabs/singularity/internal/pkg/client/cache/oci.go:31:	OciTempImage		0.0%
github.com/sylabs/singularity/internal/pkg/client/cache/oci.go:38:	OciTempExists		0.0%
github.com/sylabs/singularity/internal/pkg/client/cache/shub.go:19:	Shub			100.0%
github.com/sylabs/singularity/internal/pkg/client/cache/shub.go:24:	ShubImage		0.0%
github.com/sylabs/singularity/internal/pkg/client/cache/shub.go:31:	ShubImageExists		0.0%
[...]

PR #3511, opened 2019-05-07
Revisit the cache implementation to improve testing

gvallee:

Revisit the cache implementation with the following goals:
- improve testing coverage (now at 90.6%),
- make it easier to write tests involving caches, including error cases, making it easier to have good testing coverage,
- avoid the case where the configuration of cache could change during the execution of a command by modifying the appropriate environment variable,
- avoid deleting user data when the user misunderstand the semantic of the environment variable to specify the location of a custom cache.

- Addresses #3308 

### DIFF
--- a/cmd/internal/cli/actions.go
+++ b/cmd/internal/cli/actions.go
@@ -58,9 +58,16 @@ func handleOCI(cmd *cobra.Command, u string) (string, error) {
 	}
 
 	name := uri.GetName(u)
-	imgabs := cache.OciTempImage(sum, name)
+	c, err := cache.Create()
+	if c == nil || err != nil {
+		return "", fmt.Errorf("unable to create cache object")
+	}
+	imgabs, err := c.OciTempImage(sum, name)
+	if err != nil {
+		return "", fmt.Errorf("unable to get image's data from OCI temp")
+	}
 
-	if exists, err := cache.OciTempExists(sum, name); err != nil {
+	if exists, err := c.OciTempExists(sum, name); err != nil {
 		return "", fmt.Errorf("unable to check if %v exists: %v", imgabs, err)
 	} else if !exists {
 		sylog.Infof("Converting OCI blobs to SIF format")
@@ -98,9 +105,18 @@ func handleLibrary(u, libraryURL string) (string, error) {
 	}
 
 	imageName := uri.GetName(u)
-	imagePath := cache.LibraryImage(libraryImage.Hash, imageName)
 
-	if exists, err := cache.LibraryImageExists(libraryImage.Hash, imageName); err != nil {
+	c, err := cache.Create()
+	if c == nil || err != nil {
+		return "", fmt.Errorf("unable to create cache object")
+	}
+
+	imagePath, err := c.LibraryImage(libraryImage.Hash, imageName)
+	if err != nil {
+		return "", fmt.Errorf("unable to get image's information from library cache")
+	}
+
+	if exists, err := c.LibraryImageExists(libraryImage.Hash, imageName); err != nil {
 		return "", fmt.Errorf("unable to check if %v exists: %v", imagePath, err)
 	} else if !exists {
 		sylog.Infof("Downloading library image")
@@ -120,9 +136,18 @@ func handleLibrary(u, libraryURL string) (string, error) {
 
 func handleShub(u string) (string, error) {
 	imageName := uri.GetName(u)
-	imagePath := cache.ShubImage("hash", imageName)
 
-	exists, err := cache.ShubImageExists("hash", imageName)
+	c, err := cache.Create()
+	if c == nil || err != nil {
+		return "", fmt.Errorf("unable to create cache object")
+	}
+
+	imagePath, err := c.ShubImage("hash", imageName)
+	if err != nil {
+		return "", fmt.Errorf("unable to retrieve image data from shub cache")
+	}
+
+	exists, err := c.ShubImageExists("hash", imageName)
 	if err != nil {
 		return "", fmt.Errorf("unable to check if %v exists: %v", imagePath, err)
 	}
@@ -142,9 +167,17 @@ func handleShub(u string) (string, error) {
 func handleNet(u string) (string, error) {
 	refParts := strings.Split(u, "/")
 	imageName := refParts[len(refParts)-1]
-	imagePath := cache.NetImage("hash", imageName)
 
-	exists, err := cache.NetImageExists("hash", imageName)
+	c, err := cache.Create()
+	if c == nil || err != nil {
+		return "", fmt.Errorf("unable to create cache object")
+	}
+	imagePath, err := c.NetImage("hash", imageName)
+	if err != nil {
+		return "", fmt.Errorf("unable to get image's data from net cache")
+	}
+
+	exists, err := c.NetImageExists("hash", imageName)
 	if err != nil {
 		return "", fmt.Errorf("unable to check if %v exists: %v", imagePath, err)
 	}
--- a/cmd/internal/cli/actions.go
+++ b/cmd/internal/cli/actions.go
@@ -60,7 +60,7 @@ func handleOCI(cmd *cobra.Command, u string) (string, error) {
 	name := uri.GetName(u)
 	// Create a cache handle, which will provide access to an existing cache
 	// or create a new cache based on the current configuration.
-	c, err := cache.Create()
+	c, err := cache.NewHandle()
 	if c == nil || err != nil {
 		return "", fmt.Errorf("unable to create cache object")
 	}
@@ -110,7 +110,7 @@ func handleLibrary(u, libraryURL string) (string, error) {
 
 	// Create a cache handle, which will provide access to an existing cache
 	// or create a new cache based on the current configuration.
-	c, err := cache.Create()
+	c, err := cache.NewHandle()
 	if c == nil || err != nil {
 		return "", fmt.Errorf("unable to create cache object")
 	}
@@ -143,7 +143,7 @@ func handleShub(u string) (string, error) {
 
 	// Create a cache handle, which will provide access to an existing cache
 	// or create a new cache based on the current configuration.
-	c, err := cache.Create()
+	c, err := cache.NewHandle()
 	if c == nil || err != nil {
 		return "", fmt.Errorf("unable to create cache object")
 	}
@@ -176,7 +176,7 @@ func handleNet(u string) (string, error) {
 
 	// Create a cache handle, which will provide access to an existing cache
 	// or create a new cache based on the current configuration.
-	c, err := cache.Create()
+	c, err := cache.NewHandle()
 	if c == nil || err != nil {
 		return "", fmt.Errorf("unable to create cache object")
 	}
--- a/cmd/internal/cli/actions.go
+++ b/cmd/internal/cli/actions.go
@@ -58,6 +58,8 @@ func handleOCI(cmd *cobra.Command, u string) (string, error) {
 	}
 
 	name := uri.GetName(u)
+	// Create a cache handle, which will provide access to an existing cache
+	// or create a new cache based on the current configuration.
 	c, err := cache.Create()
 	if c == nil || err != nil {
 		return "", fmt.Errorf("unable to create cache object")
@@ -106,6 +108,8 @@ func handleLibrary(u, libraryURL string) (string, error) {
 
 	imageName := uri.GetName(u)
 
+	// Create a cache handle, which will provide access to an existing cache
+	// or create a new cache based on the current configuration.
 	c, err := cache.Create()
 	if c == nil || err != nil {
 		return "", fmt.Errorf("unable to create cache object")
@@ -137,6 +141,8 @@ func handleLibrary(u, libraryURL string) (string, error) {
 func handleShub(u string) (string, error) {
 	imageName := uri.GetName(u)
 
+	// Create a cache handle, which will provide access to an existing cache
+	// or create a new cache based on the current configuration.
 	c, err := cache.Create()
 	if c == nil || err != nil {
 		return "", fmt.Errorf("unable to create cache object")
@@ -168,6 +174,8 @@ func handleNet(u string) (string, error) {
 	refParts := strings.Split(u, "/")
 	imageName := refParts[len(refParts)-1]
 
+	// Create a cache handle, which will provide access to an existing cache
+	// or create a new cache based on the current configuration.
 	c, err := cache.Create()
 	if c == nil || err != nil {
 		return "", fmt.Errorf("unable to create cache object")
--- a/internal/app/singularity/cache_clean_linux.go
+++ b/internal/app/singularity/cache_clean_linux.go
@@ -16,9 +16,14 @@ import (
 )
 
 func cleanLibraryCache() error {
-	sylog.Debugf("Removing: %v", cache.Library())
+	c, err := cache.Create()
+	if c == nil || err != nil {
+		return fmt.Errorf("unable to create the cache object")
+	}
 
-	err := os.RemoveAll(cache.Library())
+	sylog.Debugf("Removing: %v", c.Library)
+
+	err = os.RemoveAll(c.Library)
 	if err != nil {
 		return fmt.Errorf("unable to clean library cache: %v", err)
 	}
@@ -27,9 +32,14 @@ func cleanLibraryCache() error {
 }
 
 func cleanOciCache() error {
-	sylog.Debugf("Removing: %v", cache.OciTemp())
+	c, err := cache.Create()
+	if c == nil || err != nil {
+		return fmt.Errorf("unable to create the cache object")
+	}
 
-	err := os.RemoveAll(cache.OciTemp())
+	sylog.Debugf("Removing: %v", c.OciTemp)
+
+	err = os.RemoveAll(c.OciTemp)
 	if err != nil {
 		return fmt.Errorf("unable to clean oci-tmp cache: %v", err)
 	}
@@ -38,9 +48,14 @@ func cleanOciCache() error {
 }
 
 func cleanBlobCache() error {
-	sylog.Debugf("Removing: %v", cache.OciBlob())
+	c, err := cache.Create()
+	if c == nil || err != nil {
+		return fmt.Errorf("unable to create the cache object")
+	}
 
-	err := os.RemoveAll(cache.OciBlob())
+	sylog.Debugf("Removing: %v", c.OciTemp)
+
+	err = os.RemoveAll(c.OciTemp)
 	if err != nil {
 		return fmt.Errorf("unable to clean oci-blob cache: %v", err)
 	}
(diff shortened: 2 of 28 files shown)


internal/app/singularity/cache_clean_linux.go
```diff
@@ -51,6 +66,11 @@ func cleanBlobCache() error {
 
 // CleanCache : clean a type of cache (cacheType string). will return a error if one occurs.
```
WestleyK: Can you fix this silly godoc comment thing? I know it was me that wrote this a while ago...

```suggestion
// CleanCache will clean a type of cache (cacheType). The possible options are: 'library', 'oci', 'blob', and 'all'.
Will return a err if one occurs.
```

internal/app/singularity/cache_clean_linux.go
```diff
@@ -51,6 +66,11 @@ func cleanBlobCache() error {
...
 func CleanCache(cacheType string) error {
+	c, err := cache.Create()
```
WestleyK: Why does this need to create cache here?

internal/app/singularity/cache_list_linux.go
```diff
@@ -97,26 +102,31 @@ func listBlobCache(printList bool) (int, int64, error) {
 	count := 0
 	var totalSize int64
 
-	_, err := os.Stat(filepath.Join(cache.OciBlob(), "/blobs"))
+	c, err := cache.Create()
+	if c == nil || err != nil {
+		return 0, 0, nil
```
WestleyK: Why does this return a nil?

internal/pkg/build/sources/conveyorPacker_library.go
```diff
@@ -53,11 +53,19 @@ func (cp *LibraryConveyorPacker) Get(b *types.Bundle) (err error) {
...
 
 	imageName := uri.GetName(libURI)
-	imagePath := cache.LibraryImage(libraryImage.Hash, imageName)
+	c, err := cache.Create()
+	if c == nil || err != nil {
+		return fmt.Errorf("Failed to create cache object")
+	}
+	imagePath, err := c.LibraryImage(libraryImage.Hash, imageName)
+	if err != nil {
+		return fmt.Errorf("Unable to get path to image")
```
WestleyK:
```suggestion
		return fmt.Errorf("unable to get path to image")
```

internal/pkg/build/sources/conveyorPacker_library.go
```diff
@@ -53,11 +53,19 @@ func (cp *LibraryConveyorPacker) Get(b *types.Bundle) (err error) {
 	}
 
 	imageName := uri.GetName(libURI)
-	imagePath := cache.LibraryImage(libraryImage.Hash, imageName)
+	c, err := cache.Create()
+	if c == nil || err != nil {
+		return fmt.Errorf("Failed to create cache object")
```
WestleyK:
```suggestion
		return fmt.Errorf("failed to create cache object")
```

internal/pkg/build/sources/conveyorPacker_library.go
```diff
@@ -68,6 +76,8 @@ func (cp *LibraryConveyorPacker) Get(b *types.Bundle) (err error) {
 		} else if cacheFileHash != libraryImage.Hash {
 			return fmt.Errorf("Cached File Hash(%s) and Expected Hash(%s) does not match", cacheFileHash, libraryImage.Hash)
```
WestleyK:
```suggestion
 			return fmt.Errorf("cached File Hash(%s) and expected hash(%s) does not match", cacheFileHash, libraryImage.Hash)
```

internal/pkg/client/cache/dir.go
```diff
@@ -22,69 +22,251 @@ const (
...
+	if c.Library == "" || c.Net == "" || c.OciTemp == "" || c.OciBlob == "" || c.Shub == "" {
+		return false
 	}
+
+	return true
 }
 
-func updateCacheSubdir(subdir string) string {
-	updateCacheRoot()
+// Clean : wipes all files in the cache directory, will return a error if one occurs
```
WestleyK:
```suggestion
// Clean wipes all files in the cache directory, will return a error if one occurs
```

internal/pkg/client/cache/dir.go
```diff
@@ -22,69 +22,251 @@ const (
...
+	}
+
+	absdir, err := filepath.Abs(filepath.Join(c.Root, subdir))
+	if err != nil {
+		return "", fmt.Errorf("Unable to get abs filepath: %v", err)
 	}
 
 	if err := initCacheDir(absdir); err != nil {
-		sylog.Fatalf("Unable to initialize caching directory: %v", err)
+		return "", fmt.Errorf("Unable to initialize caching directory: %v", err)
```
WestleyK:
```suggestion
		return "", fmt.Errorf("unable to initialize caching directory: %v", err)
```

WestleyK (changes requested):
Thanks for fixing this up! I have a couple comments/suggestion/questions...

internal/app/singularity/cache_clean_linux.go
gvallee: It does not really create a cache but create a cache handle. An existing cache can be used under the hood or a new one created depending on the environment variable and it it points to a new location.

internal/app/singularity/cache_list_linux.go
gvallee: Good catch.

internal/app/singularity/cache_clean_linux.go
gvallee: done

internal/app/singularity/cache_clean_linux.go
gvallee: I included some comments, do you think it is enough or would prefer to switch from `Create()` to something else?

gvallee: I also just realized that I may have misunderstood your comment in the first place and slightly change the code to make it cleaner. Let me know what you think about it.

internal/app/singularity/cache_list_linux.go
gvallee: Fixed

internal/pkg/build/sources/conveyorPacker_library.go
gvallee: fixed

internal/pkg/build/sources/conveyorPacker_library.go
gvallee: fixed

internal/pkg/client/cache/dir.go
gvallee: fixed

internal/pkg/client/cache/dir.go
gvallee: fixed

internal/pkg/client/cache/dir.go
```diff
@@ -22,69 +22,251 @@ const (
...
+	basedir, err := getCacheBasedir()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get root of the cache: %s", err)
+	}
+
+	return Init(basedir)
+}
+
+// Init initializes a new cache within a given directory
+func Init(baseDir string) (*SingularityCache, error) {
```
mem: I'm trying to figure out if this should be exported or not. Gut feeling is "no". Is there a reason why it should?

internal/pkg/client/cache/dir.go
```diff
@@ -22,69 +22,251 @@ const (
...
+
+	// Net provides the location of the Net cache
+	Net string
+
+	// Shub provides the location of the Shub cache
+	Shub string
+}
+
+// Create a new Singularity cache
+func Create() (*SingularityCache, error) {
```
mem: Because of the comments in the place of usage, I'm thinking this should have a different name.

Possibly: GetHandle, NewHandle.

The comments suggest this is not _creating_ a cache, but just getting a handle to an existing one.

internal/pkg/client/cache/dir.go
```diff
@@ -22,69 +22,251 @@ const (
...
 
-	return root
+	return newCache, nil
 }
 
-// Clean : wipes all files in the cache directory, will return a error if one occurs
-func Clean() error {
-	sylog.Debugf("Removing: %v", Root())
+// Destroy a specific Singularity cache
+func (c *SingularityCache) Destroy() error {
```
mem, line 149: in the same sense, this _is_ destroying the cache (L#155) but comments suggest more than one cache could be created at the same time.

internal/pkg/client/cache/dir.go
```diff
@@ -22,69 +22,251 @@ const (
...
+	// directory such as $HOME
+	usr, err := user.Current()
+	if err != nil {
+		return false
+	}
+
+	// The root cannot be empty of the Home directory. If the root was
+	// to be the home directory, destroying the cache would delete
+	// user's data
+	if c.Root == "" || c.Root == usr.HomeDir {
```
mem, line 185: this couples the notion of validity with the user of the cache.

Why would a cache be _invalid_ if it's in the user's home directory? My guess is that the problem is not the function, but the name of this function. `IsTransient`, maybe?

internal/pkg/client/cache/dir.go
```diff
@@ -22,69 +22,251 @@ const (
...
+	if c.Library == "" || c.Net == "" || c.OciTemp == "" || c.OciBlob == "" || c.Shub == "" {
+		return false
+	}
+
+	return true
+}
+
+// Clean wipes all files in the cache directory, will return a error if one occurs
+// Since renamed Destroy() but kept for backward compatibility
+func (c *SingularityCache) Clean() error {
```
mem, line 199: there's a single use of Clean outside this package, in `internal/app/singularity/cache_clean_linux.go`

For that specific use, the name `Clean` seems appropriate.

Maybe unexport `Destroy`?

mem (changes requested):

cmd/internal/cli/actions.go
```diff
@@ -58,9 +58,18 @@ func handleOCI(cmd *cobra.Command, u string) (string, error) {
 	}
 
 	name := uri.GetName(u)
-	imgabs := cache.OciTempImage(sum, name)
+	// Create a cache handle, which will provide access to an existing cache
+	// or create a new cache based on the current configuration.
+	c, err := cache.Create()
```
mem: this is the reason why I think the functions Create and Destroy might need renaming: if I'm calling Create and I see a Destroy function, I sort of expect that that needs to be called as well. Looking at what destroy actually does, I know that not only it doesn't need to be called, but that it should not be called.

internal/pkg/client/cache/dir.go
gvallee: At the moment, I think you are right, it does not need to be exported. I will fix it.

internal/pkg/client/cache/dir.go
gvallee: I am okay with that suggestion, I will switch to NewHandle().

internal/pkg/client/cache/dir.go
gvallee: Good point, I will fix it.

internal/pkg/client/cache/dir.go
gvallee: I have to disagree with this since the current semantic of clean is to delete the root. If it is the user directory, we would delete their home (it happens to me, I misunderstood the semantic of `Clean()` at first). So we either change the semantic of `Clean()` or keep the proposed modification. I think simply switching to `IsTransient()` would be extremely dangerous.

internal/pkg/client/cache/dir.go
gvallee: To me, `Clean()` does not match the actual semantic of the function, which erases everything. I kept both to not be too intrusive (introduce new API -> deprecate old one -> delete old one). If one function would have to not export one, I would not export `Clean()` instead of `Destroy()`.

internal/pkg/build/sources/conveyorPacker_library.go
gvallee: fixed

internal/pkg/client/cache/dir.go
```diff
@@ -22,69 +22,251 @@ const (
...
+// Figure out where the cache directory is.
+func getCacheBasedir() (string, error) {
+	// If the user defined the special environment variable, we use its value
+	usr, err := user.Current()
+	if err != nil {
+		return "", fmt.Errorf("couldn't determine user home directory: %s", err)
+	}
+
+	// Assuming the user set the environment variable
+	basedir := os.Getenv(DirEnv)
```
mem, line 219: presumably if this is set, it's an absolute path, so I would reverse the order here:

- look in the environment, if set, return that
- rest of the logic looking up user, home dir, etc.

internal/pkg/client/cache/dir.go
```diff
@@ -22,69 +22,251 @@ const (
...
 	}
+
+	return basedir, nil
 }
 
-func updateCacheSubdir(subdir string) string {
-	updateCacheRoot()
+// getCacheRoot figures out what the root directory is
+func getCacheRoot(basedir string) string {
+	return path.Join(basedir, RootDefault)
```
mem, line 239: should this not look in the environment variable?

Or how is the environment variable reconciled with this?

internal/pkg/client/cache/dir.go
```diff
@@ -22,69 +22,251 @@ const (
...
 
 	sylog.Debugf("Caching directory set to %s", absdir)
-	return absdir
+
+	return absdir, nil
 }
 
+// initCacheDir initializes a sub-cache within a cache, e.g., the shub sub-cache.
 func initCacheDir(dir string) error {
 	if _, err := os.Stat(dir); os.IsNotExist(err) {
```
mem: since this is already paying the price of calling stat, it should use the returned os.FileInfo to ask if `dir` is in fact a directory (`fi.IsDir()`)

internal/pkg/client/cache/dir_test.go
```diff
@@ -42,25 +30,276 @@ func TestRoot(t *testing.T) {
...
+	}
+
+	// Clean up by removing the file we created
+	err = os.Remove(cachePath)
+	if err != nil {
+		t.Fatalf("failed to delete %s: %s", cachePath, err)
+	}
+}
+
+func TestCreate(t *testing.T) {
```
mem, line 107: shouldn't this use the environment variable to specify a custom cache directory? so that the test does not file because there's an existing directory under the user's control?

cmd/internal/cli/actions.go
gvallee: I am renaming `Create()` to `NewHandle()`. Does it address that concern?

internal/pkg/client/cache/dir.go
gvallee: It is already the implemented semantic but the code is not organized correctly leading to confusions. I am fixing it.

internal/pkg/client/cache/dir.go
gvallee: The environment variable sets `basedir` not `root`. These two are distinct. The difference between the two is described in the code, let me know if it is unclear. I will improve the comments to clarify this a little better.

internal/pkg/client/cache/dir.go
gvallee: I was checking if it is a directory right after but can change it to use the FileInfo for sure. Good catch.

internal/pkg/client/cache/dir_test.go
gvallee: That is exactly my intent: the `TestCreate()` function creates a temporary directory that is then used as a base directory for the cache. Am I misunderstanding your comment? I will for sure improve the comments to avoid any confusion.

internal/app/singularity/cache_clean_linux.go
gvallee: Based on @mem, I rename the `Create()` function to `NewHandle()`, which greatly helps clarifying the situation.

internal/pkg/client/cache/dir.go
gvallee: fixed

internal/pkg/client/cache/dir.go
gvallee: done

internal/pkg/client/cache/dir.go
gvallee: fixed

cmd/internal/cli/actions.go
gvallee: Done.

internal/pkg/client/cache/dir.go
gvallee: Done.

internal/pkg/client/cache/dir.go
gvallee: Done.

internal/pkg/client/cache/dir.go
mem: I went back to the definition of Root several times, and I fail to see its point.

In the existing code (before this change), root defaults to ~/.singularity/cache, and it's possible to manipulate a variable and point it to ~/.

You are of course correct to say that it would be bad to remove the user's home directory ;-)

I think the check that needs to happen is not "am I deleting the user's home directory?" but instead "am I pointing the cache to the user's home directory?". In my mind that means that `Init` (or whatever it's called in the end) refuses to create a cache directly in the user's home directory. Since `IsValid` is a method on a `SingularityCache` instance, `c.Root` should _never_ point to the user's home directory. This also means that `Root` should not be exported, because that could be manipulated after the `SingularityCache` instance is created. This assumes that the _package's_ code, which does have access to unexported symbols, behaves in a sane way.

Taking a step back, I am still confused as to why `Root` is needed. The only directory that the cache should care about is the cache itself (not the parent directory or anything else outside the cache). This means the request should not be "please create a cache exactly {here}" but instead "please create a cache _inside_ {this directory}". This means, I don't ask for "~/.singularity/cache", I ask for "a cache in ~/.singularity", and the code then creates "~/.singularity/cache". External code should not have to care about what that directory is called (it could ask for items in the cache and get back absolute filenames).

Does this make sense?

gvallee: I am sorry but it does not make sense to me. Root is designed to be a way to protect the code to do bad thing and I read here that it is not a valuable feature:
- `Init` already refuses to create the cache is the home directory is used (L#141-142),
- it is possible, once the cache created to programmatically change the root (`myCache.Root = myHomeDir`), which is why a `IsValid()` function is so necessary: check that everything is fine before performing any action with the handle. While testing I had a few problems related to this and I do not feel confortable not including so a feature.
- if we do not separate `basedir` and `root`, we will always have the risk to have user mis-understand the semantic behind the environment variable and we will delete user's data and I personally do not believe that this is acceptable.

I am going to pull out my PR, I do not feel comfortable removing the protection that `Root` vs `Basedir` offers and do not see how I could make things better. Sorry.

As to not export `Root`, I agree with your point.

Issue #3304, opened 2019-04-15
Improve test coverage of pkg/sypgp


gvallee:
## Version of Singularity:

Improve test coverage of pkg/sypgp

### Actual behavior

`
$ go test -coverprofile=coverage.out && go tool cover -func=coverage.out
-----BEGIN PGP PUBLIC KEY BLOCK-----

xsBNBFy05GQBCADlz3KKYV9GuofpTAMfq783oGtmQ8BPn4WniV5NtMOke1ABHqA1
UvuaukIoVlAvinEs0WBq8bUIUN+M7n5ZQ1jIybJgLnmzyofVvgb+0XPQbLDZJhiv
1V1Ry3D60NT6KTo+quT//IuI1Zsk/II57YJ6Q6KsCshLgvVIy/XM4Af/l6b0kNQ1
PZRYUdl2Kxzo1YX50IJqfxArqUUWqcIpI5Erb0rDe0KQCxcHsK84iEud0oB2TQ/B
nr8TvsDLSMEBz26VmdVj954I6FwQF30Adoa6MzJejudBcNCq6lz73tQhiwZVoM38
b+pRAOtJxhMzQ2I/Q+xpmk7XoO6gl8SlRgIxABEBAAHNIFRlc3QgTmFtZSAoYmxh
aCkgPHRlc3RAdGVzdC5jb20+wsBiBBMBCAAWBQJctORkCRBwr/Ij5cxQ1gIbAwIZ
AQAAhakIAFYaJK6ia2r+mACSbiKTAmzGQ2Cc9fF/QVl/d/J064LFltN7AjTqoVVQ
YIrbkfD0ZdtqjD/F1rvR8sKBHC2w6w97b3Y5oVE74XCfGRdKPnVfUgYl0QzgUFbl
t9hdyIOrkCX94aBXyJBlvg7G927y5yvqFiKT/F2RTNL27PuSdxHNgA+7rAahQx9d
oeGrz9Uf4shCFgj4KLbAi5CXIeIpSAAtESzIRxU69bf5rfrF8aYf8FjC9WhvsFGB
TBKi7hVl9gEmwULudZNnHlkRYT5cRTksrPM2VPLcGKPy4GCssXs8yoJcV87DesWg
DwKa9mWz2DJmxUjdl3cqnfPEfydv7+LOwE0EXLTkZAEIANO2zTC31GqHrD/5opKl
[...]

PR #3524, opened 2019-05-09
Break up PathsCheck to make it testable

mem:
In order to make PathsCheck testable, break it into pieces that can be
individually tested.

- Add ensureDirPrivate and ensureFilePrivate which are testable using
  temporary directories
- Add tests for ensureDirPrivate and ensureFilePrivate

This addresses part of #3304 (it does not fix it).

### DIFF
--- a/pkg/sypgp/sypgp.go
+++ b/pkg/sypgp/sypgp.go
@@ -20,6 +20,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"syscall"
 
 	jsonresp "github.com/sylabs/json-resp"
 	"github.com/sylabs/scs-key-client/client"
@@ -71,28 +72,35 @@ func AskQuestionNoEcho(format string, a ...interface{}) (string, error) {
 	return string(response), nil
 }
 
-// GetTokenFile returns a string describing the path to the stored token file
-func GetTokenFile() string {
+// getSingularityDir returns the directory where the user's singularity
+// configuration and data is located.
+func getSingularityDir() string {
 	user, err := user.GetPwUID(uint32(os.Getuid()))
 	if err != nil {
-		sylog.Warningf("could not lookup user's real home folder %s", err)
-		sylog.Warningf("using current directory for %s", filepath.Join(".singularity", "sylabs-token"))
-		return filepath.Join(".singularity", "sylabs-token")
+		sylog.Warningf("Could not lookup user's real home directory: %s", err)
+
+		cwd, err := os.Getwd()
+		if err != nil {
+			sylog.Warningf("Could not get current working directory: %s", err)
+			return ".singularity"
+		}
+
+		dir := filepath.Join(cwd, ".singularity")
+		sylog.Warningf("Using current directory: %s", dir)
+		return dir
 	}
 
-	return filepath.Join(user.Dir, ".singularity", "sylabs-token")
+	return filepath.Join(user.Dir, ".singularity")
+}
+
+// GetTokenFile returns a string describing the path to the stored token file
+func GetTokenFile() string {
+	return filepath.Join(getSingularityDir(), "sylabs-token")
 }
 
 // DirPath returns a string describing the path to the sypgp home folder
 func DirPath() string {
-	user, err := user.GetPwUID(uint32(os.Getuid()))
-	if err != nil {
-		sylog.Warningf("could not lookup user's real home folder %s", err)
-		sylog.Warningf("using current directory for %s", filepath.Join(".singularity", "sypgp"))
-		return filepath.Join(".singularity", "sypgp")
-	}
-
-	return filepath.Join(user.Dir, ".singularity", "sypgp")
+	return filepath.Join(getSingularityDir(), "sypgp")
 }
 
 // SecretPath returns a string describing the path to the private keys store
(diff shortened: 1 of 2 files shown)


pkg/sypgp/sypgp.go
```diff
@@ -105,61 +113,84 @@ func PublicPath() string {
...
+
+	fs, err := os.OpenFile(fn, os.O_RDWR|os.O_CREATE, mode)
+
+	// restore umask...
+	syscall.Umask(oldumask)
+
+	// ... and check if there was an error
 	if err != nil {
 		return err
 	}
```
WestleyK, line 161: Should this err check go right under `fs, err := os.OpenFile(fn, os.O_RDWR|os.O_CREATE, mode)`? Line 153?
*ie. move `syscall.Umask(oldumask)` under the err check.*

pkg/sypgp/sypgp.go
```diff
@@ -105,61 +113,84 @@ func PublicPath() string {
...
-	if fsinfo.Mode() != 0600 {
-		sylog.Warningf("file mode (%v) on %v needs to be 0600, fixing that...", fsinfo.Mode(), SecretPath())
-		if err = fs.Chmod(0600); err != nil {
+
+	if currentMode := fsinfo.Mode(); currentMode != mode {
+		sylog.Warningf("file mode (%o) on %s needs to be %o, fixing that...", currentMode, fn, mode)
+		if err = fs.Chmod(mode); err != nil {
```
WestleyK: Probably better to add a `:`

```suggestion
		if err := fs.Chmod(mode); err != nil {
```

Because we are declaring a *new* `err` in this section.

pkg/sypgp/sypgp.go
```diff
@@ -71,28 +72,35 @@ func AskQuestionNoEcho(format string, a ...interface{}) (string, error) {
...
-func GetTokenFile() string {
+// getSingularityDir returns the directory where the user's singularity
+// configuration and data is located.
+func getSingularityDir() string {
 	user, err := user.GetPwUID(uint32(os.Getuid()))
 	if err != nil {
-		sylog.Warningf("could not lookup user's real home folder %s", err)
-		sylog.Warningf("using current directory for %s", filepath.Join(".singularity", "sylabs-token"))
-		return filepath.Join(".singularity", "sylabs-token")
+		sylog.Warningf("could not lookup user's real home directory: %s", err)
```
WestleyK: I'm fine with this warning starting with a non-cap, but to be consistent with the rest of the code, it might be better to start this with a cap:

```suggestion
		sylog.Warningf("Could not lookup user's real home directory: %s", err)
```

pkg/sypgp/sypgp.go
```diff
@@ -71,28 +72,35 @@ func AskQuestionNoEcho(format string, a ...interface{}) (string, error) {
...
+
+		cwd, err := os.Getwd()
+		if err != nil {
+			sylog.Warningf("could not get current working directory: %s", err)
```
WestleyK: Here too:

```suggestion
			sylog.Warningf("Could not get current working directory: %s", err)
```

pkg/sypgp/sypgp.go
```diff
@@ -71,28 +72,35 @@ func AskQuestionNoEcho(format string, a ...interface{}) (string, error) {
...
+			return ".singularity"
+		}
+
+		dir := filepath.Join(cwd, ".singularity")
+		sylog.Warningf("using current directory: %s", dir)
```
WestleyK: Here too:

```suggestion
		sylog.Warningf("Using current directory: %s", dir)
```

pkg/sypgp/sypgp.go
```diff
@@ -105,61 +113,84 @@ func PublicPath() string {
...
+	err := os.MkdirAll(dn, mode)
+
+	// restore umask...
+	syscall.Umask(oldumask)
+
+	// ... and check if there was an error
+
+	if err != nil {
 		return err
 	}
```
WestleyK, line 130: This err check should probably go right under line 121 (`err := os.MkdirAll(dn, mode)`). *Move `syscall.Umask(oldumask)` right under the err check.*

pkg/sypgp/sypgp.go
```diff
@@ -105,61 +113,84 @@ func PublicPath() string {
...
 	if err != nil {
 		return err
 	}
-	if dirinfo.Mode() != os.ModeDir|0700 {
-		sylog.Warningf("directory mode (%v) on %v needs to be 0700, fixing that...", dirinfo.Mode(), DirPath())
-		if err = os.Chmod(DirPath(), 0700); err != nil {
+
+	if currentMode := dirinfo.Mode(); currentMode != os.ModeDir|mode {
+		sylog.Warningf("directory mode (%o) on %s needs to be %o, fixing that...", currentMode & ^os.ModeDir, dn, mode)
+		if err = os.Chmod(dn, mode); err != nil {
```
WestleyK: Maybe?...

```suggestion
		if err := os.Chmod(dn, mode); err != nil {
```

WestleyK (changes requested):
Just a couple suggestions/comments :+1:

pkg/sypgp/sypgp.go
```diff
@@ -71,28 +72,35 @@ func AskQuestionNoEcho(format string, a ...interface{}) (string, error) {
 	return string(response), nil
 }
 
-// GetTokenFile returns a string describing the path to the stored token file
-func GetTokenFile() string {
+// getSingularityDir returns the directory where the user's singularity
+// configuration and data is located.
+func getSingularityDir() string {
```
gvallee, line 77: Maybe it should be pushed into a utility package, I have seen similar cache in the cache management, the default cache location being a `cache` directory in `getSingularityDir()`.

pkg/sypgp/sypgp.go
```diff
@@ -105,61 +113,84 @@ func PublicPath() string {
 	return filepath.Join(DirPath(), "pgp-public")
 }
 
-// PathsCheck creates the sypgp home folder, secret and public keyring files
-func PathsCheck() error {
-	// create the sypgp base directory
-	if err := os.MkdirAll(DirPath(), 0700); err != nil {
+func ensureDirPrivate(dn string) error {
```
gvallee, line 116: Maybe move this to `internal/pkg/util/fs`? That would avoid code duplication? I can see similar code in `internal/app/singularity/capability_manage_linux.go`, `internal/app/singularity/capability_list_linux.go`, `pkg/ocibundle/tools/oci.go`, `pkg/ocibundle/tools/overlay_linux.go`, and `pkg/util/unix/socket.go`.
That said, I have no problem opening an issue about having a utility package that includes such functionality, merging that current code, finish the current work `sygpg` and update later.

pkg/sypgp/sypgp.go
```diff
@@ -105,61 +113,84 @@ func PublicPath() string {
...
 			return err
 		}
 	}
 
-	// create or open the secret OpenPGP key cache file
-	fs, err := os.OpenFile(SecretPath(), os.O_RDWR|os.O_CREATE, 0600)
+	return nil
+}
+
+func ensureFilePrivate(fn string) error {
```
gvallee, line 147: Same here, maybe we should move it to a utility package. It would help with testing as well.

pkg/sypgp/sypgp.go
mem: no, I want to restore the old umask before potentially bailing out of the function. on line 129.

pkg/sypgp/sypgp.go
mem: no, same reason

pkg/sypgp/sypgp.go
WestleyK: Okay, that makes sense :+1:

pkg/sypgp/sypgp.go
```diff
@@ -105,61 +113,84 @@ func PublicPath() string {
...
 	fsinfo, err := fs.Stat()
 	if err != nil {
 		return err
 	}
-	if fsinfo.Mode() != 0600 {
-		sylog.Warningf("file mode (%v) on %v needs to be 0600, fixing that...", fsinfo.Mode(), SecretPath())
-		if err = fs.Chmod(0600); err != nil {
+
+	if currentMode := fsinfo.Mode(); currentMode != mode {
+		sylog.Warningf("file mode (%o) on %s needs to be %o, fixing that...", currentMode, fn, mode)
```
```suggestion
		sylog.Warningf("File mode (%o) on %s needs to be %o, fixing that...", currentMode, fn, mode)
```

pkg/sypgp/sypgp.go
```diff
@@ -105,61 +113,84 @@ func PublicPath() string {
...
+	dirinfo, err := os.Stat(dn)
 	if err != nil {
 		return err
 	}
-	if dirinfo.Mode() != os.ModeDir|0700 {
-		sylog.Warningf("directory mode (%v) on %v needs to be 0700, fixing that...", dirinfo.Mode(), DirPath())
-		if err = os.Chmod(DirPath(), 0700); err != nil {
+
+	if currentMode := dirinfo.Mode(); currentMode != os.ModeDir|mode {
+		sylog.Warningf("directory mode (%o) on %s needs to be %o, fixing that...", currentMode & ^os.ModeDir, dn, mode)
```
```suggestion
		sylog.Warningf("Directory mode (%o) on %s needs to be %o, fixing that...", currentMode & ^os.ModeDir, dn, mode)
```

WestleyK (changes requested):
Last suggestion, otherwise looks good :+1:

gvallee (commented):
A few comments...

pkg/sypgp/sypgp.go
mem: I am trying hard to not modify the package beyond what's needed to make it testable. I'll be happy to do as you suggest _after_ the tests have been added. As it is now, the package is very brittle and prone to breaking.

WestleyK (approved):
Looks good to me :+1:

PR #3526, opened 2019-05-10
Add initial support for pushing SIF to OCI registries (ORAS)

### LINKED ISSUE #2831: oras:// URI to support OCI registry [e.g. azurecr.io] as SIF storage endpoint

ikaneshiro:

This PR is to replace #3360, since I rebased the changed due to significant conflicts from recent CLI changes. This PR picks up where work in #3360 left off.

This work:
- Leverages ` github.com/deislabs/oras` in order to allow SIF images to be stored in OCI registries. 
  - Specifically v0.4.0
- Adds a new transport type to singularity called `oras` in order to facilitate this.
- Adds a basic `e2e` test
- Changes `e2e` tests to have a persistent oci registry in a docker container in order to allow multiple testing packages to leverage the same registry

Example usage(assumes local registry at `localhost:5000`):
```
$ singularity push oras.sif oras://localhost:5000/test_sif:latest
$ singularity pull oras://localhost:5000/test_sif:latest
INFO:    Download complete: test_sif_latest.sif
$ ./test_sif_latest.sif 
SIF container
```

This will push an image with the following config:
```
$ curl -s -H 'Accept: application/vnd.oci.image.manifest.v1+json' http://localhost:5000/v2/test_sif/manifests/latest | jq
{
  "schemaVersion": 2,
  "config": {
    "mediaType": "application/vnd.sylabs.sif.config.v1+json",
    "digest": "sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
    "size": 0
  },
  "layers": [
    {
      "mediaType": "appliciation/vnd.sylabs.sif.layer.tar",
      "digest": "sha256:5c7af839f0e52d9c5383bf949a4d04706192b4aab54d885dd991a0e3146660c1",
      "size": 2719744,
      "annotations": {
        "org.opencontainers.image.title": "oras.sif"
      }
    }
  ]
}
```

- Fixes #2831 

cc: @bauerm97 @SteveLasker @sajayantony


SteveLasker:
Hey folks.  Great to see all this work. 
Some quick feedback:
As much as we love ORAS, and the maintainers, ORAS is a library for pushing artifacts to an OCI Registry. I’d suggest using registry as the type. singularity push registry:
So registries understand your type; you’ll want to change the manifest.config.mediaTyoe to something like application/vnd.cncf.singularity.config.v1+json 
we actually already loaded these into acr. here’s the current types we support 

https://github.com/Azure/acr/blob/master/docs/artifact-media-types.json

ikaneshiro:
@bauerm97 @GodloveD thoughts on changing the transport from `oras` to `registry`?

ikaneshiro:
@SteveLasker, quoting some messages from @bauerm97 since he is currently on a plane:

> Basically, we already are interacting with docker and oci registries in the form of pulling from those registries to convert. If we were to use a generic name for the URI here, I would suggest it to be `oci`, because we're pulling from an oci registry (we're just pulling a sif artifact instead of a docker artifact).

>But we can't use `oci` because it's already being used. And I think adding `registry` would be quite confusing if we had bot that and `oci`

>Ideally, `oci` URI would automatically first look at which kind of artifact exists at the location which is referenced and then either pull and convert a docker image or just pull and extract the SIF.

>But I think that's way more effort than this as a MVP so it's better to start here and then polish it later

ikaneshiro:
Requesting reviews from @jscook2345 or @ArangoGutierrez to look over the `e2e` testing tweaks

SteveLasker:
On the oras, oci and registry types, it is a jumble. 
I do wonder if you still need the docker format. ow that we’ll have artifact sif formats support. I suppose ORAS does stand for OCI Registry As Storage, so it’s not inaccurate. And i wasn’t aware of all the other options you already support.

cmd/internal/cli/pull.go
```diff
@@ -304,6 +316,61 @@ func pullRun(cmd *cobra.Command, args []string) {
 		if err != nil {
 			sylog.Fatalf("%v\n", err)
 		}
+	case OrasProtocol:
```
gvallee, line 318: Would it be possible to move that code to a sub-function? I think it would keep the code more readable by limiting the length of the switch statement. It would also ease testing since we could have a unit test to cover some of the error cases for that function.
(Ultimately, it might be good to do the same for the library case)

cmd/internal/cli/push.go
```diff
@@ -60,20 +82,34 @@ var PushCmd = &cobra.Command{
 	Args:                  cobra.ExactArgs(2),
 	PreRun:                sylabsToken,
 	Run: func(cmd *cobra.Command, args []string) {
-		handlePushFlags(cmd)
+		file, dest := args[0], args[1]
```
gvallee, line 85: Could we put the code for the switch statements in separate functions so we can write unit tests for error cases later on? I can see that the library and ORAS cases start to be a little lengthy.

e2e/internal/e2e/docker.go
```diff
@@ -0,0 +1,44 @@
...
+package e2e
+
+import (
+	"os/exec"
+	"strings"
+	"testing"
+)
+
+func PrepRegistry(t *testing.T) {
+	commands := [][]string{
```
gvallee, line 15: I did not look at all the details about the E2E framework but do we need to call `test.EnsurePrivilege(t)` or `test.DropPrivilege(t)` and `defer test.ResetPrivilege(t)`?

e2e/internal/e2e/docker.go
```diff
@@ -0,0 +1,44 @@
...
+	for _, command := range commands {
+		cmd := exec.Command("docker", command...)
+		if b, err := cmd.CombinedOutput(); err != nil {
+			t.Logf(string(b))
+			t.Fatalf("command failed: %v", strings.Join(command, " "))
+		}
+	}
+}
+
+func KillRegistry(t *testing.T) {
```
gvallee, line 31: I did not look at all the details about the E2E framework but do we need to call `test.EnsurePrivilege(t)` or `test.DropPrivilege(t)` and `defer test.ResetPrivilege(t)`?

gvallee (changes requested):
General remark: this PR includes a lot of modifications related to `vendor`. Is it the intent?

cmd/internal/cli/pull.go
bauerm97: Agree with @gvallee here. @ikaneshiro suggested we handle this in a follow-up PR, which I think is fine. Before we merge this PR let's make some issues tracking this as tech-debt so we make sure we follow up

cmd/internal/cli/push.go
bauerm97: Same comment as with the pull switch statement. Let's create some tech-debt issues before we merge this PR. Then follow up and fix

bauerm97 (dismissed):
I think this looks good. @ikaneshiro once you create those tech debt issues I'm happy to give this the thumbs up for merge

cmd/internal/cli/pull.go
```diff
@@ -6,14 +6,24 @@
...
 import (
+	"context"
 	"fmt"
 	"io"
 	"os"
 	"os/signal"
 	"path/filepath"
 	"syscall"
 
+	"strings"
```
bauerm97, line 15: Small comment, this should be one line up (no whitespace btw `"syscall"` and `"strings"`)

ikaneshiro:
@bauerm97 @gvallee tech debt issue created: #3532

e2e/internal/e2e/docker.go
ikaneshiro: @gvallee I do not think so as the `e2e` framework is always ran as root

ikaneshiro:
@gvallee 
> General remark: this PR includes a lot of modifications related to `vendor`. Is it the intent?

Yes, we want to ensure that all dependencies are committed to `vendor` so that singularity can be built from source without an internet connection

gvallee (approved):

PR #3527, opened 2019-05-10
Try to make TestLogger more robust

mem:
TestLogger is failing every now and then because the log file has not
received all the writes yet.


ikaneshiro (approved):
lgtm

internal/pkg/instance/logger.go
```diff
@@ -127,16 +136,18 @@ func (l *Logger) scan(stream string, pr *io.PipeReader, pw *io.PipeWriter, dropC
...
 	for scanner.Scan() {
 		l.fileMutex.Lock()
+		t := scanner.Text()
 		if !dropCRNL {
-			fmt.Fprint(l.file, l.formatter(stream, r.Replace(scanner.Text())))
+			fmt.Fprint(l.file, l.formatter(stream, r.Replace(t)))
 		} else {
-			fmt.Fprint(l.file, l.formatter(stream, scanner.Text()))
+			fmt.Fprint(l.file, l.formatter(stream, t))
 		}
```
cclerget, line 144: Calling `l.file.Sync()` here doesn't work ? To avoid use of `WaitGroup`

cclerget (changes requested):

internal/pkg/instance/logger.go
mem: no, it does not work

cclerget: I'm a bit reluctant with the use of `WaitGroup` here (unless it's not possible otherwise) just to pass tests, and it implies to call `Close` for all writers to not stuck on `l.wg.Wait()` when calling `sync`, it could be good to call `Sync` from code without closing pipe, I would go for an exposed function with something like :

```
func (l *Logger) Sync() {
    l.fileMutex.Lock()
    l.file.Sync()
    l.fileMutex.Unlock()
}
```

And I may miss something, in test if `writer.Close` is called `scanner.Scan` loop end, then calling `logger.Sync()` just after close should be the same without using `WaitGroup` as there is only one writer ?

mem: Please feel free to test and push your changes to this PR

mem: Just to expand, the code is doing a number of dubious things.

1. It's opening a file and then opening it again, and expecting that writes to the second file handle are immediately available for reading via the first file handle (as if opening a file twice for reading and writing behaved like a pipe).
2. It starting a goroutine to _write_ data to the _Go_ pipe without any way of waiting for the reader end to have actually read the data. There's no synchronization between the two goroutines involved in this.
3. The loop in the scan method has no way to end, since the only thing that would cause it to end is a read error (won't happen in normal circumstances) or the other end of the pipe getting closed, so that the reader end gets EOF. Since the loop does not break, the reader and writer end of the pipe are never closed.

cclerget: Follow up in #3551

mem:
Already fixed in the other PR

PR #3538, opened 2019-05-13
Add askYNQuestion to ask y/n questions

mem:
It checks that the answer is either yes or no, and returns a default
value if the user does not provide an explicit answer.


WestleyK (approved):
Looks good to me :+1:

Issue #3199, opened 2019-04-03
key list output should start at one


jscook2345:
## Version of Singularity:

3.1-2

### Expected behavior

```
$ singularity keys list
Public key listing (/home/justin/.singularity/sypgp/pgp-public):

1) U: Sylabs Admin () <support@sylabs.io>
   C: 2018-11-01 18:20:02 -0400 EDT
   F: 8883491F4268F173C6E5DC49EDECE4F3F38D871E
   L: 4096
   --------
2) U: Justin Cook (test) <justin@sylabs.io>
   C: 2019-04-03 09:59:08 -0400 EDT
   F: DDAFFF474C3788E23986CB3F7922FD42029A6B27
   L: 4096
   --------
```

### Actual behavior

```
$ singularity keys list
Public key listing (/home/justin/.singularity/sypgp/pgp-public):

0) U: Sylabs Admin () <support@sylabs.io>
   C: 2018-11-01 18:20:02 -0400 EDT
   F: 8883491F4268F173C6E5DC49EDECE4F3F38D871E
   L: 4096
   --------
1) U: Justin Cook (test) <justin@sylabs.io>
   C: 2019-04-03 09:59:08 -0400 EDT
   F: DDAFFF474C3788E23986CB3F7922FD42029A6B27
   L: 4096
   --------
```

WestleyK:
Honestly, I agree with this. A couple hundred years ago when I was first making newpair / signing containers, I was expecting the list to start from `1`. It confused me a bit when the list started at `0`.

carterpeel:
Hello,

This is a templated response that is being sent out to all open issues.  We are working hard on 'rebuilding' the Singularity community, and a major task on the agenda is finding out what issues are still outstanding.


**Please consider the following:**

1. Is this issue a duplicate, or has it been fixed/implemented since being added?
2. Is the issue still relevant to the current state of Singularity's functionality?
3. Would you like to continue discussing this issue or feature request?


Thanks,
Carter

pedroalvesbatista:
@jscook2345 Still getting this problem ? Have this been solved already ? If yes, what work around have you followed and applied ?

We're looking into the issue carefully, soon will bring to community and discuss ways to better solve as well address this. Thankyou for keeping the interest in the subject.

kmuriki:
Old Singularity repo issues are now closed. If this is critical please open issue under new Apptainer repo.

PR #3539, opened 2019-05-13
[E2E] Tests for 'key'

### LINKED ISSUE #2839: Create e2e tests for `singularity key`
Create e2e tests for each sub-command and option for the `singularity key` command.

<pre>
[vagrant@localhost ~]$ singularity help keys
Manage OpenPGP key stores

Usage:
  singularity keys [keys options...] <subcommand>

Description:
  The 'key' command allows you to manage local OpenPGP key stores by creating
  a new store and new key pairs. You can also list available keys from the
  default store. Finally, the key command offers subcommands to communicate
  with an HKP key server to fetch and upload public keys.

Options:
  -h, --help   help for keys

Available Commands:
  list        List keys from the default key store
  newpair     Create a new OpenPGP key pair
  pull        Fetch an OpenPGP public key from a key server
  push        Upload an OpenPGP public key to a key server
  search      Search for keys matching string argument

Examples:
  All group commands have their own help output:

  $ singularity help key newpair
  $ singularity key list --help

For additional help or support, please visit https://www.sylabs.io/docs/
</pre>

### LINKED ISSUE #3533: [E2E] Add tests for key importing/exporting
### Expected behavior

CCI/Travis is testing the key importing/exporting.

### Actual behavior

CCI/Travis does not test the key import/export.

WestleyK:

This PR adds E2E test for the following key command:

1. `list`
2. `export`
3. `import`
4. `remove`
5. `pull`
6. `newpair`
7. `search`
8. ~`push`~ Not until CCI/Travis has an access token...

*Which is _almost_ all the `key` commands :tada:*

- Fixes #3533 
- Fixes #2839

<br>

This PR must be merged after https://github.com/sylabs/singularity/pull/3575.

<br>


e2e/internal/e2e/keyexec.go
```diff
@@ -0,0 +1,89 @@
+// Copyright (c) 2019, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package e2e
```
WestleyK: This file is added here to provide some `singularity key [...]` commands.

e2e/suite.go
```diff
@@ -16,14 +16,15 @@ import (
 	"syscall"
 	"testing"
 
-	"github.com/sylabs/singularity/e2e/actions"
```
jchavezlavalle: need to uncomment some lines around here and just add `"github.com/sylabs/singularity/e2e/key"` ?

e2e/internal/e2e/keyexec.go
```diff
@@ -0,0 +1,206 @@
...
+// BackupSecretKeyring ...
+func BackupSecretKeyring(t *testing.T) {
+	err := os.Rename(filepath.Join(HomeDir(), ".singularity/sypgp/pgp-secret"), backupSypgp)
+	if err != nil {
+		t.Fatalf("Unable to rename secret keyring: %v", err)
+	}
+}
+
+// RecuverSecretKeyring ...
+func RecuverSecretKeyring(t *testing.T) {
```
jchavezlavalle: What does this function do? @WestleyK

e2e/internal/e2e/keyexec.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+}
+
+// ExportPrivateKey will import a private key from kpath.
+func ExportPrivateKey(t *testing.T, kpath string) ([]byte, error) {
+	LoadEnv(t, &testenv)
+
+	s := getE(0, kpath, "", "e2etests")
+
+	t.Log("EEEEEEEEEEEEEEEEEE: ", s)
```
jchavezlavalle: remove this log

e2e/key/keyprivate/keyprivate.go
```diff
@@ -0,0 +1,131 @@
...
+
+	for i, line := range lines {
+		if strings.Contains(line, "B") {
+			lines[i] = "P"
+		}
+	}
+	output := strings.Join(lines, "\n")
+	err = ioutil.WriteFile(kpath, []byte(output), 0644)
+	if err != nil {
+		t.Fatalf("Unabel to write to file: %v", err)
```
jchavezlavalle: typo on word "Unable"

e2e/key/keyprivate/keyprivate.go
```diff
@@ -0,0 +1,131 @@
...
+						b, err := e2e.ImportKey(t, defaultKeyFile)
+						if err == nil {
+							t.Fatalf("Unexpected success: %s", string(b))
+						}
+					}))
+				} else {
+					if err == nil {
+						t.Log(string(out))
+						//t.Fatalf("Unexpected success when running: %s", cmd)
+					}
```
jchavezlavalle: remove commented code

e2e/key/keyprivate/keyprivate.go
```diff
@@ -0,0 +1,131 @@
...
+	t.Run("importing_test_key", test.WithoutPrivilege(func(t *testing.T) { e2e.ImportTestKey(t, "") }))
+
+	keyPath = testenv.TestDir
+	defaultKeyFile = filepath.Join(keyPath, "exported_private_key")
+
+	// Pull the default public key
+	t.Run("pull_default_key", test.WithoutPrivilege(func(t *testing.T) { e2e.PullDefaultPublicKey(t) }))
+
+	// Run the tests
+	t.Run("pubic_key", testPrivateKey)
```
jchavezlavalle: typo on word public

e2e/key/keyprivate/keyprivate.go
```diff
@@ -0,0 +1,131 @@
...
+
+	// Recuver the secret keyring
+	t.Run("recuvering_secret_keyring", test.WithoutPrivilege(func(t *testing.T) { e2e.RecuverSecretKeyring(t) }))
```
jchavezlavalle: maybe typo as well on "Recovering" ?

e2e/key/keypublic/keypublic.go
```diff
@@ -0,0 +1,147 @@
...
+
+	for i, line := range lines {
+		if strings.Contains(line, "B") {
+			lines[i] = "P"
+		}
+	}
+	output := strings.Join(lines, "\n")
+	err = ioutil.WriteFile(kpath, []byte(output), 0644)
+	if err != nil {
+		t.Fatalf("Unabel to write to file: %v", err)
```
jchavezlavalle: typo on Unable

e2e/suite.go
```diff
@@ -89,19 +90,21 @@ func Run(t *testing.T) {
 
 	// RunE2ETests by functionality
 
-	t.Run("BUILD", imgbuild.RunE2ETests)
+	t.Run("KEYCMD", key.RunE2ETests)
 
-	t.Run("ACTIONS", actions.RunE2ETests)
+	//	t.Run("BUILD", imgbuild.RunE2ETests)
```
jchavezlavalle: need as well to uncomment all these lines from previous tests

jchavezlavalle (changes requested):
left some comments, thanks @WestleyK

e2e/internal/e2e/keyexec.go
WestleyK: This function `RecoverSecretKeyring` will recover your old keyring. When the tests run you keyring is backup to a tmp file, after the tests run your keyring is recovered.

jchavezlavalle: okay good, there's a typo in this function name

e2e/key/key.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+//RunE2ETests is the main func to trigger the test suite
+func RunE2ETests(t *testing.T) {
+	err := envconfig.Process("E2E", &testenv)
+	if err != nil {
+		t.Fatal(err.Error())
+	}
+
+	t.Run("GenralKeyCmdTest", testGenralKeyCmd)
```
jchavezlavalle: There's a typo in `GeneralKeyCmdTest` and `testGeneralKeyCmd`

e2e/key/keyprivate/keyprivate.go
```diff
@@ -0,0 +1,131 @@
...
+	CmdPath     string `split_words:"true"`
+	TestDir     string `split_words:"true"`
+	RunDisabled bool   `default:"false"`
+}
+
+var testenv testingEnv
+var keyPath string
+var defaultKeyFile string
+
+//  coruptKey will take a ASCII key (kpath) and change some chars in it (corrupt it).
```
jchavezlavalle: typo: corruptKey

e2e/internal/e2e/keyexec.go
```diff
@@ -0,0 +1,204 @@
...
+`, armor, kpath, num, psk)
+}
+
+var backupSypgp = filepath.Join(HomeDir(), ".singularity/sypgp/secret-keyring-backup")
+
+// PullDefaultPublicKey will pull the public Sylabs Admin key
+func PullDefaultPublicKey(t *testing.T) {
+	LoadEnv(t, &testenv)
+
+	argv := []string{"key", "pull", "F69C21F759C8EA06FD32CCF4536523CE1E109AF3"}
```
mem: this is working against ~/.singularity/sypgp.

I can see that you have this set up with backup / restore to make sure that this key is not already present in that directory.

I would prefer it if the code (`DirPath` in `pkg/sypgp/sypgp.go` and related functions) is modified so that it either uses an environment variable to obtain the name of the `.singularity` directory, or it uses an environment variable to obtain the name of the `.singularity/sypgp` directory. Then you set this environment variable to a temporary directory that you know for a fact that does NOT contain this key.

I prefer the first option even if it's much more work, but the second is acceptable for now.

e2e/internal/e2e/keyexec.go
```diff
@@ -0,0 +1,208 @@
...
+// BackupSecretKeyring will take your secret keyring file, and back it up. This gets ran before the private
+// key testing.
+func BackupSecretKeyring(t *testing.T) {
+	err := os.Rename(filepath.Join(HomeDir(), ".singularity/sypgp/pgp-secret"), backupSypgp)
+	if err != nil {
+		t.Fatalf("Unable to rename secret keyring: %v", err)
+	}
+}
+
+// RecoverSecretKeyring will recover your secret keyring, this gets ran after the private key test are complete.
```
jchavezlavalle: gets **run**

e2e/internal/e2e/keyexec.go
```diff
@@ -0,0 +1,208 @@
...
+func RecoverSecretKeyring(t *testing.T) {
+	if err := os.Remove(filepath.Join(HomeDir(), ".singularity/sypgp/pgp-secret")); err != nil {
+		t.Fatalf("Unable to remove secret keyring: %v", err)
+	}
+	if err := os.Rename(backupSypgp, filepath.Join(HomeDir(), ".singularity/sypgp/pgp-secret")); err != nil {
+		t.Fatalf("Unable to rename secret keyring: %v", err)
+	}
+}
+
+// RemoveSecretKeyring will delete your secret keyring :O
```
jchavezlavalle: remove surprised emoji at the end of this comment lol

e2e/internal/e2e/keyexec.go
```diff
@@ -0,0 +1,208 @@
...
+func ImportKey(t *testing.T, kpath string) ([]byte, error) {
+	LoadEnv(t, &testenv)
+
+	argv := []string{"key", "import", kpath}
+	execKey := exec.Command(testenv.CmdPath, argv...)
+
+	return execKey.CombinedOutput()
+}
+
+// ImportTestKey will take a private key file (kpath) and import it.
```
jchavezlavalle: Update comment: ImportTestKey -> ImportPrivateKey

e2e/key/keypublic/keypublic.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	keyPath = testenv.TestDir
+	defaultKeyFile = filepath.Join(keyPath, "exported_key")
+
+	// Pull the default public key
+	t.Run("pull_default_key", test.WithoutPrivilege(func(t *testing.T) { e2e.PullDefaultPublicKey(t) }))
+
+	// Run the tests
+	t.Run("pubic_key", testPublicKey)
```
jchavezlavalle: typo on public_key

jchavezlavalle (changes requested):
looking good, added some comments around @WestleyK

e2e/internal/e2e/keyexec.go
```diff
@@ -0,0 +1,204 @@
...
+		t.Log(string(out))
+		t.Fatalf("Unable to pull key: %v", err)
+	}
+}
+
+// RemoveDefaultPublicKey will pull the public Sylabs Admin key
+func RemoveDefaultPublicKey(t *testing.T) {
+	LoadEnv(t, &testenv)
+
+	argv := []string{"key", "remove", "F69C21F759C8EA06FD32CCF4536523CE1E109AF3"}
```
mem: even if the above is a nitpick, this is not.

How do you ensure that the key exists? This needs fixing.

e2e/key/keyprivate/keyprivate.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// TestAll is trigered by ../key.go, that is trigered by suite.go in the e3e test directory
+func TestAll(t *testing.T) {
+	err := envconfig.Process("E2E", &testenv)
+	if err != nil {
+		t.Fatal(err.Error())
+	}
+
+	t.Run("backingup_secret_keyring", test.WithoutPrivilege(func(t *testing.T) { e2e.BackupSecretKeyring(t) }))
```
mem: this is not a test, this should not be running with t.Run

e2e/key/keyprivate/keyprivate.go
```diff
@@ -0,0 +1,136 @@
...
+						t.Log(string(out))
+						t.Fatalf("Unexpected success")
+					}
+				}
+			}
+		}))
+	}
+}
+
+// TestAll is trigered by ../key.go, that is trigered by suite.go in the e3e test directory
```
mem: this is documenting how the function works (internal documentation) not what the function does (external documentation). You can move this inside the function.

e2e/key/keyprivate/keyprivate.go
```diff
@@ -0,0 +1,136 @@
...
+	err := envconfig.Process("E2E", &testenv)
+	if err != nil {
+		t.Fatal(err.Error())
+	}
+
+	t.Run("backingup_secret_keyring", test.WithoutPrivilege(func(t *testing.T) { e2e.BackupSecretKeyring(t) }))
+	t.Run("importing_test_key", test.WithoutPrivilege(func(t *testing.T) {
+		b, err := e2e.ImportPrivateKey(t, "./key/testdata/e2e_test_key.asc")
+		if err != nil {
+			t.Log(string(b))
```
mem: Please add something to the output what indicates what it is, for example, "output of running <whatever>:\n"

e2e/key/keypublic/keypublic.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// TestAll is trigered by ../key.go, that is trigered by suite.go in the e3e test directory
+func TestAll(t *testing.T) {
+	err := envconfig.Process("E2E", &testenv)
+	if err != nil {
+		t.Fatal(err.Error())
+	}
+
+	keyPath = testenv.TestDir
```
mem, line 162: please don't use globals here.

e2e/key/keypublic/keypublic.go
```diff
@@ -0,0 +1,147 @@
...
+	defaultKeyFile = filepath.Join(keyPath, "exported_key")
```
mem, line 163: please don't use globals here.

e2e/key/keyprivate/keyprivate.go
```diff
@@ -0,0 +1,136 @@
...
+			t.Fatalf("Unable to import test key: %v", err)
+		}
+	}))
+
+	keyPath = testenv.TestDir
```
mem, line 169: please don't use globals here.

e2e/key/keyprivate/keyprivate.go
```diff
@@ -0,0 +1,136 @@
...
+	defaultKeyFile = filepath.Join(keyPath, "exported_private_key")
```
mem, line 170: please don't use globals here.

e2e/key/keyprivate/keyprivate.go
```diff
@@ -0,0 +1,136 @@
...
+
+	// Pull the default public key
+	t.Run("pull_default_key", test.WithoutPrivilege(func(t *testing.T) { e2e.PullDefaultPublicKey(t) }))
+
+	// Run the tests
+	t.Run("public_key", testPrivateKey)
+
+	// Recover the secret keyring
+	t.Run("recovering_secret_keyring", test.WithoutPrivilege(func(t *testing.T) { e2e.RecoverSecretKeyring(t) }))
```
mem: this is also not a test, this is teardown.

If you need to set up and tear down for each test, you can wrap the second argument to t.Run with a function that takes a func (t *testing.T) function and returns a func (t *testing.T) function, just like test.WithoutPrivilege does.

For example:

```
func withEmptySecretKeyring(f func (t *testing.T)) {
    return func (t *testing.T) {
        backupSecretKeyring(t)
        defer restoreSecretKeyringt(t)
        f(t)
    }
}
```

`defer` is required becuase defer runs even if the function f panics.

e2e/internal/e2e/keyexec.go
```diff
@@ -0,0 +1,204 @@
...
+send "%v\r"
+
+expect "Enter key passphrase : "
+send "%s\r"
+
+expect eof
+`, armor, kpath, num, psk)
+}
+
+var backupSypgp = filepath.Join(HomeDir(), ".singularity/sypgp/secret-keyring-backup")
```
mem: there's an exported function in the sypgp package that returns "~/.singularity/sypgp".

e2e/internal/e2e/keyexec.go
```diff
@@ -0,0 +1,204 @@
...
+
+// PullDefaultPublicKey will pull the public Sylabs Admin key
+func PullDefaultPublicKey(t *testing.T) {
```
mem: this does not need to be exported.

e2e/internal/e2e/keyexec.go
mem: Does it belong in e2e/internal/e2e, though? It's very specific to key testing.

e2e/internal/e2e/keyexec.go
```diff
@@ -0,0 +1,204 @@
...
+	if err != nil {
+		t.Fatalf("Unable to write to file: %v", err)
+	}
+	err = exportScript.Close()
+	if err != nil {
+		t.Fatalf("Unable to clise file: %v", err)
+	}
+
+	argv := []string{exportScript.Name()}
+	execImport := exec.Command("/usr/bin/expect", argv...)
```
mem: is it necessary to hardcode the path like this?

e2e/internal/e2e/keyexec.go
```diff
@@ -0,0 +1,204 @@
...
+
+	return execImport.CombinedOutput()
```
mem: you should check the exit code of the command, no?

e2e/internal/e2e/keyexec.go
```diff
@@ -0,0 +1,204 @@
...
+	if err != nil {
+		t.Fatalf("%v", err)
+	}
+
+	_, err = stdinRun.Seek(0, os.SEEK_SET)
+	if err != nil {
+		t.Fatalf("%v", err)
+	}
+
+	execKey.Stdin = stdinRun
```
mem: note that Cmd.Stdin is an io.Reader. A bytes.Buffer is also an io.Reader (also strings.Reader).

This means you can place `stdin` in a bytes.Buffer and you are set. Since `stdin` is a `string` and not a `[]byte`, you can use `strings.NewReader` to obtain an io.Reader.

e2e/key/key.go
```diff
@@ -0,0 +1,102 @@
...
+
+func testGeneralKeyCmd(t *testing.T) {
+	tests := []struct {
+		name    string
+		args    []string
+		file    string
+		succeed bool
+	}{
+		{
+			name:    "key_list",
```
mem: you omit the underscore. When running with -v that will read more naturally.

e2e/key/key.go
```diff
@@ -0,0 +1,102 @@
...
+		{
+			name:    "key_flag",
+			args:    []string{"--notaflag"},
+			file:    "",
+			succeed: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run("key_run "+tt.name, test.WithoutPrivilege(func(t *testing.T) {
```
mem: no need for the "key_run" prefix, as it will automatically get prefixed with the subtest that is calling this.

e2e/key/key.go
```diff
@@ -0,0 +1,102 @@
...
+			cmd, out, err := e2e.RunKeyCmd(t, testenv.CmdPath, tt.args, tt.file, "")
+			if tt.succeed {
```
mem, line 70: please see e2e/pull/pull.go around line ~ 302 (switch) for an example of how to handle this. You need to consider four cases: expecting failure + failed; expecting success + success; expecting failure + success; expecting success + failure and my recommendation is to use a switch to make it explicit that you are handling all four cases.

e2e/key/keyprivate/keyprivate.go
```diff
@@ -0,0 +1,136 @@
...
+)
+
+type testingEnv struct {
+	// base env for running tests
+	CmdPath     string `split_words:"true"`
+	TestDir     string `split_words:"true"`
+	RunDisabled bool   `default:"false"`
+}
+
+var testenv testingEnv
```
mem, line 25: please no globals in tests.

e2e/key/keyprivate/keyprivate.go
```diff
@@ -0,0 +1,136 @@
...
+var keyPath string
+var defaultKeyFile string
+
+// corruptKey will take a ASCII key (kpath) and change some chars in it (corrupt it).
+func corruptKey(t *testing.T, kpath string) {
```
mem: this is a good idea for a test, but my gut feeling is that this is better suited for a unit test.

Is there something that cannot be tested as a unit test?

e2e/key/keyprivate/keyprivate.go
```diff
@@ -0,0 +1,136 @@
...
+		args    []string
+		stdin   string
+		file    string
+		corrupt bool
+		succeed bool
+	}{
+		{
+			name:    "export_private",
+			args:    []string{"export", "--secret"},
+			stdin:   "0\n", // TODO: this will need to be '1' at some point in time -> issue #3199
```
mem, line 86: can you be more explicit as to what's the meaning of that 0 and why it's changing to a 1?

e2e/key/keyprivate/keyprivate.go
```diff
@@ -0,0 +1,136 @@
...
+			file:    defaultKeyFile,
+			succeed: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run("key_run "+tt.name, test.WithoutPrivilege(func(t *testing.T) {
```
mem: same here with "key_run"

e2e/key/keyprivate/keyprivate.go
```diff
@@ -0,0 +1,136 @@
...
+			os.RemoveAll(filepath.Join(keyPath, defaultKeyFile))
+			out, err := e2e.ExportPrivateKey(t, tt.file)
+			if tt.succeed {
```
mem: same here

e2e/key/keypublic/keypublic.go
```diff
@@ -0,0 +1,147 @@
...
+	TestDir     string `split_words:"true"`
+	RunDisabled bool   `default:"false"`
+}
+
+var testenv testingEnv
+var keyPath string
+var defaultKeyFile string
+
+// corruptKey will take a ASCII key (kpath) and change some chars in it (corrupt it).
+func corruptKey(t *testing.T, kpath string) {
```
mem: this looks like it's the same code as the other package

e2e/key/keypublic/keypublic.go
```diff
@@ -0,0 +1,147 @@
...
+	tests := []struct {
+		name    string
+		args    []string
+		stdin   string
+		file    string
+		corrupt bool
+		succeed bool
+	}{
+		{
+			name:    "export_public",
```
mem: same comment with the underscores.

e2e/key/keypublic/keypublic.go
```diff
@@ -0,0 +1,147 @@
...
+			args:    []string{"export", "--armor"},
+			stdin:   "0\n",
+			file:    defaultKeyFile,
+			corrupt: true,
+			succeed: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run("key_run "+tt.name, test.WithoutPrivilege(func(t *testing.T) {
```
mem: same with the "key_run" label

mem (changes requested):

e2e/internal/e2e/keyexec.go
WestleyK: Umm, it used in `e2e/key/keypublic/keypublic.go` to pull the key.

e2e/internal/e2e/keyexec.go
WestleyK: First, the command will not fail if theres no key. Second, if there is no key when this function is called, theres a another problem. So I dont think its necessary to check if the key exist first...

e2e/internal/e2e/keyexec.go
WestleyK: Well, it seems to work without the full path :smile:

e2e/internal/e2e/keyexec.go
WestleyK: This is checked where the function is called, in some test cases, this is supposed to fail.

e2e/key/key.go
WestleyK: Okay, I'll remove the `_` :+1:

e2e/key/keyprivate/keyprivate.go
WestleyK: Take a look at this issue https://github.com/sylabs/singularity/issues/3199.

mem: I know I can look at the issue. The issue is too far away when I'm looking at code. It's good to document that there's an open issue related to this, but it's not enough.

In this specific case, something like "the first entry in the selection list is item 0; see issue #3199"

e2e/internal/e2e/keyexec.go
mem: Uhm...

in order to help with automation, the command _should_ return with a non-zero exit code if the key to be removed does not exist. If that's currently not the case, an issue should be opened for that, and a test should be added, so that we the code changes, the test catches it.

As to the "other problem"... what might that be?

My original point is that in a test you should make sure that all the preconditions for the test are met. If I see this test without any other context, I would expect that there's something making sure this particular key exists. I don't see that here, which makes the test code harder to understand.

WestleyK: >First, the command will not fail if theres no key.

Nevermined, it does fail if theres no key.

Any ideas on how to check if this key is in the keyring?

e2e/key/keypublic/keypublic.go
WestleyK: ? This is in its own package, is it imported not to use globals? :thinking:

e2e/internal/keyexec/keyexec.go
```diff
@@ -0,0 +1,244 @@
...
+	c.Send("e2etests\n")
+
+	err = cmd.Wait()
+	cm := fmt.Sprintf("%s %s", testenv.CmdPath, strings.Join(exportCmd, " "))
+
+	return cm, outErr.Bytes(), err
+}
+
+// CorruptKey will take a ASCII key (kpath) and change some chars in it (corrupt it).
+func CorruptKey(t *testing.T, kpath string) {
```
WestleyK: Would it be better to use `sed` and call a `exec.Command()`?

WestleyK: Or even better is to put a already corrupted key in the `e2e/key/testdata`?... <- I vote for this.

e2e/internal/e2e/env.go
```diff
@@ -15,3 +16,8 @@ func LoadEnv(t *testing.T, env interface{}) {
 		t.Fatalf("Failed to load environment: %+v\n", err)
 	}
 }
+
+// HomeDir will return the users home directory.
+func HomeDir() string {
```
mem: by default, please do not export functions. It increases the size of the API and makes things more difficult to use.

It's perfectly fine to export functions if you need to use the code from multiple packages, I just can't find where this is being used, so I can't really say if this needs to be exported or not.

e2e/internal/keyexec/keyexec.go
```diff
@@ -0,0 +1,216 @@
...
+	CmdPath     string `split_words:"true"`
+	TestDir     string `split_words:"true"`
+	RunDisabled bool   `default:"false"`
+}
+
+var testenv testingEnv
+
+// PullDefaultPublicKey will pull the public Sylabs Admin key
+func PullDefaultPublicKey(t *testing.T) {
+	e2e.LoadEnv(t, &testenv)
```
mem, line 42: this function is not using values from testenv, so you don't need this.

e2e/internal/keyexec/keyexec.go
```diff
@@ -0,0 +1,361 @@
...
+
+	argv := []string{"key", "pull", "F69C21F759C8EA06FD32CCF4536523CE1E109AF3"}
```
mem: since the fingerprint is going to be used multiple times across the test (or across multiple tests?) can you please put this is a package-level constant?

e2e/internal/e2e/env.go
WestleyK: Actually, its not being used anymore, I'll just delete it.

e2e/internal/keyexec/keyexec.go
WestleyK: Wait, it does actually use it. I dont know why CCI/Travis passed, it failed locally for me.

WestleyK:
Lol @ArangoGutierrez! You updated my branch just as I was trying to push! I was wondering why it kept failing when I pushed it :smile:

e2e/key/keyprivate/keyprivate.go
```diff
@@ -0,0 +1,202 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, test.WithoutPrivilege(func(t *testing.T) {
+			t.Run("remove_private_keyring_before_newpair", test.WithoutPrivilege(func(t *testing.T) { keyexec.RemoveKeyring(t) }))
+			c, b, err := keyexec.KeyNewPair(t, tt.user, tt.email, tt.note, tt.psk1, tt.psk2, tt.push)
+			if tt.succeed {
+				if err != nil {
+					t.Log("Command that failed: ", c)
+					t.Log(string(b))
+					t.Fatalf("unepexted failure: %v", err)
```
jchavezlavalle: typo on word unexpected

e2e/key/keyprivate/keyprivate.go
```diff
@@ -0,0 +1,202 @@
...
+				}
+				//t.Run("testing_new_key_from_newpair", keyexec.QuickTestKey)
```
jchavezlavalle: remove line 71 commented code

e2e/key/keyprivate/keyprivate.go
```diff
@@ -0,0 +1,202 @@
...
+				keyexec.QuickTestKey(t)
+			} else {
+				if err == nil {
+					t.Log(string(b))
+					t.Fatalf("unexpected succees running: %v", err)
```
jchavezlavalle: typo on line 76 word success

e2e/internal/keyexec/keyexec.go
```diff
@@ -0,0 +1,361 @@
...
+
+	err = cmd.Start()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Send the passcode to singularity
+	c.Send("e2etests\n")
+	c.Send("e2etests\n")
+	c.Send("e2etests\n")
```
jchavezlavalle: why 3 times?

e2e/internal/keyexec/keyexec.go
```diff
@@ -0,0 +1,361 @@
...
+
+	exportCmd = append(exportCmd, kpath)
+
+	outErr := bytes.NewBuffer(nil)
+
+	cmd := exec.Command(testenv.CmdPath, exportCmd...)
+	cmd.Stdin = c.Tty()
+
+	cmd.Stderr = outErr
+	cmd.Stdout = outErr
```
jchavezlavalle, line 169: why are these 2 sent as error and output with the same return value?

e2e/internal/keyexec/keyexec.go
```diff
@@ -0,0 +1,361 @@
...
+
+	execKey.Stdin = stdinRun
+	out, err := execKey.CombinedOutput()
+
+	return cmd, out, err
+}
+
+// QuickTestKey will export a private, and public key (0), and then import them. This is used after
+// generating a newpair.
+func QuickTestKey(t *testing.T) {
```
jchavezlavalle: Better to call it `QuickTestExportImportKey` to know exactly what this test is doing beforehand

e2e/internal/keyexec/keyexec.go
```diff
@@ -0,0 +1,361 @@
...
+		succeed bool
+	}{
+		{
+			name:    "quick test public",
+			private: false,
+			armor:   false,
+			succeed: true,
+		},
+		{
+			name:    "quick test public  armor",
```
jchavezlavalle: additional space on name (between public and armor)

e2e/internal/keyexec/keyexec.go
```diff
@@ -0,0 +1,361 @@
...
+			if tt.private {
+				c, b, err = ExportPrivateKey(t, filepath.Join(tmpTestDir, "export_key.asc"), "0\n", tt.armor)
+			} else {
+				c, b, err = RunKeyCmd(t, testenv.CmdPath, []string{"export"}, filepath.Join(tmpTestDir, "export_key.asc"), "0\n")
+			}
+			if tt.succeed {
+				if err != nil {
+					t.Log("Command that failed: ", c)
+					t.Log(string(b))
+					t.Fatalf("unepexted failure: %v", err)
```
jchavezlavalle: typo on line 290

e2e/internal/keyexec/keyexec.go
```diff
@@ -0,0 +1,361 @@
...
+					c, b, err := ImportPrivateKey(t, filepath.Join(tmpTestDir, "export_key.asc"))
+					if err != nil {
+						t.Log("command that failed: ", c, string(b))
+						t.Fatalf("Unable to import key: %v", err)
+					}
+				}))
+			} else {
+				if err == nil {
+					t.Log(string(b))
+					t.Fatalf("unexpected succees running: %v", c)
```
jchavezlavalle: typo on line 307

jchavezlavalle (changes requested):

e2e/internal/keyexec/keyexec.go
WestleyK: Because when you import a private key, it will ask you for the old password, then a new password, and then to retype you password. For this test, its all using the same password.

e2e/internal/keyexec/keyexec.go
WestleyK: This is to capture the stdout, and stderr, and returns it. Useful for debugging or when a test fails.

e2e/internal/keyexec/keyexec.go
jchavezlavalle: cool, could you add a comment to that? so that when somebody needs to maintain it in the future it could be clear

e2e/internal/keyexec/keyexec.go
jchavezlavalle: why not using the combined output instead?

e2e/internal/keyexec/keyexec.go
```diff
@@ -0,0 +1,361 @@
...
+				}
+			}
+		}))
+	}
+}
+
+func KeyNewPair(t *testing.T, user, email, note, psk1, psk2 string, push bool) (string, []byte, error) {
```
jchavezlavalle: add comments on all functions before the definition, so that it can be clear what they're doing :+1:

jchavezlavalle (changes requested):

e2e/internal/keyexec/keyexec.go
WestleyK: Because `cmd.CombinedOutput` will wait for the command to finish running, and we want to use `c.Send()` to send input for the command thats running, (we cant do that if the command is already done running). So here we use `cmd.Start()` to start the command, then send the input, and then wait for the command to finish running by `cmd.Wait()`. *hope that makes some sense...*

e2e/key/keyprivate/keyprivate.go
```diff
@@ -0,0 +1,202 @@
...
+type testingEnv struct {
+	// base env for running tests
+	CmdPath     string `split_words:"true"`
+	TestDir     string `split_words:"true"`
+	RunDisabled bool   `default:"false"`
+}
+
+var testenv testingEnv
+var keyPath string
+var defaultKeyFile string
```
jchavezlavalle, line 28: remove globals `keyPath` and `defaultKeyFile` from global scope and define them where needed

jchavezlavalle (changes requested):

e2e/internal/keyexec/keyexec.go
```diff
@@ -0,0 +1,347 @@
...
+	// base env for running tests
+	CmdPath     string `split_words:"true"`
+	TestDir     string `split_words:"true"`
+	RunDisabled bool   `default:"false"`
+}
+
+var testenv testingEnv
+
+// PullDefaultPublicKey will pull the public Sylabs Admin key
+func PullDefaultPublicKey(t *testing.T) {
```
mem, line 41: looks to me that this is not really a test, but a helper function, maybe don't export it?

e2e/internal/keyexec/keyexec.go
```diff
@@ -0,0 +1,347 @@
...
+
+	out, err := execKey.CombinedOutput()
+	if err != nil {
+		t.Log(string(out))
+		t.Fatalf("Unable to pull key: %v", err)
+	}
+}
+
+// RemoveDefaultPublicKey will pull the public Sylabs Admin key
+func RemoveDefaultPublicKey(t *testing.T) {
```
mem, line 56: same here about not exporting

e2e/internal/keyexec/keyexec.go
```diff
@@ -0,0 +1,347 @@
...
+
+	out, err := execKey.CombinedOutput()
+	if err != nil {
+		t.Log(string(out))
+		t.Fatalf("Unable to pull key: %v", err)
+	}
+}
+
+func RemovePublicKeyring(t *testing.T) {
+	err := os.Remove(sypgp.PublicPath())
```
mem, line 70: for this to succeed, sypgp.PublicPath() ist be empty. Maybe document this as part of the function's documentation?

e2e/internal/keyexec/keyexec.go
```diff
@@ -0,0 +1,347 @@
...
+	execKey := exec.Command(testenv.CmdPath, argv...)
+
+	out, err := execKey.CombinedOutput()
+	if err != nil {
+		t.Log(string(out))
+		t.Fatalf("Unable to pull key: %v", err)
+	}
+}
+
+func RemovePublicKeyring(t *testing.T) {
```
mem: don't export

e2e/internal/keyexec/keyexec.go
```diff
@@ -0,0 +1,347 @@
...
+	err := os.Remove(sypgp.PublicPath())
+	if err != nil {
+		t.Fatalf("Unable to remove public keyring: %v", err)
+	}
+}
+
+// RemoveSecretKeyring will delete your secret keyring.
+func RemoveSecretKeyring(t *testing.T) {
```
mem, line 77: don't export

e2e/internal/keyexec/keyexec.go
```diff
@@ -0,0 +1,347 @@
...
+	err := os.Remove(sypgp.SecretPath())
```
mem, line 78: same comment regarding sypgp.SecretPath() having to be empty.

e2e/internal/keyexec/keyexec.go
```diff
@@ -0,0 +1,347 @@
...
+	if err != nil {
+		t.Fatalf("Unable to remove secret keyring: %v", err)
+	}
+}
+
+func RemoveKeyring(t *testing.T) {
```
mem, line 84: don't export

e2e/internal/keyexec/keyexec.go
```diff
@@ -0,0 +1,347 @@
...
+	err := os.RemoveAll(sypgp.DirPath())
+	if err != nil {
+		t.Fatalf("Unable to remove keyring directory: %v", err)
+	}
+}
+
+// ImportKey will import a key from kpath.
+func ImportKey(t *testing.T, kpath string) (string, []byte, error) {
```
mem, line 92: don't export

I'll stop adding this comment. Most (all?) the functions don't need to be exported.

e2e/internal/keyexec/keyexec.go
```diff
@@ -0,0 +1,347 @@
...
+	e2e.LoadEnv(t, &testenv)
+
+	argv := []string{"key", "import", kpath}
+	execKey := exec.Command(testenv.CmdPath, argv...)
+
+	cm := fmt.Sprintf("%s %s", testenv.CmdPath, strings.Join(argv, " "))
+
+	b, err := execKey.CombinedOutput()
+
+	return cm, b, err
```
mem, line 102: shouldn't you call t.Error if err != nil?

e2e/internal/keyexec/keyexec.go
```diff
@@ -0,0 +1,347 @@
...
+	c, err := expect.NewConsole()
+	if err != nil {
+		t.Fatal("Unable to start new console: ", err)
+	}
+	defer c.Close()
+
+	exportCmd := []string{"key", "import", kpath}
+
+	cmd := exec.Command(testenv.CmdPath, exportCmd...)
+	cmd.Stdin = c.Tty()
```
mem, line 118: why?

e2e/internal/keyexec/keyexec.go
```diff
@@ -0,0 +1,347 @@
...
+	}()
+
+	err = cmd.Start()
+	if err != nil {
+		t.Fatal("Unable to start command: ", err)
+	}
+
+	// Send the passcode to singularity. The first one is the old
+	// password, the next two are the new passowrd.
+	c.Send("e2etests\n")
```
mem: the key you are importing will always have this passphrase? Shouldn't this be a parameter to the function?

e2e/internal/keyexec/keyexec.go
```diff
@@ -0,0 +1,347 @@
...
+
+	err = cmd.Wait()
+	cm := fmt.Sprintf("%s %s", testenv.CmdPath, strings.Join(exportCmd, " "))
+
+	return cm, outErr.Bytes(), err
+}
+
+// RunKeyCmd will run a 'singularty key' command, with any args that are set in commands.
+func RunKeyCmd(t *testing.T, cmdPath string, commands []string, file, stdin string) (string, []byte, error) {
```
mem: why is the "file" argument special? It's used only once. The user of this function could pass it as part of the arguments (commands)

e2e/internal/keyexec/keyexec.go
```diff
@@ -0,0 +1,347 @@
...
+			os.RemoveAll(tmpTestDir)
+			os.MkdirAll(tmpTestDir, os.ModePerm)
+			var c string
+			var b []byte
+			var err error
+
+			if tt.private {
+				c, b, err = ExportPrivateKey(t, filepath.Join(tmpTestDir, "export_key.asc"), "0\n", tt.armor)
+			} else {
+				c, b, err = RunKeyCmd(t, testenv.CmdPath, []string{"export"}, filepath.Join(tmpTestDir, "export_key.asc"), "0\n")
```
mem: what's the 0 entry?

the bit that is confusing is what does the keyring contain when this is running and why?

e2e/key/keyprivate/keyprivate.go
```diff
@@ -0,0 +1,166 @@
...
+		psk2    string
+		push    bool
+		succeed bool
+	}{
+		{
+			name:    "newpair",
+			user:    "genbye2etests\n",
+			email:   "westley@sylabs.io\n",
+			note:    "test key generated by e2e tests\n",
+			psk1:    "e2etests\n",
```
mem: psk2 is not set, is that intentional?

e2e/key/keyprivate/keyprivate.go
```diff
@@ -0,0 +1,166 @@
...
+		t.Run(tt.name, test.WithoutPrivilege(func(t *testing.T) {
+			t.Run("remove_private_keyring_before_newpair", test.WithoutPrivilege(func(t *testing.T) { keyexec.RemoveKeyring(t) }))
+			c, b, err := keyexec.KeyNewPair(t, tt.user, tt.email, tt.note, tt.psk1, tt.psk2, tt.push)
+			if tt.succeed {
+				if err != nil {
+					t.Log("Command that failed: ", c)
+					t.Log(string(b))
+					t.Fatalf("unexpected failure: %v", err)
+				}
+				keyexec.QuickTestExportImportKey(t)
```
mem: I'm going to guess that QuickTestExportImportKey _depends_ on being called exactly here.

What I mean is that since this function receives only t as parameter, it's tightly coupled with the key ring that is being created or modified here.

mem (changes requested):

e2e/internal/keyexec/keyexec.go
WestleyK: Its used in `e2e/key/keypublic/keypublic.go:172`.

e2e/internal/keyexec/keyexec.go
WestleyK: It is used in `e2e/key/keypublic/keypublic.go:144`. Should I just put the whole `RemoveDefaultPublicKey` function in keypublic.go?

e2e/internal/keyexec/keyexec.go
WestleyK: `sypgp.PublicPath()` should return a file, not a directory.

e2e/internal/keyexec/keyexec.go
WestleyK: Also should return a file.

e2e/internal/keyexec/keyexec.go
WestleyK: But these functions are used in different files, like `e2e/key/keypublic/keypublic.go`, Should I put these function(s) in that file? _that would make it very big..._

e2e/internal/keyexec/keyexec.go
WestleyK: I dont think so... Because in some tests, we are expecting it to fail.

e2e/internal/keyexec/keyexec.go
WestleyK: Why to what? Connecting `cmd.Stdin` to `c.Tty()` ?

e2e/internal/keyexec/keyexec.go
WestleyK: For simplicity its I hard coded it _for now_. Right now its not testing passwords/wrong passwords. Should I change this now?

e2e/internal/keyexec/keyexec.go
WestleyK: Its slightly easier to read in the table tests, imo. But it could be part of the `commands []string`.

e2e/internal/keyexec/keyexec.go
WestleyK: Key 0 is the newly created key by `newpair`. The keyring is cleared before generating a newpair. This function just tests that key from `newpair` can do what its supposed to.

e2e/key/keyprivate/keyprivate.go
WestleyK: Yes, if `psk2` is not set, `keyexec.KeyNewPair()` will use the same password as `psk1`.

e2e/internal/keyexec/keyexec.go
WestleyK: Okay, this function is only used in this file :+1:

e2e/internal/keyexec/keyexec.go
mem: you can place this value (`e2etests`, without the newline) in a variable so that at least there's a hint that this will change / needs to change.

e2e/internal/keyexec/keyexec.go
mem: please add a comment indicating that the keyring is empty before running this and that the newly added key gets to be entry 0 in all cases. Also, I believe there's an issue for the keys being zero-based, no?

e2e/key/keyprivate/keyprivate.go
mem: uhg.

:-)

e2e/internal/keyexec/keyexec.go
WestleyK: Yes, I should of done that before, but it does need the newline. I'll add a comment about that.

mem: you can add the new line as part of sending the password using expect.

What I mean is that the password does not contain the newline, so to make it clear that the variable contains the password (and not something else), it should not contain the newline.

WestleyK: Got it! okay :+1:

e2e/internal/keyexec/keyexec.go
WestleyK: Fixed now! :+1:

e2e/key/keyprivate/keyprivate.go
WestleyK: Okay, I'll remove `psk2`, its not necessary right now.

e2e/internal/keyexec/keyexec.go
WestleyK: Okay, I changed this test to export all the keys first, then try to import them, one by one. This way the keyring _could_ still have another entity in it.

e2e/key/keyprivate/keyprivate.go
WestleyK: I also included another parameter for this function, that will tell `QuickTestExportImportKey` what key to test.

e2e/internal/keyexec/keyexec.go
WestleyK: This makes it possible to use `c.Send()` to provide a standard input to the command, to my knowledge...

e2e/internal/keyexec/keyexec.go
```diff
@@ -172,118 +177,127 @@ func ExportPrivateKey(t *testing.T, kpath, num string, armor bool) (string, []by
...
 
-	return cmd, out, err
+	return cm, outErr.Bytes(), err
 }
 
 // QuickTestExportImportKey will export a private, and public key (0), and then import them. This is used after
-// generating a newpair.
-func QuickTestExportImportKey(t *testing.T) {
+// generating a newpair. keyNum is the key number to test. Its a string, so the number must be used in "", and
+// end with a '\n', eg. "1\n" will test key 1.
```
mem: it would be better to make keyNum an int, and have this function convert it to a string and add the necessary newline.

mem (changes requested):

GodloveD:
This PR will likely need to be redone now that the testing framework has been modified.

jscook2345:
If we could break it up a bit to that would be ideal. `key list` pr, `key pull` pr, etc.

WestleyK:
Closing. I will start from scratch and breakup this PR into smaller, cleaner PRs using the new E2E framework.

PR #3542, opened 2019-05-13
Clean up DecryptKey

mem:


pkg/sypgp/sypgp.go
```diff
@@ -515,20 +516,20 @@ func GenKeyPair(keyServiceURI string, authToken string) (entity *openpgp.Entity,
 
 // DecryptKey decrypts a private key provided a pass phrase
 func DecryptKey(k *openpgp.Entity, message string) error {
+	if !k.PrivateKey.Encrypted {
+		return errNotEncrypted
+	}
```
WestleyK, line 521: I still honestly think this should not return a err, but thats my opinion and maybe I'm wrong.

WestleyK (approved):
Looks good to me :+1:

PR #3549, opened 2019-05-14
[WIP] Improvement cache

### LINKED ISSUE #3525: Clarify the semantics behind the image cache
## Version of Singularity:

Master

### Expected behavior

Protect from deleting user's data when cleaning the cache and the location of the cache is specified by the user. Right now, everything in the location is deleted. For instance, if the user sets the environment variable to point at the home directory, everything in the home directory is lost. Furthermore, even after reading the documentation, I was expecting that Singularity will create a directory in the location that I specify through the environment variable and Singularity to handle this directory.

### Actual behavior

The user can specify right now an environment variable to change the location of the image cache. The current implementation had a problem: when the cache is cleaned, everything in the cache (files and directories) are simply deleted, including the user's data.

Set the environment variable to your home; clean your cache; the content of your entire home directory is gone.

gvallee:

Clarify the semantics related to the image cache. 
Note that the code is very similar to a previous PR. The goals are:
- avoid the deletion of user data when the environment variable that specifies the location of a cache is misused (previous the entire content of the directory specified by the environment variable was deleted; if it pointed to the home directory of the user, the entire home directory was deleted),
- improve the test of the `internal/pkg/client/cache` package,
- provide a new set of API for easing the testing of other packages relying on caches (e.g., `internal/app/singularity/cache_clean_linux.go` and `internal/app/singularity/cache_list_linux.go`).

- Fixes #3525 


internal/pkg/client/cache/dir.go
```diff
@@ -22,75 +46,268 @@ const (
...
+	}
+
+	return false, nil
+}
+
+// Figure out where the cache directory is.
+func getCacheBasedir() (string, error) {
+	// If the user defined the special environment variable, we use its value as base directory.
+	basedir := os.Getenv(DirEnv)
+	if basedir == "" {
```
mem: Idiomatic Go prefers to "shift left", meaning:

```
basedir := os.Getenv(DirEnv)
if basedir != "" {
    return basedir, nil
}
// all the code in the basedir == "" case goes here
```

internal/pkg/client/cache/dir.go
```diff
@@ -22,75 +46,268 @@ const (
...
+
+// Init initializes a new cache within a given directory
+func hdlInit(baseDir string) (*SingularityCache, error) {
+	rootDir := getCacheRoot(baseDir)
+	if err := initCacheDir(rootDir); err != nil {
+		return nil, fmt.Errorf("failed initializing caching directory: %s", err)
+	}
+
+	newCache := new(SingularityCache)
+	savedDirEnv := os.Getenv(DirEnv)
```
mem, line 142: what is the purpose of this and the next two lines? (I can read what the code is doing, I don't understand why it's doing that)

internal/pkg/client/cache/dir.go
```diff
@@ -22,75 +46,268 @@ const (
...
+	// by the user.
+	rootDir string
+
+	// State of the cache. We enable manual change of the state mainly for
+	// testing.
+	State string
+
+	// Default specifies if the handle points at the default image cache or
+	// not. This enables quick lookup.
+	Default bool
```
mem, line 95: I can't find uses of this...

What's the idea here?

My expectation was that there would be a single variable pointing to the location of the cache (rootDir above) and nothing else. I can see why Library, OciTemp,OciBlob, Net and Shub are useful. State could be useful, but quickly searching thru the code, it seems that State = Initialized is equivalent with rootDir != ""

In the same sense, BaseDir does not seem to be useful as it's equivalent with rootDir != ""

internal/pkg/client/cache/dir.go
```diff
@@ -22,75 +46,268 @@ const (
...
 
+// initCacheDir initializes a sub-cache within a cache, e.g., the shub sub-cache.
 func initCacheDir(dir string) error {
-	if _, err := os.Stat(dir); os.IsNotExist(err) {
+	if fInfo, err := os.Stat(dir); os.IsNotExist(err) {
 		sylog.Debugf("Creating cache directory: %s", dir)
 		if err := fs.MkdirAll(dir, 0755); err != nil {
 			return fmt.Errorf("couldn't create cache directory %v: %v", dir, err)
 		}
+	} else if fInfo != nil && !fInfo.IsDir() {
```
mem: this should check that there was no error.

The cases are:

- Error, check that it's "not exists", create directory
- Error, something else, error out because stat can fail for a number of reasons, most unrecoverable here
- No error. `dir` exists, check that it is a directory.

In other words, don't check that fInfo is not nil, check that err is not nil.

```
switch {
case os.IsNotExist(err):
    // create directory
case err != nil:
    // bail out
case !fInfo.IsDir():
    // bail out, dir exists but it is not a directory
default:
    // dir exists, it's a directory
}
```

internal/pkg/client/cache/dir.go
```diff
@@ -22,75 +46,268 @@ const (
...
+	// RootDefault is the default cache root directory created within the base
+	// directory. This value is not supposed to be set by the user.
+	rootDefault = "cache"
+
+	// StateInitialized represents the state of a give cache after successful
+	// initialization
+	StateInitialized = "initialized"
+
+	// StateInvalid represents the state of an invalid cache
+	StateInvalid = "invalid"
```
mem: this value is only used in tests, it's never used in the code for this package. To me that means that the zero value of the State field should signify invalid, and this is not needed. Since this is exported and the state field is exported, this also means that the client code could set State to StateInvalid, which is probably not useful.

What this means is that State should be a boolean since there are only two states with opposite meanings.

But please see below first.

mem (changes requested):
Only looked at internal/pkg/client/cache

internal/pkg/client/cache/dir.go
mem: Sorry, I had never noticed that these comments are not presented in file-order but in write-order. I wrote this after the comment on line 87. "below" refers to that.

internal/pkg/client/cache/dir.go
gvallee: I will not argue about this and see your point; my goal was to preserve the initial code as much as possible (too many modifications increase the risk of introducing new bugs). I will change the code to a switch statement, the code will be more readable.

internal/pkg/client/cache/dir.go
gvallee: Yes one of the main goals is to ease testing, that is 100% correct. I will switch to a boolean, the semantic remains the same.

internal/pkg/client/cache/dir.go
gvallee: `Default` is used in the context of testing. `Default` is mainly used in `cache_test.go` which states `A set of functions internal to the package for testing. Public helper functions for testinng are in testing.go`, more specifically when cleaning a temporary cache.
Comments for the `cleanupCache()` state it as:
```
// cleanupCache will free/destroy a cache ONLY if it is NOT the default cache.                                                                  
// The goal here is not interfer with the default cache that is populating
// while using Singularity when running tests. Used in conjunction with setupCache().
```
So basically what it means is that I am trying to *not* clean the cache of developers when executing unit tests. In my mind developers may not want unit tests to mess up their image cache (by deleting it entirely). Obviously I will update the comments associated with `Default` to avoid having to read the entire code to understand, it is very unclear at the moment.
 
As BaseDir not being useful, you are referring to your expectations and I personally do not know what they are but I disagree with this: it is a key piece for testing and it makes the cache handle self-contained (`internal/pkg/client/cache/testing.go`). This is in fact the initial motivation for modifying the implementation: I will use this to revisit the testing code in `internal/app/singulairty/`, we do *not* have all the required capabilities to implement adequate tests relying on caches. BaseDir is *not* equivalent to `rootDir != ""` since the value of BaseDir is necessary. So if it is not used right now, it will be very soon (in fact, as soon as this PR is merged).

internal/pkg/client/cache/dir.go
gvallee: It creates a new cache handle and save the environment variable associated to it (we could save `BaseDir`) to avoid conflicts between the` baseDir` specified as parameter of the `hdlInit()` function and the value of the environment variable. This was initially done when `hdlInit()` was an exported function. 

The basic problem I try to address is:
1/ the developer sets the `DirEnv` environment variable for development,
2/ a unit test within Singularity *needs* to setup a temporary cache for testing. Obviously, it would be better if we do not actually use the cache specified by the user but instead create a new cache from scratch (control over the context of the test execution).
3/ By having that code, we can create a new cache by calling `hdlInit()`, save the current value of the environment variable, set the environment variable to point to the new temporary cache (control over the test's context which should be as close as possible to what a normal case would be), and once the test completed, restore the environment variable.
In other terms, the code is meant to include some mechanisms to overly safe regarding the execution of a test, as well as a good control over the execution context of a test.

Since it is not an exported function, we have two options:
- let the caller *always* ensures that there is no conflict between the environment variable and the `baseDir` passed to `hdlInit()`, even in the context of the creation of a temporary cache for testing,
- have a few lines of code, as it is now, to ensure everything is fine.

Once we decide which way is better, I will improve both the implementation and comments.

internal/pkg/client/cache/dir.go
gvallee: Sure I can change this but FYI, I am using quite a few linters to check my code and none of them complained about it. Do you know a linter that actually catches these things?

gvallee: The code has been modified

internal/pkg/client/cache/dir.go
gvallee: Comments added to the code

internal/pkg/client/cache/dir.go
gvallee: The code has been modified.

internal/pkg/client/cache/dir.go
gvallee: The code is now using a boolean for `State`

internal/pkg/client/cache/dir.go
mem: If I understand correctly what you are saying, you are trying to protect against this:

```
$ cat test.go
package main

import (
        "fmt"
        "os"
)

func main() {
        oldfoo := os.Getenv("FOO")
        os.Setenv("FOO", "new value "+oldfoo)
        newfoo := os.Getenv("FOO")
        fmt.Printf("oldfoo=%q\nnewfoo=%q\n", oldfoo, newfoo)
}
```

A process cannot modify the environment of its parent. If you run the above program, you'll see:

```
$ export FOO="old value"
$ echo $FOO
old value
$ go run test.go
oldfoo="old value"
newfoo="new value old value"
$ echo $FOO
old value
```

If this is _not_ what you are saying, then obviously I still don't understand what this code is supposed to be doing. If this code is really necessary, it should have a very clear comment explaining what is its purpose.

To try to explain what it is that I don't understand: ideally, upon creation of the cache structure, the code should look at the environment exactly _once_. If the variable is set, it should use that value. If the variable is not set, it should use the default. From that point on, the rest of the code should either use the value for the cache base directory set in the cache structure or it should use whatever is passed as arguments to functions. In particular, it should never look at the environment variable again. With that in mind, what I don't understand is why the code should be modifying the environment, much less saving old values.

internal/pkg/client/cache/dir.go
mem: Unit tests should _never_ operate against the default cache directory (as it is impossible to assert the state of that directory). That will leave a particular path of code untested, yes, but that path should be as small as:

```
basedir := os.Getenv("SINGULARITY_CACHEDIR")
if basedir == "" {
    // this branch never gets tested during unit testing
    // defaultCacheDir() itself _could_ be tested by itself
    basedir = defaultCacheDir()
}
cacheDir := filepath.Join(basedir, "cache")
```

I'm judging the lack of usefulness of BaseDir from the point of view of separation of concerns, not personal preference. That's the reason why I am asking what the idea is here.

internal/pkg/client/cache/dir.go
```diff
@@ -78,12 +71,17 @@ type SingularityCache struct {
 	// by the user.
 	rootDir string
 
-	// State of the cache. We enable manual change of the state mainly for
-	// testing.
-	State string
+	// ValidState specifies if the cache is in a valid state or not. This is
+	// mainly used for testing, where a unit test can switch a cache's state
+	// from valid to invalid in order to reach error cases.
+	ValidState bool
```
mem, line 84: if this exists for testing, it should not be exported.

internal/pkg/client/cache/dir.go
```diff
@@ -78,12 +71,17 @@ type SingularityCache struct {
...
 
 	// Default specifies if the handle points at the default image cache or
-	// not. This enables quick lookup.
+	// not. This enables quick lookup. This is for instance used in the
+	// context of unit tests execution to make sure we do not delete the
+	// image cache in '$HOME/.singularity/cache' (developers may not
+	// appreciate that some unit tests always delete their default image
+	// cache).
 	Default bool
```
mem, line 95: again, if this exists for testing, it should not be exported.

mem (commented):
Only looking at internal/pkg/client/cache/dir.go

internal/pkg/client/cache/dir.go
gvallee: I would agree with you that we should never use the default cache but so far we were most of the time running tests using the default cache directory; it is exactly what I am trying to change here. But the point here is: having a base directory allows us to change the configuration of the cache to reach some error cases. Not having this will create a headache for testing. If you disagree with having the base directory there, I do not really see a path forward at the moment.
I really do not know what to do: if we remove the base directory, I am back to the initial situation when it is about reaching error caches and I do not see alternative options.

Finally, my intent was absolutely not to question your motivation. I never suggested or even considered that your comments were based on personal preferences but at the same time, it is not possible for me to understand what your expectation precisely mean, and will not take the liberty to guess. I hope I do not go through as having a negative attitude that was clearly not my intent.

internal/pkg/client/cache/dir.go
gvallee: No I do not try to prevent this, I know how environment variables are handled in the context of processes. I try to prevent the situation where a complex test is composed of two parts, one focusing on a higher-level capability which ultimately uses a cache under the cover. An example would be unit tests for `internal/app/singularity/cache_clean|list_linux.go`. Furthermore, I want to make sure that I can "trick" the configuration of the cache to reach corner cases. All this in the context of a single unit test, i.e., a single process. So the high-level part of the test might set an environment variable (it is what privilege functions such as `DropPrivilege()` actually do) and once we reach the portion of the code dealing with the cache, the entire cache mechanism will deal with all the detail to create a temporary cache from scratch in a safe manner.

internal/pkg/client/cache/dir.go
gvallee: It is also meant to be used in the context of testing of *other* packages that rely on caches (e.g., `singularity/internal/app/singularity/cache_clean_linux.go`) so if it is not exported, I cannot reach all required error cases. The initial motivation is to better that the current package, other packages and avoid the deletion of packages. This was better described in the initial PR so I will add it here, as well as improving the comments in the code.

internal/pkg/client/cache/dir.go
gvallee: Again, this is also meant to be used for testing other packages relying on cache. Not only the testing of `internal/pkg/client/cache`.

internal/pkg/client/cache/dir.go
gvallee: Comments have been updated

internal/pkg/client/cache/dir.go
gvallee: Comments have been updated

gvallee:
I will split the PR based on the three goals: a first PR to avoid the deletion of user's data; a second PR to improve testing this package; and finally a third PR to see how we can facilitate testing of other packages relying on caches. As a note about the third PR, we may need to have a broader discussion in order to avoid writing too much code that is only for testing, outside of unit tests.

I will also mark this PR as WIP so it is not deleted but not merged either and keep it open for reference until completion of these tasks.

gvallee:

PR #3551, opened 2019-05-15
Add a logger Close method and fix logger test sync issue

cclerget:

Explicitly close logger to force a flush of data to the log file

- Duplicate fix of #3527 as suggested by @mem 

### DIFF
--- a/internal/pkg/instance/logger.go
+++ b/internal/pkg/instance/logger.go
@@ -55,12 +55,11 @@ var LogFormats = map[string]LogFormatter{
 
 // Logger defines a file logger.
 type Logger struct {
-	file      *os.File
 	fm        sync.Mutex // protect file
-	filename  string
+	file      *os.File
 	formatter LogFormatter
-	closers   []closer
 	cm        sync.Mutex // protect closers array
+	closers   []closer
 }
 
 // NewLogger instantiates a new logger with formatter and return it.
@@ -68,7 +67,6 @@ func NewLogger(logPath string, formatter LogFormatter) (*Logger, error) {
 	logger := &Logger{
 		formatter: formatter,
 		closers:   make([]closer, 0),
-		filename:  logPath,
 	}
 
 	if logger.formatter == nil {
@@ -128,7 +126,7 @@ func (l *Logger) scan(stream string, pr *io.PipeReader, pw *io.PipeWriter, dropC
 		scanner.Split(l.scanOutput)
 	}
 
-	var wg sync.WaitGroup
+	wg := new(sync.WaitGroup)
 
 	wg.Add(1)
 
@@ -139,6 +137,7 @@ func (l *Logger) scan(stream string, pr *io.PipeReader, pw *io.PipeWriter, dropC
 			l.fm.Lock()
 			// means ReOpenFile has failed proceed with cleanup
 			if l.file == nil {
+				l.fm.Unlock()
 				break
 			}
 			if !dropCRNL {
@@ -184,11 +183,10 @@ func (l *Logger) Close() {
 // ReOpenFile closes and re-open log file (eg: log rotation).
 func (l *Logger) ReOpenFile() error {
 	l.fm.Lock()
-
+	filename := l.file.Name()
 	l.file.Sync()
 	l.file.Close()
-
-	err := l.openFile(l.filename)
+	err := l.openFile(filename)
 	l.fm.Unlock()
 
 	if err != nil {
--- a/internal/pkg/instance/logger.go
+++ b/internal/pkg/instance/logger.go
@@ -56,10 +56,11 @@ var LogFormats = map[string]LogFormatter{
 // Logger defines a file logger.
 type Logger struct {
 	file      *os.File
+	fm        sync.Mutex // protect file
 	filename  string
 	formatter LogFormatter
 	closers   []closer
-	sync.Mutex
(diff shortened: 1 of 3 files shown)


internal/pkg/instance/logger.go
```diff
@@ -114,39 +111,68 @@ func (l *Logger) scanOutput(data []byte, atEOF bool) (advance int, token []byte,
...
 		scanner.Split(l.scanOutput)
 	}
 
-	for scanner.Scan() {
-		l.fileMutex.Lock()
-		if !dropCRNL {
-			fmt.Fprint(l.file, l.formatter(stream, r.Replace(scanner.Text())))
-		} else {
-			fmt.Fprint(l.file, l.formatter(stream, scanner.Text()))
+	var wg sync.WaitGroup
```
mem: please see [the documentation for WaitGroup](https://godoc.org/sync#WaitGroup) and note where it says: "A WaitGroup must not be copied after first use."

By creating a closure on L#149, the code is copying the WaitGroup instance after the first use on L#128.

The proper fix for this is to use `wg := new(sync.WaitGroup)`

```suggestion
	wg := new(sync.WaitGroup)
```

internal/pkg/instance/logger.go
```diff
@@ -55,14 +56,17 @@ var LogFormats = map[string]LogFormatter{
 // Logger defines a file logger.
 type Logger struct {
 	file      *os.File
-	fileMutex sync.Mutex
+	filename  string
 	formatter LogFormatter
+	closers   []closer
+	sync.Mutex
```
mem: the recommended pattern when using mutexes is that they protect exactly one thing, and that they are declared before the thing they protect, for example:

```
type s struct{
    m sync.Mutex // this mutex protects access to the next field
    f int
}
```

because of the usage, I'm going to guess that this mutex is protecting the entire Logger. A comment to that effect would be useful so that people looking at this code in the future don't have to guess.

internal/pkg/instance/logger.go
```diff
@@ -76,27 +80,20 @@ func NewLogger(logPath string, formatter LogFormatter) (*Logger, error) {
...
 
-func closeFile(file *os.File) {
-	file.Close()
-}
-
 func (l *Logger) openFile(path string) (err error) {
 	oldmask := syscall.Umask(0)
 	defer syscall.Umask(oldmask)
 
+	l.Lock()
```
mem: If os.OpenFile returns an error, the statement l.file.Name() will panic.

It would be better to do this:

```
fh, err := os.OpenFile(...)
if err != nil {
    return err
}

l.Lock()
l.file = fh
l.filename = fh.Name()
l.Unlock()
```

internal/pkg/instance/logger.go
```diff
@@ -114,39 +111,68 @@ func (l *Logger) scanOutput(data []byte, atEOF bool) (advance int, token []byte,
...
-			fmt.Fprint(l.file, l.formatter(stream, r.Replace(scanner.Text())))
-		} else {
-			fmt.Fprint(l.file, l.formatter(stream, scanner.Text()))
+	var wg sync.WaitGroup
+
+	wg.Add(1)
+
+	go func() {
+		for scanner.Scan() {
+			l.Lock()
```
mem: this is the bit where I'm a bit confused. What's this lock protecting? The client calling ReOpenFile while this is running? A comment to that effect would be needed here. The other thing this could be protecting is having multiple writers (with NewWriter) trying to write at the same time. Same thing: this needs a comment.

Also, if that's what's going on here, note that calling ReOpenFile might end up in a situation where the file handle is no longer valid (l.file was closed, a new file handle could not be obtained). After `l.Lock()` there should be a check like `if l.file == nil { break }`.

internal/pkg/instance/logger.go
```diff
@@ -114,39 +111,68 @@ func (l *Logger) scanOutput(data []byte, atEOF bool) (advance int, token []byte,
...
 // ReOpenFile closes and re-open log file (eg: log rotation).
 func (l *Logger) ReOpenFile() {
-	l.fileMutex.Lock()
-	defer l.fileMutex.Unlock()
-
-	path := l.file.Name()
-
+	l.Lock()
+	l.file.Sync()
 	l.file.Close()
```
mem, line 188: this bit here invalidates the file handle and releases the lock. Since l.Unlock() is a goroutine yield point, it's entirely possible that the scan gorotines waiting on the lock will get unblocked and proceed trying to write to the now-closed file handle.

This means that openFile has to be changed to that it must be called with the lock held.

mem (changes requested):

internal/pkg/instance/logger.go
cclerget: https://golang.org/ref/spec#Function_literals mentioned that :

> Function literals are closures: they may refer to variables defined in a surrounding function. Those variables are then shared between the surrounding function and the function literal, and they survive as long as they are accessible.

I may not fully understand what you mean by copy, because `scan` function can be called multiple time ?

internal/pkg/instance/logger.go
cclerget: Ah right !

internal/pkg/instance/logger.go
cclerget: Will add

mem (approved):

PR #3564, opened 2019-05-16
Add syfs.ConfigDir() to obtain singularity's config directory

mem:
There are multiple locations in the code that have their own method to
define where the singulary configuration directory exists, all trying to
do more or less the same thing.

Move that code to a single location that can be reused by the rest of
the code. Call the package "syfs" to express that this represents
singularity's file system layout.

Change direct references to .singularity to syfs.ConfigDir().

### DIFF
--- a/pkg/syfs/syfs.go
+++ b/pkg/syfs/syfs.go
@@ -1,0 +1,70 @@
+// Copyright (c) 2019, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+// Package syfs provides functions to access singularity's file system
+// layout.
+package syfs
+
+import (
+	"os"
+	"os/user"
+	"path/filepath"
+	"sync"
+
+	"github.com/sylabs/singularity/internal/pkg/sylog"
+)
+
+const singularityDir = ".singularity"
+
+// cache contains the information for the current user
+var cache struct {
+	sync.Once
+	configDir string // singularity user configuration directory
+}
+
+// ConfigDir returns the directory where the singularity user
+// configuration and data is located.
+func ConfigDir() string {
+	cache.Do(func() {
+		cache.configDir = configDir()
+		sylog.Debugf("Using singularity directory %q", cache.configDir)
+	})
+
+	return cache.configDir
+}
+
+func configDir() string {
+	user, err := user.Current()
+
+	if err != nil {
+		sylog.Warningf("Could not lookup the current user's information: %s", err)
+
+		cwd, err := os.Getwd()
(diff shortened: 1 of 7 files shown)


pkg/syfs/syfs.go
```diff
@@ -0,0 +1,64 @@
+// Pakcage syfs provides functions to access singularity's file system
+// layout
```
WestleyK: Is there that copyright thing thats supposed to go here?

```
```

pkg/syfs/syfs.go
```diff
@@ -0,0 +1,64 @@
...
+func ConfigDir() string {
+	cache.Do(func() {
+		cache.configDir = configDir()
+		sylog.Debugf("Using singularity directory %q", cache.configDir)
+	})
+
+	return cache.configDir
+}
+
+func configDir() string {
```
jchavezlavalle, line 38: shouldn't this function return the error as well? Also alternatively, you could also merge both functions `configDir` and `ConfigDirForUsername` into one, when the parameter passed is null then obtaining the current one, otherwise do the lookup on the passed username @mem

jchavezlavalle (changes requested):

pkg/syfs/syfs.go
ikaneshiro: I agree that we should return an error if we cannot lookup the current user. I don't like the idea of potentially creating user config directories at different locations based on `cwd` if we cannot resolve the user. I think that's enough reason to return an error and bail on execution.

mem: Thanks for looking at this.

The reason why configDir does not return an error is because it covers all the possible error conditions and provides fallbacks for it. I don't agree with what this is doing, but it's been ported over from the sypgp package. In particular, the choice of falling back to a .singularity directory in the current directory if the current user information is not available seems strange to me. The choice to return a relative .singularity path if the current directory cannot be obtained seems even stranger.

ConfigDirForUsername behaves differently because it's possible that the username in the argument does not exist, so it makes sense to return an error in that case. The _current_ user _should_ exist, and to be consistent with the behavior of ConfigDir, no error is returned in that case ever, it simply falls back to whatever ConfigDir returns. It would be strange to have ConfigDirForUsername(current_user) return something different than ConfigDir().

pkg/syfs/syfs.go
mem: Right! Fixing.

WestleyK (approved):
Looks good to me :+1:

jchavezlavalle (approved):

jscook2345:
@mem 

You broke a test when you made your changes.

`internal/pkg/client/cache/dir_test.go:28:40: undefined: RootDefault`

You can probably just import your new pkg and use that instead.

Thanks for this PR, good way to avoid tech debt!

PR #3574, opened 2019-05-17
Create scripts/go-test script during build

mem:
scripts/go-test is simply a wrapper for go test that uses the same flags
as the singularity build (-tags, -mod, etc), and also passes flags that
make sense in this context (-count=1, -cover, etc).

Use a Go program to do the replacement in scripts/go-test.in in order to
avoid having to mess around with quotes and whatnot.


scripts/expand-env.go
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) 2019, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+// expand-env is a simple program that reads stdin, replaces all
+// occurences of @VAR@ by the corresponding value of VAR in the current
+// environment, and writes the result to stdout.
+package main
```
WestleyK, line 9: It it worth putting a newline right before the `package ...`

```suggestion

package main
```

WestleyK (approved):
Looks good to my knowledge.

scripts/expand-env.go
mem: Thanks for looking at this.

> It it worth putting a newline right before the `package ...`

If you add a blank before `package` that becomes a comment. If you omit the blank, that's the package's documentation. It does not make a difference in this case because this piece of documentation cannot be accessed with the usual tools, but it's nice to follow conventions anyways.

WestleyK: Good to know, okay :+1:

scripts/expand-env.go
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) 2019, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+// expand-env is a simple program that reads stdin, replaces all
+// occurences of @VAR@ by the corresponding value of VAR in the current
```
mem:
```suggestion
// occurrences of @VAR@ by the corresponding value of VAR in the current
```

PR #3576, opened 2019-05-17
Change default log level to info

### LINKED ISSUE #3569: debug output occurs due to log file setting race condition
## Version of Singularity:

master

```
$ singularity version
DEBUG   [U=1000,P=569]     func1()                       Using singularity directory "/home/vagrant/.singularity"
```

This is occurring because the log level (which defaults to `debug`) is not being set to normal before the first use of `syfs.ConfigDir()`, which occurs very early because it's called to assign a value to a constant:

```
$ ack 'ConfigDir'
cmd/internal/cli/singularity.go
127:    return path.Join(syfs.ConfigDir(), "sylabs-token")

$ ack 'getDefaultTokenFile'
31:var defaultTokenFile = getDefaultTokenFile()
```

Since the value of the token file must be determined at run-time, it probably should not be a constant.

mem:
Parts of the code that do early logging (like those called before main
via init functions or variable initialization) are seeing a default log
level of debug and therefore logging at that level even if -d wasn't
specified.

Change this to info, which is the default using by the CLI during flag
processing.

If you need to see those early debug messages, you cannot use -d, you
MUST use SINGULARITY_MESSAGELEVEL=5.

Fixes: #3569

### DIFF
--- a/internal/pkg/sylog/sylog.go
+++ b/internal/pkg/sylog/sylog.go
@@ -68,11 +68,11 @@ var loggerLevel messageLevel
 func init() {
 	_level, ok := os.LookupEnv("SINGULARITY_MESSAGELEVEL")
 	if !ok {
-		loggerLevel = debug
+		loggerLevel = info
 	} else {
 		_levelint, err := strconv.Atoi(_level)
 		if err != nil {
-			loggerLevel = debug
+			loggerLevel = info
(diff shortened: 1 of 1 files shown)


WestleyK (approved):
Looks good to me, and seems to work! :+1:

PR #3578, opened 2019-05-17
Port to scs library client

EmmEff:

Modifications to use `github.com/sylabs/scs-library-client`

- Fixes # 3424


internal/pkg/library/library.go
```diff
@@ -0,0 +1,126 @@
...
+
+const defaultTag = "latest"
+
+type progressCallback func(int64, io.Reader, io.Writer) error
+
+// NormalizeLibraryRef strips off leading "library://" prefix, if any, and
+// appends the default tag (latest) if none specified.
+func NormalizeLibraryRef(libraryRef string) string {
+	var ir string
+	if strings.HasPrefix(libraryRef, "library://") {
```
ikaneshiro: This seems like a good place to use `strings.TrimPrefix()` instead

internal/pkg/library/library.go
```diff
@@ -0,0 +1,126 @@
...
+	if !strings.Contains(ir, ":") {
+		return ir + ":" + defaultTag
+	}
+	return ir
+}
+
+// DownloadImage is a helper function to wrap library image download operation
+func DownloadImage(ctx context.Context, c *client.Client, imagePath, libraryRef string, callback progressCallback) error {
+	// reassemble "stripped" library ref for scs-library-client
+	validLibraryRef := "library:///" + libraryRef
```
ikaneshiro, line 35: I assume the three `/`'s is intentional here? Will we ever have a host in the reference? If so, this will make it part of the path correct?

ikaneshiro (commented):
Looking good to me, just a suggestion and a question about library URIs

internal/pkg/library/library.go
EmmEff: The `scs-library-client` supports URIs containing a host name, however it is not currently used. My intention here is to normalize the library ref on the Singularity side, such that the common Parse() function in the `scs-library-client` can be used going forward. I am not sure how/if this "enhanced" URI format will be used in Singularity at this point.

internal/pkg/library/library.go
EmmEff: You learn something new every day :) that's exactly what I need. Thanks!

ikaneshiro (approved):

PR #3583, opened 2019-05-20
Don't call naked singularity

mem:
In tests, to make sure that we are using the correct singularity binary
(the one under test), we must use testenv.CmdPath, instead of a naked
'singularity'.


GodloveD:
davidcensored

GodloveD:
I have a naive question.  Instead of fixing this in every single line where `singularity` is called, is it possible to make it a feature of the test harness itself that the appropriate Singularity binary is always used?

ArangoGutierrez (approved):

WestleyK (approved):
Looks good to me :+1:

mem:
> I have a naive question. Instead of fixing this in every single line where `singularity` is called, is it possible to make it a feature of the test harness itself that the appropriate Singularity binary is always used?

The test harness already provides a variable that points to the correct singularity binary. This is what this PR is doing.

I guess you are asking why don't we modify the PATH so that `singularity` picks up the correct singularity binary.

In general I try to avoid setting up things using the environment. If you see this in log output:

    singularity version

you _cannot_ just cut and paste it, because the environment under which that command is running might be different from the environment you have.

If you see this instead:

    /some/path/singularity version

you _can_ cut and paste that and expect the output to be the same. The output might still depend on environment variables, but at least you know the program you are running is the same that the test ran.

GodloveD:
Yes, I understand what you are saying.  My question is actually this:

Why doesn't the Singularity test harness automatically prepend the correct path to the installed version of Singularity to every `singularity` call?  In this PR you had to copy the same change into 9 different lines.  Seems antithetical to the concept of Don't Repeat Yourself (DRY).  

I assume that we will only ever want to run a single version of the installed singularity binary with the e2e test harness.  So why not have the harness expressly prepend the path every time `singularity` is invoked instead of having to format that string in every single line?  

EDIT:  I agree that you don't want to do this by manipulating the PATH.  But the test harness already creates this env var, so it seems that it should just prepend it to all the `singularity` calls.

PR #3584, opened 2019-05-20
Add exec.Command wrapper for testing

mem:
In order to make sure we handle error conditions uniformly and log all
the required output, command and arguments, introduce a exec.Command
wrapper that takes care of those details.

### DIFF
--- a/e2e/internal/exec/exec.go
+++ b/e2e/internal/exec/exec.go
@@ -1,0 +1,89 @@
+package exec
+
+import (
+	"bytes"
+	stdexec "os/exec"
+	"strings"
+	"syscall"
+	"testing"
+)
+
+type Cmd struct {
+	*stdexec.Cmd
+	cmdStr         string
+	combinedOutput bytes.Buffer
+}
+
+func Command(name string, arg ...string) *Cmd {
+	return &Cmd{
+		Cmd: stdexec.Command(name, arg...),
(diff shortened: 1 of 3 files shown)


e2e/internal/exec/exec.go
```diff
@@ -0,0 +1,89 @@
+package exec
```
WestleyK, line 1: I think it need that copyright thing up here:

```
```

e2e/internal/exec/exec_test.go
```diff
@@ -0,0 +1,28 @@
+package exec
```
WestleyK, line 1: Here too

WestleyK (changes requested):
Couple suggestions :+1:

jscook2345 (changes requested):
If the plan is to use this instead of modifying the code I added under `internal/pkg/test/cmd.go`, perhaps we could remove that code and replace the work in the `help` e2e test package with this?

mem:
> If the plan is to use this instead of modifying the code I added under `internal/pkg/test/cmd.go`, perhaps we could remove that code and replace the work in the `help` e2e test package with this?

Aha!

I didn't find that there. Let me take a look and see how much overlap there is, but after a quick glance I think we can expand on the already existing code to handle other details.

Issue #3592, opened 2019-05-21
Improve cache management in the context of `DropPrivilege()` and `EnsurePrivilege()`


gvallee:
## Version of Singularity:

Master.

### Expected behavior

When implementing unit tests, we should always use either `DropPrivilege()` or `EnsurePrivilege()`, which should provide a clean environment for every tests, including a clean image cache specific to the test.

### Actual behavior

`DropPrivilege()` and `EnsurePrivilege()` create a specific directory in `/tmp` that is used for the execution of a test. Furthermore, that directory is used as location for the image cache. At the moment, that directory is not entirely cleaned at the end a given unit tests and re-used by the next unit test to be executed. Using a directory for a unit test that is not empty increases the risk for failures and decrease the level of control over the execution of the tests.

### Steps to reproduce behavior

First execute `make testall`; then `ls -l /tmp/WithPrivilege` or `ls -l /tmp/WithoutPrivilege`, you can see an image cache from the unit tests.

gvallee:
Since I am working on issue #3525 , I was able to add code to the unit tests to force the creation and use of a clean temporary image cache for each test that was failing, and I can confirm that it solves the failures I could see while running `make testall`.

I think the two questions to be answered are:
1. Do we want to always use the same directories for the unit tests, i.e., `/tmp/WithPrivilege` or `/tmp/WithoutPrivilege`? If so, should we ensure that the entire content of these directories is wiped out after the execution of every single unit tests? If no, do we want to rely on `ioutil.TempDir()` to create a unique temporary directory for each unit test, maybe using a `WithPriv` or `WithoutPriv` prefix to differentiate privileged and unprivileged unit tests?
2. Whichever option we choose in the previous point, do we want to give the option to test developers to create temporary clean image caches, instead of assuming that a cache is provided by default? It would help if unit tests end up using more than one cache.

gvallee:
We should also check with people who worked on `DropPrivilege()` and `EnsurePrivilege()` (@cclerget?) what are the requirements for having predefined directories and reuse them instead of dynamically create these directories.

gvallee:
Comment from @mem during a discussion: may we do not want these two functions to do anything about the cache because their names do not let developers know that it does anything about configuring the testing environment.
So the questions to answer are now:
1/ Should these two functions set anything regarding the test context, e.g., the cache location? If not do we want to have another function that would do that?
[...]

PR #3595, opened 2019-05-21
Revisit the implementation of image cache to prevent image deletion and inconsistent caches

gvallee:

This PR addresses two problems:
1/ Avoid the deletion of user data when the environment variable used to specify the cache location is not used appropriately.
2/ Ensure that the cache location cannot be changed during the execution of a command.

This prevents data deletion by adding the concept of cache root, i.e., a folder managed by Singularity to actually host the cache. When the root is deleted while cleaning the cache, no user data can be deleted.

This prevents the change of the cache location by creating an handle to manage cache. The configuration of the cache is set while creating the handle and remain valid while the handle is valid. This prevents from having the change of the value of the environment variable for the location of the cache impact the cache during the execution of a command.

This PR does *not*:
- address issue #3592, which is not introduced by this PR. In fact, this PR could be used to implement work arounds (but are not included into this PR),
- improve testing coverage of the package, another PR focusing on this will follow once merged,
- improve testing coverage of other packages, which will come later. 

- Addresses #3525 


internal/pkg/client/cache/dir.go
```diff
@@ -3,93 +3,308 @@
...
+		return basedir, nil
+	}
+
+	// If the environment variable is not set, we use the default cache.
+	usr, err := user.Current()
 	if err != nil {
-		sylog.Fatalf("Unable to get abs filepath: %v", err)
+		return "", fmt.Errorf("couldn't determine user home directory: %s", err)
+	}
+	basedir = path.Join(usr.HomeDir, basedirDefault)
```
mem: Everything between lines 223-230 should be replaced by `return syfs.ConfigDir(), nil`

If you do that, this function can no longer return an error (because syfs.ConfigDir() does not return an error)

internal/pkg/client/cache/dir.go
```diff
@@ -3,93 +3,308 @@
...
 	// singularity data directory where images are cached in by
 	// default.
 	// Uses "~/.singularity/cache" which will not clash with any 2.x cache
 	// directory.
-	cacheDir = "cache"
+	CacheDir = "cache"
 )
 
-var root string
+// SingularityCache is an opaque structure representing a cache
```
mem: "opaque" in Go means the struct has no exported fields.

internal/pkg/client/cache/dir.go
```diff
@@ -3,93 +3,308 @@
...
-// Clean : wipes all files in the cache directory, will return a error if one occurs
-func Clean() error {
-	sylog.Debugf("Removing: %v", Root())
+// NewHandle creates a new Singularity cache handle that can then be used to
+// interact with a cache. The location of the cache is driven by the DirEnv
+// directory. If it points to a location where a cache is not already present,
+// a new cache will implicitly be created; otherwise, the new cache handle
+// will point at the existing cache.
+func NewHandle() (*SingularityCache, error) {
+	// Singularity makes the following assumptions:
```
mem: this is a useful comment, but I feel it should be moved to getCacheBasedir() instead of being here.

internal/pkg/client/cache/dir.go
```diff
@@ -3,93 +3,308 @@
...
+	}
+
+	return hdlInit(basedir)
+}
 
-	if err := os.RemoveAll(Root()); err != nil {
-		return fmt.Errorf("unable to clean all cache: %s", err)
+// Init initializes a new cache within a given directory
+func hdlInit(baseDir string) (*SingularityCache, error) {
+	rootDir := getCacheRoot(baseDir)
```
mem, line 96: Now that the field is not exported and basedir does not exist, shouldn't the check for blank or homedir happen here instead of in the IsValid function?

internal/pkg/client/cache/dir.go
```diff
@@ -3,93 +3,308 @@
...
+	fmt.Println("Setting cache root to: ", rootDir)
+	if err := initCacheDir(rootDir); err != nil {
+		return nil, fmt.Errorf("failed initializing caching directory: %s", err)
 	}
 
+	newCache := new(SingularityCache)
+	os.Setenv(DirEnv, baseDir)
```
mem, line 117: I remember this was mentioned in the other review.

Can you please add a comment here as to why this is needed?

internal/pkg/client/cache/dir.go
```diff
@@ -3,93 +3,308 @@
...
+	}
+	basedir = path.Join(usr.HomeDir, basedirDefault)
+
+	return basedir, nil
+}
+
+// getCacheRoot figures out what the root directory is.
+func getCacheRoot(basedir string) string {
+	// Note: basedir and root are different and described earlier.
+	// The DirEnv environment variable can be used to set the basedir (or it
```
mem: The environment variable influences the return value of getCacheBasedir(), so most (all?) of this comment should be in that function, not here. Also, most of this comment seems to belong in the documentation for the function, not just in an internal comment.

internal/pkg/client/cache/dir.go
```diff
@@ -3,93 +3,308 @@
...
+// initCacheDir initializes a sub-cache within a cache, e.g., the shub
+// sub-cache.
 func initCacheDir(dir string) error {
-	if _, err := os.Stat(dir); os.IsNotExist(err) {
+	fInfo, err := os.Stat(dir)
+	switch {
+	case os.IsNotExist(err):
+		// The directory does not exist, we create it
 		sylog.Debugf("Creating cache directory: %s", dir)
 		if err := fs.MkdirAll(dir, 0755); err != nil {
```
mem: I'm sorry I didn't realize this before, but os.MkdirAll is already doing most of this: it begins by stat'ing dir to verify it exists, etc. Could you please check that we are not doing more work than necessary? I get the impression that simply calling fs.MkdirAll is enough.

internal/pkg/client/cache/dir.go
```diff
@@ -3,93 +3,308 @@
...
+		return nil
 	}
+}
 
-	return nil
+// checkImageHash checks whether the sha256 hash of an image in the cache
+// matches the sum passed in.
+func checkImageHash(path, sum string) bool {
+	cacheSum, err := client.ImageHash(path)
+	if err != nil {
```
mem: Notice how this is checking booleans to return booleans. This means the whole thing can be written as:

    return err == nil && sum == cacheSum

mem (changes requested):

internal/pkg/client/cache/dir.go
gvallee: Correct, I will fix this. Just something left from my initial modifications, sorry about that.

internal/pkg/client/cache/dir.go
gvallee: Good point, I was not aware of that helper function. It will simply the code. +1

internal/pkg/client/cache/dir.go
gvallee: These comments were explicitly requested in the context of another review of a previous PR. I think, after moving comments as you requested, that `getCacheBasedir()` is now adequately commented so I am wondering if we could leave these comments here (in my mind, it is a note, it does not hurt to have a reminder note to explain the context). Does it make sense?

internal/pkg/client/cache/dir.go
gvallee: fixed

gvallee:
FYI, CI fails because of the problem highlighted in issue #3592. I think a good way to proceed is to wait before merging this PR while I prepare a fix for #3592.

gvallee:

PR #3598, opened 2019-05-21
Move icmd wrapper to internal/pkg/test/exec

mem:
Rework some of the code to make it more similar to exec.Command, so that
moving from exec.Command to this version is simpler.

Rework some of the code already using this.


jscook2345 (approved):
LGTM! Great work @mem

PR #3603, opened 2019-05-23
Add 'SINGULARITY_DISABLE_CACHE'

### LINKED ISSUE #2519: SINGULARITY_DISABLE_CACHE is not a thing in 3.0
## Version of Singularity:

3.0.1

### Expected behavior

```
$ export SINGULARITY_DISABLE_CACHE=1
$ singularity build blah.sif docker://busybox
[...snip...]
$ ls -l ~/.singularity/cache/oci/blobs/sha256/
total 0
```
### Actual behavior

```
$ export SINGULARITY_DISABLE_CACHE=1
$ singularity build blah.sif docker://busybox
[...snip...]
$ ls -l ~/.singularity/cache/oci/blobs/sha256/
total 748
-rw-r--r-- 1 vagrant vagrant 755724 Jan  4 18:29 57c14dd66db0390dbf6da578421c077f6de8e88edd0815b4caa94607ba5f4c09
-rw-r--r-- 1 vagrant vagrant    347 Jan  4 18:36 8d7f536d20a5503e6bf967b88c6ea4f2a0a4fac8739f4e88f577531bf5300e66
-rw-r--r-- 1 vagrant vagrant    575 Jan  4 18:36 9bf7f625d0785fbcd9e2d07695e22485360d37a83164fc12832d06b7db27da53
```

This was brought to my attention in DM on Slack by @0xaf1f.

WestleyK:

This PR adds 'SINGULARITY_DISABLE_CACHE' to Singularity when building containers.

Example:

```bash
$ sudo singularity build --disable-cache test.sif test.def 
INFO:    Starting build...
INFO:    Downloading library image to tmp cache: /tmp/single_use_cache_424786444
INFO:    Creating SIF file...
```

- Fixes #2519 

<br>


internal/pkg/envconfig/envconfig.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"fmt"
+	"os"
+)
+
+func NoCache() bool {
+	env := os.Getenv("SINGULARITY_NO_CACHE")
+
+	fmt.Println("DEBIG: ENV: ", env)
```
jmstover: `DEBUG:` ?

WestleyK: Opps, yeah, its supposed to be `debug` :smile: This is very much a WIP, I'll remove it soon!

internal/pkg/envconfig/envconfig.go
```diff
@@ -0,0 +1,24 @@
...
+
+	if env == "1" || env == "true" {
```
jmstover: Should `SINGULARITY_NO_CACHE` work if it's set to _anything_ ... i.e. as long as `SINGULARITY_NO_CACHE` is set, then don't cache anything.

WestleyK: Okay, I'll fix that :+1: Thanks!

WestleyK: Although, wont it be confusing to new users if they set `SINGULARIY_DISABLE_CACHE=false`, they might expect it to be off, (disabled) ?...

jmstover: If the default is to cache ... why would you set `SINGULARITY_DISABLE_CACHE=false` ? Doesn't that seem ... counter intuitive to define something that you do by default?

WestleyK: Yeah, agree, I was just thinking out loud.

jmstover: Agree. It may be wanted to have a specific value like the `1` or `true` ... From my point of view though in some cases like this it's enough that the env variable is set ... not that it's set to a specific value. Hell, it could be set to `Bob` ... and IMO we should still not cache, because the variable is there.

internal/pkg/envconfig/envconfig.go
```diff
@@ -8,17 +8,24 @@
...
+//
+
+// NoCache will return true if 'SINGULARITY_DISABLE_CACHE' is set.
 func NoCache() bool {
 	env := os.Getenv("SINGULARITY_NO_CACHE")
-
-	fmt.Println("DEBIG: ENV: ", env)
-
-	if env == "1" || env == "true" {
+	if env != "" {
```
jmstover:
```
$ export SINGULARITY_NO_CACHE=""
$ env | grep 'NO_CACHE'
SINGULARITY_NO_CACHE=
```

Being set like:  `export SINGULARITY_NO_CACHE=""` ... makes the variable set ... but an empty string. `os.LookupEnv` should probably be used in this case to differentiate between an empty value, and unset variable.

internal/pkg/envconfig/envconfig.go
```diff
@@ -11,18 +11,23 @@ import (
...
+	CacheDir     = "SINGULARITY_CACHEDIR"
+)
 
 // NoCache will return true if 'SINGULARITY_DISABLE_CACHE' is set.
 func NoCache() bool {
-	env := os.Getenv("SINGULARITY_NO_CACHE")
-	return env != ""
+	env, ok := os.LookupEnv(DisableCache)
+
+	if !ok || env == "" {
```
jmstover:
```
if !ok {
    return false
}
```

Env being "" should be fine. As long as variable is set.

WestleyK: Wait, if `SINGULARITY_DISABLE_CACHE` is set to `""`, then it *should* use cache? Right now it works like:

```bash
$ sudo SINGULARITY_DISABLE_CACHE="" singularity build -F test.sif test.def 
INFO:    Starting build...
# uses the cache image
/root/.singularity/cache/library/sha256.63259fd0a2acb88bb652702c08c1460b071df51149ff85dc88db5034532a14a0/alpine_latest.sif
INFO:    Creating SIF file...

$ sudo SINGULARITY_DISABLE_CACHE="1" singularity build -F test.sif test.def 
INFO:    Starting build...
# downloads a new image
INFO:    Downloading library image to tmp cache: /tmp/single_use_cache_520818175
INFO:    Creating SIF file...
```

jmstover: This comes down to... Do you view the variable being set as enough to disable the cache? Even if it's a empty value? 

The `SINGULARITY_DISABLE_CACHE` variable is set, even though it's an empty string. You're setting the environment variable that disables the use of cache.

WestleyK: Imo, if `SINGULARITY_DISABLE_CACHE` is set to `""`, then thats the same as a unset variable. So if `SINGULARITY_DISABLE_CACHE` have a value, then singularity wont use the cache. That makes sense to me, imo.

internal/pkg/build/sources/conveyorPacker_library.go
```diff
@@ -63,22 +65,52 @@ func (cp *LibraryConveyorPacker) Get(b *types.Bundle) (err error) {
...
+	noCache := envconfig.NoCache()
+
 	imageName := uri.GetName("library://" + imageRef)
-	imagePath := cache.LibraryImage(libraryImage.Hash, imageName)
 
-	if exists, err := cache.LibraryImageExists(libraryImage.Hash, imageName); err != nil {
-		return fmt.Errorf("unable to check if %v exists: %v", imagePath, err)
-	} else if !exists {
-		sylog.Infof("Downloading library image")
+	if !noCache {
```
ikaneshiro: Lets switch the order of these cases so that we can avoid a double negative, makes it easier to read imo

internal/pkg/build/sources/conveyorPacker_library.go
```diff
@@ -63,22 +65,52 @@ func (cp *LibraryConveyorPacker) Get(b *types.Bundle) (err error) {
...
+		imagePath = cache.LibraryImage(libraryImage.Hash, imageName)
 
-		if err = library.DownloadImageNoProgress(context.TODO(), libraryClient, imagePath, imageRef); err != nil {
-			return fmt.Errorf("unable to Download Image: %v", err)
+		fmt.Println("IMAGE_PATH: ", imagePath)
```
ikaneshiro: Lets use sylog here, either `INFO` or `DEBUG` whatever's appropriate

internal/pkg/build/sources/conveyorPacker_library.go
```diff
@@ -63,22 +65,52 @@ func (cp *LibraryConveyorPacker) Get(b *types.Bundle) (err error) {
...
+			if cacheFileHash, err := client.ImageHash(imagePath); err != nil {
+				return fmt.Errorf("Error getting ImageHash: %v", err)
+			} else if cacheFileHash != libraryImage.Hash {
+				return fmt.Errorf("Cached File Hash(%s) and Expected Hash(%s) does not match", cacheFileHash, libraryImage.Hash)
+			}
 		}
+	} else {
+		tmpPath := envconfig.GetCacheDir()
+		if tmpPath == "" {
+			tmpPath = "/tmp"
```
ikaneshiro: Theres a `tmpdir` parameter in the bundle options, we should use that here to that this respects the `--tmpdir` flag

internal/pkg/build/sources/conveyorPacker_library.go
```diff
@@ -63,22 +65,52 @@ func (cp *LibraryConveyorPacker) Get(b *types.Bundle) (err error) {
...
+			return fmt.Errorf("unable to make tmp dir: %v", err)
+		}
+
+		imagePath = file.Name()
 
-		if cacheFileHash, err := client.ImageHash(imagePath); err != nil {
-			return fmt.Errorf("Error getting ImageHash: %v", err)
-		} else if cacheFileHash != libraryImage.Hash {
-			return fmt.Errorf("Cached File Hash(%s) and Expected Hash(%s) does not match", cacheFileHash, libraryImage.Hash)
+		fmt.Println("newFileName: ", imagePath)
```
ikaneshiro: Same sylog comment as above

ikaneshiro (changes requested):
This should also impact `conveyorPacker_oci.go` as well since that caches oci image layers

internal/pkg/build/sources/conveyorPacker_library.go
WestleyK: Yeah, this is temporary, just for debugging.

internal/pkg/build/sources/conveyorPacker_library.go
```diff
@@ -63,22 +66,44 @@ func (cp *LibraryConveyorPacker) Get(b *types.Bundle) (err error) {
...
-		return fmt.Errorf("unable to check if %v exists: %v", imagePath, err)
-	} else if !exists {
-		sylog.Infof("Downloading library image")
+	if cp.NoCache {
+		tmpPath := cp.TmpDir
+		if tmpPath == "" {
+			tmpPath = "/tmp"
+		}
+		// TODO: Remove file.Name() when done with it
+		file, err := ioutil.TempFile(tmpPath, "single_use_cache_")
```
ikaneshiro: Instead of the tmpdir logic above, we can simply feed `cp.TmpDir` into the first arg of `TempFile` since it defaults to the system tmp dir if that field is an empty string. Ref: https://golang.org/pkg/io/ioutil/#TempFile

internal/pkg/build/sources/conveyorPacker_oci.go
```diff
@@ -121,7 +123,7 @@ func (cp *OCIConveyorPacker) Get(b *sytypes.Bundle) (err error) {
 	}
 
 	// Grab the modified source ref from the cache
-	cp.srcRef, err = ociclient.ConvertReference(cp.srcRef, cp.sysCtx)
+	cp.srcRef, err = ociclient.ConvertReference(cp.TmpDir, cp.NoCache, cp.srcRef, cp.sysCtx)
```
ikaneshiro: I think the changes here and to `internal/pkg/client/oci/oci.go` are not necessary for this, if you want to not cache the image, you can skip this `ConvertReference()` function that changes `cp.srcRef` and it will download the blobs to `cp.b.Path` which is in the `tmpDir` if set or the system tmpdir if not set

cmd/internal/cli/build.go
```diff
@@ -158,6 +159,14 @@ var buildTmpdirFlag = cmdline.Flag{
 	EnvKeys:      []string{"TMPDIR"},
 }
 
+// SINGULARITY_DISABLE_CACHE
+var buildDisableCacheFlag = cmdline.Flag{
```
ikaneshiro, line 163: I think we should initialize this hidden flag like this:
```
var buildDisableCacheFlag = cmdline.Flag{
	ID:           "buildDisableCacheFlag",
	Value:        &disableCache,
	DefaultValue: false,
	Name:         "disablecache",
	Hidden:       true,
	EnvKeys:      []string{"DISABLE_CACHE"},
}
```

As is, it looks like this is creating a flag with the name "", which could cause issues if we register two flags in this manner since they will have a name collision

cmd/internal/cli/build_linux.go
```diff
@@ -66,16 +66,19 @@ func run(cmd *cobra.Command, args []string) {
 					sylog.Fatalf("Unable to create definition for sandbox build: %v", err)
 				}
 
+				fmt.Println("BUILDING: LINUX: sandbox")
```
ikaneshiro: Seems like a left over debug statement

ikaneshiro (changes requested):

cmd/internal/cli/build.go
WestleyK: Does it even need to be a hidden flag? *I think not...*

internal/pkg/build/sources/conveyorPacker_library.go
WestleyK: Should this use `ioutil.TempDir()` for this? Instead of `TempFile()`?

cmd/internal/cli/build.go
ikaneshiro: I think its fine to have this be exposed, but make sure to add usage info

internal/pkg/build/sources/conveyorPacker_library.go
ikaneshiro: Yes, also i would change the sylog message below to `Downloading library image to tmp location...`

ikaneshiro: Well, actually `TempFile()` works fine since you do need to pass a file path to `library.DownloadImageNoProgress`

WestleyK: :+1:

jchavezlavalle (approved):

cmd/internal/cli/actions.go
```diff
@@ -59,26 +60,31 @@ func handleOCI(cmd *cobra.Command, u string) (string, error) {
 		return "", fmt.Errorf("failed to get SHA of %v: %v", u, err)
```
ikaneshiro: The statement using `ImageSHA()` should be part of the else case of the if statement below since it is not used if the cache is disables

cmd/internal/cli/actions.go
```diff
@@ -59,26 +60,31 @@ func handleOCI(cmd *cobra.Command, u string) (string, error) {
...
+	imgabs := ""
 	name := uri.GetName(u)
-	imgabs := cache.OciTempImage(sum, name)
 
-	if exists, err := cache.OciTempExists(sum, name); err != nil {
-		return "", fmt.Errorf("unable to check if %v exists: %v", imgabs, err)
-	} else if !exists {
-		sylog.Infof("Converting OCI blobs to SIF format")
+	if DisableCacheActions {
+		file, err := ioutil.TempDir(tmpDir, "single_use_cache_")
```
ikaneshiro: We can change `file` to `imgabs` and avoid the assignment 5 lines below.

also why are you using `TempDir()` instead of `TempFile()`? If you meant to use `TempFile()` then the assignment below is not avoidable.

cmd/internal/cli/actions.go
```diff
@@ -111,22 +147,40 @@ func handleLibrary(u, libraryURL string) (string, error) {
 		return "", err
 	}
 
-	imageName := uri.GetName("library://" + imageRef)
-	imagePath := cache.LibraryImage(libraryImage.Hash, imageName)
+	imagePath := ""
+	if DisableCacheActions {
+		// TODO: Remove file.Name() when done with it
```
ikaneshiro: Is this still a TODO?

cmd/internal/cli/pull.go
```diff
@@ -217,7 +230,8 @@ func pullRun(cmd *cobra.Command, args []string) {
 			sylog.Fatalf("While pulling library image: %v", err)
 		}
 	case ShubProtocol:
-		err := shub.DownloadImage(name, args[i], force, noHTTPS)
+		// TODO:  <--
```
ikaneshiro: Same here

internal/app/singularity/pull.go
```diff
@@ -75,23 +76,42 @@ func LibraryPull(name, ref, transport, fullURI, libraryURI, keyServerURL, authTo
...
-	if err != nil {
-		return fmt.Errorf("unable to check if %s exists: %v", imagePath, err)
-	}
-	if !exists {
-		sylog.Infof("Downloading library image")
+	imagePath := ""
+
+	if noCache {
+		// Dont use cached image
+		// TODO: Remove file.Name() when done with it
```
ikaneshiro: Here too

cmd/internal/cli/actions.go
WestleyK: Theres probably a better name for `single_use_cache_`?...

internal/app/singularity/pull.go
```diff
@@ -261,20 +281,39 @@ func OciPull(name, imageURI, tmpDir string, ociAuth *ocitypes.DockerAuthConfig,
...
 
-	exists, err := cache.OciTempExists(sum, imgName)
-	if err != nil {
-		return fmt.Errorf("unable to check if %s exists: %s", imgName, err)
-	}
-	if !exists {
-		sylog.Infof("Converting OCI blobs to SIF format")
-		if err := convertDockerToSIF(imageURI, cachedImgPath, tmpDir, noHTTPS, ociAuth); err != nil {
+	if noCache {
+		// TODO: Remove file.Name() when done with it
```
ikaneshiro: Here too

internal/pkg/build/sources/conveyorPacker_library.go
```diff
@@ -69,28 +72,46 @@ func (cp *LibraryConveyorPacker) Get(b *types.Bundle) (err error) {
...
+	imagePath := ""
 	imageName := uri.GetName("library://" + imageRef)
-	imagePath := cache.LibraryImage(libraryImage.Hash, imageName)
 
-	if exists, err := cache.LibraryImageExists(libraryImage.Hash, imageName); err != nil {
-		return fmt.Errorf("unable to check if %v exists: %v", imagePath, err)
-	} else if !exists {
-		sylog.Infof("Downloading library image")
+	if cp.NoCache {
+		// TODO: Remove file.Name() when done with it
```
ikaneshiro: Here too

pkg/client/shub/pull.go
```diff
@@ -23,7 +23,7 @@ const pullTimeout = 7200
 
 // DownloadImage will retrieve an image from the Container Singularityhub,
 // saving it into the specified file
-func DownloadImage(filePath string, shubRef string, force, noHTTPS bool) (err error) {
+func DownloadImage(filePath string, shubRef, tmpDir string, force, noHTTPS, noCache bool) (err error) {
```
ikaneshiro: Why add these parameters if they are unused?

ikaneshiro (changes requested):

cmd/internal/cli/actions.go
ikaneshiro: i would use `-` instead of `_` since i think most of the tmp paths we create use those. Maybe call it `tmp-cache-`?

WestleyK: > also why are you using `TempDir()` instead of `TempFile()`? If you meant to use `TempFile()` then the assignment below is not avoidable.

Because this function pulls a docker image, and that uses a tmp directory to store the layers, afaik (could be wrong)

cmd/internal/cli/actions.go
WestleyK: Yes...

cmd/internal/cli/pull.go
WestleyK: Ohh, this is a reminder about the shub caching. I will remove it.

internal/app/singularity/pull.go
WestleyK: Yup, still a TODO, I will look into it a bit more...

pkg/client/shub/pull.go
WestleyK: Because I was starting on the noCache for shub, but since shub dose not cache right now, I left it, will remove now.

internal/app/singularity/pull.go
WestleyK: Not a TODO anymore :smile:

internal/pkg/build/sources/conveyorPacker_library.go
WestleyK: Fixed!

cmd/internal/cli/actions.go
WestleyK: Fixed now! :taco:

GodloveD:
@ikaneshiro, @jmstover, what's the status of this?  Has @WestleyK addressed your comments fully?

internal/pkg/build/build.go
```diff
@@ -299,16 +299,21 @@ func runBuildEngine(b *types.Bundle) error {
 	return starterCmd.Run()
 }
 
-func getcp(def types.Definition, libraryURL, authToken string) (ConveyorPacker, error) {
+func getcp(def types.Definition, libraryURL, authToken string, noCache bool, tmpDir string) (ConveyorPacker, error) {
 	switch def.Header["bootstrap"] {
 	case "library":
-		return &sources.LibraryConveyorPacker{}, nil
+		return &sources.LibraryConveyorPacker{
```
ikaneshiro: If the bundle contains a `NoCache` option you shouldnt need to set that here. It should be accessible from the conveyorpacker through the bundle.

internal/pkg/build/build.go
```diff
@@ -299,16 +299,21 @@ func runBuildEngine(b *types.Bundle) error {
...
+			NoCache: noCache,
+		}, nil
 	case "oras":
 		return &sources.OrasConveyorPacker{}, nil
 	case "shub":
 		return &sources.ShubConveyorPacker{}, nil
 	case "docker", "docker-archive", "docker-daemon", "oci", "oci-archive":
-		return &sources.OCIConveyorPacker{}, nil
+		return &sources.OCIConveyorPacker{
```
ikaneshiro: Same here

ikaneshiro (changes requested):

ikaneshiro (approved):

PR #3623, opened 2019-05-24
Create clean & unique caches for unit tests

### LINKED ISSUE #3592: Improve cache management in the context of `DropPrivilege()` and `EnsurePrivilege()`

gvallee:

This PR adds a new testing function that enables the creation of a unique image cache for unit tests. This avoids the problem of unit tests sharing cache from one test to another.

The typical usage is:
```
test.EnsurePrivilege(t)
test.SetCacheDir(t)
defer test.CleanCacheDir(t)
```

Overview:
- This PR introduces a new environment variable: `SINGULARITY_TESTBASEDIR` that is set either to `/tmp/WithPrivilege` or `/tmp/WithoutPrivilege`; this allows us to separate privilege management and cache management but let the code managing the cache know where the cache should be created.
- `SetCacheDIr(t)` creates a temporary directory and set `SINGULARITY_CACHEDIR` to point at the temporary directory.
- `CleanCacheDir()` will delete the temporary directory used as image cache.

Note: this PR is based on the idea of limiting changes to the existing code so the `/tmp/WithPrivilege` and `/tmp/WithoutPrivilege` are kept and unique temporary image caches are created within these directories. It would be possible to not have these two directories and just create temporary directories since these two directories are currently only used for image caches.

- Fixes #3592 
- Addresses #3525 

### DIFF
--- a/internal/pkg/test/cache.go
+++ b/internal/pkg/test/cache.go
@@ -8,36 +8,44 @@ package test
 import (
 	"io/ioutil"
 	"os"
+	"syscall"
 	"testing"
 )
 
-// SetCacheDir creates a new image cache in the context of testing
-// and set the appropriate environment variable to ensure that the
-// temporary cache is used by the caller. Note that the context of
-// the directory that is passed in will be deleted when calling
-// CleanCacheDir().
+// SetCacheDir creates a new temporary image cache directory in the
+// directory basedir and returns the path of the new directory. If
+// basedir is an empty string, SetCacheDir uses the default directory
+// for temporary files (see os.TempDir). It is the caller's responsibility
+// to remove the directory when no longer needed, which can be accomplished
+// by calling CleanCacheDir
 func SetCacheDir(t *testing.T, basedir string) string {
-	if basedir == "" {
-		// We create a unique temporary directory for the image cache since Go run
-		// tests concurrently and since SINGULARITY_TESTBASEDIR is a static directory.
-		dir, err := ioutil.TempDir("", "image_cache-")
-		if err != nil {
-			t.Fatalf("failed to create temporary directory: %s", err)
-		}
-
-		return dir
+	// We create a unique temporary directory for the image cache since Go run
+	// tests concurrently.
+	dir, err := ioutil.TempDir(basedir, "image_cache-")
+	if err != nil {
+		t.Fatalf("failed to create temporary directory: %s", err)
 	}
 
-	return basedir
+	// Update the access mode to the directory since ioutil.TempDir()
+	// creates a new directory with a different access mode than the
+	// one we want here.
+	oldmask := syscall.Umask(0)
+	defer syscall.Umask(oldmask)
+	err = os.Chmod(dir, 0755)
+	if err != nil {
+		t.Fatalf("failed to change permission of %s: %s", dir, err)
+	}
+
+	return dir
 }
 
-// CleanCacheDir deleted the temporary cache that was created for
-// testing purposes.
-func CleanCacheDir(t *testing.T, basedir string) {
-	// We clean the image cache only if we are not using the
-	// default cache; we do not want to clean the developer's
-	// default cache.
-	if basedir != "" {
-		os.RemoveAll(basedir)
+// CleanCacheDir deleted the temporary image cache that was created for
+// testing purposes. CleanCacheDir will fail the test if an error occurs
+// during its execution.
+func CleanCacheDir(t *testing.T, path string) {
+	// Note: if path is empty os.RemoveAll() will fail
+	err := os.RemoveAll(path)
+	if err != nil {
+		t.Fatalf("failed to remove directory %s: %s", path, err)
 	}
 }
--- a/internal/pkg/test/cache.go
+++ b/internal/pkg/test/cache.go
@@ -6,59 +6,38 @@
 package test
 
 import (
-	"fmt"
 	"io/ioutil"
 	"os"
-	"syscall"
 	"testing"
 )
 
 // SetCacheDir creates a new image cache in the context of testing
 // and set the appropriate environment variable to ensure that the
-// temporary cache is used by the caller.
-func SetCacheDir(t *testing.T) {
-	basedir := os.Getenv("SINGULARITY_TESTBASEDIR")
-
-	if basedir != "" {
-		// We make sure the base directory for the test is created.
-		// we rely only on package from Golang to avoid dependency cycles
-		oldmask := syscall.Umask(0)
-		defer syscall.Umask(oldmask)
-		err := os.MkdirAll(basedir, 0755)
-		// This is an error only outside of the case where the directory already exists
-		if err != nil && !os.IsExist(err) {
-			t.Fatalf("failed to create test base directory %s: %s", basedir, err)
+// temporary cache is used by the caller. Note that the context of
+// the directory that is passed in will be deleted when calling
+// CleanCacheDir().
+func SetCacheDir(t *testing.T, basedir string) string {
+	if basedir == "" {
+		// We create a unique temporary directory for the image cache since Go run
+		// tests concurrently and since SINGULARITY_TESTBASEDIR is a static directory.
+		dir, err := ioutil.TempDir("", "image_cache-")
+		if err != nil {
+			t.Fatalf("failed to create temporary directory: %s", err)
 		}
-	} else {
-		t.Fatal("SINGULARITY_TESTBASEDIR not set")
+
+		return dir
 	}
 
-	// We create a unique temporary directory for the image cache since Go run
-	// tests concurrently and since SINGULARITY_TESTBASEDIR is a static directory.
-	dir, err := ioutil.TempDir(basedir, "image_cache-")
-	if err != nil {
-		t.Fatalf("failed to create temporary directory: %s", err)
-	}
-
-	// We do not use cache.DirEnv to avoid a dependency cycle
-	err = os.Setenv("SINGULARITY_CACHEDIR", dir)
-	if err != nil {
-		t.Fatalf("failed to set SINGULARITY_CACHEDIR")
-	}
+	return basedir
 }
 
 // CleanCacheDir deleted the temporary cache that was created for
 // testing purposes.
-func CleanCacheDir(t *testing.T) {
-	// We do not use cache.DirEnv to avoid a dependency cycle
-	cacheDir := os.Getenv("SINGULARITY_CACHEDIR")
-
+func CleanCacheDir(t *testing.T, basedir string) {
 	// We clean the image cache only if we are not using the
 	// default cache; we do not want to clean the developer's
 	// default cache.
-	if cacheDir != "" {
-		os.RemoveAll(cacheDir)
-	} else {
-		fmt.Println("WARNING! the test is using the default image cache")
+	if basedir != "" {
+		os.RemoveAll(basedir)
 	}
 }
--- a/internal/pkg/test/cache.go
+++ b/internal/pkg/test/cache.go
@@ -1,0 +1,64 @@
+// Copyright (c) 2019, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package test
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"syscall"
+	"testing"
+)
+
+// SetCacheDir creates a new image cache in the context of testing
+// and set the appropriate environment variable to ensure that the
+// temporary cache is used by the caller.
+func SetCacheDir(t *testing.T) {
+	basedir := os.Getenv("SINGULARITY_TESTBASEDIR")
+
+	if basedir != "" {
+		// We make sure the base directory for the test is created.
+		// we rely only on package from Golang to avoid dependency cycles
+		oldmask := syscall.Umask(0)
+		defer syscall.Umask(oldmask)
+		err := os.MkdirAll(basedir, 0755)
+		// This is an error only outside of the case where the directory already exists
+		if err != nil && !os.IsExist(err) {
+			t.Fatalf("failed to create test base directory %s: %s", basedir, err)
+		}
+	} else {
+		t.Fatal("SINGULARITY_TESTBASEDIR not set")
+	}
+
+	// We create a unique temporary directory for the image cache since Go run
+	// tests concurrently and since SINGULARITY_TESTBASEDIR is a static directory.
+	dir, err := ioutil.TempDir(basedir, "image_cache-")
+	if err != nil {
+		t.Fatalf("failed to create temporary directory: %s", err)
+	}
+
+	// We do not use cache.DirEnv to avoid a dependency cycle
+	err = os.Setenv("SINGULARITY_CACHEDIR", dir)
+	if err != nil {
+		t.Fatalf("failed to set SINGULARITY_CACHEDIR")
+	}
+}
+
+// CleanCacheDir deleted the temporary cache that was created for
+// testing purposes.
+func CleanCacheDir(t *testing.T) {
+	// We do not use cache.DirEnv to avoid a dependency cycle
+	cacheDir := os.Getenv("SINGULARITY_CACHEDIR")
+
+	// We clean the image cache only if we are not using the
+	// default cache; we do not want to clean the developer's
+	// default cache.
+	if cacheDir != "" {
+		os.RemoveAll(cacheDir)
+	} else {
+		fmt.Println("WARNING! the test is using the default image cache")
+	}
+}
--- a/internal/pkg/test/privilege_linux.go
+++ b/internal/pkg/test/privilege_linux.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, Sylabs Inc. All rights reserved.
+// Copyright (c) 2018-2019, Sylabs Inc. All rights reserved.
 // This software is licensed under a 3-clause BSD license. Please consult the
 // LICENSE.md file distributed with the sources of this project regarding your
 // rights to use or distribute this software.
@@ -21,10 +21,10 @@ var origUID, origGID, unprivUID, unprivGID int
 var origHome, unprivHome string
 
 const (
-	// CacheDirPriv is the directory the cachedir gets set to when running privileged
-	CacheDirPriv = "/tmp/WithPrivilege"
-	// CacheDirUnpriv is the directory the cachedir gets set to when running unprivileged
-	CacheDirUnpriv = "/tmp/WithoutPrivilege"
+	// BaseDirPriv is the directory used by tests when running privileged
+	BaseDirPriv = "/tmp/WithPrivilege"
+	// BaseDirUnpriv is the directory used by tests when running unprivileged
+	BaseDirUnpriv = "/tmp/WithoutPrivilege"
 )
 
 // EnsurePrivilege ensures elevated privileges are available during a test.
@@ -33,6 +33,8 @@ func EnsurePrivilege(t *testing.T) {
 	if uid != 0 {
 		t.Fatal("test must be run with privilege")
 	}
+
+	os.Setenv("SINGULARITY_TESTBASEDIR", BaseDirPriv)
 }
 
 // DropPrivilege drops privilege. Use this at the start of a test that does
@@ -59,8 +61,8 @@ func DropPrivilege(t *testing.T) {
 		}
 	}
 
-	// set SINGULARITY_CACHEDIR
-	os.Setenv("SINGULARITY_CACHEDIR", CacheDirUnpriv)
+	// set SINGULARITY_CACHEDIR; we do not refer to cache.DirEnv to avoid dependency cycles.
+	os.Setenv("SINGULARITY_TESTBASEDIR", BaseDirUnpriv)
 }
 
 // ResetPrivilege returns effective privilege to the original user.
@@ -77,8 +79,7 @@ func ResetPrivilege(t *testing.T) {
 
 	runtime.UnlockOSThread()
 
-	// set SINGULARITY_CACHEDIR
-	os.Setenv("SINGULARITY_CACHEDIR", CacheDirPriv)
+	os.Setenv("SINGULARITY_TESTBASEDIR", BaseDirPriv)
 }
 
 // WithPrivilege wraps the supplied test function with calls to ensure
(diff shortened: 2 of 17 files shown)


internal/pkg/test/privilege_linux.go
```diff
@@ -77,8 +79,7 @@ func ResetPrivilege(t *testing.T) {
 
 	runtime.UnlockOSThread()
 
-	// set SINGULARITY_CACHEDIR
-	os.Setenv("SINGULARITY_CACHEDIR", CacheDirPriv)
+	os.Setenv("SINGULARITY_TESTBASEDIR", BaseDirPriv)
```
tri-adam: This is making the assumption that `origUID`/`origGID` is a privileged user, but this isn't _technically_ correct. Certain package tests can be run unprivileged, and it is useful to be able to do that when, for example, running tests directly from VS Code rather than the command line.

internal/pkg/test/privilege_linux.go
```diff
@@ -21,10 +21,10 @@ var origUID, origGID, unprivUID, unprivGID int
...
 
 const (
-	// CacheDirPriv is the directory the cachedir gets set to when running privileged
-	CacheDirPriv = "/tmp/WithPrivilege"
-	// CacheDirUnpriv is the directory the cachedir gets set to when running unprivileged
-	CacheDirUnpriv = "/tmp/WithoutPrivilege"
+	// BaseDirPriv is the directory used by tests when running privileged
+	BaseDirPriv = "/tmp/WithPrivilege"
+	// BaseDirUnpriv is the directory used by tests when running unprivileged
+	BaseDirUnpriv = "/tmp/WithoutPrivilege"
```
tri-adam: We're implicitly making the assumption here that `/tmp` is accessible with and without privilege. Since this is the case, let's create the temporary cache dirs directly in `/tmp`, which removes the need for the state that is being shared in `SINGULARITY_TESTBASEDIR`, and removes the coupling between privilege and cachedir handling.

tri-adam (changes requested):
I think the privilege handling functions are violating the single responsibility principle, and by this I mean that `SetCacheDir()` and `CleanCacheDir()` rely on `WithPrivilege()`/`WithoutPrivilege()` (or similar) to be called in order to function properly. Consider the case where a test requires a cache directory, but can execute properly with or without privilege. In this case, all the caller really needs is a directory that is owned by the _effective UID/GID_.

The use of environment variables to share state also has the downside of being process global, when the effective UID/GID is actually thread level. This means that while `WithPrivilege()`/`WithoutPrivilege()` and friends were originally safe for use in parallel tests, now that they have environmental side-effects, this is no longer true.

Here's what I propose as a way forward. Make it the caller's problem to provide a `baseDir` that is accessible with and without permissions, and then make the inputs/outputs to these functions explicit:

```
func GetCacheDir(t *testing.T, baseDir string) string
func CleanCacheDir(t *testing.T, cacheDir)
```

`baseDir` can be optional, and if an empty string is provided, [os.TempDir](https://golang.org/pkg/os/#TempDir) will be used . `GetCacheDir()` can then simply call [osutil.TempDir](https://golang.org/pkg/io/ioutil/#TempDir) directly. We're not changing any assumption here, since currently we're assuming we can access `/tmp` with/without privilege. This will remove the extra layer (`/tmp/WithPrivilege/image_cache-XXXXX` -> `/tmp/image_cache-XXXXX`).

I'd also suggest that setting of the environment be left up to the calling test. I realize that at the moment, the `cache` package relies on this magic environment variable, and this will necessitate a helper function in that package's test suite. As we are seeing though, this environment variable usage makes the package hard to test (and impossible, concurrently). IMO, environment variable handling should be done in the same place as other user input (`cmd/internal/cli`). `initCacheDir()` could then be made public and explicit, and functions within the package could return meaningful errors if the package had not been initialized. I'm not implying this needs to be done as part of this PR, but if you agree with the above, maybe it's worth considering as a future enhancement?

gvallee:
> I think the privilege handling functions are violating the single responsibility principle, and by this I mean that `SetCacheDir()` and `CleanCacheDir()` rely on `WithPrivilege()`/`WithoutPrivilege()` (or similar) to be called in order to function properly. Consider the case where a test requires a cache directory, but can execute properly with or without privilege. In this case, all the caller really needs is a directory that is owned by the _effective UID/GID_.
> 

And it is the way to always has been since `/tmp/WithPrivilege` and `/tmp/WithoutPrivilege` are currently mainly used for the image cache. I am not opposed in changing that but this PR is meant to keep the same semantics and address the problems we have when running concurrent tests: here each process/thread would create a temporary directory that is used during the execution of a test. Yes we currently rely on a environment variable, not because I think it is the best option, but because this is the closest option from the initial code. I have been proposing code (including the current PR #3595 and what we are talking about here was basically put on hold with strong questioning about the necessity of having public APIs exposed by the package implementing cache management only for the purpose of cache management.

> The use of environment variables to share state also has the downside of being process global, when the effective UID/GID is actually thread level. This means that while `WithPrivilege()`/`WithoutPrivilege()` and friends were originally safe for use in parallel tests, now that they have environmental side-effects, this is no longer true.
> 

Agree but again, the initial goal here was to stay close to initial implementation which only relied on environment variable. My intent was to do further modifications in the context of issue #3525. Because we were not able to reach consensus, I must address problems in small pieces; my initial PR that was addressing all these problems at once could not be approved.

So I would suggest here the minimum set of modifications so we can move forward. We want to have a base directory, which I think is a great idea, let's open an issue and do it in a different PR. I am facing a chicken and egg problem: I need to split the problem in small chunks to get consensus but small PRs do not get consensus because they do not address the entire problem.

> Here's what I propose as a way forward. Make it the caller's problem to provide a `baseDir` that is accessible with and without permissions, and then make the inputs/outputs to these functions explicit:
> 
> ```
> func GetCacheDir(t *testing.T, baseDir string) string
> func CleanCacheDir(t *testing.T, cacheDir)
> ```
> 
> `baseDir` can be optional, and if an empty string is provided, [os.TempDir](https://golang.org/pkg/os/#TempDir) will be used . `GetCacheDir()` can then simply call [osutil.TempDir](https://golang.org/pkg/io/ioutil/#TempDir) directly. We're not changing any assumption here, since currently we're assuming we can access `/tmp` with/without privilege. This will remove the extra layer (`/tmp/WithPrivilege/image_cache-XXXXX` -> `/tmp/image_cache-XXXXX`).
> 

Again, agreed and basically what I already proposed in various PRs but consensus was not reached within the group. This PR was intentionally very small and only addressing the problems I was experiencing while running tests. I would suggest to merge this PR, open an issue on the specific point you are mentioning and do it very soon. Does it make sense?

> I'd also suggest that setting of the environment be left up to the calling test. I realize that at the moment, the `cache` package relies on this magic environment variable, and this will necessitate a helper function in that package's test suite. As we are seeing though, this environment variable usage makes the package hard to test (and impossible, concurrently). IMO, environment variable handling should be done in the same place as other user input (`cmd/internal/cli`). `initCacheDir()` could then be made public and explicit, and functions within the package could return meaningful errors if the package had not been initialized. I'm not implying this needs to be done as part of this PR, but if you agree with the above, maybe it's worth considering as a future enhancement?

I personally believe that most unit tests should not even have to rely on an environment variable, there are not end-to-end tests. Using environment variable always create headaches when using threads (go routines) and concurrent executions. Please also note that with the current code, it is possible to change the environment variable during execution, effectively changing the location of the cache in the middle of a test, creating a very inconsistent context. But these problems are not addressed in this PR (cf. related issues).

internal/pkg/test/privilege_linux.go
gvallee: Agreed even if outside of the initial scope I had for this PR. Will do.

internal/pkg/test/privilege_linux.go
gvallee: Okay but I am sure that I am missing something because this is the assumption by the original code, not my modification, right? `ResetPrivilege()` is meant to be called after `DropPrivilege()` initially in the context of a privilege process. Furthermore, if not executed in the context of a privileged context in the first place, the behavior following the modification will still be the same.

I am a little confused here: the intent of this PR is not to revisit how privileged/unprivileged tests are executed but simply to avoid concurrent and unwanted sharing of caches which are creating problems.

gvallee:
I pushed the modifications that I think address your comments. I have concerns with this approach: 
- The unit tests are still supposed to deal with some of the cache setting (i.e., call the two new functions, set the environment variable), while the new API only really create a temporary directory. If the unit tests are already dealing with cache setting, why do not they create the temporary directory instead of relying on the new API from the `test` package?
- This still assumes that we set part of the cache, but not the entire cache, which will actually be created implicitly at some point (and as the person who is writing the test, we do not know when). If the environment variable changes for whatever reason, it may or not, change the location of the cache, and ultimately, may or not, corrupt the image cache. I would prefer to have the concept of cache handle like suggested in #3525 so the cache is set once, the cache configuration is captured by the handle and used during the entire duration of the test. I am aware that this point in beyond this but this PR goes, I think, against that idea (the code in the `test` package cannot return a cache handle because of dependency cycle).

That said, if you think this is the way to go, I am okay with it.

gvallee:
@tri-adam Something I do not understand just by reading the code: is the cache supposed to be shared between the privileged and unprivileged parts of the tests? Or is each supposed to have their own cache?
Based on my testing, some unit tests are failing because the cache directory was created in privileged mode and the tests are using the cache in unprivileged mode leading to failures because they do not have permission to access the directory. So what is the intent?

internal/pkg/test/privilege_linux.go
tri-adam: > ResetPrivilege() is meant to be called after DropPrivilege()

Yes, this is true.

> ...initially in the context of a privilege process.

This assumption is not necessarily true. `ResetPrivilege()` sets effective privilege to the _original_ user, which could be privileged or unprivileged. Many packages can be tested as an unprivileged user, and this is common when working within a package using an editor such as VS Code.

> I am a little confused here: the intent of this PR is not to revisit how privileged/unprivileged tests are executed but simply to avoid concurrent and unwanted sharing of caches which are creating problems.

Yes, but `SetCacheDir` and `CleanCacheDir` rely on side-effects of functions such as `ResetPrivilege`, and thus they must be considered. Consider the following sequence of events:

- `/tmp/WithPrivilege` directory exists and is writable by `root` only.
- `go test github.com/sylabs/singularity/blah/blah` is run as unprivileged user, on command line or via editor.
- one of the tests in the package calls `DropPrivilege()` and then `ResetPrivilege()`. These are effectively no-ops, since the test was run as an unprivileged user. `ResetPrivilege()` `SINGULARITY_TESTBASEDIR` to `/tmp/WithPrivilege`.
- a test calls `SetCacheDir()`, which attempts to call `ioutil.TempDir()` on `/tmp/WithPrivilege`, which fails with an access denied error.

tri-adam:
> * The unit tests are still supposed to deal with some of the cache setting (i.e., call the two new functions, set the environment variable), while the new API only really create a temporary directory. If the unit tests are already dealing with cache setting, why do not they create the temporary directory instead of relying on the new API from the `test` package?

Really this is just a pair of convenience functions to create a cache and delete it. Nothing preventing copying this code to the unit tests themselves, although that'd result in some duplication of code?

> * This still assumes that we set part of the cache, but not the entire cache, which will actually be created implicitly at some point (and as the person who is writing the test, we do not know when). If the environment variable changes for whatever reason, it may or not, change the location of the cache, and ultimately, may or not, corrupt the image cache. I would prefer to have the concept of cache handle like suggested in #3525 so the cache is set once, the cache configuration is captured by the handle and used during the entire duration of the test. I am aware that this point in beyond this but this PR goes, I think, against that idea (the code in the `test` package cannot return a cache handle because of dependency cycle).

Are you suggesting the API of the cache package itself should change? Explicitly passing around a handle that represents the cache rather than using global state (`var`, environment variable, etc.) would solve a lot of things here. The user could supply via an environment variable (with a sensible default), and the unit tests could explicitly allocate temporary directories that can be used safely in concurrently run tests. Best of both world in my opinion... maybe a follow-up PR?

tri-adam:
> @tri-adam Something I do not understand just by reading the code: is the cache supposed to be shared between the privileged and unprivileged parts of the tests? Or is each supposed to have their own cache?
> Based on my testing, some unit tests are failing because the cache directory was created in privileged mode and the tests are using the cache in unprivileged mode leading to failures because they do not have permission to access the directory. So what is the intent?

I guess it depends on the test, could you point to an example that is failing? If a single test is running code in both unprivileged and privileged modes, and attempting to share a cache directory, that would seem to me to be a bug in the test (or the code itself)?

internal/pkg/test/cache.go
```diff
@@ -0,0 +1,43 @@
...
+		dir, err := ioutil.TempDir("", "image_cache-")
+		if err != nil {
+			t.Fatalf("failed to create temporary directory: %s", err)
+		}
+
+		return dir
+	}
+
+	return basedir
+}
```
tri-adam, line 40: I think `SetCacheDir` should _always_ create a temporary directory, because it's important the test always runs with the same precondition (an empty cache.) The intention behind `basedir` is to allow the user to override the location where the temporary directory will be created. I think all we need to make this happen is:

```
	dir, err := ioutil.TempDir(basedir, "image_cache-")
	if err != nil {
		t.Fatalf("failed to create cache directory: %s", err)
	}
	return dir
```

internal/pkg/test/cache.go
```diff
@@ -0,0 +1,43 @@
...
+
+// CleanCacheDir deleted the temporary cache that was created for
+// testing purposes.
+func CleanCacheDir(t *testing.T, basedir string) {
+	// We clean the image cache only if we are not using the
+	// default cache; we do not want to clean the developer's
+	// default cache.
+	if basedir != "" {
+		os.RemoveAll(basedir)
+	}
```
tri-adam, line 50: Mirroring the suggestion in `basedir`... since a temporary directory is always created, this function should always delete `basedir`. I'd also suggest renaming `basedir`, to avoid any implication that this is the same value that was passed in `basedir` to `SetCacheDir`. Maybe just `dir` or `path`?

internal/pkg/test/cache.go
```diff
@@ -0,0 +1,43 @@
...
+	"io/ioutil"
+	"os"
+	"testing"
+)
+
+// SetCacheDir creates a new image cache in the context of testing
+// and set the appropriate environment variable to ensure that the
+// temporary cache is used by the caller. Note that the context of
+// the directory that is passed in will be deleted when calling
+// CleanCacheDir().
```
tri-adam: Think this comment needs updating, something like:

```
// SetCacheDir creates a new temporary cache directory in the directory basedir
// and returns the path of the new directory. If basedir is the empty string,
// SetCacheDir uses the default directory for temporary files (see os.TempDir).
// It is the caller's responsibility to remove the directory when no longer
// needed, which can be accomplished by calling CleanCacheDir.
```

internal/pkg/test/cache.go
```diff
@@ -0,0 +1,43 @@
...
+		}
+
+		return dir
+	}
+
+	return basedir
+}
+
+// CleanCacheDir deleted the temporary cache that was created for
+// testing purposes.
```
tri-adam: Maybe update this to make it clear everything in the dir will be deleted?

```
// CleanCacheDir removes the temporary cache directory at path, and any children
// it contains.
```

tri-adam (changes requested):
Few small comments, LMK what you think!

gvallee:
> 
> Are you suggesting the API of the cache package itself should change? Explicitly passing around a handle that represents the cache rather than using global state (`var`, environment variable, etc.) would solve a lot of things here. The user could supply via an environment variable (with a sensible default), and the unit tests could explicitly allocate temporary directories that can be used safely in concurrently run tests. Best of both world in my opinion... maybe a follow-up PR?

Yes, the initial motivation being explained there: https://github.com/sylabs/singularity/issues/3525; and this work that has been going on for a few weeks now: https://github.com/sylabs/singularity/pull/3595. Currently the goal is to break issues done to smaller pieces, this issue being one.

gvallee:
> > @tri-adam Something I do not understand just by reading the code: is the cache supposed to be shared between the privileged and unprivileged parts of the tests? Or is each supposed to have their own cache?
> > Based on my testing, some unit tests are failing because the cache directory was created in privileged mode and the tests are using the cache in unprivileged mode leading to failures because they do not have permission to access the directory. So what is the intent?
> 
> I guess it depends on the test, could you point to an example that is failing? If a single test is running code in both unprivileged and privileged modes, and attempting to share a cache directory, that would seem to me to be a bug in the test (or the code itself)?

Yes, and it is actually some of the modifications in this PR since this is sometimes creating problems for me when running `make testall` on Ubuntu (and sometimes creating CI failures with cryptic error messages).
A good example is there: https://github.com/sylabs/singularity/pull/3623/files#diff-e3d82a5dda0c2a08df54250f8e6fea61
I actually put some comments in there to explain the problem. I am also tracking at least one other problem that appears to be related to this so it might be a problem that kind of spread across the unit tests (to be audited).

internal/pkg/test/cache.go
gvallee: I just finally understand what you are actually saying. It makes sense to me. I will update the code accordingly.

internal/pkg/test/cache.go
gvallee: I will rename to `path`, no problem. Always deleting the content is a problem when `path` is an empty string, since `os.RemoveAll()` would fail but we do not really handle errors here (the function does not return errors). Can you confirm that you think `path` should never be empty there? If so, my question here is how do we want the function to handle errors? Should we fail the test if `path` is empty? Just call `os.RemoveAll()` and fail the test when it returns an error? I do not have a preference on the topic but would like to make sure we clearly define the error case(s) in the context of that function.

internal/pkg/test/cache.go
gvallee: Agreed. The only points that I think needs clarification is the error case: do we always fail the test if an error occurs the execution of the function?

internal/pkg/test/privilege_linux.go
gvallee: BTW, I misunderstood at first what you had in mind, I think we are now on the same page, I believe this is resolved.

internal/pkg/test/privilege_linux.go
gvallee: Sorry it seems that I am not precise enough here, my intent was not to talk about a general case. And I start to believe that we ultimately are saying the same thing and we just identified a missing piece.

I have the following code path in mind, which happens all the time when manually running `make testall`:
- A unit test in privileged creates a new temporary image cache for testing, usually `EnsurePrivilege()` is called; `SINGULARITY_CACHEDIR` is set to point at this new image cache.
- The unit test switches to unprivileged mode by calling `DropPrivilege()`.
- After switching to unprivileged mode, a new image cache is usually *not* set with the current implementation of the unit test **and** the environment variable still points at the one that was created In privileged mode because `SINGULARITY_CACHEDIR` is set to point to it.
- The unit test is calling functions that ultimately rely on the image cache (it is usually not explicit, more like a side effect) and if the cache is not set to be shared between privileged and unprivileged mode, the test fails.

I believe the key point here is `SINGULARITY_TESTBASEDIR`. That concept was not there before and I actually removed it because I thought you said it was not what you had in mind, while I know believe it was just not implemented correctly to reflect what you had in mind.

So here is how I would phrase the semantics we are looking for here and let's see if we are in agreement:
`SINGULARITY_DEFAULT_TESTBASEDIR` (a.k.a `SINGULARITY_TESTBASEDIR` but i renamed it to reflect better what it means) sets the default location that we use as base directory to set temporary image caches for unit tests. Unit test developers can override that location by specifying the base directory when calling `SetCacheDir()`. When switching between privileged and unprivileged mode within a unit test (by calling functions from `internal/pkg/test`), `SINGULARITY_DEFAULT_TESTBASEDIR` will automatically be set to the appropriate default location, i.e., `/tmp/WithPrivilege` or `/tmp/WithoutPrivilege`).
It is the developer's responsibility to correctly initialize a new temporary image cache or update the access mode of an existing image cache, as appropriate, when switching between privileged and unprivileged mode, by properly calling `SetCacheDir()` and properly setting `SINGULARITY_CACHEDIR` for the temporary image cache to be used by the unit test. It is also the developer's responsibility to correctly delete all temporary image caches that have been set in the context of unit tests.

gvallee: I still wonder if we need `SINGULARITY_DEFAULT_TESTBASEDIR` since `SetCacheDir()` will either rely on the path that is passed in or simply create a new temporary directory.

I guess my question would be: do we want to have all the temporary image caches in `/tmp/WithPrivilege` and `/tmp/WithoutPrivilege` by default? or just start to put them wherever `ioutil.TempDir("", ...)` would put them?

internal/pkg/test/cache.go
gvallee: I updated the code

internal/pkg/test/cache.go
gvallee: I renamed it

ikaneshiro: I think it's fine to let `os.RemoveAll()` fail on an empty string. I would suggest changing the formatting of the path in the fatal output to be quoted though so that it is easier to understand the error if it occurs

gvallee:
Notes from discussion with @ikaneshiro 
- There is no obvious need to keep using `/tmp/WithPrivilege` and `/tmp/WithoutPrivilege`, instead we could use temporary directories.
- It would be better to include the modifications that enables the creation of a cache handle with a base directory. This would let us have a clean API to create/destroy caches, as well as limit the reliance on environment variables in the context of unit tests (environment variables are meant to capture options when using the command line, not to drive the behavior of unit tests).
- We need to understand better what the current unit tests are doing and what they are potentially not doing correctly (for instance a misuse of cache images). This includes the following questions:
  - Do all unit tests need an image cache? If not, do we need a new API to deactivate the cache? Right now, the cache is always assumed to be active in the unit tests.
  - Sharing an image cache between a privileged and unprivileged context seems reasonable but do we need a specific API to enable this? Right now, the access is set in a way that does not really allow sharing of a cache.
  - Do all the unit tests correctly initialize the image cache when necessary? We already identified unit tests were it was not the case but it is still unclear whether to capture all such unit tests since the reliance on the image cache may be hidden deep in the function call graph.

gvallee:

PR #3631, opened 2019-05-27
[E2E] Tests for version: verifying semver

jchavezlavalle:

E2E Tests for version: verification of semantic version.


e2e/version/version.go
```diff
@@ -57,13 +60,17 @@ func testEqualVersion(t *testing.T) {
 			if err != nil {
 				t.Fatalf("Failed to obtain version: %+v", err)
```
WestleyK, line 61: Should you print the output here:

```go
t.Log(string(out))
```

e2e/version/version.go
```diff
@@ -38,8 +39,10 @@ func testSemanticVersion(t *testing.T) {
...
 			}
-
-			if semanticVersion, err := semver.Make(string(out)); err != nil {
+			outputVersion := strings.Replace(string(out), "singularity version ", "", -1)
```
ikaneshiro: I think `strings.TrimPrefix()` is a better util function to use here

e2e/version/version.go
```diff
@@ -38,8 +39,10 @@ func testSemanticVersion(t *testing.T) {
...
+			outputIndex := strings.Index(string(out), "-")
```
ikaneshiro: Why are you splitting the version here? It seems like `semver` will properly handle the build metadata?

ikaneshiro (changes requested):

e2e/version/version.go
jchavezlavalle: No it doesn't, that's why I'm just considering the version with this library until the patch to verify the version. `singularity --version` outputs : `singularity version 3.2.0-541.g4e601d8d9` so from the `-` to the end, this library does not detect the tag (it just handles major version, minor version and patch) Also, consider that `singularity version` (command) gives as output: `3.2.0-541.g4e601d8d9` so if I need to consider if both versions are the same, I need to split the `singularity version` string  if present, which is the output from `singularity --version`

ikaneshiro (approved):
LGTM

PR #3639, opened 2019-05-28
Improve Remote Command Workflow

### LINKED ISSUE #3597: Remote command flow
## Version of Singularity:

3.2+

### Expected behavior

- [ ] `singularity push` when not logged in should direct users to `singularity remote login` explicitly.  Currently it says `log in with singularity remote commands`

- [x] `singularity remote login` should use the default endpoint and optionally specify a different endpoint 

- [ ] `singularity remote add` should prompt for token after adding.  

- [x] All of the `singularity remote` help output incorrect lists subcommands, but they should be options.  
Current:   `singularity [global options...] remote <subcommand> login <remote_name>`
Expected: `  singularity [global options...] remote <options> login <remote_name>`

Related: #3540 

EDIT: Turned the list into check boxes to track what's been finished

ikaneshiro:

This pr addresses 1, 3 and 4 of #3597.
Point 2 has been addressed by #3554.

- Adds automatic `login` flow to the `add` command
- Adds `--no-login` flag to `add` command
- Updates docs to correctly list commands, sub-commands, options and new workflow
- Better info messages

- Fixes #3597 


GodloveD (changes requested):
I would argue that `list` and `status` should also have the `[* options]` strings added since they *all* technically can at least accept the `--help` option.  

Other than that I think this is good to go.

ikaneshiro:
@GodloveD I'd argue that the `--help` flag is a special case and can be ignored in this context because it applies to every command and this info is only accessible through the `--help` flag or the `help` command. I think saying `[* options]` when the `--help` flag is the only option doesn't really give a user any useful information imho

rherban:
Did a quick test to add an endpoint, but it didn't prompt for auth token:
```
[rherban@test]:~/tmp/singularity/builddir$ ./singularity version
5fac9d2bb
[rherban@test]:~/tmp/singularity/builddir$ ./singularity remote list
NAME           URI              GLOBAL
[SylabsCloud]  cloud.sylabs.io  YES
laptop         10.1.1.21        NO
sylabs         cloud.sylabs.io  NO
[rherban@test]:~/tmp/singularity/builddir$ ./singularity remote add ornl cloud-sylabs.marble.ccs.ornl.gov
[rherban@test]:~/tmp/singularity/builddir$
```

ikaneshiro:
@rherban I think you're on the wrong commit. `5fac9d2bb` is too old
```
commit 5fac9d2bb811c9859f072385ca18243ab4ae2e69 (HEAD, origin/master, origin/HEAD)
Merge: af9d50148 abc07f855
Author: Ian Kaneshiro <iankane@umich.edu>
Date:   Thu May 23 10:08:51 2019 -0400

    Merge pull request #3602 from mem/fix_push_to_oras
    
    Allow arbitrary locations when pushing to oras

```

rherban:
Adding a global endpoint doesn't prompt for auth token.  Maybe that makes sense because an admin would be doing it on behalf of the cluster, or maybe it got overlooked.

```
[rherban@test]:~/tmp/singularity/builddir$ ./singularity remote list
NAME  URI  GLOBAL
[rherban@test]:~/tmp/singularity/builddir$ sudo ./singularity remote add -g sylabs cloud.sylabs.io
INFO:    Remote "sylabs" added.
[rherban@test]:~/tmp/singularity/builddir$ ./singularity remote list
NAME    URI              GLOBAL
sylabs  cloud.sylabs.io  YES
```

ikaneshiro:
@rherban That should be done outside the context of `-g` by the root user logging in like any other user imo

rherban:
I don't think it's a problem and I'm not advocating blocking the merge because of it, just wanted to mention in case it was overlooked.

cmd/internal/cli/remote.go
```diff
@@ -63,6 +64,15 @@ var remoteTokenFileFlag = cmdline.Flag{
 	Usage:        "path to the file holding token",
 }
 
+// --no-login
+var remoteNoLoginFlag = cmdline.Flag{
+	ID:           "remoteNoLoginFlag",
+	Value:        &remoteNoLogin,
+	DefaultValue: false,
+	Name:         "no-login",
```
mem, line 72: this creates the following:

```
singularity remote --no-login=false
```

quickly: what does that do?

I think it would be better to have:

```
singularity remote --login=false
singularity remote --login=true
```

and default to true.

If there's a desire to keep the value-less flag, I would suggest looking for another form.

cmd/internal/cli/remote.go
```diff
@@ -110,47 +124,69 @@ var RemoteAddCmd = &cobra.Command{
...
+		name := args[0]
+		uri := args[1]
+		if err := singularity.RemoteAdd(remoteConfig, name, uri, global); err != nil {
 			sylog.Fatalf("%s", err)
 		}
+		sylog.Infof("Remote %q added.", name)
+
+		// ensure that this was not called with global flag, otherwise this will store the token in the
+		// world readable config
+		if !global && !remoteNoLogin {
```
mem: Even if the flag is not changed, I would change the name of the variable: `!remoteNoLogin` means "perform login". This would be easier to read as:

```
if remoteDoLogin && !global {
```

and when written in this form I have to wonder what the case `remoteDoLogin=true, global=true` does (since it's not handled here).

cmd/internal/cli/remote.go
ikaneshiro: That would set it as if the flag wasn't used. I would anticipate this used without a value specification as just `--no-login` which is clearer than `--login=false` imo.

cmd/internal/cli/remote.go
ikaneshiro: I think changing the the variable name will need a change in the flag name or an extra variable just to imply the opposite behavior. Any suggestions on this? I have also improved the logic to handle the case you presented explicitly instead of silently like before.

GodloveD (approved):
lgtm

PR #3650, opened 2019-05-30
Only clean cache directories that we create

WestleyK:

This PR fixes the issue when you set `SINGULARITY_CACHEDIR` to (for example your home directory `$HOME`), then run `singularity cache clean --all`, not it will not remove your home directory. Instead this will only remove the directories that we create, eg. `library`, `shub`, `oci`, `oci-tmp`.

This is most likely a temporary fix, until https://github.com/sylabs/singularity/pull/3623 gets merged.

- Fixes #

<br>

### DIFF
--- a/internal/pkg/client/cache/dir.go
+++ b/internal/pkg/client/cache/dir.go
@@ -45,10 +45,10 @@ func Root() string {
 // Clean wipes all files in the cache directory, will return a error if one occurs
 func Clean() error {
 	// TODO: add oras here
-	cacheDirs := append([]string{""}, Library(), OciTemp(), OciBlob(), Shub())
+	cacheDirs := append([]string{}, Library(), OciTemp(), OciBlob(), Shub())
 
 	for _, c := range cacheDirs {
-		sylog.Debugf("Removeing: %s", c)
+		sylog.Debugf("Removeing: %s ...", c)
 		err := os.RemoveAll(c)
 		if err != nil {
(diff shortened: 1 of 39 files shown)


gvallee:
If I remember correctly, this will cause a failure in `cmd/internal/cli/actions_linux.go` when running `make testall` on Ubuntu (to be confirmed). The following link may help: https://github.com/sylabs/singularity/pull/3595/files#diff-896d1f4723d0a072f845c405328b7f1e

ikaneshiro (approved):
lgtm will let @gvallee have a look too before merging

internal/pkg/client/cache/dir.go
```diff
@@ -42,12 +42,17 @@ func Root() string {
...
 func Clean() error {
-	sylog.Debugf("Removing: %v", Root())
-
-	if err := os.RemoveAll(Root()); err != nil {
-		return fmt.Errorf("unable to clean all cache: %s", err)
+	// TODO: add oras here
+	cacheDirs := append([]string{}, Library(), OciTemp(), OciBlob(), Shub())
+
+	for _, c := range cacheDirs {
+		sylog.Debugf("Removeing: %s ...", c)
```
gvallee:
```suggestion
		sylog.Debugf("Removing: %s ...", c)
```

gvallee (changes requested):
Exactly what I was suggesting, thanks for doing it. I only found a minor typo.

gvallee (approved):

PR #3659, opened 2019-05-31
Add 'shub' to cache list/clean + improved '--name' handling

WestleyK:

This PR adds support for `cache list/clean` for `shub`. And improves the `--name` handling.

#### Example:

You cant do this before:

```bash
$ singularity cache clean -Nalpine_latest.sif -Nalpine_3.8.sif -Ndebian_latest.sif
```

_Before it would only clean `debian_latest.sif`, but now it will clean all that are specified._

#### Shub cache example:

```bash
$ singularity exec shub://GodloveD/busybox true
INFO:    Downloading shub image
[...]

$ singularity cache list 
NAME                     DATE CREATED           SIZE             TYPE
busybox_latest.sif       2019-05-31 12:19:49    618.53 kB        shub

There 1 containers using: 618.53 kB, 0 oci blob file(s) using 0.00 kB of space.
Total space used: 618.53 kB
```

- Related to #3660 

<br>

### DIFF
--- a/cmd/internal/cli/cache_clean_linux.go
+++ b/cmd/internal/cli/cache_clean_linux.go
@@ -48,7 +48,7 @@ var cacheCleanTypesFlag = cmdline.Flag{
 var cacheCleanNameFlag = cmdline.Flag{
 	ID:           "cacheCleanNameFlag",
 	Value:        &cacheName,
-	DefaultValue: []string{""},
+	DefaultValue: []string{},
 	Name:         "name",
 	ShortHand:    "N",
 	Usage:        "specify a container cache to clean (will clear all cache with the same name)",
--- a/cmd/internal/cli/cache_clean_linux.go
+++ b/cmd/internal/cli/cache_clean_linux.go
@@ -19,7 +19,7 @@ import (
 var (
 	cleanAll        bool
 	cacheCleanTypes []string
-	cacheName       string
+	cacheName       []string
 )
 
 // -a|--all
@@ -40,7 +40,7 @@ var cacheCleanTypesFlag = cmdline.Flag{
 	DefaultValue: []string{"blob"},
 	Name:         "type",
 	ShortHand:    "T",
-	Usage:        "clean cache type, choose between: library, oci, and blob",
+	Usage:        "clean cache type, choose between: library, oci, shub, and blob",
 	EnvKeys:      []string{"TYPE"},
 }
 
@@ -48,7 +48,7 @@ var cacheCleanTypesFlag = cmdline.Flag{
 var cacheCleanNameFlag = cmdline.Flag{
 	ID:           "cacheCleanNameFlag",
 	Value:        &cacheName,
-	DefaultValue: "",
+	DefaultValue: []string{""},
 	Name:         "name",
 	ShortHand:    "N",
 	Usage:        "specify a container cache to clean (will clear all cache with the same name)",
@@ -61,7 +61,7 @@ func init() {
 	cmdManager.RegisterFlagForCmd(&cacheCleanNameFlag, CacheCleanCmd)
 }
 
-// CacheCleanCmd : is `singularity cache clean' and will clear your local singularity cache
+// CacheCleanCmd is 'singularity cache clean' and will clear your local singularity cache
 var CacheCleanCmd = &cobra.Command{
 	DisableFlagsInUseLine: true,
 	Run: func(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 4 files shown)


internal/app/singularity/cache_clean_linux.go
```diff
@@ -120,11 +131,36 @@ func cleanOciCacheName(cacheName string) (bool, error) {
 	return foundMatch, nil
 }
 
-// CleanCacheName : will clean a container with the same name as cacheName (in the cache directory).
+func cleanShubCacheName(cacheName string) (bool, error) {
+	foundMatch := false
+	blobs, err := ioutil.ReadDir(cache.Shub())
+	if err != nil {
+		return false, fmt.Errorf("unable to opening shub cache folder: %v", err)
```
jchavezlavalle: unable to open

internal/app/singularity/cache_clean_linux.go
```diff
@@ -46,28 +57,28 @@ func cleanBlobCache() error {
...
+		case "blob", "blobs":
+			return cleanBlobCache()
+		case "all":
+			return cache.Clean()
+		default:
+			// The caller checks the returned error and will exit as required
+			return fmt.Errorf("not a valid type: %s", cacheType)
+		}
 	}
+	return fmt.Errorf("no cache to clean; non specified")
```
jchavezlavalle: non -> none ?

internal/app/singularity/cache_clean_linux.go
```diff
@@ -46,28 +57,28 @@ func cleanBlobCache() error {
 	}
 
 	return nil
-
 }
```
jchavezlavalle: Ok, `cleanLibraryCache`, `cleanBlobCache`, `cleanShubCache` and `cleanOciCache` all do the same, they only differ on the param passed on `os.RemoveAll`, which is the cache type. Wouldn't it be better to instead create a single function where it receives the type and it removes the specified cache? I mean something like:

```
 func cleanSpecifiedCache(cacheType string) error{
            switch cacheType {
	        case "library":
                    cacheToRemove:= cache.Library()
                case "blob":
                    cacheToRemove:= cache.OciBlob()
                case "shub":
                    cacheToRemove:= cache.Shub()
                // add any additional cases needed

              sylog.Debugf("Removing: %v", cacheToRemove)

	err := os.RemoveAll(cacheToRemove)
	if err != nil {
		return fmt.Errorf("unable to clean %s cache: %v", cacheType , err)
	}

	return nil
}
```
That would remove functions:  `cleanLibraryCache`, `cleanBlobCache`, `cleanShubCache` and `cleanOciCache`

jchavezlavalle (dismissed):

cmd/internal/cli/cache_clean_linux.go
```diff
@@ -40,15 +40,15 @@ var cacheCleanTypesFlag = cmdline.Flag{
...
+	Usage:        "clean cache type, choose between: library, oci, shub, and blob",
 	EnvKeys:      []string{"TYPE"},
 }
 
 // -N|--name
 var cacheCleanNameFlag = cmdline.Flag{
 	ID:           "cacheCleanNameFlag",
 	Value:        &cacheName,
-	DefaultValue: "",
+	DefaultValue: []string{""},
```
ikaneshiro: This seems like an odd initial value, I think it should either be `[]string{}` or `nil`

internal/app/singularity/cache_clean_linux.go
```diff
@@ -15,59 +15,30 @@ import (
...
+		case "shub":
+			cacheToRemove = cache.Shub()
+		case "blob", "blobs":
+			cacheToRemove = cache.OciBlob()
+		case "all":
+			cacheToRemove = cache.Root()
+		default:
+			return fmt.Errorf("not a valid type: %s", c)
+		}
+		err := os.RemoveAll(cacheToRemove)
```
ikaneshiro, line 40: We already talked about this offline but I'm adding it here so that it is recorded somewhere. This introduces the same problem as #3650 is trying to address. The fix is to use `cache.Clean()` in the `all` case and return

ikaneshiro (changes requested):

cmd/internal/cli/cache_clean_linux.go
WestleyK: I tried to use `[]string{}`, but that panic... I'll try `nil` then.

WestleyK: Fixed now

ikaneshiro (approved):
lgtm

PR #3660, opened 2019-05-31
Support for Shub caching when using 'pull'

### LINKED ISSUE #3402: 3.2.0 claims shub caching but only works for action commands
## Version of Singularity:

3.2.0rc2

This only works with actions (ex: `exec`, `run`). Was this supposed to work with other commands such as `pull` or `build`? The way the changelog describes it makes me think all commands will cache the image.

This functionality was added in #2776.

WestleyK:

This PR adds the support for caching Shub images when using the `pull` command.

#### Example:

```bash
$ singularity pull shub://michael-tn/mpi-hello-world:latest
INFO:    Downloading shub image
 194.44 MiB / 194.44 MiB [========================================] 100.00% 4.12 TiB/s 0s

$ rm mpi-hello-world_latest.sif 
$ singularity pull shub://michael-tn/mpi-hello-world:latest
INFO:    Use image from cache
```

- Fixes #3402 

- Related to #3659 

<br>

### DIFF
--- a/internal/app/singularity/pull.go
+++ b/internal/app/singularity/pull.go
@@ -152,9 +152,9 @@ func LibraryPull(name, ref, transport, fullURI, libraryURI, keyServerURL, authTo
 	return retErr
 }
 
-// PullShubImage will download a image from shub, and cache it. Next time
+// PullShub will download a image from shub, and cache it. Next time
 // that container is downloaded this will just use that cached image.
-func PullShubImage(filePath string, shubRef string, force, noHTTPS bool) (err error) {
+func PullShub(filePath string, shubRef string, force, noHTTPS bool) (err error) {
 	if !force {
 		if _, err := os.Stat(filePath); err == nil {
 			return fmt.Errorf("image file already exists: %q - will not overwrite", filePath)
--- a/internal/app/singularity/pull.go
+++ b/internal/app/singularity/pull.go
@@ -29,6 +29,7 @@ import (
 	"github.com/sylabs/singularity/internal/pkg/sylog"
 	"github.com/sylabs/singularity/internal/pkg/util/uri"
 	"github.com/sylabs/singularity/pkg/build/types"
+	shub "github.com/sylabs/singularity/pkg/client/shub"
 	"github.com/sylabs/singularity/pkg/signing"
 	"github.com/sylabs/singularity/pkg/sypgp"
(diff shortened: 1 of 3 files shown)


pkg/client/shub/pull.go
```diff
@@ -13,17 +13,68 @@ import (
...
+	}
+	defer dstFile.Close()
+
+	srcFile, err := os.Open(imagePath)
+	if err != nil {
+		return fmt.Errorf("while opening cached image: %v", err)
+	}
+	defer srcFile.Close()
+
+	// Copy SIF from cache
```
ikaneshiro: I would change this from `SIF` to `image` because shub hosts older image formats as well

ikaneshiro (changes requested):
I think we should leave `DownloadImage()` as is (unless there is a bug fix I am missing) and put all the code you have added in `PullShubImage()` in a function within `internal/app/singularity/pull.go`. I think its a better design pattern to have this client package focus on handling the downloading of images and let our internal cli functions handle working with our image cache.

internal/app/singularity/pull.go
```diff
@@ -151,6 +152,55 @@ func LibraryPull(name, ref, transport, fullURI, libraryURI, keyServerURL, authTo
 	return retErr
 }
 
+// PullShubImage will download a image from shub, and cache it. Next time
+// that container is downloaded this will just use that cached image.
+func PullShubImage(filePath string, shubRef string, force, noHTTPS bool) (err error) {
```
ikaneshiro: Lets change this to `PullShub()` to match the other pull functions

pkg/client/shub/pull.go
```diff
@@ -41,12 +41,6 @@ func DownloadImage(filePath string, shubRef string, force, noHTTPS bool) (err er
 		sylog.Infof("Download filename not provided. Downloading to: %s\n", filePath)
 	}
 
-	if !force {
```
ikaneshiro, line 44: We should either keep this check, or remove the `force` parameter from the function call so that it doesnt silently do nothing

ikaneshiro (changes requested):

pkg/client/shub/pull.go
WestleyK: Oh yeah, I'll put the check back...

ikaneshiro (approved):
lgtm

PR #3668, opened 2019-06-01
Do not manipulate "environ" directly, use C API

mem:
Reading thru the docs, it seems the environ variable is not meant to be
used to _remove_ or _change_ values in the environment, but only to
iterate over them.

A recent change introduced the "cleanenv" function, which deletes all
environment variables except SINGULARITY_MESSAGELEVEL. It does this by
setting the pointers in the environ array to NULL. While this seems to
work with some implementations of the C library, with others it's
causing a segmentation fault during the initialization of the C library.

To avoid assuming a particular behavir, change the cleanenv function so
that it obtains the current value of SINGULARITY_MESSAGELEVEL, clears
all the environment variables using the clearenv function and sets
SINGULARITY_MESSAGELEVEL back to its previous value using setenv.

Since the original change did not specify the conditions under which a
bug was observed, this has received only limited testing. All the
existing tests are passing with this change.

### DIFF
--- a/cmd/starter/c/starter.c
+++ b/cmd/starter/c/starter.c
@@ -832,31 +832,17 @@ void do_exit(int sig) {
 }
 
 /*
- * cleanenv set environ pointer to NULL, while it works
- * in C context, it doesn't have any effect with the Go
- * runtime using the real pointer, so we need to work
- * directly with environ array.
+ * cleanenv clears all the environment variables except for MSGLVL_ENV
  */
 static void cleanenv(void) {
-    extern char **environ;
-    char **e;
-    char *p = NULL;
-
-    if ( environ == NULL || *environ == NULL ) {
-        fatalf("no environment variables set\n");
+    char * cur = getenv(MSGLVL_ENV);
+    if (cur) {
+        cur = strdup(cur);
     }
-
-    /* keep only SINGULARITY_MESSAGELEVEL for GO runtime */
-    for (e = environ; *e != NULL; e++) {
-        if ( strncmp(MSGLVL_ENV "=", *e, sizeof(MSGLVL_ENV)) == 0 ) {
-            p = *e;
-        } else {
-            *e = NULL;
-        }
-    }
-
-    if ( p != NULL ) {
-        *environ = p;
+    clearenv();
+    if (cur) {
+        setenv(MSGLVL_ENV, cur, 1);
+        free(cur);
     }
 }
 


cclerget:
@mem Regarding this fix, `clearenv` won't clear environment variables in Go context because `clearenv` just set the pointer `environ = NULL` to fake an empty environment, and after `setenv` call `environ` will point to another memory location (allocated under the hood by C library) which won't affect Go runtime using the stack pointer value to access to the environment array, that's why we need to manipulate environment stack array directly to also affect Go from C code.

But looks like I forgot to set `SINGULARITY_MESSAGELEVEL` pointer to `NULL`, what if you try with this code : 

```
static void cleanenv(void) {
    extern char **environ;
    char **e;
    char *p = NULL;

    if ( environ == NULL || *environ == NULL ) {
        fatalf("no environment variables set\n");
    }

    /* keep only SINGULARITY_MESSAGELEVEL for GO runtime */
    for (e = environ; *e != NULL; e++) {
        if ( strncmp(MSGLVL_ENV "=", *e, sizeof(MSGLVL_ENV)) == 0 ) {
            /* record string pointer before remove */
            p = *e;
        }
        *e = NULL;
    }

    /* set SINGULARITY_MESSAGELEVEL as the first and only environment variable */
    *environ = p;
}
```

cclerget (changes requested):
See previous comment

mem:
> @mem Regarding this fix, `clearenv` won't clear environment variables in Go context because `clearenv` just set the pointer `environ = NULL` to fake an empty environment, and after `setenv` call `environ` will point to another memory location (allocated under the hood by C library) which won't affect Go runtime using the stack pointer value to access to the environment array, that's why we need to manipulate environment stack array directly to also affect Go from C code.
> 
> But looks like I forgot to set `SINGULARITY_MESSAGELEVEL` pointer to `NULL`, what if you try with this code :
> 
> ```
> static void cleanenv(void) {
>     extern char **environ;
>     char **e;
>     char *p = NULL;
> 
>     if ( environ == NULL || *environ == NULL ) {
>         fatalf("no environment variables set\n");
>     }
> 
>     /* keep only SINGULARITY_MESSAGELEVEL for GO runtime */
>     for (e = environ; *e != NULL; e++) {
>         if ( strncmp(MSGLVL_ENV "=", *e, sizeof(MSGLVL_ENV)) == 0 ) {
>             /* record string pointer before remove */
>             p = *e;
>         }
>         *e = NULL;
>     }
> 
>     /* set SINGULARITY_MESSAGELEVEL as the first and only environment variable */
>     *environ = p;
> }
> ```

Thanks, I'll try that and post the result.

Meanwhile, can you please share a method to show that the bug you are fixing with this change is indeed fixed? I'm flying blind here because I don't know what the original problem was.

cclerget:
@mem The quicker (but not accurate) is to add a check in `init()` of `cmd/starter/main_linux.go` like this :

```
if len(os.Environ()) > 1 {
    sylog.Fatalf("environment variables was not cleaned correctly")
}
```

And test with :
```
echo "" | PIPE_EXEC_FD=0 SINGULARITY_MESSAGELEVEL=5 FOO=BAR BAR=FOO /usr/local/libexec/singularity/bin/starter-suid
```

This way you can check if environment was cleaned correctly from C.

The original problem was to avoid the need to clean environment variables both in C and Go, that's I mentioned in the first comment, it's also a security concern because we need to clean environment variable in C without the need to also clean environment variables from Go in order to avoid that GODEBUG or another variable alter/modify setuid behaviour (and because it could be potentially too late to clean envs in Go `init()` function)

mem:
> The original problem was to avoid the need to clean environment variables both in C and Go, that's I mentioned in the first comment, it's also a security concern because we need to clean environment variable in C without the need to also clean environment variables from Go in order to avoid that GODEBUG or another variable alter/modify setuid behaviour (and because it could be potentially too late to clean envs in Go `init()` function)

Ah, I'm slowly starting to understand now.

The "environment" is a chunk of memory laid out by the kernel when starting a process. Go assumes that this chunk of memory [is located right after the last argument](https://github.com/golang/go/blob/0a7bc8f430e3e4017910780a898a5f20d337895b/src/runtime/runtime1.go#L76). The same assumption is made in the C library. It reads the key=val strings from there and _copies_ that to (its own representation of the environment)[https://github.com/golang/go/blob/master/src/syscall/env_unix.go#L36]. That `copyenvs` function is executed once per process, the first time one of the environment-related functions is called.

The C library has a variable `environ` which initially points to this chunk of memory. If `setenv` is called, a _new_ array of environment variables is called, and the original is not touched. Further calls to `setenv` will free the `environ` array and allocate a new one.

This means that manipulating the environment from C (especially *adding* variables) won't have an effect in Go (because Go is not reading the `environ` variable from the C library, or using getenv from the C library) and manipulating the environment in Go won't have an effect in the C library.

PR #3675, opened 2019-06-04
Fixes key search output bug when using another key server

### LINKED ISSUE #3369: Key Search Output Issue
## Version of Singularity:

`3.2.0-rc1.42.gc92e3bd01`

### Expected behavior

`singularity key search` should supply the "machine readable" option to obtain search results from the key server, parse the results, and then output consistently formatted search results.

### Actual behavior

`singularity key search` does not supply the "machine readable" option, and thus the format is determined by the key server. The Singularity key server returns plaintext by default, but in some other key servers, this is raw HTML, which makes for an inconsistent user experience.

Run `singularity key search` against an arbitrary key server. For example, using the [SKS Keyserver Pool](https://sks-keyservers.net/overview-of-pools.php):

```
$ singularity key search -u http://p80.pool.sks-keyservers.net Adrock
WARNING: No default remote in use, falling back to: http://p80.pool.sks-keyservers.net
<?xml version="1.0" encoding="utf-8"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd" >
<html xmlns="http://www.w3.org/1999/xhtml">
<head>
<title>Search results for 'adrock'</title>
<meta http-equiv="Content-Type" content="text/html;charset=utf-8" />
<style type="text/css">
/*<![CDATA[*/
 .uid { color: green; text-decoration: underline; }
 .warn { color: red; font-weight: bold; }
/*]]>*/
</style></head><body><h1>Search results for 'adrock'</h1><pre>Type bits/keyID     Date       User ID
</pre><hr /><pre>
pub  2048R/<a href="/pks/lookup?op=get&amp;search=0x11119EF262D8E447">62D8E447</a> 2013-07-11 <a href="/pks/lookup?op=vindex&amp;fingerprint=on&amp;search=0x11119EF262D8E447">king adrock &lt;thekingadrockisminame@gmail.com&gt;</a>
                               onetyme &lt;onetyme@me.com&gt;
                               skin88p &lt;skin88pride@gmail.com&gt;
                               onetyme88 &lt;onetyme888@gmail.com&gt;
                               zhawk1188 &lt;zhawk1188@hotmail.com&gt;
                               whatchawant2 &lt;whatchawant2@hotmail.com&gt;
                               whatchawant88 &lt;whatchawant88@yahoo.com&gt;
                               sippenongin &lt;sippenongin@privatdemail.net&gt;
	 Fingerprint=BD11 1C62 F290 C191 AAAE  480E 1111 9EF2 62D8 E447 
</pre><hr /><pre>
pub  1024D/<a href="/pks/lookup?op=get&amp;search=0xECF0204C07E5513A">07E5513A</a> 2006-02-24 <a href="/pks/lookup?op=vindex&amp;fingerprint=on&amp;search=0xECF0204C07E5513A">Adam Parrott (none) &lt;adrock@devjunky.com&gt;</a>
	 Fingerprint=0FA8 B65C 0AD7 FA73 9703  C190 ECF0 204C 07E5 513A 
</pre><hr /><pre>
pub  1024D/<a href="/pks/lookup?op=get&amp;search=0x38DA6D5A9BC42A3A">9BC42A3A</a> 2001-09-24 *** KEY REVOKED *** [not verified]
                               <a href="/pks/lookup?op=vindex&amp;fingerprint=on&amp;search=0x38DA6D5A9BC42A3A">Adam Young (AY) &lt;adrock@mountaincable.net&gt;</a>
	 Fingerprint=EAB6 3898 379C DE17 3E1F  5EE3 38DA 6D5A 9BC4 2A3A 
</pre><hr /><pre>
pub   768D/<a href="/pks/lookup?op=get&amp;search=0x5E848BAB1F096AF7">1F096AF7</a> 1999-08-30 <a href="/pks/lookup?op=vindex&amp;fingerprint=on&amp;search=0x5E848BAB1F096AF7">Krzysiek &lt;adrock@polbox.com&gt;</a>
                               Krzysztof Za??ski &lt;krzysiek_z@poczta.onet.pl&gt;
	 Fingerprint=CAB2 4614 FC72 1773 FDB0  5834 5E84 8BAB 1F09 6AF7 
</pre></body></html>
```

jchavezlavalle:

Whenever searching a key on another keyserver like for instance doing:

```
$ singularity key search -u http://p80.pool.sks-keyservers.net Adrock
```
 the output was embedded inside HTML tags. this PR fixes that.

- Fixes #3369 


pkg/sypgp/sypgp.go
```diff
@@ -593,8 +595,13 @@ func SearchPubkey(search, keyserverURI, authToken string) error {
 		Size: 256,
 	}
 
+	//set the machine readable output on
+	var options []string
```
mem: this is

```
options := []string{"mr"}
```

pkg/sypgp/sypgp.go
```diff
@@ -593,8 +595,13 @@ func SearchPubkey(search, keyserverURI, authToken string) error {
...
+	options = append(options, "mr")
 	// Retrieve first page of search results from Key Service.
-	keyText, err := c.PKSLookup(context.TODO(), &pd, search, client.OperationIndex, true, false, nil)
+	keyText, err := c.PKSLookup(context.TODO(), &pd, search, client.OperationIndex, true, false, options)
```
mem, line 761: this is fine, with the understanding that `context.TODO()` is called that because it doubles as documentation: there's a pending TODO item here. I'm going to assume that the pending item is adding context to SearchPubKey(...). If yes, it would be worthwhile adding a comment to that effect.

pkg/sypgp/sypgp.go
```diff
@@ -612,6 +619,140 @@ func SearchPubkey(search, keyserverURI, authToken string) error {
 	return nil
 }
 
+//function to obtain the algorithm name for key encryption
```
mem: convention is to use the format:

```
// getEncryptionAlgorithmName obtains the ...
```

in other workds:

```
// functionName does something
```

note the blank after `//`

pkg/sypgp/sypgp.go
```diff
@@ -612,6 +619,140 @@ func SearchPubkey(search, keyserverURI, authToken string) error {
...
+		algorithmName = "DSA"
+	case 18:
+		algorithmName = "Elliptic Curve"
+	case 19:
+		algorithmName = "ECDSA"
+	case 20:
+		algorithmName = "Reserved"
+	case 21:
+		algorithmName = "Diffie-Hellman"
+	}
```
mem, line 809: _maybe_ add:

```
default:
    algorithmName = "unknown"
```

pkg/sypgp/sypgp.go
```diff
@@ -612,6 +619,140 @@ func SearchPubkey(search, keyserverURI, authToken string) error {
...
+	ret := "NULL"
+	if s == "" {
+		return ret
+	}
+	if s == "none" {
+		return s + "		  	       "
+	}
+
+	c, _ := strconv.ParseInt(s, 10, 64)
+	ret = fmt.Sprintf("%v", time.Unix(c, 0))
```
mem: this is using [time.Time.String()](https://godoc.org/time#Time.String) to format the time.Time value returned from time.Unix into a string. Note the comment in the documentation that says:

> The returned string is meant for debugging; for a stable serialized representation, use t.MarshalText, t.MarshalBinary, or t.Format with an explicit format string.

you might want to use time.Format instead.

pkg/sypgp/sypgp.go
```diff
@@ -612,6 +619,140 @@ func SearchPubkey(search, keyserverURI, authToken string) error {
...
+
+	return strings.TrimSpace(ret)
```
mem: not quite sure why strings.TrimSpace is needed here. The format used here does not have leading or trailing blanks.

pkg/sypgp/sypgp.go
```diff
@@ -612,6 +619,140 @@ func SearchPubkey(search, keyserverURI, authToken string) error {
...
+}
+
+//function to obtain a date format from linux epoch time
+func date(s string) string {
+	ret := "NULL"
+	if s == "" {
+		return ret
+	}
+	if s == "none" {
+		return s + "		  	       "
```
mem: this has embedded tabs. It would be better to use \t for those instead.

pkg/sypgp/sypgp.go
```diff
@@ -612,6 +619,140 @@ func SearchPubkey(search, keyserverURI, authToken string) error {
...
+		algorithmName = "Diffie-Hellman"
+	}
+	return algorithmName, nil
+}
+
+//function to obtain a date format from linux epoch time
+func date(s string) string {
+	ret := "NULL"
+	if s == "" {
+		return ret
```
mem: it might be simpler to use `return "NULL"` since the "NULL" value is used only here.

pkg/sypgp/sypgp.go
```diff
@@ -612,6 +619,140 @@ func SearchPubkey(search, keyserverURI, authToken string) error {
...
+	c, _ := strconv.ParseInt(s, 10, 64)
+	ret = fmt.Sprintf("%v", time.Unix(c, 0))
+
+	return strings.TrimSpace(ret)
+}
+
+//function to reformat key search output in machine readable format
+func reformatMachineReadableOutput(keyText string) string {
+
+	var output = "      		FINGERPRINT			ALGORITHM  SIZE (BITS)	       CREATION DATE			EXPIRATION DATE		  STATUS		NAME/EMAIL" + "\n"
```
mem: same comment regarding tabs here.

pkg/sypgp/sypgp.go
```diff
@@ -612,6 +619,140 @@ func SearchPubkey(search, keyserverURI, authToken string) error {
...
+		return s + "		  	       "
+	}
+
+	c, _ := strconv.ParseInt(s, 10, 64)
+	ret = fmt.Sprintf("%v", time.Unix(c, 0))
+
+	return strings.TrimSpace(ret)
+}
+
+//function to reformat key search output in machine readable format
```
mem: please follow convention for documentation

pkg/sypgp/sypgp.go
```diff
@@ -612,6 +619,140 @@ func SearchPubkey(search, keyserverURI, authToken string) error {
...
+func reformatMachineReadableOutput(keyText string) string {
+
+	var output = "      		FINGERPRINT			ALGORITHM  SIZE (BITS)	       CREATION DATE			EXPIRATION DATE		  STATUS		NAME/EMAIL" + "\n"
+
+	rePubkey := regexp.MustCompile("pub:(.*)\n")
```
mem, line 832: I'm not quite sure I understand what the input is here.

Is it multiple lines that look like:

```
pub:something here
pub:something else
```

are there leading blanks?

The rest of the code makes me think that this is NOT the case. You could please include a sample input as a comment in the code, so that other people looking at the code in the future know what the code is working with?

pkg/sypgp/sypgp.go
```diff
@@ -612,6 +619,140 @@ func SearchPubkey(search, keyserverURI, authToken string) error {
...
+	}
+
+	c, _ := strconv.ParseInt(s, 10, 64)
+	ret = fmt.Sprintf("%v", time.Unix(c, 0))
+
+	return strings.TrimSpace(ret)
+}
+
+//function to reformat key search output in machine readable format
+func reformatMachineReadableOutput(keyText string) string {
```
tri-adam: I'm not sure this shouldn't return `(string, error)`. rather than just `string`. If `keyText` is malformed, the caller has no way to detect that a problem occurred.

pkg/sypgp/sypgp.go
```diff
@@ -612,6 +619,140 @@ func SearchPubkey(search, keyserverURI, authToken string) error {
...
+func date(s string) string {
+	ret := "NULL"
+	if s == "" {
+		return ret
+	}
+	if s == "none" {
+		return s + "		  	       "
+	}
+
+	c, _ := strconv.ParseInt(s, 10, 64)
```
tri-adam, line 821: I don't think it's a good idea to ignore `err` here. When `ParseInt` fails:

> If s is empty or contains invalid digits, err.Err = ErrSyntax and the returned value is 0; if the value corresponding to s cannot be represented by a signed integer of the given size, err.Err = ErrRange and the returned value is the maximum magnitude integer of the appropriate bitSize and sign.

This will make for some interesting behaviour when these `c` values are interpreted by `time.Unix`. I'd strongly prefer to raise the error to the caller when invalid input is detected.

pkg/sypgp/sypgp.go
```diff
@@ -612,6 +619,140 @@ func SearchPubkey(search, keyserverURI, authToken string) error {
...
+			} else {
+				emailList = detailsEmail[1] + "\n"
+			}
+
+		}
+
+		algorithm, err := getEncryptionAlgorithmName(featuresKey[2])
+
+		if err != nil {
+			return ""
```
tri-adam: Would prefer returning the error explicitly rather than empty output.

pkg/sypgp/sypgp.go
```diff
@@ -593,8 +595,13 @@ func SearchPubkey(search, keyserverURI, authToken string) error {
...
 	}
 
+	//set the machine readable output on
+	var options []string
+	options = append(options, "mr")
 	// Retrieve first page of search results from Key Service.
-	keyText, err := c.PKSLookup(context.TODO(), &pd, search, client.OperationIndex, true, false, nil)
+	keyText, err := c.PKSLookup(context.TODO(), &pd, search, client.OperationIndex, true, false, options)
+
+	keyText = reformatMachineReadableOutput(keyText)
```
tri-adam: There's no point using `keyText` until you've checked `err != nil`. Suggest removing this after that check.

tri-adam (changes requested):
A few comments!

pkg/sypgp/sypgp.go
```diff
@@ -607,11 +611,157 @@ func SearchPubkey(search, keyserverURI, authToken string) error {
...
+	ret := fmt.Sprintf("%v", time.Unix(c, 0))
+
+	return ret, nil
+}
+
+// reformatMachineReadableOutput reformats the key search output that is in machine readable format
+// see the output format in: https://tools.ietf.org/html/draft-shaw-openpgp-hkp-00#section-5.2
+func reformatMachineReadableOutput(keyText string) (string, error) {
+
+	var output = "\n\t\tFINGERPRINT\t\tALGORITHM  SIZE (BITS)\t\t  CREATION DATE\t\t\tEXPIRATION DATE\t\t\tSTATUS\t\t\tNAME/EMAIL" + "\n"
```
WestleyK, line 830: I'd prefer to use spaces here, instead of tabs. Depending on your `tabs` settings, it could use 4 spaces as a tab, instead of 8.

pkg/sypgp/sypgp.go
```diff
@@ -607,11 +611,157 @@ func SearchPubkey(search, keyserverURI, authToken string) error {
...
+	c, _ := strconv.ParseInt(s, 10, 64)
+	ret := fmt.Sprintf("%v", time.Unix(c, 0))
+
+	return ret, nil
+}
+
+// reformatMachineReadableOutput reformats the key search output that is in machine readable format
+// see the output format in: https://tools.ietf.org/html/draft-shaw-openpgp-hkp-00#section-5.2
+func reformatMachineReadableOutput(keyText string) (string, error) {
+
```
WestleyK: Extra newline

```suggestion
```

pkg/sypgp/sypgp.go
```diff
@@ -607,11 +611,157 @@ func SearchPubkey(search, keyserverURI, authToken string) error {
...
+		// regular expression to obtain the email or emails from every key
+		reFormatEmail := regexp.MustCompile("uid:" + "\\w(.)+::")
+		userEmails := reFormatEmail.FindAllString(emails[0], -1)
+
+		for _, email := range userEmails {
+			detailsEmail := strings.Split(email, ":")
+
+			if emailList != "" {
+				emailList = emailList + "\n\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t" + detailsEmail[1] + "\n"
```
WestleyK, line 864: Same here about the tabs.

pkg/sypgp/sypgp.go
```diff
@@ -607,11 +611,157 @@ func SearchPubkey(search, keyserverURI, authToken string) error {
...
+			}
+			if featuresKey[6] == "e" {
+				status = "expired"
+			}
+		}
+
+		featuresKey = []string{}
+
+		if status == "revoked" {
+			output = output + "\n" + fingerprint + "  " + algorithm + "	      " + size + "      " + creationTimestamp + "	  " + expirationTimestamp + "\t" + status + "\t\t" + emailList
```
WestleyK, line 916: Also here

WestleyK (changes requested):
Looks good, left a couple suggestions about using spaces vs tabs.

pkg/sypgp/sypgp.go
jchavezlavalle: @WestleyK it was spaces before, I changed it to tabs after @mem 's review, lmk which is the best approach and I'll do the appropriate changes :+1:

pkg/sypgp/sypgp.go
jchavezlavalle: See the comment before :top: thanks @WestleyK

jchavezlavalle:
@mem @WestleyK any updates on the tabs vs. spaces topic? it was spaces before then changed to tabs after Marcelo's review, lmk and I will update as expected :+1: thanks

mem (approved):

WestleyK (approved):
In my opinion, its better to use spaces instead of tabs in the formatting output. By using tabs, different systems with different `tabs` settings will print tabs differently. For example `tabs` could be set to 4 spaces,  instead of the more common 8 spaces per tab. But, @mem gave the :+1: so we will ship with this. We could always change this later if its a problem. 

Thanks @devprincess for tackling this! :+1:

jchavezlavalle:
@WestleyK @mem @tri-adam I think this one is ready to go, just rebased and all reviews have been addressed, just reminding, thanks!

tri-adam (approved):
LGTM

PR #3676, opened 2019-06-04
Starter cleanup

cclerget:

- Split starter configuration structure into smaller and descriptive structure
- Replace chdir/file descriptor sharing magic by explicit call to `KeepFileDescriptor`
- Rename prepare_stage by apply_container_privileges
- Rework few functions prototype for better readability/understanding
- Remove useless code and cleanup 

- Fixes #


cmd/starter/c/starter.c
```diff
@@ -298,63 +270,38 @@ static int create_namespace(int nstype) {
 }
 
 static int enter_namespace(char *nspath, int nstype) {
+    int ret = -1;
```
bauerm97: What's the point of declaring `int ret = -1` if we are never doing `return(ret)`? This line doesn't seem to have an effect to me

cmd/starter/c/starter.c
```diff
@@ -371,38 +318,30 @@ static int enter_namespace(char *nspath, int nstype) {
     return(0);
 }
 
-static void setup_userns_mappings(struct cConfig *config, pid_t pid, const char *setgroup) {
+static void setup_userns_mappings(struct privileges *privileges) {
```
bauerm97, line 337: Slightly confused here. In previous function we are writing to `/proc/$pid/$file` where `$file` is either `setgroups`, `gid_map`, or `uid_map`. It appears in the updated version of the function we are writing only to `$file` in the current directory. Is this function intended to be called from `/proc/$pid`? If so, I think we should add that as a comment to the function to clarify its expected usage

cmd/starter/c/starter.c
```diff
@@ -42,77 +42,72 @@
...
 #include "include/capability.h"
 #include "include/message.h"
 #include "include/starter.h"
 
-#define CLONE_STACK_SIZE    1024*1024
-#define BUFSIZE             512
-
 /* C and JSON configuration */
-struct cConfig *config;
+struct starterConfig *sconfig;
```
bauerm97, line 50: Change comment on this variable to match new naming convention:

```
/* starter configuration struct */
struct cConfig *config;	struct starterConfig *sconfig;
```

cmd/starter/c/starter.c
```diff
@@ -893,104 +826,128 @@ __attribute__((constructor)) static void init(void) {
...
+     */
+    pipe_fd = get_pipe_exec_fd();
+
     /* cleanup environment variables */
     cleanenv();
 
-    // initialize starter configuration and share it with child processes
-    config = (struct cConfig *)mmap(NULL, sizeof(struct cConfig), PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_SHARED, -1, 0);
-    if ( config == MAP_FAILED ) {
+    /* initialize starter configuration and share it with child processes */
```
bauerm97: I would prefer to be slightly more verbose here: `/* initialize starter configuration in shared memory to later share with child processes */`

cmd/starter/c/starter.c
```diff
@@ -893,104 +826,128 @@ __attribute__((constructor)) static void init(void) {
...
+     *  step below.
      */
-    stage_pid = fork_ns(CLONE_FILES|CLONE_FS);
-    if ( stage_pid == 0 ) {
+    process = fork_ns(CLONE_FILES);
+    if ( process == 0 ) {
         /*
-         *  stage1 is responsible for singularity configuration file parsing, handle user input,
-         *  read capabilities, check what namespaces is required.
+         *  stage1 is responsible for singularity configuration file parsing
```
bauerm97: Missing a comma at end of line here ;)

cmd/starter/c/starter.c
```diff
@@ -893,104 +826,128 @@ __attribute__((constructor)) static void init(void) {
...
+         *  handle user input, read capabilities, check what namespaces is required.
```
bauerm97: `* handling user input, reading capabilities, and checking what namespaces are required`

cmd/starter/c/starter.c
```diff
@@ -893,104 +826,128 @@ __attribute__((constructor)) static void init(void) {
...
          */
-        if ( config->container.isSuid ) {
-            priv_escalate();
-            execute = prepare_stage(STAGE1, config);
-        } else {
-            set_parent_death_signal(SIGKILL);
-            execute = STAGE1;
+        if ( sconfig->starter.isSuid && uid != 0 ) {
```
bauerm97, line 927: I'm a little bit confused here. I think intention here is to permanently drop privileges back to the original uid/gid? If that's the case, maybe it's best to create a secondary function similar to `priv_drop()` which drops privileges permanently in this manner?

`static void priv_drop_permanent(void)`

This way it's a bit more clear in the code that we are intentionally dropping all privileges permanently, instead of just calling `setresguid`/`setresuid` directly

sashayakovtseva (dismissed):

cmd/starter/c/starter.c
```diff
@@ -813,22 +751,34 @@ static void fix_streams(void) {
     }
 }
 
-static void exit_with_status(const char *name, int status) {
+static void wait_child(const char *name, pid_t child_pid, bool stop) {
```
bauerm97: Can you clarify what `bool stop` argument is doing and why we are using it? Perhaps I just haven't gotten to the part in the code which will call with `bool stop = true` as argument, so I keep reading. But a comment on this function `wait_child` which explains usage of `bool stop` would be welcomed

cmd/starter/c/starter.c
```diff
@@ -813,22 +751,34 @@ static void fix_streams(void) {
...
+    int status;
+
+    while ( 1 ) {
```
bauerm97: I see that we're introducing a `while ( 1 )` loop here, which will only break when `waitpid(...)` returns the value of `child_pid`. Is it somehow possible for a race condition or other weird circumstance to trigger infinite loop of triggering the `else {...}` block in this loop indefinitely and never breaking? If not, perhaps a comment explaining why this should never happen to aid a layman like myself in reading

cmd/starter/c/starter.c
```diff
@@ -813,22 +751,34 @@ static void fix_streams(void) {
...
+        } else if ( pid == child_pid ) {
+            break;
+        } else {
+            debugf("Got a terminated child with pid %d", pid);
+        }
+    }
     if ( WIFEXITED(status) ) {
-        verbosef("%s exited with status %d\n", name, WEXITSTATUS(status));
-        exit(WEXITSTATUS(status));
+        if ( stop || WEXITSTATUS(status) != 0 ) {
```
bauerm97: This still reads a little magical to me, as I'm not sure what intention of `stop` variable is.

cmd/starter/c/starter.c
```diff
@@ -860,27 +810,10 @@ static void cleanenv(void) {
...
-    int sync_pipe[2];
-    struct fdlist *fd_before;
-    struct fdlist *fd_after;
-
-    verbosef("Starter initialization\n");
-
-#ifndef SINGULARITY_NO_NEW_PRIVS
-    fatalf("Host kernel is outdated and does not support PR_SET_NO_NEW_PRIVS!\n");
-#endif
+static int get_pipe_exec_fd(void) {
```
bauerm97, line 837: I think adding a comment would be useful here:

```
// get_pipe_exec_fd returns the pipe file descriptor stored in the PIPE_EXEC_FD environment variable 
static int get_pipe_exec_fd(void) {
```

Or something similar. Just for some more verbosity

cmd/starter/c/starter.c
```diff
@@ -893,104 +826,128 @@ __attribute__((constructor)) static void init(void) {
...
-    }
+    wait_child("stage 1", process, 0);
 
-    if ( WIFEXITED(status) && WEXITSTATUS(status) != 0 ) {
-        verbosef("stage 1 exited with status %d\n", WEXITSTATUS(status));
-        exit(WEXITSTATUS(status));
-    } else if ( WIFSIGNALED(status) ) {
-        verbosef("stage 1 interrupted by signal number %d\n", WTERMSIG(status));
-        kill(getpid(), WTERMSIG(status));
+    /* is stage 1 want to change current working directory ? */
```
bauerm97: `/* change current working directory if requested by stage1 */`

cmd/starter/c/starter.c
```diff
@@ -1002,182 +959,235 @@ __attribute__((constructor)) static void init(void) {
...
-            if (sigprocmask(SIG_UNBLOCK, &usrmask, NULL) == -1) {
+            if ( sigprocmask(SIG_UNBLOCK, &usrmask, NULL) == -1 ) {
                 fatalf("Unblock signals error: %s\n", strerror(errno));
             }
-            while ( waitpid(forked, &status, 0) <= 0 ) {
-                continue;
-            }
-            exit_with_status("instance", status);
+            /* loop until master process exits with error */
+            wait_child("instance", process, 1);
```
bauerm97: I originally had some confusion regarding the intention of this function call here, and whether it would ever return and unintentionally continue to execute `init()`. After a slack conversation with @cclerget I think I understand what's going on, which I'll try to recap:

- When requested to run as an instance, we `fork()` to create a parent/child
- The child process calls `umask(0)` and then continues through `init()`
- The parent process exists to monitor if `master` process succeeded
- Parent process creates a handler for `SIGUSR1`. Master process will send `SIGUSR1` if master stage is successful. The signal handler declared will then essentially `exit(0)` as it was successful.
- `wait_child("instance", process, 1);` is called. The third arg of `wait_child` signifies if the function should exit after waiting on the child has finished.
- There is thus no case where `wait_child` call should return and allow the parent process to continue through `init()` which was my initial worry.

I would suggest that to increase clarity and intent, we do this:

```
wait_child("instance", process, 1);

/* we should never return from the previous wait_child call */
exit(1);
```

This way we are signaling our intent in code & comment

cmd/starter/c/starter.c
```diff
@@ -699,96 +630,103 @@ static struct fdlist *list_fd(void) {
...
-     */
-    for ( i = 0; i < fd_after->num; i++ ) {
-        struct stat st;
-        int found;
+    if ( ( dir = fdopendir(fd_proc) ) == NULL ) {
+        fatalf("Failed to list /proc/self/fd directory: %s\n", strerror(errno));
+    }
 
-        if ( fd_after->fds[i] == master_socket[0] || fd_after->fds[i] == master_socket[1] ) {
+    while ( ( dirent = readdir(dir ) ) ) {
```
bauerm97: Extra space after `dir` here

internal/app/starter/master_linux.go
```diff
@@ -65,9 +65,6 @@ func Master(rpcSocket, masterSocket int, isInstance bool, containerPid int, engi
 		}); ok {
 			n, err := conn.Read(data)
 			if (err != nil && err != io.EOF) || n == 0 || data[0] == 'f' {
-				if isInstance && os.Getppid() == ppid {
```
bauerm97, line 68: I'm assuming that we remove this because we're now only sending a signal if master succeeds (`SIGUSR1`) and otherwise issuing a non-zero exit code to signal to `init()` C function that master process failed. If that's the case, then shouldn't we not just `return` from here but actually either `exit(1)` or `sylog.Fatalf()` or similar?

bauerm97: Or, similar to Line 85 in the new code do `fatalChan <- fmt.Errorf(...)`?

bauerm97 (changes requested):
Sorry for leaving so many comments :P was a lot to go through! Mostly looks really good, there's a few areas where I think we need small tweaks/tunes for clarity but overall it's awesome

cmd/starter/c/starter.c
cclerget: Ah right, dead code

cmd/starter/c/starter.c
cclerget: Will do

cmd/starter/c/starter.c
cclerget: Will fix that

cmd/starter/c/starter.c
cclerget: I will remove while loop, because we should (based on man read) have only two conditions (error or child_pid) with how `waitpid` is called currently

cmd/starter/c/starter.c
cclerget: will find a better name and add comments

internal/app/starter/master_linux.go
cclerget: I will look

bauerm97 (approved):
I think all of my comments have been addressed. Looks good to me, ready to merge!

PR #3677, opened 2019-06-04
Add a command to print the command tree for singularity

### LINKED ISSUE #3570: Need a method to dump cobra command and flags for singularity
## Version of Singularity:

master

We need a way to dump all of the commands and options `singularity` uses via the `cobra` pkg.

I think we can use the methods that are used to generate docs to do this.

jscook2345:

Add a command to dump the command tree for singularity

- Fixes #3570


mem (approved):

mem (approved):

GodloveD:
@jscook2345 sorry for the late review here (after being merged).  I think it would be good if this command were made hidden since we don't really expect it to be useful to a normal user and don't want it cluttering up the help.  Can someone open another PR and add the hidden flag to cobra?

jscook2345:
@GodloveD 

It didn't get added as a 'cobra' command really. It's more like a `go` script similar to how our man pages are.

You'd run it by doing `go run <options> ./cmd/docs/cmds/cmdtree.go`. The output is just the JSON. I didn't think it needed any more fancying up quite yet until we can generate coverage from our tests to.

GodloveD:
Oh perfect.  Thanks for the info.

Issue #2864, opened 2019-03-12
plugin/hooks: add appropriate hooks for exposing CVMFS transparently


bauerm97:
cc @DrDaveD 

As we discussed during the break at SUG19, we should investigate and understand how to add a set of plugin hooks which can give something like CVMFS the appropriate features to transparently and automatically mount itself into the container. Once we have an understanding of the specific hooks, I will update title and post to reference them

DrDaveD:
Adding @sfiligoi

bauerm97:
@DrDaveD Thank you!

poquirion:
That would be great! We use https://github.com/cooperative-computing-lab/cctools to get cvmfs working and that can leads to problems when kernel calls are involved...

DrDaveD:
Over the last few days I have learned that libfuse added an extremely helpful new relevant [feature](https://github.com/libfuse/libfuse/pull/291) not very long ago in release [3.3.0](https://github.com/libfuse/libfuse/releases/tag/fuse-3.3.0) that will make this quite easy for singularity to do along with an upgrade of [cvmfs to fuse3](https://sft.its.cern.ch/jira/browse/CVM-1710).   An option could be added to singularity, I suggest `--fusecmd 'command'`, that would open /dev/fuse, associate that file descriptor N with the mountpoint that is the last parameter to 'command', and replace the last parameter with /dev/fd/N.   Then it will execute the command unprivileged inside the container, and if the command starts a FUSE filesystem daemon based on libfuse >= 3.3.0, it will just work.  

As a shorthand, there could also be a new `--fusemount type:source:destination[:options]` that translates to `--fusecmd 'mount -t type -o options source destination`.   At that point, a cvmfs plugin could be added as simply another shorthand option that translates `--cvmfs repo1[:repo2:...:repoN]` to a series of `--fusemount cvmfs:repoN:/cvmfs/repoN` options.  On the other hand we could skip having a cvmfs plugin and just use --fusemount options.

[...]

PR #3679, opened 2019-06-04
Add support for mounting FUSE filesystems in the container

mem:
Since it's not possible to completely delegate the operations to a
plugin, add support for mounting FUSE filesystems in the container
WITHOUT adding support for any specific driver or method of operation.

People wishing to support specific FUSE filesystems would need to add a
plugin that:

1. Adds a flag to specify the mount point
2. Adds a flag to specify the FUSE driver to use (along with parameters)

The reason for doing it like this is that depending on the specifics of
the FUSE filesystem that needs to be supported, more or less support
would need to be added in singularity. By adding just the necessary
support to open the /dev/fuse device and run the FUSE driver, we make
sure that we don't under- or over-specify the assumptions as to what the
driver needs or supports.

Users wishing to use FUSE filesystem in their containers would need to
make sure that the _container_ provides the necessary support to run the
driver (e.g. by shipping the driver with the container image or by
setting up bind mounts to make sure the driver is in the expected
location).

This extends the Engine configuration by adding a Plugin object which
contains entries for each of the plugins. Nothing is assumed about the
configuration. The FUSE portion *does* require that the configuration
of each plugin has a "Fuse" key pointing to an object that has _at_
least a DevFuseFd, MountPoint and Program values. Plugins that do not
support FUSE filesystems do not need to provide these keys or objects.

### DIFF
--- a/internal/pkg/runtime/engines/singularity/container_linux.go
+++ b/internal/pkg/runtime/engines/singularity/container_linux.go
@@ -30,6 +30,7 @@ import (
 	"github.com/sylabs/singularity/internal/pkg/util/user"
 	"github.com/sylabs/singularity/pkg/image"
 	"github.com/sylabs/singularity/pkg/network"
+	singularity "github.com/sylabs/singularity/pkg/runtime/engines/singularity/config"
 	"github.com/sylabs/singularity/pkg/util/fs/proc"
 	"github.com/sylabs/singularity/pkg/util/loop"
 	"github.com/sylabs/singularity/pkg/util/namespaces"
@@ -169,6 +170,9 @@ func create(engine *EngineOperations, rpcOps *client.RPC, pid int) error {
 		return err
 	}
 	if err := c.addHostnameMount(system); err != nil {
+		return err
+	}
+	if err := c.addFuseMount(system); err != nil {
 		return err
 	}
 
@@ -1895,4 +1899,66 @@ func (c *container) prepareNetworkSetup(system *mount.System, pid int) (func() e
 		c.engine.EngineConfig.Network = setup
 		return nil
 	}, nil
+}
+
+// addFuseMount transforms the plugin configuration into a series of
+// mount requests for FUSE filesystems
+func (c *container) addFuseMount(system *mount.System) error {
+	for i, name := range c.engine.EngineConfig.GetPluginFuseMounts() {
+		var cfg struct {
+			Fuse singularity.FuseInfo
+		}
+
+		if err := c.engine.EngineConfig.GetPluginConfig(name, &cfg); err != nil {
+			sylog.Debugf("Failed getting plugin config: %+v\n", err)
+			return err
+		}
+
+		// we cannot check this because the mount point might
+		// not exist outside the container with the name that
+		// it's going to have _inside_ the container, so we
+		// simply assume that it is a directory.
+		//
+		// In a normal situation we would stat the mount point
+		// to obtain the mode, and bitwise-and the result with
+		// S_IFMT.
+		rootmode := syscall.S_IFDIR & syscall.S_IFMT
+
+		// we assume that the file descriptor we obtained by
+		// opening /dev/fuse before is valid in the RPC server,
+		// where the actual mount operation is going to be
+		// executed.
+		opts := fmt.Sprintf("fd=%d,rootmode=%o,user_id=%d,group_id=%d",
+			cfg.Fuse.DevFuseFd,
+			rootmode,
+			os.Getuid(),
+			os.Getgid())
+
+		// mount file system in three steps: first create a
+		// dedicated session directory for each FUSE filesystem
+		// and use that as the mount point.
+		fuseDir := fmt.Sprintf("/fuse/%d", i)
+		if err := c.session.AddDir(fuseDir); err != nil {
+			return err
+		}
+		fuseDir, _ = c.session.GetPath(fuseDir)
+		sylog.Debugf("Add FUSE mount %s for %s with options %s", fuseDir, name, opts)
+		if err := system.Points.AddFS(
+			mount.BindsTag,
+			fuseDir,
+			"fuse",
+			syscall.MS_NOSUID|syscall.MS_NODEV,
+			opts); err != nil {
+			sylog.Debugf("Calling AddFS: %+v\n", err)
+			return err
+		}
+
+		// second, add a bind-mount the session directory into
+		// the destination mount point inside the container.
+		if err := system.Points.AddBind(mount.OtherTag, fuseDir, cfg.Fuse.MountPoint, 0); err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
--- a/internal/pkg/runtime/engines/singularity/prepare_linux.go
+++ b/internal/pkg/runtime/engines/singularity/prepare_linux.go
@@ -799,6 +799,42 @@ func (e *EngineOperations) PrepareConfig(starterConfig *starter.Config) error {
 	// determine if engine need to propagate signals across processes
 	e.checkSignalPropagation()
 
+	// We must call this here because at this point we haven't
+	// spawned the master process nor the RPC server. The assumption
+	// is that this function runs in stage 1 and that even if it's a
+	// separate process, it's created in such a way that it's
+	// sharing its file descriptor table with the wrapper / stage 2.
+	//
+	// At this point we do not have elevated privileges. We assume
+	// that the user running singularity has access to /dev/fuse
+	// (typically it's 0666, or 0660 belonging to a group that
+	// allows the user to read and write to it).
+	if err := openDevFuse(e, starterConfig); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// openDevFuse is a helper function that opens /dev/fuse once for each
+// plugin that wants to mount a FUSE filesystem.
+func openDevFuse(e *EngineOperations, starterConfig *starter.Config) error {
+	for _, name := range e.EngineConfig.GetPluginFuseMounts() {
+		fd, err := syscall.Open("/dev/fuse", syscall.O_RDWR, 0)
+		if err != nil {
+			sylog.Debugf("Calling open: %+v\n", err)
+			return err
+		}
+
+		err = e.EngineConfig.SetPluginFuseFd(name, fd)
+		if err != nil {
+			sylog.Debugf("Unable to setup plugin %s fd: %+v\n", name, err)
+			return err
+		}
+
+		starterConfig.KeepFileDescriptor(fd)
+	}
+
 	return nil
 }
 
--- a/internal/pkg/runtime/engines/singularity/process_linux.go
+++ b/internal/pkg/runtime/engines/singularity/process_linux.go
@@ -9,6 +9,7 @@ import (
 	"debug/elf"
 	"encoding/binary"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net"
 	"os"
@@ -22,6 +23,7 @@ import (
 	"unsafe"
 
 	"github.com/sylabs/singularity/internal/pkg/security"
+	singularity "github.com/sylabs/singularity/pkg/runtime/engines/singularity/config"
 
 	"github.com/sylabs/singularity/internal/pkg/util/user"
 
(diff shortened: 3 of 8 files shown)


examples/plugins/test-plugin/main.go
```diff
@@ -32,18 +32,39 @@ type pluginImplementation struct {
...
 			fmt.Printf("Received flag value: %s\n", f.Value.String())
+
+			config := pluginConfig{
+				Fuse: FuseConfig{
+					MountPoint: f.Value.String(),
+					Program:    []string{"/usr/local/bin/fuse-example"},
+				},
+			}
+			if err := cfg.SetPluginConfig("test-plugin", config); err != nil {
+				fmt.Printf("Cannot set plugin configuration: %+v\n", err)
```
WestleyK: Should this print to stderr?

```suggestion
				fmt.Fprintf(os.Stderr, "Cannot set plugin configuration: %+v\n", err)
```

examples/plugins/test-plugin/main.go
```diff
@@ -32,18 +32,39 @@ type pluginImplementation struct {
...
+				return
+			}
+
```
WestleyK: Extra newline

```suggestion
```

internal/pkg/runtime/engines/singularity/container_linux.go
```diff
@@ -1742,3 +1745,55 @@ func (c *container) addActionsMount(system *mount.System) error {
...
+	for _, name := range c.engine.EngineConfig.GetPluginFuseMounts() {
+		var cfg struct {
+			Fuse struct {
+				DevFuseFd  int
+				MountPoint string
+			}
+		}
+
+		if err := c.engine.EngineConfig.GetPluginConfig(name, &cfg); err != nil {
+			sylog.Debugf("Getting plugin config: %+v\n", err)
```
WestleyK: Should this be:

```suggestion
			sylog.Debugf("Failed getting plugin config: %+v\n", err)
```
Or something like that?

internal/pkg/runtime/engines/singularity/prepare_linux.go
```diff
@@ -705,6 +705,40 @@ func (e *EngineOperations) PrepareConfig(starterConfig *starter.Config) error {
...
+	for _, name := range e.EngineConfig.GetPluginFuseMounts() {
+		fd, err := syscall.Open("/dev/fuse", syscall.O_RDWR, 0)
+		if err != nil {
+			sylog.Debugf("Calling open: %+v\n", err)
+			return err
+		}
+
+		err = e.EngineConfig.SetPluginFuseFd(name, fd)
+		if err != nil {
+			sylog.Debugf("Setting plugin %s fd: %+v\n", name, err)
```
WestleyK: Why are these `DEGUB`s not saying something like:

```suggestion
			sylog.Debugf("Unable to setup plugin %s fd: %+v\n", name, err)
```

WestleyK (commented):
Couple questions, otherwise looks good to my knowledge. Maybe @bauerm97, or @ikaneshiro can take a better look :+1:

internal/pkg/runtime/engines/singularity/process_linux.go
```diff
@@ -145,8 +145,69 @@ func (engine *EngineOperations) checkExec() error {
...
+				MountPoint string
+				Program    []string
+			}
+		}
+
+		if err := engine.EngineConfig.GetPluginConfig(name, &cfg); err != nil {
+			return err
+		}
+
+		fh := os.NewFile(uintptr(cfg.Fuse.DevFuseFd), "fd-"+name)
```
ikaneshiro: There is a chance this could return a `nil` value causing the subsequent dereference to panic

pkg/runtime/engines/singularity/config/config.go
```diff
@@ -485,3 +489,98 @@ func (e *EngineConfig) GetDeleteImage() bool {
...
+// SetPluginFuseFd sets the /dev/fuse file descriptor fd for the
+// specified plugin
+//
+// This function tries to make sure that any additional configuration
+// already found in the "Fuse" object is preserved.
+func (e *EngineConfig) SetPluginFuseFd(name string, fd int) error {
+	raw, found := e.Plugin[name]
+	if !found {
+		// named plugin does not have a configuration
+		// entry, skip
```
ikaneshiro: Why are these comments indicating that these calls are skipped when `openDevFuse()` is catching and stopping iteration over plugin fuse mounts if an error is encountered

ikaneshiro (changes requested):

DrDaveD:
@mem I don't fully understand -- what is it about the generic --fusecmd option that I proposed in #2864 was not good?

Also, do you have an example plugin that uses this for sshfs?

internal/pkg/runtime/engines/singularity/process_linux.go
```diff
@@ -145,8 +145,69 @@ func (engine *EngineOperations) checkExec() error {
...
+		// descriptors to be passed to the new process.
+		//
+		// ExtraFiles is an array of *os.File, with the position
+		// of each entry determining the resulting file
+		// descriptor number. It must correspond to the file
+		// descriptor we have so that the above /dev/fd/%d
+		// works. Use fd-3 because we must skip over stdin,
+		// stdout, stderr; +1 because the second argument to
+		// make is the number of elements.
+		cmd.ExtraFiles = make([]*os.File, fd-3+1)
```
cclerget: What if you just pass `fh` to `ExtraFiles` and set `/dev/fd/3` above instead ?

internal/pkg/runtime/engines/singularity/process_linux.go
```diff
@@ -145,8 +145,69 @@ func (engine *EngineOperations) checkExec() error {
...
+	}
+
+	return nil
+}
+
 // StartProcess starts the process
 func (engine *EngineOperations) StartProcess(masterConn net.Conn) error {
+	// perform all the necessary "pre" operations, right before
+	// starting the user's process
+	if err := preStartProcess(engine); err != nil {
```
cclerget, line 243: As we already discussed, I have concerns about executing FUSE server inside container :
- it implies to compile FUSE server statically
- some FUSE server may need to access host filesystem or specific device which may not be made with few singularity flags like `--contain` enabled
- when container finish FUSE server can be killed without having time to flush pending data
- seccomp filter when enable could potentially affect FUSE server

internal/pkg/runtime/engines/singularity/container_linux.go
```diff
@@ -1742,3 +1745,55 @@ func (c *container) addActionsMount(system *mount.System) error {
...
+		// opening /dev/fuse before is valid in the RPC server,
+		// where the actual mount operation is going to be
+		// executed.
+		opts := fmt.Sprintf("fd=%d,rootmode=%o,user_id=%d,group_id=%d",
+			cfg.Fuse.DevFuseFd,
+			rootmode,
+			os.Getuid(),
+			os.Getgid())
+
+		err := system.Points.AddFSWithSource(
```
cclerget: You can use `AddFS` as source and fstype are identical here

cclerget (changes requested):

examples/plugins/test-plugin/main.go
```diff
@@ -32,18 +32,39 @@ type pluginImplementation struct {
 
 var impl = pluginImplementation{}
 
+type FuseConfig struct {
```
mem: oops, this shouldn't be here.

internal/pkg/runtime/engines/singularity/process_linux.go
mem: You are strictly correct: if the filedescriptor is invalid (or a really large value?), it's possible that the various casts inside NewFile produce an error. That's the only path that I could find that could cause this to return nil. I'll add a check for this.

DrDaveD:
I got the test-plugin to work with sshfs, by binding in a script to /usr/local/bin/fuse-example that executes "sshfs <myfavoritehost>: "$@", and adding a test invocation option "-z /srv".  

However when I had errors in that script it just died with `FATAL:   exit status 1`.  Stdout and stderr of the command went nowhere apparently, and there was no indication that the error was in the mount.  I suppose the latter is a flaw in the test-plugin, but I don't understand why stdout & stderr weren't seen.  There wasn't even any indication in -vvvv that it was about to execute the command.  It did successfully prompt for a passphrase when I had the right syntax in fuse-example, I saw the prompt.

internal/pkg/runtime/engines/singularity/process_linux.go
mem: Yeah, that should work, thanks.

internal/pkg/runtime/engines/singularity/process_linux.go
mem: And as I have already said, that's a problem for the person implementing the plugin. That's precisely the reason why this is implemented like this.

pkg/runtime/engines/singularity/config/config.go
mem: LOL... that's fair.

I mean "skip" in the sense that this won't be doing anything with this plugin. The "skip" part was probably more accurate in the original context where this code existed. I'll fix this. Thanks.

mem:
> I got the test-plugin to work with sshfs, by binding in a script to /usr/local/bin/fuse-example that executes "sshfs : "$@", and adding a test invocation option "-z /srv".
> 
> However when I had errors in that script it just died with `FATAL: exit status 1`. Stdout and stderr of the command went nowhere apparently, and there was no indication that the error was in the mount. I suppose the latter is a flaw in the test-plugin, but I don't understand why stdout & stderr weren't seen. There wasn't even any indication in -vvvv that it was about to execute the command. It did successfully prompt for a passphrase when I had the right syntax in fuse-example, I saw the prompt.

Sorry about the trouble. For other people looking for this, I have provided an example here: https://github.com/mem/singularity-fusemnt

What does -z do? I can't find that option in the version of sshfs that I have.

mem:
Also, I think you identified a bug with the way the driver is started. Looking at it.

DrDaveD:
I meant the singularity action command -z option, from the test-plugin, which does a fuse mount with /usr/local/bin/fuse-example as the command.

ikaneshiro (approved):
LGTM, @cclerget?

DrDaveD:
@mem The version of this that I checked out last week worked, but the current version is broken.  If the --fusecmd command I run (using my plugin from pr #3693) has some debug commands in it including a "mount" command (with no parameters), I can see that it has not chrooted into the container space.  I see it mounted at /var/singularity/mnt/session/rootfs, instead of /.  It looks like one of the force pushes you did since then (presumably because you rebased?) broke it.  Maybe the place that the chroot happened in the base branch moved.

cclerget:
> @mem The version of this that I checked out last week worked, but the current version is broken. If the --fusecmd command I run (using my plugin from pr #3693) has some debug commands in it including a "mount" command (with no parameters), I can see that it has not chrooted into the container space. I see it mounted at /var/singularity/mnt/session/rootfs, instead of /. It looks like one of the force pushes you did since then (presumably because you rebased?) broke it. Maybe the place that the chroot happened in the base branch moved.

@DrDaveD The FUSE program is not executed inside container anymore since last changes but from master process which is not impacted by chroot and also by command line options that user could pass (pid namespace, contain ...). If running FUSE program inside container is a requirement, maybe a field like `RunInside` would help to determine if FUSE program request to run inside or outside container, @mem what do you think ?

internal/pkg/runtime/engines/singularity/container_linux.go
```diff
@@ -1766,3 +1771,130 @@ func (c *container) prepareNetworkSetup(system *mount.System, pid int) (func() e
...
+		return nil
+	}
+}
+
+// addFuseMount transforms the plugin configuration into a series of
+// mount requests for FUSE filesystems
+func (c *container) addFuseMount(system *mount.System) error {
+	for i, name := range c.engine.EngineConfig.GetPluginFuseMounts() {
+		var cfg struct {
+			Fuse struct {
```
cclerget: @mem Based on comment https://github.com/sylabs/singularity/pull/3679#issuecomment-501989673, wouldn't be good to have a public package with a defined FUSE config structure and also use this package could used as a doc ?

DrDaveD:
Running the fuse command inside the container is a requirement.  That has been the idea since the beginning.  It would be very challenging to install the software outside of the container in such a way that it is useful inside.  With it inside, it's a simple standard package install.

In fact I'm not very happy with the whole plugin approach at all for this.  We originally were thinking plugin, but when I discovered the new fuse3 library functionality and I proposed the --fusecmd option, I was thinking we could bypass the whole plugin complication along with its security risks.  The --fusecmd option could be implemented in entirely general way, much more simply than a plugin (it could be disabled by default in singularity.conf for suspicious people).  I think giving plugins access to do anything privileged is very risky.

DrDaveD:
The current version works again with cvmfs. 

It does not however work with the simpler-to-test sshfs.  I put an strace around the ssh call that sshfs makes, and it is getting stuck on a SIGTTOU signal.  This did not happen with the original version that I tested on June 6.  The original one prompted for my ssh password.

internal/pkg/runtime/engines/singularity/container_linux.go
```diff
@@ -528,6 +531,7 @@ func (c *container) mountGeneric(mnt *mount.Point) (err error) {
 			defer c.rpcOps.SetFsID(os.Getuid(), os.Getgid())
 		}
 	}
+	sylog.Debugf("src=%s dst=%s type=%s flags=%v opts=%s", source, dest, mnt.Type, flags, optsString)
```
cclerget: To make flags more readable maybe something like :

```
sylog.Debugf("src=%s dst=%s type=%s opts=%s", source, dest, mnt.Type, strings.Join(options, ","))
```

internal/pkg/runtime/engines/singularity/process_linux.go
```diff
@@ -146,8 +147,106 @@ func (engine *EngineOperations) checkExec() error {
...
+	// how many and which file descriptors are open at this point,
+	// simply assume that it's possible to map the existing file
+	// descriptor to the name number in the new process.
+	//
+	// "newFd" should be the same as "fd", but do not assume that
+	// either.
+	newFd := fh.Fd()
+	fdDevice := fmt.Sprintf("/dev/fd/%d", newFd)
+	args := append(program, fdDevice)
+	sylog.Debugf("args=%v", args)
```
cclerget:
```suggestion
	sylog.Debugf("args=%s", strings.Join(args, ","))
```

internal/pkg/runtime/engines/singularity/process_linux.go
```diff
@@ -146,8 +147,106 @@ func (engine *EngineOperations) checkExec() error {
...
+	if err := setupFuseDrivers(engine); err != nil {
+		return err
+	}
+
+	return nil
+}
+
 // StartProcess starts the process
 func (engine *EngineOperations) StartProcess(masterConn net.Conn) error {
+	if err := preStartProcess(engine); err != nil {
```
cclerget, line 243: @DrDaveD mentioned an issue with signals, you will need to queue them before executing the FUSE program

cclerget (changes requested):
Small suggestions, the signal issue however need to be addressed

internal/pkg/runtime/engines/singularity/process_linux.go
DrDaveD: @cclerget I don't recall exactly what you're referring to, unless you mean the issue of leaving behind running FUSE processes if the singularity container is killed.  I deal with that issue by always enabling --pid.  In fact the [pull request on this branch in mem's repository](https://github.com/mem/singularity/pull/2) that I submitted yesterday automatically enables --pid whenever --fusecmd is used.  With that, I don't see a need for doing anything special with signals here.  The fuse command should put itself into a separate session.

On the other hand, there is still a problem with starting sshfs if it requires prompting for a passphrase.  I think the thing to do there is to simply wait for the process.  Ordinarily a FUSE process should put itself into the background.

cclerget: IIRC there was an issue with defunc processes not correctly catched by PID 1 process ?

DrDaveD: Good recalling.  I had forgotten about that but quickly discovered it again when I tried using cmd.Run() instead of cmd.Start() for the fuse command.  I think the solution to that is to just have sinit wait for any orphan children before it starts, as I put into [this pull request in mem's repository](https://github.com/mem/singularity/pull/3).

internal/pkg/runtime/engines/singularity/container_linux.go
mem: The reason why I tried to avoid doing as suggested is because this is the configuration for a plugin. That's why it's a struct inside an anonymous struct: the outside struct is to be defined by the plugin.

I have added a global definition for the inner struct. Not happy about it, but it's there.

internal/pkg/runtime/engines/singularity/process_linux.go
mem: I don't know what you guys are talking about...

```
builddir/singularity shell --fusemnt /mnt ./images/fuse-example.sif
WARNING: Cannot find the fuse-example driver in $PATH.
Singularity fuse-example.sif:~/devel/sylabs/singularity/src/github.com/sylabs/singularity> cat /mnt/hello
Hello World!
```

the driver is fuse-example:

```
$ pgrep fuse-example
10377
$ ps `pgrep fuse-example`
  PID TTY      STAT   TIME COMMAND
10377 pts/5    Sl     0:00 /bin/fuse-example -f /dev/fd/5
```

after exiting the shell by typing `exit` and the command line:

```
$ pgrep fuse-example
$
```

the program is properly stopped.

The reason why this is started with `cmd.Start()` and not `cmd.Run()` is because the driver should *not* background itself. If there's a reason why a driver should background itself, we can talk about it after this merges.

DrDaveD: The fuse drivers that I am familiar with (sshfs & cvmfs) do background themselves so I assume it is normal.  If singularity doesn't wait, then sshfs doesn't work if it needs to prompt for a password, as it does when I test it.  Is fuse-example a realistic example?  Where does it come from?

cclerget: By default FUSE program run in background, if foreground is the expected way to run FUSE program, it may be safer to document that somewhere further or better to enforce this mode in the code. Additionally it will be the plugin responsibility to also reset the `--no-init` option in order to always reach the wait process code in `StartProcess` otherwise the condition `(!isInstance && !shimProcess)` will be met when `--no-init` is set, fixable by setting it to `false` in `setupFuseDrivers` loop. I approve this PR as is, feel free to consider or not those suggestions for further work.

cclerget (approved):

PR #3681, opened 2019-06-05
Fix configuration bad default value and remove old configuration template file

### LINKED ISSUE #3669: Bad "default" values in singularity.conf
## Version of Singularity:

3.1.1 (but seems to be still present in current master)

### Expected behavior

Documented `DEFAULT: ` in config file matches real default. 

### Actual behavior

It is wrong for at least `ROOT DEFAULT CAPABILITIES`. 

Read config file after installation. 

### More details

As seen in https://github.com/sylabs/singularity/blob/master/pkg/runtime/engines/singularity/config/config.go , the default for `RootDefaultCapabilities` is `full`. 
However, as seen in https://github.com/sylabs/singularity/blob/master/etc/singularity.conf.in , the documentation in `singularity.conf` says:
`# DEFAULT: no`
It would be best to just use:
`# DEFAULT: @ROOT_DEFAULT_CAPABILITIES_DEFAULT@`
in that file to have things correct by design, instead of mixing hardcoding something and using macros for some other options in that file. Otherwise, things will break regularly, as is right now the case with at least `ROOT DEFAULT CAPABILITIES`.

cclerget:

- Fixes #3669 


pkg/runtime/engines/singularity/config/data/singularity.conf
```diff
@@ -206,7 +206,7 @@ autofs bug path = {{$path}}
 always use nv = {{ if eq .AlwaysUseNv true }}yes{{ else }}no{{ end }}
 
 # ROOT DEFAULT CAPABILITIES: [full/file/no]
-# DEFAULT: no
+# DEFAULT: full
```
olifre, line 209: Wouldn't it be safer to use:
```
# DEFAULT: {{ .RootDefaultCapabilities }}
```
here (and also in the other places, where possible)? This would prevent the need to hardcode the default at two different places.

cclerget: No because this template is also used at build/installation time, we parse the currently installed configuration file (in `/usr/local/etc/singularity/singularity.conf` or any configured path), use the values grabbed from the old configuration file and write a new configuration file based on the template, that means if `root default capabilities` was set to `file` in old configuration file by example we would get a `DEFAULT: file` instead of full, so we need to hardcode them. Not a big deal, changes are pretty low in this file

olifre: > we parse the currently installed configuration file (in `/usr/local/etc/singularity/singularity.conf` or any configured path), use the values grabbed from the old configuration file and write a new configuration file based on the template,

Do I understand correctly that this means that when installing singularity on a system on which singularity was already installed and configured before, the generated and installed config file will look different (i.e. similar to the old one) and not be `vanilla` anymore? 

Is there a way to have singularity install itself with the vanilla / "upstream" default configuration independent of what is present on the system? 
Otherwise, this breaks configuration update systems of downstream Linux distributions, since the config file of a "second" installation looks different than the original config file of the first installation before user modifications, and users will not be alerted of changed defaults.

cclerget: Yes but this behaviour can be disabled for package maintainers

To take old values:

```
./mconfig
make -C builddir
...
```

For a vanilla configuration:
```
./mconfig
make old_config= -C builddir
...
```

olifre: Thanks, that explains it very well :smile:.

WestleyK (approved):
Looks good to me :+1:

PR #3682, opened 2019-06-05
Add caching for pulls with oras transport

### LINKED ISSUE #3656: Pull should cache OCI registry (ORAS) images
## Version of Singularity:

Master branch (build: daaf5e7).

### Expected behavior

The pull command should cache the OCI registry (ORAS) images just like library://, shub:// or docker:// images. 

### Actual behavior

Always downloads from OCI registry.

```
$ singularity pull -F example.sif oras://registry.azurecr.io/example_sif:latest
INFO:    Download complete: example.sif
$ singularity pull -F example.sif oras://registry.azurecr.io/example_sif:latest
INFO:    Download complete: example.sif
$ singularity cache list
NAME                     DATE CREATED           SIZE             TYPE

There 0 containers using: 0.00 kB, 0 oci blob file(s) using 0.00 kB of space.
Total space used: 0.00 kB
```

ikaneshiro:

This moves some code into a new package to allow image downloading
through oras to be used by non-cli packages in the future.

- Fixes #3656


internal/pkg/oras/oras.go
```diff
@@ -0,0 +1,183 @@
...
+			return l.Digest.String(), nil
+		}
+	}
+
+	return "", fmt.Errorf("no layer found corresponding to SIF image")
+}
+
+// ImageHash returns the appropriate hash for a provided image file
+//   e.g. sha256:<sha256>
+func ImageHash(filePath string) (result string, err error) {
```
gvallee, line 235: We start to duplicate what was in the client package for library images and that is now in https://github.com/sylabs/scs-library-client/blob/cec4defe222a191539b6a777c33df76bb33637bb/client/util.go#L131
Is it okay? I just do not know...

internal/pkg/oras/oras.go
```diff
@@ -0,0 +1,183 @@
...
+	// ensure container is executable
+	if err := os.Chmod(imagePath, 0755); err != nil {
+		return fmt.Errorf("unable to set image perms: %s", err)
+	}
+
+	return nil
+}
+
+// EnsureSIF checks for a SIF image at filepath and returns an error if it is not, or an error is encountered
+func EnsureSIF(filepath string) error {
```
gvallee: Any reason why this is in `oras.go`? It seems to me that it is independent from ORAS.

internal/app/singularity/pull.go
```diff
@@ -177,71 +170,66 @@ func downloadImageCallback(totalSize int64, r io.Reader, w io.Writer) error {
...
-		return nil, nil
+	} else {
+		sylog.Infof("Using cached image")
 	}
-	pullHandler := oras.WithPullBaseHandler(images.HandlerFunc(handlerFunc))
 
-	_, _, err = oras.Pull(orasctx.Background(), resolver, spec.String(), store, allowedMediaTypes, pullHandler)
+	// Perms are 777 *prior* to umask in order to allow image to be
+	// executed with its leading shebang like a script
+	dstFile, err := os.OpenFile(name, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, 0777)
```
mem: To prevent against a race, if force == false, you should add O_EXCL to the flags here. In that way, if the file didn't exist when you called os.Stat above, but now the file does exist, this will fail, honoring the missing --force.

internal/app/singularity/pull.go
```diff
@@ -177,71 +170,66 @@ func downloadImageCallback(totalSize int64, r io.Reader, w io.Writer) error {
...
+	sum, err := oras.ImageSHA(ref, sysCtx)
+	if err != nil {
+		return fmt.Errorf("failed to get checksum for %s: %s", ref, err)
 	}
-	resolver := docker.NewResolver(docker.ResolverOptions{Credentials: credFn})
 
-	wd, err := os.Getwd()
+	imageName := uri.GetName("oras:" + ref)
+
+	imagePath := cache.OrasImage(sum, imageName)
```
mem: _I_ would call this cacheFilename or cachedImage or some variation of that, to underline that this is the cached file, not the image file that the user is asking for.

internal/app/singularity/pull.go
```diff
@@ -177,71 +170,66 @@ func downloadImageCallback(totalSize int64, r io.Reader, w io.Writer) error {
...
 	}
 
-	store := content.NewFileStore(wd)
-	defer store.Close()
+	if !exists {
+		sylog.Infof("Downloading image with ORAS")
 
-	store.AllowPathTraversalOnWrite = true
-	store.DisableOverwrite = !force
+		if err := oras.DownloadImage(imagePath, ref, ociAuth); err != nil {
```
mem: I suppose this is downloading the image to imagePath?

In that case, I would create a temporary filename _first_, download and verify, and only if that succeeds rename it to the final name in the cache. Also note that in case of error, the invalid file should be deleted.

1. assign temporary filename
2. download to temporary filename
3. verify temporary filename
4. if anything above fails, delete the temporary file, otherwise rename to the final name

If you move this logic to a function, you could do:

```
func downloadToCache(...) error {
    tmpfn := ...
    defer os.Remove(tmpfn)
    if err := download(src, tmpfn, ...) {
        return err // this deletes the temporary file if it was created
    }
    if err := verify(tmpfn, ...) {
        return err // this deletes the temporary
    }
    return os.Rename(tmpfn, finalfn) // if this fails, tmpfn gets deleted
}
```

internal/pkg/client/cache/oras.go
```diff
@@ -0,0 +1,53 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/sylabs/singularity/internal/pkg/oras"
+	"github.com/sylabs/singularity/internal/pkg/sylog"
+)
+
+const (
+	// OrasDir is the directory inside the cache.Dir where oras images are cached
+	OrasDir = "oras"
```
mem, line 18: does this need to be exported?

internal/pkg/client/cache/oras.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+// Oras returns the directory inside the cache.Dir() where oras images are cached
+func Oras() string {
+	return updateCacheSubdir(OrasDir)
+}
+
+// OrasImage creates a directory inside cache.Dir() with the name of the SHA sum of the image
+func OrasImage(sum, name string) string {
+	updateCacheSubdir(filepath.Join(OrasDir, sum))
```
mem: doesn't this return something like filepath.Join(cacheDir, OrasDir, sum)? If yes, maybe you can save that and use it in the next line as `return filepath.Jon(dir, name)`?

mem (changes requested):

internal/pkg/oras/oras.go
ikaneshiro: This is very similar but there are two caveats that creating a function here addresses.
1. the `scs-library-client` gives a sha with a `.` separator when I want a `:` to match the format of OCI layer digests.
2. That client function has a note to change to returning the sif uuid in the future which I want to ensure does not happen here since detecting identical images is reliant on the `sha` sum due to using the OCI layer digest to identify unique images

internal/pkg/oras/oras.go
ikaneshiro: No reason specifically. Not sure if this is something we want in `pkg/image/image.go`. Maybe this could go into `internal/pkg/util`?

internal/pkg/client/cache/oras.go
ikaneshiro: No, I'm just following the convention of the other cache types

internal/pkg/client/cache/oras.go
ikaneshiro: That works ill change that

internal/app/singularity/pull.go
ikaneshiro: The current semantics for the cache is that if a corrupt image exists in the cache `<CacheType>ImageExists()` will indicate that it does not exist in the cache and it is to be overwritten. I think for this PR its best leave this the same between the different cache types, but since we are in the process of revisiting how we work with the cache in issues like #3525 this should be become the new workflow. I agree that it doesnt make sense to allow a corrupt image to be in the cache

internal/pkg/oras/oras.go
ikaneshiro: After thinking a bit more on this, I'm going to move `OrasPush()` into this `oras` package and make `EnsureSIF()` an internal utility function

mem (commented):
LGTM, @gvallee?

ikaneshiro:
@mem @gvallee this is currently not ready to merge as it will fail the e2e test suite. The issue is related to automatically detecting localhost and using http instead of https. Currently pulling from a local registry will fail because it expects a https response but will get http. I will let you guys know when this is ready

ikaneshiro:
@gvallee @mem I have updated this PR to fix the aforementioned issue and have removed the WIP label. For reference, the issue was related to parsing the oci reference with two different packages, one that automatically detected and removed tls for `localhost` and one that did not unless manually specified. This has been resolved by using the package that detects `localhost` throughout the flow. The critical changes since the last review are reflected in `oras.ImageSHA()`.

gvallee (approved):

mem (approved):

PR #3687, opened 2019-06-06
Clear the cache by deleting the individual caches

mem:
Since it's possible for users to nuke their home directory (or in
general other directories in the system that they have access to) by
setting the environment variable SINGULARITY_CACHEDIR to the (in)correct
value, reimplement the way the "cache clean --all" command works by
iterating over the known cache types. This creates a maintainance burden
since the list of known types has to be kept in sync in multiple places,
but this chage is trying to be as small as possible while still being
correct.

### DIFF
--- a/internal/pkg/client/cache/dir.go
+++ b/internal/pkg/client/cache/dir.go
@@ -42,17 +42,6 @@ func Root() string {
 	return root
 }
 
-// Clean : wipes all files in the cache directory, will return a error if one occurs
-func Clean() error {
-	sylog.Debugf("Removing: %v", Root())
-
-	if err := os.RemoveAll(Root()); err != nil {
-		return fmt.Errorf("unable to clean all cache: %s", err)
-	}
-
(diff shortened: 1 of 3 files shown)


internal/pkg/client/cache/dir.go
```diff
@@ -42,17 +42,6 @@ func Root() string {
 	return root
 }
 
-// Clean : wipes all files in the cache directory, will return a error if one occurs
```
gvallee, line 45: By removing this public function, it breaks at least the unit tests. I do not believe that the goal at the moment is to revisit the API, especially the public API, but rather implement a "quick fix". Changing the public API would also impact all the other efforts to correctly image caches and it is unclear to me how that would be beneficial from a timing point of view and coordination of efforts.

gvallee (changes requested):

mem:
This one has been addressed in a different pull request

PR #3695, opened 2019-06-07
Slightly improved 'key pull' usability

### LINKED ISSUE #2893: Pull Public Key by Short ID
## Version of Singularity:

```
3.1.0-309.g795a70e6
```

### Expected behavior

You should be able to pull a key from the output of the verify command:

```
$ singularity verify alpine_latest.sif 
Verifying image: alpine_latest.sif
	Sylabs Admin <support@sylabs.io>, KeyID EDECE4F3F38D871E

$ singularity key pull EDECE4F3F38D871E
WARNING: No default remote in use, falling back to: https://keys.sylabs.io
0 key(s) fetched and stored in local cache /home/westleyk/.singularity/sypgp/pgp-public
```

### Actual behavior

```
$ singularity verify alpine_latest.sif 
Verifying image: alpine_latest.sif
	Sylabs Admin <support@sylabs.io>, KeyID EDECE4F3F38D871E

$ singularity key pull EDECE4F3F38D871E
WARNING: No default remote in use, falling back to: https://keys.sylabs.io
ERROR:   pull failed: unexpected fingerprint length of 8 (expected 20)
```

```
$ singularity verify alpine_latest.sif 

$ singularity key pull <key_id_from_that_container>
```

WestleyK:

This PR slightly improves the 'key pull' function.

#### Example

```bash
# in this PR
$ singularity key pull CE4F3F38D871E
# will pull the Sylabs Admin key

# before this PR it will error out and say its to long, or short
$ singularity key pull CE4F3F38D871E
ERROR:   pull failed: unable to pull key from server: failed to decode fingerprint: encoding/hex: odd length hex string

# or with one less char
$ singularity key pull E4F3F38D871E
ERROR:   pull failed: unable to pull key from server: not a valid key lenth: only accepts 8, or 40 chars

```

- Fixes #2893

<br>

### DIFF
--- a/pkg/sypgp/sypgp.go
+++ b/pkg/sypgp/sypgp.go
@@ -618,10 +618,13 @@ func FetchPubkey(fingerprint, keyserverURI, authToken string, noPrompt bool) (op
 		return nil, fmt.Errorf("not a valid key lenth: %s: must be over 8 chars", fingerprint)
 	}
 
-	// Harvest the last 8 chars of the fingerprint string
-	sylog.Debugf("Getting the last 8 chars from: %s", fingerprint)
-	fingerprint = fingerprint[len(fingerprint)-8:]
-	sylog.Debugf("Results: %s", fingerprint)
+	// Only chop the fingerprint if its less then 40 chars
+	if len(fingerprint) < 40 {
+		// Harvest the last 8 chars of the fingerprint string
+		sylog.Debugf("Getting the last 8 chars from: %s", fingerprint)
+		fingerprint = fingerprint[len(fingerprint)-8:]
+		sylog.Debugf("Results: %s", fingerprint)
+	}
 
 	// Decode fingerprint and ensure proper length.
 	var fp []byte
--- a/pkg/sypgp/sypgp.go
+++ b/pkg/sypgp/sypgp.go
@@ -614,17 +614,20 @@ func SearchPubkey(search, keyserverURI, authToken string) error {
 
 // FetchPubkey pulls a public key from the Key Service.
 func FetchPubkey(fingerprint, keyserverURI, authToken string, noPrompt bool) (openpgp.EntityList, error) {
+	if len(fingerprint) < 8 {
+		return nil, fmt.Errorf("not a valid key lenth: %s: must be over 8 chars", fingerprint)
+	}
+
+	// Harvest the last 8 chars of the fingerprint string
+	sylog.Debugf("Getting the last 8 chars from: %s", fingerprint)
+	fingerprint = fingerprint[len(fingerprint)-8:]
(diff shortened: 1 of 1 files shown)


pkg/sypgp/sypgp.go
```diff
@@ -614,6 +614,17 @@ func SearchPubkey(search, keyserverURI, authToken string) error {
...
 func FetchPubkey(fingerprint, keyserverURI, authToken string, noPrompt bool) (openpgp.EntityList, error) {
+	if len(fingerprint) < 8 {
+		return nil, fmt.Errorf("not a valid key lenth: %s: must be over 8 chars", fingerprint)
+	}
+
+	// Only chop the fingerprint if its less then 40 chars
+	if len(fingerprint) < 40 {
+		// Harvest the last 8 chars of the fingerprint string
+		sylog.Debugf("Getting the last 8 chars from: %s", fingerprint)
+		fingerprint = fingerprint[len(fingerprint)-8:]
```
mem, line 625: not sure I agree with this.

The user is providing a specific input, and in an effort to be "user friendly" we would be _discarding_ part of the input.

The problem with this is that we would be making the search less specific, potentially returning results the user was NOT looking for.

For clarity, the fingerprint a 160-bit hash of part of the public key; the long key ID is the lowest-order 64 bits of this hash; the short key ID is the lowest-order 32 bits of this hash. In hexadecimal this means the fingerprint is 40 digits, the short key ID is 8 digits and the long key ID is 16 digits.

Consider a "highly hypothetical" case of two keys sharing the same _short key ID_. If the user intended to type a fingerprint, but misses a digit, so instead of entering 40 digits enters 39, this code would remove 31 of those digits and keep only 8. Instead of getting an error, the user would get zero, one or two results, _depending on the keyserver it's talking to_. In a normal situation the user would still get the intended key, but this is highly dependent on external factors. What if the user gets the _other_ key though? Meaning, with the original query the user would get a "not found message", but because of the typo they get something.

I would argue this is _less_ desirable than the error.

My take is that this code should be strict about checking for 40, 16 or 8 digits. I would go a step further and refuse to use short key IDs, as collisions in short key IDs are in fact not "highly hypothetical" (http://www.asheesh.org/note/debian/short-key-ids-are-bad-news.html).

mem (changes requested):

pkg/sypgp/sypgp.go
GodloveD: So it seems like the issue is really with the SCS API right?  Maybe this should be handled the same way git handles hashes on the KeyStore side.  It will accept a fingerprint of any length and then will either return the key or barf because there is a collision and the user must provide additional chars.

WestleyK:
Closing. This should be handled differently.

PR #3696, opened 2019-06-07
Fix parent process signal propagation

cclerget:

This PR add signal propagation support to relay signal from master process to container process and when `sinit` shim process is running to also relay signal to container process.
Signals are propagated only if Singularity processes are not in the same process group as the one controlling the terminal or if Singularity doesn't have at least one of standard file descriptor pointing to a terminal. Which means that Singularity master process will propagate signal if run like :

```
singularity exec image sleep 100 &
setsid singularity exec image sleep 100
```

### DIFF
--- a/internal/app/starter/master_linux.go
+++ b/internal/app/starter/master_linux.go
@@ -24,8 +24,14 @@ import (
 func Master(rpcSocket, masterSocket int, isInstance bool, containerPid int, engine *engines.Engine) {
 	var fatal error
 	var status syscall.WaitStatus
-
 	fatalChan := make(chan error, 1)
+
+	// we could receive signal from child with CreateContainer call so we
+	// set the signal handler earlier to queue signals until MonitorContainer
+	// is called to handle them
+	signals := make(chan os.Signal, 1)
+	signal.Notify(signals)
+
 	ppid := os.Getppid()
 
 	go func() {
@@ -109,11 +115,6 @@ func Master(rpcSocket, masterSocket int, isInstance bool, containerPid int, engi
 
 	go func() {
 		var err error
-
-		// catch all signals
-		signals := make(chan os.Signal, 1)
-		signal.Notify(signals)
-
 		status, err = engine.MonitorContainer(containerPid, signals)
 		fatalChan <- err
 	}()
--- a/internal/pkg/runtime/engines/singularity/monitor.go
+++ b/internal/pkg/runtime/engines/singularity/monitor.go
@@ -25,6 +25,12 @@ func (engine *EngineOperations) MonitorContainer(pid int, signals chan os.Signal
 				continue
 			}
 			return status, nil
+		default:
+			if engine.EngineConfig.GetSignalPropagation() {
+				if err := syscall.Kill(pid, s.(syscall.Signal)); err != nil {
+					return status, fmt.Errorf("interrupted by signal %s", s.String())
+				}
+			}
 		}
 	}
 }
--- a/internal/pkg/runtime/engines/singularity/prepare_linux.go
+++ b/internal/pkg/runtime/engines/singularity/prepare_linux.go
@@ -17,6 +17,7 @@ import (
 	"syscall"
 
 	"github.com/sylabs/singularity/pkg/util/fs/proc"
+	"golang.org/x/sys/unix"
 
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/sylabs/singularity/internal/pkg/buildcfg"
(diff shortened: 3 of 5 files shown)


internal/app/starter/master_linux.go
```diff
@@ -24,8 +24,12 @@ import (
 func Master(rpcSocket, masterSocket int, isInstance bool, containerPid int, engine *engines.Engine) {
 	var fatal error
 	var status syscall.WaitStatus
-
 	fatalChan := make(chan error, 1)
+
+	// catch all signals
```
mem: you really need to add a *comment in the code* as to why this needs to be here instead of where it was. The next person going over this code is going to have the impulse to move it back. Even if they dig thru the commit history, they won't be any wiser, because there's nothing hinting at why this change is required.

pkg/runtime/engines/singularity/config/config.go
```diff
@@ -485,3 +486,15 @@ func (e *EngineConfig) GetDeleteImage() bool {
 func (e *EngineConfig) SetDeleteImage(delete bool) {
 	e.JSON.DeleteImage = delete
 }
+
+// SetSignalPropagation sets if parent processes (master and sinit)
```
mem: I think the wording of the documentation for both functions could be improved, but I'm having a hard time finding a suggestion.

Looking at the code that resends the signal (L#30 in monitor.go above), it uses the PID passed to Master, and chasing that arround, that seems to correspond to the PID of the master process.

So signals are propagated from the monitor process to the master process? Am I reading the code wrong?

internal/pkg/runtime/engines/singularity/monitor.go
```diff
@@ -25,6 +25,12 @@ func (engine *EngineOperations) MonitorContainer(pid int, signals chan os.Signal
 				continue
 			}
 			return status, nil
+		default:
+			if engine.EngineConfig.GetSignalPropagation() {
+				if err := syscall.Kill(pid, s.(syscall.Signal)); err != nil {
```
mem, line 30: this pid here is the "container PID", but the internal documentation does not describe what that is. Chasing code around this seems to be the PID of the master process.

Please see below.

WestleyK (approved):
Looks good to my knowledge :+1: Maybe @bauerm97, @ikaneshiro or @mem can have a look.

internal/pkg/runtime/engines/singularity/prepare_linux.go
```diff
@@ -720,6 +721,20 @@ func (e *EngineOperations) PrepareConfig(starterConfig *starter.Config) error {
 		starterConfig.SetCapabilities(capabilities.Ambient, e.EngineConfig.OciConfig.Process.Capabilities.Ambient)
 	}
 
+	// check if there is a controlling terminal
```
mem:
```suggestion
	// obtain the process group ID of the associated controlling
	// terminal (if there's one).
```

This little bit here could be its own function, just for clarity's sake.

internal/pkg/runtime/engines/singularity/prepare_linux.go
```diff
@@ -720,6 +721,20 @@ func (e *EngineOperations) PrepareConfig(starterConfig *starter.Config) error {
...
+	if pgrp == 0 {
+		pgrp, _ = unix.IoctlGetInt(1, unix.TIOCGPGRP)
+		if pgrp == 0 {
+			pgrp, _ = unix.IoctlGetInt(2, unix.TIOCGPGRP)
+		}
+	}
+	// cases we need to propagate signals to container process:
+	// - when pgrp == 0 because container won't run in a terminal
+	// - when process group is different from the process group controlling terminal
+	if pgrp == 0 || (pgrp > 0 && pgrp != syscall.Getpgrp()) {
```
mem, line 777: reading thru [ioctl_tty](http://man7.org/linux/man-pages/man2/ioctl_tty.2.html) and [tcgetpgrp](http://man7.org/linux/man-pages/man3/tcgetpgrp.3.html), the bit that stands out is:

>  and some value larger than 1 that is not presently a process group ID otherwise

is that the behavior of the system call or the GNU libc wrapper?

From those docs there seem to be three cases:

- If the file descriptor refers to the controlling terminal, and there's a foreground process associated with that terminal, the process group ID of that process is returned
- If the file descriptor refers to the controlling terminal, and there is no foreground process associated with that terminal, _some_ process group ID (associated with that terminal?) is returned
- if the file descriptor does not refer to the controlling terminal, -1 is returned

I can't find anything regarding 0.

If the behavior of the system call is different from the GNU libc wrapper, this should be documented so that other people don't have to go around chasing documentation to figure out if this is correct or not.

internal/pkg/runtime/engines/singularity/process_linux.go
```diff
@@ -321,6 +321,11 @@ func (engine *EngineOperations) StartProcess(masterConn net.Conn) error {
 						sylog.Debugf("No child process, exiting ...")
 						os.Exit(128 + int(signal))
 					}
+				} else if engine.EngineConfig.GetSignalPropagation() {
+					if err := syscall.Kill(cmd.Process.Pid, signal); err == syscall.ESRCH {
```
mem, line 331: technically, it's possible that there's an error other than syscall.ESRCH. There should be a test for a non-nil error.

mem (changes requested):

pkg/runtime/engines/singularity/config/config.go
cclerget: monitor is called from master process, I think you missed this line https://github.com/sylabs/singularity/blob/f6ad552faa6e8a8427bb687d6ad100dfd8877aca/internal/app/starter/master_linux.go#L116

internal/pkg/runtime/engines/singularity/process_linux.go
cclerget: technically yes but they are unlikely to happen in this context, `EINVAL` means Go runtime received an invalid signal number. `EPERM` means PID 1 process can't send signal to its direct child which is not possible

internal/pkg/runtime/engines/singularity/prepare_linux.go
cclerget: This is syscall behaviour, in kernel source code 0 can be returned based on this code https://elixir.bootlin.com/linux/latest/source/kernel/pid.c#L416, I checked this is like that since a very long time. Now if there is any error with above code, the ioctl referenced value won't be affected and will follow the Go idiomatic way of a function returning an error where we should get zero default value

GodloveD:
ping @mem

pkg/runtime/engines/singularity/config/config.go
mem: thanks

mem (approved):
thanks!

PR #3716, opened 2019-06-11
Improve integration into SUSE SLE

e4t:

Add support for singularity containers on SLE to 'zypper' bootstrap:
- Add support for header variables spanning multiple lines
- Add support for 'indexed' header variables
- Add "product, user, regcode, productpgp, registerurl, modules, otherurl (indexed) 

**This fixes or addresses the following issues:**
- Fix rpmDB handling for later (open)SUSE versions
- Handle install script failures more gracefully.

- None


internal/pkg/build/sources/conveyorPacker_zypper.go
```diff
@@ -74,6 +79,124 @@ func (cp *ZypperConveyorPacker) Get(b *types.Bundle) (err error) {
 	// add aaa_base to start of include list by default
 	include = `aaa_base ` + include
 
+	// get mirrorURL, OSVerison, and Includes components to definition
+	osversion, osversion_ok := cp.b.Recipe.Header["osversion"]
```
WestleyK: >don't use underscores in Go names; var `osversion_ok` should be `osversionOk` (golint)

 From golint.

internal/pkg/build/sources/conveyorPacker_zypper.go
```diff
@@ -74,6 +79,124 @@ func (cp *ZypperConveyorPacker) Get(b *types.Bundle) (err error) {
...
+		if len(array) > 1 {
+			osminor = "." + array[1]
+		}
+		if iosmajor > 12 && !mirrorurl_ok {
+			return fmt.Errorf("For SLE version > 12 'MirrorURL' must be defined and point to the installer\n")
+		}
+		osservicepack := ""
+		tmp, err := strconv.Atoi(array[1])
+		if err != nil {
+			return fmt.Errorf("Cannot convert minor version string to integer: %v", err)
```
WestleyK: To be consistent (and its good practice), you should't use caps in return errors, this should be:

```suggestion
			return fmt.Errorf("cannot convert minor version string to integer: %v", err)
```

WestleyK (dismissed):
Left a couple suggestion comments, thanks for contributing!

ikaneshiro:
Hey @e4t, sorry for the delay on this from me. This is looking good just have a couple questions.

Can you run `make check` on your code? It looks like CI is giving some formatting/style errors and that will help track them down.

Do you need to store the pgp key in the header like this or could that parameter reference a file to prevent the need to add line continuation for header parameters?

Will this add any restrictions on the version of zypper used?

e4t:
> Hey @e4t, sorry for the delay on this from me. This is looking good just have a couple questions.
> 
> Can you run `make check` on your code? It looks like CI is giving some formatting/style errors and that will help track them down.

I've got the fixes ready, I'm sorry that I haven't dalt with this, yet - ISC19 got in the way.
I will submit them ASAP.

> Do you need to store the pgp key in the header like this or could that parameter reference a file to prevent the need to add line continuation for header parameters?

The GPG key is needed on SLE-12 only. On SLE-15 you need to provide a URL to the installation medium which will provide the needed data. The reason why I've added the line continuation for header parameters was that I wanted to avoid multiple files: it seems to be much more convenient if container descriptions are self contained.
> 
> Will this add any restrictions on the version of zypper used?

zypper will require the key in a separate file anyway. This is generated from the data in the header parameter by the zypper-specific code.
The reason for adding this to the description was pure convenience while attempting to maintain readability.

internal/pkg/build/sources/conveyorPacker_zypper.go
```diff
@@ -74,6 +79,124 @@ func (cp *ZypperConveyorPacker) Get(b *types.Bundle) (err error) {
...
+			return fmt.Errorf("malformed Product setting")
+		}
+		if slepgpOk {
+			tmpfile, err := ioutil.TempFile("/tmp", "singularity-pgp")
+			if err != nil {
+				return fmt.Errorf("cannot create pgp-file: %v", err)
+			}
+			pgpfile = tmpfile.Name()
+
+			if _, err = tmpfile.WriteString(slepgp); err != nil {
```
ikaneshiro: I ran into this error when trying to import the gpg key from the example definition:
```
ian@linux-of0y:~/singularity> sudo singularity build test.sif examples/sle/Singularity 
[sudo] password for root: 
INFO:    Starting build...
error: /tmp/singularity-pgp233218452: key 1 not an armored public key.
FATAL:   While performing build: conveyor failed to get: while importing pgp keys: exit status 1
```

Looks like this was caused by a missing newline at the end og `slepgp`. 

There appears to be two ways to fix this:
1. always append a newline here
2. add `\n` to the last line of the key in the definition and change the parser to always replace any trailing `\n` characters with a newline character

ikaneshiro (changes requested):
@e4t Sorry this took so long for me to review.

Tested this a bit and only ran into one issue. I'm not a SUSE customer so I don't think I have the means test the `SUSEconnect` portion of this myself, but I assume you have and it's working.

internal/pkg/build/sources/conveyorPacker_zypper.go
e4t: Indeed, I had not tested this part any more after I had done changes. As soon as a MirrorImage is provided (which is the case for SLE-15 upwards), this key is no longer required. While retesting this I came across some other issues as well. All have been addressed in my latest push.

ikaneshiro (approved):
lgtm!

Issue #3717, opened 2019-06-12
Inconsistent -T flag for cache commands


mem:
## Version of Singularity:

61b2ea101308bb9d51d36f41862a5ed32bb3458a

### Expected behavior

The -T flags (as in `singularity cache list -T ...`) behaves consistently across cache commands (`singularity cache list ...`, `singularity cache clean ...`)

### Actual behavior

The default value and the accepted values for -T are different across commands.

### Steps to reproduce behavior

Look at help text for `singularity cache clean` and `singularity cache list`

PR #3720, opened 2019-06-12
Simplify code cache listing code

### LINKED ISSUE #3718: Inconsistent behavior of singularity cache list --summary
## Version of Singularity:

61b2ea101308bb9d51d36f41862a5ed32bb3458a

### Expected behavior

The behavior of the --summary flag is consistent

### Actual behavior

Sometimes adding --summary to `singularity cache list` causes the cache entries to not be listed, some times they are.

Use `singularity cache list -T ... --summary` with different values for the -T flag (all, blob, oci, etc)

mem:
Remove some of the code repetition by moving all the common code to a
single location.

The summary for the listing is printed will all the listings, and the
--summary flag consistently means "only print the summary", without
printing the invidual entries.

Remove the "blobSum" type, as it's used to request a summary for the
blobs, which is the same thing --summary does when passing the -T blob
flag.

Make the default -T value for list and clean the same.

Fixes: #3718, #3717

### DIFF
--- a/cmd/internal/cli/cache_clean_linux.go
+++ b/cmd/internal/cli/cache_clean_linux.go
@@ -17,48 +17,45 @@ import (
 )
 
 var (
-	cleanAll        bool
+	cacheCleanForce bool
 	cacheCleanTypes []string
-	cacheName       []string
+	cacheCleanNames []string
 )
-
-// -a|--all
-var cacheCleanAllFlag = cmdline.Flag{
-	ID:           "cacheCleanAllFlag",
-	Value:        &cleanAll,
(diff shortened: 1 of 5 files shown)


mem:
[output before this change](https://github.com/sylabs/singularity/files/3290850/cache-list-old.txt)
[output after this change](https://github.com/sylabs/singularity/files/3290851/cache-list-new.txt)

WestleyK (commented):
After some discussing with @mem, we will go with this workflow, and see what the community thinks about it. Thanks @mem! :+1:

WestleyK (approved):

PR #3724, opened 2019-06-13
Clean environment by setting variables to empty string rather than NULL

### LINKED ISSUE #3703: Segmentation fault when executing a basic command
## Version of Singularity:

I traced the bug back to commit 95381813bff402469409dc52acab53a243e986e0 where it was first introduced. It exists from that point forward up to the current master fae33077e3270a624e673e0940341dc8b4b6c20c.

### Expected behavior

Executing commands should work.

### Actual behavior

Executing commands results in a segmentation fault.

```
[jacob@gideon tmp]$ singularity exec library://centos:7 echo hi
Segmentation fault
```

phpHavok:

This PR cleans the environment in starter by setting existing/unwanted environment variables to the empty string rather than NULL. This fixes a weird bug based on how the Go runtime discovers `auxv` (See https://github.com/golang/go/blob/82cf8bca9cf20297bc0edf481cc530c9b3f4bf1e/src/runtime/os_linux.go#L192-L195).

- Fixes #3703 

### DIFF
--- a/cmd/starter/c/starter.c
+++ b/cmd/starter/c/starter.c
@@ -814,21 +814,20 @@ void do_exit(int sig) {
 static void cleanenv(void) {
     extern char **environ;
     char **e;
-    char *p = NULL;
 
     if ( environ == NULL || *environ == NULL ) {
         fatalf("no environment variables set\n");
     }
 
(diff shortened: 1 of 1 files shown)


cclerget (approved):
Tested on debian testing and worked as expected :) ! Thank you for fixing this annoying bug !

PR #3729, opened 2019-06-13
Encrypted File System inside SIF

### LINKED ISSUE #3285: Encrypting SIF images
## Version of Singularity:

3.3.0
### Expected behavior

Currently SIFs are not encrypted. Idea here is the encrypt the squashfs portion of SIF while building and decrypting that portion via a loopback device at runtime.

### Actual behavior
Encryption is not currently supported.

None

### LINKED ISSUE #3517: encrypt the squashfs partition of a SIF image
In order to mount and run encrypted SIF images, one must first be able to simply encrypt the file system.  This is the first little step toward that greater goal.  

This doesn't even really need to be integrated into the Singularity CLI or anything yet.  Just carried out as a POC.

### LINKED ISSUE #3518: mount an encrypted SIF container to the loop device
This is one step in the overall process of encrypting and running encrypted containers.  Don't know if it will be a simple thing to do or if it will require multiple sub-issues.  Once the POC is carried out, will need to integrate into Singularity runtime code as well.

### LINKED ISSUE #3519: Read the contents of an encrypted and mounted SIF image
This is one more POC step to implementing encrypted containers.  After the container is mounted it must be demonstrated that the contents can be accessed.

### LINKED ISSUE #3605: Add command line support for encrypted filesystem
## Version of Singularity:

3.5

### Expected behavior

Add singularity command support to specify if the filesystem needs to be encrypted.
Subsequently read the password during encryption and decryption phases.

### Actual behavior

Not implemented

NA

### LINKED ISSUE #3662: Add a command line argument to encrypt file system
### Expected behavior

sudo singularity build -e test.sif test.recipe

Enter password to encrypt filesystem: 

### Actual behavior

Not supported yet

### LINKED ISSUE #3664: Root privilege during cleanup
### Expected behavior

During cleanup (exiting the container), "cryptsetup luksClose <dev>" needs to be executed so that we don't end up with stale encrypted devices and loopback devices. CleanupContainer() is run for the regular runtime to perform cleanup operations but it doesn't have root privileges to perform the above operation. 

### Actual behavior

NA

NA

schebro:
Description of the Pull Request (PR):

Implemented review feedback

Locking /dev/mapper before creating crypt devices to avoid races

Deleted encryptfs.go, which is not necessary

Generate random number for crypt device

Ext3->Squashfs for encrypted FS

This fixes or addresses the following GitHub issues:

Fixes #3605
Fixes #3662
Fixes #3664
Fixes #3285
Fixes #3517
Fixes #3518
Fixes #3519
NOTE: CI tests are still failing. I am resolving the failures now. Pushing the code to get feedback on the latest changes while I resolve other issues.

### DIFF
--- a/cmd/internal/cli/build_linux.go
+++ b/cmd/internal/cli/build_linux.go
@@ -157,6 +157,7 @@ func run(cmd *cobra.Command, args []string) {
 					LibraryAuthToken: authToken,
 					DockerAuthConfig: authConf,
 					Fakeroot:         fakeroot,
+					Encrypted:        encrypt,
 				},
 			})
 		if err != nil {
--- a/internal/pkg/build/assemblers/assembler_sif.go
+++ b/internal/pkg/build/assemblers/assembler_sif.go
@@ -25,13 +25,14 @@ import (
 	"github.com/sylabs/singularity/internal/pkg/sylog"
 	"github.com/sylabs/singularity/pkg/build/types"
 	singularityConfig "github.com/sylabs/singularity/pkg/runtime/engines/singularity/config"
+	"github.com/sylabs/singularity/pkg/util/crypt"
 )
 
 // SIFAssembler doesnt store anything
 type SIFAssembler struct {
 }
 
-func createSIF(path string, definition, ociConf []byte, squashfile string) (err error) {
+func createSIF(path string, definition, ociConf []byte, squashfile string, encrypted bool) (err error) {
 	// general info for the new SIF file creation
 	cinfo := sif.CreateInfo{
 		Pathname:   path,
@@ -90,7 +91,13 @@ func createSIF(path string, definition, ociConf []byte, squashfile string) (err 
 	parinput.Fp = fp
 	parinput.Size = fi.Size()
 
-	err = parinput.SetPartExtra(sif.FsSquash, sif.PartPrimSys, sif.GetSIFArch(runtime.GOARCH))
+	sifType := sif.FsSquash
+
+	if encrypted {
+		sifType = sif.FsEncryptedSquashfs
+	}
+
+	err = parinput.SetPartExtra(sifType, sif.PartPrimSys, sif.GetSIFArch(runtime.GOARCH))
 	if err != nil {
 		return
 	}
@@ -139,19 +146,17 @@ func getMksquashfsPath() (string, error) {
 func (a *SIFAssembler) Assemble(b *types.Bundle, path string) (err error) {
 	sylog.Infof("Creating SIF file...")
 
+	var fsPath string
+
 	mksquashfs, err := getMksquashfsPath()
 	if err != nil {
 		return fmt.Errorf("While searching for mksquashfs: %v", err)
 	}
-
 	f, err := ioutil.TempFile(b.Path, "squashfs-")
-	squashfsPath := f.Name() + ".img"
+	fsPath = f.Name()
 	f.Close()
-	os.Remove(f.Name())
-	os.Remove(squashfsPath)
-	defer os.Remove(squashfsPath)
-
-	args := []string{b.Rootfs(), squashfsPath, "-noappend"}
+	defer os.Remove(fsPath)
+	args := []string{b.Rootfs(), fsPath, "-noappend"}
 
 	// build squashfs with all-root flag when building as a user
 	if syscall.Getuid() != 0 {
@@ -177,7 +182,29 @@ func (a *SIFAssembler) Assemble(b *types.Bundle, path string) (err error) {
 		return fmt.Errorf("While running mksquashfs: %v: %s", err, strings.Replace(string(errOut), "\n", " ", -1))
 	}
 
-	err = createSIF(path, b.Recipe.Raw, b.JSONObjects["oci-config"], squashfsPath)
+	if b.Opts.Encrypted {
+		// A dm-crypt device needs to be created with squashfs
+		cryptDev := &crypt.Device{}
+
+		key, err := cryptDev.ReadKeyFromStdin(true)
+		if err != nil {
+			return fmt.Errorf("unable to read key from stdin")
+		}
+		loopPath, cryptName, err := cryptDev.FormatCryptDevice(fsPath, key)
+		if err != nil {
+			return fmt.Errorf("unable to format crypt device: %s", cryptName)
+		}
+
+		defer os.Remove(loopPath)
+
+		err = cryptDev.CloseCryptDevice(cryptName)
+		if err != nil {
+			return fmt.Errorf("unable to close crypt device: %s", cryptName)
+		}
+		fsPath = loopPath
+	}
+
+	err = createSIF(path, b.Recipe.Raw, b.JSONObjects["oci-config"], fsPath, b.Opts.Encrypted)
 	if err != nil {
 		return fmt.Errorf("While creating SIF: %v", err)
 	}
--- a/internal/pkg/build/assemblers/assembler_sif.go
+++ b/internal/pkg/build/assemblers/assembler_sif.go
@@ -186,6 +186,10 @@ func (a *SIFAssembler) Assemble(b *types.Bundle, path string) (err error) {
 		// A dm-crypt device needs to be created with squashfs
 		cryptDev := &crypt.Device{}
 
+		// TODO (schebro): Fix #3876
+		// Detach the following code from the squashfs creation. SIF can be
+		// created first and encrypted after. This gives the flexibility to
+		// encrypt an existing SIF
 		key, err := cryptDev.ReadKeyFromStdin(true)
 		if err != nil {
 			return fmt.Errorf("unable to read key from stdin")
--- a/internal/pkg/runtime/engines/singularity/cleanup_linux.go
+++ b/internal/pkg/runtime/engines/singularity/cleanup_linux.go
@@ -6,12 +6,17 @@
 package singularity
 
 import (
+	"fmt"
 	"os"
+	"path/filepath"
+	"runtime"
 	"syscall"
 
+	"github.com/sylabs/singularity/internal/pkg/buildcfg"
 	"github.com/sylabs/singularity/internal/pkg/instance"
 	"github.com/sylabs/singularity/internal/pkg/sylog"
 	"github.com/sylabs/singularity/internal/pkg/util/priv"
+	"github.com/sylabs/singularity/pkg/util/crypt"
 )
 
 /*
@@ -21,7 +26,6 @@ import (
 
 // CleanupContainer cleans up the container
 func (engine *EngineOperations) CleanupContainer(fatal error, status syscall.WaitStatus) error {
-	sylog.Debugf("Cleanup container")
 
 	if engine.EngineConfig.GetDeleteImage() {
 		image := engine.EngineConfig.GetImage()
@@ -56,6 +60,45 @@ func (engine *EngineOperations) CleanupContainer(fatal error, status syscall.Wai
 			return err
 		}
 		return file.Delete()
+	}
+
+	if engine.EngineConfig.CryptDev != "" {
+		cleanupCrypt(engine.EngineConfig.CryptDev)
+	}
+
+	return nil
+}
+
+func cleanupCrypt(path string) error {
+
+	// Elevate the privilege to unmount and delete the crypt device
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	uid := os.Getuid()
+	err := syscall.Setresuid(uid, 0, uid)
+	if err != nil {
+		return fmt.Errorf("error while setting root uid")
+	}
+
+	defer syscall.Setresuid(uid, uid, 0)
+
+	err = syscall.Unmount(filepath.Join(buildcfg.SESSIONDIR, "final"), syscall.MNT_DETACH)
+	if err != nil {
+		return fmt.Errorf("failed while unmounting final session directory: %s", err)
+	}
+
+	err = syscall.Unmount(filepath.Join(buildcfg.SESSIONDIR, "rootfs"), syscall.MNT_DETACH)
+	if err != nil {
+		return fmt.Errorf("error while unmounting rootfs session directory: %s", err)
+	}
+
+	devName := filepath.Base(path)
+
+	cryptDev := &crypt.Device{}
+	err = cryptDev.CloseCryptDevice(devName)
+	if err != nil {
+		return fmt.Errorf("unable to delete crypt device: %s", devName)
 	}
 
 	return nil
--- a/internal/pkg/runtime/engines/singularity/cleanup_linux.go
+++ b/internal/pkg/runtime/engines/singularity/cleanup_linux.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, Sylabs Inc. All rights reserved.
+// Copyright (c) 2018-2019, Sylabs Inc. All rights reserved.
 // This software is licensed under a 3-clause BSD license. Please consult the
 // LICENSE.md file distributed with the sources of this project regarding your
 // rights to use or distribute this software.
--- a/internal/pkg/runtime/engines/singularity/container_linux.go
+++ b/internal/pkg/runtime/engines/singularity/container_linux.go
@@ -599,10 +599,29 @@ func (c *container) mountImage(mnt *mount.Point) error {
 	}
 
 	path := fmt.Sprintf("/dev/loop%d", number)
-	sylog.Debugf("Mounting loop device %s to %s\n", path, mnt.Destination)
-	_, err = c.rpcOps.Mount(path, mnt.Destination, mnt.Type, flags, optsString)
+
+	sylog.Debugf("Mounting loop device %s to %s of type %s\n", path, mnt.Destination, mnt.Type)
+
+	mountType := mnt.Type
+
+	if mountType == "encryptfs" {
+		cryptDev, err := c.rpcOps.Decrypt(offset, path)
+
+		if err != nil {
+			return fmt.Errorf("Unable to decrypt the file system: %s", err)
+		}
+
+		path = cryptDev
+
+		// Save this device to cleanup later
+		c.engine.EngineConfig.CryptDev = cryptDev
+
+		// Currently we only support encrypted squashfs file system
+		mountType = "squashfs"
+	}
+	_, err = c.rpcOps.Mount(path, mnt.Destination, mountType, flags, optsString)
 	if err != nil {
-		return fmt.Errorf("failed to mount %s filesystem: %s", mnt.Type, err)
+		return fmt.Errorf("failed to mount %s filesystem: %s", mountType, err)
 	}
 
 	return nil
@@ -666,11 +685,15 @@ func (c *container) addRootfsMount(system *mount.System) error {
 	offset := imageObject.Partitions[0].Offset
 	size := imageObject.Partitions[0].Size
 
+	sylog.Debugf("Image type is %v", imageObject.Partitions[0].Type)
+
 	switch imageObject.Partitions[0].Type {
 	case image.SQUASHFS:
 		mountType = "squashfs"
 	case image.EXT3:
 		mountType = "ext3"
+	case image.ENCRYPTSQUASHFS:
+		mountType = "encryptfs"
 	case image.SANDBOX:
 		sylog.Debugf("Mounting directory rootfs: %v\n", rootfs)
 		flags |= syscall.MS_BIND
--- a/internal/pkg/runtime/engines/singularity/rpc/client/client.go
+++ b/internal/pkg/runtime/engines/singularity/rpc/client/client.go
@@ -8,6 +8,7 @@ package client
 import (
 	"net/rpc"
 	"os"
+	"path/filepath"
 
 	args "github.com/sylabs/singularity/internal/pkg/runtime/engines/singularity/rpc"
 	"github.com/sylabs/singularity/pkg/util/loop"
@@ -19,7 +20,7 @@ type RPC struct {
 	Name   string
 }
 
-// Mount calls tme mount RPC using the supplied arguments.
+// Mount calls the mount RPC using the supplied arguments.
 func (t *RPC) Mount(source string, target string, filesystem string, flags uintptr, data string) (int, error) {
 	arguments := &args.MountArgs{
 		Source:     source,
@@ -30,6 +31,20 @@ func (t *RPC) Mount(source string, target string, filesystem string, flags uintp
 	}
 	var reply int
 	err := t.Client.Call(t.Name+".Mount", arguments, &reply)
+	return reply, err
+}
+
+// Decrypt calls the DeCrypt RPC using the supplied arguments.
+func (t *RPC) Decrypt(offset uint64, path string) (string, error) {
+	loopdev := filepath.Base(path)
+	arguments := &args.CryptArgs{
+		Offset:  offset,
+		Loopdev: loopdev,
+	}
+
+	var reply string
+	err := t.Client.Call(t.Name+".Decrypt", arguments, &reply)
+
 	return reply, err
 }
 
--- a/internal/pkg/runtime/engines/singularity/rpc/server/server_linux.go
+++ b/internal/pkg/runtime/engines/singularity/rpc/server/server_linux.go
@@ -17,6 +17,7 @@ import (
 	"github.com/sylabs/singularity/internal/pkg/sylog"
 	"github.com/sylabs/singularity/internal/pkg/util/mainthread"
 	"github.com/sylabs/singularity/internal/pkg/util/user"
+	"github.com/sylabs/singularity/pkg/util/crypt"
 	"github.com/sylabs/singularity/pkg/util/loop"
 )
 
@@ -30,6 +31,21 @@ func (t *Methods) Mount(arguments *args.MountArgs, reply *int) (err error) {
 	mainthread.Execute(func() {
 		err = syscall.Mount(arguments.Source, arguments.Target, arguments.Filesystem, arguments.Mountflags, arguments.Data)
 	})
+	return err
+}
+
+// DeCrypt decrypts the loop device
+func (t *Methods) Decrypt(arguments *args.CryptArgs, reply *string) (err error) {
+	cryptDev := &crypt.Device{}
+
+	key, err := cryptDev.ReadKeyFromStdin(false)
+	if err != nil {
+		return fmt.Errorf("Unable to read key from stdin")
+	}
+	cryptName, err := cryptDev.GetCryptDevice(key, arguments.Loopdev)
+
+	*reply = "/dev/mapper/" + cryptName
+
 	return err
 }
 
@@ -134,6 +150,7 @@ func (t *Methods) LoopDevice(arguments *args.LoopArgs, reply *int) error {
 	} else {
 		var err error
 		image, err = os.OpenFile(arguments.Image, arguments.Mode, 0600)
+		defer image.Close()
 		if err != nil {
 			return fmt.Errorf("could not open image file: %v", err)
 		}
--- a/internal/pkg/util/fs/mount/mount_linux.go
+++ b/internal/pkg/util/fs/mount/mount_linux.go
@@ -141,8 +141,9 @@ var authorizedTags = map[AuthorizedTag]struct {
 }
 
 var authorizedImage = map[string]fsContext{
-	"ext3":     {true},
-	"squashfs": {true},
+	"encryptfs": {true},
+	"ext3":      {true},
+	"squashfs":  {true},
 }
 
 var authorizedFS = map[string]fsContext{
--- a/pkg/build/types/bundle.go
+++ b/pkg/build/types/bundle.go
@@ -52,6 +52,8 @@ type Options struct {
 	LibraryAuthToken string `json:"libraryAuthToken"`
 	// contains docker credentials if specified
 	DockerAuthConfig *ocitypes.DockerAuthConfig
+	// Encrypted specifies if the filesystem needs to be encrypteded
+	Encrypted bool `json:"encrypt"`
 	// noTest indicates if build should skip running the test script
 	NoTest bool `json:"noTest"`
 	// force automatically deletes an existing container at build destination while performing build
@@ -69,8 +71,8 @@ type Options struct {
 	ImgCache *cache.Handle
 }
 
-// NewBundle creates a Bundle environment
-func NewBundle(bundleDir, bundlePrefix string) (b *Bundle, err error) {
+// Common code between NewBundle and NewEncryptedBundle
+func bundleCommon(encrypted bool, bundleDir, bundlePrefix string) (b *Bundle, err error) {
 	b = &Bundle{}
 	b.JSONObjects = make(map[string][]byte)
 
@@ -88,6 +90,8 @@ func NewBundle(bundleDir, bundlePrefix string) (b *Bundle, err error) {
 		"rootfs": "fs",
 	}
 
+	b.Opts.Encrypted = encrypted
+
 	for _, fso := range b.FSObjects {
 		if err = os.MkdirAll(filepath.Join(b.Path, fso), 0755); err != nil {
 			return
@@ -95,6 +99,17 @@ func NewBundle(bundleDir, bundlePrefix string) (b *Bundle, err error) {
 	}
 
 	return b, nil
+
+}
+
+// NewEncryptedBundle creates an Encrypted Bundle environment
+func NewEncryptedBundle(bundleDir, bundlePrefix string) (b *Bundle, err error) {
+	return bundleCommon(true, bundleDir, bundlePrefix)
+}
+
+// NewBundle creates a Bundle environment
+func NewBundle(bundleDir, bundlePrefix string) (b *Bundle, err error) {
+	return bundleCommon(false, bundleDir, bundlePrefix)
 }
 
 // Rootfs give the path to the root filesystem in the Bundle
--- a/pkg/runtime/engines/singularity/config/config_linux.go
+++ b/pkg/runtime/engines/singularity/config/config_linux.go
@@ -18,4 +18,5 @@ type EngineConfig struct {
 	File      *FileConfig      `json:"-"`
 	Network   *network.Setup   `json:"-"`
 	Cgroups   *cgroups.Manager `json:"-"`
+	CryptDev  string           `json:"-"`
 }
--- a/pkg/util/crypt/crypt_dev.go
+++ b/pkg/util/crypt/crypt_dev.go
@@ -1,0 +1,278 @@
+// Copyright (c) 2018-2019, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package crypt
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"syscall"
+
+	uuid "github.com/satori/go.uuid"
+	"github.com/sylabs/singularity/internal/pkg/sylog"
+	"github.com/sylabs/singularity/pkg/util/fs/lock"
+	"github.com/sylabs/singularity/pkg/util/loop"
+	"golang.org/x/crypto/ssh/terminal"
+)
+
+// Device describes a crypt device
+type Device struct{}
+
+// createLoop attaches the file to the next available loop device and
+// sets the sizelimit on it
+func createLoop(file *os.File, offset, size uint64) (string, error) {
+	loopDev := &loop.Device{
+		MaxLoopDevices: 256,
+		Shared:         true,
+		Info: &loop.Info64{
+			SizeLimit: size,
+			Offset:    offset,
+			Flags:     loop.FlagsAutoClear,
+		},
+	}
+	idx := 0
+	if err := loopDev.AttachFromFile(file, os.O_RDWR, &idx); err != nil {
+		return "", fmt.Errorf("failed to attach image %s: %s", file.Name(), err)
+	}
+	return fmt.Sprintf("/dev/loop%d", idx), nil
+}
+
+// CloseCryptDevice closes the crypt device
+func (crypt *Device) CloseCryptDevice(path string) error {
+
+	cmd := exec.Command("/sbin/cryptsetup", "luksClose", path)
+	cmd.SysProcAttr = &syscall.SysProcAttr{}
+	cmd.SysProcAttr.Credential = &syscall.Credential{Uid: 0, Gid: 0}
+	fd, err := lock.Exclusive("/dev/mapper")
+	if err != nil {
+		return err
+	}
+	defer lock.Release(fd)
+	err = cmd.Run()
+	if err != nil {
+		sylog.Debugf("Unable to delete the crypt device %s", err)
+		return err
+	}
+
+	return nil
+}
+
+// ReadKeyFromStdin reads key from terminal and returns it
+func (crypt *Device) ReadKeyFromStdin(confirm bool) (string, error) {
+
+	fmt.Print("Enter the Key: ")
+	password, err := terminal.ReadPassword(int(syscall.Stdin))
+	if err != nil {
+		sylog.Fatalf("Error parsing the key: %s", err)
+	}
+
+	input := string(password)
+	fmt.Println()
+	if confirm {
+		fmt.Print("Confirm the Key: ")
+		password2, err := terminal.ReadPassword(int(syscall.Stdin))
+		if err != nil {
+			sylog.Fatalf("Error parsing the key: %s", err)
+		}
+		input2 := string(password2)
+		fmt.Println()
+		if input != input2 {
+			return "", errors.New("Keys don't match")
+		}
+	}
+
+	return input, nil
+}
+
+// FormatCryptDevice allocates a loop device, encrypts, and returns the loop device name, and encrypted device name
+func (crypt *Device) FormatCryptDevice(path, key string) (string, string, error) {
+
+	f, err := os.Stat(path)
+	if err != nil {
+		return "", "", fmt.Errorf("failed getting size of %s", path)
+	}
+
+	fSize := f.Size()
+
+	// Create a temporary file to format with crypt header
+	cryptF, err := ioutil.TempFile("", "crypt-")
+	if err != nil {
+		sylog.Debugf("Error creating temporary crypt file")
+		return "", "", err
+	}
+	defer cryptF.Close()
+
+	// Truncate the file taking the squashfs size and crypt header into account
+	// Crypt header is around 2MB in size. Slightly over-allocate to be safe
+	devSize := fSize + 4*1024*1024 // 4MB for LUKS header
+
+	err = os.Truncate(cryptF.Name(), devSize)
+	if err != nil {
+		sylog.Debugf("Unable to truncate crypt file to size %d", devSize)
+		return "", "", err
+	}
+
+	// NOTE: This routine runs with root privileges. It's not necessary
+	// to explicitly set cmd's uid or gid here
+	cmd := exec.Command("/sbin/cryptsetup", "luksFormat", cryptF.Name())
+	stdin, err := cmd.StdinPipe()
+
+	if err != nil {
+		return "", "", err
+	}
+
+	go func() {
+		defer stdin.Close()
+		io.WriteString(stdin, key)
+	}()
+
+	err = cmd.Run()
+	if err != nil {
+		return "", "", fmt.Errorf("unable to format crypt device: %s", cryptF.Name())
+	}
+
+	// Associate the temporary crypt file with a loop device
+	loop, err := createLoop(cryptF, 0, uint64(devSize))
+	if err != nil {
+		return "", "", err
+	}
+
+	loopdev := filepath.Base(loop)
+
+	fd, err := lock.Exclusive("/dev/mapper")
+	if err != nil {
+		return "", "", fmt.Errorf("unable to acquire lock on /dev/mapper")
+	}
+	defer lock.Release(fd)
+
+	nextCrypt := getNextAvailableCryptDevice()
+	cmd = exec.Command("/sbin/cryptsetup", "luksOpen", loopdev, nextCrypt)
+	cmd.Dir = "/dev"
+	stdin, err = cmd.StdinPipe()
+	if err != nil {
+		return "", "", err
+	}
+
+	go func() {
+		defer stdin.Close()
+		io.WriteString(stdin, key)
+	}()
+
+	err = cmd.Run()
+	if err != nil {
+		sylog.Verbosef("Unable to open crypt device: %s", nextCrypt)
+		return "", "", err
+	}
+
+	copyDeviceContents(path, "/dev/mapper/"+nextCrypt, fSize)
+
+	// Open a new Temp file to stash the loop contents
+	loopF, err := ioutil.TempFile("", "loop-")
+	if err != nil {
+		sylog.Debugf("Error creating temporary crypt file")
+		return "", "", err
+	}
+
+	copyDeviceContents("/dev/"+loopdev, loopF.Name(), devSize)
+
+	return loopF.Name(), nextCrypt, err
+}
+
+// copyDeviceContents copies the contents of source to destination.
+// source and dest can either be a file or a block device
+func copyDeviceContents(source, dest string, size int64) error {
+	sourceFd, err := syscall.Open(source, syscall.O_RDWR, 0777)
+	if err != nil {
+		return fmt.Errorf("Unable to open the file %s", source)
+	}
+	defer syscall.Close(sourceFd)
+
+	destFd, err := syscall.Open(dest, syscall.O_RDWR, 0777)
+	if err != nil {
+		return fmt.Errorf("Unable to open the file: %s", dest)
+	}
+	defer syscall.Close(destFd)
+
+	var writtenSoFar int64
+
+	buffer := make([]byte, 1024)
+	for writtenSoFar < size {
+		numRead, err := syscall.Read(sourceFd, buffer)
+		if err != nil {
+			return fmt.Errorf("Unable to read the the file %s", source)
+		}
+		numWritten, err := syscall.Write(destFd, buffer)
+		if err != nil {
+			return fmt.Errorf("Unable to write to destination %s", dest)
+		}
+
+		if numRead != numWritten {
+			sylog.Debugf("numRead != numWritten")
+		}
+		writtenSoFar += int64(numWritten)
+	}
+
+	return nil
+}
+
+func getNextAvailableCryptDevice() string {
+	return (uuid.NewV4()).String()
+}
+
+// GetCryptDevice returns the next available device in /dev/mapper for encryption/decryption
+func (crypt *Device) GetCryptDevice(key, loopDev string) (string, error) {
+	fd, err := lock.Exclusive("/dev/mapper")
+	if err != nil {
+		sylog.Debugf("Unable to acquire lock on /dev/mapper while decrypting")
+		return "", err
+	}
+	defer lock.Release(fd)
+
+	maxRetries := 3 // Arbitrary number of retries.
+
+retry:
+	numRetries := 0
+	nextCrypt := getNextAvailableCryptDevice()
+	if nextCrypt == "" {
+		return "", errors.New("Crypt Device not available")
+	}
+
+	cmd := exec.Command("/sbin/cryptsetup", "luksOpen", loopDev, nextCrypt)
+	cmd.Dir = "/dev"
+	cmd.SysProcAttr = &syscall.SysProcAttr{}
+	cmd.SysProcAttr.Credential = &syscall.Credential{Uid: 0, Gid: 0}
+	stdin, err := cmd.StdinPipe()
+	if err != nil {
+		return "", err
+	}
+
+	go func() {
+		defer stdin.Close()
+		io.WriteString(stdin, key)
+	}()
+
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		if strings.Contains(string(out), "No key available") {
+			sylog.Debugf("Invalid password")
+		}
+		if strings.Contains(string(out), "Device already exists") {
+			numRetries++
+			if numRetries < maxRetries {
+				goto retry
+			}
+		}
+		return "", err
+	}
+	sylog.Debugf("Decrypted the FS successfully")
+
+	return nextCrypt, nil
+}
--- a/pkg/util/crypt/crypt_dev.go
+++ b/pkg/util/crypt/crypt_dev.go
@@ -66,6 +66,10 @@ func (crypt *Device) CloseCryptDevice(path string) error {
 }
 
 // ReadKeyFromStdin reads key from terminal and returns it
+// TODO (schebro): Fix #3816, #3851
+// Currently keys are being read interactively from the terminal.
+// Keys should be non-interactive, preferably in a keyfile that can
+// be passed to cryptsetup utility
 func (crypt *Device) ReadKeyFromStdin(confirm bool) (string, error) {
 
 	fmt.Print("Enter the Key: ")
@@ -122,6 +126,17 @@ func (crypt *Device) FormatCryptDevice(path, key string) (string, string, error)
 
 	// NOTE: This routine runs with root privileges. It's not necessary
 	// to explicitly set cmd's uid or gid here
+	// TODO (schebro): Fix #3818, #3821
+	// Currently we are relying on host's cryptsetup utility to encrypt and decrypt
+	// the SIF. The possiblity to saving a version of cryptsetup inside the container should be
+	// investigated. To do that, at least one additional partition is required, which is
+	// not encrypted.
+
+	// TODO (schebro): Fix #3819
+	// If we choose not to save a version of cryptsetup in container, host's cryptsetup utility's
+	// paty should be saved in a configuration file at build time (similar to mksquashfs) for
+	// security reasons
+
 	cmd := exec.Command("/sbin/cryptsetup", "luksFormat", cryptF.Name())
 	stdin, err := cmd.StdinPipe()
 
--- a/vendor/github.com/sylabs/sif/pkg/sif/fmt.go
+++ b/vendor/github.com/sylabs/sif/pkg/sif/fmt.go
@@ -89,6 +89,8 @@ func fstypeStr(ftype Fstype) string {
 		return "Archive"
 	case FsRaw:
 		return "Raw"
+	case FsEncryptedSquashfs:
+		return "Encrypted squashfs"
 	}
 	return "Unknown fs-type"
 }
--- a/vendor/github.com/sylabs/sif/pkg/sif/sif.go
+++ b/vendor/github.com/sylabs/sif/pkg/sif/sif.go
@@ -144,10 +144,11 @@ type Fstype int32
 
 // List of supported file systems
 const (
-	FsSquash  Fstype = iota + 1 // Squashfs file system, RDONLY
-	FsExt3                      // EXT3 file system, RDWR (deprecated)
-	FsImmuObj                   // immutable data object archive
-	FsRaw                       // raw data
+	FsSquash            Fstype = iota + 1 // Squashfs file system, RDONLY
+	FsExt3                                // EXT3 file system, RDWR (deprecated)
+	FsImmuObj                             // immutable data object archive
+	FsRaw                                 // raw data
+	FsEncryptedSquashfs                   // Encrypted Squashfs file system, RDONLY
 )
 
 // Parttype represents the different SIF container partition types (system and data)
--- a/cmd/internal/cli/build.go
+++ b/cmd/internal/cli/build.go
@@ -25,6 +25,7 @@ import (
 
 var (
 	remote         bool
+	encrypt        bool
 	builderURL     string
 	detached       bool
 	libraryURL     string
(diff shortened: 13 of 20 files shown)


internal/pkg/runtime/engines/singularity/cleanup_linux.go
```diff
@@ -51,5 +52,43 @@ func (engine *EngineOperations) CleanupContainer(fatal error, status syscall.Wai
 		return file.Delete()
 	}
 
+	if engine.EngineConfig.CryptMount != "" {
+		// Elevate the privilege to unmount and delete the crypt device
+		uid := os.Getuid()
+		err := syscall.Setresuid(uid, 0, uid)
```
cclerget: You need to lock os thread with `runtime.LockOSThread()` before escalating privileges and call `runtime.UnlockOSThread()` after privileges drop, otherwise runtime scheduler can execute routine on this thread which could lead to potential side effects

internal/pkg/runtime/engines/singularity/cleanup_linux.go
```diff
@@ -51,5 +52,43 @@ func (engine *EngineOperations) CleanupContainer(fatal error, status syscall.Wai
...
+		if err != nil {
+			sylog.Debugf("Err setting suid")
+			return err
+		}
+
+		err = syscall.Unmount("/usr/local/var/singularity/mnt/session/final", syscall.MNT_DETACH)
```
cclerget: You can use `filepath.Join(buildcfg.SESSIONDIR, "final")` to match user custom installation

internal/pkg/runtime/engines/singularity/cleanup_linux.go
```diff
@@ -51,5 +52,43 @@ func (engine *EngineOperations) CleanupContainer(fatal error, status syscall.Wai
...
+		if err != nil {
+			syscall.Setresuid(uid, uid, 0)
+			sylog.Debugf("Error while unmounting overlay FS: %s", err)
+			return err
+		}
+
+		err = syscall.Unmount("/usr/local/var/singularity/mnt/session/rootfs", syscall.MNT_DETACH)
```
cclerget: same with `filepath.Join(buildcfg.SESSIONDIR, "rootfs")`

internal/pkg/runtime/engines/singularity/container_linux.go
```diff
@@ -612,11 +628,16 @@ func (c *container) addRootfsMount(system *mount.System) error {
 	offset := imageObject.Partitions[0].Offset
 	size := imageObject.Partitions[0].Size
 
+	sylog.Debugf("Image type is %v", imageObject.Partitions[0].Name)
+
 	switch imageObject.Partitions[0].Type {
 	case image.SQUASHFS:
 		mountType = "squashfs"
+		sylog.Debugf("Mount type is squashfs")
```
cclerget: you can remove this debug, it's redundant with line 631

internal/pkg/runtime/engines/singularity/container_linux.go
```diff
@@ -855,6 +876,7 @@ func (c *container) addKernelMount(system *mount.System) error {
 	} else {
 		sylog.Verbosef("Skipping /sys mount")
 	}
+	err = system.Points.AddBind(mount.KernelTag, "/sbin", "/sbin", bindFlags)
```
cclerget: Need to be removed

internal/pkg/runtime/engines/singularity/rpc/server/server_linux.go
```diff
@@ -33,6 +38,52 @@ func (t *Methods) Mount(arguments *args.MountArgs, reply *int) (err error) {
...
 
+// Crypt decrypts the loop device
+func (t *Methods) Decrypt(arguments *args.CryptArgs, reply *string) (err error) {
+
+	sylog.Debugf("In Crypt RPC")
+	sylog.Debugf("Crypt RPC running in PID %d", os.Getpid())
+	sylog.Debugf("Loop device is %s", arguments.Loopdev)
+
+	fmt.Print("Enter the password to decrypt File System: ")
+	password, err := terminal.ReadPassword(int(syscall.Stdin))
```
cclerget: I would rather ask password from master process and pass it as argument in https://github.com/sylabs/singularity/pull/3729/files#diff-784d2c712e51a05593054a6d797ab51bR39

internal/pkg/runtime/engines/singularity/cleanup_linux.go
```diff
@@ -51,5 +52,43 @@ func (engine *EngineOperations) CleanupContainer(fatal error, status syscall.Wai
...
+		if err != nil {
+			syscall.Setresuid(uid, uid, 0)
+			sylog.Debugf("Error while unmounting Rootfs: %s", err)
+			return err
+		}
+
+		cmd := exec.Command("/sbin/cryptsetup", "luksClose", engine.EngineConfig.CryptMount)
```
cclerget: Maybe use `exec.LookPath` instead of hardcoded `/sbin/cryptsetup`

internal/pkg/runtime/engines/singularity/rpc/server/server_linux.go
```diff
@@ -33,6 +38,52 @@ func (t *Methods) Mount(arguments *args.MountArgs, reply *int) (err error) {
...
+
+	crypt_dev := &crypt.Device{
+		MaxDevices: 256,
+	}
+
+	cdev_str, err := crypt_dev.GetCryptDevice()
+
+	sylog.Debugf("Crypt device is %s\n", cdev_str)
+
+	cmd := exec.Command("/sbin/cryptsetup", "luksOpen", arguments.Loopdev, cdev_str, "-v", "--debug")
```
cclerget: Maybe use `exec.LookPath` instead of hardcoded `/sbin/cryptsetup`

internal/pkg/util/fs/mount/mount_linux.go
```diff
@@ -581,6 +584,7 @@ func (p *Points) AddImage(tag AuthorizedTag, source string, dest string, fstype
 	if flags&(syscall.MS_BIND|syscall.MS_REMOUNT|syscall.MS_REC) != 0 {
 		return fmt.Errorf("MS_BIND, MS_REC or MS_REMOUNT are not valid flags for image mount points")
 	}
+	sylog.Debugf("FStype is %s", fstype)
```
cclerget: Can also be removed

pkg/build/types/bundle.go
```diff
@@ -64,8 +76,25 @@ type Options struct {
 	NoCleanUp bool `json:"noCleanUp"`
 }
 
+func createLoop(file *os.File, offset, size uint64) (string, error) {
+	loopDev := &loop.Device{
+		MaxLoopDevices: 256,
+		Shared:         true,
```
cclerget: `Shared` is meaningful when images are open in read-only mode, so it's not required here

pkg/build/types/bundle.go
```diff
@@ -83,10 +112,116 @@ func NewBundle(bundleDir, bundlePrefix string) (b *Bundle, err error) {
...
+		}
+
+		// Create a sparse file in tmp dir
+		f, err := os.Create(b.Path + "/sparse_fs.loop")
+		if err != nil {
+			return nil, err
+		}
+
+		// Create a 500MB sparse file
+		err = f.Truncate(5 * 1e8)
```
cclerget: We will really need to determine final size dynamically

pkg/runtime/engines/singularity/config/config_linux.go
```diff
@@ -13,9 +13,10 @@ import (
...
-	OciConfig *oci.Config      `json:"ociConfig"`
-	File      *FileConfig      `json:"-"`
-	Network   *network.Setup   `json:"-"`
-	Cgroups   *cgroups.Manager `json:"-"`
+	JSON       *JSONConfig      `json:"jsonConfig"`
+	OciConfig  *oci.Config      `json:"ociConfig"`
+	File       *FileConfig      `json:"-"`
+	Network    *network.Setup   `json:"-"`
+	Cgroups    *cgroups.Manager `json:"-"`
+	CryptMount string           `json:"_"`
```
cclerget:
```suggestion
	CryptMount string           `json:"-"`
```

vendor/github.com/sylabs/sif/pkg/sif/fmt.go
```diff
@@ -89,6 +89,8 @@ func fstypeStr(ftype Fstype) string {
 		return "Archive"
 	case FsRaw:
 		return "Raw"
+	case FsEncrypt:
```
cclerget: This change requires a PR in https://github.com/sylabs/sif repository. We don't touch vendor

vendor/github.com/sylabs/sif/pkg/sif/sif.go
```diff
@@ -148,6 +148,7 @@ const (
 	FsExt3                      // EXT3 file system, RDWR (deprecated)
 	FsImmuObj                   // immutable data object archive
 	FsRaw                       // raw data
+	FsEncrypt                   // Encrypted File System
```
cclerget: same

vendor/github.com/sylabs/sif/pkg/sif/sif.go
```diff
@@ -299,6 +300,7 @@ type DescriptorInput struct {
 	Link      uint32   // link to be set for new descriptor
 	Size      int64    // size of the data object for the new descriptor
 	Alignment int      // Align requirement for data object
+	Encrypt   bool
```
cclerget: same

pkg/build/types/bundle.go
```diff
@@ -83,10 +112,116 @@ func NewBundle(bundleDir, bundlePrefix string) (b *Bundle, err error) {
...
+		}()
+
+		out, err = cmd.CombinedOutput()
+		if err != nil {
+			sylog.Verbosef("Out is %s, err is %s", out, err)
+			return nil, err
+		}
+
+		// Create an EXT3 FS in the mapped device
+		cmd = exec.Command("mkfs.ext3", cdevStr)
```
cclerget: Disable journal ?

internal/pkg/runtime/engines/singularity/rpc/server/server_linux.go
```diff
@@ -33,6 +38,52 @@ func (t *Methods) Mount(arguments *args.MountArgs, reply *int) (err error) {
 	return err
 }
 
+// Crypt decrypts the loop device
+func (t *Methods) Decrypt(arguments *args.CryptArgs, reply *string) (err error) {
+
+	sylog.Debugf("In Crypt RPC")
```
cclerget: The whole code could be wrapped in a package function in `pkg/util/crypt/`

pkg/image/encryptfs.go
```diff
@@ -0,0 +1,49 @@
+package image
```
cclerget: Looks like this file is not required ?

cmd/singularity/cli.go
```diff
@@ -16,7 +16,6 @@ func main() {
 	// In cmd/internal/cli/singularity.go
 	cli.ExecuteSingularity()
 }
-
```
cclerget: Need a blank line here

internal/pkg/build/assemblers/assembler_sif.go
```diff
@@ -139,45 +144,68 @@ func getMksquashfsPath() (string, error) {
...
+	if encrypted == true {
+		err = syscall.Unmount(b.Rootfs(), 0)
+		if err != nil {
+			return fmt.Errorf("Unable to unmount Rootfs: %s err: %s", b.Rootfs(), err)
+		}
 
-	if err := mksquashfsCmd.Start(); err != nil {
-		return fmt.Errorf("While starting mksquashfs: %v", err)
-	}
+		cmd := exec.Command("/sbin/cryptsetup", "luksClose", b.LoopPath)
```
cclerget: Use `exec.LookPath`

internal/pkg/runtime/engines/imgbuild/create_linux.go
```diff
@@ -57,7 +58,10 @@ func (engine *EngineOperations) CreateContainer(pid int, rpcConn net.Conn) error
 	// sensible mount point options to avoid accidental system settings override
 	flags := uintptr(syscall.MS_BIND | syscall.MS_NOSUID | syscall.MS_NOEXEC | syscall.MS_NODEV | syscall.MS_RDONLY)
 
-	sylog.Debugf("Mounting image directory %s\n", rootfs)
+	bpath := engine.EngineConfig.Bundle.Path
+	// Create a sparse file in /tmp
+	ioutil.TempFile(bpath, "temp_loop")
```
cclerget: dead code ?

internal/pkg/runtime/engines/singularity/container_linux.go
```diff
@@ -612,11 +628,16 @@ func (c *container) addRootfsMount(system *mount.System) error {
 	offset := imageObject.Partitions[0].Offset
 	size := imageObject.Partitions[0].Size
 
+	sylog.Debugf("Image type is %v", imageObject.Partitions[0].Name)
```
cclerget:
```suggestion
	sylog.Debugf("Image type is %v", imageObject.Partitions[0].Type)
```

cclerget (changes requested):
Overall, there is some related crypto code which could find a place in `pkg/util/crypt` package. Did you give a try with squashfs approach which seems simpler imo: https://martin.elwin.com/blog/2008/05/backups-with-squashfs-and-luks/

schebro:
> Overall, there is some related crypto code which could find a place in `pkg/util/crypt` package. Did you give a try with squashfs approach which seems simpler imo: https://martin.elwin.com/blog/2008/05/backups-with-squashfs-and-luks/

I will go through it today. It sounds interesting. I tried this approach before but gave up when I couldn't reliably create a LUKS container to stash the squashfs.

pkg/image/encryptfs.go
schebro: yes. forgot to remove it.

internal/pkg/runtime/engines/singularity/rpc/server/server_linux.go
schebro: Any specific reason to do that instead of reading here?

internal/pkg/runtime/engines/singularity/cleanup_linux.go
schebro: Runtime doesn't seem to have PATH set. The only env variable I see in runtime is SINGULARITY_MESSAGELEVEL=5. PATH is set inside the container though.

internal/pkg/runtime/engines/singularity/rpc/server/server_linux.go
cclerget: Well actually this won't work when running a Singularity instance because there is no terminal available at this stage, this is also the case in master process but that could be fixed more easily for master process in the future. Additionally I'm not comfortable to let user interact with a program executed with root privileges

internal/pkg/runtime/engines/singularity/cleanup_linux.go
cclerget: Runtime could set a default PATH like `/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`

vendor/github.com/sylabs/sif/pkg/sif/sif.go
schebro: Should I make these changes separately in sif repository and push?

WestleyK: I would fork the [sif repo](https://github.com/sylabs/sif), then checkout your branch and add these changes, push to your branch and open a PR. _Thats how I would do it..._

Then you can update the vendor in this PR.

internal/pkg/runtime/engines/singularity/cleanup_linux.go
schebro: We cleanup all the env variables in cleanenv() in starter.c. Wonder if it's for any specific reason. Otherwise, I can just add them back.

internal/pkg/build/assemblers/assembler_sif.go
```diff
@@ -90,7 +91,11 @@ func createSIF(path string, definition, ociConf []byte, squashfile string) (err
 	parinput.Fp = fp
 	parinput.Size = fi.Size()
 
-	err = parinput.SetPartExtra(sif.FsSquash, sif.PartPrimSys, sif.GetSIFArch(runtime.GOARCH))
+	if encrypted == false {
+		err = parinput.SetPartExtra(sif.FsSquash, sif.PartPrimSys, sif.GetSIFArch(runtime.GOARCH))
```
mem: this is changing the first argument of the function. It would be preferred to do this:

```
if encrypted {
    sifType = sif.FsEncrypt
} else {
    sifType = sif.FsSquash
}
```

internal/pkg/build/assemblers/assembler_sif.go
```diff
@@ -139,45 +144,68 @@ func getMksquashfsPath() (string, error) {
...
-	defer os.Remove(squashfsPath)
+	sylog.Debugf("Encrypted loop device is %s", b.LoopPath)
 
-	args := []string{b.Rootfs(), squashfsPath, "-noappend"}
-
-	// build squashfs with all-root flag when building as a user
-	if syscall.Getuid() != 0 {
-		args = append(args, "-all-root")
+	encrypted := false
+	if b.LoopPath != "" {
```
mem: `LoopPath` is too generic to signify that this bundle. The `Opts` member of `types.Bundle` seems more appropriate to contain a flag to indicate that the result should be encrypted.

internal/pkg/build/assemblers/assembler_sif.go
```diff
@@ -139,45 +144,68 @@ func getMksquashfsPath() (string, error) {
...
+		err = syscall.Unmount(b.Rootfs(), 0)
+		if err != nil {
+			return fmt.Errorf("Unable to unmount Rootfs: %s err: %s", b.Rootfs(), err)
+		}
 
-	if err := mksquashfsCmd.Start(); err != nil {
-		return fmt.Errorf("While starting mksquashfs: %v", err)
-	}
+		cmd := exec.Command("/sbin/cryptsetup", "luksClose", b.LoopPath)
+		cmd.Dir = "/dev/mapper"
```
mem: can you please add a comment indicating why you need to execute the command in this directory?

internal/pkg/build/assemblers/assembler_sif.go
```diff
@@ -46,6 +46,7 @@ func createSIF(path string, definition, ociConf []byte, squashfile string) (err
 		Groupid:  sif.DescrDefaultGroup,
 		Link:     sif.DescrUnusedLink,
 		Data:     definition,
+		Encrypt:  true,
```
mem: shouldn't this use the `encrypted` argument to `createSIF`?

internal/pkg/build/assemblers/assembler_sif.go
```diff
@@ -139,45 +144,68 @@ func getMksquashfsPath() (string, error) {
...
+	if b.LoopPath != "" {
+		encrypted = true
 	}
 
-	mksquashfsCmd := exec.Command(mksquashfs, args...)
-	stderr, err := mksquashfsCmd.StderrPipe()
-	if err != nil {
-		return fmt.Errorf("While setting up stderr pipe: %v", err)
-	}
+	if encrypted == true {
```
mem: can you please move both branches of this `if` to separate functions? They seem to be doing two completely different things that need to be documented on their own.

internal/pkg/build/assemblers/assembler_sif.go
```diff
@@ -139,45 +144,68 @@ func getMksquashfsPath() (string, error) {
 func (a *SIFAssembler) Assemble(b *types.Bundle, path string) (err error) {
```
mem, line 146: It seems to me that you have two somewhat similar but different enough functions to warrant two different types.

One would be the existing SIFAssembler, that creates unencrypted images, and the other would be SIFEncryptedAssembler that creates encrypted images.

This probably makes it easier to stack the assemblers, too. At some point in the future what you want to do is create a squashfs filesystem, push the data to it, set up an encrypted LUKS container, and push the squashfs data to it. In order to do that, you need to know how big the LUKS container needs to be, and that means you need to create the squashfs filesystem first. This is easier if you have two separate assemblers.

internal/pkg/runtime/engines/imgbuild/create_linux.go
mem: not quite sure how this got resolved?

Since ioutil.TempFile returns an open file handle, this doesn't seem to be doing anything with that new file (since it's not storing the variable anywhere)

internal/pkg/runtime/engines/singularity/cleanup_linux.go
```diff
@@ -51,5 +52,43 @@ func (engine *EngineOperations) CleanupContainer(fatal error, status syscall.Wai
 		return file.Delete()
 	}
 
+	if engine.EngineConfig.CryptMount != "" {
```
mem: it seems to be this whole bit of code belongs in a separate function, possibly part of `config.Common`.

internal/pkg/runtime/engines/singularity/cleanup_linux.go
mem: looks like you want to place dropping privilege in a defer'ed function

internal/pkg/runtime/engines/singularity/cleanup_linux.go
```diff
@@ -51,5 +52,43 @@ func (engine *EngineOperations) CleanupContainer(fatal error, status syscall.Wai
...
+
+		err = syscall.Unmount("/usr/local/var/singularity/mnt/session/rootfs", syscall.MNT_DETACH)
+		if err != nil {
+			syscall.Setresuid(uid, uid, 0)
+			sylog.Debugf("Error while unmounting Rootfs: %s", err)
+			return err
+		}
+
+		cmd := exec.Command("/sbin/cryptsetup", "luksClose", engine.EngineConfig.CryptMount)
+		cmd.Dir = "/dev/mapper"
```
mem: also needs a comment mentioning why this needs to run in this directory.

internal/pkg/runtime/engines/singularity/cleanup_linux.go
```diff
@@ -51,5 +52,43 @@ func (engine *EngineOperations) CleanupContainer(fatal error, status syscall.Wai
...
+		cmd.SysProcAttr.Credential = &syscall.Credential{Uid: 0, Gid: 0}
+		_, err = cmd.CombinedOutput()
+		if err != nil {
+			sylog.Debugf("Unable to delete the crypt device %s", err)
+			syscall.Setresuid(uid, uid, 0)
+			return err
+		}
+
+		// Restore the privilege
+		err = syscall.Setresuid(uid, uid, 0)
```
mem: this is dropping this error, no?

internal/pkg/runtime/engines/singularity/container_linux.go
```diff
@@ -545,10 +544,27 @@ func (c *container) mountImage(mnt *mount.Point) error {
...
+	sylog.Debugf("Mounting loop device %s to %s of type %s\n", path, mnt.Destination, mnt.Type)
+
+	if mnt.Type == "encryptfs" {
+		sp := strings.Split(path, "/")
+		loopdev := sp[len(sp)-1]
+		dev, err := c.rpcOps.Decrypt(offset, loopdev)
+
+		sylog.Debugf("Device to be mounted is %s", dev)
+
+		_, err = c.rpcOps.Mount("/dev/mapper/"+dev, mnt.Destination, "ext3", flags, optsString)
```
mem: assuming we change the format of the encrypted data to squashfs (and this is something we have to either express in the SIF metadata or simply assume, which would require documenting), the encrypted/non-encrypted bit of info should be a boolean stored as part of mount.Point.

If that's done that way, we can call the open operation (via Decrypt, I guess?) and obtain the device. We can override path with that value we obtained from the RPC server. At that point the mount operation is the same for encrypted / non-encrypted images, so line 564 does not need to be on an else branch.

internal/pkg/runtime/engines/singularity/container_linux.go
```diff
@@ -545,10 +544,27 @@ func (c *container) mountImage(mnt *mount.Point) error {
...
-	_, err = c.rpcOps.Mount(path, mnt.Destination, mnt.Type, flags, optsString)
-	if err != nil {
-		return fmt.Errorf("failed to mount %s filesystem: %s", mnt.Type, err)
+
+	sylog.Debugf("Mounting loop device %s to %s of type %s\n", path, mnt.Destination, mnt.Type)
+
+	if mnt.Type == "encryptfs" {
+		sp := strings.Split(path, "/")
+		loopdev := sp[len(sp)-1]
+		dev, err := c.rpcOps.Decrypt(offset, loopdev)
```
mem: it's better if this receives path as argument, not "loopdev". It's also better if the call returns the full path to the device, not just the name used with device mapper.

Basically, this code can become:

```
newPath, err := c.rpcOps.Decrypt(offset, path)

File: internal/pkg/runtime/engines/singularity/rpc/server/server_linux.go [hunk 34]
```diff
@@ -33,6 +38,52 @@ func (t *Methods) Mount(arguments *args.MountArgs, reply *int) (err error) {
...
+	if err != nil {
+		sylog.Fatalf("Error parsing input: %s", err)
+	}
+	fmt.Println()
+
+	crypt_dev := &crypt.Device{
+		MaxDevices: 256,
+	}
+
+	cdev_str, err := crypt_dev.GetCryptDevice()
```
mem: please, no snake_case. Go uses camelCase.

internal/pkg/runtime/engines/singularity/rpc/server/server_linux.go
mem: let's not depend on PATH lookups inside privileged code.

If you need to resolve the path to cryptsetup, do it at build time, and introduce a configuration variable so that the admin that override that location.

internal/pkg/runtime/engines/singularity/rpc/server/server_linux.go
```diff
@@ -33,6 +38,52 @@ func (t *Methods) Mount(arguments *args.MountArgs, reply *int) (err error) {
...
+
+	sylog.Debugf("Crypt device is %s\n", cdev_str)
+
+	cmd := exec.Command("/sbin/cryptsetup", "luksOpen", arguments.Loopdev, cdev_str, "-v", "--debug")
+	cmd.Dir = "/dev"
```
mem: can you please document why you need this?

internal/pkg/runtime/engines/singularity/rpc/server/server_linux.go
```diff
@@ -134,6 +185,7 @@ func (t *Methods) LoopDevice(arguments *args.LoopArgs, reply *int) error {
 	} else {
 		var err error
 		image, err = os.OpenFile(arguments.Image, arguments.Mode, 0600)
+		defer image.Close()
```
mem, line 153: is this necessary? "image" is used below as an argument to loopdev.AttachFromFile, and that sets things up so that this file descriptor is closed. There's no explicit call to image.Close(), that's right. And there might be some error paths where this is not closed, too.

Did you find a case where this was needed?

pkg/build/types/bundle.go
```diff
@@ -64,8 +76,25 @@ type Options struct {
...
+	idx := 0
+	if err := loopDev.AttachFromFile(file, os.O_RDWR, &idx); err != nil {
+		return "", fmt.Errorf("failed to attach image %s: %s", file.Name(), err)
+	}
+	return fmt.Sprintf("/dev/loop%d", idx), nil
+}
+
 // NewBundle creates a Bundle environment
-func NewBundle(bundleDir, bundlePrefix string) (b *Bundle, err error) {
+func NewBundle(encrypt bool, bundleDir, bundlePrefix string) (b *Bundle, err error) {
```
mem: see above

pkg/build/types/bundle.go
```diff
@@ -83,10 +112,116 @@ func NewBundle(bundleDir, bundlePrefix string) (b *Bundle, err error) {
 		"rootfs": "fs",
 	}
 
+	var input = "Default"
+	if encrypt == true {
```
mem: this branch is too long. Please move it to a separate function.

I was thinking it should be possible to set up the contents of the file system as usual, create the squashfs filesystem as usual, and then create a loop device of the correct size (squashfs + ~ 16 MB overhead, exact number is 16384 for the metadata + 16744448 key slots, it might be possible to reduce the size of the key slot storage), format it and the copy the squashfs data into the encrypted volume (e.g. `dd if=squashfs.img of=/dev/mapper/luks-volume ...`).

pkg/util/crypt/crypt_dev.go
```diff
@@ -0,0 +1,44 @@
...
+
+func (e *errCryptDeviceUnavailable) Error() string {
+	return e.message
+}
+
+// GetCryptDevice returns the next available device in /dev/mapper for encryption/decryption
+func (crypt *Device) GetCryptDevice() (string, error) {
+	// Return the next available crypt device
+	for i := 0; i < crypt.MaxDevices; i++ {
+		retStr := fmt.Sprintf("singularity_crypt_%d", i)
```
mem: it might be better to use a random ID (e.g. uuid) to reduce the chance of collisions.

mem (changes requested):

internal/pkg/runtime/engines/singularity/rpc/server/server_linux.go
schebro: The flags are unnecessary. I will delete them.

internal/pkg/runtime/engines/singularity/rpc/server/server_linux.go
schebro: Sorry.. My leftover C habits. :-) Will change.

internal/pkg/runtime/engines/singularity/rpc/server/server_linux.go
schebro: Didn't really find any cases that are going wrong with this image not being closed. This change has nothing to do with the overall change. I just thought image should be closed before we exit this function.

internal/pkg/build/assemblers/assembler_sif.go
schebro: In the new code, I am using squashfs image to create an encrypted one.

internal/pkg/build/assemblers/assembler_sif.go
```diff
@@ -139,19 +147,19 @@ func getMksquashfsPath() (string, error) {
...
 	sylog.Infof("Creating SIF file...")
 
+	var fsPath string
+
 	mksquashfs, err := getMksquashfsPath()
 	if err != nil {
 		return fmt.Errorf("While searching for mksquashfs: %v", err)
 	}
-
 	f, err := ioutil.TempFile(b.Path, "squashfs-")
```
mem, line 155: I missed this before.

The idea of using ioutil.TempFile is that you get a file handle for a file that did not exist before, and that you can use to create a new file without worrying about races. Simplest case is that you want to create an image, and _always_ use the name "/tmp/new_image" for the temporary file. Someone can create a symlink from "/tmp/new_image" to an interesting file so that your process clobbers it. So to mitigate that, you remove "/tmp/new_image" before opening it for writing. Now you have a race: there's a window of time between removing the file and creating the file that the attacker can use to create the symlink again. To mitigate that, you use a non-fixed filename, e.g. "/tmp/new_image_123". You still have a race: the attacker can try creating _all_ the possible filenames.

ioutil.TempFile deals with all of that: it repeatedly calls open with O_CREATE|O_EXCL so that it knows that the file did not exist before it created it. By _removing_ the file before creating it again (line 159), this code is opening up the window for the race again. By using a different filename than the one ioutil.TempFile returned (line 157) it's opening the window again. Just use whatever ioutil.TempFile returns, do _not_ remove the file and pass `-noappend` to mksquashfs (this will cause it to truncate the file, which is fine, because the file is empty anyways). After you are done, call `os.Rename` to rename the file to its final name if needed.

pkg/util/crypt/crypt_dev.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	return nil
+}
+
+// FormatCryptDevice allocates a loop device, encrypts, and returns the loop device name, and encrypted device name
+func (crypt *Device) FormatCryptDevice(path string) (string, string, error) {
+
+	// Read the password from terminal
+	fmt.Print("Enter the password to encrypt the filesystem: ")
```
mem: before we can merge this, can you please remove the interactive prompts from here and simply pass the secret as an argument to this function?

pkg/util/crypt/crypt_dev.go
```diff
@@ -0,0 +1,306 @@
...
+		sylog.Fatalf("Error parsing the password: %s", err)
+	}
+	input2 := string(password2)
+	fmt.Println()
+
+	if input != input2 {
+		return "", "", errors.New("Passwords don't match")
+	}
+
+	squashF, err := os.Stat(path)
```
mem: there's nothing in this that is specific to squashfs.

pkg/util/crypt/crypt_dev.go
```diff
@@ -0,0 +1,306 @@
...
+	// NOTE: This routine runs with root privileges. It's not necessary
+	// to explicitly set cmd's uid or gid here
+	cmd := exec.Command("/sbin/cryptsetup", "luksFormat", cryptF.Name())
+	stdin, err := cmd.StdinPipe()
+
+	if err != nil {
+		return "", "", err
+	}
+
+	go func() {
```
mem, line 147: you don't need a goroutine for this. You can call cmd.Start(), which does not block.

pkg/util/crypt/crypt_dev.go
```diff
@@ -0,0 +1,306 @@
...
+
+	err = os.Truncate(cryptF.Name(), size)
+	if err != nil {
+		sylog.Debugf("Unable to truncate crypt file to size %d", size)
+		return "", "", err
+	}
+
+	// NOTE: This routine runs with root privileges. It's not necessary
+	// to explicitly set cmd's uid or gid here
+	cmd := exec.Command("/sbin/cryptsetup", "luksFormat", cryptF.Name())
```
mem, line 140: it's probably a good idea to pass --key-file, to not rely on the interactive prompting. The file passed can be "-" so that it still reads from stdin.

pkg/util/crypt/crypt_dev.go
```diff
@@ -0,0 +1,306 @@
...
+		return "", "", err
+	}
+
+	// Associate the temporary crypt file with a loop device
+	loop, err := createLoop(cryptF, 0, uint64(size))
+	if err != nil {
+		return "", "", err
+	}
+
+	sp := strings.Split(loop, "/")
```
mem: guessing the bit of code to obtain a name for the dm device should be its own function.

pkg/util/crypt/crypt_dev.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+	nextCrypt := getNextAvailableCryptDevice()
+	cmd = exec.Command("/sbin/cryptsetup", "luksOpen", loopdev, nextCrypt)
+	cmd.Dir = "/dev"
+	stdin, err = cmd.StdinPipe()
+	if err != nil {
+		return "", "", err
+	}
+
+	go func() {
```
mem, line 179: same here

pkg/util/crypt/crypt_dev.go
```diff
@@ -0,0 +1,306 @@
...
+	sp := strings.Split(loop, "/")
+	loopdev := sp[len(sp)-1]
+
+	fd, err := lock.Exclusive("/dev/mapper")
+	if err != nil {
+		sylog.Debugf("Unable to acquire lock on /dev/mapper")
+		return "", "", err
+	}
+	nextCrypt := getNextAvailableCryptDevice()
+	cmd = exec.Command("/sbin/cryptsetup", "luksOpen", loopdev, nextCrypt)
```
mem, line 172: --key-file here, too

pkg/util/crypt/crypt_dev.go
```diff
@@ -0,0 +1,306 @@
...
+	copyDeviceContents("/dev/"+loopdev, loopF.Name(), size)
+
+	return loopF.Name(), nextCrypt, err
+}
+
+// copyDeviceContents copies the contents of source to destination.
+// source and dest can either be a file or a block device
+func copyDeviceContents(source string, dest string, size int64) error {
+
+	squashFd, err := syscall.Open(source, syscall.O_RDWR, 0777)
```
mem: why do you need to use the system calls directly? Open the two files and use io.Copy?

mem (changes requested):
about halfway thru

pkg/util/crypt/crypt_dev.go
schebro: I tried that first. io.Copy doesn't work with block devices. Will try once more.

pkg/util/crypt/crypt_dev.go
schebro: Ok. What you are suggesting is to take our the interactive password entry and replace with a keyfile, right?

pkg/util/crypt/crypt_dev.go
schebro: This go func is providing the password from stdin to the command. But if I use keyfile, these go away.

pkg/util/crypt/crypt_dev.go
schebro: Right. "path" here belongs to squashfs. That's why I named it as such. Will change.

internal/pkg/runtime/engines/imgbuild/create_linux.go
```diff
@@ -60,7 +60,6 @@ func (engine *EngineOperations) CreateContainer(pid int, rpcConn net.Conn) error
 		flags = uintptr(syscall.MS_BIND | syscall.MS_REC)
 	}
 
-	sylog.Debugf("Mounting image directory %s\n", rootfs)
```
cclerget: Why is it removed ?

internal/pkg/runtime/engines/singularity/cleanup_linux.go
```diff
@@ -58,5 +63,47 @@ func (engine *EngineOperations) CleanupContainer(fatal error, status syscall.Wai
...
+	return nil
+}
+
+func cleanupCrypt(path string) error {
+
+	// Elevate the privilege to unmount and delete the crypt device
+	runtime.LockOSThread()
+	uid := os.Getuid()
+	err := syscall.Setresuid(uid, 0, uid)
+	defer syscall.Setresuid(uid, uid, 0)
```
cclerget: I didn't notice during first review, but `defer syscall.Setresuid(uid, uid, 0)` defer call must be after `defer runtime.UnlockOSThread()`, `UnlockOSThread` call need to be the last (the first defer call is the last executed in the stack) and I believe `syscall` functions trigger a scheduler yield, so `syscall.Setresuid(uid, uid, 0)` could be scheduled on a different thread and drop privileges for a different OS thread leaving privileges for the previously locked thread

internal/pkg/runtime/engines/singularity/cleanup_linux.go
```diff
@@ -58,5 +63,47 @@ func (engine *EngineOperations) CleanupContainer(fatal error, status syscall.Wai
...
+	defer runtime.UnlockOSThread()
+	if err != nil {
+		sylog.Debugf("Err setting suid")
+		return err
+	}
+
+	err = syscall.Unmount(filepath.Join(buildcfg.SESSIONDIR, "final"), syscall.MNT_DETACH)
+	if err != nil {
+		sylog.Debugf("Error while unmounting overlay FS: %s", err)
```
cclerget:
```suggestion
		sylog.Debugf("Error while unmounting final session directory: %s", err)
```

internal/pkg/runtime/engines/singularity/cleanup_linux.go
```diff
@@ -58,5 +63,47 @@ func (engine *EngineOperations) CleanupContainer(fatal error, status syscall.Wai
...
+		return err
+	}
+
+	err = syscall.Unmount(filepath.Join(buildcfg.SESSIONDIR, "rootfs"), syscall.MNT_DETACH)
+	if err != nil {
+		sylog.Debugf("Error while unmounting Rootfs: %s", err)
```
cclerget:
```suggestion
		sylog.Debugf("Error while unmounting rootfs session directory: %s", err)
```

internal/pkg/runtime/engines/singularity/cleanup_linux.go
```diff
@@ -58,5 +63,47 @@ func (engine *EngineOperations) CleanupContainer(fatal error, status syscall.Wai
...
+		return err
+	}
+
+	sp := strings.Split(path, "/")
```
cclerget: you can use `devName := filepath.Base(path)` and drop line 99

internal/pkg/runtime/engines/singularity/container_linux.go
```diff
@@ -571,10 +570,29 @@ func (c *container) mountImage(mnt *mount.Point) error {
...
 
 	path := fmt.Sprintf("/dev/loop%d", number)
-	sylog.Debugf("Mounting loop device %s to %s\n", path, mnt.Destination)
-	_, err = c.rpcOps.Mount(path, mnt.Destination, mnt.Type, flags, optsString)
+
+	sylog.Debugf("Mounting loop device %s to %s of type %s\n", path, mnt.Destination, mnt.Type)
+
+	mountType := mnt.Type
+
+	if mnt.Type == "encryptfs" {
```
cclerget:
```suggestion
	if mountType == "encryptfs" {
```

internal/pkg/runtime/engines/singularity/rpc/client/client.go
```diff
@@ -33,6 +34,21 @@ func (t *RPC) Mount(source string, target string, filesystem string, flags uintp
 	return reply, err
 }
 
+// Decrypt calls the DeCrypt RPC using the supplied arguments.
+func (t *RPC) Decrypt(offset uint64, path string) (string, error) {
+	sp := strings.Split(path, "/")
```
cclerget: same `loopdev := filepath.Base(path)`

pkg/build/types/bundle.go
```diff
@@ -88,13 +90,28 @@ func NewBundle(bundleDir, bundlePrefix string) (b *Bundle, err error) {
 		"rootfs": "fs",
 	}
 
+	if encrypted {
```
cclerget: `b.Opts.Encrypted = encrypted` ?

pkg/util/crypt/crypt_dev.go
```diff
@@ -0,0 +1,306 @@
...
+func (crypt *Device) CloseCryptDevice(path string) error {
+
+	cmd := exec.Command("/sbin/cryptsetup", "luksClose", path)
+	cmd.SysProcAttr = &syscall.SysProcAttr{}
+	cmd.SysProcAttr.Credential = &syscall.Credential{Uid: 0, Gid: 0}
+	fd, err := lock.Exclusive("/dev/mapper")
+	if err != nil {
+		return err
+	}
+	err = cmd.Run()
```
cclerget, line 59: just before this line, I would call `defer lock.Release(fd)` in order to release lock if an error occur during `cmd.Run()`

pkg/util/crypt/crypt_dev.go
```diff
@@ -0,0 +1,306 @@
...
+	if err != nil {
+		sylog.Debugf("Unable to delete the crypt device %s", err)
+		return err
+	}
+	err = lock.Release(fd)
```
cclerget: previous comment

pkg/util/crypt/crypt_dev.go
```diff
@@ -0,0 +1,306 @@
...
+
+	sp := strings.Split(loop, "/")
+	loopdev := sp[len(sp)-1]
+
+	fd, err := lock.Exclusive("/dev/mapper")
+	if err != nil {
+		sylog.Debugf("Unable to acquire lock on /dev/mapper")
+		return "", "", err
+	}
+	nextCrypt := getNextAvailableCryptDevice()
```
cclerget, line 171: call `defer lock.Release(fd)` here

pkg/util/crypt/crypt_dev.go
cclerget: `loopdev := filepath.Base(loop)`

pkg/util/crypt/crypt_dev.go
```diff
@@ -0,0 +1,306 @@
...
+	b := make([]byte, n)
+	if _, err := rand.Read(b); err != nil {
+		sylog.Debugf("Something went wrong while generating random string")
+		return ""
+	}
+	return fmt.Sprintf("%x", b)
+}
+
+func getNextAvailableCryptDevice() string {
+	return getRandomString(15)
```
cclerget: or you can import `github.com/satori/go.uuid` and use `uuid.NewV4()`

pkg/util/crypt/crypt_dev.go
```diff
@@ -0,0 +1,306 @@
...
+		return "", "", err
+	}
+
+	err = lock.Release(fd)
+	if err != nil {
+		sylog.Debugf("Unable to release lock on /dev/mapper")
+		return "", "", err
+	}
+
+	copyDeviceContents(path, "/dev/mapper/"+nextCrypt, squashFsSize)
```
cclerget: eligible for `filepath.Join("/dev/mapper", nextCrypt)`

pkg/util/crypt/crypt_dev.go
```diff
@@ -0,0 +1,306 @@
...
+
+	// Open a new Temp file to stash the loop contents
+	loopF, err := ioutil.TempFile("", "loop-")
+	if err != nil {
+		sylog.Debugf("Error creating temporary crypt file")
+		return "", "", err
+	}
+
+	copyDeviceContents("/dev/"+loopdev, loopF.Name(), size)
```
cclerget: eligible for `filepath.Join("/dev", loopdev)`

vendor/github.com/sylabs/sif/pkg/sif/fmt.go
cclerget: need to revert this change

vendor/github.com/sylabs/sif/pkg/sif/sif.go
cclerget: same

vendor/github.com/sylabs/sif/pkg/sif/sif.go
cclerget: same

pkg/util/crypt/crypt_dev.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	go func() {
+		defer stdin.Close()
+		io.WriteString(stdin, string(password))
+	}()
+
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		if strings.Contains(string(out), "No key available") {
```
cclerget, line 279: Just a question, is cryptsetup return different exit code which could help to determine the error ? Relying on error output could have side effects between cryptsetup version, `LANG` environment variable can alter error output and return error message in another language, ok it's not actually the case because this function is only called by RPC server which run with only one environment variable, but if this function is called directly by build command or something else, that's something to consider.

cclerget (changes requested):

internal/pkg/runtime/engines/imgbuild/create_linux.go
schebro: Oops. Thought I added that. I should have checked the final diff. Will restore.

internal/pkg/runtime/engines/singularity/cleanup_linux.go
schebro: Will fix that. Thanks for pointing out. I didn't know the order of defers.

internal/pkg/runtime/engines/singularity/cleanup_linux.go
schebro: Cool. Will fix.

vendor/github.com/sylabs/sif/pkg/sif/fmt.go
schebro: Yes. I will revert the vendor changes once the PR in SIF repo is merged.

pkg/util/crypt/crypt_dev.go
schebro: cryptsetup error codes are not as descriptive as the output. That's why I was checking for the output to know what exactly the error is. Here are cryptsetup error codes.

0 - success, otherwise fail
1 - wrong parameters
2 - no permission
3 - out of memory
4 - wrong device specified
5 - device already exists or device is busy

If the password is wrong or for any other wrong parameter, the error code would be 1. For the latter case where the device exists, the code would be 5. I can replace the second string compare by checking the error code for 5 but error code 1 is not very descriptive.

cmd/internal/cli/build_linux.go
```diff
@@ -23,6 +23,7 @@ import (
 )
 
 func run(cmd *cobra.Command, args []string) {
+
```
WestleyK: Why the extra newline?

```suggestion
```

internal/pkg/build/assemblers/assembler_sif.go
```diff
@@ -25,13 +25,15 @@ import (
...
+	"github.com/sylabs/singularity/pkg/util/crypt"
 )
 
 // SIFAssembler doesnt store anything
 type SIFAssembler struct {
 }
 
-func createSIF(path string, definition, ociConf []byte, squashfile string) (err error) {
+func createSIF(path string, definition, ociConf []byte, squashfile string, encrypted bool) (err error) {
+
```
WestleyK: Extra newline here too?

```suggestion
```

internal/pkg/build/assemblers/assembler_sif.go
```diff
@@ -177,7 +185,28 @@ func (a *SIFAssembler) Assemble(b *types.Bundle, path string) (err error) {
 		return fmt.Errorf("While running mksquashfs: %v: %s", err, strings.Replace(string(errOut), "\n", " ", -1))
 	}
 
-	err = createSIF(path, b.Recipe.Raw, b.JSONObjects["oci-config"], squashfsPath)
+	if b.Opts.Encrypted {
+
```
WestleyK: And here? Are these necessary?

internal/pkg/runtime/engines/singularity/cleanup_linux.go
```diff
@@ -58,5 +63,47 @@ func (engine *EngineOperations) CleanupContainer(fatal error, status syscall.Wai
...
+
+	// Elevate the privilege to unmount and delete the crypt device
+	runtime.LockOSThread()
+	uid := os.Getuid()
+	err := syscall.Setresuid(uid, 0, uid)
+	defer syscall.Setresuid(uid, uid, 0)
+	defer runtime.UnlockOSThread()
+	if err != nil {
+		sylog.Debugf("Err setting suid")
+		return err
```
WestleyK: I think it would also be good to add more debugging in the error message:

```suggestion
		return fmt.Errorf("unable to setup suid: %v", err)
```

_Or something like that..._

internal/pkg/runtime/engines/singularity/cleanup_linux.go
WestleyK: Should that message also be returned in the `error`? Like:

```go
return fmt.Errorf("failed unmounting final session directory: %s", err)
```

internal/pkg/runtime/engines/singularity/container_linux.go
```diff
@@ -571,10 +570,29 @@ func (c *container) mountImage(mnt *mount.Point) error {
...
+		cryptDev, err := c.rpcOps.Decrypt(offset, path)
+
+		if err != nil {
+			return fmt.Errorf("Unable to decrypt the file system: %s", err)
```
WestleyK, line 611: Its better to not use caps in error messages

```suggestion
			return fmt.Errorf("unable to decrypt the file system: %s", err)
```

internal/pkg/runtime/engines/singularity/rpc/server/server_linux.go
```diff
@@ -13,6 +13,8 @@ import (
 	"strings"
 	"syscall"
 
+	"github.com/sylabs/singularity/pkg/util/crypt"
+
```
WestleyK: Extra newline here

```suggestion
```

internal/pkg/runtime/engines/singularity/rpc/server/server_linux.go
```diff
@@ -33,6 +35,18 @@ func (t *Methods) Mount(arguments *args.MountArgs, reply *int) (err error) {
 	return err
 }
 
+// DeCrypt decrypts the loop device by using dm-crypt interface
+func (t *Methods) Decrypt(arguments *args.CryptArgs, reply *string) (err error) {
+
```
WestleyK: Extra newline here too

```suggestion
```

pkg/util/crypt/crypt_dev.go
```diff
@@ -0,0 +1,306 @@
...
+	"strings"
+	"syscall"
+
+	"github.com/sylabs/singularity/internal/pkg/sylog"
+	"github.com/sylabs/singularity/pkg/util/fs/lock"
+	"github.com/sylabs/singularity/pkg/util/loop"
+	"golang.org/x/crypto/ssh/terminal"
+)
+
+// Device is
```
WestleyK: Device is what? Is there more that needs to go here?

pkg/util/crypt/crypt_dev.go
```diff
@@ -0,0 +1,306 @@
...
+	fmt.Println()
+
+	if input != input2 {
+		return "", "", errors.New("Passwords don't match")
+	}
+
+	squashF, err := os.Stat(path)
+	if err != nil {
+		sylog.Debugf("Error acquiring size of %s", path)
+		return "", "", err
```
WestleyK: I think this should also return some debugging like the above line

```suggestion
		return "", "", fmt.Errorf("failed getting size of: %s: ", path, err)
```

pkg/util/crypt/crypt_dev.go
```diff
@@ -0,0 +1,306 @@
...
+
+	go func() {
+		defer stdin.Close()
+		io.WriteString(stdin, input)
+	}()
+
+	err = cmd.Run()
+	if err != nil {
+		sylog.Verbosef("Unable to format crypt device")
+		return "", "", err
```
WestleyK: Should also return some debugging to?

pkg/util/crypt/crypt_dev.go
```diff
@@ -0,0 +1,306 @@
...
+
+	copyDeviceContents("/dev/"+loopdev, loopF.Name(), size)
+
+	return loopF.Name(), nextCrypt, err
+}
+
+// copyDeviceContents copies the contents of source to destination.
+// source and dest can either be a file or a block device
+func copyDeviceContents(source string, dest string, size int64) error {
+
```
WestleyK: Extra newline

pkg/util/crypt/crypt_dev.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	copyDeviceContents("/dev/"+loopdev, loopF.Name(), size)
+
+	return loopF.Name(), nextCrypt, err
+}
+
+// copyDeviceContents copies the contents of source to destination.
+// source and dest can either be a file or a block device
+func copyDeviceContents(source string, dest string, size int64) error {
```
WestleyK: Not necessary, but may look nicer...

```suggestion
func copyDeviceContents(source, dest string, size int64) error {
```

pkg/util/crypt/crypt_dev.go
```diff
@@ -0,0 +1,306 @@
...
+
+	squashFd, err := syscall.Open(source, syscall.O_RDWR, 0777)
+	if err != nil {
+		sylog.Debugf("Unable to open the file %s", source)
+		return err
+	}
+	defer syscall.Close(squashFd)
+
+	cryptFd, err := syscall.Open(dest, syscall.O_RDWR, 0777)
+
```
WestleyK, line 212: Extra newline?

pkg/util/crypt/crypt_dev.go
```diff
@@ -0,0 +1,306 @@
...
+	if err != nil {
+		sylog.Debugf("Unable to open the file: %s", dest)
+		return err
```
WestleyK: Again, this should probably return some debugging with the error message:

```suggestion
		return fmt.Errorf("unable to open file: %s: %v", dest, err)
```

pkg/util/crypt/crypt_dev.go
```diff
@@ -0,0 +1,306 @@
...
+}
+
+func getNextAvailableCryptDevice() string {
+	return getRandomString(15)
+}
+
+// GetCryptDevice returns the next available device in /dev/mapper for encryption/decryption
+func (crypt *Device) GetCryptDevice(loopDev string) (string, error) {
+
+	fmt.Print("Enter the password to decrypt the File System: ")
```
WestleyK: Maybe there should also be a argument to pass the password?

WestleyK (commented):
Looks good, just a couple suggestion that may help :+1:

cmd/internal/cli/build_linux.go
schebro: Sorry. Not intentional. I was relying on vscode to format the extra lines but apparently it doesn't. Will clean up. Thanks for pointing out.

WestleyK: Sure! No problem :+1:

GodloveD:
Wow!  113 comments and counting!!

Let's really try to get this merged as a top priority so that we can use it as the basis for more work toward that goal.

schebro:
Most of these are already addressed in the latest version. I am going to
open a separate issue for keyfile support. I don't want it to delay
existing changes. Will push other changes tomorrow. Let's get it merged.

pkg/util/crypt/crypt_dev.go
schebro: I will change this behavior as part of a separate issue. There need to be some changes both on the encryption and decryption side w.r.t providing extra arguments(key-file) to the command.

pkg/util/crypt/crypt_dev.go
schebro: Changed this to filepath.Base

internal/pkg/build/assemblers/assembler_sif.go
schebro: Not necessary. Coming from C background, I am used to use new lines liberally for readability. Will remove

internal/pkg/runtime/engines/singularity/rpc/server/server_linux.go
schebro: I created #3851 to supply encryption keys non-interactively.

pkg/util/crypt/crypt_dev.go
schebro: Created #3851 to address it. Passing secret as argument requires some changes in encryption and decryption path such as supplying a --key-file from the command line. Instead of holding this PR up, I prefer to address it as a separate issue if that's fine.

pkg/util/crypt/crypt_dev.go
schebro: #3851

pkg/util/crypt/crypt_dev.go
schebro: Used that instead

internal/pkg/runtime/engines/singularity/cleanup_linux.go
```diff
@@ -6,12 +6,17 @@
 package singularity
```
WestleyK, line 6: Missing copyright thing?

```

package singularity
```

_EDIT: not necessary, it already exist. my bad_

WestleyK (commented):
Have one more suggestion, not necessary, we can fix this in a followup PR.

internal/pkg/runtime/engines/singularity/cleanup_linux.go
WestleyK: Sorry, my bad. There already one, but it should be `2018-2019`. (not `2018`)

cclerget (approved):

mem (changes requested):
I still think the reading of the passphrase is happening too late and too deep in the code (this is related to but not the same as not reading the passphrase interactively). I also think this would be better done in two steps: create the unencrypted SIF file and later creating a second SIF that's encrypted. That would allow us to introduce an "encrypt" operation for existing SIF files. The more important consequence of that is that you can split the encryption code from the squashfs creation code. An issue should be opened for that.

Given the above, before merging this, let's do one last thing: all the opened issues have a counterpart in this review (e.g. reading the passphrase from the terminal is happening in a couple specific locations).

I would strongly prefer to annotate the _source code_ with the corresponding issue numbers. For example:

```
// TODO(schebro): fix issue #3818 
// We expect the cryptsetup utility to be installed in the host.
// We need to add checks in mconfig for that purpose and we need
// to provide a configuration variable for the sysadmin to specify
// its location.
```

In order words, please add internal documentation that documents what the issue is, both as a reference to the existing issue and with enough description for someone else to understand what the problem is. This will allow us to split the effort if necessary, and not miss any changes related to those issues.

schebro:
> I still think the reading of the passphrase is happening too late and too deep in the code (this is related to but not the same as not reading the passphrase interactively). I also think this would be better done in two steps: create the unencrypted SIF file and later creating a second SIF that's encrypted. That would allow us to introduce an "encrypt" operation for existing SIF files. The more important consequence of that is that you can split the encryption code from the squashfs creation code. An issue should be opened for that.
> 
> Given the above, before merging this, let's do one last thing: all the opened issues have a counterpart in this review (e.g. reading the passphrase from the terminal is happening in a couple specific locations).
> 
> I would strongly prefer to annotate the _source code_ with the corresponding issue numbers. For example:
> 
> ```
> // TODO(schebro): fix issue #3818 
> // We expect the cryptsetup utility to be installed in the host.
> // We need to add checks in mconfig for that purpose and we need
> // to provide a configuration variable for the sysadmin to specify
> // its location.
> ```
> 
> In order words, please add internal documentation that documents what the issue is, both as a reference to the existing issue and with enough description for someone else to understand what the problem is. This will allow us to split the effort if necessary, and not miss any changes related to those issues.

I will annotate the code with necessary comments and issues. I agree that the key reading can happen much earlier. The method of key reading is going to change in the coming weeks anyway. I will address it then.

internal/pkg/runtime/engines/singularity/cleanup_linux.go
schebro: Fixed

mem (approved):
thanks, let's go with this.

PR #3614, opened 2019-05-24
Fixed panic when pulling a container that does not exist

WestleyK:
## Description of the Pull Request (PR):

This PR fixes a panic when pulling a image that does not exist.

Example:

```bash
# Before this PR:
$ singularity pull foobar
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xb8 pc=0xf75540]

goroutine 1 [running]:
github.com/sylabs/singularity/internal/app/singularity.LibraryPull(0xc000226520, 0x11, 0x7ffc9b8ea25c, 0x6, 0x0, 0x0, 0x7ffc9b8ea25c, 0x6, 0xc000388740, 0x19, ...)
[...]

# After this PR:
$ singularity pull library://foobar
FATAL:   While pulling library image: image does not exist in the library
```

<br>

PR #3736, opened 2019-06-14
Fixed panic when exec/run a container in the library that does not exist

### LINKED ISSUE #3735: segfault on library container that doesn't exist
## Version of Singularity:

3.2.1-660.g4c8a84050

### Expected behavior
```
$ singularity exec library://no-container true
FATAL:   Unable to handle library://no-container uri: failed to find library://no-container: Error container does not exist.
I'm sorry Dave, I'm afraid I can't do that...
```
(Or something along those lines.)
### Actual behavior

```
$ singularity exec library://no-container true
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xb0 pc=0xff9fd9]

goroutine 1 [running]:
github.com/sylabs/singularity/cmd/internal/cli.handleLibrary(0x7ffe4d2e03a1, 0x16, 0xc0003a0140, 0x19, 0x7ffe4d2e03a9, 0xe, 0xc3, 0x30)
	cmd/internal/cli/actions.go:149 +0x2f9
github.com/sylabs/singularity/cmd/internal/cli.replaceURIWithImage(0x1c18ca0, 0xc00000eda0, 0x2, 0x2)
	cmd/internal/cli/actions.go:227 +0x3ff
github.com/sylabs/singularity/cmd/internal/cli.actionPreRun(0x1c18ca0, 0xc00000eda0, 0x2, 0x2)
	cmd/internal/cli/actions.go:43 +0x125
github.com/spf13/cobra.(*Command).execute(0x1c18ca0, 0xc0000320e0, 0x2, 0x2, 0x1c18ca0, 0xc0000320e0)
	vendor/github.com/spf13/cobra/command.go:755 +0x4e2
github.com/spf13/cobra.(*Command).ExecuteC(0x1c21fe0, 0x4, 0xc0004f4240, 0x27)
	vendor/github.com/spf13/cobra/command.go:850 +0x2bb
github.com/sylabs/singularity/cmd/internal/cli.ExecuteSingularity()
	cmd/internal/cli/singularity.go:246 +0x13c
main.main()
	cmd/singularity/cli.go:17 +0x20
```
ewww

```
$ singularity exec library://no-container true
```

WestleyK:

This PR fixes the panic when using actions command on a library image that does not exist.

- Fixes #3735 
- Related to #3614

### DIFF
--- a/cmd/internal/cli/actions.go
+++ b/cmd/internal/cli/actions.go
@@ -140,9 +140,16 @@ func handleLibrary(u, libraryURL string) (string, error) {
 
 	imageRef := libraryhelper.NormalizeLibraryRef(u)
 
-	libraryImage, _, err := c.GetImage(ctx, imageRef)
+	libraryImage, existOk, err := c.GetImage(ctx, imageRef)
 	if err != nil {
 		return "", err
+	}
+	if !existOk {
+		return "", fmt.Errorf("image does not exist in the library: %s", imageRef)
+	}
+	if libraryImage == nil {
+		// It should never get here
+		return "", fmt.Errorf("failed getting image from the library: %s", imageRef)
 	}
 
 	imageName := uri.GetName("library://" + imageRef)
--- a/cmd/internal/cli/actions.go
+++ b/cmd/internal/cli/actions.go
@@ -147,10 +147,6 @@ func handleLibrary(u, libraryURL string) (string, error) {
 	if !existOk {
 		return "", fmt.Errorf("image does not exist in the library: %s", imageRef)
 	}
-	if libraryImage == nil {
-		// It should never get here
-		return "", fmt.Errorf("failed getting image from the library: %s", imageRef)
-	}
 
 	imageName := uri.GetName("library://" + imageRef)
 	imagePath := cache.LibraryImage(libraryImage.Hash, imageName)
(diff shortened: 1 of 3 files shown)


cmd/internal/cli/actions.go
```diff
@@ -140,10 +140,17 @@ func handleLibrary(u, libraryURL string) (string, error) {
 
 	imageRef := libraryhelper.NormalizeLibraryRef(u)
 
-	libraryImage, _, err := c.GetImage(ctx, imageRef)
+	libraryImage, existOk, err := c.GetImage(ctx, imageRef)
 	if err != nil {
 		return "", err
 	}
+	if !existOk {
```
mem, line 147: this is fine, but why is a non-existent / failure to get an image giving err == nil?

Skimming thru the code either err == nil and found == true, in which case libraryImage will have _something_, or err == nil and found == false, and libraryImage will be nil.

_I would prefer_ to not have the extra `if libraryImage == nil` check, as it makes reading the code a little harder, and as the comment states, that path is never hit.

It's interesting that the code goes thru several layers of handling "not exist" as an error or not. At some level, not found is in fact a non-nil error, which gets translated to a nil error one layer above it. It would be more idiomatic to report not found as an error.

mem (commented):
Please let me know what you want to do with this.

cmd/internal/cli/actions.go
WestleyK: Okay, I'll remove the `if libraryImage == nil {...}` check :+1:

ikaneshiro (approved):
lgtm

PR #3737, opened 2019-06-14
E2E testing for 'inspect'

### LINKED ISSUE #2121: Add basic `singularity inspect` test

WestleyK:

This PR adds E2E tests for `inspect` command.

- Fixes #2121 

<br>


e2e/inspect/inspect.go
```diff
@@ -0,0 +1,182 @@
...
+		"apps": "",
+		"environment": {
+			"90-environment.sh": "#!/bin/sh\n#Custom environment shell code should follow\n\n\nexport test=\"testing\"\nexport e2e=\"e2e testing\"\n\n\n"
+		}
+	},
+	"type": "container"
+}`
+
+func runInspectCommand(inspectType string) ([]byte, error) {
+	argv := []string{"inspect", "--json", inspectType, "testdata/test.sif"}
```
mem: the name "test.sif" is too generic. If this file is specific to this set of tests, its name should reflect that.

Also, if there's something special about this file, instructions as to how to recreate it should be included. Alternatively the file should be created as part of the test. Or use the standard test file provided in the environment variable E2E_IMAGE_PATH.

e2e/inspect/inspect.go
```diff
@@ -0,0 +1,182 @@
...
+	cmd := exec.Command("singularity", argv...)
+
+	return cmd.CombinedOutput()
+}
+
+func singularityInspect(t *testing.T) {
+	tests := []struct {
+		name      string
+		insType   string
+		json      []string
```
mem: can you please add comments documenting what these fields represent. json and expectedOut in particular are not clear to me.

Reading thru the code, `json` seems to be the path into the JSON output from singularity inspect, starting from the outermost object. `expectOut` seems to be a dump of the whole output. This seems strange. I would have expected to see the value of the object specified by `json`.

e2e/inspect/inspect.go
```diff
@@ -0,0 +1,182 @@
...
+			if err != nil {
+				t.Fatalf("Unexpected failure: %s: %s", string(out), err)
+			}
+
+			// Check the E2E label in test.sif, does it match our expected output
+			v, err := jsonparser.GetString(out, tt.json...)
+			if err != nil {
+				fmt.Println("ERROR: ", err)
+			}
+			// Get the expected output, and compair them
```
mem:
```suggestion
			// Get the expected output, and compare them
```

e2e/inspect/inspect.go
```diff
@@ -0,0 +1,182 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, test.WithPrivilege(func(t *testing.T) {
+			out, err := runInspectCommand(tt.insType)
+			if err != nil {
+				t.Fatalf("Unexpected failure: %s: %s", string(out), err)
+			}
+
+			// Check the E2E label in test.sif, does it match our expected output
+			v, err := jsonparser.GetString(out, tt.json...)
```
mem, line 129: this bit here is extracting the object specified by tt.json from out. That's fine.

e2e/inspect/inspect.go
```diff
@@ -0,0 +1,182 @@
...
+			if err != nil {
+				fmt.Println("ERROR: ", err)
+			}
+			// Get the expected output, and compair them
+			e, err := jsonparser.GetString([]byte(tt.expectOut), tt.json...)
```
mem: but this bit here is extracting the same object from the stored dump. This seems strange. This is where I would have expected to compare v above with a specific value.

e2e/inspect/inspect.go
```diff
@@ -0,0 +1,182 @@
...
+		t.Run(tt.name, test.WithPrivilege(func(t *testing.T) {
+			out, err := runInspectCommand(tt.insType)
+			if err != nil {
+				t.Fatalf("Unexpected failure: %s: %s", string(out), err)
+			}
+
+			// Check the E2E label in test.sif, does it match our expected output
+			v, err := jsonparser.GetString(out, tt.json...)
+			if err != nil {
+				fmt.Println("ERROR: ", err)
```
mem: this should mark the test as failed and bail out (of the subtest), not print the error and continue.

e2e/inspect/inspect.go
```diff
@@ -0,0 +1,182 @@
...
+		}))
+	}
+
+}
+
+// RunE2ETests is the main func to trigger the test suite
+func RunE2ETests(t *testing.T) {
+	e2e.LoadEnv(t, &testenv)
+
+	// try to build from a non existen path
```
mem: I don't understand what this comment refers to. The next line is executing the inspect test.

e2e/inspect/inspect.go
```diff
@@ -0,0 +1,182 @@
...
+	t.Run("singularityEnv", singularityInspect)
```
mem: "singularityEnv" seems a little too generic.

e2e/suite.go
```diff
@@ -16,13 +16,15 @@ import (
 	"syscall"
 	"testing"
 
+	singularityinspect "github.com/sylabs/singularity/e2e/inspect"
```
mem, line 19: this is fine, but why do you need to use a name other than "inspect"?

e2e/suite.go
```diff
@@ -92,11 +94,15 @@ func Run(t *testing.T) {
...
-	singularitye2e.PrepRegistry(t)
-	defer singularitye2e.KillRegistry(t)
+	//	singularitye2e.PrepRegistry(t)
+	//	defer singularitye2e.KillRegistry(t)
 
 	// RunE2ETests by functionality
 
+	t.Run("INSPECT", singularityinspect.RunE2ETests)
+
+	return
```
mem: accident, I guess?

mem (changes requested):

e2e/inspect/inspect.go
WestleyK: Probably a copy-paste error, I'll remove it :+1:

e2e/inspect/inspect.go
WestleyK: Also copy-paste error, should be 

```go
t.Run("singularityInspect", singularityInspect)
```

e2e/suite.go
WestleyK: Go complained about it if I didn't have `singularityinspect`, not really sure why...

e2e/suite.go
WestleyK: Yeah, I just did that to "disable" the other tests, so its quicker :+1:

e2e/inspect/inspect.go
WestleyK: This is just getting the expected out (`tt.expectOut`). Then is comparing it on line ~177

WestleyK: I guess we _could_ have a parameter in the table tests setting the expected output?

e2e/suite.go
WestleyK: I think I figured out why Go did not like it before, but I'll leave it as it is.

e2e/inspect/inspect.go
```diff
@@ -0,0 +1,215 @@
...
+			json:      []string{"attributes", "environment", "90-environment.sh"},
+			expectOut: expectedEnvironmentJSON,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, test.WithPrivilege(func(t *testing.T) {
+			out, err := runInspectCommand(tt.insType)
+			if err != nil {
+				t.Fatalf("Unexpected failure: %s: %s", string(out), err)
```
gvallee:
```suggestion
				t.Fatalf("unexpected failure: %s: %s", string(out), err)
```

e2e/inspect/inspect.go
```diff
@@ -0,0 +1,215 @@
...
+			}
+
+			// Check the E2E label in test.sif, does it match our expected output
+			v, err := jsonparser.GetString(out, tt.json...)
+			if err != nil {
+				t.Fatalf("Unable to get expected output from json: %v", err)
```
gvallee:
```suggestion
				t.Fatalf("unable to get expected output from json: %v", err)
```

e2e/inspect/inspect.go
```diff
@@ -0,0 +1,215 @@
...
+			}
+			// Get the expected output, and compare them
+			e, err := jsonparser.GetString([]byte(tt.expectOut), tt.json...)
+			if err != nil {
+				t.Fatalf("Unable to get expected output from json: %v", err)
```
gvallee:
```suggestion
				t.Fatalf("unable to get expected output from json: %v", err)
```

e2e/inspect/inspect.go
```diff
@@ -0,0 +1,215 @@
...
+			}
+			if v != e {
+				t.Fatalf("Unexpected faulure: got: %s, expecting: %s", v, e)
```
gvallee:
```suggestion
				t.Fatalf("unexpected failure: got: %s, expecting: %s", v, e)
```

gvallee (changes requested):
LGTM, I only have a couple of minor suggestions.

e2e/inspect/inspect.go
```diff
@@ -0,0 +1,215 @@
...
+		{
+			name:      "environment",
+			insType:   "--environment",
+			json:      []string{"attributes", "environment", "90-environment.sh"},
+			expectOut: expectedEnvironmentJSON,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, test.WithPrivilege(func(t *testing.T) {
```
gvallee: If I understand correctly, this test is supposed to only run in privileged mode so I would suggest to add `test.EnsurePrivilege(t)` at the beginning of the test rather than escalating privilege for each sub-test

gvallee (changes requested):

e2e/inspect/inspect.go
WestleyK: Actually it should probably be `test.WithoutPrivilege`. Not sure how it changed :confused:

e2e/inspect/inspect.go
```diff
@@ -3,13 +3,9 @@
 // LICENSE.md file distributed with the sources of this project regarding your
 // rights to use or distribute this software.
 
-// This test sets singularity image specific environment variables and
-// verifies that they are properly set.
-
-package singularityenv
+package singularityinspect
```
mem: usually packages are named after the directory they live in, this might be why you need a named import elsewhere. Since this file lives in a directory called "inspect" the expectation is that the package is also called "inspect".

e2e/inspect/inspect.go
```diff
@@ -25,16 +21,19 @@ type testingEnv struct {
...
+const containerTesterSIF = "testdata/inspecter_container.sif"
+
+const expectedLabelsJSON = `
 {
 	"attributes": {
 		"apps": "",
 		"labels": {
 			"E2E": "AWSOME",
 			"HI": "\"HELLO WORLD\"",
+			"MAINTAINER": "\"WestleyK \u003cwestley@sylabs.io\u003e\"",
```
mem: 0x3c and 0x3e are `<` and `>` respectively.

It should be possible to use those characters instead of unicode code points.

Also, if you don't want to escape the quotes, you can use backticks. ``"MAINTAINER": `"Westely ..."`,``

e2e/inspect/inspect.go
mem: the bit that is strange about this is that you have a whole JSON object stored in `expectOut`, but you are looking at a specific value within that object for the test.

I think I know what's bugging me here.

The code is checking that each one of the labels is correct, and it has one test case per label. At a higher level, what you want to do is check that all the labels are present, and that their values are correct. That's what you have in `expectedLabelsJSON`, the whole set of expected labels.

The problem with comparing JSON object as strings is that nothing says their order should be the same. For example `{"foo": 1, "bar": 2}` is the same object as `{"bar": 2, "foo": 1}`, even if those are different strings.

If you want two compare two JSON objects represented as strings, you can do [this](https://play.golang.org/p/zD81n-nX7uI). In this way you can have a single test case for each `--labels`, `--runscript`, etc.

mem (changes requested):
LGTM. I added a suggestion as to how to make the test simpler.

e2e/inspect/inspect.go
WestleyK: This output is unmodified output from `singularity inspect inspecter_container.sif`, I did not change anything.

And yes, it does work if I change `\u003c` to `<`.

e2e/inspect/inspect.go
WestleyK: I'll go back to your first suggestion, just adding the expected output in the table.

mem (approved):

e2e/inspect/inspect.go
```diff
@@ -0,0 +1,148 @@
...
+			if err != nil {
+				t.Fatalf("unexpected failure: %s: %s", string(out), err)
+			}
+
+			// Parse the output
+			v, err := jsonparser.GetString(out, tt.json...)
+			if err != nil {
+				t.Fatalf("unable to get expected output from json: %v", err)
+			}
+			// Compair the output, with the expected output
```
gvallee:
```suggestion
			// Compare the output, with the expected output
```

gvallee (changes requested):
Found a typo otherwise lgtm

gvallee (approved):

PR #3749, opened 2019-06-18
Update help docs to include oras transport

### LINKED ISSUE #3767: No help for Singularity ORAS
## Version of Singularity:

3.3.0-rc1

### Expected behavior

A help doc for `singularity oras --help`

ikaneshiro:

Fixes: #3767

### DIFF
--- a/docs/content.go
+++ b/docs/content.go
@@ -60,7 +60,8 @@ Enterprise Performance Computing (EPC)`
 
       library://  an image library (default https://cloud.sylabs.io/library)
       docker://   a Docker registry (default Docker Hub)
-      shub://     a Singularity registry (default Singularity Hub)`
+      shub://     a Singularity registry (default Singularity Hub)
+      oras://     a supporting OCI registry`
 
 	BuildExample string = `
 
@@ -441,7 +442,9 @@ Enterprise Performance Computing (EPC)`
 
   docker://*          A container hosted on Docker Hub
 
-  shub://*            A container hosted on Singularity Hub`
+  shub://*            A container hosted on Singularity Hub
+
+  oras://*            A container hosted on a supporting OCI registry`
 	ExecUse   string = `exec [exec options...] <container> <command>`
 	ExecShort string = `Run a command within a container`
 	ExecLong  string = `
@@ -556,7 +559,10 @@ Enterprise Performance Computing (EPC)`
       docker://user/image:tag
     
   shub: Pull an image from Singularity Hub to CWD
-      shub://user/image:tag`
+      shub://user/image:tag
+
+  oras: Pull a SIF image from a supporting OCI registry
+      oras://registry/namespace/image:tag`
 	PullExample string = `
   From Sylabs cloud library
   $ singularity pull alpine.sif library://alpine:latest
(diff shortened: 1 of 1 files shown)


docs/content.go
```diff
@@ -565,20 +571,36 @@ Enterprise Performance Computing (EPC)`
...
+  The 'push' command allows you to upload a SIF container to a given
+  URI.  Supported URIs include:
+
+  library:
+      library://user/collection/container[:tag]
+
+  oras:
+      oras://registry/namespace/repo:tag
+
+
```
WestleyK, line 594: Extra newline here? Is this supposed to be here?

WestleyK (approved):
Looks good to me, just one comment about a "extra" newline :+1:

docs/content.go
ikaneshiro: That was intentional, I wanted some space before the note, but if you think it would look better without I can remove it

WestleyK: Imo, its looks better without the extra newline, but I'm cool either way.

docs/content.go
```diff
@@ -60,7 +60,8 @@ Enterprise Performance Computing (EPC)`
 
       library://  an image library (default https://cloud.sylabs.io/library)
       docker://   a Docker registry (default Docker Hub)
-      shub://     a Singularity registry (default Singularity Hub)`
+      shub://     a Singularity registry (default Singularity Hub)
+      oras://     a supporting OCI registry (no default)`
```
jscook2345: Remove `(no default)`

docs/content.go
```diff
@@ -556,7 +559,10 @@ Enterprise Performance Computing (EPC)`
       docker://user/image:tag
     
   shub: Pull an image from Singularity Hub to CWD
-      shub://user/image:tag`
+      shub://user/image:tag
+
+  oras: Pull an SIF image from a supporting OCI registry
```
jscook2345: Change `Pull an SIF` to `Pull a SIF`.

docs/content.go
```diff
@@ -556,7 +559,10 @@ Enterprise Performance Computing (EPC)`
...
+      oras://registry/namespace/repo:tag`
```
jscook2345: Should `repo` be `image` instead?

docs/content.go
```diff
@@ -565,20 +571,36 @@ Enterprise Performance Computing (EPC)`
   $ singularity pull tensorflow.sif docker://tensorflow/tensorflow:latest
 
   From Shub
-  $ singularity pull singularity-images.sif shub://vsoch/singularity-images`
+  $ singularity pull singularity-images.sif shub://vsoch/singularity-images
+
+  From supporting OCI registry (e.g. Azure Container Registry)
+  $ singularity pull image.sif oras://<username>.azurecr.io/namespace/repo:tag`
```
jscook2345: #1. Can we put a real example here?
#2. `repo` vs `image` again.

docs/content.go
```diff
@@ -565,20 +571,36 @@ Enterprise Performance Computing (EPC)`
...
+  NOTE: It's always good practice to sign your containers before
+  pushing them to the library. An auth token is required to push to the library,
+  so you may need to configure it first with 'singularity remote'.`
 	PushExample string = `
-  $ singularity push /home/user/my.sif library://user/collection/my.sif:latest`
+  To Library
+  $ singularity push /home/user/my.sif library://user/collection/my.sif:latest
+
+  To supported OCI registry
+  $ singularity push /home/user/my.sif oras://registry/namespace/repo:tag`
```
jscook2345: `repo` vs `image` again

jscook2345 (changes requested):
I put some comments in for review. Ping me when you need a re-review! Thanks!

docs/content.go
ikaneshiro: The nomenclature I've seen more is to call that a repo and then the image is the repo plus tag.

i.e. https://hub.docker.com/_/golang, the repo is `golang` and then you specify which image you want based on the tag

docs/content.go
ikaneshiro: 1. If we want a public acr registry for people to pull from to try out, then we would need to set that up ourselves. I tested using a trial and am not planning on making the registry public or extending it once it expires.

2. See above.

ikaneshiro:
@jscook2345 ping!

jscook2345 (approved):
LGTM

PR #3758, opened 2019-06-19
Revisit the implementation of image caches 

### LINKED ISSUE #3525: Clarify the semantics behind the image cache

### LINKED ISSUE #3557: make testall fails because of library_test.go
## Version of Singularity:

Master on Ubuntu 18.10.

### Expected behavior

`make testall` should successfully pass.

### Actual behavior

```
--- FAIL: TestLibraryImageExists (0.02s)
    library_test.go:117: command failed: exit status 1 - stdout:  - stderr: Usage:
          singularity [global options...] pull [pull options...] [output file] <URI>
        
        
FAIL
```

`make testall`

### LINKED ISSUE #3592: Improve cache management in the context of `DropPrivilege()` and `EnsurePrivilege()`

### LINKED ISSUE #3722: Unit tests cannot rely on the SINGULARITY_CACHEDIR environment variable because of thread safety problems
## Version of Singularity:

Master

### Expected behavior

The development team seems to agree that it would be beneficial for unit tests to be able to set and use their own private temporary image cache. An option to implement this would be to create a handle that represents an image cache and use it throughout the unit test(s) as required. In other terms, be able to create a temporary image cache, use it, potentially share it and clean it at the end.

### Actual behavior

Relying on the `SINGULARITY_CACHEDIR` environment variable in unit tests to specify the location of the image cache is not thread safe. As a result, a unit test that would like to change the value of the environment is required either to not be multi-threaded (Go will run unit tests in parallel) or rely only on a handle/object that will describe a given image cache (see #3592).
It seems to be the main reason why unit tests were initially shared a single location as image cache, i.e., `/tmp/WithPrivilege` or `/tmp/WithPrivilege`. As a result, it is also unsafe to clean a cache in the context of a test since it may impact the behavior of a concurrent test relying on the same image cache.

An effect of this problem is that unit tests are *not* supposed to call the Singularity CLI (for instance to pull an image) because the image cache location can only be specified by the unit test through the environment variable and therefore make it safe-unsafe.

Experienced failures while running the test in `internal/pkg/client/cache` on a SMP Ubuntu machine.

gvallee:

This PR fixes a few issues with the image cache:
- From purely a technical point-of-view, image caches are now accessed through a handle which let us control precisely which image cache to use.
- It fixes thread-safety problems by avoiding to use a unique directory for all image caches that is shared between uni tests. This is mainly done by avoiding to use the `SINGULARITY_CACHEDIR` environment variable and create a unique temporary directory as image cache for all unit tests (instead of `/tmp/WithPrivilege` and `/tmp/WithoutPrivilege`).
- It avoids changing the image cache location during the execution of a CLI command. To do so, CLI commands will capture the value of the `SINGULARITY_CACHEDIR` environment variable, figure out the image cache location based on it, and create an handle that will then be used for all image cache operations. This required to change the assembler and conveyor since the location of the image cache was before implicit in the sense that the code would repeatedly refer to the environment variable to access the image cache every time necessary.
- I checked how unit tests are using the image cache and everything seems now to be fine after ensuring that each unit tests has its own unique image cache (I was experiencing failures on my development system with the latest Ubuntu that the CI did not report).

I would need feedback about the following points (Is the approach okay? Is the implementation of the approach okay? And do we want/need to open new issues as a follow up?):
- In order to use the image cache handle throughout the entire code path, I added a field to the structure capturing the build configuration to store a pointer to a image cache handle. This modification lets us get the value of the `SINGULARITY_CACHEDIR` environment variable early on during the execution of a CLI command involving the image cache, create a handle and store it for future use. This pointer is then later on passed on to the conveyor and assembler, for actually accessing and using the image cache. The conveyor interface has been extended to be able to set an internal image cache handle pointer. Some conveyors do not need to use the image cache, in which case the pointer is set to nil. Feedback from someone who worked on the conveyor and the build path code would be valuable.
- Unit tests and the code is now thread-safe in the sense that it does not rely on the `SINGULARITY_CACHEDIR` environment variable anymore. However, the code is `cmd/singularity/actions_test.go` may need to be revisited. The assumptions made in the file are basically the following: a privileged user create an image (relying on an image cache) and a unprivileged user is using the image to perform various tests. While it is now thread-safe, I still encountered what appeared to be random problems. I also updated the code to make the sharing of an image between a privilege and unprivileged user explicit. Ultimately, it should move to the e2e framework

- Fixes #3592 
- Fixes #3722
- Fixes #3557 
- Fixes #3525 


cmd/internal/cli/actions.go
```diff
@@ -39,10 +39,16 @@ func actionPreRun(cmd *cobra.Command, args []string) {
 	os.Setenv("USER_PATH", userPath)
 	os.Setenv("PATH", defaultPath)
 
-	replaceURIWithImage(cmd, args)
+	// create an handle for the current image cache
+	imgCache, err := cache.HdlInit(os.Getenv(cache.DirEnv))
+	if imgCache == nil || err != nil {
```
mem: the `HdlInit` function never returns a nil image cache in the absence of an error. What I mean is that functionally these two things are equivalent:

```
    h, err := cache.HdlInit(dir)
    if h == nil || err != nil {
```

and

```
    h, err := cache.HdlInit(dir)
    if err != nil {
```
as there's never a case where `err == nil` and `h == nil`.

While having the extra check seems prudent, it makes the code harder to read, because the reader to stop and ask themselves why the additional check is required. The second version is idiomatic: if there's an error, I don't look at h, and if there's no error, h has a valid value.

Also, since all the uses of this in this package (`cli`) end up calling `syslog.Fatal`, this _could_ be moved to a function of its own:

```
func getCacheHandle() *cache.Handle {
    h, err := cache.HdlInit(os.Getenv(cache.DirEnv))
    if err != nil {
        sylog.Fatal(...)
    }
    return h
}
```

I'm even tempted to say this could use `sync.Once`, but that's probably going too far for this PR.

internal/pkg/client/cache/dir.go
```diff
@@ -22,42 +23,186 @@ const (
...
+}
+
+// HdlInit initializes a new cache within a given directory. It does not set
+// the environment variable to specify the location of the cache, the caller
+// being in charge of doing so. This also allows us to have a thread-safe
+// function, changing the value of the environment variable potentially
+// impacting other threads (e.g., while running unit tests). If baseDir is an
+// empty string, the image cache will be located to the default location, i.e.,
+// $HOME/.singularity.
+func HdlInit(baseDir string) (*ImgCache, error) {
```
mem: I would prefer to use something like NewHandle (with a return type of *Handle, not *ImgCache) just because that's more conventional in Go. With that change, the client code reads like:

```
    h := &cache.Handle{}
    h, err := cache.NewHandle(...)
```

cmd/singularity/build_test.go
```diff
@@ -110,14 +114,22 @@ func TestBuild(t *testing.T) {
 				}
 			}
 
+			// Set a clean image cache for every test
+			imgCacheDir := test.SetCacheDir(t, "")
```
mem: since this bit of code is repeated several times, maybe something like this helps:

```
func setupCache(t *testing.T) (*cache.Handle, func()) {
    dir := test.SetCacheDir(t, "")
    h, err := cache.HdlInit(dir)
    if err != nil {
        test.CleanCacheDir(t, dir)
        t.Fatal("...")
    }
    return h, func() { test.CleanCacheDir(t, dir) }
}
```

and then:

```
    h, cleanup := setupCache(t)
    defer cleanup()
    // use h
```

cmd/singularity/build_test.go
```diff
@@ -996,7 +1039,19 @@ func verifyEnv(t *testing.T, imagePath string, env []string, flags []string) err
 	}
 	args = append(args, imagePath, "env")
 
+	// We always prefer to run tests with a clean temporary image cache rather
+	// than using the cache of the user running the test.
+	// In order to unit test using the singularity cli that is thread-safe,
+	// we prepare a temporary cache that the process running the command will
+	// use.
+	tmpImgCache, err := ioutil.TempDir("", "image-cache-")
```
mem: this bit of code is also repeated a couple of times...

Maybe:

```
func setupCmdCache(t *testing.T, cmd *exec.Cmd, tag string) {
    dir, err := ioutil.TempDir("", tag + "-")
    if err != nil {
        t.Fatalf("...")
    }
    env := cache.DirEnv + "=" + dir
    if cmd.Env == nil {
        cmd.Env = os.Environ
    }
    cmd.Env = append(cmd.Env, env)
}
```

and then you do:

```
    cmd := exec.Command(cmdPath, args...)
    setupCmdCache(t, cmd, "image-cache")
    out, err := cmd.CombinedOuptut()
```

internal/pkg/client/cache/dir.go
```diff
@@ -22,42 +23,186 @@ const (
...
+	newCache.Oras, err = getOrasCachePath(newCache)
+	if err != nil {
+		return nil, fmt.Errorf("failed getting the path to the ORAS cache")
+	}
+
+	return newCache, nil
+}
+
+// getCacheBaseDir figures out where the cache base directory is.
+// Singularity makes the following assumptions:
```
mem, line 135: the formatter for the Go documentation will consider everything before a blank line a single paragraph, so this would get formatted in a way similary to this:

> getCacheBaseDir figures out where the cache base directory is. Singularity makes the following assumptions: - a user ...

just add a blank between the first and second lines (but include the leading //)

internal/pkg/client/cache/dir.go
```diff
@@ -22,42 +23,186 @@ const (
...
+//   change the location
+// - a user can change the location of a cache at any time
+// - but in the context of a Singularity command, the cache location
+//   cannot change once the command starts executing
+func getCacheBasedir() string {
+	// If the user defined the special environment variable, we use its value
+	// as base directory.
+	basedir := os.Getenv(DirEnv)
+	if basedir != "" {
+		fmt.Println("cache basedir set to:", basedir)
```
mem: we probably don't want this here, or we want sylog.Debugf

internal/pkg/client/cache/dir.go
```diff
@@ -22,42 +23,186 @@ const (
...
 
-var root string
+// ImgCache is an structure representing a cache
+type ImgCache struct {
+	// Basedir is the base directory of the image cache. By default, it is set
+	// to $HOME/.singularity. Users can also set the SINGULARITY_CACHEDIR
+	// environment variable to set BaseDir. BaseDir is also used when the code
+	// sets the SINGULARITY_CACHEDIR for a child process that needs to use the
+	// image cache.
+	BaseDir string
```
mem: I _still_ think this should not be exported.

I was going to comment this elsewhere, but it probably makes more sense to do it here.

The problem with exporting all these fields instead of having functions to access the values is this:

```
package cache

type ImgCache struct {
    Library string
}
```

and then later:

```
package foo

func bar() {
    c := ImgCache{Library: "some value"}
}

func quux() {
    h := getCacheHandle()
    h.Library = "some value"
}
```

if you do this instead:

```
package cache

type ImgCache struct {
    library string
}
```

and then later:

```
package foo

func bar() {
    c := ImgCache{library: "some value"} // this is a compilation error
}

func quux() {
    h := getCacheHandle()
    h.library = "some value" // this is a compilation error
}
```

also, since all the fields in the struct would not be exported, and since the zero value of the struct is not useful, users are forced to go thru `cache.NewHandle` to obtain a handle. Of course within the cache package you can still do all sort of nasty stuff, because it would have access to all the unexported fields.

The following does not need to be implemented, I'm mentioning it for the rationale.

Defining functions like this also allows you to define interfaces. For example:

```
package shub

type ShubCache interface {
    Shub() string
}

func Download(cache ShubCache, ...) {
    // ...
    dir := cache.Shub()
    // ...
}
```

and other code could pass an instance of cache.Handle to this function.

internal/pkg/client/cache/dir.go
```diff
@@ -85,13 +230,14 @@ func initCacheDir(dir string) error {
 
 // cleanAllCaches is an utility function that wipes all files in the
 // cache directory, will return a error if one occurs
-func cleanAllCaches() {
+func (c *ImgCache) cleanAllCaches() {
 	// TODO: add oras here
```
mem: this TODO is done, no?

internal/pkg/client/cache/library.go
```diff
@@ -1,38 +1,53 @@
...
 	"github.com/sylabs/scs-library-client/client"
-	"github.com/sylabs/singularity/internal/pkg/sylog"
 )
 
 const (
 	// LibraryDir is the directory inside the cache.Dir where library images are cached
 	LibraryDir = "library"
 )
 
+/*
```
mem: please delete code, don't comment it out

mem (changes requested):

internal/pkg/client/cache/dir.go
gvallee: Not possible: we need `BaseDir` to correctly set the environment variable when starting a new process via `exec` as the comments are stating. We can have a `GetBasedir()` function tho. @mem would you prefer that option? Your point about not letting anyone change the value of `BaseDir` is very valid.

internal/pkg/client/cache/dir.go
gvallee: I do not know this PR has nothing to do with it. It seems that it is done but i am not quite sure. I will remove the comment.

internal/pkg/client/cache/library.go
gvallee: Agreed. Just something i missed before creating the PR.

internal/pkg/client/cache/dir.go
mem: I'm sorry I'm not being clear: change `BaseDir` to `baseDir` and add a function to _read_ the value. I'm not asking you to remove it.

internal/app/singularity/pull.go
```diff
@@ -343,12 +343,18 @@ func OciPull(name, imageURI, tmpDir string, ociAuth *ocitypes.DockerAuthConfig,
 	return nil
 }
 
-func convertDockerToSIF(image, cachedImgPath, tmpDir string, noHTTPS bool, authConf *ocitypes.DockerAuthConfig) error {
+func convertDockerToSIF(imgCache *cache.Handle, image, cachedImgPath, tmpDir string, noHTTPS bool, authConf *ocitypes.DockerAuthConfig) error {
+	if imgCache == nil {
+		fmt.Println("Image cache is undefined")
+		os.Exit(42)
```
ikaneshiro: What's the significance of the `42` exit code? I would think this should return an error instead

internal/pkg/build/assemblers/assembler_sandbox.go
```diff
@@ -28,8 +31,15 @@ func (a *SandboxAssembler) Assemble(b *types.Bundle, path string) (err error) {
 	if _, err := os.Stat(path); err == nil {
 		os.RemoveAll(path)
 	}
+
 	var stderr bytes.Buffer
 	cmd := exec.Command("mv", b.Rootfs(), path)
+	// If the assemble is associated to a specific image cache, we make
+	// sure that the environment variable points to it for the process
+	// that will create.
```
ikaneshiro: I think this is outside the functionality of the assembler. All it should really be doing is taking the root file system created during the build process and package it into whatever format is specified. It shouldn't need to interact with the image cache

internal/pkg/build/build.go
```diff
@@ -195,6 +199,11 @@ func (b *Build) Full() error {
 			}
 		} else {
 			// regular build or force, start build from scratch
+			if b.Conf.ImgCache == nil {
+				fmt.Println("WARNING!!! imgCache is undefined")
+				os.Exit(42)
```
ikaneshiro: Given I saw this above I assume these are debugging statements. I think an error is appropriate here.

internal/pkg/client/cache/dir.go
```diff
@@ -22,42 +23,185 @@ const (
...
+func getCacheBasedir() string {
+	// If the user defined the special environment variable, we use its value
+	// as base directory.
+	basedir := os.Getenv(DirEnv)
+	if basedir != "" {
+		return basedir
+	}
+
+	// If the environment variable is not set, we use the default cache.
+	fmt.Println("env var not set, using default image cache")
```
ikaneshiro: Lets use `sylog.Infof()` here

internal/pkg/client/cache/library_test.go
```diff
@@ -68,26 +85,69 @@ func TestLibraryImage(t *testing.T) {
...
+	}
+
+	// Update the path to the file with the valid hash value
+	newSumPath := filepath.Join(c.Library, sum)
+	err = os.Rename(sumPath, newSumPath)
+	if err != nil {
+		t.Fatalf("failed to rename directory %s to %s: %s", sumPath, newSumPath, err)
+	}
+
+	fmt.Println("File is:", filepath.Join(newSumPath, filename))
```
ikaneshiro: `t.Logf()` is probably the way to go here

internal/pkg/client/cache/library_test.go
```diff
@@ -68,26 +85,69 @@ func TestLibraryImage(t *testing.T) {
...
 func TestLibraryImageExists(t *testing.T) {
 	test.DropPrivilege(t)
 	defer test.ResetPrivilege(t)
 
+	imageCacheDir, err := ioutil.TempDir("", "image-cache-")
+	if err != nil {
+		t.Fatal("failed to create temporary image cache directory")
+	}
+	//defer os.RemoveAll(imageCacheDir)
+	fmt.Println("Cache is there: ", imageCacheDir)
```
ikaneshiro: Same here, also I assume this defer should be uncommented

internal/pkg/build/build.go
```diff
@@ -47,6 +48,9 @@ type Build struct {
 
 // Config defines how build is executed, including things like where final image is written.
 type Config struct {
+	// Image cache handle
+	ImgCache *cache.Handle
```
ikaneshiro: I would consider including this in the options struct in https://github.com/sylabs/singularity/blob/master/pkg/build/types/bundle.go#L43. This way you can access the handle without needing to expand the conveyorPacker interface and without needing to set dummy functions for converyorPackers that dont use the image cache. Any conveyorPacker that needs it can simply get the handle from the bundle in their `Get()` method.

ikaneshiro (dismissed):
This is looking good. Just one suggestion related to image build system, the rest are nitpicks

internal/app/singularity/pull.go
gvallee: None, i use 42 during debugging and simply forgot to remove it. Sorry about that.

mem (approved):

internal/pkg/build/assemblers/assembler_sandbox.go
```diff
@@ -11,12 +11,15 @@ import (
...
+	"github.com/sylabs/singularity/internal/pkg/client/cache"
 	"github.com/sylabs/singularity/internal/pkg/sylog"
 	"github.com/sylabs/singularity/pkg/build/types"
 )
 
-// SandboxAssembler doesnt store anything
+// SandboxAssembler stores data required to assemble the image, for instannce the
+// image cache
 type SandboxAssembler struct {
+	ImgCache *cache.Handle
```
ikaneshiro: Do we need to keep this?

internal/pkg/build/conveyorPacker.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, Sylabs Inc. All rights reserved.
+// Copyright (c) 2018-2019, Sylabs Inc. All rights reserved.
```
ikaneshiro: Nitpick: these dates shouldn't be changed since we arent adding code anymore

internal/pkg/build/sources/conveyorPacker_arch.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, Sylabs Inc. All rights reserved.
+// Copyright (c) 2018-2019, Sylabs Inc. All rights reserved.
```
ikaneshiro: Same here

internal/pkg/build/sources/conveyorPacker_busybox.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, Sylabs Inc. All rights reserved.
+// Copyright (c) 2018-2019, Sylabs Inc. All rights reserved.
```
ikaneshiro: here

internal/pkg/build/sources/conveyorPacker_debootstrap.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, Sylabs Inc. All rights reserved.
+// Copyright (c) 2018-2019, Sylabs Inc. All rights reserved.
```
ikaneshiro: here

internal/pkg/build/sources/conveyorPacker_scratch.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, Sylabs Inc. All rights reserved.
+// Copyright (c) 2018-2019, Sylabs Inc. All rights reserved.
```
ikaneshiro: here

internal/pkg/build/sources/conveyorPacker_shub.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, Sylabs Inc. All rights reserved.
+// Copyright (c) 2018-2019, Sylabs Inc. All rights reserved.
```
ikaneshiro: here

internal/pkg/build/sources/conveyorPacker_yum.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, Sylabs Inc. All rights reserved.
+// Copyright (c) 2018-2019, Sylabs Inc. All rights reserved.
```
ikaneshiro: here

internal/pkg/client/cache/oras.go
```diff
@@ -18,21 +18,36 @@ const (
 	OrasDir = "oras"
 )
 
+/*
 // Oras returns the directory inside the cache.Dir() where oras images are cached
```
ikaneshiro: We can probably get rid of this function right?

ikaneshiro (changes requested):
Looks like just a bit of tidying up is needed then we're good to merge

internal/pkg/build/assemblers/assembler_sandbox.go
gvallee: No we do not

ikaneshiro (approved):
lgtm

PR #3771, opened 2019-06-21
Network CNI plugins dependencies update

### LINKED ISSUE #3766: Instance networking firewall issues
## Version of Singularity:

3.2.1

This is a known issue that @cclerget is working on, and I wanted to document the commands to reproduce it here since Slack messages get lost in some deep bucket of "If you paid us for the service you would see it..."

Specifically, it seems to be the case that on linux machines with Docker installed, (we are hypothesizing) that Docker has done manipulation of the iptables so that a firewall rule prevents different container instances on the same bridge from seeing one another. A simple example to reproduce was provided by @cclerget :+1: 

```bash
sudo singularity instance start --net --network-args="IP=10.22.10.10" docker://alpine A
sudo singularity instance start --net --network-args="IP=10.22.10.11" docker://alpine B
sudo singularity exec instance://A ping 10.22.10.11
sudo singularity exec instance://B ping 10.22.10.10
```

### Expected behavior
The expected behavior for one container to ping another should be a stream of responses (packets sent and then received).

### Actual behavior

However, the actual behavior is that we hang - the message is sent without receipt. 

```bash
sudo singularity exec instance://A ping 10.22.10.11
PING 10.22.10.11 (10.22.10.11): 56 data bytes
```

Note that this is also the case when we stop all docker services,

```bash
$ sudo service docker stop
```
I was also able to generate the same application I was originally testing, but [dumped into one container](https://github.com/singularityhub/singularity-compose-simple) and it functioned as expected.

@cclerget please let me know if/when you would like a tester! My computer is fully infiltrated by the Docker bugs and ready :)

cclerget:

This is an attempt to fix an issue when firewall rules are already set (system, docker ...) and conflicts with ones set by CNI plugins

This PR also :
- Updates network CNI plugins dependencies
- Applies changes in network package to use the new API allowing to provide a timeout context for CNI plugins execution
- Updates Singularity network configuration files by adding a new meta plugin `firewall` which will add a forward rule for traffic passthrough between host <=> container

- Fixes #3766


vsoch:
Here is a testing of the above PR, first we pull the code here, and install:
```bash
./mconfig 
make -C builddir
cd builddir/
sudo make install
```
Then we create two instances, as previously done

```bash
sudo singularity instance start --net --network-args="IP=10.22.10.10" docker://alpine A
sudo singularity instance start --net --network-args="IP=10.22.10.11" docker://alpine B
sudo singularity exec instance://A ping 10.22.10.11
sudo singularity exec instance://B ping 10.22.10.10
```

And... it works!!! :D :D :D 

https://asciinema.org/a/252939?speed=2

Just curious, are the vendor modules supposed to be served alongside the code, or is this just for testing and the user will install them? I'm going to do the last testing with the singularity-compose-example (with multiple containers) to confirm that everything is good. This one likely will take me a bit longer, so please stay tuned!

cclerget:
@vsoch Yes, users doesn't need to install them, vendor is provided when you pull a branch or with git archives and network plugins are compiled and installed in one pass

vsoch:
Okay, working beautifully for the multiple container example! This will fix the issues for any user that has Docker installed, and doesn't have a squeaky clean iptables. I'll add a note to the README that it's recommended to install this branch, given having Docker on the machine.

vsoch:
Full example (with the PR) is working and ready to go! https://github.com/singularityhub/singularity-compose-example

pkg/network/network_linux.go
```diff
@@ -448,13 +450,19 @@ func (m *Setup) command(command string) error {
 
 	config := &libcni.CNIConfig{Path: []string{m.cniPath.Plugin}}
 
+	// provide a timeout context for CNI plugin execution
```
gvallee:
```suggestion
	// set a timeout context for the execution of the CNI plugin
```

pkg/network/network_linux.go
```diff
@@ -448,13 +450,19 @@ func (m *Setup) command(command string) error {
...
+	// to interrupt the execution if plugins tooks more
```
gvallee:
```suggestion
	// to interrupt its execution if it takes more than 5 seconds
```

pkg/network/network_linux.go
```diff
@@ -448,13 +450,19 @@ func (m *Setup) command(command string) error {
...
+	// than 5 seconds to do their job
```
gvallee:
```suggestion
```

gvallee (changes requested):
LGTM, I only suggest a few fixes including a slightly better way to phrase a couple of comments. Feel free to ignore.

WestleyK (approved):
Looks good to me :+1:

gvallee (approved):
LGTM!

vsoch:
Sorry for a late ping on this - I'm finding that using Docker on my machine is totally failing - running containers (on port 3000, for example) with Docker that worked before, many times, are giving me a firewall "This site cannot be reached" error. @cclerget is there anything that was done here to possibly get in the way of Docker working?

cclerget:
@vsoch I can't figure out how CNI plugins responsible of setting firewall rules prevent docker from working, basically what the plugin does is :

```
iptables -N CNI-FORWARD
iptables -I FORWARD 1 -j CNI-FORWARD
iptables -A CNI-FORWARD -s <container_IP> -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
iptables -A CNI-FORWARD -d <container_IP> -j ACCEPT
```

* `iptables -N CNI-FORWARD`: it creates a user defined chain `CNI-FORWARD`
* `iptables -I FORWARD 1 -j CNI-FORWARD`: insert a rule in the chain `FORWARD` (docker has a similar chain called `DOCKER-ISOLATION` or something like that) so all packet will go through `CNI-FORWARD` user defined chain to be checked against rules below
* `iptables -A CNI-FORWARD -s <container_IP> -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT`: allow traffic forwarding from the container to host
* `iptables -A CNI-FORWARD -d <container_IP> -j ACCEPT`: allow traffic forwarding from host to the container

Just to confirm I read this https://netfilter.org/documentation/HOWTO/packet-filtering-HOWTO-7.html#ss7.4, so when docker firewall are set, forwarded packets will go first through `CNI-FORWARD`, if those packets come from/go to a singularity container, they are flagged as accepted, if the packets come from/go to a docker container, they should just be ignored by `CNI-FORWARD` rules and continue to go through next rules in `FORWARD` chains, logically the next user defined chain in `FORWARD` is `DOCKER-ISOLATION`, so those packets should find a docker rule in `DOCKER-ISOLATION` chain accepting them.

Do you run Singularity and Docker containers in the same network address space ?

vsoch:
hey @cclerget ! I'm not sure what the issue was, but I wound up deleting and reinstalling docker (and all seems to be working). I haven't tested Singularity again but let's just be optimistic and (based on the logic above) say that all is well! Thanks for the detailed debugging. :)

PR #3780, opened 2019-06-24
Allow execution from a parent (rootless) user namespace

### LINKED ISSUE #3753: fake root requires Singularity to be installed as root
## Version of Singularity:

3.3.0-rc1

### Expected behavior

76> singularity build --fakeroot test_centos_basic_fakeroot.sif test_centos_basic.def

successful build

### Actual behavior

INFO:    Starting build...
Skipping fetch of repeat blob sha256:8ba884070f611d31cb2c42eddb691319dc9facf5e0ec67672fcfa135181ab3df
Copying config sha256:b9a1b1f0b2baaec83946a26d7045e4028f11eccc8b0e5b3514568e56a391beb2
Storing signatures
FATAL:   While performing build: while running engine: fakeroot feature requires to install Singularity as root

See above.
Note, I su'd to root to do make install. Does setuid need to be allowed. I configured with nosuid? I want to be able to leverage Linux namespaces, and not require suid.

cclerget:

Users may not want to install Singularity setuid binaries on their system and consequently prevent them to use `fakeroot` feature especially for image build. This PR fixes few issues preventing to build image from a parent user namespace already in rootless mode.

- Fixes #3753

### DIFF
--- a/cmd/internal/cli/actions_linux.go
+++ b/cmd/internal/cli/actions_linux.go
@@ -21,6 +21,7 @@ import (
 	"github.com/sylabs/singularity/internal/pkg/plugin"
 	"github.com/sylabs/singularity/pkg/image"
 	"github.com/sylabs/singularity/pkg/image/unpacker"
+	"github.com/sylabs/singularity/pkg/util/namespaces"
 	"github.com/sylabs/singularity/pkg/util/nvidia"
 
 	"github.com/spf13/cobra"
@@ -107,6 +108,7 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 
 	uid := uint32(os.Getuid())
 	gid := uint32(os.Getgid())
+	insideUserNs, _ := namespaces.IsInsideUserNamespace(os.Getpid())
 
 	// Are we running from a privileged account?
 	isPrivileged := uid == 0
@@ -124,7 +126,13 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 
 	syscall.Umask(0022)
 
-	starter := buildcfg.LIBEXECDIR + "/singularity/bin/starter-suid"
+	starter := filepath.Join(buildcfg.LIBEXECDIR, "singularity/bin/starter-suid")
+	// setuid workflow is not required for root user or if we
+	// are already running inside a user namespace
+	if uid == 0 || insideUserNs {
+		starter = filepath.Join(buildcfg.LIBEXECDIR, "singularity/bin/starter")
+	}
+	sylog.Debugf("Use starter binary %s", starter)
 
 	engineConfig := singularityConfig.NewConfig()
 
@@ -425,9 +433,10 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 
 	generator.AddProcessEnv("SINGULARITY_APPNAME", AppName)
 
-	// convert image file to sandbox if image contains
-	// a squashfs filesystem
-	if UserNamespace && fs.IsFile(image) {
+	// convert image file to sandbox if we are using user
+	// namespace or if we are currently running inside a
+	// user namespace
+	if (UserNamespace || insideUserNs) && fs.IsFile(image) {
 		unsquashfsPath := ""
 		if engineConfig.File.MksquashfsPath != "" {
 			d := filepath.Dir(engineConfig.File.MksquashfsPath)
--- a/internal/pkg/runtime/engines/imgbuild/engine_linux.go
+++ b/internal/pkg/runtime/engines/imgbuild/engine_linux.go
@@ -19,6 +19,7 @@ import (
 	"github.com/sylabs/singularity/internal/pkg/util/fs"
 	singularity "github.com/sylabs/singularity/pkg/runtime/engines/singularity/config"
 	"github.com/sylabs/singularity/pkg/util/capabilities"
+	"github.com/sylabs/singularity/pkg/util/namespaces"
 )
 
 // EngineOperations implements the engines.EngineOperations interface for
@@ -91,6 +92,13 @@ func (e *EngineOperations) PrepareConfig(starterConfig *starter.Config) error {
 		starterConfig.SetHybridWorkflow(true)
 		starterConfig.SetAllowSetgroups(true)
 
+		starterConfig.SetTargetUID(0)
+		starterConfig.SetTargetGID([]int{0})
+	} else if insideUserNs, setgroups := namespaces.IsInsideUserNamespace(os.Getpid()); insideUserNs && setgroups {
+		// we are currently running in a user namespace, if setgroups is
+		// allowed we assume this process can fake the root user, if it's not
+		// the case, it will fail later in C starter code
+		e.EngineConfig.Bundle.Opts.Fakeroot = true
 		starterConfig.SetTargetUID(0)
 		starterConfig.SetTargetGID([]int{0})
 	}
--- a/internal/pkg/runtime/engines/singularity/container_linux.go
+++ b/internal/pkg/runtime/engines/singularity/container_linux.go
@@ -34,6 +34,7 @@ import (
 	"github.com/sylabs/singularity/pkg/network"
 	"github.com/sylabs/singularity/pkg/util/fs/proc"
 	"github.com/sylabs/singularity/pkg/util/loop"
+	"github.com/sylabs/singularity/pkg/util/namespaces"
 	"github.com/sylabs/singularity/pkg/util/nvidia"
 	"golang.org/x/crypto/ssh/terminal"
 )
@@ -103,6 +104,14 @@ func create(engine *EngineOperations, rpcOps *client.RPC, pid int) error {
 		c.sessionSize = int(engine.EngineConfig.File.SessiondirMaxSize)
 	} else if engine.EngineConfig.GetAllowSUID() && !c.userNS {
 		c.suidFlag = 0
+	}
+
+	// user namespace was not requested but we need to check
+	// if we are currently running in a user namespace and set
+	// value accordingly to avoid remount errors while running
+	// inside a user namespace
+	if !c.userNS {
+		c.userNS, _ = namespaces.IsInsideUserNamespace(os.Getpid())
 	}
 
 	p := &mount.Points{}
--- a/internal/pkg/runtime/engines/singularity/process_linux.go
+++ b/internal/pkg/runtime/engines/singularity/process_linux.go
@@ -382,11 +382,6 @@ func (engine *EngineOperations) PostStartProcess(pid int) error {
 			return err
 		}
 
-		file.Config, err = json.Marshal(engine.CommonConfig)
-		if err != nil {
-			return err
-		}
-
 		pw, err := user.GetPwUID(uint32(uid))
 		if err != nil {
 			return err
@@ -396,6 +391,11 @@ func (engine *EngineOperations) PostStartProcess(pid int) error {
 		file.PPid = os.Getpid()
 		file.Image = engine.EngineConfig.GetImage()
 
+		// by default we add all namespaces except the user namespace which
+		// is added conditionally. This delegates checks to the C starter code
+		// which will determine if a namespace needs to be joined by
+		// comparing namespace inodes
+		path := fmt.Sprintf("/proc/%d/ns", pid)
 		namespaces := []struct {
 			nstype string
 			ns     specs.LinuxNamespaceType
@@ -407,12 +407,6 @@ func (engine *EngineOperations) PostStartProcess(pid int) error {
 			{"cgroup", specs.CgroupNamespace},
 			{"net", specs.NetworkNamespace},
 		}
-
-		// by default we add all namespaces except user namespace which
-		// is added conditionally. This delegates checks to C starter code
-		// which will determine if a namespace need to be joined by
-		// comparing namespace inodes
-		path := fmt.Sprintf("/proc/%d/ns", pid)
 		for _, n := range namespaces {
 			nspath := filepath.Join(path, n.nstype)
 			engine.EngineConfig.OciConfig.AddOrReplaceLinuxNamespace(string(n.ns), nspath)
@@ -424,6 +418,12 @@ func (engine *EngineOperations) PostStartProcess(pid int) error {
 				file.UserNs = true
 				break
 			}
+		}
+
+		// grab configuration to store in instance file
+		file.Config, err = json.Marshal(engine.CommonConfig)
+		if err != nil {
+			return err
 		}
 
 		return file.Update()
--- a/pkg/util/namespaces/user_linux.go
+++ b/pkg/util/namespaces/user_linux.go
@@ -1,0 +1,60 @@
+// Copyright (c) 2019, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package namespaces
+
+import (
+	"bufio"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strconv"
+	"strings"
+)
+
+// IsInsideUserNamespace checks if a process is already running in a
+// user namespace and also returns if the process has permissions to use
+// setgroups in this user namespace.
+func IsInsideUserNamespace(pid int) (bool, bool) {
+	// default values returned in case of error
+	insideUserNs := false
+	setgroupsAllowed := false
+
+	// can fail if the kernel doesn't support user namespace
+	r, err := os.Open(fmt.Sprintf("/proc/%d/uid_map", pid))
+	if err != nil {
+		return insideUserNs, setgroupsAllowed
+	}
+	defer r.Close()
+
+	scanner := bufio.NewScanner(r)
+	// we are interested only by the first line of
+	// uid_map which would give us the answer quickly
+	// based on the value of size field
+	if scanner.Scan() {
+		fields := strings.Fields(scanner.Text())
+
+		// trust values returned by procfs
+		size, _ := strconv.ParseUint(fields[2], 10, 32)
+
+		// a size of 4294967295 means the process is running
+		// in the host user namespace
+		if uint32(size) == ^uint32(0) {
+			return insideUserNs, setgroupsAllowed
+		}
+
+		// process is running inside user namespace
+		insideUserNs = true
+
+		// should not fail if open call passed
+		d, err := ioutil.ReadFile(fmt.Sprintf("/proc/%d/setgroups", pid))
+		if err != nil {
+			return insideUserNs, setgroupsAllowed
+		}
+		setgroupsAllowed = string(d) == "allow\n"
+	}
+
+	return insideUserNs, setgroupsAllowed
+}


jscook2345:
@DrDaveD

internal/pkg/runtime/engines/singularity/process_linux.go
```diff
@@ -386,7 +381,12 @@ func (engine *EngineOperations) PostStartProcess(pid int) error {
 		file.PPid = os.Getpid()
 		file.Image = engine.EngineConfig.GetImage()
 
-		namespaces := []struct {
+		// by default we add all namespaces except user namespace which
```
gvallee:
```suggestion
		// by default we add all namespaces except the user namespace which
```

internal/pkg/runtime/engines/singularity/process_linux.go
```diff
@@ -386,7 +381,12 @@ func (engine *EngineOperations) PostStartProcess(pid int) error {
...
+		// is added conditionally. This delegates checks to C starter code
```
gvallee:
```suggestion
		// is added conditionally. This delegates checks to the C starter code
```

internal/pkg/runtime/engines/singularity/process_linux.go
```diff
@@ -386,7 +381,12 @@ func (engine *EngineOperations) PostStartProcess(pid int) error {
...
+		// which will determine if a namespace need to be joined by
```
gvallee:
```suggestion
		// which will determine if a namespace needs to be joined by
```

gvallee (changes requested):
LGTM, I only have a few minor suggestions

DrDaveD:
@cclerget What's an example way to get a parent user namespace in rootless mode?  Using podman?  How exactly does this address #3753?

cmd/internal/cli/actions_linux.go
```diff
@@ -125,6 +127,10 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 	syscall.Umask(0022)
 
 	starter := buildcfg.LIBEXECDIR + "/singularity/bin/starter-suid"
+	if uid == 0 || parentUserNs {
+		starter = buildcfg.LIBEXECDIR + "/singularity/bin/starter"
```
mem: Let's please document the logic here: "if ... then ..., which means we can use the non-suid version of starter instead."

cmd/internal/cli/actions_linux.go
```diff
@@ -125,6 +127,10 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 	syscall.Umask(0022)
 
 	starter := buildcfg.LIBEXECDIR + "/singularity/bin/starter-suid"
+	if uid == 0 || parentUserNs {
```
mem: this bit (`uid == 0`) is called `isPrivileged` a few lines up.

Can `isPrivileged` be redefined as `uid == 0 || parentUserNs` or is that not equivalent?

And if it indeed equivalent, maybe that can serve as a guide for a better name for `parentUserNs`?

cmd/internal/cli/actions_linux.go
```diff
@@ -107,6 +108,7 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 
 	uid := uint32(os.Getuid())
 	gid := uint32(os.Getgid())
+	parentUserNs, _ := namespaces.ParentUserNamespace(os.Getpid())
```
mem: I have not looked at what the function does, but I'm guessing `parentUserNs` is a boolean? If that's the case maybe both the variable and the function could have a different name? Sorry I can't provide a concrete suggestion right now.

cmd/internal/cli/actions_linux.go
```diff
@@ -352,9 +358,9 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 			}
 			generator.SetProcessArgs([]string{"/sbin/init"})
 		}
-		pwd, err := user.GetPwUID(uint32(os.Getuid()))
+		pwd, err := user.GetPwUID(uint32(uid))
```
mem: note that on line 159 uid changed. Just making sure that that is not passing unnoticed.

cmd/internal/cli/actions_linux.go
```diff
@@ -427,7 +433,7 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 
 	// convert image file to sandbox if image contains
 	// a squashfs filesystem
-	if UserNamespace && fs.IsFile(image) {
+	if (UserNamespace || parentUserNs) && fs.IsFile(image) {
```
mem: the variable `UserNamespace` means a user namespace is going to be set up.

The comment above this line needs updating to reflect the conditions that are being checked in order to enter this branch.

internal/pkg/runtime/engines/imgbuild/engine_linux.go
```diff
@@ -91,6 +92,13 @@ func (e *EngineOperations) PrepareConfig(starterConfig *starter.Config) error {
 		starterConfig.SetHybridWorkflow(true)
 		starterConfig.SetAllowSetgroups(true)
 
+		starterConfig.SetTargetUID(0)
+		starterConfig.SetTargetGID([]int{0})
+	} else if parent, setgroups := namespaces.ParentUserNamespace(os.Getpid()); parent && setgroups {
+		// we run in a parent user namespace and if setgroups is allowed
+		// we assume this process can fake root user, otherwise let it
```
mem: the "otherwise let ..." part...

does it mean that it _will_ fail unless we do this, or that it might still fail even if we do this?

internal/pkg/runtime/engines/singularity/container_linux.go
```diff
@@ -105,6 +106,10 @@ func create(engine *EngineOperations, rpcOps *client.RPC, pid int) error {
 		c.suidFlag = 0
 	}
 
+	if !c.userNS {
+		c.userNS, _ = namespaces.ParentUserNamespace(os.Getpid())
```
mem: Please add a comment here. I'm guessing something like "even if the use of a user namespace was not requested by the user, we are already running in a user namespace, so set things up as if the user requested that mode of operation".

Or whatever is right. I'm guessing.

internal/pkg/runtime/engines/singularity/process_linux.go
```diff
@@ -386,7 +381,12 @@ func (engine *EngineOperations) PostStartProcess(pid int) error {
 		file.PPid = os.Getpid()
 		file.Image = engine.EngineConfig.GetImage()
 
-		namespaces := []struct {
+		// by default we add all namespaces except the user namespace which
+		// is added conditionally. This delegates checks to the C starter code
+		// which will determine if a namespace needs to be joined by
+		// comparing namespace inodes
+		path := fmt.Sprintf("/proc/%d/ns", pid)
+		for _, n := range []struct {
```
mem: I can't see why this change is needed (the fix in the comment is fine, thanks).

Is the variable name "namespaces" clashing with something else? Can you just rename the variable?

pkg/util/namespaces/user_linux.go
```diff
@@ -0,0 +1,59 @@
...
+		// trust values returned by procfs
+		size, _ := strconv.ParseUint(fields[2], 10, 32)
+
+		// a size of 4294967295 means the process is running
+		// in the host user namespace
+		if uint32(size) == ^uint32(0) {
+			return parent, setgroups
+		}
+
+		// we are in a parent user namespace
```
mem: I'm having trouble with the phrase "a parent user namespace".

From this function I gather it means "we are running in a user namespace" or "this user namespace is not the root one".

pkg/util/namespaces/user_linux.go
```diff
@@ -0,0 +1,59 @@
...
+		parent = true
+
+		// should not fail if open call passed
+		d, err := ioutil.ReadFile(fmt.Sprintf("/proc/%d/setgroups", pid))
+		if err != nil {
+			return parent, setgroups
+		}
+		setgroups = string(d) == "allow\n"
+
+		break
```
mem: this means this is only going to read one line from the uid_map file. Maybe omit the for loop and do `if scanner.Scan()` instead?

Also, please add a comment as to why reading one line is enough.

pkg/util/namespaces/user_linux.go
```diff
@@ -0,0 +1,59 @@
...
+		// a size of 4294967295 means the process is running
+		// in the host user namespace
+		if uint32(size) == ^uint32(0) {
+			return parent, setgroups
+		}
+
+		// we are in a parent user namespace
+		parent = true
+
+		// should not fail if open call passed
```
mem, line 51: shouldn't this be a different function? Like CanSetGroups or something like that? I understand that the two things are related, but the new code shows there are cases where you care about the first value and not the second.

pkg/util/namespaces/user_linux.go
```diff
@@ -0,0 +1,59 @@
...
+	"io/ioutil"
+	"os"
+	"strconv"
+	"strings"
+)
+
+// ParentUserNamespace checks if a process is already running in a
+// parent user namespace, also returns if the process has permissions
+// to use setgroups.
+func ParentUserNamespace(pid int) (bool, bool) {
```
mem: Maybe call this IsInsideUserNamespace?

mem (changes requested):

cclerget:
> @cclerget What's an example way to get a parent user namespace in rootless mode

You can use rootlesskit https://github.com/rootless-containers/rootlesskit or the following script I didn't include in this PR because a discussion is required about user mappings configuration/control, should we use our own mappings or use `/etc/sub[uid|gid]` files or both ... ? From users feedbacks, there is very diverging expectations with this feature in term of security/configuration.

> How exactly does this address #3753?

Regarding this issue and with this PR and script below, you could use `singularity build` inside parent rootless namespace with help of `newuidmap/newgidmap` to have same effect as with `--fakeroot`.

Here the script :

```
#!/bin/sh
# Copyright (c) 2019, Sylabs Inc. All rights reserved.
# This software is licensed under a 3-clause BSD license. Please consult the
# LICENSE.md file distributed with the sources of this project regarding your
# rights to use or distribute this software.

if [ ! -z "${_FIFO_MAP}" ]; then
    pid=`cat ${_FIFO_MAP} 2>/dev/null`

    uid=`id -u`
    username=`id -un`
    gid=`id -g`
    groupname=`id -gn`

    subuid=`grep "^${username}:" /etc/subuid`
    if [ -z "${subuid}" ]; then
        subuid=`grep "^${uid}:" /etc/subuid`
        if [ -z "${subuid}" ]; then
            echo "FATAL: you're user is not set in /etc/subuid file" > ${_FIFO_MAP}
            exit 255
        fi
    fi
    uidrange=`echo ${subuid}|cut -d ":" -f2`
    uidsize=`echo ${subuid}|cut -d ":" -f3`

    subgid=`grep "^${groupname}:" /etc/subgid)`
    if [ -z "${subgid}" ]; then
        subgid=`grep "^${gid}:" /etc/subgid`
        if [ -z "${subgid}" ]; then
            echo "FATAL: you're group is not set in /etc/subgid file" > ${_FIFO_MAP}
            exit 255
        fi
    fi
    gidrange=`echo ${subgid}|cut -d ":" -f2`
    gidsize=`echo ${subgid}|cut -d ":" -f3`

    err=`newgidmap ${pid} 0 ${gid} 1 1 ${gidrange} ${gidsize} 2>&1`
    if [ ! -z "${err}" ]; then
        echo "FATAL: ${err}" > ${_FIFO_MAP}
        exit 255
    fi
    err=`newuidmap ${pid} 0 ${uid} 1 1 ${uidrange} ${uidsize} 2>&1`
    if [ ! -z "${err}" ]; then
        echo "FATAL: ${err}" > ${_FIFO_MAP}
        exit 255
    fi
    echo "success" > ${_FIFO_MAP}
elif [ ! -z "${_FIFO_FINAL}" ]; then
    home=${HOME}
    echo "$$" > ${_FIFO_FINAL}
    err=`cat ${_FIFO_FINAL} 2>/dev/null`
    rm -f ${_FIFO_FINAL}
    if [ "${err}" != "success" ]; then
        echo "${err}"
        exit 255
    fi
    if ! mount --bind ${home} /root 2>&1 >/dev/null; then
        echo "FATAL: failed to mount ${home} to /root"
        exit 255
    fi
    unset _FIFO_FINAL
    exec "$@"
else
    if ! command -v unshare 2>&1 >/dev/null; then
        echo "FATAL: unshare is not installed on your system"
        exit 255
    fi
    if ! command -v newuidmap 2>&1 >/dev/null; then
        echo "FATAL: newuidmap is not installed on your system"
        exit 255
    fi
    if ! command -v newgidmap 2>&1 >/dev/null; then
        echo "FATAL: newgidmap is not installed on your system"
        exit 255
    fi
    if ! command -v mkfifo 2>&1 >/dev/null; then
        echo "FATAL: mkfifo is not installed on your system"
        exit 255
    fi
    if ! command -v mktemp 2>&1 >/dev/null; then
        echo "FATAL: mktemp is not installed on your system"
        exit 255
    fi
    fifo=`mktemp -u 2>/dev/null`
    if ! mkfifo $fifo 2>&1 >/dev/null; then
        echo "FATAL: failed to create fifo for process communication"
        exit 255
    fi
    _FIFO_MAP="${fifo}" $0 &
    _FIFO_FINAL="${fifo}" exec unshare -U -m $0 "$@"
fi
```

cmd/internal/cli/actions_linux.go
cclerget: Good catch ! I will revert this

internal/pkg/runtime/engines/singularity/process_linux.go
cclerget: I can restore it as it was before, I changed during tests

cmd/internal/cli/actions_linux.go
cclerget: I will defer this change to https://github.com/sylabs/singularity/pull/3801

cmd/internal/cli/actions_linux.go
cclerget: not equivalent, will look at better naming

pkg/util/namespaces/user_linux.go
cclerget: No because it could be used wrongly because `/proc/pid/setgroups` is set to `allow` by default for process doesn't living inside a user namespace, so we only check read setgroups value once we have confirmation that we are indeed in a user namespace

cmd/internal/cli/actions_linux.go
```diff
@@ -125,6 +127,12 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 	syscall.Umask(0022)
 
 	starter := buildcfg.LIBEXECDIR + "/singularity/bin/starter-suid"
+	// setuid workflow is not required for root user or if we
+	// are already running inside a user namespace
+	if uid == 0 || insideUserNs {
+		starter = buildcfg.LIBEXECDIR + "/singularity/bin/starter"
```
gvallee: Would it make sense here to use `filepath.Join()`?

internal/pkg/runtime/engines/imgbuild/engine_linux.go
```diff
@@ -91,6 +92,13 @@ func (e *EngineOperations) PrepareConfig(starterConfig *starter.Config) error {
 		starterConfig.SetHybridWorkflow(true)
 		starterConfig.SetAllowSetgroups(true)
 
+		starterConfig.SetTargetUID(0)
+		starterConfig.SetTargetGID([]int{0})
+	} else if insideUserNs, setgroups := namespaces.IsInsideUserNamespace(os.Getpid()); insideUserNs && setgroups {
+		// we are currently running in a user namespace, if setgroups is
+		// allowed we assume this process can fake root user, if it's not
```
gvallee:
```suggestion
		// allowed we assume this process can fake the root user, if it's not
```

gvallee (changes requested):
I only have a couple of minor comments...

cmd/internal/cli/actions_linux.go
cclerget: Yep !

DrDaveD:
@cclerget Thanks for the script.  I found two bugs, both on this line:
```
subgid=`grep "^${groupname}:" /etc/subgid)`
```
Remove the close parentheses, and change groupname to username.  The corresponding error message should then be changed to "your user" instead of "you're group".

Isn't it kind of cheating to depend on an external script, however?  Wouldn't it be a whole lot friendlier if the --fakeroot option just worked, without requiring nesting of namespaces (assuming singularity is run inside this script, as this PR is making possible).

pkg/util/namespaces/user_linux.go
mem: good, thanks

mem (approved):
thanks!

internal/pkg/runtime/engines/imgbuild/engine_linux.go
```diff
@@ -91,6 +92,13 @@ func (e *EngineOperations) PrepareConfig(starterConfig *starter.Config) error {
...
+		starterConfig.SetTargetUID(0)
+		starterConfig.SetTargetGID([]int{0})
+	} else if insideUserNs, setgroups := namespaces.IsInsideUserNamespace(os.Getpid()); insideUserNs && setgroups {
+		// we are currently running in a user namespace, if setgroups is
+		// allowed we assume this process can fake the root user, if it's not
+		// the case, it will fail later in C starter code
+		e.EngineConfig.Bundle.Opts.Fakeroot = true
 		starterConfig.SetTargetUID(0)
 		starterConfig.SetTargetGID([]int{0})
 	}
```
gvallee, line 104:
```suggestion
	}
	starterConfig.SetTargetUID(0)
	starterConfig.SetTargetGID([]int{0})
```

internal/pkg/runtime/engines/imgbuild/engine_linux.go
```diff
@@ -91,6 +92,13 @@ func (e *EngineOperations) PrepareConfig(starterConfig *starter.Config) error {
...
 		starterConfig.SetAllowSetgroups(true)
 
+		starterConfig.SetTargetUID(0)
+		starterConfig.SetTargetGID([]int{0})
+	} else if insideUserNs, setgroups := namespaces.IsInsideUserNamespace(os.Getpid()); insideUserNs && setgroups {
+		// we are currently running in a user namespace, if setgroups is
+		// allowed we assume this process can fake the root user, if it's not
+		// the case, it will fail later in C starter code
+		e.EngineConfig.Bundle.Opts.Fakeroot = true
 		starterConfig.SetTargetUID(0)
```
gvallee, line 102:
```suggestion
```

gvallee, line 102:
```suggestion
```

internal/pkg/runtime/engines/imgbuild/engine_linux.go
```diff
@@ -91,6 +92,13 @@ func (e *EngineOperations) PrepareConfig(starterConfig *starter.Config) error {
 		starterConfig.SetHybridWorkflow(true)
 		starterConfig.SetAllowSetgroups(true)
 
+		starterConfig.SetTargetUID(0)
+		starterConfig.SetTargetGID([]int{0})
```
gvallee, line 96:
```suggestion
```

internal/pkg/runtime/engines/imgbuild/engine_linux.go
```diff
@@ -91,6 +92,13 @@ func (e *EngineOperations) PrepareConfig(starterConfig *starter.Config) error {
 		starterConfig.SetHybridWorkflow(true)
 		starterConfig.SetAllowSetgroups(true)
 
+		starterConfig.SetTargetUID(0)
```
gvallee, line 95:
```suggestion
```

gvallee, line 95: A few minor modifications to avoid some code duplication (I am being picky)... see following suggestions.

gvallee (changes requested):
Just a minor suggestion to avoid some minor code duplication.

internal/pkg/runtime/engines/imgbuild/engine_linux.go
cclerget: As discussed on Slack, putting `starterConfig.SetTargetGID([]int{0})` outside of those two conditions could fail later in C starter code when calling `setgroups` if for some reasons setgroups usage is denied

gvallee (approved):
LGTM

PR #3784, opened 2019-06-24
ORAS: discard logrus warnings depending on log level

### LINKED ISSUE #3773: Update ORAS push output
## Version of Singularity:

master

### Expected behavior
```
$ singularity push test.sif oras://localhost:5000/test/sif:newtest
INFO: Upload complete
```
### Actual behavior

Currently we are getting warning output from `logrus` used internally within the `oras` package, and do not give info indicating a successful push.
```
$ singularity push test.sif oras://localhost:5000/test/sif:newtest
$
```
or
```
$ singularity push test.sif oras://localhost:5000/test/sif:newtest
WARN[0000] encountered unknown type appliciation/vnd.sylabs.sif.layer.tar; children may not be fetched 
WARN[0000] encountered unknown type application/vnd.sylabs.sif.config.v1+json; children may not be fetched 
WARN[0000] reference for unknown type: appliciation/vnd.sylabs.sif.layer.tar  digest="sha256:9c8f9816e66e9fa6a3577c360f6cbd3676ae1f4ccaaa993ea2fe76b81bc14a50" mediatype=appliciation/vnd.sylabs.sif.layer.tar size=2723840
WARN[0000] reference for unknown type: application/vnd.sylabs.sif.config.v1+json  digest="sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855" mediatype=application/vnd.sylabs.sif.config.v1+json size=0
$
```

ikaneshiro:

The `containerd` packages used by `oras` to push images currently outputs warnings. This PR discards them unless a `-v` or `-d` sylog level is specified.

Also adds output on successful push with oras

- Fixes #3773 

### DIFF
--- a/cmd/internal/cli/push.go
+++ b/cmd/internal/cli/push.go
@@ -93,6 +93,7 @@ var PushCmd = &cobra.Command{
 			if err := oras.UploadImage(file, ref, ociAuth); err != nil {
 				sylog.Fatalf("Unable to push image to oci registry: %v", err)
 			}
+			sylog.Infof("Upload complete.")
 		}
 	},
 
--- a/internal/pkg/sylog/sylog.go
+++ b/internal/pkg/sylog/sylog.go
@@ -191,3 +191,14 @@ func Writer() io.Writer {
 
 	return os.Stderr
 }
+
+// LevelWriter returns an io.Writer to pass to an external packages logging utility.
+// This allows you to set the threshold for when output should be discarded
+// i.e if passed a level of 1, this function returns ioutil.Discard unless --debug is specified
+func LevelWriter(threshold int) io.Writer {
+	if int(loggerLevel) <= threshold {
+		return ioutil.Discard
+	}
+
+	return os.Stderr
+}
--- a/go.mod
+++ b/go.mod
@@ -78,7 +78,7 @@ require (
 	github.com/safchain/ethtool v0.0.0-20180504150752-6e3f4faa84e1 // indirect
 	github.com/satori/go.uuid v1.2.0
 	github.com/seccomp/libseccomp-golang v0.9.0
-	github.com/sirupsen/logrus v1.0.5 // indirect
+	github.com/sirupsen/logrus v1.0.5
 	github.com/spf13/cobra v0.0.0-20190321000552-67fc4837d267
 	github.com/spf13/pflag v1.0.3
 	github.com/stevvooe/resumable v0.0.0-20180830230917-22b14a53ba50 // indirect
--- a/internal/pkg/oras/oras.go
+++ b/internal/pkg/oras/oras.go
@@ -26,6 +26,7 @@ import (
 	"github.com/deislabs/oras/pkg/oras"
 	"github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/sirupsen/logrus"
 	"github.com/sylabs/singularity/internal/pkg/sylog"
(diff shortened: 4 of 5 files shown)


cmd/internal/cli/push.go
```diff
@@ -93,6 +93,7 @@ var PushCmd = &cobra.Command{
 			if err := oras.UploadImage(file, ref, ociAuth); err != nil {
 				sylog.Fatalf("Unable to push image to oci registry: %v", err)
 			}
+			sylog.Infof("Upload complete.")
```
WestleyK, line 96: Whats the `.` for? I dont think we use any _dots_ in our other Info messages...

internal/pkg/sylog/sylog.go
```diff
@@ -191,3 +191,14 @@ func Writer() io.Writer {
 
 	return os.Stderr
 }
+
+// LevelWriter returns an io.Writer to pass to an external packages logging utility.
+// This allows you to set the threshold for when output should be discarded
+// i.e if passed a level of 1, this function returns ioutil.Discard unless --debug is specified
+func LevelWriter(threshold int) io.Writer {
```
mem: can we do this the other way around and pass a sylog.messageLevel

cmd/internal/cli/push.go
ikaneshiro: Not necessary, ill change that

internal/pkg/sylog/sylog.go
mem, line 198: I'm not sure I follow the logic here (FWIW, I also don't follow the logic in the current Writer function).

Let me try to figure out which part I am misunderstanding:

* `loggerLevel` is set once per singularity run, either from the environment or the command line flags.
* there's an equivalency between sylog's levels and logrus' levels
* if the current `loggerLevel` is below the specified threshold, a discard io.Writer is returned
* if the current `loggerLevel` is above the threshold, return an io.Writer that writes to stderr

to me the above means that for each message that the external logger wants to produce, they would be a call to sylog.LevelWriter with the appropriate level. So if the external logger wants to produce a debug message, there's a call to sylog.LevelWriter(sylog.debug) to obtain an io.Writer that discards or writes the message as a function of the sylog's `loggerLevel`.

That does not seem to be what this code is doing.

Shouldn't this PR be calling `SetLevel` and translate from sylog's level into logrus' level?

Also, "1" corresponds to sylog.info, so this change causes messages to be shown only if the sylog level is at `verbose`, `verbose2`, `verbose3` or `debug`. If the sylog level is at sylog.info and the code using logrus logs an error, that is not shown, right?

mem (approved):

mem (changes requested):
Sorry, hit approve, I meant to hit comment.

internal/pkg/sylog/sylog.go
ikaneshiro: This isn't intended to match the log levels between the two loggers. The idea is to conditionally squash the output of logrus because it is throwing warnings that are relevant to `containerd` about unknown oci mediaTypes, but not to us. I have also looked and the code causing this output and it is not configurable to registering a new mediaType. So, I decided that we can just squash these warnings under normal circumstances, but have them appear when we want more verbose output. 

If logrus logs an error that will not be shown, I assume that an error level message to logrus from that package would occur in conjunction with the functions involved returning an error and it can be debugged from there using a flag like `-d`

ikaneshiro:
Closing this PR as I have found the correct solution and placed it in #3791

PR #3789, opened 2019-06-24
Remove the incomplete-downloaded cache if interrupted while pulling

### LINKED ISSUE #3776: cache is clocking re-downloading of images
## Version of Singularity:

3.3.0-rc1

### Expected behavior

If my previous pull was incomplete/interrupted - I should be able to pull again

### Actual behavior

INFO:    Use image from cache (Which was an incomplete download)

```
$ singularity pull shub://michael-tn/mpi-hello-world:latest
```
While its downloading, interrupt the download
Run the pull command agian, you'll see
```
INFO:    Use image from cache
```

### LINKED ISSUE #3788: Interrupting a 'pull' will not delete the old image from cache
## Version of Singularity:

```
3.3.0-rc.1.29.g8900c2d22
```

### Expected behavior

After the interrupting the download, next time you download that image, it should restart. Not give you the half-downloaded image from cache.

### Actual behavior

```bash
$ singularity pull library://ubuntu:latest
# Press ctrl-c while downloading it

$ singularity pull library://ubuntu:latest
# and it will fail
```

<br>

Related to: #3776 

Also related to #3758, I think...

WestleyK:

This PR fixes the issue when you pull a image, then interrupt the pull, then pull it again.

- Fixes #3788 
- Fixes #3776 

<br>


internal/pkg/client/cache/library.go
```diff
@@ -53,7 +55,7 @@ func (c *Handle) LibraryImageExists(sum, name string) (bool, error) {
 		return false, err
 	}
 	if cacheSum != sum {
-		return false, fmt.Errorf("cached file sum (%s) and expected sum (%s) does not match", cacheSum, sum)
+		return false, ErrCacheHashNotTheSame
```
ikaneshiro: Can we do the same in `oras.go` if it's returning an error for this case too

internal/pkg/client/cache/library.go
```diff
@@ -6,13 +6,15 @@
...
 import (
-	"fmt"
+	"errors"
 	"os"
 	"path/filepath"
 
 	"github.com/sylabs/scs-library-client/client"
 )
 
+var ErrCacheHashNotTheSame = errors.New("hash does not match")
```
ikaneshiro: Maybe we should put this in `dir.go` since this error will now be part of the package API. Might also be worth calling it something like `ErrBadChecksum` so it's a little less verbose.

internal/app/singularity/pull.go
```diff
@@ -311,6 +351,17 @@ func OciPull(imgCache *cache.Handle, name, imageURI, tmpDir string, ociAuth *oci
 	}
 	if !exists {
 		sylog.Infof("Converting OCI blobs to SIF format")
+
+		// monitor for OS signals and remove invalid file
+		c := make(chan os.Signal)
+		signal.Notify(c, os.Interrupt, syscall.SIGTERM)
+		go func(fileName string) {
```
ikaneshiro: Maybe we should start considering a general signal handler implementation. This is very similar to the signal handling in `build.go` and we can probably create a package to do what we want. I wouldnt let this block the PR, but make a tech debt issue if you dont want to address it in this PR

ikaneshiro (dismissed):

internal/app/singularity/interrupt_cleanup.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/sylabs/singularity/internal/pkg/sylog"
+)
+
+// InterruptCleanup will watch for a interrupt signal, if theres
```
gvallee:
```suggestion
// InterruptCleanup will watch for a interrupt signal, if there's
```

gvallee (approved):
I just found a typo, otherwise LGTM. Thanks for implementing that.

internal/app/singularity/interrupt_cleanup.go
mem: please don't export unless you need it in a different package.

internal/app/singularity/interrupt_cleanup.go
```diff
@@ -0,0 +1,30 @@
...
+// one detected, then it will remove all the specified file(s)
+func InterruptCleanup(files ...string) {
+	c := make(chan os.Signal)
```
mem: the channel used to receive signal notifications should be buffered. A buffer size of 1 is enough in this case.

internal/app/singularity/pull.go
```diff
@@ -68,11 +68,19 @@ func LibraryPull(imgCache *cache.Handle, name, ref, transport, fullURI, libraryU
...
+	} else if err != nil {
 		return fmt.Errorf("unable to check if %s exists: %v", imagePath, err)
 	}
+
 	if !exists {
 		sylog.Infof("Downloading library image")
+		go InterruptCleanup(imagePath)
 
 		// call library download image helper
 		if err = library.DownloadImage(context.TODO(), libraryClient, imagePath, imageRef, downloadImageCallback); err != nil {
```
mem, line 85: since this is using a context, the signal handler should use it to notify the download process that it should stop. I would expect that if the download is stopped, this functions returns an error, and you can let the normal code flow continue from here (dealing with the error).

This also hints at the signal handler being installed "too low". It should probably be installed higher in the call stack so that you can pass a context down and handle cancellation in that way, as it is likely that all clients need the same handling.

mem (dismissed):

mem (commented):
Removing my request for changes, but not approving.

jscook2345:
@mem @gvallee 

Can we dismiss the approving review if this fix will be postponed? I don't want anyone to accidentally merge it seeing it was approved. Thanks!

internal/app/singularity/pull.go
GodloveD: This is a good idea, but I think it should be added as an additional improvement (see #3899).  Since this is fixing a bug, I think it can be added as it already is, and the context can be fixed later.

PR #3794, opened 2019-06-25
Add some unit testing to internal/app/starter

gvallee:

Add some unit testing to `internal/app/starter/master_linux.go` and reorganize slightly the code to do so. Since I do not know whether modifying the code of `master_linux.go` to improve testing in the way I did it is acceptable, this is only a limited number of modifications and tests. If accepted, I will add more test cases in a separate future PR.

My modifications of `internal/app/starter/master_linux.go` intends to have shorter functions so testing is easier.

- Addresses #3315 


internal/app/starter/master_linux.go
```diff
@@ -20,93 +20,119 @@ import (
...
+
+	err = engine.PostStartProcess(containerPid)
+	if err != nil {
+		fatalChan <- fmt.Errorf("post start process failed: %s", err)
+		return
+	}
+	if isInstance && os.Getppid() == ppid {
+		// sleep a bit to see if child exit
+		time.Sleep(100 * time.Millisecond)
+
```
mem: could you please delete this new newline? For this PR I would prefer to keep changes to a strict minimum.

internal/app/starter/master_linux.go
```diff
@@ -20,93 +20,119 @@ import (
...
+	if engine == nil || containerPid < 0 {
+		fatalChan <- fmt.Errorf("invalid parameter(s)")
+		return
+	}
+
+	ppid := os.Getppid()
+
+	go func() {
+		doRPCthread(rpcSocket, containerPid, engine, fatalChan)
+		runtime.Goexit()
```
mem: what's the logic of keeping this here vs in the doRPCthread function? I can guess, but I would prefer to have it documented so that people looking at this in the future don't have to.

internal/app/starter/master_linux.go
```diff
@@ -20,93 +20,119 @@ import (
 	"github.com/sylabs/singularity/internal/pkg/util/mainthread"
 )
 
-// Master initializes a runtime engine and runs it
-func Master(rpcSocket, masterSocket int, isInstance bool, containerPid int, engine *engines.Engine) {
-	var fatal error
-	var status syscall.WaitStatus
+func doRPCthread(rpcSocket int, containerPid int, engine *engines.Engine, fatalChan chan error) {
```
mem: I would prefer _not_ to call these functions "thread".

internal/app/starter/master_linux.go
```diff
@@ -20,93 +20,119 @@ import (
...
-			fatalChan <- fmt.Errorf("container creation failed: %s", err)
-		} else {
-			rpcConn.Close()
-		}
+	rpcConn.Close()
+}
 
-		runtime.Goexit()
-	}()
+func doMasterThread(masterSocket int, isInstance bool, containerPid int, ppid int, engine *engines.Engine, fatalChan chan error) {
```
mem: same here

internal/app/starter/master_linux_test.go
```diff
@@ -0,0 +1,171 @@
...
+			rpcSocket:    -1,
+			containerPid: -1,
+			engine:       nil,
+			shallPass:    false,
+		},
+		{
+			name:         "fake engine struct",
+			rpcSocket:    -1,
+			containerPid: -1,
+			engine:       &fakeEngine,
```
mem: please create a new engine (`&engines.Engine{}` or `new(engines.Engine)`) for each test case, otherwise the code under test might end up modifying the engine for the next test.

internal/app/starter/master_linux_test.go
```diff
@@ -0,0 +1,171 @@
...
+	}{
+		{
+			name:         "invalid case",
+			rpcSocket:    -1,
+			containerPid: -1,
+			engine:       nil,
+			shallPass:    false,
+		},
+		{
+			name:         "fake engine struct",
```
mem: if I'm understanding correctly, this case is hitting the "invalid RPC socket" path. It's probably better to name the test something along those lines.

internal/app/starter/master_linux_test.go
```diff
@@ -0,0 +1,171 @@
...
+			rpcSocket:    -1,
+			containerPid: -1,
+			engine:       &fakeEngine,
+			shallPass:    false,
+		},
+		{
+			name:         "fake engine; fake rpcSocket",
```
mem: not sure what this is testing. That the PID value of -1 is checked and "detected" as invalid?

internal/app/starter/master_linux_test.go
```diff
@@ -0,0 +1,171 @@
...
+			containerPid: -1,
+			engine:       &fakeEngine,
+			shallPass:    false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			go doRPCthread(tt.rpcSocket, tt.containerPid, tt.engine, fatalChan)
+			fatal = <-fatalChan
```
mem, line 59: interesting... since the code does not have an obvious way to make it stop (IIRC there's a magic value that you can write to the RPC socket that will cause it to exit), it's not evident whether this test will hang here forever or not.

Maybe add a comment?

internal/app/starter/master_linux_test.go
```diff
@@ -0,0 +1,171 @@
...
+			if tt.shallPass && fatal != nil {
```
mem, line 60: I wrote [this](https://github.com/sylabs/singularity/wiki/End-to-end-testing#check-all-cases) with E2E in mind, but it applies the same here. Can you please read that section and do what it says there regarding handling all cases?

internal/app/starter/master_linux_test.go
```diff
@@ -0,0 +1,171 @@
...
+			engine:       nil,
+			shallPass:    false,
+		},
+		{
+			name:         "fake engine",
+			masterSocket: -1,
+			isInstance:   false,
+			containerPid: -1,
+			ppid:         -1,
+			engine:       &fakeEngine,
```
mem: same

internal/app/starter/master_linux_test.go
```diff
@@ -0,0 +1,171 @@
...
+			shallPass:    false,
+		},
+		{
+			name:         "fake engine; fake masterSocket",
+			masterSocket: 42000,
+			isInstance:   false,
+			containerPid: -1,
+			ppid:         -1,
+			engine:       &fakeEngine,
```
mem: same

internal/app/starter/master_linux_test.go
```diff
@@ -0,0 +1,171 @@
...
+			name:         "invalid case",
+			masterSocket: -1,
+			isInstance:   false,
+			containerPid: -1,
+			ppid:         -1,
+			engine:       nil,
+			shallPass:    false,
+		},
+		{
+			name:         "fake engine",
```
mem: same

internal/app/starter/master_linux_test.go
```diff
@@ -0,0 +1,171 @@
...
+			masterSocket: -1,
+			isInstance:   false,
+			containerPid: -1,
+			ppid:         -1,
+			engine:       &fakeEngine,
+			shallPass:    false,
+		},
+		{
+			name:         "fake engine; fake masterSocket",
```
mem: same

internal/app/starter/master_linux_test.go
```diff
@@ -0,0 +1,171 @@
...
+			engine:       &fakeEngine,
+			shallPass:    false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			go doMasterThread(tt.masterSocket, tt.isInstance, tt.containerPid, tt.ppid, tt.engine, fatalChan)
+			fatal = <-fatalChan
+			if tt.shallPass && fatal != nil {
```
mem, line 107: same

internal/app/starter/master_linux_test.go
```diff
@@ -0,0 +1,171 @@
...
+			name:         "invalid case",
+			rpcSocket:    -1,
+			masterSocket: -1,
+			isInstance:   false,
+			pid:          -1,
+			engine:       nil,
+			shallPass:    false,
+		},
+		{
+			name:         "fake engine",
```
mem: same

internal/app/starter/master_linux_test.go
```diff
@@ -0,0 +1,171 @@
...
+			rpcSocket:    -1,
+			masterSocket: -1,
+			isInstance:   false,
+			pid:          -1,
+			engine:       &fakeEngine,
```
mem: same

internal/app/starter/master_linux_test.go
```diff
@@ -0,0 +1,171 @@
...
+			shallPass:    false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			Master(tt.rpcSocket, tt.masterSocket, tt.isInstance, tt.pid, tt.engine)
```
mem: probably better to omit this until there's something that can be tested? (and by "this" I mean "this test")

mem (changes requested):

internal/app/starter/master_linux.go
gvallee: What do you want to call these functions then? I personally have no preference. Would just remove the term "thread" be okay?

mem: I don't have a good name for them. I thought about suggesting something like "masterService" or "rpcService"; or "initRPCserver". But I'm not sure if those are good names either, and I didn't want to nudge anyone in a particular direction.

I know they are not threads, though.

cclerget: @gvallee maybe just describe what they do ?

```
go createContainer(rpcSocket, containerPid, engine, fatalChan)
go startContainer(masterSocket, isInstance, containerPid, ppid, engine, fatalChan)
```

mem: that's good from my point of view

cclerget (commented):
Changes committed directly in @gvallee branch after slack discussion. Waiting @mem feedbacks

internal/app/starter/master_linux_test.go
mem: Maybe add a comment?

internal/app/starter/master_linux_test.go
mem: Sorry, I don't see a response or a code change as suggested. Can you please follow the provided link?

internal/app/starter/master_linux_test.go
mem: and here the same comment: can you please follow the provided link?

internal/app/starter/master_linux_test.go
mem: I'm still waiting on an action regarding the original request.

mem (changes requested):

internal/app/starter/master_linux_test.go
mem: Let me try yet again: if I look at this test, it's not obvious that it will stop (i.e. that something is sent over `fatalChan`, partly because the name "fatalChan" suggest that would be the case only if a failure occurs).

Can you please add a comment clarifying that the test cases are such that there's always going to be a failure (both cases have `shallPass` set to `false`; someone reading this code might think there are other cases where `shallPass` is set to `true`) or that `createContainer` always sends something on `fatalChan` no matter what or whatever it is that guarantees that this test will not get stuck waiting for a value on `fatalChan`?

internal/app/starter/master_linux_test.go
mem: good enough

mem (changes requested):

internal/app/starter/master_linux.go
```diff
@@ -11,129 +11,131 @@ import (
...
+	}
+
+	err = e.PostStartProcess(containerPid)
+	if err != nil {
+		fatalChan <- fmt.Errorf("post start process failed: %s", err)
+		return
+	}
+}
+
+// Master initializes a runtime engine and runs it
```
sashayakovtseva:
```suggestion
// Master initializes a runtime engine and runs it.
```

sashayakovtseva:
@gvallee Can you update your branch, please?

internal/app/starter/master_linux.go
```diff
@@ -11,129 +11,131 @@ import (
...
+	}
+	rpcConn, err := net.FileConn(comm)
+	comm.Close()
+	if err != nil {
+		fatalChan <- fmt.Errorf("failed to copy unix socket descriptor: %s", err)
+		return
+	}
 
-	ppid := os.Getppid()
+	err = e.CreateContainer(containerPid, rpcConn)
```
sashayakovtseva, line 34: Why there is no `runtime.LockOSThread` any more? 
In comments above I see you had some discussion with @cclerget in slack, was this discussed as well?

internal/app/starter/master_linux.go
```diff
@@ -11,129 +11,131 @@ import (
...
 
-	go func() {
-		comm := os.NewFile(uintptr(rpcSocket), "socket")
-		rpcConn, err := net.FileConn(comm)
-		comm.Close()
-		if err != nil {
-			fatalChan <- fmt.Errorf("failed to copy unix socket descriptor: %s", err)
-			return
-		}
+	rpcConn.Close()
```
sashayakovtseva, line 40: Should this call be in defer? Just like in `startContainer`

internal/app/starter/master_linux.go
```diff
@@ -11,129 +11,131 @@ import (
...
-		runtime.Goexit()
-	}()
+	data := make([]byte, 1)
 
-	go func() {
-		data := make([]byte, 1)
-		comm := os.NewFile(uintptr(masterSocket), "master-socket")
-		conn, err := net.FileConn(comm)
-		comm.Close()
+	// special path for engines which needs to stop before executing
```
sashayakovtseva, line 59:
```suggestion
	// special path for engines which need to stop before executing
```

internal/app/starter/master_linux.go
```diff
@@ -11,129 +11,131 @@ import (
...
+	// container process
+	if obj, ok := e.Operations.(interface {
+		PreStartProcess(int, net.Conn, chan error) error
```
sashayakovtseva, line 62: Since this is a cleanup, maybe worth putting it into a separate type? e.g. function scoped interface

internal/app/starter/master_linux.go
```diff
@@ -11,129 +11,131 @@ import (
...
-			} else if data[0] == 'f' {
-				// StartProcess reported an error in stage 2, don't send error via
-				// fatalChan, error will be reported by stage 2 and the process
-				// status will be set accordingly via MonitorContainer method below
-				sylog.Debugf("stage 2 process reported an error, waiting status")
-				return
-			}
-			if err := obj.PreStartProcess(containerPid, conn, fatalChan); err != nil {
-				fatalChan <- fmt.Errorf("pre start process failed: %s", err)
+
```
sashayakovtseva:
```suggestion
```

internal/app/starter/master_linux.go
```diff
@@ -11,129 +11,131 @@ import (
...
-		// byte sent is equal to 'f', it means an error occurred in
-		// StartProcess, just return by waiting error and process status
-		_, err = conn.Read(data)
-		if (err != nil && err != io.EOF) || data[0] == 'f' {
+			// EOF means something goes wrong in stage 2, don't send error via
+			// fatalChan, error will be reported by stage 2 and the process
+			// status will be set accordingly via MonitorContainer method below
+			sylog.Debugf("stage 2 process was interrupted, waiting status")
+			return
+		} else if data[0] == 'f' {
```
sashayakovtseva, line 75: Can be simplified to `if data[0] == 'f'` since you have a return statement in previous if clause

internal/app/starter/master_linux.go
```diff
@@ -11,129 +11,131 @@ import (
...
+	_, err = conn.Read(data)
+	if (err != nil && err != io.EOF) || data[0] == 'f' {
+		sylog.Debugf("stage 2 process reported an error, waiting status")
+		return
+	}
+
+	err = e.PostStartProcess(containerPid)
+	if err != nil {
+		fatalChan <- fmt.Errorf("post start process failed: %s", err)
+		return
```
sashayakovtseva, line 100: No need for this return, however for explicitness it is fine

sashayakovtseva: Also a question, why `if isInstance && os.Getppid() == ppid` check is removed?

internal/app/starter/master_linux.go
```diff
@@ -11,129 +11,131 @@ import (
...
 		var err error
 		status, err = e.MonitorContainer(containerPid, signals)
 		fatalChan <- err
 	}()
 
-	fatal = <-fatalChan
+	fatal := <-fatalChan
 
-	runtime.LockOSThread()
 	if err := e.CleanupContainer(fatal, status); err != nil {
```
sashayakovtseva, line 127: Same question, why removing `runtime.LockOSThread`?

internal/pkg/runtime/engines/singularity/process_linux.go
```diff
@@ -425,7 +426,17 @@ func (engine *EngineOperations) PostStartProcess(pid int) error {
 			return err
 		}
 
-		return file.Update()
+		err = file.Update()
+
+		// send SIGUSR1 to the parent process in order to tell it
+		// to detach container process and run as instance.
+		// Sleep a bit in case child would exit
+		time.Sleep(100 * time.Millisecond)
```
sashayakovtseva, line 530: Okay, looks like you and Cedric decided to move this here and thus `if isInstance && os.Getppid() == ppid` is removed in `master_linux.go`

internal/app/starter/master_linux_test.go
```diff
@@ -0,0 +1,118 @@
...
+			// createContainer is creating a separate thread and we sync with that
+			// thread through a channel similarly to the createContainer function itself,
+			// as well as the Master function.
+			// For more details, please refer to the master_linux.go code.
+			fatal = <-fatalChan
+			if tt.shallPass && fatal != nil {
+				t.Fatalf("test %s expected to succeed but failed: %s", tt.name, fatal)
+			} else if !tt.shallPass && fatal == nil {
+				t.Fatalf("test %s expected to fail but succeeded", tt.name)
+			} else if tt.shallPass && fatal == nil {
```
sashayakovtseva, line 64: Why having this elses? 
IMO [testify](https://github.com/stretchr/testify) is a perfect fit here.

internal/app/starter/master_linux_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			go startContainer(tt.masterSocket, tt.containerPid, tt.engine, fatalChan)
+			fatal = <-fatalChan
+			if tt.shallPass && fatal != nil {
+				t.Fatalf("test %s expected to succeed but failed: %s", tt.name, fatal)
+			} else if !tt.shallPass && fatal == nil {
+				t.Fatalf("test %s expected to fail but succeeded", tt.name)
+			} else if tt.shallPass && fatal == nil {
```
sashayakovtseva, line 111: Same as above, testify will same time and make code cleaner IMO

internal/app/starter/master_linux.go
gvallee: That is actually incorrect, the subject is `special path` so it should be `needs` not `need` or it should be `special paths`

internal/app/starter/master_linux.go
gvallee: This is outside of the scope of this PR, which focuses on unit tests. I am not saying that we should not open an issue.

internal/pkg/runtime/engines/singularity/process_linux.go
gvallee: @sasha I have nothing to do with it, I only wrote the unit test.

gvallee: And I do not understand what this comment is about. Is it a question? A suggestion?

internal/app/starter/master_linux_test.go
gvallee: @sasha Because it was required by others on the team regarding ALL unit tests in the ENTIRE codebase. It is supposed to improve readability and this topic is totally out-of-scope of this PR.

internal/app/starter/master_linux_test.go
gvallee: @sasha this type of code organization has been requested by others; you ask for the opposite. Please sync with the other members of the team, decide what to do, but please do *not* PRs for these reasons.

internal/app/starter/master_linux.go
gvallee: The linter did not complain about this, I believe this is totally minor based on personal preferences.

internal/app/starter/master_linux.go
gvallee: @sashayakovtseva  What is the request here? Add comments?

internal/app/starter/master_linux.go
gvallee: @sashayakovtseva This is incorrect, I did not discuss this point. The initial comment was that a reviewer (I do not remember who and it does not matter) did not like the naming I selected. @cclerget made a pass over the code to cleanup/ease the implementation of unit tests.
But are you asking here? The addition of comments?

internal/app/starter/master_linux.go
gvallee: @sashayakovtseva I do not understand your request here, are you asking to add comments?

internal/app/starter/master_linux.go
sashayakovtseva: ah, that changes everything... I thought subject is ‘engines’  which should be stopped before start in order to execute prestart

sashayakovtseva (approved):

internal/app/starter/master_linux.go
jscook2345: Agreed. Should this be a deferred action?

mem (approved):

internal/app/starter/master_linux.go
cclerget: The connection to RPC is closed only when container creation succeed in order to not execute stage 2 process (container process) in case of failure, then the master process will automatically killed the RPC process at exit

PR #3801, opened 2019-06-26
Support inspect/run-help commands for unprivileged installation

### LINKED ISSUE #3779: Problem with inspect command with singularity `--without-suid` installation
## Version of Singularity:

v3.2.1
v3.3.0-rc.1

### Expected behavior

I do run the command
```
singularity inspect test.sif 
```

### Actual behavior
I got the message:

```
FATAL:   Unable to prossess command: fork/exec /home/alazzaro/Work/containers/singularity/singularity_install_v3.3.0-rc.1/libexec/singularity/bin/starter-suid: no such file or directory: 
```

- Install on Ubuntu 19.04
- Install singularity rootless: `./mconfig --without-suid --prefix=${SINGULARITY} -c "gcc" -x "g++"`

### LINKED ISSUE #3996: singularity inspect returns "FATAL:   exit status 255: "
## Version of Singularity:

singularity version 3.2.1-1.el7a

### Expected behavior

Output of container metadata or meaningfull error message

### Actual behavior

Singularity prints `FATAL:   exit status 255: `, nothing more

Run `singularity inspect <myimage.sif>` on a container.

Happens only for some containers build by users on our HPC cluster. On the same image I get `FATAL:   container creation failed: mount /proc/self/fd/12->/var/singularity/mnt/session/rootfs error: can't mount image /proc/self/fd/12: failed to mount squashfs filesystem: invalid argument` so it might be a faulty image. But the non-existing inspect output makes it impossible to help them

cclerget:

This PR does :
- adds a new compilation config constant `SINGULARITY_SUID` to ease detection of setuid installation at execution
- adds `LookStarterPath` to return the location of starter binary based on runtime/configuration requirements
- CLI code was a bit refactored to remove some code redundancy and also fixes `run-help` command suffer from same issue as `inspect` with unprivileged installation

- Fixes #3779
- Fixes #3996 


cmd/internal/cli/actions_linux.go
```diff
@@ -124,20 +75,39 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
...
+	if err != nil {
+		sylog.Fatalf("%s", err)
 	}
 
 	ociConfig := &oci.Config{}
 	generator := generate.Generator{Config: &ociConfig.Spec}
 
 	engineConfig.OciConfig = ociConfig
 
+	allowedSuid := engineConfig.File.AllowSetuid
```
mem, line 89: `allowSuid` instead of `allowedSuid` (just like the original field)

```suggestion
	allowSuid := engineConfig.File.AllowSetuid
```

cmd/internal/cli/actions_linux.go
```diff
@@ -124,20 +75,39 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
...
+	// users need to use setuid workflow if they haven't requested
+	// a user namespace or if they want to use fakeroot feature
+	requireSuid := (!UserNamespace || IsFakeroot) && !isPrivileged
+	starterBinary, starterSuid := exec.LookStarterPath(requireSuid, allowedSuid)
+	if !starterSuid && requireSuid {
```
mem, line 94: I'm trying to unravel the logic here:

- starterSuid means "starter with the setuid bit is available from the installation", right?
- requireSuid means "the requirested operation requires privilege, and I need to get it from starter-suid becasue I don't have it right now", right?

so this if means "I don't have starter-setuid, but I need privilege", why is it asking if setuid is allowed? If there's no starter-suid, now is it going to get it?

I wrote the above without looking at what exec.LookStarterPath does, to show where the miscommunication can happen. My guess is the problem is with the "starterSuid" name. But that's a guess.

My first impulse was to ask this whole block to be moved to its own function, requiring four arguments (allowSetuid, isPrivileged, userNS, fakeroot) but I'm not sure if that helps the situation.

cmd/internal/cli/common_linux.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/sylabs/singularity/internal/pkg/util/fs"
+	"github.com/sylabs/singularity/pkg/image"
+	"github.com/sylabs/singularity/pkg/image/unpacker"
+	singularityConfig "github.com/sylabs/singularity/pkg/runtime/engines/singularity/config"
+)
+
+// singularityConfPath holds the path to singularity.conf file
+var singularityConfPath = filepath.Join(buildcfg.SYSCONFDIR, "singularity/singularity.conf")
+
+// convertSquashfsToSandbox take a single image file in SIF or squashfs format
```
mem, line 30:
```suggestion
// convertSquashfsToSandbox takes a single image file in SIF or squashfs format,
```

cmd/internal/cli/common_linux.go
```diff
@@ -0,0 +1,157 @@
...
+// to convert it to a sandbox image and returns the path to the sandbox.
```
mem, line 31:
```suggestion
// converts it to a sandbox image and returns the path to the resulting sandbox.
```

cmd/internal/cli/common_linux.go
```diff
@@ -0,0 +1,157 @@
...
+func convertSquashfsToSandbox(filename string, unsquashfsPath string) (string, error) {
```
mem, line 32:
```suggestion
func convertSquashfsToSandbox(filename, unsquashfsPath string) (string, error) {
```

cmd/internal/cli/common_linux.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/sylabs/singularity/internal/pkg/sylog"
+	"github.com/sylabs/singularity/internal/pkg/util/exec"
+	"github.com/sylabs/singularity/internal/pkg/util/fs"
+	"github.com/sylabs/singularity/pkg/image"
+	"github.com/sylabs/singularity/pkg/image/unpacker"
+	singularityConfig "github.com/sylabs/singularity/pkg/runtime/engines/singularity/config"
+)
+
+// singularityConfPath holds the path to singularity.conf file
+var singularityConfPath = filepath.Join(buildcfg.SYSCONFDIR, "singularity/singularity.conf")
```
mem, line 28: this looks like something better added to pkg/syfs as it looks like it's used in multiple locations in the code.

cmd/internal/cli/common_linux.go
```diff
@@ -0,0 +1,157 @@
...
+		return "", fmt.Errorf("could not open image %s: %s", filename, err)
+	}
+	defer img.File.Close()
+
+	if !img.HasRootFs() {
+		return "", fmt.Errorf("no root filesystem found in %s", filename)
+	}
+
+	// squashfs only
+	if img.Partitions[0].Type != image.SQUASHFS {
```
mem, line 44: the code is such that `img.HasRootFs()` above will look at all the partitions looking for one that is idenfitied as the root filesystem, but this part here assumes that said rootfs is the first partition. I don't see anything in the code that guarantees that that partition is going to be 0.

I think to be on the safe side, a `func (*Image) GetRootFs() *Partition` method should be added that returns `nil` if there's no root partition, and returns the pointer to the first partition it finds identified as root. Since the call on line 49 also needs this 0, the code probably needs an index instead of the pointer to the partition.

cmd/internal/cli/common_linux.go
```diff
@@ -0,0 +1,157 @@
...
+		return "", fmt.Errorf("not a squashfs root filesystem")
+	}
+
+	// create a reader for rootfs partition
+	reader, err := image.NewPartitionReader(img, "", 0)
+	if err != nil {
+		return "", fmt.Errorf("could not extract root filesystem: %s", err)
+	}
+	s := unpacker.NewSquashfs()
+	if !s.HasUnsquashfs() && unsquashfsPath != "" {
```
mem, line 54: I guess this is fine because an error will be reported on line 71, but if `s.HasUnquashfs() == false && unsquashfsPath == ""`, all the work between here and line 71 will be discarded.

cmd/internal/cli/common_linux.go
```diff
@@ -0,0 +1,157 @@
...
+		s.UnsquashfsPath = unsquashfsPath
+	}
+
+	// keep compatibility with v2
+	tmpdir := os.Getenv("SINGULARITY_LOCALCACHEDIR")
```
mem, line 59: I think this should be moved to `internal/pkg/client/cache`. The current code there does not know about `SINGULARITY_LOCALCACHEDIR`, and the string `"SINGULARITY_CACHEDIR"` is `cache.DirEnv`.

cmd/internal/cli/common_linux.go
```diff
@@ -0,0 +1,157 @@
...
+		os.RemoveAll(dir)
+		return "", fmt.Errorf("root filesystem extraction failed: %s", err)
+	}
+
+	return dir, err
+}
+
+// getfileConfig returns singularity configuration read from singularity.conf
+// file.
+func getFileConfig() (*singularityConfig.FileConfig, error) {
```
mem, line 81: this looks like something that should be part of `pkg/runtime/engines/singularity/config`, no?

cmd/internal/cli/common_linux.go
```diff
@@ -0,0 +1,157 @@
...
+	fileConfig := new(singularityConfig.FileConfig)
+
+	if err := config.Parser(singularityConfPath, fileConfig); err != nil {
+		return nil, fmt.Errorf("unable to parse %s file: %s", singularityConfPath, err)
+	}
+	return fileConfig, nil
+}
+
+// genericStarterCommand executes a single command inside container image with
```
mem, line 90: this sounds useful, but it should not be part of the cli package.

cmd/starter/c/starter.c
```diff
@@ -888,6 +888,10 @@ __attribute__((constructor)) static void init(void) {
 
     /* temporarily drop privileges while running as setuid */
     if ( sconfig->starter.isSuid ) {
+        /* value from the generated config.h file */
+        if ( SINGULARITY_SUID == 0 ) {
+            fatalf("You could not run setuid worfklow if Singularity was compiled with '--without-suid'\n");
```
mem, line 893: I'm going to guess the user is not going to be able to do anything with this message.

How does this error happen? Singularity was built using `--without-suid` and somehow `starter` now has the setuid bit set (because there should be no starter-suid in this case, right?)

internal/pkg/util/exec/starter_linux.go
```diff
@@ -0,0 +1,126 @@
...
+
+	"github.com/sylabs/singularity/internal/pkg/buildcfg"
+	"github.com/sylabs/singularity/internal/pkg/runtime/engines/config"
+	"github.com/sylabs/singularity/internal/pkg/sylog"
+)
+
+// LookStarterPath returns the starter binary path and also returns
+// if the path point to the setuid starter executable
+func LookStarterPath(suid bool, allowedSuid bool) (string, bool) {
+	starterPath := filepath.Join(buildcfg.LIBEXECDIR, "singularity/bin/starter")
```
mem, line 24: this should be in syfs as StarterPath() and StarterSetuidPath()

internal/app/singularity/oci_create_linux.go
```diff
@@ -22,7 +21,7 @@ import (
 
 // OciCreate creates a container from an OCI bundle
 func OciCreate(containerID string, args *OciArgs) error {
-	starter := buildcfg.LIBEXECDIR + "/singularity/bin/starter"
+	starterBinary, _ := exec.LookStarterPath(false, false)
```
mem, line 24: For readability's sake, you should add a function in that exec package that I'm guessing should be called `GetStarterPathNoSuid()`

internal/app/singularity/oci_exec_linux.go
```diff
@@ -6,22 +6,20 @@
...
-	"github.com/sylabs/singularity/internal/pkg/buildcfg"
 	"github.com/sylabs/singularity/internal/pkg/runtime/engines/oci"
 	"github.com/sylabs/singularity/internal/pkg/sylog"
 	"github.com/sylabs/singularity/internal/pkg/util/exec"
 )
 
 // OciExec executes a command in a container
 func OciExec(containerID string, cmdArgs []string) error {
-	starter := buildcfg.LIBEXECDIR + "/singularity/bin/starter"
+	starterBinary, _ := exec.LookStarterPath(false, false)
```
mem, line 22: same here

internal/pkg/build/build.go
```diff
@@ -251,19 +249,26 @@ func engineRequired(def types.Definition) bool {
...
 	env := []string{sylog.GetEnvVar()}
-	starter := filepath.Join(buildcfg.LIBEXECDIR, "/singularity/bin/starter")
-	if b.Opts.Fakeroot {
-		starter = filepath.Join(buildcfg.LIBEXECDIR, "/singularity/bin/starter-suid")
-		if _, err := os.Stat(starter); os.IsNotExist(err) {
-			return fmt.Errorf("fakeroot feature requires to install Singularity as root")
-		}
+
+	requireSuid := b.Opts.Fakeroot && !isPrivileged
+	// to make it proper, we should read singularity configuration file
```
mem, line 262: if there's an import cycle, it means you need to introduce a new package.

internal/pkg/build/build.go
```diff
@@ -251,19 +249,26 @@ func engineRequired(def types.Definition) bool {
...
 	sylog.Debugf("Starting build engine")
 	env := []string{sylog.GetEnvVar()}
-	starter := filepath.Join(buildcfg.LIBEXECDIR, "/singularity/bin/starter")
-	if b.Opts.Fakeroot {
-		starter = filepath.Join(buildcfg.LIBEXECDIR, "/singularity/bin/starter-suid")
-		if _, err := os.Stat(starter); os.IsNotExist(err) {
-			return fmt.Errorf("fakeroot feature requires to install Singularity as root")
-		}
+
+	requireSuid := b.Opts.Fakeroot && !isPrivileged
```
mem, line 261: deja vu.

Isn't the logic expressed on lines 252-256,261 repeated in a couple of places?

internal/pkg/util/exec/starter_linux.go
```diff
@@ -0,0 +1,126 @@
...
+	"syscall"
+
+	"github.com/sylabs/singularity/internal/pkg/buildcfg"
+	"github.com/sylabs/singularity/internal/pkg/runtime/engines/config"
+	"github.com/sylabs/singularity/internal/pkg/sylog"
+)
+
+// LookStarterPath returns the starter binary path and also returns
+// if the path point to the setuid starter executable
+func LookStarterPath(suid bool, allowedSuid bool) (string, bool) {
```
mem, line 23: this function should not be exported.

Since you have two booleans, strictly speaking you have four functions. That does not seem to be the case, though, as allowedSuid is only used if suid is true, so that's three functions (`(true, true)`, `(true, false)` and `(false, *)`).

The last one is easy: `GetStarterPath`, this calls `getStaterPath(false, false)`.

The other two are `GetStarterSuidPath`, this calls `getStaterPath(true, *)`.

Also, the second return value should be an error: if you ask for starter-suid and you don't get it, you should get an error, not the path to starter.

internal/pkg/util/exec/starter_linux.go
```diff
@@ -0,0 +1,126 @@
...
+		starterSuidPath := fmt.Sprintf("%s-suid", starterPath)
+		if _, err := os.Stat(starterSuidPath); err != nil {
+			// screwed installation, should be pretty rare
+			sylog.Verbosef("Could not use starter suid binary: %s", err)
+			return starterPath, starterSuid
+		}
+		// setuid binary is usable
+		starterPath = starterSuidPath
+		starterSuid = true
+	} else if buildcfg.SINGULARITY_SUID == 0 && suid {
```
mem, line 43: I would move this to the front and do:

```
    if !suid {
        return starter, nil
    }

    if buildcfg.SINGULARITY_SUID == 0 {
        return "", errStarterSuidNotAvailable
    }

    if !allowedSuid {
        return "", errStarterSuidNotAllowed
    }

    if _, err := os.Stat(starterSuid); err != nil {
        return "", errStarterSuidNotFound
    }

    return starterSuid, nil
```

internal/pkg/util/exec/starter_linux.go
```diff
@@ -0,0 +1,126 @@
...
+		Args: args,
+		Env:  env,
+	}
+	return c, nil
+}
+
+// setStarterPipe sets a pipe communication channel for JSON configuration data and returns
+// the file descriptor of the read side
+func setStarterPipe(config *config.Common) (int, error) {
+	fd, err := syscall.Socketpair(syscall.AF_UNIX, syscall.SOCK_STREAM|syscall.SOCK_CLOEXEC, 0)
```
mem, line 87: since you are touching this... would it be possible to add some comments as to why this is using a socket and not a pipe (given how this function is called "setStarterPipe")?

I'm guessing it's because the communication is bidirectional, but then the question becomes why is it bidirectional?

Any internal documentation regarding that will be appreciated.

mem (changes requested):

WestleyK (commented):
Ping @cclerget, whats the status on this?

cclerget:
@WestleyK I was thinking about closing it and rewrite a new proper one while addressing @mem comments. If it's a blocker for you or conflicts with your work, don't pay attention to this PR, I will rework it later.

PR #3803, opened 2019-06-26
Fix user namespace bind source located on a read-only filesystem

### LINKED ISSUE #3799: Failure binding ro source when running unprivileged -- can't remount
## Version of Singularity:

3.2.1 and 3.3.0-rc.1

### Expected behavior

Should be able to bind a directory from a read-only filesystem into a container when running unprivileged. 

### Actual behavior

When binding anything from a read-only filesystem and running unprivileged there's a "can't remount" "operation not permitted" error.  This is a regression from singularity-2.6.1.

Run the following or use any container image of your choice:
```
$ sudo mount -t tmpfs -o ro tmpfs /mnt
$ singularity shell -u -B /mnt docker://centos 
FATAL:   container creation failed: mount ->/mnt error: can't remount /mnt: operation not permitted
```

We noticed it first with cvmfs as a mount source but can reproduce it with any filesystem mounted read-only.

@cclerget

cclerget:

This PR prevent to remount user bind source directory located on a RO filesystem when user namespace is in use to avoid failure

- Fixes #3799

### DIFF
--- a/internal/pkg/runtime/engines/singularity/container_linux.go
+++ b/internal/pkg/runtime/engines/singularity/container_linux.go
@@ -540,6 +540,25 @@ func (c *container) mountGeneric(mnt *mount.Point) (err error) {
 		}
 	}
 	_, err = c.rpcOps.Mount(source, dest, mnt.Type, flags, optsString)
+	// when using user namespace we always try to apply mount flags with
+	// remount, then if we get a permission denied error, we continue
+	// execution by ignoring the error and warn user if the bind mount
+	// need to be mounted read-only
+	if remount && err != nil && c.userNS {
+		// this needs to compare error strings because we are dropping
+		// the original error value in the RPC call. In order to keep
+		// the original error values, we would have to add them to the
+		// corresponding struct passed over RPC and add the necessary
+		// GOB encoding / decoding functions.
+		if err.Error() == syscall.Errno(syscall.EPERM).Error() {
+			if flags&syscall.MS_RDONLY != 0 {
+				sylog.Warningf("Could not remount %s read-only: %s", dest, err)
+			} else {
+				sylog.Verbosef("Could not remount %s: %s", dest, err)
+			}
+			return nil
+		}
+	}
 	return err
 }
 
@@ -1292,13 +1311,14 @@ func (c *container) addUserbindsMount(system *mount.System) error {
 	devicesMounted := 0
 	devPrefix := "/dev"
 	userBindControl := c.engine.EngineConfig.File.UserBindControl
-	flags := uintptr(syscall.MS_BIND | c.suidFlag | syscall.MS_NODEV | syscall.MS_REC)
+	defaultFlags := uintptr(syscall.MS_BIND | c.suidFlag | syscall.MS_NODEV | syscall.MS_REC)
 
 	if len(c.engine.EngineConfig.GetBindPath()) == 0 {
 		return nil
 	}
 
 	for _, b := range c.engine.EngineConfig.GetBindPath() {
+		flags := defaultFlags
 		splitted := strings.Split(b, ":")
 
 		src, err := filepath.Abs(splitted[0])
@@ -1348,13 +1368,12 @@ func (c *container) addUserbindsMount(system *mount.System) error {
 
 		sylog.Debugf("Adding %s to mount list\n", src)
 
-		if err := system.Points.AddBind(mount.UserbindsTag, src, dst, flags); err != nil && err == mount.ErrMountExists {
+		if err := system.Points.AddBind(mount.UserbindsTag, src, dst, flags); err == mount.ErrMountExists {
 			sylog.Warningf("destination %s already in mount list: %s", src, err)
 		} else if err != nil {
 			return fmt.Errorf("unable to add %s to mount list: %s", src, err)
 		} else {
 			system.Points.AddRemount(mount.UserbindsTag, dst, flags)
-			flags &^= syscall.MS_RDONLY
 		}
 	}
 


internal/pkg/runtime/engines/singularity/container_linux.go
```diff
@@ -1339,13 +1341,23 @@ func (c *container) addUserbindsMount(system *mount.System) error {
 
 		sylog.Debugf("Adding %s to mount list\n", src)
 
+		if c.userNS {
+			// test if the source directory is a read-only filesystem by
+			// asking for write permission for this directory, if EROFS error
+			// is returned, we need to skip remount to avoid failure
+			if err := syscall.Access(src, 0x2); err.(syscall.Errno) == syscall.EROFS {
```
DrDaveD: Since it is running unprivileged, if the source directory is simply owned by root, won't this treat it as a read-only filesystem?  On the other hand, maybe it doesn't matter, because in that case it won't try to remount.  Or will it?  The remount seems to otherwise be always tried when AddBind() below returns err == nil.

For that matter, why do this check here and have a remount flag at all?  Why not do the check down below, just before the AddRemount, only when it is actually needed?

internal/pkg/runtime/engines/singularity/container_linux.go
```diff
@@ -1283,13 +1283,15 @@ func (c *container) addUserbindsMount(system *mount.System) error {
...
-	flags := uintptr(syscall.MS_BIND | c.suidFlag | syscall.MS_NODEV | syscall.MS_REC)
+	defaultFlags := uintptr(syscall.MS_BIND | c.suidFlag | syscall.MS_NODEV | syscall.MS_REC)
 
 	if len(c.engine.EngineConfig.GetBindPath()) == 0 {
 		return nil
 	}
 
 	for _, b := range c.engine.EngineConfig.GetBindPath() {
+		flags := defaultFlags
+		remount := true
```
DrDaveD: I suggest calling this something like "remountok" because it doesn't always result in a remount, it just allows it.

internal/pkg/runtime/engines/singularity/container_linux.go
```diff
@@ -1339,13 +1341,23 @@ func (c *container) addUserbindsMount(system *mount.System) error {
...
+		if c.userNS {
+			// test if the source directory is a read-only filesystem by
+			// asking for write permission for this directory, if EROFS error
+			// is returned, we need to skip remount to avoid failure
+			if err := syscall.Access(src, 0x2); err.(syscall.Errno) == syscall.EROFS {
+				remount = false
+			}
+		}
+
 		if err := system.Points.AddBind(mount.UserbindsTag, src, dst, flags); err != nil && err == mount.ErrMountExists {
```
DrDaveD: This comment is for the unchanged AddBind() line: the err != nil check is redundant, because if it is equal to mount.ErrMountExists it is obviously not nil.

DrDaveD (changes requested):

internal/pkg/runtime/engines/singularity/container_linux.go
DrDaveD: In version 2.6.1, the [corresponding code](https://github.com/sylabs/singularity/blob/vault/release-2.6/src/lib/runtime/mounts/userbinds/userbinds.c#L143) only sets read_only when the user requests it and then [skips the remount with a warning](https://github.com/sylabs/singularity/blob/vault/release-2.6/src/lib/runtime/mounts/userbinds/userbinds.c#L143).  It does not, as far as I see, detect anywhere that the source filesystem is read-only and try to copy that.  How's that being detected now?  Even with this patch if the user requests read only it will fail rather than warn.

In a couple of other places in this same source file, when mounting /proc and /sys, it skips the remount whenever c.userNS is set.  Why not do it the same way?  And there are a number of other places where AddRemount is called (e.g. in addHostMount) that do not check c.userNS; won't some of them have this same problem?

internal/pkg/runtime/engines/singularity/container_linux.go
cclerget: Should go away by moving check before `AddRremount`

internal/pkg/runtime/engines/singularity/container_linux.go
cclerget: `EROFS` will take precedence over `EPERM`, that's why `EPERM` is ignored here because it means the directory is owned by another user and it doesn't matter in this case.

Regarding 2.6.1 behavior, it was a lazy approach, IMO when we can we should apply mount flag correctly even with user namespace enabled (users may want to bind a directory in read-only).

I will apply fix for others. Thanks !

internal/pkg/runtime/engines/singularity/container_linux.go
```diff
@@ -531,6 +531,15 @@ func (c *container) mountGeneric(mnt *mount.Point) (err error) {
 		}
 	}
 	_, err = c.rpcOps.Mount(source, dest, mnt.Type, flags, optsString)
+	// when using user namespace we always try to apply mount flags,
+	// then if we get a permission denied error, we contine execution
+	// by ignoring the error and warn user
+	if remount && err != nil && c.userNS {
+		if err.Error() == syscall.Errno(syscall.EPERM).Error() {
+			sylog.Warningf("Could not remount %s: %s", dest, err)
```
DrDaveD: I don't think this should be a warning; verbose should be enough (and matching the behavior of 2.6), unless perhaps the remount is because the user explicitly asked for it with a :ro in a --bind option.

Are there any reasons that err might be EPERM other than a remount failure?

internal/pkg/runtime/engines/singularity/container_linux.go
```diff
@@ -531,6 +531,15 @@ func (c *container) mountGeneric(mnt *mount.Point) (err error) {
 		}
 	}
 	_, err = c.rpcOps.Mount(source, dest, mnt.Type, flags, optsString)
+	// when using user namespace we always try to apply mount flags,
+	// then if we get a permission denied error, we contine execution
+	// by ignoring the error and warn user
+	if remount && err != nil && c.userNS {
```
DrDaveD, line 547: Note that err != nil is redundant here again.

cclerget: this is required otherwise the call `err.Error` will panic when `err` is nil

internal/pkg/runtime/engines/singularity/container_linux.go
cclerget: For `EPERM` condition, this is a pretty defensive approach instead of ignoring all error, other errors could be caused by a dev mistake or something irregular

internal/pkg/runtime/engines/singularity/container_linux.go
```diff
@@ -531,6 +531,20 @@ func (c *container) mountGeneric(mnt *mount.Point) (err error) {
 		}
 	}
 	_, err = c.rpcOps.Mount(source, dest, mnt.Type, flags, optsString)
+	// when using user namespace we always try to apply mount flags with
+	// remount, then if we get a permission denied error, we contine
```
DrDaveD: CI scripts are complaining about the misspelling of continue in this comment.

DrDaveD (approved):
Other than the complaint about the misspelled comment, this looks OK to me now.  There is now only the verbose message when bind-mounting from a read-only filesystem when running unprivileged.

There is a mystery, however: if I add the ":ro" flag to a bind mount (whether the source is a read-only filesystem or read-write filesystem), this adds the MS_RDONLY flag and then the remount does not fail (there is no warning message).  I don't understand what the remount is attempting to do without the MS_RDONLY flag then, nor why it is failing.  This suggests that there may be another way to solve this that does not cause a remount failure.

internal/pkg/runtime/engines/singularity/container_linux.go
```diff
@@ -540,6 +540,20 @@ func (c *container) mountGeneric(mnt *mount.Point) (err error) {
 		}
 	}
 	_, err = c.rpcOps.Mount(source, dest, mnt.Type, flags, optsString)
+	// when using user namespace we always try to apply mount flags with
+	// remount, then if we get a permission denied error, we continue
+	// execution by ignoring the error and warn user if the bind mount
+	// need to be mounted read-only
+	if remount && err != nil && c.userNS {
+		if err.Error() == syscall.Errno(syscall.EPERM).Error() {
```
mem, line 553: this is comparing the error strings. This should be done only as a last resort...

You can type-assert err (`err.(syscall.Errno)`) and then compare that with `syscall.EPERM`.

mem (changes requested):

internal/pkg/runtime/engines/singularity/container_linux.go
cclerget: It's the last resort, here `err` is of type `ServerError`

DrDaveD:
The travis failure is unrelated to the change and transient and so could just be retried.

If MS_RDONLY is false, why is singularity even attempting a remount?  Is the failure then really because the kernel thinks it is trying to remount a read-only source as read-write, and that's not allowed when unprivileged?

cclerget:
> If MS_RDONLY is false, why is singularity even attempting a remount?

To also apply flags like MS_NOSUID, MS_NODEV ...

> Is the failure then really because the kernel thinks it is try to remount a read-only source as read-write, and that's not allowed when unprivileged?

Unprivileged you need to pass some of original flags (like MS_RDONLY) if you want to apply additional flags, so if source is read-only, you need to pass `-B /mnt:/mnt:ro` to add MS_RDONLY and if you want to also add `MS_NOSUID|MS_NODEV` with remount.

It would fail too :
```
mount -t tmpfs tmpfs -onoexec /mnt
singularity exec -u -B /mnt docker://alpine
```

DrDaveD:
> so if source is read-only, you need to pass -B /mnt:/mnt:ro to add MS_RDONLY and if you want to also add MS_NOSUID|MS_NODEV with remount.

We can't expect users to add :ro, but should singularity be detecting if the source is mounted read-only and if so add the MS_RDONLY flag?

cclerget:
> We can't expect users to add :ro, but should singularity be detecting if the source is mounted read-only and add the MS_RDONLY flag?

Yes we should, but definitively not in this PR, it involves too much changes and also need to handle MS_NOEXEC case

DrDaveD:
> Yes we should, but definitively not in this PR

That makes sense.  Shall we consider this version final then for 3.3.0?

cclerget:
> Shall we consider this version final then for 3.3.0?

Yep, re-requested review from @mem

internal/pkg/runtime/engines/singularity/container_linux.go
mem, line 553:
```suggestion
		// this needs to compare error strings because we are dropping
		// the original error value in the RPC call. In order to keep
		// the original error values, we would have to add them to the
		// corresponding struct passed over RPC and add the necessary
		// GOB encoding / decoding functions.
		if err.Error() == syscall.Errno(syscall.EPERM).Error() {
```

mem (approved):
Thanks!

DrDaveD:
The failing test in travis appears to be intermittent, please re-run the test.

PR #3807, opened 2019-06-26
Skip e2e tests for pull requests based on their commit range

### LINKED ISSUE #3430: When changes are made to the e2e directory, run all e2e tests in CI to verify
Currently e2e tests are only run when master is merged to. It would be good if a change was made to the `e2e/` directory contents that we run all the e2e tests to verify they still work.

### LINKED ISSUE #3806: When changes are made to the vendor directory, run all e2e tests in CI to verify
Currently e2e tests are only run when master is merged to. It would be good if a change was made to the vendor/ directory contents that we run all the e2e tests to verify they still work.

jscook2345:
We only want to run e2e tests under the following condtions:

* Changes are made to the 'e2e/' directory
* Changes are made to the 'vendor' directory
* Merging a pull request into 'master'

Using the 'compare-url' circleci orb, we can get a commit range to diff
against. In the future this feature may be integrated into circleci by
default so this can be removed than.

* Fixes #3430 
* Fixes #3806

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2.1
+version: 2.0
 
 orbs:
   circle-compare-url: iynere/compare-url@0.4.8
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,7 @@
-version: 2.0
+version: 2.1
+
+orbs:
+  circle-compare-url: iynere/compare-url@0.4.8
 
 defaults: &defaults
   working_directory: /go/singularity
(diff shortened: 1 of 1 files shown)


mem (approved):
let's go with this and see if that $COMMIT_RANGE needs fixing.

.circleci/config.yml
```diff
@@ -140,6 +143,20 @@ jobs:
     steps:
       - attach_workspace:
           at: ~/go
+      - compare-url/reconstruct
```
mem: guessing the problem is a missing : here?

PR #3826, opened 2019-06-28
Execute e2e tests inside their own pid and mount namespaces.

cclerget:

- Execute e2e tests inside their own pid and mount namespaces:
  - pid namespace ensure no pending process will live after tests finished
  - mount namespace will allow to manipulate binds without to worry about host mount namespace, and it allow actually to setup temporary home directories
- A temporary filesystem is now setup and home directories are
created in this temporary filesystem and bound to the original
ones, that would ease e2e tests and avoid to screw them up during
tests execution (think about singularity cache, library tokens, PGP
keys ...)
- Remove unnecessary call to `EnsureImage`, cleanup code a bit.

### DIFF
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -7,6 +7,16 @@ package e2e
 
 import (
 	"testing"
+
+	// This import will execute a CGO section with the help of a C
+	// constructor section "init". As we always require to run e2e
+	// tests as root, the C part is responsible of finding the original
+	// user who executes tests; it will also create a dedicated pid
+	// and mount namespace for e2e tests, and will finally restore
+	// identity to the original user but will retain privileges for
+	// Privileged method enabling the execution of a function with root
+	// privileges when required
+	_ "github.com/sylabs/singularity/e2e/internal/e2e/init"
 )
 
 func TestE2E(t *testing.T) {
--- a/e2e/internal/e2e/docker.go
+++ b/e2e/internal/e2e/docker.go
@@ -6,10 +6,12 @@
 package e2e
 
 import (
+	"net"
 	"path/filepath"
 	"sync"
 	"sync/atomic"
 	"testing"
+	"time"
 )
 
 const dockerInstanceName = "e2e-docker-instance"
@@ -49,6 +51,25 @@ func PrepRegistry(t *testing.T, env TestEnv) {
 			WithArgs("-w", "-B", "/sys", dockerImage, dockerInstanceName),
 			ExpectExit(0),
 		)
+
+		// start script in e2e/testdata/Docker_registry.def will listen
+		// on port 5111 once docker registry is up and initialized, so
+		// we are trying to connect to this port until we got a response,
+		// without any response after 10 seconds we abort tests execution
+		// because the start script probably failed
+		retry := 0
+		for {
+			conn, err := net.Dial("tcp", "127.0.0.1:5111")
+			if err == nil {
+				conn.Close()
+				break
+			}
+			time.Sleep(100 * time.Millisecond)
+			retry++
+			if retry == 100 {
+				t.Fatalf("docker registry unreachable after 10 seconds: %s", err)
+			}
+		}
 
 		RunSingularity(
 			t,
--- a/e2e/internal/e2e/home_linux.go
+++ b/e2e/internal/e2e/home_linux.go
@@ -1,0 +1,122 @@
+// Copyright (c) 2019, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package e2e
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"syscall"
+	"testing"
+
+	"github.com/sylabs/singularity/internal/pkg/buildcfg"
+	"github.com/sylabs/singularity/internal/pkg/util/user"
+)
+
+// rpmMacrosContent contains required content to
+// place in $HOME/.rpmmacros file for yum bootstrap
+// build
+var rpmMacrosContent = `
+%_var /var
+%_dbpath %{_var}/lib/rpm
+`
+
+// SetupHomeDirectories creates temporary home directories for
+// privileged and unprivileged users and bind mount those directories
+// on top of real ones. It's possible because e2e tests are executed
+// in a dedicated mount namespace.
+func SetupHomeDirectories(t *testing.T) {
+	var unprivUser, privUser *user.User
+
+	sessionDir := buildcfg.SESSIONDIR
+	unprivUser = CurrentUser(t)
+
+	Privileged(func(t *testing.T) {
+		// there is no cleanup here because everything done (tmpfs, mounts)
+		// in our dedicated mount namespace will be automatically discarded
+		// by the kernel once all test processes exit
+
+		privUser = CurrentUser(t)
+
+		// create the temporary filesystem
+		if err := syscall.Mount("tmpfs", sessionDir, "tmpfs", 0, "mode=0777"); err != nil {
+			t.Fatalf("failed to mount temporary filesystem")
+		}
+
+		// want the already resolved current working directory
+		cwd, err := os.Readlink("/proc/self/cwd")
+		if err != nil {
+			t.Fatalf("could not readlink /proc/self/cwd: %s", err)
+		}
+		unprivResolvedHome, err := filepath.EvalSymlinks(unprivUser.Dir)
+		if err != nil {
+			t.Fatalf("could not resolve %s: %s", unprivUser.Dir, err)
+		}
+		privResolvedHome, err := filepath.EvalSymlinks(privUser.Dir)
+		if err != nil {
+			t.Fatalf("could not resolve %s: %s", privUser.Dir, err)
+		}
+
+		// prepare user temporary homes
+		unprivSessionHome := filepath.Join(sessionDir, unprivUser.Name)
+		privSessionHome := filepath.Join(sessionDir, privUser.Name)
+
+		oldUmask := syscall.Umask(0)
+		defer syscall.Umask(oldUmask)
+
+		if err := os.Mkdir(unprivSessionHome, 0700); err != nil {
+			t.Fatalf("failed to create temporary home %s: %s", unprivSessionHome, err)
+		}
+		if err := os.Chown(unprivSessionHome, int(unprivUser.UID), int(unprivUser.GID)); err != nil {
+			t.Fatalf("failed to set temporary home %s owner: %s", unprivSessionHome, err)
+		}
+		if err := os.Mkdir(privSessionHome, 0700); err != nil {
+			t.Fatalf("failed to create temporary home %s: %s", privSessionHome, err)
+		}
+
+		sourceDir := buildcfg.SOURCEDIR
+
+		// re-create the current source directory if it's located in the user
+		// home directory and bind it. Root home directory is not checked because
+		// the whole test suite can not run from there as we are dropping privileges
+		if strings.HasPrefix(sourceDir, unprivResolvedHome) {
+			trimmedSourceDir := strings.TrimPrefix(sourceDir, unprivResolvedHome)
+			sessionSourceDir := filepath.Join(unprivSessionHome, trimmedSourceDir)
+			if err := os.MkdirAll(sessionSourceDir, 0755); err != nil {
+				t.Fatalf("failed to create temporary home source directory %s: %s", sessionSourceDir, err)
+			}
+			if err := syscall.Mount(sourceDir, sessionSourceDir, "", syscall.MS_BIND, ""); err != nil {
+				t.Fatalf("failed to bind %s to %s: %s", sourceDir, sessionSourceDir, err)
+			}
+		}
+
+		// finally bind temporary homes on top of real ones
+		// in order to not screw them by accident during e2e
+		// tests execution
+		if err := syscall.Mount(unprivSessionHome, unprivResolvedHome, "", syscall.MS_BIND|syscall.MS_REC, ""); err != nil {
+			t.Fatalf("failed to bind %s to %s: %s", unprivSessionHome, unprivResolvedHome, err)
+		}
+		if err := syscall.Mount(privSessionHome, privResolvedHome, "", syscall.MS_BIND, ""); err != nil {
+			t.Fatalf("failed to bind %s to %s: %s", privSessionHome, privResolvedHome, err)
+		}
+		// change to the "new" working directory if above mount override
+		// the current working directory
+		if err := os.Chdir(cwd); err != nil {
+			t.Fatalf("failed to change directory to %s: %s", cwd, err)
+		}
+
+		// create .rpmmacros files for yum bootstrap builds
+		macrosFile := filepath.Join(unprivSessionHome, ".rpmmacros")
+		if err := ioutil.WriteFile(macrosFile, []byte(rpmMacrosContent), 0444); err != nil {
+			t.Fatalf("could not write %s: %s", macrosFile, err)
+		}
+		macrosFile = filepath.Join(privSessionHome, ".rpmmacros")
+		if err := ioutil.WriteFile(macrosFile, []byte(rpmMacrosContent), 0444); err != nil {
+			t.Fatalf("could not write %s: %s", macrosFile, err)
+		}
+	})(t)
+}
--- a/e2e/internal/e2e/init/init_linux.go
+++ b/e2e/internal/e2e/init/init_linux.go
@@ -1,0 +1,168 @@
+// Copyright (c) 2019, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package init
+
+/*
+#define _GNU_SOURCE
+#include <unistd.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+#include <sched.h>
+#include <sys/mount.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+
+#define SIZE    128
+
+// getProcInfo returns the parent PID, UID, and GID associated with the
+// supplied PID.
+static pid_t getProcInfo(pid_t pid, uid_t *uid, gid_t *gid) {
+	FILE *status;
+	char procPath[SIZE];
+	char *line = NULL;
+	size_t len = 0;
+	pid_t ppid = 1;
+
+	memset(procPath, 0, SIZE);
+	if ( snprintf(procPath, SIZE-1, "/proc/%d/status", pid) > SIZE-1 ) {
+		// set returned PID to 1 to trigger error from getUnprivIDs call
+		return 1;
+	}
+
+	status = fopen(procPath, "r");
+	if ( status == NULL ) {
+		// set returned PID to 1 to trigger error from getUnprivIDs call
+		return 1;
+	}
+
+	while ( getline(&line, &len, status) != -1 ) {
+		if ( ppid == 1 ) {
+			sscanf(line, "PPid:\t%d", &ppid);
+		}
+		if ( *uid == 0 ) {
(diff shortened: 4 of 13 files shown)


e2e/internal/e2e/docker.go
```diff
@@ -50,6 +51,13 @@ func PrepRegistry(t *testing.T, env TestEnv) {
 			ExpectExit(0),
 		)
 
+		// sleep a bit to let time to registry to start
+		// and setup everything. The start script already
+		// sleep for 2 seconds before pushing a busybox image in
```
WestleyK: It looks like its sleeping for 5 seconds

```suggestion
		// sleep for 5 seconds before pushing a busybox image in
```

WestleyK (commented):
Looks good to me, just one suggestion...

e2e/internal/e2e/docker.go
cclerget: Oh I refer to this line https://github.com/sylabs/singularity/blob/master/e2e/testdata/Docker_registry.def#L19

cclerget: I will mention the file to avoid confusion

jscook2345:
Can I ask why you want to do this? CircleCI already runs tests in a VM or a container. Is this for the local case where you're running `make test` on your local linux system?

cclerget:
> Can I ask why you want to do this? CircleCI already runs tests in a VM or a container. Is this for the local case where you're running `make test` on your local linux system?

Yes essentially, additionally user/root homes are replaced by an empty one where e2e tests could do what they needs : play with cache, PGP keys, token, running/stopping all instances ..., while this is fine for CI where we start from scratch, this is less good on a dev machine where some image are already cached that I would like to keep, PGP keys that I need to keep, prevent few instances currently running from being stopped by a test executing `instance stop --all`. Also I would like to not mess with host installation and/or other tests by using a custom temporary binding on top `/usr/local/etc/singularity` to test different directives value in `singularity.conf`.

e2e/internal/e2e/docker.go
```diff
@@ -50,6 +51,13 @@ func PrepRegistry(t *testing.T, env TestEnv) {
 			ExpectExit(0),
 		)
 
+		// Sleep a bit to let time to registry to start and setup
+		// everything. The start script already sleep for 2 seconds
+		// (in e2e/testdata/Docker_registry.def) before pushing a
+		// busybox image in the docker registry. So it's safer to
+		// sleep few seconds at this point
+		time.Sleep(5 * time.Second)
```
mem: Sleeping is inherently brittle. Is there something we could try (multiple times, sleeping a little in between if needed) to test that the registry is up or not? From quickly browsing the docs, I see that contacting the login end point should be sufficient, as it should reply with either 200 or 401 if the registry is up (in other words, you don't need to login, you just need to try it).

e2e/internal/e2e/home_linux.go
```diff
@@ -0,0 +1,117 @@
...
+
+// SetupHomeDirectories creates temporary home directories for
+// privileged and unprivileged users and bind mount those directories
+// on top of real ones. It's possible because e2e tests are executed
+// in a dedicated mount namespace.
+func SetupHomeDirectories(t *testing.T) {
+	var unprivUser, privUser *user.User
+
+	sessionDir := buildcfg.SESSIONDIR
+	unprivUser = CurrentUser(t)
```
mem, line 36: seeing how the code below depends on this really being a different user, shouldn't this code assert that 1) this is in fact an unprivileged user; 2) that this is different from the privileged user below?

e2e/internal/e2e/home_linux.go
```diff
@@ -0,0 +1,117 @@
...
+		if err := os.Chown(unprivHome, int(unprivUser.UID), int(unprivUser.GID)); err != nil {
+			t.Fatalf("failed to set temporary home %s owner: %s", unprivHome, err)
+		}
+		if err := os.Mkdir(privHome, 0700); err != nil {
+			t.Fatalf("failed to create temporary home %s: %s", privHome, err)
+		}
+
+		// re-create current working directory if tests are run from the user
+		// home directory and bind it
+		if strings.HasPrefix(cwd, unprivRealHome) {
```
mem: If you where to give this bit of logic a name, it would be easier to read.

Let's see if I can unravel this correctly.

Let's say

```
sessionDir := "/a/b"
unprivUser.Name = "joe"
unprivHome := "/a/b/joe"
unprivRealHome := "/x/y"
cwd := "/x/y/z/1/2/3"    // filepath.Dir(cwd) == "/x/y/z/1/2"
trimmed := "/z/1/2"
path := "/a/b/joe/z/1/2"
```

correct?

so it's extracting (part of) the current working directory relative to the home directory of the current (?) user and reparenting on top of the temporary home?

And why is this necessary only if the current working directory is within the current user's home? (why is there a condition on line 83)

Also, what's special about the directory "3" above? Why it is kept out of this whole thing?

e2e/internal/e2e/home_linux.go
```diff
@@ -0,0 +1,117 @@
...
+			t.Fatalf("failed to bind %s to %s: %s", unprivHome, unprivRealHome, err)
+		}
+		if err := syscall.Mount(privHome, privRealHome, "", syscall.MS_BIND, ""); err != nil {
+			t.Fatalf("failed to bind %s to %s: %s", privHome, privRealHome, err)
+		}
+		if err := os.Chdir(cwd); err != nil {
+			t.Fatalf("failed to change directory to %s: %s", cwd, err)
+		}
+
+		// create .rpmmacros files
```
mem: up to this point everything was pretty generic.

What is this bit about? Seems pretty unrelated to setting up home directories.

e2e/internal/e2e/home_linux.go
```diff
@@ -0,0 +1,117 @@
...
+		// finally bind temporary homes on top of real ones
+		// in order to not screw them by accident during e2e
+		// tests execution
+		if err := syscall.Mount(unprivHome, unprivRealHome, "", syscall.MS_BIND|syscall.MS_REC, ""); err != nil {
+			t.Fatalf("failed to bind %s to %s: %s", unprivHome, unprivRealHome, err)
+		}
+		if err := syscall.Mount(privHome, privRealHome, "", syscall.MS_BIND, ""); err != nil {
+			t.Fatalf("failed to bind %s to %s: %s", privHome, privRealHome, err)
+		}
+		if err := os.Chdir(cwd); err != nil {
```
mem, line 108: oh, this is subtle. Please add a comment as to why this is needed.

e2e/internal/e2e/home_linux.go
```diff
@@ -0,0 +1,117 @@
...
+// privileged and unprivileged users and bind mount those directories
+// on top of real ones. It's possible because e2e tests are executed
+// in a dedicated mount namespace.
+func SetupHomeDirectories(t *testing.T) {
+	var unprivUser, privUser *user.User
+
+	sessionDir := buildcfg.SESSIONDIR
+	unprivUser = CurrentUser(t)
+
+	Privileged(func(t *testing.T) {
```
mem, line 38: just a general comment about this anonymous function: it's doing almost the same thing twice, once for the privileged case and once for the unprivileged case. Since there's no real logic overlap between the two cases, the code would be easier to read if you did one part first (say the privileged one) and the other part second (say the unprivileged one).

e2e/internal/e2e/init/init_linux.go
```diff
@@ -0,0 +1,145 @@
...
+// supplied PID.
+static pid_t getProcInfo(pid_t pid, uid_t *uid, gid_t *gid) {
+	FILE *status;
+	char procPath[SIZE];
+	char *line = NULL;
+	size_t len = 0;
+	pid_t ppid = 1;
+
+	memset(procPath, 0, SIZE);
+	snprintf(procPath, SIZE-1, "/proc/%d/status", pid);
```
mem: this is probably never going to be the case with this code, but just to be correct, please check the return value of snprintf to make sure that the string did in fact fit in the provided buffer.

e2e/internal/e2e/init/init_linux.go
```diff
@@ -0,0 +1,145 @@
...
+	status = fopen(procPath, "r");
+	if ( status == NULL ) {
+		return 1;
+	}
+
+	while ( getline(&line, &len, status) != -1 ) {
+		if ( ppid == 1 ) {
+			sscanf(line, "PPid:\t%d", &ppid);
+		}
+		if ( *uid == 0 ) {
```
mem, line 47: since this code assumes that uid == 0 before entering this function, shouldn't it check that that's the case?

e2e/internal/e2e/init/init_linux.go
```diff
@@ -0,0 +1,145 @@
...
+			sscanf(line, "Uid:\t%d", uid);
+		}
+		if ( *gid == 0 ) {
```
mem, line 50: same here

e2e/internal/e2e/init/init_linux.go
```diff
@@ -0,0 +1,145 @@
...
+	free(line);
+	fclose(status);
+
+	return ppid;
+}
+
+// getUnprivIDs searches recursively up the process parent chain to find a
+// process with a non-root UID, then returns the UID and GID of that process.
+static int getUnprivIDs(pid_t pid, uid_t *uid, gid_t *gid) {
+	if ( pid == 1 ) {
```
mem, line 67: This is quite subtle as a terminating condition. At first I thought it was just a special case but then I realized that things are wired up so that getUnprivIDs returns 1 on errors. So this bit of code is playing two roles: if the code is run by root, it will terminate here because eventually it will walk up to PID 1; if the code is not run by root, but there's an error, it will terminate here because of the error.

Can you please add a comment?

e2e/internal/e2e/init/init_linux.go
```diff
@@ -0,0 +1,145 @@
...
+		return -1;
+	}
+	pid_t ppid = getProcInfo(pid, uid, gid);
+	if ( *uid == 0 ) {
+		return getUnprivIDs(ppid, uid, gid);
+	}
+	return 0;
+}
+
+__attribute__((constructor)) static void init(void) {
```
mem: can you please add a comment as to why this needs to be set up line like this? (I'm talking about the attribute constructor). This construction makes reading the code really hard, because you start at e2e/suite.go, where `func main` is, and you can continue reading without ever noticing that this thing ran first.

I would say _this_ function needs a comment and `func main` also needs a comment refering this.

e2e/internal/e2e/init/init_linux.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	// use a mount namespace in order to bind a temporary filesystem
+	// on top of home directories and not screw them up by accident
+	// during tests execution
+	if ( unshare(CLONE_FS) < 0 ) {
+		fprintf(stderr, "failed to unshare filesystem: %s\n", strerror(errno));
+		exit(1);
+	}
+	if ( unshare(CLONE_NEWNS) < 0 ) {
```
mem: since this requires root (or CAP_SYS_ADMIN, same difference), maybe check for that before trying to do this in order to produce a better error message. The error message below is pretty generic and it will make things hard to debug for other people.

e2e/internal/e2e/init/init_linux.go
```diff
@@ -0,0 +1,145 @@
...
+		fprintf(stderr, "failed to set private mount propagation: %s\n", strerror(errno));
+		exit(1);
+	}
+	// use a PID namespace if possible to avoid to left some processes
+	// once tests are done. Child process won't catch orphaned child processes like
+	// instances, and we can't really catch them correctly to avoid conflicts during
+	// `cmd.Wait()` calls. But this is not a big deal compared to detached processes
+	// that could keep running on host machine after the tests execution.
+	if ( unshare(CLONE_NEWPID) == 0 ) {
+		pid_t forked = fork();
```
mem, line 84: the body of this if is independent of the rest of the code. Please move it to its own function, just for readability's sake.

I would even say everything from the unshare call down should be its own function.

e2e/internal/e2e/init/init_linux.go
```diff
@@ -0,0 +1,145 @@
...
+		if ( forked > 0 ) {
+			// parent process will wait that tests execution finished
+			int status, exit_status = 0;
+			pid_t child;
+
+			child = waitpid(forked, &status, 0);
+			if ( child < 0 ) {
+				fprintf(stderr, "unexpected error while waiting childs: %s\n", strerror(errno));
```
mem:
```suggestion
				fprintf(stderr, "unexpected error while waiting for children: %s\n", strerror(errno));
```

e2e/internal/e2e/init/init_linux.go
```diff
@@ -0,0 +1,145 @@
...
+	if ( getUnprivIDs(getppid(), &uid, &gid) < 0 ) {
+		fprintf(stderr, "failed to retrieve user information\n");
+		exit(1);
+	}
+	if ( uid == 0 || gid == 0 ) {
+		fprintf(stderr, "failed to retrieve user information\n");
+		exit(1);
+	}
+
+	// use a mount namespace in order to bind a temporary filesystem
```
mem: the bit between here and line 97 could be moved to its own function (setup_mount_namespace or some such).

mem (changes requested):

e2e/internal/e2e/docker.go
cclerget: Yeah, I will check how to do that better

e2e/internal/e2e/home_linux.go
cclerget: Can avoid that by simply adding a root check in constructor in `init_linux.go`

e2e/internal/e2e/home_linux.go
cclerget: > correct?
so it's extracting (part of) the current working directory relative to the home directory of the current (?) user and reparenting on top of the temporary home?

Right

> And why is this necessary only if the current working directory is within the current user's home? (why is there a condition on line 83)

As we use go module, the source could be everywhere, if the source base is in `/tmp/singularity` we don't need to bind CWD on top of the user home directory, but if source code is indeed somewhere in user directory we need to bind it on top of the temporary one in order to execute tests correctly.

> Also, what's special about the directory "3" above? Why it is kept out of this whole thing?

It would have been easier to have something like `buildcfg.SOURCEDIR` in order to determine the source code location, but as tests are executed relative to the directory where `_test.go` is located, so in `$SOURCEDIR/e2e`, `filepath.Dir(cwd)` is equivalent to source directory

e2e/internal/e2e/home_linux.go
cclerget: `SetupHomeDirectories` and everything they can/need to contain

e2e/internal/e2e/home_linux.go
cclerget: will do

e2e/internal/e2e/home_linux.go
cclerget: I think that would make the code more complicated, for the unprivileged part there is also privileged operation to execute and that could be confusing

e2e/internal/e2e/init/init_linux.go
cclerget: Will do

e2e/internal/e2e/init/init_linux.go
cclerget: Not required, `uid` and `gid` are initialized to 0 : https://github.com/sylabs/singularity/blob/11503481784c21a6637e59afe239f7494e71d2ee/e2e/internal/e2e/init/init_linux.go#L71

and there is already a check at https://github.com/sylabs/singularity/blob/11503481784c21a6637e59afe239f7494e71d2ee/e2e/internal/e2e/init/init_linux.go#L64

e2e/internal/e2e/init/init_linux.go
cclerget: will do

e2e/internal/e2e/init/init_linux.go
cclerget: will add a generic check at the beginning for root user

e2e/internal/e2e/init/init_linux.go
cclerget: will do

e2e/e2e_test.go
```diff
@@ -7,6 +7,16 @@ package e2e
 
 import (
 	"testing"
+
+	// This import will execute a CGO section with the help of a C
+	// constructor section "init". As we always requires to run e2e
```
gvallee:
```suggestion
	// constructor section "init". As we always require to run e2e
```

e2e/e2e_test.go
```diff
@@ -7,6 +7,16 @@ package e2e
...
+	// tests as root, the C part is responsible of finding original
```
gvallee:
```suggestion
	// tests as root, the C part is responsible of finding the original
```

e2e/e2e_test.go
```diff
@@ -7,6 +7,16 @@ package e2e
...
+	// user which execute tests, it will also create a dedicated pid
```
gvallee:
```suggestion
	// user who executes tests; it will also create a dedicated pid
```

e2e/e2e_test.go
```diff
@@ -7,6 +7,16 @@ package e2e
...
+	// and mount namespace for e2e tests, and will finally restore
+	// identity to the original user but will retain privileges for
+	// Privileged method allowing to execute a function with root
```
gvallee:
```suggestion
	// Privileged methods enabling the execution of a function with root
```

e2e/internal/e2e/docker.go
```diff
@@ -50,6 +52,25 @@ func PrepRegistry(t *testing.T, env TestEnv) {
 			ExpectExit(0),
 		)
 
+		// start script in e2e/testdata/Docker_registry.def will listen
+		// on port 5111 once docker registry is up and initialized, so
+		// we are trying to connect to this port until we got a response,
+		// without any response after 10 seconds we abort tests execution
+		// because start script probably failed
```
gvallee:
```suggestion
		// because the start script probably failed
```

e2e/internal/e2e/init/init_linux.go
```diff
@@ -0,0 +1,168 @@
...
+		return -1;
+	}
+	pid_t ppid = getProcInfo(pid, uid, gid);
+	if ( *uid == 0 ) {
+		return getUnprivIDs(ppid, uid, gid);
+	}
+	return 0;
+}
+
+// create and use a PID namespace if possible to avoid to left some processes
```
gvallee:
```suggestion
// create and use a PID namespace if possible to avoid leaving some processes
```

gvallee (changes requested):
LGTM, I only change a few things in a few comments.

gvallee (approved):
LGTM!

WestleyK (approved):
Looks good to me :+1:

e2e/internal/e2e/home_linux.go
mem: > As we use go module, the source could be everywhere, if the source base is in `/tmp/singularity` we don't need to bind CWD on top of the user home directory, but if source code is indeed somewhere in user directory we need to bind it on top of the temporary one in order to execute tests correctly.

Ah, got it. That makes sense.

> > Also, what's special about the directory "3" above? Why it is kept out of this whole thing?
> 
> It would have been easier to have something like `buildcfg.SOURCEDIR` in order to determine the source code location, but as tests are executed relative to the directory where `_test.go` is located, so in `$SOURCEDIR/e2e`, `filepath.Dir(cwd)` is equivalent to source directory

I think I understand what you are saying.

I would not mind adding SOURCEDIR to buildcfg (I actually would need this to test the plugin stuff; the closest right now is buildcfg.BUILDDIR, which by default contains a directory one deeper than SOURCEDIR, but that can be changed). You could think of it as a debugging tool, it's pretty common to record that kind of information.

If the reason for this logic is getting to SOURCEDIR from e2e, I would prefer to make that change (add SOURCEDIR to buildcfg).

mem (commented):
This looks good to me. If you could add buildcfg.SOURCEDIR, that would be great.

cclerget:
> This looks good to me. If you could add buildcfg.SOURCEDIR, that would be great.

@mem Yep, added

mem (approved):
OK, let's go with this.

PR #3840, opened 2019-06-28
Add mandatory argument to RunSingularity to specify binary to use

mem:
Add mandatory argument to RunSingularity to specify binary to use

RunSingularity is using "singularity" as the command it wants to run,
which means it's picking up singularity from the user's PATH, which
might or might not correspond to the singularity binary that matches the
code to be tested (e.g. mconfig --prefix=/usr/local/test, and then run
run the E2E suite without adding /usr/local/test/bin to the path).

Fix this by adding a cmdPath argument to RunSingularity which specifies
where the singularity binary is located. This plays nice with the
existing code that records the location of this binary in the
E2E_CMD_PATH environment variable. It's the responsability of the test
calling RunSingularity to pass the correct value derived from this
environment variable.

This is replacing the existing "name" argument, which is mandatory, but
it's only used if a subtest run is requested, which is optional. Convert
this into an option "WithSubtest", which takes the name of the subtest as
an argument and is self-documenting.

### DIFF
--- a/e2e/actions/actions.go
+++ b/e2e/actions/actions.go
@@ -61,7 +61,7 @@ func (c *actionTests) actionRun(t *testing.T) {
 	for _, tt := range tests {
 		e2e.RunSingularity(
 			t,
-			tt.name,
+			e2e.AsSubtest(tt.name),
 			e2e.WithCommand("run"),
 			e2e.WithArgs(tt.argv...),
 			e2e.ExpectExit(tt.exit),
@@ -229,7 +229,7 @@ func (c *actionTests) actionExec(t *testing.T) {
 	for _, tt := range tests {
 		e2e.RunSingularity(
 			t,
-			tt.name,
+			e2e.AsSubtest(tt.name),
 			e2e.WithCommand("exec"),
 			e2e.WithDir("/tmp"),
 			e2e.WithArgs(tt.argv...),
@@ -293,7 +293,7 @@ func (c *actionTests) actionShell(t *testing.T) {
 	for _, tt := range tests {
 		e2e.RunSingularity(
 			t,
-			tt.name,
+			e2e.AsSubtest(tt.name),
 			e2e.WithCommand("shell"),
 			e2e.WithArgs(tt.argv...),
 			e2e.ConsoleRun(tt.consoleOps...),
@@ -378,7 +378,7 @@ func (c *actionTests) STDPipe(t *testing.T) {
 	for _, tt := range stdinTests {
 		e2e.RunSingularity(
 			t,
-			tt.name,
+			e2e.AsSubtest(tt.name),
 			e2e.WithCommand(tt.command),
 			e2e.WithArgs(tt.argv...),
 			e2e.WithStdin(&input),
@@ -430,7 +430,7 @@ func (c *actionTests) STDPipe(t *testing.T) {
 	for _, tt := range stdoutTests {
 		e2e.RunSingularity(
 			t,
-			tt.name,
+			e2e.AsSubtest(tt.name),
 			e2e.WithCommand(tt.command),
 			e2e.WithArgs(tt.argv...),
 			e2e.ExpectExit(
@@ -620,7 +620,7 @@ func (c *actionTests) RunFromURI(t *testing.T) {
 	for _, tt := range tests {
 		e2e.RunSingularity(
 			t,
-			tt.name,
+			e2e.AsSubtest(tt.name),
 			e2e.WithCommand(tt.command),
 			e2e.WithArgs(tt.argv...),
 			e2e.ExpectExit(tt.exit),
@@ -754,7 +754,7 @@ func (c *actionTests) PersistentOverlay(t *testing.T) {
 	for _, tt := range tests {
 		e2e.RunSingularity(
 			t,
-			tt.name,
+			e2e.AsSubtest(tt.name),
 			e2e.WithCommand("exec"),
 			e2e.WithArgs(tt.argv...),
 			e2e.WithPrivileges(tt.privileged),
--- a/e2e/actions/actions.go
+++ b/e2e/actions/actions.go
@@ -59,7 +59,7 @@ func (c *actionTests) actionRun(t *testing.T) {
 	}
 
 	for _, tt := range tests {
-		e2e.RunSingularity(
+		c.env.RunSingularity(
 			t,
 			e2e.AsSubtest(tt.name),
 			e2e.WithCommand("run"),
@@ -227,7 +227,7 @@ func (c *actionTests) actionExec(t *testing.T) {
 	}
 
 	for _, tt := range tests {
-		e2e.RunSingularity(
+		c.env.RunSingularity(
 			t,
 			e2e.AsSubtest(tt.name),
 			e2e.WithCommand("exec"),
@@ -291,7 +291,7 @@ func (c *actionTests) actionShell(t *testing.T) {
 	}
 
 	for _, tt := range tests {
-		e2e.RunSingularity(
+		c.env.RunSingularity(
 			t,
 			e2e.AsSubtest(tt.name),
 			e2e.WithCommand("shell"),
@@ -376,7 +376,7 @@ func (c *actionTests) STDPipe(t *testing.T) {
 	var input bytes.Buffer
 
 	for _, tt := range stdinTests {
-		e2e.RunSingularity(
+		c.env.RunSingularity(
 			t,
 			e2e.AsSubtest(tt.name),
 			e2e.WithCommand(tt.command),
@@ -428,7 +428,7 @@ func (c *actionTests) STDPipe(t *testing.T) {
 		},
 	}
 	for _, tt := range stdoutTests {
-		e2e.RunSingularity(
+		c.env.RunSingularity(
 			t,
 			e2e.AsSubtest(tt.name),
 			e2e.WithCommand(tt.command),
@@ -618,7 +618,7 @@ func (c *actionTests) RunFromURI(t *testing.T) {
 	}
 
 	for _, tt := range tests {
-		e2e.RunSingularity(
+		c.env.RunSingularity(
 			t,
 			e2e.AsSubtest(tt.name),
 			e2e.WithCommand(tt.command),
@@ -752,7 +752,7 @@ func (c *actionTests) PersistentOverlay(t *testing.T) {
 	}
 
 	for _, tt := range tests {
-		e2e.RunSingularity(
+		c.env.RunSingularity(
 			t,
 			e2e.AsSubtest(tt.name),
 			e2e.WithCommand("exec"),
--- a/e2e/env/env.go
+++ b/e2e/env/env.go
@@ -90,7 +90,7 @@ func (c *ctx) singularityEnv(t *testing.T) {
 	}
 
 	for _, tt := range tests {
-		e2e.RunSingularity(
+		c.env.RunSingularity(
 			t,
 			e2e.WithPrivileges(false),
 			e2e.WithCommand("exec"),
--- a/e2e/env/env.go
+++ b/e2e/env/env.go
@@ -92,7 +92,6 @@ func (c *ctx) singularityEnv(t *testing.T) {
 	for _, tt := range tests {
 		e2e.RunSingularity(
 			t,
-			tt.name,
 			e2e.WithPrivileges(false),
 			e2e.WithCommand("exec"),
 			e2e.WithEnv(tt.env),
--- a/e2e/imgbuild/imgbuild.go
+++ b/e2e/imgbuild/imgbuild.go
@@ -59,7 +59,6 @@ func (c *imgBuildTests) buildFrom(t *testing.T) {
 
 		e2e.RunSingularity(
 			t,
-			tt.name,
 			e2e.WithPrivileges(true),
 			e2e.WithCommand("build"),
 			e2e.WithArgs(args...),
@@ -136,7 +135,6 @@ func (c *imgBuildTests) nonRootBuild(t *testing.T) {
 
 		e2e.RunSingularity(
 			t,
-			tt.name,
 			e2e.WithPrivileges(false),
 			e2e.WithCommand("build"),
 			e2e.WithArgs(args...),
@@ -179,7 +177,6 @@ func (c *imgBuildTests) buildLocalImage(t *testing.T) {
 
 	e2e.RunSingularity(
 		t,
-		"test-sandbox",
 		e2e.WithPrivileges(true),
 		e2e.WithCommand("build"),
 		e2e.WithArgs("--sandbox", sandboxImage, c.env.ImagePath),
@@ -209,19 +206,17 @@ func (c *imgBuildTests) buildLocalImage(t *testing.T) {
 
 	for i, tt := range tests {
 		imagePath := filepath.Join(tmpdir, fmt.Sprintf("image-%d", i))
-		t.Run(tt.name, e2e.Privileged(func(t *testing.T) {
-			e2e.RunSingularity(
-				t,
-				tt.name,
-				e2e.WithPrivileges(true),
-				e2e.WithCommand("build"),
-				e2e.WithArgs(imagePath, tt.buildSpec),
-				e2e.PostRun(func(t *testing.T) {
-					e2e.ImageVerify(t, c.env.CmdPath, imagePath)
-				}),
-				e2e.ExpectExit(0),
-			)
-		}))
+		e2e.RunSingularity(
+			t,
+			e2e.AsSubtest(tt.name),
+			e2e.WithPrivileges(true),
+			e2e.WithCommand("build"),
+			e2e.WithArgs(imagePath, tt.buildSpec),
+			e2e.PostRun(func(t *testing.T) {
+				e2e.ImageVerify(t, c.env.CmdPath, imagePath)
+			}),
+			e2e.ExpectExit(0),
+		)
 	}
 }
 
@@ -229,7 +224,6 @@ func (c *imgBuildTests) badPath(t *testing.T) {
 	imagePath := path.Join(c.env.TestDir, "container")
 	e2e.RunSingularity(
 		t,
-		"bad path",
 		e2e.WithPrivileges(true),
 		e2e.WithCommand("build"),
 		e2e.WithArgs(imagePath, "/some/dumb/path"),
@@ -426,7 +420,6 @@ func (c *imgBuildTests) buildMultiStageDefinition(t *testing.T) {
 
 		e2e.RunSingularity(
 			t,
-			tt.name,
 			e2e.WithPrivileges(true),
 			e2e.WithCommand("build"),
 			e2e.WithArgs(args...),
@@ -724,7 +717,6 @@ func (c *imgBuildTests) buildDefinition(t *testing.T) {
 
 		e2e.RunSingularity(
 			t,
-			tt.name,
 			e2e.WithPrivileges(true),
 			e2e.WithCommand("build"),
 			e2e.WithArgs(args...),
--- a/e2e/imgbuild/imgbuild.go
+++ b/e2e/imgbuild/imgbuild.go
@@ -57,7 +57,7 @@ func (c *imgBuildTests) buildFrom(t *testing.T) {
 		}
 		args = append(args, imagePath, tt.buildSpec)
 
-		e2e.RunSingularity(
+		c.env.RunSingularity(
 			t,
 			e2e.WithPrivileges(true),
 			e2e.WithCommand("build"),
@@ -133,7 +133,7 @@ func (c *imgBuildTests) nonRootBuild(t *testing.T) {
 		}
 		args = append(args, imagePath, tt.buildSpec)
 
-		e2e.RunSingularity(
+		c.env.RunSingularity(
 			t,
 			e2e.WithPrivileges(false),
 			e2e.WithCommand("build"),
@@ -175,7 +175,7 @@ func (c *imgBuildTests) buildLocalImage(t *testing.T) {
 
 	sandboxImage := path.Join(tmpdir, "test-sandbox")
 
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
 		e2e.WithPrivileges(true),
 		e2e.WithCommand("build"),
@@ -206,7 +206,7 @@ func (c *imgBuildTests) buildLocalImage(t *testing.T) {
 
 	for i, tt := range tests {
 		imagePath := filepath.Join(tmpdir, fmt.Sprintf("image-%d", i))
-		e2e.RunSingularity(
+		c.env.RunSingularity(
 			t,
 			e2e.AsSubtest(tt.name),
 			e2e.WithPrivileges(true),
@@ -222,7 +222,7 @@ func (c *imgBuildTests) buildLocalImage(t *testing.T) {
 
 func (c *imgBuildTests) badPath(t *testing.T) {
 	imagePath := path.Join(c.env.TestDir, "container")
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
 		e2e.WithPrivileges(true),
 		e2e.WithCommand("build"),
@@ -418,7 +418,7 @@ func (c *imgBuildTests) buildMultiStageDefinition(t *testing.T) {
 		}
 		args = append(args, imagePath, defFile)
 
-		e2e.RunSingularity(
+		c.env.RunSingularity(
 			t,
 			e2e.WithPrivileges(true),
 			e2e.WithCommand("build"),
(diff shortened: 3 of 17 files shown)


e2e/internal/e2e/singularitycmd.go
```diff
@@ -109,7 +109,7 @@ type singularityCmd struct {
 	envs       []string
 	dir        string
 	privileged bool
-	subTest    bool
+	noSubTest  bool
```
mem: no, scratch this. The way the `RunSingularity` function works requires a name for the subtest, even if the subtest is optional. I'll change that instead so that it makes sense.

jscook2345: Sounds good. Ping me when you'd like the review.

mem: thanks @jscook2345 	

I had to change more than I had initially expected. There are two commits in this PR now: the first one removes the "name" argument, as mentioned in the cover letter. The second one adds a "cmdPath" argument, which is the path to the singularity binary, as obtained from the E2E_CMD_PATH environment variable.

e2e/actions/actions.go
```diff
@@ -65,7 +65,8 @@ func actionRun(t *testing.T) {
 	for _, tt := range tests {
 		e2e.RunSingularity(
 			t,
-			tt.name,
+			testenv.CmdPath,
+			e2e.WithSubtest(tt.name),
```
cclerget: I like that with name argument, just it would be more expressive with `AsSubTest(name)`

e2e/actions/actions.go
```diff
@@ -65,7 +65,8 @@ func actionRun(t *testing.T) {
 	for _, tt := range tests {
 		e2e.RunSingularity(
 			t,
-			tt.name,
+			testenv.CmdPath,
```
cclerget: We should get rid of `testenv.CmdPath`, because it's a superficial argument and instead we could append `filepath.Join(buildcfg.LIBEXECDIR, "singularity/bin")` to `PATH` in `Run` function: https://github.com/sylabs/singularity/blob/master/e2e/suite.go#L38 and just restore `singularity` value in `RunSingularity`

e2e/suite.go
```diff
@@ -102,8 +102,8 @@ func Run(t *testing.T) {
 	singularitye2e.EnsureImage(t)
 
 	// Start registry for tests
-	singularitye2e.PrepRegistry(t, name)
-	defer singularitye2e.KillRegistry(t)
+	singularitye2e.PrepRegistry(t, cmdPath, name)
```
cclerget: another reason to get rid of `cmdpath`

cclerget (changes requested):

e2e/actions/actions.go
jscook2345: @cclerget 

I think the argument is that we may not run the installed singularity binary each time. I think that makes sense to make as an input to testing.

jscook2345 (approved):
LGTM

e2e/actions/actions.go
cclerget: After slack discussion, this argument is not required and instead we could set `PATH` conditionally to handle case for a different binary locations (pro release tests ...) :

```
path := os.Getenv("PATH")
if path == "" {
    path = "/usr/local/bin:/usr/local/sbin:/usr/sbin:/usr/bin:/sbin:/bin"
}
// to customize path for both singularity and external binaries
// used in e2e tests
e2ePath := os.Getenv("E2E_PREPEND_PATH")
if e2ePath == "" {
    e2ePath = buildcfg.BINDIR
}
os.Setenv("PATH", fmt.Sprintf("%s:%s", e2ePath, path))
```

mem: No, messing with PATH to get the "right" value is the wrong approach here.

Keep in mind that this is a test and the thing being tested is singularity: the path to the singularity binary under test is an input to the test. That binary under test should be capable of finding its components (confguration file, starter binaries, etc) on its own. How the top level test harness obtains the location of this binary should not be a concern of the test.

I suspect the problem you are trying to work around of is the fact that we are passing a *testing.T argument to the test functions, when we probably should pass something like a *e2e.T which embeds (or includes a field that is) a *testing.T. This *e2e.T would contain the information that we want to pass from the top level test harness down to the tests. If we want to do discovery of test functions, it's probably a good idea to define a single type for the functions that implement tests (just like the testing package does).

All that said, what I'm trying to fix in this PR is the incorrect assumption that the singularity binary under test is the first singularity program found in PATH because that's messing with what I actually want to do.

e2e/actions/actions.go
mem: will do

e2e/suite.go
mem: sorry?

e2e/actions/actions.go
cclerget: There is confusion, the snippet above is intended to be in `e2e/suite.go` not in `RunSingularity`, then you could just restore value to `"singularity"` in `RunSingularity` and `LookPath` in `exec.Command` will does its job by finding the right singularity binary based on `PATH` initialization above and would avoid to pass this argument everytime. Now if you really need to pass a custom singularity binary, you could add a function like `WithBinary(path)` to `RunSingularity`

mem: No, I understood the suggestion correctly. I still say it's going the wrong way about it: it is up to the test harness to locate the binary that is going to be tested. singularity, for better or worse, is built in such a way that it will look in specific paths for the binaries it needs. From the point of view of the E2E tests, you must install singularity (as in `make install`) and from there singularity itself will locate its stuff. The E2E harness is using buildcfg to locate the install location of the singularity binary, so it's testing the binary that matches the current build. That does not need to be that way, but it is that way right now. By looking for singularity in the PATH, as the current code does, it is making a strong and wrong assumption that whatever singularity binary is found that way will match the singularity binary from the current build. In general, that's false.

cclerget: My bad, I edited code snippet above to use the right `buildcfg` path value.

I'm just pointing that we can get rid of this useless argument for `RunSingularity` ... I suggested to initialize `PATH` like above because we can put a custom path first in `PATH` where `LookPath` during call to `exec.Command("singularity" ...)` in `RunSingularity` will consider the first singularity binary found as the one to execute for e2e tests, this doesn't differ of what was done previously with `cmdPath := filepath.Join(buildcfg.BINDIR, "singularity")`.

If you are not happy with `PATH`, this is not a problem, just back to the original issue and the thing I pointed out in my first comment is: why passing singularity binary path as an argument of `RunSingularity` ? which always has the same value and which could involve more work in case of a refactoring, when `exec.Command` in `RunSingularity` could use an initialized variable directly like `exec.Command(e2e.SingularityPath, ...)` or `PATH` or whatever ...

And like I said, if a test require to use another singularity binary for some reasons but I can't figure out a use case for that, there is the possibility to use something like `WithBinary(path)` as a `SingularityCmdOp` argument for `RunSingularity`.

GodloveD: If I understand, @cclerget is trying to make the following point:

Adding the `cmdPath` argument to the `RunSingularity()` function violates the principle of DRY (Don't repeat yourself).  There are ~50 instances of the `RunSingularity()` function that need to be altered in this PR after the argument is added.  And in all 50 calls, the variable is going to be the same right?  Seems tedious and error prone.  So why not just add something like this _into the `RunSingularity()` function itself._
```
cmdPath := filepath.Join(buildcfg.BINDIR, "singularity")
```
And then use it there once instead of setting it outside the function and passing it in 50 times?  (Or you could also just make it a constant in the `e2e` package and import it into the `RunSingularity()` function or whatever.  The point remains the same.)

Is that basically what you are trying to say @cclerget or am I totally misunderstanding your point?

cclerget: Yes that's the point @GodloveD

e2e/env/env.go
```diff
@@ -90,9 +90,8 @@ func (c *ctx) singularityEnv(t *testing.T) {
 	}
 
 	for _, tt := range tests {
-		e2e.RunSingularity(
+		c.env.RunSingularity(
 			t,
-			tt.name,
```
cclerget, line 95: missing `e2e.AsSubtest` call

e2e/imgbuild/imgbuild.go
```diff
@@ -57,9 +57,8 @@ func (c *imgBuildTests) buildFrom(t *testing.T) {
 		}
 		args = append(args, imagePath, tt.buildSpec)
 
-		e2e.RunSingularity(
+		c.env.RunSingularity(
 			t,
-			tt.name,
```
cclerget, line 62: missing `e2e.AsSubtest`

e2e/imgbuild/imgbuild.go
```diff
@@ -134,9 +133,8 @@ func (c *imgBuildTests) nonRootBuild(t *testing.T) {
...
```
cclerget, line 139: missing `e2e.AsSubtest`

e2e/imgbuild/imgbuild.go
```diff
@@ -177,9 +175,8 @@ func (c *imgBuildTests) buildLocalImage(t *testing.T) {
 
 	sandboxImage := path.Join(tmpdir, "test-sandbox")
 
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
-		"test-sandbox",
```
cclerget, line 182: missing `e2e.AsSubtest`

e2e/imgbuild/imgbuild.go
```diff
@@ -424,9 +418,8 @@ func (c *imgBuildTests) buildMultiStageDefinition(t *testing.T) {
 		}
 		args = append(args, imagePath, defFile)
 
-		e2e.RunSingularity(
+		c.env.RunSingularity(
 			t,
-			tt.name,
```
cclerget, line 429: missing `e2e.AsSubtest`

e2e/imgbuild/imgbuild.go
```diff
@@ -722,9 +715,8 @@ func (c *imgBuildTests) buildDefinition(t *testing.T) {
...
```
cclerget, line 727: missing `e2e.AsSubtest`

e2e/instance/instance.go
```diff
@@ -34,9 +34,8 @@ func (c *ctx) testBasicEchoServer(t *testing.T) {
 	args := []string{c.env.ImagePath, instanceName, strconv.Itoa(instanceStartPort)}
 
 	// Start the instance.
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
-		instanceName,
```
cclerget, line 39: missing `e2e.AsSubtest`

e2e/instance/instance.go
```diff
@@ -61,9 +60,8 @@ func (c *ctx) testCreateManyInstances(t *testing.T) {
 		port := instanceStartPort + i
 		instanceName := "echo" + strconv.Itoa(i+1)
 
-		e2e.RunSingularity(
+		c.env.RunSingularity(
 			t,
-			instanceName,
```
cclerget, line 66: missing `e2e.AsSubtest`

e2e/instance/instance.go
```diff
@@ -106,9 +104,8 @@ func (c *ctx) testBasicOptions(t *testing.T) {
 	}
 
 	// Start an instance with the temporary directory as the home directory.
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
-		instanceName,
```
cclerget, line 111: missing `e2e.AsSubtest`

e2e/instance/instance.go
```diff
@@ -156,9 +153,8 @@ func (c *ctx) testContain(t *testing.T) {
 	defer os.RemoveAll(dir)
 
 	// Start the instance.
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
-		instanceName,
```
cclerget, line 161: missing `e2e.AsSubtest`

e2e/instance/instance.go
```diff
@@ -212,9 +208,8 @@ func (c *ctx) testInstanceFromURI(t *testing.T) {
 
 	for _, i := range instances {
 		args := []string{i.uri, i.name}
-		e2e.RunSingularity(
+		c.env.RunSingularity(
 			t,
-			i.name,
```
cclerget, line 217: missing `e2e.AsSubtest`

e2e/instance/instance_utils.go
```diff
@@ -31,9 +31,8 @@ type instanceList struct {
 func (c *ctx) listInstance(t *testing.T, listArgs ...string) (stdout string, stderr string, success bool) {
 	var args []string
 
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
-		"List",
```
cclerget, line 36: missing `e2e.AsSubtest`

e2e/instance/instance_utils.go
```diff
@@ -53,9 +52,8 @@ func (c *ctx) stopInstance(t *testing.T, instance string, stopArgs ...string) (s
 		args = append(args, instance)
 	}
 
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
-		"Stop",
```
cclerget, line 58: missing `e2e.AsSubtest`

e2e/instance/instance_utils.go
```diff
@@ -78,9 +76,8 @@ func (c *ctx) execInstance(t *testing.T, instance string, execArgs ...string) (s
 		name = fmt.Sprintf("%s_%s", name, execArgs[0])
 	}
 
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
-		name,
```
cclerget, line 83: missing `e2e.AsSubtest`

e2e/instance/instance_utils.go
```diff
@@ -106,9 +103,8 @@ func (c *ctx) expectedNumberOfInstances(t *testing.T, n int) {
 		nbInstances = len(instances.Instances)
 	}
 
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
-		"GetNumberOfInstances",
```
cclerget, line 111: missing `e2e.AsSubtest`

e2e/internal/e2e/env.go
```diff
@@ -14,3 +16,11 @@ type TestEnv struct {
 	TestRegistry  string
 	KeyringDir    string
 }
+
+// RunSingularity is a convinience wrapper for the standalone
+// RunSingularity function, ensuring that RunSingularity gets called
+// with the correct singularity path as specified by the test
+// environment.
+func (env TestEnv) RunSingularity(t *testing.T, cmdOps ...SingularityCmdOp) {
```
cclerget, line 24: I left that here as a reminder (or it's up to you if you want to tackle it in this one), to avoid confusion between different function prototypes and also for consistency with other helpers in `e2e/internal` (to pass a `TestEnv` argument instead of currently by example for a function pass a `cmdPath` and for another pass `TestEnv`), I would suggest to rename `RunSingularity(t *testing.T, cmdPath string, cmdOps ...SingularityCmdOp)` in `e2e/internal/e2e/singularitycmd.go` to `runSingularity(t *testing.T, cmdPath string, cmdOps ...SingularityCmdOp)` for internal use only by `TestEnv`, so each helpers using `RunSingularity` would take a `TestEnv` argument

e2e/internal/e2e/imageverify.go
```diff
@@ -75,7 +75,8 @@ func ImageVerify(t *testing.T, cmdPath string, imagePath string) {
 	for _, tt := range tests {
 		RunSingularity(
 			t,
-			tt.name,
+			cmdPath,
```
cclerget, line 78: what I was saying about helpers and `cmdPath` or `TestEnv`

e2e/oci/oci.go
```diff
@@ -37,7 +37,7 @@ func checkOciState(t *testing.T, containerID string, state string) {
 
 	e2e.RunSingularity(
 		t,
-		"State",
```
cclerget, line 40: missing `AsSubtest`

e2e/oci/oci.go
```diff
@@ -52,9 +52,8 @@ func genericOciMount(t *testing.T, c *ctx) (string, func()) {
 	}
 	ociConfig := filepath.Join(bundleDir, "config.json")
 
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
-		"OciMount",
```
cclerget, line 57: missing `e2e.AsSubtest`

e2e/oci/oci.go
```diff
@@ -81,9 +80,8 @@ func genericOciMount(t *testing.T, c *ctx) (string, func()) {
 	)
 
 	cleanup := func() {
-		e2e.RunSingularity(
+		c.env.RunSingularity(
 			t,
-			"OciUmount",
```
cclerget, line 86: missing `e2e.AsSubtest`

e2e/oci/oci.go
```diff
@@ -140,9 +138,8 @@ func (c *ctx) testOciAttach(t *testing.T) {
 	// umount bundle
 	defer umountFn()
 
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
-		"OciAttachCreate",
```
cclerget, line 145: missing `e2e.AsSubtest`

e2e/oci/oci.go
```diff
@@ -153,29 +150,27 @@ func (c *ctx) testOciAttach(t *testing.T) {
...
+				checkOciState(t, c.env.CmdPath, containerID, ociruntime.Created)
 			}
 		}),
 		e2e.ExpectExit(0),
 	)
 
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
-		"OciAttachStart",
```
cclerget, line 164: missing `e2e.AsSubtest`

e2e/oci/oci.go
```diff
@@ -153,29 +150,27 @@ func (c *ctx) testOciAttach(t *testing.T) {
...
+				checkOciState(t, c.env.CmdPath, containerID, ociruntime.Running)
 			}
 		}),
 		e2e.ExpectExit(0),
 	)
 
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
-		"OciAttachInteractive",
```
cclerget, line 178: missing `e2e.AsSubtest`

e2e/oci/oci.go
```diff
@@ -186,15 +181,14 @@ func (c *ctx) testOciAttach(t *testing.T) {
...
+				checkOciState(t, c.env.CmdPath, containerID, ociruntime.Stopped)
 			}
 		}),
 		e2e.ExpectExit(0),
 	)
 
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
-		"OciAttachDelete",
```
cclerget, line 197: missing `e2e.AsSubtest`

e2e/oci/oci.go
```diff
@@ -211,9 +205,8 @@ func (c *ctx) testOciBasic(t *testing.T) {
 	// umount bundle
 	defer umountFn()
 
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
-		"OciBasicCreate",
```
cclerget, line 216: missing `e2e.AsSubtest`

e2e/oci/oci.go
```diff
@@ -224,29 +217,27 @@ func (c *ctx) testOciBasic(t *testing.T) {
...
+				checkOciState(t, c.env.CmdPath, containerID, ociruntime.Created)
 			}
 		}),
 		e2e.ExpectExit(0),
 	)
 
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
-		"OciBasicStart",
```
cclerget, line 235: missing `e2e.AsSubtest`

e2e/oci/oci.go
```diff
@@ -224,29 +217,27 @@ func (c *ctx) testOciBasic(t *testing.T) {
...
+				checkOciState(t, c.env.CmdPath, containerID, ociruntime.Running)
 			}
 		}),
 		e2e.ExpectExit(0),
 	)
 
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
-		"OciPause",
```
cclerget, line 249: missing `e2e.AsSubtest`

e2e/oci/oci.go
```diff
@@ -256,15 +247,14 @@ func (c *ctx) testOciBasic(t *testing.T) {
...
+				checkOciState(t, c.env.CmdPath, containerID, ociruntime.Paused)
 			}
 		}),
 		e2e.ExpectExit(0),
 	)
 
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
-		"OciResume",
```
cclerget, line 267: missing `e2e.AsSubtest`

e2e/oci/oci.go
```diff
@@ -274,72 +264,65 @@ func (c *ctx) testOciBasic(t *testing.T) {
...
+				checkOciState(t, c.env.CmdPath, containerID, ociruntime.Running)
 			}
 		}),
 		e2e.ExpectExit(0),
 	)
 
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
-		"OciBasicAlreadyStarted",
```
cclerget, line 285: missing `e2e.AsSubtest`

e2e/oci/oci.go
```diff
@@ -274,72 +264,65 @@ func (c *ctx) testOciBasic(t *testing.T) {
...
 		e2e.WithPrivileges(true),
 		e2e.WithCommand("oci start"),
 		e2e.WithArgs(containerID),
 		e2e.ExpectExit(255),
 	)
 
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
-		"OciBasicExecId",
```
cclerget, line 294: missing `e2e.AsSubtest`

e2e/oci/oci.go
```diff
@@ -274,72 +264,65 @@ func (c *ctx) testOciBasic(t *testing.T) {
...
 		e2e.WithPrivileges(true),
 		e2e.WithCommand("oci exec"),
 		e2e.WithArgs(containerID, "id"),
 		e2e.ExpectExit(0),
 	)
 
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
-		"OciBasicKill",
```
cclerget, line 303: missing `e2e.AsSubtest`

e2e/oci/oci.go
```diff
@@ -274,72 +264,65 @@ func (c *ctx) testOciBasic(t *testing.T) {
...
+				checkOciState(t, c.env.CmdPath, containerID, ociruntime.Stopped)
 			}
 		}),
 		e2e.ExpectExit(0),
 	)
 
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
-		"OciBasicDelete",
```
cclerget, line 317: missing `e2e.AsSubtest`

e2e/oci/oci.go
```diff
@@ -274,72 +264,65 @@ func (c *ctx) testOciBasic(t *testing.T) {
...
 		e2e.WithPrivileges(true),
 		e2e.WithCommand("oci delete"),
 		e2e.WithArgs(containerID),
 		e2e.ExpectExit(0),
 	)
 
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
-		"OciBasicStateAfterDelete",
```
cclerget, line 326: missing `e2e.AsSubtest`

e2e/oci/oci.go
```diff
@@ -274,72 +264,65 @@ func (c *ctx) testOciBasic(t *testing.T) {
...
 		e2e.WithPrivileges(true),
 		e2e.WithCommand("oci state"),
 		e2e.WithArgs(containerID),
 		e2e.ExpectExit(255),
 	)
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
-		"OciBasicKillAfterDelete",
```
cclerget, line 334: missing `e2e.AsSubtest`

e2e/oci/oci.go
```diff
@@ -274,72 +264,65 @@ func (c *ctx) testOciBasic(t *testing.T) {
...
 		e2e.WithPrivileges(true),
 		e2e.WithCommand("oci kill"),
 		e2e.WithArgs(containerID),
 		e2e.ExpectExit(255),
 	)
-	e2e.RunSingularity(
+	c.env.RunSingularity(
 		t,
-		"OciBasicStartAfterDelete",
```
cclerget, line 342: missing `e2e.AsSubtest`

e2e/version/version.go
```diff
@@ -38,9 +38,8 @@ func (c *ctx) testSemanticVersion(t *testing.T) {
 			}
 		}
 
-		e2e.RunSingularity(
+		c.env.RunSingularity(
 			t,
-			tt.name,
```
cclerget, line 43: missing `e2e.AsSubtest`

e2e/version/version.go
```diff
@@ -80,9 +79,8 @@ func (c *ctx) testEqualVersion(t *testing.T) {
...
```
cclerget, line 85: missing `e2e.AsSubtest`

cclerget (dismissed):

PR #3849, opened 2019-07-02
Return ErrArchMismatch if SIF arch is incompatible with host's

sashayakovtseva:
Also add periods in the end of comments. 
Simplify AuthorizedGroup and AuthorizedOwner funcs.

### DIFF
--- a/pkg/image/image.go
+++ b/pkg/image/image.go
@@ -44,9 +44,16 @@ var registeredFormats = []struct {
 	{"ext3", &ext3Format{}},
 }
 
+// ErrArchMismatch denotes that the image was built for use on a
+// different platform and the host architecture is incompatible.
+var ErrArchMismatch = fmt.Errorf("image arch doesn't match host's")
+
 // format describes the interface that an image format type must implement.
 type format interface {
 	openMode(bool) int
+	// initializer tries to init Image based on the format. If passed
+	// file is not of the format, an error is returned. When format is correct,
+	// but image was built for use on a different platform ErrArchMismatch is returned.
 	initializer(*Image, os.FileInfo) error
 }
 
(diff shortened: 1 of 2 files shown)


pkg/image/image.go
```diff
@@ -192,14 +197,18 @@ func Init(path string, writable bool) (*Image, error) {
...
 
 		err = rf.format.initializer(img, fileinfo)
+		if err == ErrArchMismatch {
+			_ = img.File.Close()
+			return nil, ErrArchMismatch
+		}
 		if err != nil {
-			sylog.Debugf("%s format initializer returns: %s", rf.name, err)
-			img.File.Close()
+			sylog.Errorf("%s format initializer returned: %s", rf.name, err)
```
cclerget: It will display annoying false error message for each executed initializer until it found the correct format, should be `Debugf` or `Verbosef`

pkg/image/image.go
```diff
@@ -44,9 +44,16 @@ var registeredFormats = []struct {
 	{"ext3", &ext3Format{}},
 }
 
+// ErrArchMismatch denotes that the image was built for use on a
+// different platform and the host architecture is incompatible.
+var ErrArchMismatch = fmt.Errorf("image arch doesn't match host's")
```
cclerget: Ok it eases detection but we loose information. The previous error message displayed current arch and detected arch in image, we should keep this information and display it as error for users. Why not use the suggestion made in slack channel ?

cclerget (changes requested):

pkg/image/sif.go
```diff
@@ -41,15 +40,21 @@ func (f *sifFormat) initializer(img *Image, fileinfo os.FileInfo) error {
...
+	// 3. if definition file contains a 'executable' section, the architecture check should
+	// occur (or delegate to runtime which would fail during execution).
+	// The current code will be called by the starter which will cover most of the
+	// workflow described above. However, SIF is currently build upon the assumption
+	// that the architecture is assigned based on the architecture defined by a Go
+	// runtime, which is not 100% compliant with the intended workflow.
 	sifArch := string(fimg.Header.Arch[:sif.HdrArchLen-1])
 	if sifArch != sif.HdrArchUnknown && sifArch != sif.GetSIFArch(runtime.GOARCH) {
-		sylog.Fatalf("the image's architecture (%s) is incompatible with the host (%s)", sif.GetGoArch(sifArch), runtime.GOARCH)
+		return ErrArchMismatch
```
pisarukv: what about to use errors.Wrapf(ErrArchMismatch, "image arch: %s host arch: %s", sif.GetGoArch(sifArch), runtime.GOARCH)

cclerget (approved):
LGTM

PR #3853, opened 2019-07-03
Pass test environment information as function arguments

mem:
Originally the intention was to run each test as a separate program
(there's some important amount of rewriting required to do that).
Because of this some test parameters are being passed in the
environment, in a way that is not immediately obvious. Since that
pattern is causing some confusion and people are outright ignoring it in
some cases, remove it and replace it by adding a test context that
contains all the relevant test information (path to singularity, test
directory, test image, etc).

In order to keep the diff to a minimum, this change has been made almost
mechanically, so there's room for improvement in the future.


e2e/pull/pull.go
```diff
@@ -268,26 +263,69 @@ func getImageNameFromURI(imgURI string) string {
 	return uri.GetName(imgURI)
 }
 
-func testPullCmd(t *testing.T) {
+func (c *ctx) setup(t *testing.T) {
```
mem, line 266: this one is the only change that is bigger than it needs to be. There was a bunch of code in the set up portion of the test, so I gave it a name. The code is unchanged.

e2e/internal/e2e/image.go
```diff
@@ -10,17 +10,10 @@ import (
...
-	ImagePath string `split_words:"true"` // base image for tests
-}{}
-
 // EnsureImage checks if e2e test image is already built or built
 // it otherwise.
-func EnsureImage(t *testing.T) {
-	LoadEnv(t, &testenv)
-
-	switch _, err := os.Stat(testenv.ImagePath); {
+func EnsureImage(t *testing.T, env TestEnv) {
```
jscook2345, line 15: instead of `env` should this be `env.ImagePath`?

mem: I guess you are right, but...

This function needs three things:
1. The path to the test image
2. The path to singularity in case the image needs to be built
3. The path to the directory with test data

That's more than half of the information contained in `TestEnv`. In fact I wonder if it should care about `RunDisabled`.

The first one is obviously being used. The second one _should_ be used, but because of the way `RunSingularity` works, it's not obvious that this function depends on that. The last one is tricky, because there's an implicit requirement that the current working directory is one above the directory with the test data.

I can change it to take an image path if you think that's better, but I'm pretty sure it's going to change back to this, or something like this. I thought about making this function a method of TestEnv, but did not implement that.

jscook2345 (approved):

PR #3857, opened 2019-07-03
Error if no keys are in keyring when trying to export them

WestleyK:

This PR fixes the issue when you try to export a key, and there are no keys in your keyring. (before it would panic)

- Fixes #

<br>


cclerget (approved):
LGTM

pkg/sypgp/sypgp.go
```diff
@@ -564,6 +564,9 @@ func EncryptKey(k *openpgp.Entity, pass string) error {
 
 // SelectPubKey prints a public key list to user and returns the choice
 func SelectPubKey(el openpgp.EntityList) (*openpgp.Entity, error) {
+	if el == nil {
+		return nil, fmt.Errorf("no key in public keyring")
```
sashayakovtseva: maybe add a separate error for that cases? e.g `ErrEmptyKeyList`
also you can check if a list is empty with `len(el)` to eliminate case where slice is not nil but still empty.

pkg/sypgp/sypgp.go
```diff
@@ -564,6 +564,9 @@ func EncryptKey(k *openpgp.Entity, pass string) error {
...
+	}
 	PrintPubKeyring()
```
sashayakovtseva: Just looked into this func, it reads file to fetch `openpgp.EntityList` to print, but we already have one in `SelectXXXKey` arguments... I think that should be refactored. Should we create a separate issue for that, what do you think?

pkg/sypgp/sypgp.go
WestleyK: Good idea, I will need to use two variables, `errPublicKeyringEmpty`, and `errPrivateKeyringEmpty` for the public and private keyring.

sashayakovtseva: I think one exported error would be enough, the context will be clear depending on function being called

pkg/sypgp/sypgp.go
WestleyK: I know @mem and @gvallee was doing some refactoring on this package, I want to change as less as possible. And I would prefer to change that in a separate PR. (Small PRs are more likely to get merged quicker :smile:)

pkg/sypgp/sypgp.go
```diff
@@ -564,8 +568,8 @@ func EncryptKey(k *openpgp.Entity, pass string) error {
 
 // SelectPubKey prints a public key list to user and returns the choice
 func SelectPubKey(el openpgp.EntityList) (*openpgp.Entity, error) {
-	if el == nil {
-		return nil, fmt.Errorf("no key in public keyring")
+	if len(el) < 1 {
```
sashayakovtseva: Comparing with 0 is more natural way to check for emptiness

pkg/sypgp/sypgp.go
```diff
@@ -587,9 +594,12 @@ func SelectPubKey(el openpgp.EntityList) (*openpgp.Entity, error) {
 
 // SelectPrivKey prints a secret key list to user and returns the choice
 func SelectPrivKey(el openpgp.EntityList) (*openpgp.Entity, error) {
+	if len(el) <= 0 {
```
sashayakovtseva: Sorry to bother, but length cannot be less than 0 😃

So the common way to check slices regardless of nil is `if len(slice) == 0`. Just to look Go native. Otherwise lgtm!

WestleyK: Okay, I will fix in a bit :+1:

But, is it possible for it to return `-1`? as a error maybe? I guess Go does not do that...

sashayakovtseva: no, len never returns negative, it is in go spec

pkg/sypgp/sypgp.go
```diff
@@ -45,6 +45,10 @@ var errTooManyRetries = errors.New("too many retries while getting a passphrase"
 var errNotEncrypted = errors.New("key is not encrypted")
 var errInvalidChoice = errors.New("invalid choice")
 
+// ErrEmptyKeyring is the error when the public, or private keyring
+// empty.
+var ErrEmptyKeyring = errors.New("no key in keyring")
```
mem:
```suggestion
var ErrEmptyKeyring = errors.New("no keys in keyring")
```

or:

```suggestion
var ErrEmptyKeyring = errors.New("keyring is empty")
```

mem (approved):

WestleyK:
Closing, https://github.com/sylabs/singularity/pull/3889 is merged.

PR #3858, opened 2019-07-03
Always check the return of the Scan() function

### LINKED ISSUE #3714: Always check to return of `Scan()`
## Version of Singularity:

Master.

### Expected behavior

@mem pointed out that the `Scan` method returns a boolean, which should be inspected before trying to obtain the text:

```
Scan advances the Scanner to the next token, which will then be available
through the Bytes or Text method. It returns false when the scan stops,
either by reaching the end of the input or an error. After Scan returns
false, the Err method will return any error that occurred during scanning,
except that if it was io.EOF, Err will return nil. Scan panics if the split
function returns too many empty tokens without advancing the input. This is
a common error mode for scanners.
```

Our code base includes a few calls where we do not check the boolean returned by `Scan()`

### Actual behavior

Write here.

Write here.

gvallee:

Always check the return of the Scan() function.

- Fixes #3714 


pkg/util/fs/proc/proc.go
```diff
@@ -144,7 +144,10 @@ func ReadIDMap(path string) (uint32, uint32, error) {
 	defer r.Close()
```
WestleyK, line 145: This files copyright year should be `2018-2019`

WestleyK (commented):
Looks almost good to me; can you change the copyright year to `2018-2019` in this file. Otherwise looks good, thanks :+1:

WestleyK (approved):
Looks good to me! :+1:

pkg/util/fs/proc/proc.go
```diff
@@ -144,7 +144,10 @@ func ReadIDMap(path string) (uint32, uint32, error) {
...
 
 	scanner := bufio.NewScanner(r)
-	scanner.Scan()
+	tok := scanner.Scan()
+	if !tok {
```
sashayakovtseva: Can be simplified a bit with `if !scanner.Scan()`, since `tok` is used for this single check only

pkg/util/fs/proc/proc.go
```diff
@@ -144,7 +144,10 @@ func ReadIDMap(path string) (uint32, uint32, error) {
...
+		return 0, 0, scanner.Err()
```
sashayakovtseva: Btw error can be nil here if the file was empty. I am not sure how that will affect further logic since it may become `0`  for uid, so better ask @cclerget.

cclerget:
```suggestion
		return 0, 0, fmt.Errorf("scanner error: %s", scanner.Err())
```

cclerget (changes requested):

pkg/util/fs/proc/proc.go
sashayakovtseva: hm, that might be confusing, `scanner error: nil`
but assuming that files are never empty...

sashayakovtseva (approved):

cclerget (approved):

WestleyK:
Whats the status on this? Looks like it can be merged now.

PR #3861, opened 2019-07-04
Add workaround for false-negative in ShellExit test

mem:
Add "cd /" to work around issue where a long working directory name
causes the test to fail because the "Singularity" that we are looking
for is chopped from the front.

### DIFF
--- a/e2e/actions/actions.go
+++ b/e2e/actions/actions.go
@@ -255,6 +255,7 @@ func (c *actionTests) actionShell(t *testing.T) {
 				// to fail because the "Singularity" that
 				// we are looking for is chopped from the
 				// front.
+				// TODO(mem): This test was added back in 491a71716013654acb2276e4b37c2e015d2dfe09
 				e2e.ConsoleSendLine("cd /"),
 				e2e.ConsoleExpect("Singularity"),
 				e2e.ConsoleSendLine("exit"),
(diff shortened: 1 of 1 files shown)


e2e/actions/actions.go
```diff
@@ -250,6 +250,12 @@ func (c *actionTests) actionShell(t *testing.T) {
 			name: "ShellExit",
 			argv: []string{c.env.ImagePath},
 			consoleOps: []e2e.SingularityConsoleOp{
+				// "cd /" to work around issue where a long
+				// working directory name causes the test
+				// to fail because the "Singularity" that
+				// we are looking for is chopped from the
+				// front.
```
ArangoGutierrez, line 257: Could you ad as TODO the commit hash that introduced the error, just to track it, and be able to circle back in the future and fix it.

ArangoGutierrez (changes requested):

e2e/actions/actions.go
mem, line 257:
```suggestion
				// front.
				// TODO(mem): This test was added back in 491a71716013654acb2276e4b37c2e015d2dfe09
```

ArangoGutierrez (approved):

ArangoGutierrez (approved):

jscook2345 (approved):

PR #3865, opened 2019-07-05
Disable tests or test cases that access singularity hub

mem:
While singularity hub is down, disable the tests that try to access it.
Every disabled test or test case has been marked with a comment
indicating the reason for disabling it is the fact that singularity hub
is down. This is done in order to avoid confusion in case the code is
reworked before the tests can be reenabled again.


ArangoGutierrez:
#3866

ArangoGutierrez (approved):
#3866

PR #3868, opened 2019-07-05
Add some "Require" test helpers

cclerget:

Add a `internal/pkg/test/tool/require` package with :

- `UserNamespace` to check if user namespaces are supported
- `Filesytem` to check if a filesystem is supported
- `CgroupMount` to check if cgroup mount is present
- `Command` to check if command is in `PATH`

### DIFF
--- a/internal/pkg/test/tool/require/require.go
+++ b/internal/pkg/test/tool/require/require.go
@@ -1,0 +1,83 @@
+// Copyright (c) 2019, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package require
+
+import (
+	"os/exec"
+	"sync"
+	"syscall"
+	"testing"
+
+	"github.com/sylabs/singularity/pkg/util/fs/proc"
+)
+
+var hasUserNamespace bool
+var hasUserNamespaceOnce sync.Once
+
+// UserNamespace checks that the current test could use
+// user namespace, if user namespaces are not enabled or
+// supported, the current test is skipped with a message.
+func UserNamespace(t *testing.T) {
+	// not performance critical, just save extra execution
+	// to get the same result
+	hasUserNamespaceOnce.Do(func() {
+		// user namespace is a bit special, as there is no simple
+		// way to detect if it's supported or enabled via a call
+		// on /proc/self/ns/user, the easiest and reliable way seems
+		// to directly execute a command by requesting user namespace
+		cmd := exec.Command("/bin/true")
+		cmd.SysProcAttr = &syscall.SysProcAttr{
+			Cloneflags: syscall.CLONE_NEWUSER,
+		}
+		// no error means user namespaces are enabled
+		hasUserNamespace = cmd.Run() == nil
+	})
+	if !hasUserNamespace {
+		t.Skipf("user namespaces seems not enabled or supported")
+	}
+}
+
+// Filesystem checks that the current test could use the
+// corresponding filesystem, if the filesystem is not
+// listed in /proc/filesystems, the current test is skipped
+// with a message.
+func Filesystem(t *testing.T, fs string) {
+	has, err := proc.HasFilesystem(fs)
+	if err != nil {
+		t.Fatalf("error while checking filesystem presence: %s", err)
+	}
+	if !has {
+		t.Skipf("%s filesystem seems not supported", fs)
+	}
+}
+
+// CgroupMount checks that /sys/fs/cgroup mount is present,
+// if no cgroup mount is found, the current test is skipped
+// with a message.
+func CgroupMount(t *testing.T) {
+	// check first that cgroup is enabled
+	Filesystem(t, "cgroup")
+
+	mounts, err := proc.ParseMountInfo("/proc/self/mountinfo")
+	if err != nil {
+		t.Fatalf("could not obtain mount information: %s", err)
+	}
+	// checks if /sys/fs/cgroup is mounted
+	_, hasCgroupMount := mounts["/sys/fs/cgroup"]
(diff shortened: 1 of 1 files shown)


internal/pkg/test/tool/require.go
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) 2019, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package tool
```
mem: Sorry to make the same comment as the other code review...

Can you please add documentation for this package stating what it provides (https://rakyll.org/style-packages/ has some good advise on how to document packages, near the bottom of the article; https://github.com/golang/go/wiki/CodeReviewComments#package-comments has more of that, too).

What I would like to avoid is a split where some things are added to "test" (one directory above this) and others are added to "tool" (this).

mem (changes requested):

internal/pkg/test/tool/require.go
cclerget: I will add a `doc.go` in other PR

jscook2345 (approved):
I'm really happy to see these type of functions showing up, thanks so much!

internal/pkg/test/tool/require.go
```diff
@@ -0,0 +1,87 @@
...
+	if !hasUserNamespace {
+		t.Skipf("user namespaces seems not enabled or supported")
+	}
+}
+
+// RequireFilesystem checks that the current test could
+// use the corresponding filesystem, if the filesystem is
+// not listed in /proc/filesystems, the current test is
+// skipped with a message.
+func RequireFilesystem(t *testing.T, fs string) {
```
mem: this is fine, but I'm not sure I understand the need for it. I can see that you are using this below to check for cgroup, but I suspect this is going beyond that.

Also, from this PR alone, this is a helper for the next function. I would not export it. I would instead make wrappers like the below one. But then again, I don't know the actual use case for this...

internal/pkg/test/tool/require.go
```diff
@@ -0,0 +1,87 @@
...
+	has, err := proc.HasFilesystem(fs)
+	if err != nil {
+		t.Logf("error while checking filesystem presence: %s", err)
```
mem: I would guess that if this fails, the test cannot continue, as it seems the only way that `proc.HasFilesystem` can fail is if it cannot open /proc/filesystems... which means there's no proc?

internal/pkg/test/tool/require.go
```diff
@@ -0,0 +1,87 @@
...
+// RequireCgroupMount checks that the current test could
+// manipulate cgroup, if there is no cgroup mount found,
+// the current test is skipped with a message.
+func RequireCgroupMount(t *testing.T) {
+	// check first that cgroup is enabled
+	RequireFilesystem(t, "cgroup")
+
+	mounts, err := proc.ParseMountInfo("/proc/self/mountinfo")
+	if err != nil {
+		t.Logf("could not use cgroup mount: %s", err)
```
mem: "cannot obtain mount info" perhaps? the error should not have anything to do with cgroups directly.

Also, if this fails, shouldn't this be fatal?

internal/pkg/test/tool/require.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	// checks if /sys/fs/cgroup is mounted
+	_, hasCgroupMount := mounts["/sys/fs/cgroup"]
+	if !hasCgroupMount {
+		t.Skipf("no /sys/fs/cgroup mount found")
+	}
+}
+
+// RequireCommand checks if the provided command is found
+// in one the path defined in the PATH environment variable,
```
mem:
```suggestion
// in one the directories defined in the PATH environment variable.
```

internal/pkg/test/tool/require.go
```diff
@@ -0,0 +1,87 @@
...
+// if not found the current test is skipped with a message.
```
mem:
```suggestion
// If not found the current test is skipped with a message.
```

internal/pkg/test/tool/require.go
```diff
@@ -0,0 +1,87 @@
...
+	has, err := proc.HasFilesystem(fs)
+	if err != nil {
+		t.Logf("error while checking filesystem presence: %s", err)
+	}
+	if !has {
+		t.Skipf("%s filesystem seems not supported", fs)
+	}
+}
+
+// RequireCgroupMount checks that the current test could
```
mem: I want to say that this is not what this function does, but I'll leave it up to you.

mem (changes requested):

internal/pkg/test/tool/require.go
mem: > I will add a `doc.go` in other PR

There's no need for an additional PR, please add the documentation here as well.

mem: As for the name of the package...

Maybe naming the package `require` helps readability, as that would lead to code that reads like `require.CgroupMount` for example, and it establishes that the contents of the package are assert-type functions.

internal/pkg/test/tool/require.go
cclerget: useful for e2e tests using overlay: `RequireFilesytem(t, "overlay")`

internal/pkg/test/tool/require.go
cclerget: just defensive

internal/pkg/test/tool/require.go
cclerget: will fix that

internal/pkg/test/tool/require.go
cclerget: after comma ?

mem (approved):

PR #3872, opened 2019-07-05
Move call to e2e.EnsureImage to the tests that need it

mem:
EnsureImage creates the test image if it's missing. This is done to
avoid creating the image multiple times for each test.

The call to EnsureImage MUST be in the tests that require said image in
order to avoid paying the cost of creating the image when it is not
required: many tests do not require this image and can operate without
it, and while developing a test or working on individual failure cases
it's useful to be able to run a single test repeatedly. If said test
requires the image, there's no way around it, but if the test does not
require the test image it does not make sense to build it over and over
again.

### DIFF
--- a/e2e/suite.go
+++ b/e2e/suite.go
@@ -114,7 +114,7 @@ func Run(t *testing.T) {
 	testenv.ImagePath = imagePath
 	defer os.Remove(imagePath)
 
-	// XXX(mem): Please DO NOT add a call to e2e.EnsureImage here.
+	// WARNING(Sylabs-team): Please DO NOT add a call to e2e.EnsureImage here.
 	// If you need the test image, add the call at the top of your
 	// own test.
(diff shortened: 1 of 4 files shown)


e2e/suite.go
```diff
@@ -114,8 +114,9 @@ func Run(t *testing.T) {
 	testenv.ImagePath = imagePath
 	defer os.Remove(imagePath)
 
-	// build test image
-	singularitye2e.EnsureImage(t, testenv)
+	// XXX(mem): Please DO NOT add a call to e2e.EnsureImage here.
```
ArangoGutierrez:
```suggestion
	// WARNING(Sylabs-team): Please DO NOT add a call to e2e.EnsureImage here.
```

ArangoGutierrez (approved):

PR #3875, opened 2019-07-05
Changed SINGULARITY_NAME => SINGULARITY_PULL_NAME

WestleyK:

This variable name was interfering will 'SINGULARITY_NAME' when shell-ing into a container. This is a problem if you are running singularity in a singularity container, because every time you pull a container, it will be the name of the container you are shelling into, (unexpected behavior)

Besides, `SINGULARITY_NAME` is kinda undescriptive for the pull name.

<br>

### DIFF
--- a/cmd/internal/cli/pull.go
+++ b/cmd/internal/cli/pull.go
@@ -79,7 +79,7 @@ var pullNameFlag = cmdline.Flag{
 	Name:         "name",
 	Hidden:       true,
 	Usage:        "specify a custom image name",
-	EnvKeys:      []string{"NAME"},
+	EnvKeys:      []string{"PULL_NAME"},
 }
 
 // --dir


cmd/internal/cli/pull.go
```diff
@@ -79,7 +79,7 @@ var pullNameFlag = cmdline.Flag{
 	Name:         "name",
 	Hidden:       true,
 	Usage:        "specify a custom image name",
-	EnvKeys:      []string{"NAME"},
+	EnvKeys:      []string{"PULL_NAME"},
```
WestleyK, line 82: @cclerget, maybe `SINGULARITY_NAME` was intentional? Was it supported to interfere with the other `SINGULARITY_NAME`?

ikaneshiro (approved):
LGTM

cmd/internal/cli/pull.go
ikaneshiro: These variables should not collide

WestleyK: :+1:

PR #3879, opened 2019-07-05
Remove call to e2e.PrepRegistry from toplevel test

mem:
Just like with e2e.EnsureImage, the call to e2e.PrepRegistry should not
be in the toplevel test. Instead each test that needs to access the
registry MUST call this function to make sure the registry is ready for
use.


jscook2345:
Testing in progress

jscook2345 (approved):
Approved and tested locally. +1

PR #3883, opened 2019-07-07
Replace offline pritunl Arch Linux mirror

louwers:

I replaced the pritunl mirror (`http://mirror.pritunl.com/archlinux/2018/08/11/$repo/os/$arch`) that is not working with an official mirror. Just like the replaced mirror, this mirror is a snapshot of the Arch Linux package archive at a specific date. Seehttps://wiki.archlinux.org/index.php/Arch_Linux_Archive. Right now, the commands don't work as stated because the provided server address returns 403. 

Since this is such a minor change I did not add myself to the contributors file.

### DIFF
--- a/examples/arch/README.md
+++ b/examples/arch/README.md
@@ -11,6 +11,7 @@ docker run -it  --privileged pritunl/archlinux bash
 ```
 
 ```bash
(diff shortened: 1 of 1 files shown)


louwers:
Even with the official mirror, troubles are ahead down the line. I'll see what I can do upstream.

PR #3885, opened 2019-07-07
Minor improvments to error handling

WestleyK:

This PR improves some error handling.

- Fixes #

<br>


pkg/util/crypt/crypt_dev.go
```diff
@@ -222,13 +210,14 @@ func copyDeviceContents(source, dest string, size int64) error {
...
 		numWritten, err := syscall.Write(destFd, buffer)
 		if err != nil {
-			return fmt.Errorf("Unable to write to destination %s", dest)
+			return fmt.Errorf("unable to write to destination: %s: %s", dest, err)
 		}
 
+		// TODO: is this really necessary???
 		if numRead != numWritten {
 			sylog.Debugf("numRead != numWritten")
 		}
```
WestleyK: @schebro, is this necessary? Should it be removed?

schebro: I added it so that we see when there is some discrepancy b/w read and written data. Not strictly necessary. But, I prefer to leave it in for debugging purposes.

schebro (approved):

pkg/util/crypt/crypt_dev.go
```diff
@@ -71,29 +70,22 @@ func (crypt *Device) CloseCryptDevice(path string) error {
 // Keys should be non-interactive, preferably in a keyfile that can
 // be passed to cryptsetup utility
 func (crypt *Device) ReadKeyFromStdin(confirm bool) (string, error) {
-
-	fmt.Print("Enter the Key: ")
-	password, err := terminal.ReadPassword(int(syscall.Stdin))
+	pass1, err := sypgp.AskQuestionNoEcho("Enter a passphrase: ")
 	if err != nil {
-		sylog.Fatalf("Error parsing the key: %s", err)
+		return "", fmt.Errorf("unable getting input: %s", err)
```
jchavezlavalle: Avoid using general words like "input" on error messages .. maybe something better would be: "unable to get passphrase"

pkg/util/crypt/crypt_dev.go
```diff
@@ -71,29 +70,22 @@ func (crypt *Device) CloseCryptDevice(path string) error {
...
 
-	input := string(password)
-	fmt.Println()
 	if confirm {
-		fmt.Print("Confirm the Key: ")
-		password2, err := terminal.ReadPassword(int(syscall.Stdin))
+		pass2, err := sypgp.AskQuestionNoEcho("Confirm the passphrase: ")
 		if err != nil {
-			sylog.Fatalf("Error parsing the key: %s", err)
+			return "", fmt.Errorf("unable parsing input: %s", err)
```
jchavezlavalle: same here.. "unable to parse the passphrase"

pkg/util/crypt/crypt_dev.go
```diff
@@ -71,29 +70,22 @@ func (crypt *Device) CloseCryptDevice(path string) error {
...
 		}
-		input2 := string(password2)
-		fmt.Println()
-		if input != input2 {
-			return "", errors.New("Keys don't match")
+		if pass1 != pass2 {
```
jchavezlavalle: using variable names like pass1, pass2, pass3 are not a good practice. Better to have in this case a "passphrase" variable instead of "pass1" variable and then a "confirmPassphrase" variable instead of "pass2" variable to avoid the added consecutive numbers..

jchavezlavalle (changes requested):
some comments around , thanks @WestleyK :+1:

pkg/util/crypt/crypt_dev.go
WestleyK: Thanks :+1: fixed

pkg/util/crypt/crypt_dev.go
```diff
@@ -227,11 +215,11 @@ func copyDeviceContents(source, dest string, size int64) error {
...
-			return fmt.Errorf("Unable to read the the file %s", source)
+			return fmt.Errorf("unable to read the the file: %s: %s", source, err)
 		}
 		numWritten, err := syscall.Write(destFd, buffer)
 		if err != nil {
-			return fmt.Errorf("Unable to write to destination %s", dest)
+			return fmt.Errorf("unable to write to destination: %s: %s", dest, err)
 		}
 
 		if numRead != numWritten {
```
jchavezlavalle, line 225: Just one single suggestion more @WestleyK ..On line 226: `sylog.Debugf("numRead != numWritten")` I don't think on debugging logs we should output code conditions, a developer may not know what these variables mean. Better to log out the meaning of this condition, so an explanation like "the byte size read is different than the ones written", please reword as needed :+1:  thanks!

jchavezlavalle (changes requested):

pkg/util/crypt/crypt_dev.go
WestleyK: I already asked @schebro if that was necessary (https://github.com/sylabs/singularity/pull/3885#discussion_r301214749). Maybe I could change it  to something like this?

```go
sylog.Debugf("numRead = %v, numWritten = %v", numRead, numWrite)
```

Thoughts?

jchavezlavalle: better, but still doesn't explain what does `numRead` and `numWritten` variables even mean

schebro: @WestleyK: @mem's 3930 fixed this part.

WestleyK:
Closing. Some/most of these fixes not necessary anymore, and too much conflict. I'll make a new PR if later if necessary.

PR #3900, opened 2019-07-09
Check for cryptsetup at build time

### LINKED ISSUE #3819: Save location of cryptsetup during buildtime in config files
## Version of Singularity:

Currently cryptsetup utility is directly called with full path (/sbin/cryptsetup) as PATH variable is not set in Go runtime. We can address the problem in a similar way to mksquashfs by saving the location of cryptsetup in a configuration file during build time and using that location at runtime.

### Expected behavior

NA

### Actual behavior

NA

Write here.

mem:
Currently the location of cryptsetup is hardcoded to /sbin/cryptsetup.
Check for the actual path at build time, and record the result in
buildcfg. At runtime, the path is retrieved from the configuration file
first, and if that's empty, it falls back to the path recorded at build
time. This gives the system administrator full control over which
cryptsetup program is used.

Fixes: #3819


internal/pkg/util/bin/bin.go
```diff
@@ -0,0 +1,70 @@
...
+	sync.Once
+	cryptsetup string
+	err        error
+}
+
+// Cryptsetup looks for the "cryptsetup" program returning the absolute
+// path to it. If the cryptsetup program is not available, this function
+// returns a non-nil error.
+func Cryptsetup() (string, error) {
+	cache.Do(func() {
```
schebro, line 37: This is nice. Didn't know "Do" before.

internal/pkg/util/bin/bin.go
```diff
@@ -0,0 +1,70 @@
...
+			return "", errors.New("unable to obtain path to cryptsetup program")
+		}
+
+		path = buildcfg.CRYPTSETUP_PATH
+	} else {
+		switch fi, err := os.Stat(path); {
+		case err != nil:
+			return "", errors.Wrapf(err, "unable to stat %s", path)
+
+		case fi.IsDir():
```
schebro, line 76: Should we force one way or the other? Either store directory or the absolute path to the binary?

pkg/util/crypt/crypt_dev.go
```diff
@@ -169,7 +174,7 @@ func (crypt *Device) FormatCryptDevice(path, key string) (string, string, error)
 	defer lock.Release(fd)
 
 	nextCrypt := getNextAvailableCryptDevice()
-	cmd = exec.Command("/sbin/cryptsetup", "luksOpen", loopdev, nextCrypt)
+	cmd = exec.Command(cryptsetup, "luksOpen", loopdev, nextCrypt)
 	cmd.Dir = "/dev"
```
schebro, line 178: Running cryptsetup in "/dev" is harmless in this case I feel. But, as you noted in slack, you can delete this and pass "/dev"+loopdev I guess

pkg/util/crypt/crypt_dev.go
```diff
@@ -260,7 +265,12 @@ retry:
...
 	}
 
-	cmd := exec.Command("/sbin/cryptsetup", "luksOpen", loopDev, nextCrypt)
+	cryptsetup, err := bin.Cryptsetup()
+	if err != nil {
+		return "", err
+	}
+
+	cmd := exec.Command(cryptsetup, "luksOpen", loopDev, nextCrypt)
 	cmd.Dir = "/dev"
```
schebro, line 274: Same as above.

schebro (changes requested):

internal/pkg/util/bin/bin.go
mem: I'm not happy with this but I'm following the example set by mksquashfs: in the configuration file the user can provide a directory or a filename. If it's a directory, it's the directory containing the program, just like this code does.

pkg/util/crypt/crypt_dev.go
mem: I'll fix that in a different PR.

pkg/util/crypt/crypt_dev.go
mem: ditto

internal/pkg/util/bin/bin.go
mem: :-)

mlocal/scripts/find_program
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
```
jscook2345: Is there a good reason to use this over standard programs like `which` and `command -v`?

mlocal/scripts/find_program
```diff
@@ -0,0 +1,23 @@
...
+for dir in $2 ; do
+	fn="${dir}/$1"
+	if test -x "${fn}" ; then
+		# found
+		echo "${fn}"
+		exit 0
+	fi
+done
+
+exit 1
```
jscook2345: This will cause a failure in the mconfig (what I'm seeing in CircleCI) if the package is not installed rather than with the correct output when it's not found.

Bad
```
checking: cryptsetup... Exited with code 1
```

Good
```
 checking: cryptsetup... no

unable to find the cryptsetup program, need package cryptsetup-bin
```

jscook2345:
Does this package now need to be included with our RPMs and other CI services?

mlocal/scripts/find_program
mem: Fixed

mlocal/scripts/find_program
mem: When I was doing this I had something else in mind, and I ended up backtracking. So, right, the script is equivalent to command -v (except that command -v considers aliases, too, but I'll ignore that).

schebro (approved):

PR #3925, opened 2019-07-12
Add check to ensure gzip compression

### LINKED ISSUE #2717: RFE: Always use gzip for Squashfs
## Version of Singularity:

For 3.2.0

### Expected behavior

Always use gzip.

### Actual behavior

Mksquashfs defaults to host settings.

### LINKED ISSUE #2724: Conditionally include dependent programs with Singularity installation
Dependent programs should be optionally installed in a Singularity specific directory ($libexecdir perhaps), so we know what options are available that maybe version specific (e.g. mksquashfs in #2717).

### LINKED ISSUE #3437: enable mksquashfs format compression options  at build time.
## Version of Singularity: 3.1.0

related to #1456

one of our user  provided us an image built with singularity-3.1.0 built with squashfs with default mksquasfs compression method xz 

1) will it be possible for him  pass the -comp option to mksquasfh at build time in order to build gzip compressed images instead of xz one ?

rebuild mksquashfs with COMP_DEFAULT=gzip will not be the ideal solution ;-)

as stated on the doc:  https://www.sylabs.io/guides/3.0/user-guide/build_a_container.html
```
Beware that it is possible to build an image on a host and have the 
image not work on a different host. This could be because of the default 
compressor supported by the host. For example, when building an image on 
a host in which the default compressor is |xz| and then trying to run 
that image on a CentOS 6 node, where the only compressor available is 
|gzip|.
```

allowing "containers creator" to choose the compression format for the image will be a nice addition 

regards

Eric

ikaneshiro:

- creates and uses a squashfs packer within the image build system to replace bare calls to the `mksquashfs` binary
- adds three 4K squashfs files to test this packer under different conditions
- creates a small test squashfs before building in order to parse the superblock and ensure `gzip` compression will be used in the resulting SIF

- Fixes #2717
- Fixes #3437
- Closes #2724 


internal/pkg/build/build.go
```diff
@@ -135,14 +139,113 @@ func newBuild(defs []types.Definition, conf Config) (*Build, error) {
...
+	if err != nil {
+		return false, fmt.Errorf("while creating temporary file for squashfs source: %v", err)
+	}
+
+	srcf.Write([]byte("Test File Content"))
+	srcf.Close()
+
+	f, err := ioutil.TempFile(tmpdir, "squashfs-gzip-comp-test-")
+	if err != nil {
+		return false, fmt.Errorf("while creating temporary file for squashfs: %v", err)
```
mem, line 182: this is leaving the temporary file from line 192 behind.

internal/pkg/build/build.go
```diff
@@ -135,14 +139,113 @@ func newBuild(defs []types.Definition, conf Config) (*Build, error) {
...
+	}
+	f.Close()
+
+	flags := []string{"-noappend"}
+	if err := s.Create([]string{srcf.Name()}, f.Name(), flags); err != nil {
+		return false, fmt.Errorf("while creating squashfs: %v", err)
```
mem, line 188: this is leaving both temporary files behind.

mem (changes requested):
LGTM, but there are some lingering temporary files.

internal/pkg/build/build.go
ikaneshiro: Both of these files will be taken care of when the build cleans up overall. I can see a potential debugging case where we would want at least the test squashfs to stay around if we use `--no-cleanup` in case we have a squashfs tool related error

mem: ah, I missed the tmpdir argument, thanks

mem (approved):

PR #3930, opened 2019-07-12
Sif encryption cleanup

mem:
This PR contains multiple clean ups done while trying to figure out why I was seeing a corrupted SIF file when trying to use an encrypted container.

* Obtain the key from the environment instead of interactively
* Make sure we are using LUKS 2 crypt devices
* Reduce the space required to create an encrypted container

### DIFF
--- a/pkg/util/crypt/crypt_dev.go
+++ b/pkg/util/crypt/crypt_dev.go
@@ -20,7 +20,6 @@ import (
 	"github.com/sylabs/singularity/internal/pkg/util/bin"
 	"github.com/sylabs/singularity/pkg/util/fs/lock"
 	"github.com/sylabs/singularity/pkg/util/loop"
-	"golang.org/x/crypto/ssh/terminal"
 )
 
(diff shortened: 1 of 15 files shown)


pkg/util/crypt/crypt_dev.go
```diff
@@ -115,18 +86,26 @@ func (crypt *Device) FormatCryptDevice(path, key string) (string, string, error)
...
 		sylog.Debugf("Error creating temporary crypt file")
-		return "", "", err
+		return "", err
 	}
 	defer cryptF.Close()
 
 	// Truncate the file taking the squashfs size and crypt header into account
 	// Crypt header is around 2MB in size. Slightly over-allocate to be safe
-	devSize := fSize + 4*1024*1024 // 4MB for LUKS header
+	devSize := fSize + 1*1024*1024 // 16MB for LUKS header
```
schebro, line 95: Please update the comment

schebro (approved):
Looks good to me.

PR #3945, opened 2019-07-17
Add support for public keys when encrypting images

mem:
Extend the existing SIF file encryption so that it's possible to specify
a PEM file with a public key instead of a plain text key. When the
public key is specified, a secret will be automatically generated and it
will be encrypted with the specified public key.

fixing or enhancing.

- Fixes #

### DIFF
--- a/internal/pkg/util/fs/mount/mount_linux.go
+++ b/internal/pkg/util/fs/mount/mount_linux.go
@@ -6,6 +6,7 @@
 package mount
 
 import (
+	"encoding/base64"
 	"fmt"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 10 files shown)


internal/pkg/util/fs/mount/mount_linux.go
```diff
@@ -306,13 +306,15 @@ func GetSizeLimit(options []string) (uint64, error) {
 }
 
 // GetKey returns key value for image options
-func GetKey(options []string) (string, error) {
+func GetKey(options []string) ([]byte, error) {
 	for _, opt := range options {
 		if strings.HasPrefix(opt, "key=") {
-			return strings.TrimPrefix(opt, "key="), nil
+			// TODO(mem): base64 encode the key?
```
schebro: You might need to base64 encode here. Otherwise, "," or "." in the key may create problems.

schebro (changes requested):
Looks good other than the encoding comment. I can start from this and work on private key decryption.

internal/pkg/util/fs/mount/mount_linux.go
mem: Thanks, fixed!

schebro (approved):

PR #3952, opened 2019-07-18
Client support for SCS library multi-arch tags

### LINKED ISSUE #3937: Multiple architectures per tag support with library push/pull
## Version of Singularity:

Support pulling containers from the SCS library where there are images for multiple architectures per tag.

### Expected behavior

On e.g. `armv6`, `arm64`, `ppc64le`, `amd64` hosts `singularity pull library://dtrudg-sylabs/test:latest` should return a container of the native architecture, or raise an error if one is not available.

It should be possible to specify a specific architecture with a flag to pull down a container on a host where the architecture does not match.

dtrudg:

This PR is the Singularity client-facing portion of code to support multi-arch tags in the SCS library.

Images can be built for multiple architectures, e.g. `amd64` / `arm64` / `ppc64le` and all pushed to the library using the same container name and tag. `singularity` clients pulling the tag will then retrieve the correct architecture for the local machine.

An image for a different architecture to the host can be pulled by using e.g. `singularity pull --arch amd64`

Note r.e. testing - it's not possible to add good e2e tests for this until an SCS library endpoint with archtags support is available. This is noted as needed, though.

-  Fixes #3937

- Read the Guidelines for Contributing, and this PR conforms to the stated requirements.
- Added changes to the CHANGELOG if necessary according to the Contribution Guidelines
- Added tests to validate this PR and tested this PR locally with a make testall
- Based this PR against the appropriate branch according to the Contribution Guidelines
- Added myself as a contributor to the Contributors File


cmd/internal/cli/actions.go
```diff
@@ -155,13 +156,10 @@ func handleLibrary(imgCache *cache.Handle, u, libraryURL string) (string, error)
 
 	imageRef := libraryhelper.NormalizeLibraryRef(u)
 
-	libraryImage, existOk, err := c.GetImage(ctx, imageRef)
+	libraryImage, err := c.GetImage(ctx, runtime.GOARCH, imageRef)
```
mem, line 199: this one will cause the command to ignore the arch specified on the command line: if a particular arch is requested, and the code goes to check if the image is in the cache, since it's not getting the value specified on the command line and it's instead going back to runtime.GOARCH, you will end up with two different things. The hash should prevent this image from clobbering the "right" one, but I still think you'd be getting the wrong image, unless I'm missing something.

cmd/internal/cli/actions.go
```diff
@@ -171,7 +169,7 @@ func handleLibrary(imgCache *cache.Handle, u, libraryURL string) (string, error)
 	} else if !exists {
 		sylog.Infof("Downloading library image")
 
-		if err = libraryhelper.DownloadImageNoProgress(ctx, c, imagePath, imageRef); err != nil {
+		if err = libraryhelper.DownloadImageNoProgress(ctx, c, imagePath, runtime.GOARCH, imageRef); err != nil {
```
mem, line 216: same here

cmd/internal/cli/pull.go
```diff
@@ -48,8 +49,21 @@ var (
...
 )
 
+// --library
+var pullArchFlag = cmdline.Flag{
+	ID:           "pullArchFlag",
+	Value:        &PullArch,
+	DefaultValue: runtime.GOARCH,
+	Name:         "arch",
+	Usage:        "Architecture to pull from library",
+	EnvKeys:      []string{"PULL_ARCH"},
```
mem, line 65: I don't think the environment variable is useful in this case, is it?

cmd/internal/cli/pull.go
```diff
@@ -48,8 +49,21 @@ var (
 	unauthenticatedPull bool
 	// PullDir is the path that the containers will be pulled to, if set
 	PullDir string
+	// PullArch is the architecture for which containers will be pulled from the
+	// SCS library
+	PullArch string
```
mem: this is a bad pattern that has crept into this particular part of the code: these variables don't need to be exported, we are passing them down to the appropriate functions as arguments.

go.mod
```diff
@@ -120,5 +120,6 @@ require (
 replace (
 	github.com/Sirupsen/logrus => github.com/sirupsen/logrus v1.0.5
 	github.com/opencontainers/image-tools => github.com/sylabs/image-tools v0.0.0-20181006203805-2814f4980568
+	github.com/sylabs/scs-library-client => github.com/dctrud/scs-library-client v0.3.1-0.20190717221055-79bcf0162d6b
```
mem: I don't think we want to commit this, right?

internal/app/singularity/push.go
```diff
@@ -97,5 +103,17 @@ func LibraryPush(file, dest, authToken, libraryURI, keyServerURL, remoteWarning
 	}
 	defer f.Close()
 
-	return libraryClient.UploadImage(context.Background(), f, r.Host+r.Path, r.Tags, "No Description", &progressCallback{})
+	return libraryClient.UploadImage(context.Background(), f, r.Host+r.Path, arch, r.Tags, "No Description", &progressCallback{})
```
mem, line 106: funny, I would have expected the library to look inside the file to figure out the architecture, not that the client has to tell it that information.

internal/pkg/build/remotebuilder/remotebuilder.go
```diff
@@ -147,7 +147,9 @@ func (rb *RemoteBuilder) Build(ctx context.Context) (err error) {
 
 		imageRef := library.NormalizeLibraryRef(bi.LibraryRef)
 
-		if err = library.DownloadImageNoProgress(ctx, c, rb.ImagePath, imageRef); err != nil {
+		// TODO - remotebuilder is *only* amd64 right now, but this will change
+		// in future.
+		if err = library.DownloadImageNoProgress(ctx, c, rb.ImagePath, "amd64", imageRef); err != nil {
```
mem: but then there should be a check locally that if a remote build is requested, then the requested architecture _must_ be amd64, no?

mem (changes requested):

cmd/internal/cli/actions.go
dtrudg: @mem - as I read the singularity code, handleLibrary is used for `run shell exec` etc. with a `library://` URI. Here it only makes sense to use the host arch - the `--arch` option isn't availble. You can only use that with `singularity pull` which uses a different code path for interacting with the library.

cmd/internal/cli/actions.go
dtrudg: ditto the above comment

cmd/internal/cli/pull.go
dtrudg: Do you want me to change them all to private then / or set this one just private? I was being consistent here.

go.mod
dtrudg: This is a WIP PR - It has to be tested against non-merged scs-library-client and a test library version, so I'm putting this here so it works for testing. It'll be a proper tagged scs-library-client before the WIP comes off.

internal/app/singularity/push.go
dctrud: It does - see just above where `arch` comes from...

```
arch, err := sifArch(file)
```

internal/pkg/build/remotebuilder/remotebuilder.go
dtrudg: there is no `--arch` flag for build, only pull. We can't request something else yet. Any remote build that is run (from any host) will end up as an amd64 image.

cmd/internal/cli/pull.go
dtrudg: I can't think of a concrete example - possibly some CI workflow? but I don't think there is harm?

internal/app/singularity/push.go
dtrudg: `<edit>` ahh I see what you mean...`</edit>`

The library does look into the file after upload. Since things upload into S3 directly now this has to be async, after the upload is completed. In order the arch->tag->Image is correct immediately we have to send it here. The `scs-library-client` is setting the tag on the image immediately after upload, and that tag is arch specific now.

internal/pkg/build/remotebuilder/remotebuilder.go
dtrudg: After chatting with @EmmEff who is working on the builder in this area I will need to enable the `--arch` flag only in combination with the `--remote` option of `singularity build`, and will pass-through architecture requests in an `arch` key on the `BuilderRequirements` map to future-proof things even if alternate archs aren't handled by the builder right now.

mem: Ok. Does this need changes in this PR or not? I just want to make sure nothing is falling thru the cracks.

dtrudg: @mem - yes there will be changes to this PR - hopefully this afternoon, maybe tomorrow morning. Thanks for tracking this!

cmd/internal/cli/actions.go
dtrudg: Note that `--arch` is now an option for `pull` and `build --remote` only. 

* On `pull` you should get an image back of the requested architecture.
* On `build --remote` you are asking the remote builder to make an image for the specified arch.

`--arch` isn't in any other code paths as it doesn't make sense to try and perform other actions with images that don't match the host architecture

mem (approved):

cmd/internal/cli/pull.go
```diff
@@ -47,10 +48,23 @@ var (
 	keyServerURL = "https://keys.sylabs.io"
 	// unauthenticatedPull when true; wont ask to keep a unsigned container after pulling it.
 	unauthenticatedPull bool
-	// pullDir is the path that the containers will be pulled to, if set.
+	// PullDir is the path that the containers will be pulled to, if set.
```
WestleyK: Why is this changed? It was correct before.

WestleyK (approved):
Looks good to me :+1: That one comment is unnecessary.

cmd/internal/cli/pull.go
dtrudg: oof - this PR has been around for long enough that a bunch of refactoring stuff has touched the same areas. Looks like I've missed an odd thing in the merge this time :-)

EmmEff (approved):

cmd/internal/cli/pull.go
dtrudg: Fixed in ec4f5ee

Issue #3828, opened 2019-06-28
Port e2e tests to use singularity execution context


cclerget:
Port following tests to use singularity execution context (https://github.com/sylabs/singularity/blob/master/e2e/internal/e2e/singularitycmd.go) :

- [x] docker -> https://github.com/sylabs/singularity/blob/master/e2e/docker/docker.go
- [x] env -> https://github.com/sylabs/singularity/blob/master/e2e/env/env.go
- [x] help -> https://github.com/sylabs/singularity/blob/master/e2e/help/help.go
- [x] imgbuild -> https://github.com/sylabs/singularity/blob/master/e2e/imgbuild/imgbuild.go
- [x] instance -> https://github.com/sylabs/singularity/blob/master/e2e/instance/instance.go
- [x] pull -> https://github.com/sylabs/singularity/blob/master/e2e/pull/pull.go
- [x] push -> https://github.com/sylabs/singularity/blob/master/e2e/push/push.go
- [x] remote -> https://github.com/sylabs/singularity/blob/master/e2e/remote/remote.go
- [x] version -> https://github.com/sylabs/singularity/blob/master/e2e/version/version.go
- [x] inspect -> https://github.com/sylabs/singularity/blob/master/e2e/inspect/inspect.go

PR #3962, opened 2019-07-22
Port e2e push test

### LINKED ISSUE #3837: Port e2e push tests
Port push tests https://github.com/sylabs/singularity/blob/master/e2e/push/push.go to use singularity execution context (https://github.com/sylabs/singularity/blob/master/e2e/internal/e2e/singularitycmd.go)

gvallee:

Port e2e push test to new API

- Fixes #3837 
- Addresses #3828


e2e/push/push.go
```diff
@@ -39,7 +40,7 @@ func (c *ctx) testPushCmd(t *testing.T) {
 		desc          string // case description
 		dstURI        string // destination URI for image
 		imagePath     string // src image path
-		expectSuccess bool   // singularity should exit with code 0
+		expectSuccess bool   // expection regarding the test's success
```
gvallee:
```suggestion
		expectSuccess bool   // expectation regarding the test's success
```

e2e/push/push.go
```diff
@@ -39,7 +40,7 @@ func (c *ctx) testPushCmd(t *testing.T) {
 		desc          string // case description
 		dstURI        string // destination URI for image
 		imagePath     string // src image path
-		expectSuccess bool   // singularity should exit with code 0
+		expectSuccess bool   // expectation regarding the test's success
```
jscook2345: In other tests I've seen this field changed to something like`expectExitCode` and than using that value instead of doing further checking around the `boolean` later on.

What do you think?

e2e/push/push.go
```diff
@@ -75,26 +76,24 @@ func (c *ctx) testPushCmd(t *testing.T) {
...
-
-			case tt.expectSuccess && err != nil:
-				// FAIL: expecting success, failed
-
-				t.Logf("Running command:\n%s\nOutput:\n%s\n", cmd, out)
-				t.Errorf("unexpected failure: %v", err)
-
-			case !tt.expectSuccess && err == nil:
-				// FAIL: expecting failure, succeeded
+			args := tt.dstURI
```
ikaneshiro, line 83: Can we make `args` a `[]string` and append to it so we dont need to split a string by a separator in `e2e.WithArgs()`?

ikaneshiro (changes requested):

e2e/push/push.go
gvallee: I had that question before and no it does not work. In a nutshell, the command is very sensitive to empty slots, meaning an element in the slice that would not be defined, which would make the command to fail even if everything looks fine (only a double space would appear when displaying the elements of the commands). I had very bad luck with unit tests, e2e tests and in the context of my MPI tool.
In other words, that is by far the safer way to deal with it and i did not manage to have another solution that never create a problem.

gvallee: I had that question before and the short answer is no: so far, i always end up with Go assuming that an element in the slice is empty, which will make the command fail (when displaying the args, the problem can be seen through a double space). So this is the only reliable way i know of dealing with it.
I had the problem in unit tests, e2e tests and the MPI tool.

e2e/push/push.go
gvallee: Done

e2e/push/push.go
ikaneshiro: Could you throw a comment on the code summarizing these comments so that others hopefully dont run into this issue when working with this file?

jscook2345: Not handling an empty argument seems like a short coming. @gvallee Can you open an issue for that so we can make things more consistent across the board?

jscook2345 (approved):

e2e/push/push.go
gvallee: Will do and as you can see I am having problems with Github, my comments appearing/disappearing from my view of the PR... :(

jscook2345:
@gvallee 

See @mem 's changes here: https://github.com/sylabs/singularity/pull/3840

They will need to be applied before the tests will pass.

gvallee:
I do not understand the API modification introduced in #3840, it will take me some time to figure it out.

e2e/push/push.go
```diff
@@ -75,26 +76,23 @@ func (c *ctx) testPushCmd(t *testing.T) {
...
+			// an empty elements to the list when passing it to the command, which
+			// will create a failure.
+			args := tt.dstURI
+			if tt.imagePath != "" {
+				args = tt.imagePath + " " + args
 			}
+
+			e2e.RunSingularity(
+				t,
+				tt.desc,
```
mem: replace this with `e2e.AsSubtest(tt.desc)`
```suggestion
				e2e.AsSubtest(tt.desc),
```

mem (changes requested):

e2e/push/push.go
```diff
@@ -75,26 +76,23 @@ func (c *ctx) testPushCmd(t *testing.T) {
...
+			// We create the list of arguments using a string instead of a slice of
+			// strings because using slices of strings most of the type ends up adding
+			// an empty elements to the list when passing it to the command, which
+			// will create a failure.
+			args := tt.dstURI
+			if tt.imagePath != "" {
+				args = tt.imagePath + " " + args
 			}
+
+			e2e.RunSingularity(
```
mem: if you change this to `c.env.RunSingularity`, what I wrote below actually works :-|

```suggestion
			c.env.RunSingularity(
```

e2e/push/push.go
gvallee: That is the part I do not understand, semantically, I guess I do not understand the meaning of `AsSubtest`, which, without spending more than 2 minutes on it, would infer that I want to create a subtest... which is not the case. But I will look it up, I just need time. :)

mem: Welcome to my world :-)

The function `e2e.RunSingularity` previously had a signature that was something like `func(t *testing.T, name string, ...)`.

I changed it to `func(t *testing.T, cmdPath string, ...)`

In the previous signature, `name` is the name of the subtest (as in `t.Run(name, f)`). If name was empty (`""`), then it _did not_ run a subtest. If name was not empty, it run the command as a subtest.

To make that less awkward and less magic, I changed it to the explicit `e2e.AsSubtest(name)`, which is passed along with all the other options.

But then I added the `cmdPath` argument :-(

The `cmdPath` argument specifies the path to singularity.

I guess I could add a `type CmdPath string` to make this a compiler error (at the cost of making the code a little weird).

mem: And if you don't want a subtest, just don't pass e2e.AsSubtest. I suggested that to match the behavior of the previous code.

gvallee: Thanks for the explanation, it truly helps. I like the fact that we can specify the Singularity binary and now that I have your explanation, I understand the subtest part too. I am totally okay with the new API. :)

jscook2345:
@gvallee @mem 

The PR that keeps on giving ;)

I built locally using this branch and this is what `make check` gave me on `macos`.

```
grapplehammer:singularity justin$ make -C builddir check
 CHECK golangci-lint
cmd/plugin_manifestgen/main.go:32:23: `marshalling` is a misspelling of `marshaling` (misspell)
                sylog.Fatalf("While marshalling manifest to json: %s", err)
```

gvallee:
This is not related to this PR at all but I will fix it as soon as I can.

ikaneshiro (approved):

mem (approved):

PR #3968, opened 2019-07-23
[E2E] tests for 'key' (take 2!)

### LINKED ISSUE #2839: Create e2e tests for `singularity key`

### LINKED ISSUE #3533: [E2E] Add tests for key importing/exporting

WestleyK:

This PR will add some E2E testing for the `key` command.

- Fixes: https://github.com/sylabs/singularity/issues/3533
- Fixes: https://github.com/sylabs/singularity/issues/2839

- Related to: https://github.com/sylabs/singularity/pull/3539

<br>


e2e/key/key.go
```diff
@@ -0,0 +1,235 @@
...
+
+func (c *ctx) singularityKeyNewpair(t *testing.T) {
+	tests := []struct {
+		name       string
+		consoleOps []e2e.SingularityConsoleOp
+	}{
+		{
+			name: "newpair",
+			consoleOps: []e2e.SingularityConsoleOp{
+				e2e.ConsoleSendLine("e2e test key; do not use"),
```
mem: what's the "; do not use" bit supposed to be?

Can you add a comment next to each command so that it documents what it is?

For example:

```
e2e.ConsoleSendLine("e2etests"), // password
e2e.ConsoleSendLine("e2etests"), // password confirmation
```

Also note you can have a helper function that takes a series of strings and spits out a series of e2e.SingularityConsoleOp. Something like:

```
func buildConsoleLines(lines ...string) []e2e.SingularityConsoleOp {
    consoleLines := make([]e2e.SingularityConsoleOp, 0, len(lines))
    for _, line := range lines {
        consoleLines = append(consoleLines, e2e.ConsoleSendLine(line))
    }
}
```

e2e/key/key.go
```diff
@@ -0,0 +1,235 @@
...
+
+func (c *ctx) singularityKeyList(t *testing.T) {
+
+	printOut := func(t *testing.T, r *e2e.SingularityCmdResult) {
+		t.Log("stdout from 'key list' : ", string(r.Stdout))
+	}
+
+	c.env.RunSingularity(
+		t,
+		e2e.WithPrivileges(false),
```
mem: just for the record, `e2e.WithPrivileges(false)` has no practical effect. I've been meaning to remove that argument so that it's more evident that you need to call it only to gain privilege, not to drop it.

e2e/key/key.go
```diff
@@ -0,0 +1,235 @@
...
+				e2e.ConsoleSendLine("E2E test key; do not use"),
+				e2e.ConsoleSendLine("e2etests"),
+				e2e.ConsoleSendLine("e2etests"),
+				e2e.ConsoleSendLine("n"),
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		c.env.RunSingularity(
```
mem, line 151: you should probably add `e2e.AsSubtest(tt.name)` to the list of ops.

e2e/key/key.go
```diff
@@ -0,0 +1,235 @@
...
+
+		return cmd
+	}
+
+	for _, tt := range tests {
+		c.env.RunSingularity(
+			t,
+			e2e.WithPrivileges(false),
+			e2e.WithCommand("key"),
+			e2e.WithArgs(prepCmd(tt.exportPath, tt.secret, tt.armor)...),
```
mem: notice that this would be easier to read if you add another field to the test struct that contains the args, replacing the armor and secret booleans. That field would look like this:

```
    args: []string{"export",},
    args: []string{"export", "--armor"},
    args: []string{"export", "--secret"},
    args: []string{"export", "--secret", "--armor"},
```

that makes it straightforward to see what's happening.

e2e/key/key.go
```diff
@@ -0,0 +1,235 @@
...
+func (c *ctx) singularityKeyImport(t *testing.T) {
+	tests := []struct {
+		name       string
+		exportPath string
+		consoleOps []e2e.SingularityConsoleOp
+	}{
+		{
+			name:       "import public binary",
+			exportPath: c.publicExportPath,
+			consoleOps: []e2e.SingularityConsoleOp{},
```
mem: no need to have this if it's empty. `nil` as a slice is perfectly fine.

e2e/key/key.go
```diff
@@ -0,0 +1,235 @@
...
+		},
+		{
+			name:       "import private binary",
+			exportPath: c.privateExportPath,
+			consoleOps: []e2e.SingularityConsoleOp{
+				e2e.ConsoleSendLine("e2etests"),
```
mem: the first line is the passphrase of the existing key

the second the is passphrase for the new key

the third is the passphrase confirmation for the new key

I would say the second set should use a different passphrase and it should test that the exported key has the correct passphrase.

e2e/key/key.go
```diff
@@ -0,0 +1,235 @@
...
+		{
+			name:       "import public ascii",
+			exportPath: c.publicExportASCIIPath,
+			consoleOps: []e2e.SingularityConsoleOp{},
+		},
+		{
+			name:       "import private ascii",
+			exportPath: c.privateExportASCIIPath,
+			consoleOps: []e2e.SingularityConsoleOp{
+				e2e.ConsoleSendLine("e2etests"),
```
mem: same here

e2e/key/key.go
```diff
@@ -0,0 +1,235 @@
...
+		return []string{"import", exportPath}
+	}
+
+	for _, tt := range tests {
+		c.singularityResetKeyring(t) // Remove the tmp keyring before each import
+		c.env.RunSingularity(
+			t,
+			e2e.WithPrivileges(false),
+			e2e.WithCommand("key"),
+			e2e.WithArgs(prepCmd(tt.exportPath)...),
```
mem: same comment as before

e2e/key/key.go
```diff
@@ -0,0 +1,235 @@
...
+	c := &ctx{
+		env:                    env,
+		publicExportPath:       filepath.Join(env.TestDir, "public_key.asc"),
+		publicExportASCIIPath:  filepath.Join(env.TestDir, "public_ascii_key.asc"),
+		privateExportPath:      filepath.Join(env.TestDir, "private_key.asc"),
+		privateExportASCIIPath: filepath.Join(env.TestDir, "private_ascii_key.asc"),
+		keyRing:                filepath.Join(env.TestDir, "sypgp-test-keyring"),
+	}
+
+	if err := os.Setenv("SINGULARITY_SYPGPDIR", c.keyRing); err != nil {
```
mem, line 341: this should be passed as part of the environment of the command to be run, not as part of the environment in the running test.

e2e/key/key.go
```diff
@@ -0,0 +1,235 @@
...
+		panic(fmt.Sprintf("unable to set keyring: %s", err))
+	}
+
+	return func(t *testing.T) {
+		//t.Run("singularityKeyNewpairAndExport", c.singularityKeyNewpair)
+		//t.Run("singularityKeyImportAndExport", c.singularityKeyImport)
+		t.Run("singularityKeyNewpair", c.singularityKeyNewpair) // Generate a newpair (required for the other tests)
```
mem: the "singularity" prefix is probably redundant.

e2e/key/key.go
```diff
@@ -0,0 +1,235 @@
...
+		t.Run("singularityKeyExport", c.singularityKeyExport)   // Export the newpair
+		t.Run("singularityKeyImport", c.singularityKeyImport)   // Import the newpair (this will delete the old, tmp keyring before importing)
+		t.Run("singularityKeyExport", c.singularityKeyExport)   // Re-export them, again (this will catch any issues if Singularity cant import correctly)
+		t.Run("singularityKeyImport", c.singularityKeyImport)   // Finally, Re-import them
```
mem: please do not rely on a specific order for tests. If you need for things to happen in a particular order, make them part of the same test.

mem (changes requested):

e2e/key/key.go
WestleyK: I'm setting `SINGULARITY_SYPGPDIR` to a tmp directory so it can easily be removed after each `import`. I'm not specifically trying to test `SINGULARITY_SYPGPDIR`, but its helpful for testing other `key` commands.

e2e/key/key.go
WestleyK: Yes, that would be better for testing, but I'm trying to keep this PR small and simple. Because each `import` and `export` is ran twice. We could something like:

```go
// e2eKeyPassword is the password for the e2e test key, it will change as test run.
var e2eKeyPassword = "e2etests"
```

Then just set `e2eKeyPassword` to the new password in the test. _Theres probably a simpler way to do this..._

So for now, I added a couple "import fail tests" that will use the wrong password when importing them.

e2e/key/key.go
gvallee: I think @mem is pointing at another potential problem: setting `SINGULARITY_SYPGPDIR` as it is done here will prevent us from safely run the test in parallel since the caller (i.e., this function) may end up set that environment variable and start various e2e test at the same time. We have the same problem with the image cache (I am actually working on this) but we would need code modifications to solve this problem in a clean way. I would therefore personally suggest to create an issue on the topic and just put a comment here saying that this is a temporary solution.

To be more precise, my idea to solve this type of problems is to extend the `singularityCmd` structure in `singularitycmd.go` to include both a `cacheDir` and a `syPGPDir` elements. If not defined (empty string), during the execution of `RunSingularity`, we simply create temporary directories and use them for PGP and the image cache.

e2e/key/key.go
```diff
@@ -0,0 +1,230 @@
...
+func RunE2ETests(env e2e.TestEnv) func(*testing.T) {
+	c := &ctx{
+		env:                    env,
+		publicExportPath:       filepath.Join(env.TestDir, "public_key.asc"),
+		publicExportASCIIPath:  filepath.Join(env.TestDir, "public_ascii_key.asc"),
+		privateExportPath:      filepath.Join(env.TestDir, "private_key.asc"),
+		privateExportASCIIPath: filepath.Join(env.TestDir, "private_ascii_key.asc"),
+		keyRing:                filepath.Join(env.TestDir, "sypgp-test-keyring"),
+	}
+
```
gvallee, line 340:
```suggestion

// todo: setting the environment variable here is a temporary solution until we
// have the option of creating a temporary directory on a per-test basis.
// When we will have such an option, we should remove the code setting `SINGULARITY_SYPGPDIR`.
```

gvallee (changes requested):
See comments for details but I would suggest adding a todo note in the code and open a new issue (not yet done) to have a thread-safe way to specify a temporary directory with `SINGULARITY_SYPGPDIR` (similar to #3987 but for `SYPGP`).

gvallee:
About the use of the `SINGULARITY_SYPGPDIR` environment variable, I opened issue #4143 and will be working on it.

gvallee:
@WestleyK PR #4144 should simply this e2e tests since a clean new temporary keyring is now by default created for every test.

e2e/key/key.go
WestleyK: May not be necessary anymore? https://github.com/sylabs/singularity/pull/4144 is merged.

gvallee (approved):
lgtm

mem (approved):

PR #3970, opened 2019-07-23
Fix --contain issues. Fix bind destination creation

### LINKED ISSUE #3949: "can't remount" under "cleaned up" /home and /tmp in -c mode
## Version of Singularity:
<details><summary>
Initially ran into a failure with 2.6.1 where failing behavior varies (might be related to #3530)
</summary>

```shell
/tmp/d > mkdir -p {~,/tmp}/d; touch {~,/tmp}/d/file;  for d in /etc /etc/apt ~ ~/d /tmp /tmp/d; do echo $d ; builtin cd $d; singularity exec -c -B $d /tmp/out1/containers/images/bids/bids-mriqc--0.15.0.sing ls -l $d >/dev/null; echo "exited with $?"; done                        
/etc
exited with 0
/etc/apt
exited with 0
/home/yoh
WARNING: Not mounting requested bind point (already mounted in container): /home/yoh
exited with 0
/home/yoh/d
exited with 0
/tmp
exited with 0
/tmp/d
ls: cannot access '/tmp/d': No such file or directory
exited with 2
```
</details>

and then replicated with 3.1.1 (debian system both times)

### Expected behavior

leading subdirectories will be provided within cleaned up environment, similar to how it is done to any other directory (overlays?)

### Actual behavior

Just fails to bind mount a directory under /home or /tmp whenever running with `-c` flag:
```shell
/tmp/d > mkdir -p {~,/tmp}/d; touch {~,/tmp}/d/file; for d in /etc /etc/apt /data/famface/ ~ ~/d /tmp /tmp/d; do echo $d ; builtin cd $d; singularity exec -c -B $d /tmp/out1/containers/images/bids/bids-mriqc--0.15.0.sing ls -l $d | head -n 1; done
/etc
total 2868
/etc/apt
total 72
/data/famface/
total 16
/home/yoh
total 1317320
/home/yoh/d
FATAL:   container creation failed: mount ->/home/yoh/d error: can't remount /home/yoh/d: no such file or directory
/tmp
total 172088
/tmp/d
FATAL:   container creation failed: mount ->/tmp/d error: can't remount /tmp/d: no such file or directory
```
in above - /data/famface is not in the container, everything works out just fine.  ~/d and /tmp/d -- fail to bind mount

Interestingly, if I provide `-W` path where I will have those /tmp/d and /home/yoh/d precreated -- seems to "work" for /tmp but not for home:
```shell
/tmp/d > mkdir -p {~,/tmp}/d; touch {~,/tmp}/d/file; for d in ~/d /tmp/d; do echo $d ; builtin cd $d; singularity exec -c -W /home/yoh/.tmp/singtmp.HUk5El -B $d /tmp/out1/containers/images/bids/bids-mriqc--0.15.0.sing ls -l $d | head -n 1; done                                               
/home/yoh/d
FATAL:   container creation failed: mount ->/home/yoh/d error: can't remount /home/yoh/d: no such file or directory
/tmp/d
total 0
```
but you can mention that actually `/tmp/d` lists no files (total 0) so it doesn't bind mount /tmp/d from outside which has a `file` in it

PS removed `E: could not determine git repository root` in the above dumps which were coming from my prompt setup and shouldn't be relevant here

cclerget:

- Fix --contain issue with both overlay/underlay and writable flags.
- Introduces overrideDir in session manager to simulate a bind mount during directories/files/symlinks creation, it should address many issues related to binding giving error like "no such file directory" during mount call.
- Nested binds are already allowed, with `overrideDir` destination directories will be automatically created.
- Fix RPC mount error type to now return `syscall.Errno` type

- Fixes #3949

### DIFF
--- a/internal/pkg/util/fs/layout/manager.go
+++ b/internal/pkg/util/fs/layout/manager.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, Sylabs Inc. All rights reserved.
+// Copyright (c) 2018-2019, Sylabs Inc. All rights reserved.
 // This software is licensed under a 3-clause BSD license. Please consult the
 // LICENSE.md file distributed with the sources of this project regarding your
 // rights to use or distribute this software.
@@ -49,6 +49,7 @@ type Manager struct {
 	rootPath string
 	entries  map[string]interface{}
 	dirs     []*dir
+	ovDirs   map[string]string
 }
 
 func (m *Manager) checkPath(path string, checkExist bool) (string, error) {
@@ -86,6 +87,9 @@ func (m *Manager) createParentDir(path string) {
 				d := &dir{mode: m.DirMode, uid: uid, gid: gid}
 				m.entries[p] = d
 				m.dirs = append(m.dirs, d)
+				if ovDir, ok := m.ovDirs[filepath.Dir(p)]; ok {
+					m.ovDirs[p] = filepath.Join(ovDir, filepath.Base(p))
+				}
 			}
 		}
 	}
@@ -101,6 +105,9 @@ func (m *Manager) SetRootPath(path string) error {
 		m.entries = make(map[string]interface{})
 	} else {
 		return fmt.Errorf("root path is already set")
+	}
+	if m.ovDirs == nil {
+		m.ovDirs = make(map[string]string)
 	}
 	if m.dirs == nil {
 		m.dirs = make([]*dir, 0)
@@ -150,6 +157,20 @@ func (m *Manager) AddSymlink(path string, target string) error {
 	m.createParentDir(filepath.Dir(p))
 	m.entries[p] = &symlink{uid: os.Getuid(), gid: os.Getgid(), target: target}
 	return nil
+}
+
+// overrideDir will substitute another directory to the one associated
+// to directory located by path
+func (m *Manager) overrideDir(path string, realpath string) {
+	m.ovDirs[path] = realpath
+}
+
+// GetOverridePath returns the real path for the session path
+func (m *Manager) GetOverridePath(path string) (string, error) {
+	if p, ok := m.ovDirs[path]; !ok {
+		return p, nil
+	}
+	return "", fmt.Errorf("no override directory %s", path)
 }
 
 // GetPath returns the full path of layout path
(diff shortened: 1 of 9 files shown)


gvallee (approved):
LGTM

internal/pkg/util/fs/layout/manager.go
```diff
@@ -152,6 +159,20 @@ func (m *Manager) AddSymlink(path string, target string) error {
...
+func (m *Manager) overrideDir(path string, realpath string) {
+	m.ovDirs[path] = realpath
+}
+
+// GetOverridePath returns the real path for the session path
+func (m *Manager) GetOverridePath(path string) (string, error) {
+	if p, ok := m.ovDirs[path]; !ok {
+		return p, nil
+	}
+	return "", fmt.Errorf("no override directory %s", path)
```
WestleyK, line 173:
```suggestion
	return "", fmt.Errorf("no override directory: %s", path)
```

~All~ **most** of the error returns have the `:` right before a variable, for consistency maybe better to add it

internal/pkg/util/fs/layout/session_linux.go
```diff
@@ -75,12 +78,74 @@ func (s *Session) FinalPath() string {
...
+			flags, _ := mount.ConvertOptions(point.Options)
+
+			// ignore anything which is not a bind mount point
+			if flags&syscall.MS_BIND == 0 {
+				continue
+			}
+
+			// check if the bind source exists
+			if err := syscall.Stat(point.Source, st); err != nil {
+				sylog.Warningf("skipping mount of %s: %s", point.Source, err)
```
WestleyK: Same here...

```suggestion
				sylog.Warningf("skipping mount of: %s: %s", point.Source, err)
```

WestleyK (approved):
Looks good to me, just two optional suggestions :+1:

internal/pkg/util/fs/layout/manager.go
mem: I don't think the result would be understandable in this case: `no override directory: /foo/bar`, it's better without the colon.

WestleyK: @mem, I agree, it does sound a bit better in this case.

WestleyK (approved):
Looks good to me! Thanks :+1:

PR #3979, opened 2019-07-23
Better handle the path to the Singularity system configuration file

### LINKED ISSUE #3976: Better handle path to the Singularity config file
## Version of Singularity:

All.

### Expected behavior

Paths should be built using `filepath.Join()` or something similar instead of `"/a/path"+"/a/subdir".

### Actual behavior

The code has a lot of `buildcfg.SYSCONFDIR + "/singularity/singularity.conf"`.

N/A.

gvallee:

Better handle the path to the Singularity system configuration file by calling a specific API instead of the creation of a string (which did not even use `filepath.Join()`).

- Fixes #3976 

### DIFF
--- a/cmd/internal/cli/actions_linux.go
+++ b/cmd/internal/cli/actions_linux.go
@@ -35,6 +35,7 @@ import (
 	"github.com/sylabs/singularity/internal/pkg/util/env"
 	"github.com/sylabs/singularity/internal/pkg/util/exec"
 	"github.com/sylabs/singularity/internal/pkg/util/fs"
+	"github.com/sylabs/singularity/internal/pkg/util/fs/files"
 	"github.com/sylabs/singularity/internal/pkg/util/user"
 	singularityConfig "github.com/sylabs/singularity/pkg/runtime/engines/singularity/config"
 )
@@ -132,7 +133,7 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 
 	engineConfig := singularityConfig.NewConfig()
 
-	configurationFile := buildcfg.SYSCONFDIR + "/singularity/singularity.conf"
+	configurationFile := files.GetSysConfigFile()
 	if err := config.Parser(configurationFile, engineConfig.File); err != nil {
 		sylog.Fatalf("Unable to parse singularity.conf file: %s", err)
 	}
--- a/cmd/singularity/cli_test.go
+++ b/cmd/singularity/cli_test.go
@@ -14,8 +14,8 @@ import (
 	"syscall"
 	"testing"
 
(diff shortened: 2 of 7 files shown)


cmd/internal/cli/actions_linux.go
```diff
@@ -129,7 +130,7 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 
 	engineConfig := singularityConfig.NewConfig()
 
-	configurationFile := buildcfg.SYSCONFDIR + "/singularity/singularity.conf"
+	configurationFile := files.GetSysConfigFile()
```
mem, line 136: there's the `syfs` package in `pkg/syfs`. I think that's a better place for this code.

mem (changes requested):

cmd/internal/cli/actions_linux.go
gvallee: I put it there because it is not user-facing but internal; what is in `syfs` is related to the user-level file system details (e.g., user-level config files). Here we deal with the system-level Singularity configuration file.

jscook2345 (approved):
Approve pending agreement on filesystem location.

cmd/internal/cli/actions_linux.go
mem: Having two packages that deal with the layout of the singularity files is confusing. There's some pending clean up that needs to happen in terms of what's available to other packages (if that's what you mean by "user facing") but I would rather have one problem pending of a fix and not two.

gvallee: Well, issues are there for that. I would prefer to do the right thing here, i.e., put the package in the right place. I guess we need to see if we can find consensus about this because on who review my PRs, it is once one way, once the other way and this is not good to change code based on what appears to me to be personal preferences. In other words, let's set some guidelines.

cclerget: The rule is pretty simple here, if the package use `buildcfg` constants, it's an internal package because it would be unusable as public package otherwise.

An other approach could be to add some get accessor in `buildcfg` package to have things like `buildcfg.GetSingularityConfigFile()`

gvallee: I think there is agreement that the current location is adequate. We also have issue #3995 to finalize the discussion and agree on the potential next steps.

mem: Agreement is something else.

I hadn't realized that you sneaked the change into a different unrelated code review. You could have said so.

mem (approved):
The part where the changes were requested was merged using a different PR.

WestleyK (approved):
Looks good to me! Thanks :+1:

jmstover (approved):

PR #3997, opened 2019-07-25
Add e2e fakeroot tests and Singularity execution profile

### LINKED ISSUE #3206: e2e testing for --fakeroot flag
The `--fakeroot` flag should minimally be tested using all of the action commands and also with build.

### LINKED ISSUE #3770: e2e tests fix pending issues of 3760
#3760 have some pending issues requiring fix :

- https://github.com/sylabs/singularity/pull/3760#discussion_r295723098
- https://github.com/sylabs/singularity/pull/3760#discussion_r295879331

cclerget:

Introduces Singularity execution profiles (`SingularityProfile`), in order to run tests with different profiles without rewriting tests.

Introduces a `ExecContext` holding `testing.T`, `TestEnv` and `SingularityProfile` to simplify function prototype.

Add fakeroot tests for actions, imgbuild and instances as part of Singularity execution profile.

Actions, imgbuild and instances are now run under different profiles depending of requirements and command support.

**EDIT**:

In order to generalize profile usage for all tests, actually each tests groups (actions/docker/imgbuild/... directories) have an entrypoint called `RunE2ETests`, the idea is to loop over all available profiles allow us to run all tests for a profile with `-run TestE2E/*/User/`.

So we could introduce a e2e suite manager to register test groups like :

```
var userOnlyProfile = []e2e.SingularityProfile{e2e.UserProfile}
var allProfile = e2e.Profiles

func testOne(ec *e2e.ExecContext) {
    t, env, profile := ec.Get()
    ...
}

func testTwo(ec *e2e.ExecContext) {
    t, env, profile := ec.Get()
    ...
}

func RunE2ETests(tg *e2e.TestGroup) {
    const parallel = true
    tg.AddTest("Test1", testOne, parallel, allProfile)
    tg.AddTest("Test2", testTwo, !parallel, userOnlyProfile)
}
```

and manage them from `suite.go` with something like :

```
s := e2e.NewSuite(t, env)
groups := map[string]func(*e2e.TestGroup){
    "ACTIONS":  actions.RunE2ETests,
    ...
}
for name, fn := range groups {
    s.AddGroup(name, fn)
}
s.Run() // will take care of running parallel and non parallel tests
```

- Fixes #3206
- Fixes #3770


e2e/internal/e2e/profile.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	u, err := user.GetPwUID(uint32(origUID))
+	if err != nil {
+		t.Fatalf("failed to retrieve original user information: %s", err)
+	}
+	return u
+}
+
+// Name returns the profile name.
+func (p SingularityProfile) Name() string {
```
mem, line 154: idiomatic Go code calls this function "String" because the formatter looks for that function when it has to format a type using "%s"

e2e/internal/e2e/profile.go
```diff
@@ -0,0 +1,126 @@
...
+
+import (
+	"testing"
+
+	"github.com/sylabs/singularity/internal/pkg/fakeroot"
+	"github.com/sylabs/singularity/internal/pkg/test/tool/require"
+	"github.com/sylabs/singularity/internal/pkg/util/user"
+)
+
+// SingularityProfile represents a Singularity execution profile.
```
mem: it would be nice to to have a description that does not read like "Thing is a thing". Specifically, it might be useful to clarify what a "execution profile" is.

e2e/internal/e2e/profile.go
```diff
@@ -0,0 +1,126 @@
...
+type SingularityProfile uint8
+
+const (
+	// UserProfile is the execution profile with a regular user.
+	UserProfile SingularityProfile = iota + 1
```
mem: is there a reason why you need "iota + 1"?

e2e/internal/e2e/profile.go
```diff
@@ -0,0 +1,126 @@
...
+	FakerootProfile,
+	UserNamespaceProfile,
+	// NEED FIX: disabled until tests are fixed to run it correctly
+	// RootUserNamespaceProfile,
+}
+
+// CheckRequirements checks and ensures that the corresponding
+// execution profile has all the requirements, the current test
+// is skipped if not.
+func (p SingularityProfile) CheckRequirements(t *testing.T) {
```
mem: looking at the implementation this function is not _checking_ something (in the sense that it's looking at whether something is provided or present). Instead this function is stating what the requirements are. I'm guessing this is to be used as `RootProfile.CheckRequirements(t)`. Maybe call it `RootProfile.EnsureRequierements(t)` or `RootProfile.AssertRequirements(t)` or `RootProfile.Require(t)`?

e2e/internal/e2e/profile.go
```diff
@@ -0,0 +1,126 @@
...
+	switch p {
+	case UserNamespaceProfile, RootUserNamespaceProfile:
+		require.UserNamespace(t)
+	case FakerootProfile:
+		require.UserNamespace(t)
+		// now check that current user has valid mappings
+		// in /etc/subuid and /etc/subgid
+		if _, err := fakeroot.GetIDRange(fakeroot.SubUIDFile, uint32(origUID)); err != nil {
+			t.Skipf("fakeroot configuration error: %s", err)
```
mem: if this is an error, the test should not be skipped, it should fail.

If the test is to be skipped, the message should be more along the lines of "skipping test, such and such is not available"

e2e/internal/e2e/profile.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+		if _, err := fakeroot.GetIDRange(fakeroot.SubGIDFile, uint32(origUID)); err != nil {
+			t.Skipf("fakeroot configuration error: %s", err)
```
mem: same here

e2e/internal/e2e/profile.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+	}
+}
+
+// Match returns if one the provided profiles match the current one.
+func (p SingularityProfile) Match(profiles ...SingularityProfile) bool {
```
mem: I'm trying to understand the use case here.

```
if RootProfile.Match(listOfProfiles) {
```

what would the code be trying to do in that case?

I imagine something like this would be easier to read:

```
if listOfProfiles.Match(RootProfile) {
```

but without understanding the problem this is solving I cannot say.

e2e/internal/e2e/profile.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	return false
+}
+
+// WithArgs returns Singularity arguments for the corresponding
+// profile if any.
+func (p SingularityProfile) WithArgs(args []string) []string {
+	switch p {
+	case FakerootProfile:
+		return append([]string{"--fakeroot"}, args...)
```
mem: these functions here (the previous one, this one, the next one) are what makes me think the code could be implemented differently, but I don't understand the use case enough to figure out of there's an advantage to this or not.

If you had:

```
type Profile interface {
    Privileged() bool
    Args() []string
    User() (*user.User, error)
}
```

profiles would be types implementing the interface, so for example:

```
type RootProfile struct{}

func (p RootProfile) Privileged() bool {
    return true
}

func (p RootProfile) Args() []string {
    return nil
}

func (p RootProfile) User() (*user.User, error) {
    return user.GetPwUID(0)
}
```

it's a little more typing, but the implementations are much more straightforward.

Whether this makes sense or not hinges on what the purpose of Match is.

e2e/internal/e2e/profile.go
```diff
@@ -0,0 +1,126 @@
+// Copyright (c) 2019, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package e2e
```
mem, line 6: the code feels like it should be in a separate package.

mem (changes requested):
I only looked at profile.go. I intentionally did not look at the other changes because I was trying to figure out if the completely new code stands on its own or not. What I think is that there's some missing documentation to make it clear what some of this functions are for and what SingularityProfile represents.

e2e/actions/actions.go
```diff
@@ -34,214 +106,248 @@ func (c *actionTests) actionRun(t *testing.T) {
...
 		},
 		{
+			// as /root directory is always present in images, this test
+			// could not run with RootProfile, RootUserNamespaceProfile and
+			// FakerootProfile
 			name: "NoHome",
-			argv: []string{"--no-home", c.env.ImagePath, "ls", "-ld", user.Dir},
+			argv: []string{"--no-home", env.ImagePath, "ls", "-ld", u.Dir},
 			exit: 2,
+			excludedProfiles: []e2e.SingularityProfile{
```
gvallee, line 319: If I understand correctly, this means that the spirit behind the profile is basically to assume that new profiles are supported by default, right? I am asking the question because trying to understand what work will be required when we will include a new profile, more precisely understand how it will be possible to identify all the tests that need to be updated.

e2e/actions/actions.go
```diff
@@ -631,158 +710,167 @@ func (c *actionTests) RunFromURI(t *testing.T) {
...
-		t.Run("exec", c.actionExec)
-		// stdin/stdout pipe
-		t.Run("STDPIPE", c.STDPipe)
-		// action_URI
-		t.Run("action_URI", c.RunFromURI)
-		// Persistent Overlay
-		t.Run("Persistent_Overlay", c.PersistentOverlay)
-		// shell interaction
-		t.Run("Shell", c.actionShell)
+		tests := map[string]func(*e2e.ExecContext){
```
gvallee: Maybe add a comment such as: "when executing the E2E tests, the framework will execute the tests in the context of multiple profiles, which are orthogonal to the test itself. In other words, the framework will implicitly loop over all existing profiles and execute the test in its context. The test developers are in charge of explicitly excluding profiles that may not be relevant to the test by using a `PreRun` condition."

e2e/internal/e2e/context.go
```diff
@@ -0,0 +1,30 @@
...
+// NewExecContext creates a new test execution context.
+func NewExecContext(t *testing.T, e TestEnv, p SingularityProfile) *ExecContext {
+	return &ExecContext{
+		t:       t,
+		env:     e,
+		profile: p,
+	}
+}
+
+// Get returns information embedded by the test execution
```
gvallee:
```suggestion
// Get returns information embedded in the test execution
```

e2e/internal/e2e/profile.go
```diff
@@ -0,0 +1,163 @@
...
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/sylabs/singularity/internal/pkg/fakeroot"
+	"github.com/sylabs/singularity/internal/pkg/test/tool/require"
+	"github.com/sylabs/singularity/internal/pkg/util/user"
+)
+
```
gvallee, line 16: Maybe add a comment such as: "When executing the E2E tests, the framework will execute the tests in the context of multiple profiles, which are orthogonal to the test itself. In other words, the framework will implicitly loop over all existing profiles and execute the test in its context. The test developers are in charge of explicitly excluding profiles that may not be relevant to the test by using a PreRun condition."

e2e/internal/e2e/profile.go
gvallee: I strongly disagree, this is pure E2E functionality so far, I do not see a single reason why it should be somewhere else. Moving this outside of E2E would be extremely confusing since it would be the only major capability that is specific to E2E and not in the E2E subtree and under the E2E package.

e2e/internal/e2e/profile.go
gvallee: Because 0 is no profile if I understand correctly the explanations in the rest of the code

e2e/internal/e2e/profile.go
```diff
@@ -0,0 +1,163 @@
...
+// SingularityProfile represents a Singularity execution profile.
+type SingularityProfile uint8
+
+// We start from one so we could compare SingularityProfile
+// variable with zero, meaning there is no profile
+const (
+	// UserProfile is the execution profile with a regular user.
```
gvallee, line 26:
```suggestion
	// 0 is equivalent to no profile.
	// UserProfile is the execution profile with a regular user.
```

e2e/internal/e2e/profile.go
gvallee: The PR is full of examples of this and I think fairly easy to understand (which does not mean it does not deserve a few additional comments). In a nutshell, based on my understanding, all profiles are assumed to be applicable by default so if a test is not relevant for a specific profile, the developer must explicitly exclude it. `Match` is a utility function to help developers to handle that situation by comparing the current profile with a pre-defined list of profiles. I included a suggested modification to the comments.

e2e/internal/e2e/profile.go
```diff
@@ -0,0 +1,163 @@
...
+		if _, err := fakeroot.GetIDRange(fakeroot.SubUIDFile, uint32(origUID)); err != nil {
+			t.Fatalf("fakeroot configuration error: %s", err)
+		}
+		if _, err := fakeroot.GetIDRange(fakeroot.SubGIDFile, uint32(origUID)); err != nil {
+			t.Fatalf("fakeroot configuration error: %s", err)
+		}
+	}
+}
+
+// Match returns if one the provided profiles match the current one.
```
gvallee:
```suggestion
// Match returns if one of the provided profiles match the current one.
//
// Practically, this function is part of the utility functions that are available to developers to handle
// tests that are relevant only to a sub-set of the existing profiles. Tests are assumed by default to
// be relevant in all existing profiles. If not, it is the responsibility of the developer to explicitly exclude
// profiles that are not applicable through a `PreRun()` function. This function is part of the
// capabilities provided by the E2E framework to ease such a task.
```

e2e/internal/e2e/singularitycmd.go
```diff
@@ -290,12 +290,15 @@ func WithDir(dir string) SingularityCmdOp {
 	}
 }
 
-// WithPrivileges sets whether a singularity command must be
-// executed with privileges or not. PreRun, InRun, PostRun
-// are also executed with privileges.
-func WithPrivileges(privileged bool) SingularityCmdOp {
+// WithProfile sets the Singularity execution profile, this
```
gvallee, line 294: Would it make sense to provide a `WithoutProfile` function as well? Ultimately I think the question is to decide what is the default behavior if any: are all the profiles included by default? are none of the profiles included by default? If no default is assumed, having `WithProfile()` and `WithoutProfile()` would make sense to me. If a default is assumed, we could decide to only rely on `PreRun()` functions.

e2e/internal/e2e/profile.go
gvallee: I think the real question here is what are the exact default assumption. For instance what is the intended behavior when the test developer does not specific any `PreRun()` function for filtering profiles and does not use `WithProfile()` either. It would be very useful to explicitly state it in a comment and ensure that my understanding is what @cclerget actually has in mind (and come to agreement as a group of course).
I also find the code a little confusing because of this since it is unclear what the default behavior is. For instance, just by reading the code, I thought the default behavior was to include all the profiles but the `PreRun()` code shows that it will fail is no profile is defined.

e2e/internal/e2e/singularitycmd.go
```diff
@@ -411,16 +414,30 @@ func RunSingularity(t *testing.T, cmdPath string, cmdOps ...SingularityCmdOp) {
 		t.Errorf("ExpectExit is missing in cmdOps argument")
 		return
 	}
+	if s.profile == 0 {
```
gvallee:
```suggestion
	if s.profile == 0 {
		// No profile specified
```

e2e/internal/e2e/singularitycmd.go
```diff
@@ -411,16 +414,30 @@ func RunSingularity(t *testing.T, cmdPath string, cmdOps ...SingularityCmdOp) {
...
+		availableProfiles := make([]string, len(Profiles))
+		for i, p := range Profiles {
+			availableProfiles[i] = p.String()
+		}
+		profiles := strings.Join(availableProfiles, ", ")
+		t.Errorf("you must specify a profile, available profiles are %s", profiles)
```
gvallee, line 425: So the default is no profile and leads to an error, meaning that it is assumed that a profile *must* be explicitly specified by the test?

e2e/internal/e2e/singularitycmd.go
```diff
@@ -411,16 +414,30 @@ func RunSingularity(t *testing.T, cmdPath string, cmdOps ...SingularityCmdOp) {
...
+		return
+	}
+	privileged := s.profile.privileged()
```
gvallee, line 429: This default is only because E2E is executed under a `sudo` command, right?

gvallee (changes requested):
I have one major question: what is the default behavior regarding profiles? Are some/all profiles assumed to always be applicable? Are developers to explicitly enumerate all the applicable profiles in the test itself?
Otherwise, I find the PR extremely interesting since it would allow us to drastically improve coverage and potentially minimize the time required to do so.

e2e/internal/e2e/profile.go
mem: Because from what I can see this is very tightly related functionality within the new code and is not as tightly related to the rest of the existing code in the package. Sure, there's a high level theme here (e2e) but all the new code exists for a specific purpose.

Keep in mind that in Go the name of the package becomes part of the client code. Consider the `io` package and its `Reader` type. In the client code you have `io.Reader`, which is very expressive. The `io` package provides functionality that is closely related to `sys` and `os`. They could have provided everything in the `os` package (and that was the partly the case a long time ago) and have `os.Reader`. What's an `os.Reader`? To clarify this, they would have named the `Reader` type `IOReader` or `FileReader` or something like that (and I believe that or something similar was in fact the name at some point).

What I'm saying for example is what's better from the point of view of someone reading the code: `e2e.Match` or `profile.Match`?

e2e/internal/e2e/profile.go
mem: but that means this could benefit from `InvalidProfile` or `UnknownProfile` for example.

e2e/internal/e2e/profile.go
mem: If the question that is going to be generally asked (the use case) is "does this list contain this item?" then the second form is clearer:

```
if listOfProfiles.Contains(RootProfile) {
```

If the most common question is going to be "is my current profile in this list?", then this might be clearer:

```
if currentProfile.In(listOfProfiles) {
```

(and notice that I switched from "RootProfile" to "currentProfile" in the second case).

e2e/internal/e2e/profile.go
cclerget: Will add `UnknownProfile` to be explicit

e2e/internal/e2e/profile.go
cclerget: The goal of `Match` is to match the current profile against a list of profiles, so yes `In` is fine too. The first form `listOfProfiles.Contains(RootProfile)` would involve much steps to setup and make the things more complex IMO.

e2e/internal/e2e/profile.go
cclerget: The idea is to have static and not alterable profiles, because there is nothing to set dynamically, so we can avoid allocation step

e2e/internal/e2e/profile.go
cclerget: the `profile.Match` is better IMO, and actually calls are always done with `profile.Match` or depending of "profile" variable name, since `Match` is a function of `SingularityProfile`, so it's fine to keep it as is actually

e2e/actions/actions.go
cclerget: Yeah, this is clearly not a good example, was lazy so maybe it's better to take the approach of grouping similar tests by eligible "profiles" instead of adding a filter field in test table ?

e2e/internal/e2e/singularitycmd.go
cclerget: My preference goes for always provide a profile (to avoid the question "huh wait what is the default behavior ?" when reading code. As profile in this PR are redundant with `WithPrivileges`, we now have something like when using profile :

`UserProfile` -> `WithPrivileges(false)`
`RootProfile` -> `WithPrivileges(true)`
`UserNamespaceProfile` -> `WithPrivileges(false)` inject `--userns` option for commands `shell,test,exec,run,instance start`
`FakerootProfile` -> `WithPrivileges(false)` inject `--fakeroot` option for commands `shell,test,exec,run,instance start,build`

e2e/internal/e2e/singularitycmd.go
cclerget: yes

e2e/internal/e2e/singularitycmd.go
cclerget: `sudo` is used once when running `sudo go test ...`, after if `privileged` is `true` singularity is executed with root identity via a dedicated privileged thread by calling `e2e.Privileged()`

e2e/internal/e2e/profile.go
cclerget: So actually a profile must be provided (to be explicit), for case like actions/instances, those tests are actually executed with all profiles with few exceptions, but like I said previously those exceptions could be grouped via other test functions and by compatible profiles, so a developer could just add a `if profile.Match(compatibleProfiles)` at the beginning of the test function like it's done in imgbuild tests, not necessarily in a `PreRun` to skip a whole tests table (group). It really depends of tests, this is flexible to take both approach, skip an entire test function or just few tests via `PreRun`, it's left to developer

gvallee (approved):
lgtm

WestleyK:
Ping @cclerget, @mem, whats the status on this PR?

cclerget:
Closing, remain to add e2e tests for fakeroot

PR #4007, opened 2019-07-25
[3.3] Fixed issue when exporting key and 'sypgp' does not exist

### LINKED ISSUE #3993: key export of an empty keyring creates an empty file
## Version of Singularity:

v3.3.0 rc.3

### Expected behavior

```
$ ls pub.asc
ls: cannot access pub.asc: No such file or directory

$ singularity key export ./pub.asc
ERROR:   key export command failed: unable to open local keyring: open /home/vagrant/.singularity/sypgp/pgp-public: no such file or directory

$ ls pub.asc
ls: cannot access pub.asc: No such file or directory
```

### Actual behavior

```
$ ls pub.asc
ls: cannot access pub.asc: No such file or directory

$ singularity key export ./pub.asc
ERROR:   key export command failed: unable to open local keyring: open /home/vagrant/.singularity/sypgp/pgp-public: no such file or directory

$ ls pub.asc
pub.asc
```

- The exported file is empty.
- This does not occur when using the `--secret` flag.

WestleyK:

Fixed issue when trying to export a key when `sypgp` dir does not exist.

- Fixes #3993 

<br>

### DIFF
--- a/pkg/sypgp/sypgp.go
+++ b/pkg/sypgp/sypgp.go
@@ -799,6 +799,7 @@ func ExportPrivateKey(kpath string, armor bool) error {
 	if err != nil {
 		return err
 	}
+	defer file.Close()
 
 	if !armor {
 		// Export the key to the file
@@ -811,7 +812,6 @@ func ExportPrivateKey(kpath string, armor bool) error {
 		}
 		file.WriteString(keyText)
 	}
-	defer file.Close()
 
 	if err != nil {
 		return fmt.Errorf("unable to serialize private key: %v", err)
@@ -847,6 +847,7 @@ func ExportPubKey(kpath string, armor bool) error {
 	if err != nil {
 		return fmt.Errorf("unable to create file: %v", err)
 	}
+	defer file.Close()
 
 	if !armor {
 		err = entityToExport.Serialize(file)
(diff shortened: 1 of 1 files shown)


GodloveD:
This is a pretty minor issue so I think we'll leave it for a point release if/when we do one.  3.3 is very, very overdue at this point.

pkg/sypgp/sypgp.go
```diff
@@ -840,6 +843,11 @@ func ExportPubKey(kpath string, armor bool) error {
 		return err
 	}
 
+	file, err := os.Create(kpath)
```
sashayakovtseva, line 846: Why don't you defer file.Close() right after this error check?

WestleyK: The `defer file.Close()` is on line `854` or `862`. Although it makes sense to move it right under the error check. _I'll fix that..._

jscook2345:
@WestleyK Is there a PR applied to master that can be used instead of this specific PR to 3.3? We only really want to push fixes to release branches if the master PR cannot be used.

WestleyK:
@jscook2345, Yes, that already got merged https://github.com/sylabs/singularity/pull/4073

jscook2345:
I think we can close this than we will use #4073 . Thanks for the fix!

WestleyK:
>I think we can close this than we will use #4073 . Thanks for the fix!

Do you mean cherry-pick #4073 into 3.3?

WestleyK:
Ping @GodloveD, @jscook2345

jscook2345:
> > I think we can close this than we will use #4073 . Thanks for the fix!
> 
> Do you mean cherry-pick #4073 into 3.3?

Yes

WestleyK:
@jscook2345, this is the cherry-pick for that PR. Since the sypgp package was updated since 3.3, the fix is a little different.

PR #4009, opened 2019-07-26
Check SIF squashfs/ext3 partitions at load time

### LINKED ISSUE #3996: singularity inspect returns "FATAL:   exit status 255: "

cclerget:

Actually SIF squashfs/ext3 partitions are not checked, this PR add checks for both squashfs/ext3 system/data partitions

- Related to #3996 

### DIFF
--- a/internal/pkg/runtime/engines/singularity/container_linux.go
+++ b/internal/pkg/runtime/engines/singularity/container_linux.go
@@ -614,8 +614,22 @@ func (c *container) mountImage(mnt *mount.Point) error {
 		mountType = "squashfs"
 	}
 	err = c.rpcOps.Mount(path, mnt.Destination, mountType, flags, optsString)
-	if err != nil {
-		return fmt.Errorf("failed to mount %s filesystem: %s", mountType, err)
+	switch err {
+	case syscall.EINVAL:
+		if mountType == "squashfs" {
+			return fmt.Errorf(
+				"kernel reported a bad superblock for %s image partition, "+
+					"possible causes are that your kernel doesn't support "+
+					"the compression algorithm or the image is corrupted",
+				mountType)
+		}
+		return fmt.Errorf("%s image partition contains a bad superblock (corrupted image ?)", mountType)
+	case syscall.ENODEV:
+		return fmt.Errorf("%s filesystem seems not enabled and/or supported by your kernel", mountType)
+	default:
+		if err != nil {
+			return fmt.Errorf("failed to mount %s filesystem: %s", mountType, err)
+		}
 	}
 
 	return nil
--- a/pkg/image/sif.go
+++ b/pkg/image/sif.go
@@ -13,7 +13,6 @@ import (
 	"syscall"
 
 	"github.com/sylabs/sif/pkg/sif"
-	"github.com/sylabs/singularity/internal/pkg/sylog"
 )
 
 const (
@@ -22,16 +21,41 @@ const (
 
 type sifFormat struct{}
 
+func checkPartitionType(img *Image, fstype sif.Fstype, offset int64) (uint32, error) {
+	header := make([]byte, bufferSize)
+
+	if _, err := img.File.ReadAt(header, offset); err != nil {
+		return 0, fmt.Errorf("failed to read SIF partition at offset %d: %s", offset, err)
+	}
+
+	switch fstype {
+	case sif.FsSquash:
+		if _, err := CheckSquashfsHeader(header[:]); err != nil {
+			return 0, fmt.Errorf("error while checking squashfs header: %s", err)
+		}
+		return SQUASHFS, nil
+	case sif.FsExt3:
+		if _, err := CheckExt3Header(header[:]); err != nil {
+			return 0, fmt.Errorf("error while checking ext3 header: %s", err)
+		}
+		return EXT3, nil
+	case sif.FsEncryptedSquashfs:
+		return ENCRYPTSQUASHFS, nil
+	}
+
+	return 0, fmt.Errorf("unknown filesystem type %v", fstype)
+}
+
 func (f *sifFormat) initializer(img *Image, fileinfo os.FileInfo) error {
 	if fileinfo.IsDir() {
-		return fmt.Errorf("not a sif file image")
+		return debugError("not a sif file image")
 	}
 	b := make([]byte, bufferSize)
 	if n, err := img.File.Read(b); err != nil || n != bufferSize {
-		return fmt.Errorf("can't read first %d bytes: %s", bufferSize, err)
+		return debugErrorf("can't read first %d bytes: %s", bufferSize, err)
 	}
 	if !bytes.Contains(b, []byte(sifMagic)) {
-		return fmt.Errorf("SIF magic not found")
+		return debugError("SIF magic not found")
 	}
 
 	// Load the SIF file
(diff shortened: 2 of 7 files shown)


pkg/image/sif.go
```diff
@@ -130,8 +143,20 @@ func (f *sifFormat) initializer(img *Image, fileinfo os.FileInfo) error {
 			switch fstype {
```
Flamefire: Can't this code be merged with the above? Some function returning the FS type or an error?

cclerget: Yes it could, just for now does it work as expected for you ?

Flamefire: No. I still get the same indeterminate error on inspect and mount failure on shell/run/exec

cclerget: Well it's weird that the cluster kernel doesn't support the default compression which is gzip, when image is compressed with another compression algorithm you get an info message like `squashfs image was compressed with <comp>, if it failed to run, please contact image's author`.

I think squashfs filesystem is supported by your kernel (just in case check with `grep squashfs /proc/filesystems`), otherwise you should get a `no such device` error if it was the case.

From `mount` man, the only corresponding error for this issue is `EINVAL source had an invalid superblock`

pkg/image/sif.go
```diff
@@ -127,14 +152,16 @@ func (f *sifFormat) initializer(img *Image, fileinfo os.FileInfo) error {
...
-			case sif.FsExt3:
-				partition.Type = EXT3
-			default:
-				partition.Type = uint32(fstype)
+
+			htype, err := checkPartitionType(img, fstype, desc.Fileoff)
+			if err != nil && err != ErrUnknownFsType {
+				return fmt.Errorf("while checking data partition header: %s", err)
+			} else if err == ErrUnknownFsType {
+				img.Partitions[0].Type = uint32(fstype)
```
Flamefire: I know this was here before but it strikes me as sketchy. Doesn't this assume the constants of https://github.com/sylabs/singularity/blob/c5e243fc45fc5b55f315f7e39255cb95201d4f17/pkg/image/image.go#L21 and https://github.com/sylabs/sif/blob/640d2f2a0ef5219f0066d277e465b6c18a046fc8/pkg/sif/sif.go#L147 are kept in sync? Is this ensured at another place?

cclerget: We can't really sync those constants with the SIF ones because the image code is generic and must handle different format like sandbox, ext3 only, squashfs only, SIF (which can contain ext3/squashfs image), so it doesn't hurt to have different values there.

Flamefire: But this code basically says: We have an unknown file system type, just assume that the incoming type id (type is `Fstype` from SIF package) is a valid Type here (like `SQUASHFS`) even though we may not even know it.

This works for e.g. `fstype==FsEncryptedSquashfs` because it happens to hold `uint32(FsEncryptedSquashfs) == ENCRYPTSQUASHFS`, but what if the enums(?) go out of sync?

cclerget: I initially thought to keep the entry for a possible use by the caller as it's a public package but as it's only for overlay partition, we could probably just ignore those unknow fstype entries to avoid confusion, thoughts ?

Flamefire: Not sure if I understand correctly but IMO an unknown fs type is an error and should be reported as such. So I'd remove lines 159&160 and the check for `err != ErrUnknownFsType ` from 157

pkg/image/sif.go
```diff
@@ -82,24 +107,26 @@ func (f *sifFormat) initializer(img *Image, fileinfo os.FileInfo) error {
 			continue
 		}
 
+		// checks if the partition length is greater that the file
+		// size which may reveal a corrupted image (see issue #3996)
+		if fimg.Filesize < desc.Filelen {
```
Flamefire: Shouldn't this be `desc.Filelen + desc.Fileoff` or is that already included?

cclerget: Yes and thanks for the review and for spotting my stupid mistakes, now it's time for me to end this work day :)

Flamefire: Sure :)
May I suggest adding tests that running `singularity inspect` or `singularity shell/exec` on an incomplete image returns a meaningful error message? I guess there are already tests for running those commands on correct images. To create an incomplete image I'd take a normal one and chop of the last byte. Maybe additionally tests where you destroy the squashfs header. just to make sure this code is correctly triggered and to avoid regressions.

cclerget: I will add missing unit tests for SIF images in a first time in this PR, and create another PR for e2e tests

internal/pkg/runtime/engines/singularity/container_linux.go
```diff
@@ -614,8 +614,24 @@ func (c *container) mountImage(mnt *mount.Point) error {
 		mountType = "squashfs"
 	}
 	err = c.rpcOps.Mount(path, mnt.Destination, mountType, flags, optsString)
-	if err != nil {
-		return fmt.Errorf("failed to mount %s filesystem: %s", mountType, err)
+	switch err {
+	case syscall.EINVAL:
+		switch mountType {
+		case "squashfs":
```
sashayakovtseva: No need for switch here, can be simplified to 
```
if mountType == "squashfs" {
  return ....
}
return ....
```

Flamefire:
Update on my example corrupted image:

- Using `inspect` still results in `Unable to process command: exit status 255: `
- Using `shell` results in `FATAL:   image format not recognized`
- Adding `-d` to `shell`:

```
VERBOSE [U=1000,P=14703]   init()                        Spawn stage 1
DEBUG   [U=1000,P=14703]   startup()                     singularity runtime engine selected
VERBOSE [U=1000,P=14703]   startup()                     Execute stage 1
DEBUG   [U=1000,P=14703]   StageOne()                    Entering stage 1
DEBUG   [U=1000,P=14703]   Init()                        Entering image format intializer
DEBUG   [U=1000,P=14703]   Init()                        Check for image format sif
DEBUG   [U=1000,P=14703]   Init()                        sif format initializer returned: SIF image /home/tmp.sif is corrupted: wrong partition size
DEBUG   [U=1000,P=14703]   Init()                        Check for image format sandbox
DEBUG   [U=1000,P=14703]   Init()                        sandbox format initializer returned: not a directory image
DEBUG   [U=1000,P=14703]   Init()                        Check for image format squashfs
DEBUG   [U=1000,P=14703]   Init()                        squashfs format initializer returned: while parsing squashfs super block: not a valid squashfs image
DEBUG   [U=1000,P=14703]   Init()                        Check for image format ext3
DEBUG   [U=1000,P=14703]   Init()                        ext3 format initializer returned: file is not a valid ext3 image
FATAL   [U=1000,P=14703]   StageOne()                    image format not recognized
VERBOSE [U=1000,P=14695]   wait_child()                  stage 1 exited with status 255
```

So yes it works better, although the error message is still not that helpful unless one uses `-d`. Still better :+1: BTW: See the typo at   `Entering image format intializer`?

sashayakovtseva (approved):

PR #4029, opened 2019-07-28
Fix port handling

### LINKED ISSUE #4028: fails to map network port
## Version of Singularity:

3.3.0-rc.3

### Expected behavior
```
sudo singularity instance start --net --network-args "portmap=8070:8080/tcp" --network-args "portmap=50000:50000/tcp" --writable-tmpfs $(APPNAME).sif $(APPNAME)
```
fails with

```
ERROR:   container cleanup failed: no instance found with name test
FATAL:   container creation failed: error while setting network arguments: can't convert host port '50000': strconv.ParseInt: parsing "50000": value out of range
```

changing 50000 to 8088 works.

### Actual behavior

works fine with docker when passing to the docker ```-p 50000:50000```

```
sudo singularity instance start --net --network-args "portmap=8070:8080/tcp" --network-args "portmap=50000:50000/tcp" --writable-tmpfs $(APPNAME).sif $(APPNAME)
```

jmstover:
Looking at the port value was using `ParseInt()` which used a
`int16` value. Change to `ParseUint()` to cover the full range we
require with a `uint16` value.

Fixes: #4028 

### DIFF
--- a/pkg/network/network_linux.go
+++ b/pkg/network/network_linux.go
@@ -283,7 +283,7 @@ func (m *Setup) SetArgs(args []string) error {
 				if len(ports) != 1 && len(ports) != 2 {
 					return fmt.Errorf("portmap port argument is badly formatted")
 				}
-				if n, err := strconv.ParseInt(ports[0], 0, 16); err == nil {
+				if n, err := strconv.ParseUint(ports[0], 0, 16); err == nil {
 					pm.HostPort = int(n)
 					if pm.HostPort <= 0 {
 						return fmt.Errorf("host port must be greater than zero")
@@ -292,7 +292,7 @@ func (m *Setup) SetArgs(args []string) error {
 					return fmt.Errorf("can't convert host port '%s': %s", ports[0], err)
 				}
 				if len(ports) == 2 {
-					if n, err := strconv.ParseInt(ports[1], 0, 16); err == nil {
+					if n, err := strconv.ParseUint(ports[1], 0, 16); err == nil {
 						pm.ContainerPort = int(n)
 						if pm.ContainerPort <= 0 {
 							return fmt.Errorf("container port must be greater than zero")
--- a/pkg/network/network_linux.go
+++ b/pkg/network/network_linux.go
@@ -285,8 +285,8 @@ func (m *Setup) SetArgs(args []string) error {
 				}
 				if n, err := strconv.ParseUint(ports[0], 0, 16); err == nil {
 					pm.HostPort = int(n)
-					if pm.HostPort <= 0 {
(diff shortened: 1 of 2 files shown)


cclerget (commented):
Looks good to me, thanks @jmstover . Do you plan to include unit tests in this PR or in a separated one ?

pkg/network/network_linux.go
```diff
@@ -283,7 +283,7 @@ func (m *Setup) SetArgs(args []string) error {
 				if len(ports) != 1 && len(ports) != 2 {
 					return fmt.Errorf("portmap port argument is badly formatted")
 				}
-				if n, err := strconv.ParseInt(ports[0], 0, 16); err == nil {
+				if n, err := strconv.ParseUint(ports[0], 0, 16); err == nil {
```
mike-dubman, line 286: looks good, line 288 and 297 cannot happen now, i guess.

jmstover: I left those in for now, so that just incase the `ParseUint()` gets reverted somehow, we still have an error check in place. Though... I should make another small change.

jscook2345:
@cclerget 

I had him open another issue for us to add tests since he's not familiar with e2e. Hopefully I can take care of that soon 🤞

cclerget:
@jscook2345 I was thinking about unit tests as there is a test function where few cases could be covered easily for regressions here : https://github.com/sylabs/singularity/blob/master/pkg/network/network_linux_test.go#L188

jscook2345:
@cclerget 

Ah, ok. Yes that would be good for him to add to. @jmstover Please add some unit tests where @cclerget suggests. If you need help, please let me know!

pkg/network/network_linux.go
sashayakovtseva: imo useless checks should be removed in this PR.
How `ParseUint()` can be reverted?

mike-dubman: or better check that port is in range 0-65k and fail if not. If something gets reverted - it will still be correct.

jmstover: I was going to make it a `uint16` across the board since that fits the port range nicely. But in the Golang unix.SockaddrInet4 function, the port is an *int* value. So... that just breaks that idea, so I need to keep letting it use the lines like:
```
pm.HostPort = int(n)
```

As for getting reverting, it's _possible_ that a merge could have the old `ParseInt()` code. So, when it gets merged, it reverts it back to the incorrect way. Made more likely if someone force pushes.

jscook2345 (approved):
🥇 @jmstover !

jscook2345:
r_565529_WRQcK

PR #4030, opened 2019-07-28
Add another apt lock file to wait for before continuing

jscook2345:
CircleCI is continuously failing at the following step with the following error:

```
#!/bin/bash -eo pipefail
mkdir -p ${GOBIN}
sudo apt-get update -y && sudo apt-get install -y build-essential libssl-dev uuid-dev libgpgme11-dev squashfs-tools libseccomp-dev

E: Could not get lock /var/lib/apt/lists/lock - open (11: Resource temporarily unavailable)
E: Unable to lock directory /var/lib/apt/lists/
Exited with code 100
```

So we will try and wait for that lock file in the previous step.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -115,7 +115,7 @@ jobs:
       - run:
          name: Wait until apt completion
          command: |-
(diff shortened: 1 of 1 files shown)


jmstover (approved):

jscook2345:
This seemed to have zero effect, even though it looks like it should work. Closing.

PR #4031, opened 2019-07-28
Kill the apt processes that start on boot but do not finish before CI steps

jscook2345:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -117,13 +117,13 @@ jobs:
           working_directory: /tmp
           command: |-
             mkdir -p ${GOBIN}
-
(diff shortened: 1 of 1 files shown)


jmstover (approved):

jscook2345:
This is a quick solution to a problem blocking CI from running. It's not ideal, but it appears to work. We can return and improve this at a later date.

See also:

* https://support.circleci.com/hc/en-us/articles/360021256633-Apt-Get-Update-Is-Slow-Or-Locked
* https://discuss.circleci.com/t/could-not-get-lock-var-lib-apt-lists-lock/28337/6

Issue #2445, opened 2018-12-10
Singularity 3.0.1-1.el6 dependent on the host session's environment


mboisson:
## Version of Singularity:
3.0.1-1.el6

### Expected behavior

Starting singularity should be pretty much independent on the session's environment

### Actual behavior

Running singularity depends on the session's environment. 

### Steps to reproduce behavior

```
$ /opt/software/singularity-3.0/bin/singularity --version
singularity version 3.0.1-1.el6
``` 

```
$ /opt/software/singularity-3.0/bin/singularity shell --nv pytorch-18.01-py3.simg
WARNING: ldconfig execution error: exit status 1
INFO:    Unable to capture nvidia bind points: exit status 1
ERROR  : Failed to read JSON configuration from socket: Bad file descriptor
``` 

On the same system : 

```
$ /usr/bin/env -i PATH=/usr/bin:/usr/local/sbin:/usr/sbin:/sbin /opt/software/singularity-3.0/bin/singularity shell --nv pytorch-18.01-py3.simg
WARNING: underlay of /usr/bin/nvidia-smi required more than 50 (323) bind mounts
Singularity pytorch-18.01-py3.simg:~>
``` 

This is caused by modules loaded and various environment variables that are set on a cluster.

mboisson:
This looks similar to #2060, but obviously, this is not fixed in 3.0.1 since that is what we have installed.

mboisson:
For what it's worth, we have confirmed that the version that we compiled has 
```
func ExecuteSingularity() {
  defaultEnv := "/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin"

  // backup user PATH
  userEnv := strings.Join([]string{os.Getenv("PATH"), defaultEnv}, ":")
  os.Setenv("USER_PATH", userEnv)

  os.Setenv("PATH", defaultEnv)
  if err := SingularityCmd.Execute(); err != nil {
    os.Exit(1)
  }
}
```

mboisson:
@jscook2345, actually, the error message the OP is two separate problems. The `ldconfig` error message is related to the `PATH`. I can workaround it by sanitizing the `PATH` before calling `singularity`. The JSON error is related to the overall size of the environment.

mboisson:
This seems to have been resolved in 3.1.

PR #4057, opened 2019-07-31
Patch ldconfig path

mboisson:

Currently, singularity uses "ldconfig" found in the environment provided by the user. It should be using the system ldconfig instead. In our case, the ldconfig in the user environment is not meant to be used by user-space software. 

- Fixes #
This fixes issue https://github.com/sylabs/singularity/issues/2445  which was never completely fixed. 


jmstover:
In general this looks good to me. We _should_ be cleaning the PATH early in the Singularity execution to be:
`/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin`

And being dependent on system installed binaries except for where we may explicitly specify a path.

Should being the key word here...

jscook2345:
@jmstover @cclerget 

Could / Should we make this part of the `buildcfg` pkg?

Is there ever a time where `ldconfig` should be provided as a run-time option?

jmstover:
> Is there ever a time where `ldconfig` should be provided as a run-time option?

No, the `ldconfig` executed should be the system `ldconfig` so we know that our system library cache looks like.

In theory, if we are resetting our `PATH` prior to this point, we'll be running the `ldconfig` installed on the system wherever that may happen to be ...  in either `/sbin`, or `/usr/sbin` ... we hope.

sashayakovtseva (changes requested):
This change potentially breaks old behavior. 

We have `actionPreRun` in `cmd/internal/cli/actions.go` resetting `PATH` to `/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin`, so each actions using `--nv` should use the right `ldconfig`.

mboisson:
There may very well be something that is supposed to be resetting `PATH`, but it does not work: 
```
$ which singularity
/opt/software/singularity-3.2/bin/singularity
$ singularity --version
singularity version 3.2.1-1.el7
$ which ldconfig
/cvmfs/soft.computecanada.ca/nix/var/nix/profiles/16.09/bin/ldconfig
$ singularity shell --nv pytorch-18.01-py3.simg
WARNING: Unable to capture NVIDIA bind points: could not execute ldconfig: exit status 1
Singularity pytorch-18.01-py3.simg:~> exit
exit
$ PATH=/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin:$PATH singularity shell --nv pytorch-18.01-py3.simg
Singularity pytorch-18.01-py3.simg:~> exit
exit
``` 

So, while it may potentially break old behavior, the old behavior is currently broken and needs fixing one way or another. 

We also confirmed that 3.3.0 also requires the patch.

cclerget:
@mboisson Ok the issue is because we use `USER_PATH` for binary path lookup, could you please instead do before line https://github.com/sylabs/singularity/blob/b511066f6ffe65eb7e444683d60c0d139bc4814c/pkg/util/nvidia/paths.go#L85


```
ldConfig, err := exec.LookPath("ldconfig")
if err != nil {
    sylog.Debugf("ldconfig binary not found in %s, will search in %s", err, os.Getenv("PATH"), envPath)
    ldConfig = "ldconfig"
}
```

and use `ldConfig` for :

```
out, err := exec.Command(ldConfig, "-p").Output()
```

This way we should keep the old behavior and also fix this issue

mboisson:
I made the suggested change. Will be testing soon.

pkg/util/nvidia/paths.go
```diff
@@ -82,14 +82,19 @@ func nvliblist(nvidiaDir string) ([]string, error) {
 // Paths returns list of nvidia libraries and binaries that should
 // be added to mounted into container if it needs NVIDIA GPUs.
 func Paths(nvidiaDir string, envPath string) ([]string, []string, error) {
+	ldConfig, err := exec.LookPath("ldconfig")
+	if err != nil {
+		sylog.Debugf("ldconfig binary not found in %s, will search in %s", err, os.Getenv("PATH"), envPath)
```
jmstover: In this `Debugf` you have two `%s`, but three values. Should probably be changed to something like:

```suggestion
		sylog.Debugf("%v: ldconfig binary not found in %s, will search in %s", err, os.Getenv("PATH"), envPath)
```

jmstover (dismissed):

mboisson:
The error given by travis seems not to be related to this PR, but rather to a timeout from the docker registry.

jscook2345:
@mboisson 

I restarted the jobs after verifying your comment.

mboisson:
For what it's worth, I confirm that this patch fixes the issue on singularity 3.3.0 on our system. It now uses the proper `/sbin/ldconfig`

pkg/util/nvidia/paths.go
```diff
@@ -82,14 +82,19 @@ func nvliblist(nvidiaDir string) ([]string, error) {
 // Paths returns list of nvidia libraries and binaries that should
 // be added to mounted into container if it needs NVIDIA GPUs.
 func Paths(nvidiaDir string, envPath string) ([]string, []string, error) {
+	ldConfig, err := exec.LookPath("ldconfig")
+	if err != nil {
```
sashayakovtseva: I think you should check for ErrNotFound here to be sure that `ldconfig` was not found indeed rather than any other error occurred. 
```
 ee, ok := err.(*exec.Error)
if ok && ee.Err == exec.ErrNotFound {}
```

pkg/util/nvidia/paths.go
```diff
@@ -82,14 +82,19 @@ func nvliblist(nvidiaDir string) ([]string, error) {
...
+		sylog.Debugf("%v: ldconfig binary not found in %s, will search in %s", err, os.Getenv("PATH"), envPath)
```
sashayakovtseva: 1. I don't think it is a good idea to print full PATH value here, simply `not found in PATH` is enough
2. It is common to describe error first and then clarify what was wrong. So this should be `sylog.Debugf("could not find ldconfig: %v", err)`.

sashayakovtseva (changes requested):
Minor suggestions.

sashayakovtseva:
@mboisson Looks good, just add an error check in case it is not `ErrNotFound`

mboisson:
@sashayakovtseva what should it do if the error is not `ErrNotFound` ?

Note that my Go knowledge is very limited, so I am mostly copying the code that you guys post.

sashayakovtseva:
@mboisson I believe simply returning `fmt.Errorf("could not lookup ldconfig: %v", err)` is fine

mboisson:
@sashayakovtseva ok, I think that's what I've done now

sashayakovtseva:
@mboisson Not exactly what I meant. You are not returning error when `ok` is false. If you don't mind, I will update your branch.

mboisson:
Please do :)

sashayakovtseva:
Btw formatting is also wrong and code will not be built.

sashayakovtseva (approved):

mboisson:
I'm not sure why @jmstover's requested changes are not marked as resolved ?

sashayakovtseva:
lgtm, I will dismiss @jmstover's  review since his comments are already addressed and we don't want to keep this PRs forever.

mboisson:
Will this be back-ported into all 3.x ?

sashayakovtseva:
I think @GodloveD can answer that.

jmstover:
> I'm not sure why @jmstover's requested changes are not marked as resolved ?

Sorry, I thought that I had marked it as resolved when I saw the change. :/  Thanks Sasha for just dismissing it.

GodloveD:
Just added it to the milestone.

PR #4069, opened 2019-08-01
Lock ext3 image partitions and prevent them from concurrent writes

### LINKED ISSUE #4048: Overlay image corruption with multiple writable instances on same SIF
## Version of Singularity:

3.3.0-rc4

### Expected behavior

Some sort of diagnostic or safeguard around mounting images multiple times.

### Actual behavior

I put an ext3 overlay filesystem into a `.sif` file and started two writable instances with `shell` and `exec`. I installed some software as root in one while the other one was active, `dmesg` spews warnings about deleted inodes and the filesystem is pretty much busted.

I'm guessing that it might be technically safe if all mounts are read-only.

Reproduction steps approximate, reconstructing from history.

I'm not sure if this is how one is supposed to make an persistent overlay image at all, as it's quite underdocumented but this approach got it appearing to work for single instances.

Build image:
```bash
singularity pull docker://fedora:30
truncate -s 150G data.ext3
mkfs.ext3 data.ext3
singularity sif add --datatype 4 --groupid 1 --partarch 2 --partfs 2 --parttype 4 fedora_30.sif data.ext3
```

Concurrently - launch two instance and install a bunch of software from repos in the superuser one while building software in the other one.
```bash
singularity shell -e -w fedora_30.sif
sudo singularity shell -e -w fedora_30.sif
```

`blkid` on the host sees two loop devices, both appearing to be the image:
```
/dev/loop2: TYPE="squashfs"
/dev/loop3: UUID="0c055153-8bd5-4896-a96d-395289e91883" TYPE="ext3"
/dev/loop4: TYPE="squashfs"
/dev/loop5: UUID="0c055153-8bd5-4896-a96d-395289e91883" TYPE="ext3"
```

cclerget:

Lock EXT3 image partitions to prevent from concurrent writes

- Fixes #4048 

### DIFF
--- a/internal/pkg/runtime/engines/singularity/prepare_linux.go
+++ b/internal/pkg/runtime/engines/singularity/prepare_linux.go
@@ -898,8 +898,7 @@ func (e *EngineOperations) loadImages(starterConfig *starter.Config) error {
 		}
 		// C starter code will position current working directory
 		starterConfig.SetWorkingDirectoryFd(int(img.Fd))
-	}
-	if img.Type == image.SIF {
+	} else if img.Type == image.SIF {
 		// query the ECL module, proceed if an ecl config file is found
 		ecl, err := syecl.LoadConfig(buildcfg.ECL_FILE)
 		if err == nil {
@@ -912,6 +911,16 @@ func (e *EngineOperations) loadImages(starterConfig *starter.Config) error {
 			}
 		}
 	}
+
+	// lock all ext3 partitions if any to prevent concurrent writes
+	for _, part := range img.Partitions {
+		if part.Type == image.EXT3 {
+			if err := img.LockSection(part); err != nil {
+				return fmt.Errorf("error while locking ext3 partition from %s: %s", img.Path, err)
+			}
+		}
+	}
+
 	if err := starterConfig.KeepFileDescriptor(int(img.Fd)); err != nil {
 		return err
 	}
@@ -934,6 +943,16 @@ func (e *EngineOperations) loadImages(starterConfig *starter.Config) error {
 		if err != nil {
 			return fmt.Errorf("failed to open overlay image %s: %s", splitted[0], err)
 		}
+
+		// lock all ext3 partitions if any to prevent concurrent writes
+		for _, part := range img.Partitions {
+			if part.Type == image.EXT3 {
+				if err := img.LockSection(part); err != nil {
+					return fmt.Errorf("error while locking ext3 overlay partition from %s: %s", img.Path, err)
+				}
+			}
+		}
+
 		if err := starterConfig.KeepFileDescriptor(int(img.Fd)); err != nil {
 			return err
 		}
--- a/pkg/image/image.go
+++ b/pkg/image/image.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/sylabs/singularity/internal/pkg/sylog"
 	"github.com/sylabs/singularity/internal/pkg/util/user"
+	"github.com/sylabs/singularity/pkg/util/fs/lock"
 )
 
 const (
@@ -159,6 +160,46 @@ func (i *Image) HasRootFs() bool {
 		}
 	}
 	return false
+}
+
+// LockSection puts a file byte-range lock on a section to prevent
+// from concurrent writes depending if the image is writable or
+// not. If the image is writable, calling this function will place
+// a write lock for the corresponding section preventing further use
+// if the section is used for writing or reading only, if the image is
+// not writable this function place a read lock to prevent section
+// from being written while the section is used in read-only mode.
+func (i *Image) LockSection(section Section) error {
+	fd := int(i.Fd)
+	start := int64(section.Offset)
+	size := int64(section.Size)
+
+	br := lock.NewByteRange(fd, start, size)
+
+	var err error
+
+	if i.Writable {
+		err = br.Lock()
+	} else {
+		err = br.RLock()
+	}
+
+	if err == lock.ErrByteRangeAcquired {
+		if i.Writable {
+			return fmt.Errorf("can't open %s for writing, currently in use by another process", i.Path)
+		}
+		return fmt.Errorf("can't open %s for reading, currently in use for writing by another process", i.Path)
+	} else if err == lock.ErrLockNotSupported {
+		// ENOLCK means that the underlying filesystem doesn't support
+		// lock, so we simply ignore the error in order to allow ext3
+		// images located on the underlying filesystem to run correctly
+		// and advertise user in log
+		sylog.Verbosef("Could not set lock on %s section %q, underlying filesystem seems to not support lock", i.Path, section.Name)
+		sylog.Verbosef("Data corruptions may occur if %s is open for writing by multiple processes", i.Path)
+		return nil
+	}
+
+	return err
 }
 
 // ResolvePath returns a resolved absolute path.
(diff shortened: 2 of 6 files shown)


internal/pkg/runtime/engines/singularity/prepare_linux.go
```diff
@@ -911,6 +910,23 @@ func (e *EngineOperations) loadImages(starterConfig *starter.Config) error {
 				return err
 			}
 		}
+		// lock overlay ext3 partitions if any to prevent concurrent writes
+		if len(img.Partitions) > 1 {
+			for _, part := range img.Partitions[1:] {
+				if part.Type == image.EXT3 {
+					if err := img.LockSection(part); err != nil {
+						return fmt.Errorf("error while loading ext3 overlay partition from %s: %s", img.Path, err)
```
mem: "while locking", no?

internal/pkg/runtime/engines/singularity/prepare_linux.go
```diff
@@ -911,6 +910,23 @@ func (e *EngineOperations) loadImages(starterConfig *starter.Config) error {
...
+					}
+				}
+			}
+		}
+	} else if img.Type == image.EXT3 {
+		// lock ext3 system partition if we opened it for writing
+		// or check if the image is not already opened for writing
```
mem: not sure where the "or" part of this comment is...

Is `LockSection` doing something more than just locking?

internal/pkg/runtime/engines/singularity/prepare_linux.go
```diff
@@ -911,6 +910,23 @@ func (e *EngineOperations) loadImages(starterConfig *starter.Config) error {
...
+		// by another process in case of read-only use
+		if err := img.LockSection(img.Partitions[0]); err != nil {
+			return fmt.Errorf("error while loading ext3 system partition from %s: %s", img.Path, err)
```
mem: "while locking"?

internal/pkg/runtime/engines/singularity/prepare_linux.go
```diff
@@ -934,6 +950,12 @@ func (e *EngineOperations) loadImages(starterConfig *starter.Config) error {
 		if err != nil {
 			return fmt.Errorf("failed to open overlay image %s: %s", splitted[0], err)
 		}
+		// lock overlay ext3 images to prevent concurrent writes
+		if img.Type == image.EXT3 && img.HasRootFs() {
+			if err := img.LockSection(img.Partitions[0]); err != nil {
+				return fmt.Errorf("error while loading ext3 overlay image %s: %s", img.Path, err)
```
mem: "while locking"?

pkg/image/image.go
```diff
@@ -161,6 +162,39 @@ func (i *Image) HasRootFs() bool {
 	return false
 }
 
+// LockSection put a file byte-range lock on a section to prevent
```
mem: "LockSection puts ..."

pkg/image/image.go
```diff
@@ -161,6 +162,39 @@ func (i *Image) HasRootFs() bool {
...
+	ltype := lock.ReadLock
+	if i.Writable {
+		ltype = lock.WriteLock
+	}
+	if err := br.Acquire(ltype); err == lock.ErrByteRangeAcquired {
+		if i.Writable {
+			return fmt.Errorf("can't open %s for writing, currently in use by another process", i.Path)
+		}
+		return fmt.Errorf("can't open %s for reading, currently in use for writing by another process", i.Path)
+	} else if err != nil {
```
mem: Not sure what to do with this...

The function is supposed to acquire a lock, fails to do so and no error is reported. I understand the need to support filesystems where locking is not available, but I don't think dropping all errors is the right way to go about it.

Ideally it should not be necessary to dig into the implementation details to understand the logic. I suspect this handling of errors is tightly related to implementation details.

pkg/image/image.go
```diff
@@ -161,6 +162,39 @@ func (i *Image) HasRootFs() bool {
...
+	fd := int(i.Fd)
+	start := int64(section.Offset)
+	size := int64(section.Size)
+
+	br := lock.NewByteRange(fd, start, size)
+	ltype := lock.ReadLock
+	if i.Writable {
+		ltype = lock.WriteLock
+	}
+	if err := br.Acquire(ltype); err == lock.ErrByteRangeAcquired {
```
mem: as a general comment, looking at the implementation of these functions (Acquire, etc), I think the client code would be easier to read if you had a different API. Instead of having a parameter specifying the type of lock, something more like:

```
func (ByteRange) RLock()
func (ByteRange) Lock()
func (ByteRange) Unlock()
```

If you look at the sync package, you'll see that read vs read-write is handled at the type level: you create a read lock, you create a rw lock.

pkg/util/fs/lock/lock.go
```diff
@@ -1,34 +1,105 @@
...
-	if err := syscall.Flock(fd, syscall.LOCK_UN); err != nil {
+	defer unix.Close(fd)
+	if err := unix.Flock(fd, unix.LOCK_UN); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Type defines the type of a lock
+type Type uint8
```
mem: this type and the corresponding constants are mimicking the underlying flock API. I don't think these are necessary (or at least, I don't think it's necessary to export these; obviously it's necessary with the current implementation).

pkg/util/fs/lock/lock.go
```diff
@@ -1,34 +1,105 @@
...
+	// UnLock corresponds to a released lock
+	UnLock
+)
+
+// ErrByteRangeAcquired correponds to the error returned
+// when a file byte-range is already acquired
+var ErrByteRangeAcquired = errors.New("file byte-range lock is already acquired")
+
+// ByteRange defines a file byte-range lock.
+type ByteRange struct {
```
mem, line 48: `ByteRangeLock` maybe? I don't really like it because it stutters with the package name (lock.ByteRangeLock). The standard library side-stepped this problem by calling the somewhat equivalent package `sync`, so it results in `sync.Mutex`, etc.

Within the package the problem I see with `ByteRange` is that it sounds like it's something else (something representing a byte range, not a lock on a byte range).

pkg/image/image.go
cclerget: Will just ignore `ENOLCK` error and will see if it block users

pkg/image/image.go
cclerget: will do, thanks !

pkg/util/fs/lock/lock.go
cclerget: Well we are in package `lock` and get `lock.ByteRange`, that's explicit enough at least for me and there is many package with this kind of "shortcuted" name

pkg/util/fs/lock/lock.go
```diff
@@ -1,34 +1,85 @@
...
-	defer syscall.Close(fd)
-	if err := syscall.Flock(fd, syscall.LOCK_UN); err != nil {
+	defer unix.Close(fd)
+	if err := unix.Flock(fd, unix.LOCK_UN); err != nil {
 		return err
 	}
 	return nil
 }
+
+// ErrByteRangeAcquired correponds to the error returned
```
sashayakovtseva:
```suggestion
// ErrByteRangeAcquired corresponds to the error returned
```

pkg/util/fs/lock/lock.go
```diff
@@ -1,34 +1,85 @@
...
+// when a file byte-range is already acquired
```
sashayakovtseva:
```suggestion
// when a file byte-range is already acquired.
```

sashayakovtseva (changes requested):
lgtm
as discussed offline a couple things to address:
- ErrLockNotSupported
- lock partition[0] of SIF if it is ext3 (even though it is a rare case)
- F_SETLK64 is equal to F_SETLK in linux, so probably no need to separate const and const_linux

sashayakovtseva (approved):
lgtm

PR #4075, opened 2019-08-01
Ensure descr type for verification

ikaneshiro:

Previously we were only checking for descriptors linked to the system partition and assuming they were signatures, now that we have linked partitions to store other information, that assumption is no longer correct and we need to filter linked descriptor results to only `DataSignature` types.

- Fixes #

### DIFF
--- a/pkg/signing/signing.go
+++ b/pkg/signing/signing.go
@@ -199,6 +199,16 @@ func Sign(cpath string, id uint32, isGroup bool, keyIdx int) error {
 	return nil
 }
 
+func filterSigDescrs(sigs []*sif.Descriptor) []*sif.Descriptor {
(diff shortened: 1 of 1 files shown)


WestleyK (changes requested):
Thanks for the quick fix, but this should probably be fixed in the sif repo -> https://github.com/sylabs/sif/pull/45

dtrudg (approved):
This lgtm. @WestleyK - I think the filtering should either (a) be here, (b) or in a *new* method in the `sif` code that explicitly returns only signatures, or allows the type to return to be filtered there. I don't think the approach in https://github.com/sylabs/sif/pull/45 to only return `DataSignature` from a generically named `GetFromLinkedDescr` is correct.

PR #4089, opened 2019-08-02
Rename engine receivers so that they are consistent

sashayakovtseva:


cclerget (approved):
LGTM

GodloveD:
I'm not going to block this, but in general I prefer more verbose names over single letters just because this is going to make grepping for mentions of `engine` virtually impossible.  But that's just me.  

Here's a cat being chased by ducks with cartoon arms:  

PhonyFancyAdamsstaghornedbeetle-size_restricted

pisarukv (approved):
lol, but agree short names for a close scope is ok in go

PR #4105, opened 2019-08-02
Make cache clean -f the default behavior

### LINKED ISSUE #4014: Singularity cache clean doesn't work
## Version of Singularity:

3.3.0-rc.1.438.gc5e243fc4

### Expected behavior

`singularity cache clean` removes local cache.

### Actual behavior

Cache is not removed.

```
rm -rf $HOME/.singularity/cache
singularity pull -U library://sashayakovtseva/test/image-server
ls -la $HOME/.singularity/cache/library
singularity cache clean
ls -la $HOME/.singularity/cache/library
```

sashayakovtseva:
Closes #4014


gvallee (commented):
I included @WestleyK to the review process since he spent quite a bit of time working on this.
If it is okay to change the flag, the only question I have is that just by looking at the modifications, not the entire code, it is a little unclear to me that `--force` will now be the default behavior. Can you confirm that it is (ultimately we really need an E2E test for this)?
Otherwise, LGTM.

sashayakovtseva:
Yes, force is now the default behavior.

mem (changes requested):
The problem is that we tried to make cache list and cache clean consistent with each other.

When you do:

```
singularity cache list
singularity cache clean
```

you would expect that the output of list is what gets removed by the second command.

Does that make sense?

If yes, the next question is what does get listed by `cache list` by default. After a lot of back and forth, it was determined that there are two answers that make sense:

1. Everything
2. "Containers" (meaning things that correspond to container images)

the distinction between the two is the stuff OCI adds to the cache (intermediate layers, etc). Depending on what the user does most of the time, this difference might represent ~ 50% of the cache contents.

Because of that last point, to me 1. is the most correct choice.

Now, since `cache list` and `cache clean` will work on the same thing, and `cache list` lists everything, that means `cache clean` nukes the entire cache.

That requires some confirmation from the user. That's the `-f` flag that this PR is removing.

So, where do we go from here?

GodloveD:
I would suggest that we do the following:
```
$ singularity cache clean
This will delete everything in your cache (containers from all sources and OCI blobs). 
Hint: You can see exactly what would be deleted by canceling and using the --dryrun option.
Do you want to continue? [N/y]
```
^ yes and no prompt do what you expect
```
$ singularity cache clean --force
```
^ skip the prompt and get down to business.  (Useful for scripting.)
```
$ singularity cache clean --dry-run
```
^ look before you leap

Does this workflow satisfy all requirements?

cmd/internal/cli/cache_clean_linux.go
```diff
@@ -42,20 +41,19 @@ var cacheCleanNameFlag = cmdline.Flag{
 	Usage:        "specify a container cache to clean (will clear all cache with the same name)",
 }
 
-// -f|--force
-var cacheCleanForceFlag = cmdline.Flag{
-	ID:           "cacheCleanForceFlag",
-	Value:        &cacheCleanForce,
+// --dry-run
```
GodloveD:
```suggestion
// -n|--dry-run
```
I don't really know why, but I think that `-n` is a common convention for the `--dry-run` flag in Linux programs.

cmd/internal/cli/cache_clean_linux.go
```diff
@@ -42,20 +41,19 @@ var cacheCleanNameFlag = cmdline.Flag{
...
+var cacheCleanDryFlag = cmdline.Flag{
+	ID:           "cacheCleanDryFlag",
+	Value:        &cacheCleanDry,
 	DefaultValue: false,
-	Name:         "force",
-	ShortHand:    "f",
-	Usage:        "force cleaning the cache (otherwise operate in dry run mode)",
+	Name:         "dry-run",
```
GodloveD:
```suggestion
	Name:         "dry-run",
	ShortHand:    "n",
```

cmd/internal/cli/cache_clean_linux.go
sashayakovtseva: Didn't know that, will add

sashayakovtseva:
As for me, @GodloveD's suggested flow is the best one here and will satisfy @mem's concerns about confirmation. 
Will update PR soon.

cmd/internal/cli/cache_clean_linux.go
```diff
@@ -74,15 +87,34 @@ var CacheCleanCmd = &cobra.Command{
 }
 
 func cacheCleanCmd() error {
```
pisarukv: this method doesn't return any errors

cmd/internal/cli/cache_clean_linux.go
```diff
@@ -74,15 +87,36 @@ var CacheCleanCmd = &cobra.Command{
...
-	// We create a handle to access the current image cache
+	if !cacheCleanForce {
+		ok, err := cacheCleanPrompt()
+		if err != nil {
+			return fmt.Errorf("could not prompt user: %v", err)
+		}
+		if !ok {
+			sylog.Infof("Cache cleanup canceled")
+			return nil
```
pisarukv, line 98: A bit awkward moment. Do we need to return an error if deleting was canceled? I believe that yes, since the operation was not performed. Just to think.

sashayakovtseva: We can introduce `errCleanCancelled` here and move logging to the level up.

pisarukv: I would better create a more generic err for cli operations. Something like `errOperationCancelled` So it can be used in other operations with user input.

pisarukv (approved):

GodloveD:
With this (and any other changes to the CLI), can you please make sure that the E2E tests are updated to cover any new flags and they are passing before merging the PR?  Thank you.

gvallee:
@sashayakovtseva about the e2e test, we can work on it together if you want, just let me know.

gvallee (approved):
lgtm

e2e/cache/cache.go
```diff
@@ -0,0 +1,95 @@
...
+	}{
+		{
+			name:   "clean cache normal",
+			expect: "Do you want to continue? [N/y]",
+			send:   "y",
+			exit:   0,
+		},
+	}
+
+	imagePath := c.env.TestDir + "/busybox.sif"
```
gvallee:
```suggestion
	imagePath := filepath.Join(c.env.TestDir, "busybox.sif")
```

gvallee (changes requested):
Another little suggestion.

mem (approved):

gvallee (approved):

PR #4107, opened 2019-08-02
Remove the dependency on homebrew to install go

### LINKED ISSUE #4106: remove homebrew as dependency from macos build
`homebrew` update is taking quite awhile before we build, so remove it as a dependency for now and install `go` from binary package.

we only really build and check for macos currently so `homebrew` is not a hard requirement atm.

jscook2345:
homebrew runs an update before installing any packages, which can
take awhile. Since we are only building and linting on macos (and
not doing any functional testing), remove homebrew as a dependency
to reduce build times.

In the future, we may add homebrew back when more dependencies
besides go are required.

Fixes: #4106

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,7 +83,7 @@ jobs:
           name: Update Go to 1.12.7
           working_directory: /tmp
           command: |-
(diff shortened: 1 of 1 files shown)


gvallee (approved):
lgtm

PR #4111, opened 2019-08-05
Update scs-library client

### LINKED ISSUE #3755: Unused parameters in LibraryPull function
See internal/app/singularity/pull.go:40

## Version of Singularity:

3.3.0-rc.1

### LINKED ISSUE #4014: Singularity cache clean doesn't work

sashayakovtseva:
Also minor code update: do not export vars if not needed, add periods in comments.

Closes #3755

### DIFF
--- a/internal/app/singularity/pull.go
+++ b/internal/app/singularity/pull.go
@@ -36,7 +36,7 @@ var (
 // LibraryPull will download the image specified by file from the library specified by libraryURI.
 // After downloading, the image will be checked for a valid signature and removed if it does not contain one,
 // unless specified not to by the unauthenticated bool
-func LibraryPull(imgCache *cache.Handle, name, ref, transport, fullURI, libraryURI, keyServerURL, authToken string, force, unauthenticated, noCache bool) error {
+func LibraryPull(imgCache *cache.Handle, name, fullURI, libraryURI, keyServerURL, authToken string, force, unauthenticated, noCache bool) error {
 	if !force {
 		if _, err := os.Stat(name); err == nil {
 			return fmt.Errorf("image file already exists: %q - will not overwrite", name)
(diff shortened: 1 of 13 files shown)


internal/app/singularity/pull.go
```diff
@@ -54,25 +54,25 @@ func LibraryPull(imgCache *cache.Handle, name, ref, transport, fullURI, libraryU
...
 
-	imageName := uri.GetName("library://" + imageRef)
-
 	// check if image exists in library
-	libraryImage, existOk, err := libraryClient.GetImage(context.TODO(), imageRef)
-	if err != nil {
-		return fmt.Errorf("while getting image info: %v", err)
-	}
-	if !existOk {
+	libraryImage, err := libraryClient.GetImage(context.Background(), imageRef)
```
mem: hmm...

context.TODO means there's pending work regarding adding proper context handling.

context.Background means this is the topmost context.

They are otherwise identical, the only difference is "documentation".

The key bit in the documentation is this, regarding context.TODO:

> (because the surrounding function has not yet been extended to accept a Context parameter).

What we need to do here is add a signal handler, and in case we receive SIGINT (and other similar signals) we call a cancellation function, so that the download is stopped and any temporary files are removed.

There's a signal handler buried somewhere in this code. That ought to be moved to the main function, where we can create a context.Background, and call the cancellation function as appropriate. This requires extending the CLI run handlers so that they pass this context down. There are other parts of the code that will benefit from this.

mem (changes requested):
LGTM, please revert back to context.TODO to keep track of where this needs to be changed.

mem (approved):
thanks!

pisarukv (approved):

PR #4120, opened 2019-08-06
Improved documentation for 'pull'

WestleyK:
## Description of the Pull Request (PR):

This PR imporves/adds documentation for the:
 - exit code from 'pull' when pulling a unsigned container
 - pulling from http, https protocol

### This fixes or addresses the following GitHub issues:

- ~Fixes #3955~
- Fixes #4119

<br>

PR #4124, opened 2019-08-06
Check if image has a signature before pull

### LINKED ISSUE #4123: Should we even pull images without a signature
Right now we are pulling images even they don't have signatures. In the result of such pulling singularity will exit with an error code and pulled image. It has two problems. 

1) **It's totally misleading.** When I see an app with not 0 code I'm not expecting it to do some work.  So in such case I don't expect image to be pulled. When I'm using singularity pull command manually, maybe, it's ok. But it's a kind of pain to automate it. Inside my scripts I need to check for a specific, not documented error code and after initiate some cleanup.

2) **It's takes a lot of time.** When I don't want to use unsigned images, why I have to download it first to check that it even doesn't have a signature at all? It's the part which can be checked before pulling the image.

I propose to fail fast. If an image doesn't have a signature let's exit with an error without actually pulling it.

pisarukv:
closes https://github.com/sylabs/singularity/issues/4123

Pull flag --allow-unsigned, -U now is marked as Deprecated.


sashayakovtseva (approved):
LGTM

jscook2345:
Since this will likely break a lot of scripts and workflows for users I'd like to discuss this and our defaults a bit more in a meeting.

gvallee:
Please do not merge, this is contradicting #4120

gvallee (changes requested):
We need to clarify what returned code is returned when. See #4120

WestleyK (changes requested):
Thanks @cali4888 :+1: can you just document the exit code in `pull --help` like in https://github.com/sylabs/singularity/pull/4120? Otherwise looks good!

GodloveD (changes requested):
I think there are still some problems with this.
1. **We should have consistent behavior across all commands.** 
`shell`, `run`, `exec`, etc. are all capable of doing a pull behind the scenes.  That pull should behave the same as the explicit pull command.
2. **Failing changes behavior and will break scripts.**
And what is the point really?  In the old version we issued a warning and returned an error code (which was weird).  But I don't think the answer is really to return an error.  It's just to let the user know about signed containers and that they should be using them.  I think we should just remove the error code and continue warning.

I suggest the following.  Keep the warning and change the exit code to zero (since there is no _error_).  Add the warning to every command that is capable of building an image.  Retain the `--allow-unsigned` flag and add it to all of the commands that perform an implicit pull so the user can suppress the warning.

pisarukv:
Agree with you @GodloveD But I'm not sure about the flag. Do we need a separate flag to suppress pull warnings? It seems like it can be done with log levels setup.

GodloveD:
Agree that we can ditch the flag.

Issue #4126, opened 2019-08-06
Clarify the semantics of e2e profiles


gvallee:
## Version of Singularity:

master.

### Expected behavior

The goal of a profile is to write one test and have it executed *implicitly* under different contexts, called *profiles*. In other words, a profile is a e2e execution context with which is associated one or more options that will be added to the Singularity command line.

For example, let's assume we have a test to build an image. The classic command line is `singularity build my.sif URI`. With the concept of profiles, a E2E developer will only have to encode the following (not meant to be code that can be executed, this is just an illustration):
```
  c.env.RunSingularity(
         t,
         e2e.AsSubtest(tt.desc),
         e2e.WithPrivileges(false),
         e2e.WithCommand("build"),
         e2e.WithArgs(strings.Split(argv, " ")...),
         e2e.WithEnv(append(os.Environ(), c.env.KeyringDir)),
         e2e.ExpectExit(tt.expectedExitCode))
```

Adding profiles would mean that we only need to add an option to the `RunSingularity` function to define them and the E2E framework will *automatically* run multiple Singularity commands with all the appropriate flags. For instance, it is possible to run a test with fakeroot by simply adding an option such as:

```
   // Add the profile to each test case
  tests := []struct {
     profile e2e.Profile
     ...
  }{
      { profile: e2e.RootProfile, ... },
      { profile: e2e.FakerootProfile, ...},
  }
```
and 
```
c.env.RunSingularity(
         t,
         e2e.AsSubtest(tt.desc),
         e2e.WithProfile(tt.profile),
         e2e.WithCommand("build"),
         e2e.WithArgs(strings.Split(argv, " ")...),
         e2e.WithEnv(append(os.Environ(), c.env.KeyringDir)),
         e2e.ExpectExit(tt.expectedExitCode))
```

### Actual behavior

@cclerget created a PR as a possible implementation: #3997 

The goals of the PR are to:
- minimize the amount of modifications necessary to implement profiles,
- assume that developers will always specify at least one applicable profile in their tests.
As a note: it was mentioned a few times that the PR increases the execution time in a quite drastic manner, only because it runs more tests, not because on inherent problems with the implementation.

### Steps to reproduce behavior

N/A.

gvallee:
[...]

PR #4131, opened 2019-08-06
Add "Profile" as steppingstone for testing --fakeroot

mem:
As part of testing the --fakeroot functionality, sometimes it's
necessary to run the same singularity command under multiple
"personalities" or "profiles". For example "singularity build" can be
executed with privilege or as "singularity build --fakeroot", and they
should both succeed. It's also expected that the command fails if run as
a regular user.

As part of end-to-end testing, all these personalities need to be
tested, and the "Profile" concept introduced here helps with that. The
RunSingularity driver needs to be extended to support this.

Most of this comes from Cedric; this is mostly a reorganization of that
code, with some documentation sprinkled into it. A lot of this is me
guessing and filling in gaps from #3997.


gvallee:
Related to #4126

e2e/internal/e2e/priv_linux.go
```diff
@@ -59,3 +59,13 @@ func Privileged(f func(*testing.T)) func(*testing.T) {
 		f(t)
 	}
 }
+
+// OrigUID returns the UID of the user running the test suite
```
sashayakovtseva: Period.

e2e/internal/e2e/priv_linux.go
```diff
@@ -59,3 +59,13 @@ func Privileged(f func(*testing.T)) func(*testing.T) {
...
+func OrigUID() int {
+	return origUID
+}
+
+// OrigGID returns the GID of the user running the test suite
```
sashayakovtseva: Period.

e2e/internal/e2e/profile/profile.go
```diff
@@ -0,0 +1,230 @@
...
+func (Fakeroot) Requirements(t *testing.T) {
+	require.UserNamespace(t)
+
+	// check that current user has valid mappings in /etc/subuid
+	if _, err := fakeroot.GetIDRange(fakeroot.SubUIDFile, uint32(e2e.OrigUID())); err != nil {
+		t.Fatalf("fakeroot configuration error: %s", err)
+	}
+
+	// check that current user has valid mappings in /etc/subgid
+	if _, err := fakeroot.GetIDRange(fakeroot.SubGIDFile, uint32(e2e.OrigUID())); err != nil {
```
sashayakovtseva: Shouldn't you use `e2e.OrigGID()` here?

e2e/internal/e2e/profile/profile.go
```diff
@@ -0,0 +1,230 @@
...
+		if profile == p {
+			return true
+		}
+	}
+	return false
+}
+
+// buildArgs builds the necessary arguments for singularity command cmd
+// from a list of commands that support the specified flag.
+func buildArgs(cmd []string, commands []string, flag string) []string {
```
sashayakovtseva: Overall this looks okay, but I think what you are trying to do is to check whether one slice contains at least one element from the other.
Maybe this func should be changed to smth like `Contains(s, sub []string)`? and then in `Args()` you will do e.g :

```
if Contains() {
return []string{"--fakeroot"}
}
```

e2e/internal/e2e/profile/profile.go
```diff
@@ -0,0 +1,230 @@
...
+
+func (p UserNamespace) In(profiles ...Profile) bool {
+	return in(p, profiles)
+}
+
+func (UserNamespace) String() string {
+	return "UserNamespace"
+}
+
+// RootUserNamespace implemets Profile describing a privileged user that
```
sashayakovtseva:
```suggestion
// RootUserNamespace implements Profile describing a privileged user that
```

e2e/internal/e2e/profile/profile.go
```diff
@@ -0,0 +1,230 @@
...
+// operates in a separate user namespace.
+type RootUserNamespace struct {
```
sashayakovtseva, line 204: Honestly, I don't really like this double embedding. It took some time to determine why `RootUserNamespace .Privileged()` will refer `Root.Privileged()`. 
This is diamond embedding. I understand, that this may be convenient, but really hard to understand at the first glance.

sashayakovtseva (changes requested):
A couple suggestions I want to discuss.

jscook2345 (dismissed):
Approve pending discussion of @sashayakovtseva notes. Thanks!

e2e/internal/e2e/profile/profile.go
mem: I had the exact same question! I'm glad I'm not alone :-)

The file that this is reading, the one specified by the first argument, is indexed by user name, not group name. I remember my original feedback when this function was added was that it should be two separate functions, not one function doing two different things, but some implementation details make it so that having a single function is more convenient.

I guess I can add a comment clarifying. Thanks.

jscook2345: I see now what you mean after your description. Thanks.

e2e/internal/e2e/profile/profile.go
mem: I think I see what you mean.

This function started off as something different, so I loss some perspective in the process.

This is taking something that already exists as a slice (cmd), which will in most cases contain a single element, but in some cases it will contain two, and maybe more.

Then it takes the list of commands, commands, and looks for cmd in that list. Since cmd is a slice and commands is a slice (not a slice of slices), what the code does is concatenate the elements of cmd and looks for _that_ in the commands slice. If it finds it, it then returns the flag argument a a slice.

All of this is super weird. I'll try the suggested change.

e2e/internal/e2e/profile/profile.go
mem: Yeah, I had a different version of this with a single embedded struct. Then I switched to embedding the other. And then I ended up embedding both because that's what RootUserNamespace is: one part UserNamespace, one part Root, with Root defining the user and UserNamespace defining flags and requirements. The "purely imperative" version of this (for lack of a better way to describe it) is worse, because it has different switch cases depending on the function you are looking at.

The subtlety here is that these structs do not have overlapping methods. If both structs where directly defining the same method, say Privileged, you wouldn't be able to call RootUserNamespace.Privileged() without specifying which one you want (or without implementing Privileged directly for RootUserNamespace). That's Go's solution to the diamond problem: do not allow the ambiguity at the call location.

So, while I agree with you, I'm not sure what to do about it...

sashayakovtseva (approved):

gvallee:
It would have been great to actually discuss this before merging. I have no idea how to use this, there is no example, no explanation, issue #4126 has *not* been updated and what @cclerget did is now totally irrelevant. @jscook2345 @GodloveD What is a path forward?

PR #4136, opened 2019-08-07
[E2E] create a clean temporary image cache by default to each test

### LINKED ISSUE #3987: Make e2e image cache management thread safe
## Version of Singularity:

All.

### Expected behavior

The image cache management in the e2e framework should use the new API to be thread-safe.

### Actual behavior

The current implementation heavily relies on the environment variable which is not thread-safe.

N/A.

gvallee:

This set of modifications ensures that each e2e test will have its own image clean cache by default. This fixes potential problems when running concurrent tests as well as giving us a fine grain control over the context of the test execution since we ensure that the cache will be clean by default.

- Fixes #3987 

### DIFF
--- a/e2e/internal/e2e/fileutil.go
+++ b/e2e/internal/e2e/fileutil.go
@@ -6,20 +6,11 @@
 package e2e
 
 import (
-	"fmt"
 	"io/ioutil"
-	"os"
+	"testing"
 
 	"github.com/pkg/errors"
-	"github.com/sylabs/singularity/internal/pkg/client/cache"
 	"github.com/sylabs/singularity/internal/pkg/util/fs"
-)
-
-var (
-	// cacheDirPriv is the directory the cachedir gets set to when running privileged.
-	cacheDirPriv = ""
-	// cacheDirUnpriv is the directory the cachedir gets set to when running unprivileged.
-	cacheDirUnpriv = ""
 )
 
 // WriteTempFile creates and populates a temporary file in the specified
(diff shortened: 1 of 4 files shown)


jscook2345:
How would one test the usage of the `SINGULARITY_CACHE_DIR` environment variable here?

gvallee:
@jscook2345 By default, setting manually the environment variable should be avoided by any mean since it would be done for all threads and potentially sub-processes. This PR implicitly and correctly sets the environment variable for processes that we run a single e2e test.

To write a test exercising the `SINGULARITY_CACHE_DIR` environment variable, there is no good solution at the moment; the e2e framework is not designed to easily do this. However, if a e2e developer wants to manually create a test, it would be required to specify it through `singularityCmd`, the structure used to track all test-specific details. I did not expose an API to ease this task since atm we do not have a single test that requires sharing of an image cache (and would be surprised if we ever need it outside of optimizing image downloads) and exercising `SINGULARITY_CACHE_DIR`. Practically, it means that the test would do something like:
```
c.env.RunSingularity(
  t,
  e2e.Singularity_cache_dir("/a/dir"),
  ...
)
```
But again this PR does not include something like `e2e.Singularity_cache_dir()`, that would be future work (same thing for `SINGULARITY_SYPGPDIR`).

Just to make sure that we are all on the same page: the idea here is that *by default*, e2e creates a clean image cache for every single test and it is all we need right now. So there is no need by default to set the `SINGULARITY_CACHE_DIR` environment variable. It is *not* meant to facilitate the test of the `SINGULARITY_CACHE_DIR` environment variable per-say.

gvallee:
This PR fixes the error that occurs during CI testing for PR #4105

jscook2345 (approved):

PR #4179, opened 2019-08-12
Refactor pulling from cloud library

### LINKED ISSUE #4115: LibraryPull should not accept 9 parameters
That is just way toooooo much for a single function....

sashayakovtseva:
After brainstorming session with @cali4888 we decided that it might be better to consider having a separate type for interactions with remote image registries. This PR introduces Library type for pulling images.

If this approach will be approved corresponding types for the rest of registries will be introduced as well as for push action.

Closes #4115


dtrudg:
@sashayakovtseva - could I ask that myself (and maybe @EmmEff) are pinged on this being ready for review / merge in regard to there being ongoing work in #3952 which is affected by changes to the `pull` code. Thanks.

dtrudg:
Hi @sashayakovtseva. Certainly the function sigs and other code around library pull benefits from a cleanup.

Are you intending to also refactor the calls in the other portions of code which do library pulls:

- internal/pkg/build/remotebuilder/remotebuilder.go
- internal/pkg/build/sources/conveyorPacker_library.go
- cmd/internal/cli/actions.go

I wonder if @GodloveD can indicate here also whether this cleanup is a 3.4 target, and if #3952 is? These will conflict quite a bit.

sashayakovtseva:
@dctrud Initially I was not going to update other packages other than CLI related, but it looks reasonable to update those as well. I only wonder if a new issue for each package should be created.

internal/app/singularity/registry.go
```diff
@@ -0,0 +1,146 @@
...
+
+	client *scs.Client
+	cache  *cache.Handle
+}
+
+// NewLibrary initializes and returns new Library ready to  be used.
+func NewLibrary(client *scs.Client, cache *cache.Handle, keystoreURI string) *Library {
+	return &Library{
+		keystoreURI: keystoreURI,
+		client:      client,
```
mem: this one here can be an interface if the access to the AuthToken field is replaced with a function call. All the other fields used here are functions (GetImage, DownloadImage).

the next one (cache) can be an interface without any change (LibraryImage, LibraryImageExists).

You only need to define the interfaces in this package:

```
type scsClient interface {
    GetAuthToken(...)
    GetImage(...)
    DownloadImage(...)
}
```

and declare the fields of Library to be of that type.

This would facilitate testing.

internal/app/singularity/registry.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/sylabs/singularity/internal/pkg/sylog"
+	"github.com/sylabs/singularity/internal/pkg/util/uri"
+	"github.com/sylabs/singularity/pkg/signing"
+)
+
+var errNotInCache = fmt.Errorf("image was not found in cache")
+
+// Registry represents image registries and common operations
+// performed, e.g pull and push.
+type Registry interface {
```
mem: where is this interface used?

internal/app/singularity/registry.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		return fmt.Errorf("unable to check if %s exists: %v", name, err)
+	}
+	if !exists {
+		return errNotInCache
+	}
+
+	// Perms are 777 *prior* to umask in order to allow image to be
+	// executed with its leading shebang like a script
+	dstFile, err := os.OpenFile(to, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, 0777)
```
mem: In order to deal gracefully with various errors that might happen here, this should be creating a temporary file in the destination directory, and set things up to remove that temporary file when this function exits.

If the whole operation succeeds, the temporary file should be renamed to the final name (`to` here). In that way, if something fails, the temporary file gets removed. If nothing fails, you end up with the correct contents in the destination file.

In other words, the defer'ed function should be something like:

```
func () {
    if err := dstFile.Close(); err == nil {
        os.Rename(tmpname, to)
    }
    os.Remove(tmpname)
}
```

Close might result in error and Rename might result in error, but no matter what, the file gets removed.

By defer'ing this, the error on line 137 does not result in a truncated file, for example. Same with the error on line 143.

mem (changes requested):

internal/app/singularity/registry.go
sashayakovtseva: Currently nowhere, but planned to be used in `cmd/internal/cli/pull.go` later as all of registries are implemented. Do you think this should be removed?

sashayakovtseva: Removed until we have all registries updated.

internal/app/singularity/registry.go
sashayakovtseva: That will require creation some kind of wrapper for `*scs.Client` b/c `DownloadImage` is a function in Singularity repo, which accepts `*scs.Client`. 
Also code readability will suffer, so I'd prefer to keep e2e tests and bare `*scs.Client` rather than introducing this new interface.

sashayakovtseva:
@dctrud updated PR according to @mem's comments, rfr

internal/app/singularity/registry.go
```diff
@@ -0,0 +1,127 @@
...
+// Library is a Registry implementation for Sylabs Cloud Library.
+type Library struct {
+	keystoreURI string
+
+	client *scs.Client
+	cache  *cache.Handle
+}
+
+// NewLibrary initializes and returns new Library ready to  be used.
+func NewLibrary(client *scs.Client, cache *cache.Handle, keystoreURI string) *Library {
```
ikaneshiro: Would it be better to have This accept a `client.Config{}` and call `client.NewClient()` here so that we don't need two steps every time we want a new library object?

sashayakovtseva: Well, on the one hand yes, that will save lines.
On the other hand ideally an object accepts dependencies rather than creates them itself, that was the initial motivation.

But I see that `NewClient` simply records fields from config, so it may be reasonable to update.
Unless we may want to bypass `NewClient` in tests (that are not in this PR, but will be in place soon).

dtrudg: I'm kinda in agreement with @ikaneshiro here for the reason that both...

- a `Library` is useless without a client to access it
- a `Client` is not to be used outside of the context of a Library

...seem true?

Then the internal state of a `Library` holding a `Client` is preferable to having to write it out longhand every time?

dtrudg (commented):
I like these changes - the consistent use of `from` and `to` makes the flow a lot nicer to read, and when it spreads out later to the other places we grab things from the library it'll help there also.

The only reason I am not putting a :heavy_check_mark: right now is that I think merging this will require more than tiny changes to outstanding PRs around caching that @gvallee has in the queue, and the library archtags support. I don't necessarily judge those as more urgent - I look to @GodloveD @jscook2345 etc. to consider order of integrating things there.

(so consider this a wimpy implied :heavy_check_mark:)

gvallee:
I think there is plenty to do and this PR is clearly going in the right direction. @dctrud @sashayakovtseva and anyone interested, maybe we should have a brief live discussion to figure out the best way to move forward? I think we are pretty much on the same page but it would be nice to see if we could have relatively small PRs. What do you think?

sashayakovtseva:
@dctrud @gvallee Agree, let's discuss on core meeting on Monday.

internal/app/singularity/registry.go
sashayakovtseva: @dctrud You convinced me, arguments seem true.

pisarukv (approved):

internal/app/singularity/registry.go
mem: I probably did not explain myself correctly.

It's enough to change the code like this:

```
diff --git a/internal/pkg/library/library.go b/internal/pkg/library/library.go
index 5ecde1b65..fa03d2fa8 100644
--- a/internal/pkg/library/library.go
+++ b/internal/pkg/library/library.go
@@ -29,8 +29,12 @@ func NormalizeLibraryRef(libraryRef string) string {
        return ir
 }

+type ImageDownloader interface {
+       DownloadImage(ctx context.Context, w io.Writer, path, tag string, callback func(int64, io.Reader, io.Writer) error) error
+}
+
 // DownloadImage is a helper function to wrap library image download operation
-func DownloadImage(ctx context.Context, c *client.Client, imagePath, libraryRef string, callback progressCallback) error {
+func DownloadImage(ctx context.Context, c ImageDownloader, imagePath, libraryRef string, callback progressCallback) error {
        // reassemble "stripped" library ref for scs-library-client
        validLibraryRef := "library:///" + libraryRef

@@ -66,12 +70,16 @@ func DownloadImage(ctx context.Context, c *client.Client, imagePath, libraryRef

 // DownloadImageNoProgress downloads an image from the library without
 // displaying a progress bar while doing so
-func DownloadImageNoProgress(ctx context.Context, c *client.Client, imagePath, libraryRef string) error {
+func DownloadImageNoProgress(ctx context.Context, c ImageDownloader, imagePath, libraryRef string) error {
        return DownloadImage(ctx, c, imagePath, libraryRef, nil)
 }

+type ImageSearcher interface {
+       Search(ctx context.Context, args map[string]string) (*client.SearchResults, error)
+}
+
 // SearchLibrary searches the library and outputs results to stdout
-func SearchLibrary(ctx context.Context, c *client.Client, value string) error {
+func SearchLibrary(ctx context.Context, c ImageSearcher, value string) error {
        if len(value) < 3 {
                return fmt.Errorf("bad query '%s'. You must search for at least 3 characters", value)
        }
```

you don't need other changes (and just for the record, you don't need to export the interfaces either). The rest of the code doesn't need to care about this, it can continue using a client.

(obviously this refers to the existing code, not this PR)

The point of this change is that you can write a test for these functions without requiring an actual client that will hit the network. We already have enough problems with "unit" tests hitting the network and this is a simple enough change that would help with that.

mem: And yes, I provided the example in the wrong package, sorry :-\

sashayakovtseva: Looks like a separate issue/PR, do you mind creating one?

gvallee (approved):
LGTM

internal/app/singularity/registry.go
```diff
@@ -0,0 +1,135 @@
...
+type Library struct {
+	keystoreURI string
+	arch        string
+
+	client *scs.Client
+	cache  *cache.Handle
+}
+
+// NewLibrary initializes and returns new Library ready to  be used.
+func NewLibrary(scsConfig *scs.Config, cache *cache.Handle, keystoreURI, arch string) (*Library, error) {
```
dtrudg: I'm not in favor of `arch` being associated with the `Library` struct I'm afraid. When we use `arch` currently we are asking for a specific architecture on a *single interaction* with the Library. The Library itself supports all architectures - so conceptually we could interact with it multiple times, with each request asking for a different architecture.

I would suggest that the architecture should be specified on the `Pull` request. In this way any currently unimplemented functionality that might want to handle multiple architectures in one invocation is simpler.

dtrudg (changes requested):

internal/app/singularity/registry.go
EmmEff: See @dctrud I wasn't the only one thinking this way :)

After your explanation, I do agree that library and arch are two disjoint parameters to a pull request.

sashayakovtseva: Yes, that was my concern as well. My points are:
1. I am not sure other registries will need that, and `Pull` is likely to become and interface method. Thus I didn't want to force other implementations to have/respect arch parameter.
2. A `Library` object is used to perform a single pull. If it was ever reused - no doubt that should have been a `Pull` parameter.

sashayakovtseva: If you suspect it will be useful for other registries, then yes, it is reasonable to make it a `Pull` parameter.

EmmEff: If the pull request API is going to be an interface, would it make sense to create an opaque "filter spec" (map) argument where the caller can specify arbitrary key-value pairs for filtering requests?

sashayakovtseva: Not a fan of that approach. 
Furthermore, it is really straightforward here, we don't have complex filters (only arch now), so I suggest not to complicate things before it is necessary.

sashayakovtseva: And it is going to be an interface to enable unit testing of internal CLI package, not a generic purpose interface we are talking about.

dtrudg: > If you suspect it will be useful for other registries, then yes, it is reasonable to make it a Pull parameter.

Yes - for docker hub etc. where you can have defined archs. a straight pull into a SIF for another architecture may be useful... especially thinking about deploying to air-gapped / bandwidth constrained systems....

*but* we are only concerned about library right now.

sashayakovtseva: Well, currently Docker is smart enough to pull correct arch, unless we want to support corner case where we pull incompatible arch.

sashayakovtseva: I am fine with moving arch into `Pull` parameters. Just want to figure out whether that is indeed better :)

dtrudg: > Well, currently Docker is smart enough to pull correct arch, unless we want to support corner case where we pull incompatible arch.

Exactly this is what we are doing in the library. Behind the scenes the OCI code we use for docker pulls is handling passing the arch up on the pull request for docker. You can request a different arch (if we were to expose that). For the library case here we are trying to cater to people who want to be able to pull an `arm` image on their `amd64` system, and then copy it across to their `arm` system. If you consider remote build possibly supporting multiple architectures too then we clearly have to be able to handle interactions which are 'incompatible' with the host arch.

_Edit_ - possible use case.... images of all architectures are pulled to a secure host where they are signed. Only this one machine has the signing keys... needs to be able to operate (at the metadata level) on images for all architectures.

sashayakovtseva: okay, will move arch to pull function parameters.
Either way we can change it later as we proceed with this approach with other registries.

dtrudg (approved):
This looks good to go for me. Tested against staging, production, development library servers pulling / pushing default and specified architectures.

mem (approved):

PR #4190, opened 2019-08-13
Fix a problem with disabling the image cache

### LINKED ISSUE #4189: SINGULARITY_DISABLE_CACHE not entirely behaving as expected
## Version of Singularity:

master.

### Expected behavior

Users can use the `SINGULARITY_DISABLE_CACHE` environment to disable the image caching mechanism. When using the environment variable, no file related to the cache should be created. 

### Actual behavior

PR #4188 adds an e2e test to check the correct of the behavior of the environment variable. While the image is *not* included into the cache, the cache itself is created. This creates problems with environment where users have non-writable home directories and do not wish to use the cache. Basically the need here is to disable the cache, leave it to the default location and nothing should fail. The current behavior will make it to fail because it will try to create the directory structure.

Test added by PR #4188 .

gvallee:

The PR fixes a problem when `SINGULARITY_DISABLE_CACHE` is set. It makes sure that when set, no directory is created in the base directory. So far, the sub-directories of the cache were created (e.g., `cache/library`), even if not used.

The approach implemented in this PR assumes:
- If `SINGULARITY_DISABLE_CACHE` is set `cache.NewHandle()` returns a cache handle that represents a disabled cache. This allows us to easily support disabling the cache without raising an issue and deeply modifying the code.
- The unit tests can only be executed if `SINGULARITY_DISABLE_CACHE` is not set, otherwise the tests are skipped (we cannot really test the image cache capabilities if the environment specifies that no cache should be used).
- We add an internal state to the cache handle (`disabled`) to be able to track if the handle represents a disabled cache. At the moment, that value cannot be changed after the creation of the handle.

- Fixes #4189 

### DIFF
--- a/internal/pkg/client/cache/cache_test.go
+++ b/internal/pkg/client/cache/cache_test.go
@@ -5,9 +5,33 @@
 
 package cache
 
+import (
+	"os"
+	"strconv"
+	"testing"
+)
+
 const (
 	validSHASum   = ""
 	invalidSHASum = "not a SHA sum"
 	validPath     = ""
 	invalidPath   = "not an image"
 )
+
+// checkIfCacheDisabled is used throughout the unit tests to check whether the
+// SINGULARITY_DISABLE_CACHE environment variable is set. If it is set, it will
+// skip the current tests since exercising the cache when caching is disabled is
+// not supported.
+func chechIfCacheDisabled(t *testing.T) {
+	envValue := os.Getenv(DisableEnv)
+	if envValue == "" {
+		envValue = "0"
+	}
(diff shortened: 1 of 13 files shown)


ikaneshiro (changes requested):
Lets change the env value check to use `strconv.ParseBool()` instead of only comparing to "1" so that we can have identical behavior with the cli parsing. Other than that this LGTM!

ikaneshiro (approved):

WestleyK (commented):
Looks good to me :+1: _sorry I'm a little late..._

PR #4191, opened 2019-08-13
Improve the error message if cryptsetup version is unsupported

schebro:
In current code, if the cryptsetup version present on the host is not supported, we simply display "available cryptsetup is not supported".

This change changes the error displayed in the following format:

Min required cryptsetup version: 2.0.0, Found version 2.0.0

- Fixes #

### DIFF
--- a/pkg/util/crypt/crypt_dev.go
+++ b/pkg/util/crypt/crypt_dev.go
@@ -77,30 +77,34 @@ func (crypt *Device) CloseCryptDevice(path string) error {
 	return nil
 }
 
-func checkCryptsetupVersion(cryptsetup string) error {
+func checkCryptsetupVersion(cryptsetup string) (string, error) {
 	if cryptsetup == "" {
-		return fmt.Errorf("binary path not defined")
+		return "", fmt.Errorf("binary path not defined")
 	}
 
 	cmd := exec.Command(cryptsetup, "--version")
 	out, err := cmd.CombinedOutput()
 	if err != nil {
-		return fmt.Errorf("failed to run cryptsetup --version: %s", err)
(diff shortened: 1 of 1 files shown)


gvallee (commented):
So I guess the plan is not anymore to have something like
```
err := crypt.EncryptFilesystem()
if err == crypt.ErrUnsupportedCryptsetupVersion {
```
because based on this change, the unit test does not make sense anymore as it is (reason why it failed). Is the intent here to check the version returned and use that we identify errors where the version does not match? If so, how the package calling `EncryptFilesystem()` can cleanly catch that type of error?

schebro:
> So I guess the plan is not anymore to have something like
> 
> ```
> err := crypt.EncryptFilesystem()
> if err == crypt.ErrUnsupportedCryptsetupVersion {
> ```
> 
> because based on this change, the unit test does not make sense anymore as it is (reason why it failed). Is the intent here to check the version returned and use that we identify errors where the version does not match? If so, how the package calling `EncryptFilesystem()` can cleanly catch that type of error?

One thing we can do in that case is to still return ErrUnsupportedCryptsetupVersion but simply print the current and minimum version numbers. Does that work?

tcooper:
Just built and am testing 3.4.0-rc.1 on CentOS 6. I would add that it would be great if the cryptsetup test came *early*, ideally before the image build and/or removal of an existing image with same filename has happened.

dtrudg:
This PR has gone stale and is on a branch on our main `sylabs/singularity` repo.

I'm closing, having captured the issue as #4621 and I grabbed this branch into https://github.com/dctrud/singularity/tree/schebro_cryptsetup also

PR #4197, opened 2019-08-14
Added back missing code from #3789

### LINKED ISSUE #4187: An image with mismatched checksum should be removed
Currently if an image will be pulled during man-in-the-middle attack we won't remove it, but simply display an error. I wonder whether we should remove that potentially viral image or this behavior is intended?

See [this line](https://github.com/sylabs/singularity/blob/master/internal/app/singularity/pull.go#L87).

WestleyK:

This PR adds back some missing code from #3789, in specific, the check where its
compairing the image hash, and removes the image if the hash is incorrect. Then
re-downloads it instead of failing, and returning a error.

- Fixes #4187 

<br>


internal/app/singularity/pull.go
```diff
@@ -68,7 +68,13 @@ func LibraryPull(imgCache *cache.Handle, name, fullURI, libraryURI, keyServerURL
 		imageName := uri.GetName("library://" + imageRef)
 		imagePath := imgCache.LibraryImage(libraryImage.Hash, imageName)
 		exists, err := imgCache.LibraryImageExists(libraryImage.Hash, imageName)
-		if err != nil {
+		if err == cache.ErrBadChecksum {
+			sylog.Warningf("Removing cached image: %s: cache could be corrupted", imagePath)
```
sashayakovtseva, line 73: Why cache can be corrupted in this case? I mean, you remove image and removal is pretty transparent to the user.

internal/app/singularity/pull.go
```diff
@@ -68,7 +68,13 @@ func LibraryPull(imgCache *cache.Handle, name, fullURI, libraryURI, keyServerURL
...
+			err := os.Remove(imagePath)
+			if err != nil {
+				return fmt.Errorf("unable to remove corrupted cache: %v", err)
```
sashayakovtseva: You are removing corrupted image, not cache.

sashayakovtseva (changes requested):

internal/app/singularity/pull.go
WestleyK: imagePath is the path to the cache, in the case of a download failure (and also fails to remove the non-complete downloaded cache), this will remove the corrupted cache, then redownload it.

WestleyK: The actual output image is not copied from cache until the download is complete.

sashayakovtseva: Yeah, I mean cache can be corrupted _only_ when malicious image removal fails, isn't it?

WestleyK: > cache can be corrupted _only_ when malicious image removal fails

Do you mean when the `go interruptCleanup(imagePath)` catches a interrupt?

Cache can also be corrupted if the download process fails, and `go interruptCleanup(imagePath)` was unable to cache it.

sashayakovtseva (approved):

internal/app/singularity/pull.go
gvallee:
```suggestion
				return fmt.Errorf("unable to remove corrupted image from cache: %v", err)
```

gvallee (changes requested):
Very minor suggestion

gvallee (approved):
lgtm

Issue #4094, opened 2019-08-02
[E2E] Improve coverage for run command


gvallee:
## Version of Singularity:

All.

### Expected behavior

All commands expected to succeed should be tested.
Part of issue #4078

### Actual behavior

- [ ] singularity run --add-caps
- [ ] singularity run --allow-setuid
- [ ] singularity run --apply-cgroups
- [ ] singularity run --cleanenv
- [ ] singularity run --contain
- [ ] singularity run --containall
- [ ] singularity run --containlibs
- [ ] singularity run --disable-cache
- [ ] singularity run --dns
- [ ] singularity run --docker-login
- [ ] singularity run --docker-password
- [ ] singularity run --docker-username
- [ ] singularity run --drop-caps
- [ ] singularity run --fakeroot
- [ ] singularity run --fusemount
- [ ] singularity run --home
- [ ] singularity run --hostname
- [ ] singularity run --ipc
- [ ] singularity run --keep-privs
- [ ] singularity run --net
- [ ] singularity run --network
- [ ] singularity run --network-args
- [ ] singularity run --no-home
- [ ] singularity run --no-init
- [ ] singularity run --no-nv
- [ ] singularity run --no-privs
- [ ] singularity run --nohttps
- [ ] singularity run --nonet
- [ ] singularity run --nv
- [ ] singularity run --overlay
- [ ] singularity run --pid
- [ ] singularity run --pwd
- [ ] singularity run --scratch
- [ ] singularity run --security
- [ ] singularity run --tmpdir
- [ ] singularity run --userns
- [ ] singularity run --uts
- [ ] singularity run --vm
- [ ] singularity run --vm-cpu
- [ ] singularity run --vm-err
- [ ] singularity run --vm-ram
- [ ] singularity run --workdir
- [ ] singularity run --writable
- [ ] singularity run --writable-tmpfs

### Steps to reproduce behavior

N/A.

gvallee:
Also add a test for the following use cases since it is a problem that is important for some users and it currently seems to fail (from PR #4198):
```
$ mkdir /tmp/cache

$ export SINGULARITY_CACHEDIR=/tmp/cache

$ singularity run library://godlovedc/funny/lolcow
INFO:    Downloading library image
 _________________________________________
/ Let us endeavor so to live that when we \
| come to die even the undertaker will be |
| sorry.                                  |
|                                         |
| -- Mark Twain, "Pudd'nhead Wilson's     |
\ Calendar"                               /
[...]

PR #4198, opened 2019-08-14
Make sure the cache is not used with a read-only target directory

### LINKED ISSUE #4150: 3.3.0 adds hidden requirement of a writable home directory
## Version of Singularity:

3.3.0

### Expected behavior

To be able to continue to run singularity from an account without a writable home directory, as in previous versions.

### Actual behavior

If there's no home directory, first get an error like this:
```
FATAL: Error attempting to create /grid_home/cmsprod/.singularity: mkdir /grid_home/cmsprod/.singularity: no such file or directory
```

That was due to pr #3661 from @godloved.  That's easily worked around, by changing the fatal error into a debug message.  However, then there's another error:
```
FATAL:   failed to create an image cache handle: failed initializing caching directory: couldn't create cache directory /grid_home/cmsprod/.singularity/cache: mkdir /grid_home/cmsprod: permission denied
```
That is due to pr #3758 from @gvallee, and that is much more difficult to get around.  Making an image cache handle now seems to be very deeply embedded in the process beginning in actions.go.  However, there should fundamentally be no need to make an image cache when the image is a sandbox like we in the grid community almost always use.  I wouldn't think there would be a need with a .sif file either for that matter.

Use an account that has no home directory or that is owned by someone else so there is no write access, and try to exec an image.

Attn:  @singularity-maintainers
This is a showstopper for me to get 3.3.0 released in EPEL.  I need a patch.

gvallee:

This PR introduces a new semantic to the image cache: if the target location of the image cache (will it be the default location, i.e., `~/.singularity/cache` or another location specified by `SINGULARITY_CACHEDIR`) is read-only, i.e., not writable, the cache shall be implicitly disabled when a new cache handle is created, *without* raising an error.
The key point with this PR is that it does not rely on the environment variable `SINGULARITY_DISABLE_CACHE` and that the cache is disabled without any action from the user.
Note that once the cache handle is created as disabled because the target is not writable, it is *not* possible to active it ever. This is due to the fact that we have no control over when the target could become writable and therefore we cannot guarantee correctness.

This PR also includes a e2e test to ensure that we test this configuration. The test ensures that when pulling an image while the cache directory is read-only, the cache is not created, without raising an error.

- Addresses #4150 


internal/pkg/client/cache/cache_test.go
```diff
@@ -22,7 +24,7 @@ const (
 // SINGULARITY_DISABLE_CACHE environment variable is set. If it is set, it will
 // skip the current tests since exercising the cache when caching is disabled is
 // not supported.
-func chechIfCacheDisabled(t *testing.T) {
```
DrDaveD, line 25: This should be "check", not "chech", right?

DrDaveD:
This looks like a good approach, but please test with the home directory completely missing because I keep stumbling on other problems.  First I run into this [fatal error](https://github.com/sylabs/singularity/blob/master/cmd/internal/cli/singularity.go#L199) that I first mentioned in  issue #4150, and if I turn that into a Debugf instead of Fatalf (and run it with -vvvd) I then see this:
```
DEBUG   [U=9996,P=7512]    mountGeneric()                Mounting /cloud/login/dwd/work/singularity/centos7-sandbox to /var/singularity/mnt/session/rootfs
FATAL   [U=9996,P=7512]    Master()                      container creation failed: hook function for tag prelayer returns error: failed to create /grid/home/cmsprod directory: mkdir /grid/home/cmsprod: permission denied
```
I tried finding a patch for that but the code is a maze of twisty little passages.  
A golang debugger where one could set breakpoints would probably be a big help here.  These are the few lines of strace before the error in case it helps:
```
7530  mkdirat(AT_FDCWD, "/var/singularity/mnt/session/underlay/scratch", 0755) = 0
7530  mkdirat(AT_FDCWD, "/var/singularity/mnt/session/underlay/scratch/home", 0755) = 0
7530  newfstatat(AT_FDCWD, "/grid/home/cmsprod", 0xc0003a2ac8, 0) = -1 ENOENT (No such file or directory)
7530  mkdirat(AT_FDCWD, "/grid/home/cmsprod", 0755) = -1 EACCES (Permission denied)
```
If I go ahead and create the home directory not owned by the user, then it skips the newfstatat and last mkdirat and instead creates /var/singularity/mnt/session/underlay/scratch/home/cmsprod, and successfully starts the container.

DrDaveD:
Ok by putting in debug statements including PrintStack() I found that "newfstatat" above is from line 250 of [this commit](https://github.com/sylabs/singularity/pull/3970/commits/40331a5b11185c1ed5a5e14305fdd8007a5f81c6#diff-60769b7cff9e6dae88716c3eab49b80fR250).  And shouldn't line 251 be statting m.rootPath + ovDir and setting path to that value?

DrDaveD:
That code is from pr #3970 which is only in the master branch.  This PR is addressing a problem in 3.3.0 so probably it would work better in the release-3.3 branch.

DrDaveD:
I don't think the behavior change of this PR is significant enough that it requires a Release 3.4.  This is needed as an urgent bug fix to Release 3.3.

gvallee:
@DrDaveD thanks for the feedback. I am not in charge of deciding what goes in what release. This PR is strictly about a problem I could reproduce and fortunately we will have all we need to make sure that this problem will not reappear in the future. Similarly, we usually develop in master and then port the work to a release branch, I am just following the process. :)

But it sounds like this problem could be fixed, I will see if we can get this approved.

Thanks again for your bug report and your feedback.

gvallee:
I also opened a new issue (#4208) specifically for the bug that you highlight with your trace.

DrDaveD:
@godloved marked issue #4150 as a milestone for 3.3.1, so it's already been recognized that a fix like the one in this PR is needed there.  I do understand though that it should go first into the master branch and be backported.  I tried a couple of simple ways to backport it myself (cherry-pick of the commits & applying a source diff with patch) and quickly hit major snags so I decided to leave that for later.

gvallee:
@DrDaveD No worries at all. As soon as it is merged into master, if no one object it, I will be happy to port it to the 3.3.x branch, whatever it takes. :)

internal/app/singularity/pull.go
```diff
@@ -56,7 +56,7 @@ func LibraryPull(imgCache *cache.Handle, name, fullURI, libraryURI, keyServerURL
 		return fmt.Errorf("could not get image info: %v", err)
 	}
 
-	if noCache {
+	if noCache || imgCache.IsDisabled() {
```
WestleyK: Why would one of these be true and not the other? Would't that be a issue if part of the code is detecting noCache, and the other part is not? Unless I missing something, I think there should be a check here, something like:

```go
if noCache != imgCache.IsDisabled() {
    sylog.Debugf("noCache=%v, imgCache.IsDisabled()=%v", noCache, imgCache.IsDisabled())
    return fmt.Errorf("unable to detect if cache is disabled or not")
}
```

**Or**, you can skip the `noCache` bool and only rely on `imgCache.IsDisabled()` for disabling the cache?...

gvallee: Good point! I will check but I am pretty sure you are right, `noCache` most certainly can be skipped.

WestleyK: Wait, maybe... If `imgCache.IsDisabled()` is checking if cache is disabled by only looking at the `SINGULARITY_DISABLE_CACHE`, then it will miss the `--disable-cache` command-line-option...

gvallee: Mmmm, good point again. Sounds like we need a few more e2e tests here to make sure everything will be consistent. I will work on the e2e tests in a different PR. We now have enough different code paths that it will be difficult to ensure that we do not break something without these tests.
I may also have a look at the code path the support of `--disable-cache` and make sure things are clean and consistent. I will ping you when code will be ready to get your feedback.

GodloveD:
I'm playing with this and I'm wondering about the following situation.  My cache is read-only, but it contains the image I'm trying to run.  Should we expect that use case to work?  For instance:
```
$ mkdir /tmp/cache

$ export SINGULARITY_CACHEDIR=/tmp/cache

$ singularity run library://godlovedc/funny/lolcow
 _________________________________________
/ Let us endeavor so to live that when we \
| come to die even the undertaker will be |
| sorry.                                  |
|                                         |
| -- Mark Twain, "Pudd'nhead Wilson's     |
\ Calendar"                               /
 -----------------------------------------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||

$ chmod 555 /tmp/cache

$ singularity run library://godlovedc/funny/lolcow
FATAL:   Unable to handle library://godlovedc/funny/lolcow uri: unable to download image: error opening file  for writing: open : no such file or directory
```
I kindof expect the last command to run because it shouldn't be trying to write to the cache, it should be trying to read from the cache.  What kind of behavior do others expect?

internal/pkg/util/fs/helper_linux_test.go
```diff
@@ -334,3 +334,67 @@ func TestMakeTempFile(t *testing.T) {
 		}
 	}
 }
+
+func TestIsWritable(t *testing.T) {
+	// We make a tmeporary directory where all the different cases will be tested
```
GodloveD: lol ol' typo `tmeporary `.  no big deal

internal/pkg/client/cache/dir.go
```diff
@@ -107,12 +107,22 @@ func NewHandle(baseDir string) (*Handle, error) {
...
 
+	newCache.disabled = false
+
+	/* Initialize the base directory of the cache */
 	if baseDir == "" {
 		baseDir = getCacheBasedir()
 	}
+	newCache.baseDir = baseDir
+	// We check if we can write to the basedir, if not we disable the caching mechanism
+	writable, _ := fs.IsWritable(baseDir)
```
GodloveD: `fs.IsWritable()` doesn't really check to see if the directory is writable by the _current user_.  It just checks to see if the user bit is set to writable or not.  As a consequence, I think we are getting an unexpected error in the case where the directory ownership prevents writing.  For instance:
```
$ sudo mkdir /tmp/cache 

$ sudo chmod 700 /tmp/cache

$ export SINGULARITY_CACHEDIR=/tmp/cache

$ singularity run ~/containers/lolcow.sif 
FATAL:   Failed to create an image cache handle: failed initializing caching directory: unable to stat /tmp/cache/cache: stat /tmp/cache/cache: permission denied
```
The last command is just trying to run a local container.  I assume that `IsWritable()` returns `true` even though the current user cannot write to it (because of the ownership) and then Singularity fails to disable the cache and errors out when it tries to write.

internal/pkg/util/fs/helper.go
```diff
@@ -172,3 +172,22 @@ func MakeTmpFile(basedir, pattern string, mode os.FileMode) (*os.File, error) {
...
+	info, err := os.Stat(dir)
+	if err != nil {
+		return false, err
+	}
+
+	if !info.IsDir() {
+		return false, fmt.Errorf("%s is not a directory", dir)
+	}
+
+	// Check whether the user bit for write permission is set
```
GodloveD: As stated above, we should probably check ownership too so that we can determine if the directory can be written to by the _current user_.

GodloveD (changes requested):
I think there is a corner case that's not covered here when a directory is writable by the owner but not writable by the current user.

internal/pkg/client/cache/dir.go
gvallee: Actually, it does, I just wrote a new unit test case for a 700 directory and the unit test works. With the fixes from the other issues (I suspect when I revisited slightly the pull code path), it also seems to be working now:
```
$ export SINGULARITY_CACHEDIR=/tmp/mecache/
$ chmod 700 /tmp/mecache/
$ singularity pull /tmp/lolcow.sif library://godlovedc/funny/lolcow
WARNING: Container might not be trusted; run 'singularity verify /tmp/lolcow.sif' to show who signed it
INFO:    Download complete: /tmp/lolcow.sif
```

internal/pkg/util/fs/helper.go
gvallee: That is why I check the user bit. I also added a unit test that seems to be working as expected. With all the fixes lately pushed to the cache management, that problem seems to be fixed.

internal/app/singularity/pull.go
gvallee: Fixed thanks to all the PRs pushed over the past few days related to the cache management. Yay! :)

gvallee:
> I'm playing with this and I'm wondering about the following situation. My cache is read-only, but it contains the image I'm trying to run. Should we expect that use case to work? For instance:
> 
> ```
> $ mkdir /tmp/cache
> 
> $ export SINGULARITY_CACHEDIR=/tmp/cache
> 
> $ singularity run library://godlovedc/funny/lolcow
> INFO:    Downloading library image
>  _________________________________________
> / Let us endeavor so to live that when we \
> | come to die even the undertaker will be |
> | sorry.                                  |
> |                                         |
> | -- Mark Twain, "Pudd'nhead Wilson's     |
> \ Calendar"                               /
>  -----------------------------------------
>         \   ^__^
>          \  (oo)\_______
>             (__)\       )\/\
>                 ||----w |
>                 ||     ||
> 
> $ chmod 555 /tmp/cache
> 
> $ singularity run library://godlovedc/funny/lolcow
> INFO:    Downloading library image
> FATAL:   Unable to handle library://godlovedc/funny/lolcow uri: unable to download image: error opening file  for writing: open : no such file or directory
> ```
> 
> I kindof expect the last command to run because it shouldn't be trying to write to the cache, it should be trying to read from the cache. What kind of behavior do others expect?

Good catch, it is clearly not working but it is unclear why at the moment. I will write a e2e test for this (I just added a note to issue #4094) and will investigate that.

gvallee:
> > I'm playing with this and I'm wondering about the following situation. My cache is read-only, but it contains the image I'm trying to run. Should we expect that use case to work? For instance:
> > ```
> > $ mkdir /tmp/cache
> > 
> > $ export SINGULARITY_CACHEDIR=/tmp/cache
> > 
> > $ singularity run library://godlovedc/funny/lolcow
> > INFO:    Downloading library image
> >  _________________________________________
> > / Let us endeavor so to live that when we \
> > | come to die even the undertaker will be |
> > | sorry.                                  |
> > |                                         |
> > | -- Mark Twain, "Pudd'nhead Wilson's     |
> > \ Calendar"                               /
> >  -----------------------------------------
> >         \   ^__^
> >          \  (oo)\_______
> >             (__)\       )\/\
> >                 ||----w |
> >                 ||     ||
> > 
> > $ chmod 555 /tmp/cache
> > 
> > $ singularity run library://godlovedc/funny/lolcow
> > INFO:    Downloading library image
> > FATAL:   Unable to handle library://godlovedc/funny/lolcow uri: unable to download image: error opening file  for writing: open : no such file or directory
> > ```
> > 
> > 
> > I kindof expect the last command to run because it shouldn't be trying to write to the cache, it should be trying to read from the cache. What kind of behavior do others expect?
> 
> Good catch, it is clearly not working but it is unclear why at the moment. I will write a e2e test for this (I just added a note to issue #4094) and will investigate that.

@GodloveD I just did a few minor modifications, rebased my branch against master and added a test, it seems to be working now.

GodloveD:
Hey @DrDaveD I just added this to the 3.4 milestone.  It doesn't mean I've forgotten about this for 3.3.1 but a PR can only be on 1 milestone unfortunately.

DrDaveD:
@GodloveD If you have to pick one I would think it would make more sense to choose the milestone that comes first chronologically.

internal/pkg/util/fs/helper_linux.go
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2018-2019, Sylabs Inc. All rights reserved.
```
WestleyK: This looks like this file was created in 2019

```suggestion
// Copyright (c) 2019, Sylabs Inc. All rights reserved.
```

WestleyK (commented):
Looks good to me, just one comment...

WestleyK (approved):
Looks good to me :+1:

GodloveD (approved):
lgtm

GodloveD:
Feel free to merge pending conflict resolution @gvallee .  Thank you

PR #4205, opened 2019-08-14
Improving 'key search --long-list' to use tabwriter

### LINKED ISSUE #4156: Improve 'key search --long-list' to use text/tabwriter
## Version of Singularity:

```
3.3.0
```

### Expected behavior

```
$ singularity key search --long-list
```
Will use the `text/tabwriter` package.

### Actual behavior

Does not exist yet.

Write here.

WestleyK:

This PR improves the `key search --long-list` by using the `tabwriter` package. And refactor the mr parsing.

- Fixes #4156

<br>


pkg/sypgp/sypgp.go
```diff
@@ -703,112 +702,108 @@ func getEncryptionAlgorithmName(n string) (string, error) {
...
+				keyDateExpired, err = date(nk[5])
+				if err != nil {
+					panic(err)
+				}
+				if nk[6] == "r" {
+					keyStatus = "[revoked]"
+				} else if nk[6] == "d" {
+					keyStatus = "[disabled]"
+				} else if nk[6] == "e" {
+					keyStatus = "[xpired]"
```
gvallee:
```suggestion
					keyStatus = "[expired]"
```

gvallee (changes requested):
I only have a very minor comment.

pkg/sypgp/sypgp.go
```diff
@@ -703,112 +702,108 @@ func getEncryptionAlgorithmName(n string) (string, error) {
...
-		if err != nil {
-			return "", err
-		}
+	key := strings.Split(mrString, "\n")
 
-		expirationTimestamp, err := date(expiryDate)
-		if err != nil {
-			return "", err
-		}
+	var (
```
ikaneshiro: Why declare all of these variables here when they could be within the scope of the first loop? Then you would not need to "zero out" their values after iteration on the second loop completes.

pkg/sypgp/sypgp.go
```diff
@@ -703,112 +702,108 @@ func getEncryptionAlgorithmName(n string) (string, error) {
...
+		keyReady       bool
+	)
 
-		// check the status of each key if flags are present
-		if len(featuresKey) == 7 {
-			if featuresKey[6] == "r" {
-				status = "revoked"
+	for _, k := range key {
+		nk := strings.Split(k, ":")
+		for _, n := range nk {
```
ikaneshiro, line 796: Could we break this inner loop into a function that returns the key parameters in a struct or something like that?

ikaneshiro (changes requested):

pkg/sypgp/sypgp.go
WestleyK: That wont work, the second loop can loop 1-3 times, and we dont know how much until after the `uid` section.

gvallee (approved):
lgtm

ikaneshiro (approved):

Issue #4087, opened 2019-08-02
[E2E] Improve coverage for key command


gvallee:
## Version of Singularity:

All.

### Expected behavior

All commands expected to succeed should be tested.
Part of issue #4078

### Actual behavior

- [x] singularity key export --armor
- [x] singularity key export --help
- [x] singularity key export --secret
- [x] singularity key import --help
- [x] singularity key list --help
- [x] singularity key list --secret
- [x] singularity key newpair --help
- [ ] singularity key newpair --new-password (pending https://github.com/sylabs/singularity/pull/4405)
- [x] singularity key newpair --bit-length (included in PR #4207)
- [x] singularity key pull --help
- [ ] singularity key pull --url
- [x] singularity key push --help
- [ ] singularity key push --url
- [x] singularity key remove --help
- [x] singularity key search --help
- [x] singularity key search --url
- [ ] singularity key search --long-list (included in PR #4121)

### Steps to reproduce behavior

N/A.

gvallee:
I executed a coverage test on 2019/08/29 and `key pull --help`, `key push --help` and `key remove --help` are not reported as covered while tests are supposed to be present. This needs further investigation.

dtrudg:
Closing. These "[E2E] Improve Coverage" issues have not been maintained, and are out of date with respect to test improvements, and new flags / arguments.

#5830 is a task to remove redundancy from unit/integration <-> e2e and will revisit e2e coverage to an extent.

PR #4207, opened 2019-08-14
Allow users to specify the key length of newpair

### LINKED ISSUE #4157: Ability to specify key bit length when 'newpair'
## Version of Singularity:

```
3.3.0
```

### Expected behavior

```bash
$ singularity key newpair --bit-length=1024
```

### Actual behavior

Does not exist yet...

Write here.

WestleyK:

This PR adds a flag to allow the users to specify the key length of newpair.

- Fixes #4157

<br>


pkg/sypgp/sypgp.go
```diff
@@ -486,8 +493,10 @@ func (keyring *Handle) GenKeyPair(keyServiceURI string, authToken string) (*open
 
 	fmt.Printf("Generating Entity and OpenPGP Key Pair... ")
 
-	entity, err := keyring.genKeyPair(name, comment, email, passphrase)
+	entity, err := keyring.genKeyPair(name, comment, email, passphrase, bitLen)
 	if err != nil {
+		// Print the missing newline if theres a error
```
gvallee:
```suggestion
		// Print the missing newline if there’s a error
```

gvallee (changes requested):
Minor change

gvallee:
If/when merged, we need to add the new option to issue #4087 to ensure we will write the appropriate e2e test.

pkg/sypgp/sypgp.go
```diff
@@ -486,8 +493,10 @@ func (keyring *Handle) GenKeyPair(keyServiceURI string, authToken string) (*open
 
 	fmt.Printf("Generating Entity and OpenPGP Key Pair... ")
 
-	entity, err := keyring.genKeyPair(name, comment, email, passphrase)
+	entity, err := keyring.genKeyPair(name, comment, email, passphrase, bitLen)
 	if err != nil {
+		// Print the missing newline if there’s a error
```
WestleyK: Needs to be 'an'

```suggestion
		// Print the missing newline if there’s an error
```

pkg/sypgp/sypgp.go
```diff
@@ -486,8 +493,10 @@ func (keyring *Handle) GenKeyPair(keyServiceURI string, authToken string) (*open
 
 	fmt.Printf("Generating Entity and OpenPGP Key Pair... ")
 
-	entity, err := keyring.genKeyPair(name, comment, email, passphrase)
+	entity, err := keyring.genKeyPair(name, comment, email, passphrase, bitLen)
 	if err != nil {
+		// Print the missing newline if there’s an error
+		fmt.Printf("\n")
```
sashayakovtseva, line 511: the error is printed to the stdout?

pkg/sypgp/sypgp.go
```diff
@@ -458,11 +460,16 @@ func (keyring *Handle) genKeyPair(name, comment, email, passphrase string) (*ope
...
 // GenKeyPair generates an PGP key pair and store them in the sypgp home folder
-func (keyring *Handle) GenKeyPair(keyServiceURI string, authToken string) (*openpgp.Entity, error) {
+func (keyring *Handle) GenKeyPair(keyServiceURI, authToken string, bitLen int) (*openpgp.Entity, error) {
 	if err := keyring.PathsCheck(); err != nil {
 		return nil, err
 	}
 
+	// Fail early if the bit length is invalid
+	if bitLen < 1024 || bitLen > 4096 {
+		return nil, ErrInvalidBitLen
```
sashayakovtseva: Why having this particular check?

sashayakovtseva (changes requested):

pkg/sypgp/sypgp.go
WestleyK: This only adds the newline after:

```go
fmt.Printf("Generating Entity and OpenPGP Key Pair... ") // line 487, or 494
```

In case the key gen fails, this just adds the newline so the return err is printed at the next line.

sashayakovtseva: Why not adding \n into the actual output then?

WestleyK: Nicer output, the `openpgp.NewEntity` can take some time to generate the key, and it looks nicer with the `... done` imo.

pkg/sypgp/sypgp.go
WestleyK: Should it not be here? I thought the key length was supposed to be in this range. (maybe could start a bit lower, like 7xx bits)...

sashayakovtseva: Maybe, this check should be removed? 
If I am not mistaken, RSA does not restrict key len to be in this range as well as to be a power of two.

tri-adam: Agreed. No need to artificially restrict, let the underlying RSA library enforce its conditions.

gvallee (approved):
lgtm

sashayakovtseva (approved):

PR #4211, opened 2019-08-15
Enable seccomp during oci mount testing

jscook2345:
This was initially disabled because `14.04/trusty` did not support some of the system calls in the default `seccomp` profile


dtrudg (approved):
LGTM

WestleyK (approved):
Looks good to me. Maybe just remove the line, instead of commenting it out. Or add another command on why its commented. _just thinking out loud, not necessary..._

Issue #4192, opened 2019-08-13
Cache management code should not directly handle the environment variables


gvallee:
## Version of Singularity:

master.

### Expected behavior

Checking the environment variables should be done as close as possible to the CLI code. As a result, the code for the management of image caches should not include any call to `os.Getenv()` or `os.Setenv()`.

The CLI code should therefore look like:
```
var cacheEnv cache.Env
cacheEnv.CacheDir = os.Getenv(cache.DirEnv)
cacheEnv.DisableCache = os.Getenv(cache.DisableEnv

c := cache.NewHandle(cacheEnv)
```

Then, the code that implements the image cache only need to rely on the environment that is specified in the `Env` argument.

### Actual behavior

The current implementation checks both `SINGULARITY_DISABLE_CACHE` and `SINGULARITY_CACHEDIR`.

### Steps to reproduce behavior

Write here.

carterpeel:
Hello,

This is a templated response that is being sent out to all open issues.  We are working hard on 'rebuilding' the Singularity community, and a major task on the agenda is finding out what issues are still outstanding.


**Please consider the following:**

1. Is this issue a duplicate, or has it been fixed/implemented since being added?
2. Is the issue still relevant to the current state of Singularity's functionality?
3. Would you like to continue discussing this issue or feature request?


Thanks,
Carter

pedroalvesbatista:
@gvallee Still getting this problem ? Have this been solved already ? If yes, what work around have you followed and applied ?

We're looking into the issue carefully, soon will bring to community and discuss ways to better solve as well address this. Thankyou for keeping interest over the subject.

kmuriki:
Moved to the new Apptainer repo issue. https://github.com/apptainer/apptainer/issues/1157

PR #4215, opened 2019-08-15
Change the interface to create a new image cache handle 

### LINKED ISSUE #4213: Command pull --disable-cache still creates a cache.
## Version of Singularity:

master

### Expected behavior

When running `singularity pull --disable-cache`, the cache should be entirely bypassed. This is for instance required when the location for the cache (default location or specified by user) is not writable (grid environment and such). Otherwise the command will fail.

### Actual behavior

The command will still create the directories related to the cache:
```
$ cd /tmp/e2e-imgcache-905650874/cache
$ tree
.
├── library
├── net
├── oci
├── oci-tmp
├── oras
└── shub
```

Create a temporary directory, set the cache location to that temporary directory and run `singularity pull --disable-cache test.sif library://alpine:latest`.

gvallee:

This PR changes the interface to create a new image cache handle to use a context (extensible structure) instead of a single string for the specification of the base directory. This is required because we now have various ways to impact the cache, namely environment variables (e.g., `SINGULARITY_CACHE_DIR`, `SINGULARITY_DISABLE_CACHE`) and command options (e.g., `singularity pull --disable-cache`). As a result, we need to be able to capture the context in which a new image cache handle is created.

This PR only changes the interface for the creation of a cache handle, in order to ease the its review. The goal here is to only introduce a new struct `Context` that can be used to capture the context in which a cache is created, as well as the changes to the interface to support the context, as well as all the changes necessary.

Extension to the `Context` structure to address #4213 and #4192 will be introduced in future PRs.

- Addresses #4213 
- Addresses #4192 

### DIFF
--- a/internal/pkg/client/cache/dir.go
+++ b/internal/pkg/client/cache/dir.go
@@ -38,9 +38,9 @@ const (
 	CacheDir = "cache"
 )
 
-// Context describes the context in which a new handle is created, as defined
-// by the user through command flags and environment variables.
-type Context struct {
+// Config describes the requested configuration requested when a new handle is created,
+// as defined by the user through command flags and environment variables.
+type Config struct {
 	// Location where the user wants the cache to be created.
 	BaseDir string
 }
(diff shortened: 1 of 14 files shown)


internal/pkg/client/cache/dir.go
```diff
@@ -38,6 +38,13 @@ const (
 	CacheDir = "cache"
 )
 
+// Context describes the context in which a new handle is created, as defined
+// by the user through command flags and environment variables.
+type Context struct {
```
ikaneshiro: I have a knee-jerk reaction to the name of this struct just because of the prevalence of the https://golang.org/pkg/context/ package to manage go routine lifetimes. What are your thoughts on renaming this to `Config`?

ikaneshiro (approved):
This LGTM, only one suggestion that isnt a blocker

ikaneshiro (approved):

PR #4221, opened 2019-08-16
Do not encrypt pk if pass was not provided

pisarukv:
Before ,even if passphrase was empty, we were encrypting pk. As the result, we were asking user to enter a password even it was empty.


sashayakovtseva (approved):

WestleyK (changes requested):
Thanks for adding this, but as it is it fails to export a secret key without a password:

```
$ singularity key export -s ./bar
0) U: test () <bar>
   C: 2019-08-16 08:44:48 -0700 PDT
   F: 642A128AEA64518B65E4D597AD83C68A74DE7BA5
   L: 4096
   --------
Enter # of private key to use : 0
Enter key passphrase : # pressed enter, because theres no password
ERROR:   key export command failed: key is not encrypted
```

I suggest try adding a "if key is un-encrypted then skip the password decryption" in the export function.

schebro (approved):
lgtm

pisarukv:
> Thanks for adding this, but as it is it fails to export a secret key without a password:
> 
> ```
> $ singularity key export -s ./bar
> 0) U: test () <bar>
>    C: 2019-08-16 08:44:48 -0700 PDT
>    F: 642A128AEA64518B65E4D597AD83C68A74DE7BA5
>    L: 4096
>    --------
> Enter # of private key to use : 0
> Enter key passphrase : # pressed enter, because theres no password
> ERROR:   key export command failed: key is not encrypted
> ```
> 
> I suggest try adding a "if key is un-encrypted then skip the password decryption" in the export function.

Thanks @WestleyK I have added a check in export

WestleyK (commented):
Looks good to me. Maybe there should be a `WARNING` during the `newpair` if theres no password, something like: `WARNING: if theres no password set, your key is not secure. Do you want to continue? [y/n] ` or something like that. Open to suggestions...

cali4888:
I agree that warning is required, but not

pisarukv:
Have added the requested message in a pr related to non-interactive key newpair https://github.com/sylabs/singularity/pull/4240

WestleyK (approved):
Looks good to me :+1:

PR #4240, opened 2019-08-19
Non interactive newkey

### LINKED ISSUE #4232: 'key newpair' should have a non-interactive option
## Version of Singularity:

```
3.3
```

### Expected behavior

`singularity key newpair` should be non-interactive, we should pass all the key information as arguments. But if no arguments are passed, then it should be interactive as it is now.

### Actual behavior

Only interactive right now.

Write here.

pisarukv:
closes https://github.com/sylabs/singularity/issues/4232


pkg/sypgp/sypgp.go
```diff
@@ -471,35 +481,60 @@ func (keyring *Handle) genKeyPair(name, comment, email, passphrase string) (*ope
...
-	if err != nil {
-		return nil, err
+	if opts.Password == nil {
+		// get a password
+		p, err := interactive.GetPassphrase("Enter a passphrase : ", 3)
+		if err != nil {
+			return nil, err
+		}
+		if p == "" {
+			a, err := interactive.AskYNQuestion("n", "WARNING: if there is no password set, your key is not secure. Do you want to continue? [y/n]")
```
```suggestion
			a, err := interactive.AskYNQuestion("n", "WARNING: if there is no password set, your key is not secure. Do you want to continue? [y/n] ")
```

pkg/sypgp/sypgp.go
```diff
@@ -57,6 +57,14 @@ type Handle struct {
 	path string
 }
 
+// GenKeyPairOptions parameters needed for generating new key pair.
+type GenKeyPairOptions struct {
+	Name     *string
```
WestleyK: Why are you using pointers here?

WestleyK (changes requested):
Looks good, just what I was thinking about using a struct to pass the the key information. Just a couple suggestions about the workflow...

Imo, if you just specify one (or all but one) of the key requirements with the command line, it should fail. Because if a user (or script) only passes the `--name`, `--pass`, and forgets the `--email`, the script will hang. I suggest that if not all the key requirements are passed, it should fail. Heres a example:

```bash
$ singularity key newpair --name=test --password=test123
ERROR:   Missing '--email' argument
```

Another suggestion is to change your shorthand arguments to caps, like `-e` to `-E`. The reason is that some user may only use one dash (`-`) for the long argument, that will result in a unexpected output. Heres a example:

```bash
$ singularity key newpair -email=foo@sylabs.io
# will result in a email of: "mail=foo@sylabs.io"
```

By using caps as the short arguments, that can be avoided.

pkg/sypgp/sypgp.go
pisarukv: Since i want to distinguish when user didn't pass a flag and when he specified empty field. For example, empty string for password field is a valid value and without reference we will not know whether to ask user input or not.

pkg/sypgp/sypgp.go
```diff
@@ -445,17 +453,19 @@ func (keyring *Handle) RemovePubKey(toDelete string) error {
 	return keyring.storePubKeyring(newKeyList)
 }
 
-func (keyring *Handle) genKeyPair(name, comment, email, passphrase string) (*openpgp.Entity, error) {
+func (keyring *Handle) genKeyPair(opts GenKeyPairOptions) (*openpgp.Entity, error) {
 	conf := &packet.Config{RSABits: 4096, DefaultHash: crypto.SHA384}
 
-	entity, err := openpgp.NewEntity(name, comment, email, conf)
+	entity, err := openpgp.NewEntity(*opts.Name, *opts.Comment, *opts.Email, conf)
```
sashayakovtseva: Possible nil dereference?

sashayakovtseva:
@WestleyK you are right about capital shorthands, but we already have them lowercase almost everywhere, so maybe worth sticking with lowers here as well for CLI consistency.

pkg/sypgp/sypgp.go
sashayakovtseva: Nevermind.
Maybe add a comment clarifying that `opts` should not container nil values (except pswd)?

sashayakovtseva (approved):
LGTM!

WestleyK:
> @WestleyK you are right about capital shorthands, but we already have them lowercase almost everywhere, so maybe worth sticking with lowers here as well for CLI consistency.

Actually, the only non-cap short option that requires a option is `-o, --overlay` for `run` (afaik). Everything else that requires a additional option is a cap, like `-B` for `--bind`, `-H`, for `--home`, `-W` for  `--workdir` and so on... Also, I think its more "standard" to use caps in short arguments that require a variable, like gcc, and many other programs. 

_I still recommend changing the shorthand arguments to caps._

WestleyK (commented):
I still like the idea of failing if a argument is missing, like: 

```bash
$ singularity key newpair --name=test --password=test123
ERROR:   Missing '--email' argument
```

Thats something we can change later...

**But** something I do suggest changing, is that when you do:

```bash
$ singularity key newpair --name=test --email=foobar@hush.com -Cnull -Ppsk
Would you like to push it to the keystore? [Y,n] 
```

Theres no argument to disable this question. One option is to add another argument like `-l`, `--no-push` (-l for local) to avoid that interactive question.

_sorry its taking me a little longer to review..._

pisarukv:
@WestleyK about push to the keystore we have --push(-U) flag. If the flag specified we will not ask the question.

pisarukv:
About failing if an argument is missing. Not sure that it's what user are expecting while using flags. Need to think about it a bit more. Maybe we can pass all that options via config (yaml, json). I'll create an issue for that.

WestleyK:
> @WestleyK about push to the keystore we have --push(-U) flag. If the flag specified we will not ask the question.

I saw that, but is there a flag to disable the push without a question?

pisarukv:
Got it. My bad in documenting push flag, but it suppose to be a bool flag. So if User will pass —push=false the push will be disabled without a question.

cmd/internal/cli/key_newpair.go
```diff
@@ -6,34 +6,194 @@
...
+	}
+
+	keyNewPairPush     bool
+	KeyNewPairPushFlag = &cmdline.Flag{
+		ID:           "KeyNewPairPushFlag",
+		Value:        &keyNewPairPush,
+		DefaultValue: false,
+		Name:         "push",
+		ShortHand:    "U",
+		Usage:        "push generated key to a keystore",
```
WestleyK:
```suggestion
		Usage:        "specify to push the public key to the remote keystore (default true)",
```

And make the default behavior to push/true. Because that is the default when interactively generating a newpair.

Also suggest adding a example in `docs.KeyNewPairExample` of generating a newpair non-interactively, with `--push=false`.

cmd/internal/cli/key_newpair.go
```diff
@@ -6,34 +6,194 @@
...
-			os.Exit(2)
+var (
+	keyNewPairName     string
+	KeyNewPairNameFlag = &cmdline.Flag{
+		ID:           "KeyNewPairNameFlag",
+		Value:        &keyNewPairName,
+		DefaultValue: "",
+		Name:         "name",
+		ShortHand:    "N",
+		Usage:        "keys owner name",
```
WestleyK:
```suggestion
		Usage:        "key owner name",
```

Although, `keys` (with the `s`) would be correct, because theres two keys. But `key` sounds better imo.

cmd/internal/cli/key_newpair.go
```diff
@@ -6,34 +6,194 @@
...
+	}
+
+	keyNewPairEmail     string
+	KeyNewPairEmailFlag = &cmdline.Flag{
+		ID:           "KeyNewPairEmailFlag",
+		Value:        &keyNewPairEmail,
+		DefaultValue: "",
+		Name:         "email",
+		ShortHand:    "E",
+		Usage:        "keys owner email",
```
WestleyK: Same here:

```suggestion
		Usage:        "key owner email",
```

cmd/internal/cli/key_newpair.go
```diff
@@ -6,34 +6,194 @@
...
+	}
+
+	keyNewPairComment     string
+	KeyNewPairCommentFlag = &cmdline.Flag{
+		ID:           "KeyNewPairCommentFlag",
+		Value:        &keyNewPairComment,
+		DefaultValue: "",
+		Name:         "comment",
+		ShortHand:    "C",
+		Usage:        "keys comment",
```
WestleyK: Same here

```suggestion
		Usage:        "key comment",
```

cmd/internal/cli/key_newpair.go
```diff
@@ -6,34 +6,194 @@
...
+	}
+
+	keyNewPairPassword     string
+	KeyNewPairPasswordFlag = &cmdline.Flag{
+		ID:           "KeyNewPairPasswordFlag",
+		Value:        &keyNewPairPassword,
+		DefaultValue: "",
+		Name:         "password",
+		ShortHand:    "P",
+		Usage:        "keys password",
```
WestleyK: And here

```suggestion
		Usage:        "key password",
```

WestleyK (commented):
Yes, now I understand the `--push` flag. I just have a couple more suggestions. After that, it looks good :+1:

pisarukv:
@Westley Thanks for your fixes!

WestleyK (commented):
This looks good. Can you just add this example to the key newpair help?

```go
Examples:
  $ singularity key newpair --password=psk --name=your-name --comment="key comment" --email=mail@email.com --push=false
```

After that, this looks good to go!

WestleyK (approved):
Looks good to me :+1: Thanks for fixing this up.

go.mod
```diff
@@ -100,7 +100,7 @@ require (
 	google.golang.org/grpc v1.20.0 // indirect
 	gopkg.in/cheggaaa/pb.v1 v1.0.28
 	gopkg.in/yaml.v2 v2.2.2
-	gotest.tools v2.2.0+incompatible // indirect
+	gotest.tools v2.2.0+incompatible
```
mem: please don't add this, change the code to use gotest.tools/v3

mem (changes requested):

mem (approved):

PR #4243, opened 2019-08-19
Set up CI with Azure Pipelines

mem:


jscook2345:
@mem 

How's this going? We should chat about this a bit and see what to do.

dtrudg:
It doesn't appear as though we will pursue this. I am going to move the branch to the `vault/` for any future reference need.

PR #4245, opened 2019-08-19
Add --encrypt option to build command

### LINKED ISSUE #4241: Require -e flag to build encrypted containers
As the title says. This will make it clear that the build command is created an encrypted container. If we pick up env vars for for keys/passphrases, and this flag is not specified, we should give some kind of warning. Something that is not completely clear to me is if we should have this be required for `exec/shell/run` or if we should just act on envvars and other flags for those commands

ikaneshiro:

- Also adds check at runtime as specified in https://github.com/sylabs/singularity/issues/4241#issuecomment-522734611

- Fixes #4241

### DIFF
--- a/cmd/internal/cli/actions_linux.go
+++ b/cmd/internal/cli/actions_linux.go
@@ -19,7 +19,7 @@ import (
 
 	"github.com/opencontainers/runtime-tools/generate"
 	"github.com/sylabs/singularity/internal/pkg/plugin"
-	"github.com/sylabs/singularity/pkg/image"
+	imgutil "github.com/sylabs/singularity/pkg/image"
 	"github.com/sylabs/singularity/pkg/image/unpacker"
 	"github.com/sylabs/singularity/pkg/util/crypt"
 	"github.com/sylabs/singularity/pkg/util/namespaces"
@@ -54,7 +54,7 @@ func EnsureRootPriv(cmd *cobra.Command, args []string) {
 }
 
 func convertImage(filename string, unsquashfsPath string) (string, error) {
-	img, err := image.Init(filename, false)
+	img, err := imgutil.Init(filename, false)
 	if err != nil {
 		return "", fmt.Errorf("could not open image %s: %s", filename, err)
 	}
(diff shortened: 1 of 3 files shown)


schebro (approved):

cmd/internal/cli/actions_linux.go
```diff
@@ -258,12 +258,32 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
...
 	}
+	defer img.File.Close()
+
+	if !img.HasRootFs() {
+		sylog.Fatalf("no root filesystem found in %s", engineConfig.GetImage())
+	}
+
+	// ensure we have decryption material
+	if img.Partitions[0].Type == imgutil.ENCRYPTSQUASHFS {
+		sylog.Debugf("Encrypted container filesystem detected")
```
schebro: Encrypted filesystem detected?

cmd/internal/cli/actions_linux.go
```diff
@@ -258,12 +258,32 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
...
+		if encryptionKey == "" {
```
schebro: encryptionKey is retried on line 280, right? Won't we hit this Fatalf always?

cmd/internal/cli/build_linux.go
```diff
@@ -186,6 +191,17 @@ func run(cmd *cobra.Command, args []string) {
...
 	} else {
+		// ensure passphrase or key was supplied with encrypt option
+		if encrypt && !cmd.Flags().Lookup("encryption-key").Changed {
+			sylog.Fatalf("Unable to encrypt container, no passphrase or key path specified.")
+		}
+
+		// ensure we do not build an encrypted container if encrypt option is not specified
+		if !encrypt && cmd.Flags().Lookup("encryption-key").Changed {
+			sylog.Warningf("Encryption key environment variable found, but -e was not specified. NOT encrypting container.")
+			encryptionKey = ""
```
schebro, line 202: Where is this key used?

ikaneshiro: It is in the build config below as it was before this change. This is just preventing an encrypted container build from happening if `-e` isnt there

cmd/internal/cli/actions_linux.go
ikaneshiro: No this is just catching the case of no key material upfront. I think as is we might allow for an encrypted container with an empty passphrase?

cmd/internal/cli/actions_linux.go
ikaneshiro: I think its better to be slightly more verbose here

cmd/internal/cli/actions_linux.go
schebro: Didn't think about that case. It's fine for now I guess.

schebro (approved):

PR #4262, opened 2019-08-21
Abort suid install if not root

### LINKED ISSUE #4227: INSTALL SUID should check that effective user is actually root
## Version of Singularity:

3.3.0

### Expected behavior

If mistakenly using `make install` instead of `sudo make install` and setuid install is enabled, fail with a diagnostic.

### Actual behavior

It installs the setuid binary setuid to my build user, which is worse than useless.

omit sudo

sashayakovtseva:

This PR adds a small effective uid check during starter installation. If euid is not 0 then this installation doesn't make sense. It is better to abort it then to install with wrong suid bit.

- Fixes #4227

### DIFF
--- a/mlocal/frags/build_runtime_suid.mk
+++ b/mlocal/frags/build_runtime_suid.mk
@@ -11,7 +11,7 @@ $(starter_suid): $(BUILDDIR)/.clean-starter $(singularity_build_config) $(starte
 		-o $@ $(SOURCEDIR)/cmd/starter/main_linux.go
 
 $(starter_suid_INSTALL): $(starter_suid)
-	@if [ `id -u` -ne 0 ] ; then \
+	@if [ `id -u` -ne 0 -a -z "${RPM_BUILD_ROOT}" ] ; then \
 		echo "SUID binary requires to execute make install as root, use sudo make install to finish installation"; \
 		exit 1 ; \
 	fi
--- a/mlocal/frags/build_runtime_suid.mk
+++ b/mlocal/frags/build_runtime_suid.mk
@@ -12,7 +12,7 @@ $(starter_suid): $(BUILDDIR)/.clean-starter $(singularity_build_config) $(starte
 
 $(starter_suid_INSTALL): $(starter_suid)
 	@if [ `id -u` -ne 0 ] ; then \
-		echo "SUID is useful when running as root only, aborting"; \
+		echo "SUID binary requires to execute make install as root, use sudo make install to finish installation"; \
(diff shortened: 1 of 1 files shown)


mlocal/frags/build_runtime_suid.mk
```diff
@@ -11,6 +11,10 @@ $(starter_suid): $(BUILDDIR)/.clean-starter $(singularity_build_config) $(starte
 		-o $@ $(SOURCEDIR)/cmd/starter/main_linux.go
 
 $(starter_suid_INSTALL): $(starter_suid)
+	@if [ `id -u` -ne 0 ] ; then \
+		echo "SUID is useful when running as root only, aborting"; \
```
cclerget:
```suggestion
		echo "SUID binary requires to execute make install as root, use sudo make install to finish installation"; \
```

cclerget (approved):
Small suggestion, feel free to apply otherwise LGTM

pisarukv (approved):

jmstover:
Travis is failing with:
```
SUID binary requires to execute make install as root, use sudo make install to finish installation
```
If you do just a `make install` as a normal user, it also runs `install_SUID` doesn't it? You need the check in there for `with_suid -eq 1` don't you?

I don't know if this will fix Travis though... looks like it's building an RPM, as a normal user. When the RPM is installed by root, the permissions get applied. I _think_ permissions are separate. You can build a RPM as a normal user, it just tags what the permissions _should_ be.

So... attempting this could break that behavior... :/

jmstover (changes requested):
Travis failing ... `with_suid -eq 1` check needed as well?

sashayakovtseva:
@jmstover
> If you do just a make install as a normal user, it also runs install_SUID doesn't it?
It will run it if `with_suid` was true during `./mconfig` run.

 We don't need to check `with_suid` in makefile fragment b/c if that fragment appears in the resulting Makefile that means `with_suid` *was* set to true during `./mconfig` run.

jmstover:
> We don't need to check `with_suid` in makefile fragment b/c if that fragment appears in the resulting Makefile that means `with_suid` _was_ set to true during `./mconfig` run.

`with_suid=1` by default
`with_suid=0` if `--without-suid` is set

jmstover (approved):

jscook2345:
I could see some existing scripts breaking that rely on the current behavior, but I'm willing to merge this and see what our users come back with.

Thanks for this fix @sashayakovtseva !

PR #4282, opened 2019-08-22
Improvements and bug fixes for #4179

### LINKED ISSUE #4281: 'pull' dose not respect umask
## Version of Singularity:

```
3.3
```

### Expected behavior

umask should be honored when pulling a image.

### Actual behavior

Does not honor umask, output image permission will be 777.

 - Related to: https://github.com/sylabs/singularity/pull/4179

```
$ singularity pull alpine
```

WestleyK:

This PR is a follow up for https://github.com/sylabs/singularity/pull/4179.

 - Fixes #4281
 - Related to #4187
 - Related to #4179

<br>

### DIFF
--- a/internal/app/singularity/registry.go
+++ b/internal/app/singularity/registry.go
@@ -122,7 +122,6 @@ func (l *Library) pullAndVerify(ctx context.Context, imgMeta *scs.Image, from, t
 func (l *Library) copyFromCache(hash, name, to string) error {
 	exists, err := l.cache.LibraryImageExists(hash, name)
 	if err == cache.ErrBadChecksum {
-		sylog.Debugf("cache hash does not match, removing old cache...")
 		sylog.Warningf("Removing cached image: %s: cache could be corrupted", name)
 		err := os.Remove(l.cache.LibraryImage(hash, name))
 		if err != nil {
--- a/internal/app/singularity/registry.go
+++ b/internal/app/singularity/registry.go
@@ -1,8 +1,14 @@
+// Copyright (c) 2019, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
 package singularity
 
 import (
 	"context"
 	"fmt"
+	"os"
 
 	"github.com/sylabs/scs-library-client/client"
 	scs "github.com/sylabs/scs-library-client/client"
@@ -115,17 +121,25 @@ func (l *Library) pullAndVerify(ctx context.Context, imgMeta *scs.Image, from, t
 // If so, the image will be copied to the location provided.
 func (l *Library) copyFromCache(hash, name, to string) error {
 	exists, err := l.cache.LibraryImageExists(hash, name)
-	if err != nil {
+	if err == cache.ErrBadChecksum {
+		sylog.Debugf("cache hash does not match, removing old cache...")
+		sylog.Warningf("Removing cached image: %s: cache could be corrupted", name)
+		err := os.Remove(l.cache.LibraryImage(hash, name))
+		if err != nil {
+			return fmt.Errorf("unable to remove corrupted image from cache: %s", err)
+		}
+	} else if err != nil {
 		return fmt.Errorf("unable to check if %s exists: %v", name, err)
 	}
+
 	if !exists {
 		return errNotInCache
 	}
 
 	from := l.cache.LibraryImage(hash, name)
-	// Perms are 777 *prior* to umask in order to allow image to be
+	// Perms are 755 *prior* to umask in order to allow image to be
 	// executed with its leading shebang like a script
-	err = fs.CopyFile(from, to, 0777)
+	err = fs.CopyFile(from, to, 0755)
 	if err != nil {
 		return fmt.Errorf("while copying image from cache: %v", err)
 	}
--- a/internal/pkg/util/fs/helper.go
+++ b/internal/pkg/util/fs/helper.go
@@ -176,23 +176,33 @@ func MakeTmpFile(basedir, pattern string, mode os.FileMode) (*os.File, error) {
 	return f, nil
 }
 
-// CopyFile copies file to the provided location making sure the resulting
-// file has permission bits set to the mode.
-func CopyFile(from, to string, mode os.FileMode) (err error) {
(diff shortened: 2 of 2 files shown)


internal/pkg/util/fs/helper.go
```diff
@@ -176,23 +176,33 @@ func MakeTmpFile(basedir, pattern string, mode os.FileMode) (*os.File, error) {
 	return f, nil
 }
 
-// CopyFile copies file to the provided location making sure the resulting
-// file has permission bits set to the mode.
-func CopyFile(from, to string, mode os.FileMode) (err error) {
-	dstFile, err := MakeTmpFile(filepath.Dir(to), "", mode)
+// FileExists simply checks if a path exists.
+func FileExists(path string) (bool, error) {
```
gvallee, line 180: Should we keep this one and remove the one from `e2e/internal/e2e/fileutil.go`? I can see e2e using normal packages but not normal packages depending on code from e2e. Maybe actually move both `PathExists()` and `FileExists()`. What do you think? If you agree, maybe open an issue and do it in a different PR to make it easier to review this one?

gvallee: @WestleyK and I chatted and we both think we should move the code from e2e to `helper.go` but in a separate PR, until then @WestleyK is opening an issue.

gvallee (approved):
LGTM. After talking to @WestleyK please do not merge until @sashayakovtseva has a chance to review.

internal/app/singularity/registry.go
```diff
@@ -115,17 +121,25 @@ func (l *Library) pullAndVerify(ctx context.Context, imgMeta *scs.Image, from, t
 // If so, the image will be copied to the location provided.
 func (l *Library) copyFromCache(hash, name, to string) error {
 	exists, err := l.cache.LibraryImageExists(hash, name)
-	if err != nil {
+	if err == cache.ErrBadChecksum {
+		sylog.Debugf("cache hash does not match, removing old cache...")
```
sashayakovtseva: What's the point of two log entries (debug and warning) here? 
If log level is debug, then a warning will be displayed as well, I think this line should be removed.

internal/pkg/util/fs/helper.go
```diff
@@ -176,23 +176,33 @@ func MakeTmpFile(basedir, pattern string, mode os.FileMode) (*os.File, error) {
...
-	defer func() {
-		dstFile.Close()
-		if err == nil {
-			if e := os.Rename(dstFile.Name(), to); e != nil {
-				err = fmt.Errorf("could not rename temp file: %v", err)
-			}
-		}
-		os.Remove(dstFile.Name())
-	}()
+	dstFile, err := os.OpenFile(to, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, mode)
```
sashayakovtseva, line 202: This is wrong, if `io.Copy` fails you will have corrupted file here. 
The creation of temp file and defer was added to handle this particular case, please return that code back.
See https://github.com/sylabs/singularity/pull/4179#discussion_r313509655

internal/pkg/util/fs/helper.go
```diff
@@ -176,23 +176,33 @@ func MakeTmpFile(basedir, pattern string, mode os.FileMode) (*os.File, error) {
...
+		return false, err
+	}
+
+	return true, nil
+}
+
+// CopyFile copies file to the provided location. To honor umask
+// correctly, the to file must not exist.
+func CopyFile(from, to string, mode os.FileMode) error {
+	exist, err := FileExists(to)
```
sashayakovtseva, line 194: What if file is there with correct mode? 
I think this check should be performed in the calling function, rather than here.

sashayakovtseva (changes requested):

internal/app/singularity/registry.go
WestleyK: I added it just to have a little more verbosity, but I'll remove it.

internal/pkg/util/fs/helper.go
WestleyK: This is ~wrong~ **correct**, as it will honor umask. The issue with the tmp file, is that the `os.Chmod()` will override umask. There is a way to have have them both, which we can talk about.

internal/pkg/util/fs/helper.go
WestleyK: It is, but also double checked here. The code should never get here with `to` existing. But I can remove it if you want?

internal/pkg/util/fs/helper.go
sashayakovtseva: What do you mean by `The issue with the tmp file, is that the os.Chmod() will override umask.`? Do . you mean file permissions here (because that is what `mode` is)? The same permissions are applied during temp file creation,  and afaik `os.Rename` which is `mv` for Linux does not change them.

WestleyK: os.Rename() is not the issue. The issue is that when doing the `os.Chmod()`, that will override the umask. So the solution is to create the tmp file with `os.OpenFile(...)`.

internal/pkg/util/fs/helper.go
```diff
@@ -202,6 +212,7 @@ func CopyFile(from, to string, mode os.FileMode) (err error) {
 
 	_, err = io.Copy(dstFile, srcFile)
 	if err != nil {
+		os.Remove(to)
```
sashayakovtseva: I suggest you to keep defer func and use named return parameters as discussed. 
You need to close dstFile prior to removing it.

internal/pkg/util/fs/helper.go
```diff
@@ -176,22 +176,37 @@ func MakeTmpFile(basedir, pattern string, mode os.FileMode) (*os.File, error) {
...
-		return fmt.Errorf("could not create temp file: %v", err)
+		return err
+	}
+	if exist {
+		return fmt.Errorf("file %s already exists", to)
 	}
 
+	dstFile, err := os.OpenFile(to, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, mode)
+	if err != nil {
+		return fmt.Errorf("could not open file: %s: %s", to, err)
```
sashayakovtseva:
```suggestion
		return fmt.Errorf("could not open %s: %v", to, err)
```

sashayakovtseva (approved):

ikaneshiro (approved):

PR #4284, opened 2019-08-22
search: Print the container ID if theres no tag

### LINKED ISSUE #3877: Pulling image without latest tag fails
## Version of Singularity:

3.3

### Expected behavior

Options are to download the container if there's only one, but @mem points out problems with that in sylabs/cloud-library#612

Also suggested in that thread is to have better error output telling the user there's no image with the `latest` tag and to present the list of images

### Actual behavior

```
[rherban@test]:~$ singularity search simple
Found 1 containers for 'simple'
	library://rherban/default/simple.sif
		Tags:

[rherban@test]:~$ singularity pull library://rherban/default/simple.sif
FATAL:   While pulling library image: image does not exist in the library: rherban/default/simple.sif:latest
```

At this point, it's not obvious why it's not working or what we have to do to make it work.

WestleyK:

When searching for a container, if that container has no tag, this will now print the image ID, which you can use just like a tag.

#### Example

```bash
$ singularity search simple.sif
No users found for 'simple.sif'

No collections found for 'simple.sif'

Found 1 containers for 'simple.sif'
	library://rherban/default/simple.sif
		Image ID: [5cd9efccb0c9d75347e15e13] (no tag)

# Without a tag/image ID
$ singularity pull library://rherban/default/simple.sif
FATAL:   While pulling library image: image does not exist in the library: //rherban/default/simple.sif:latest (amd64)

# Now with the image ID as the tag
$ singularity pull library://rherban/default/simple.sif:5cd9efccb0c9d75347e15e13
```

<br>

**NOTE:** The image ID is not the sha/hash. But its shorter (has less characters) and seems to work. :smile: 

Please let @EmmEff, and or @tri-adam have a look before merge.

- Fixes #3877 

<br>


EmmEff (approved):
lgtm

sashayakovtseva:
Hm, image ID is used internally to refer the image in a database? If so, I'd prefer to have SHA sums as that is what users see in cloud library, while they don't see image IDs. That will prevent any further confusion.
Also why introduce another image reference type if we already have two (tag, SHA) and SHA is a perfect fit here, even though ID is shorter. Just my opinion.

WestleyK:
@sashayakovtseva, I definitely agree! But unfortunately the sha is not getting returned when doing a search, _like for `simple` https://library.sylabs.io/v1/search?value=simple._ Maybe thats something we can add? @EmmEff? :wink:

rherban:
Can we get additional text added to a failed pull?  Its still not obvious what to do from here:

```
# Without a tag/image ID
$ singularity pull library://rherban/default/simple.sif
FATAL:   While pulling library image: image does not exist in the library: //rherban/default/simple.sif:latest (amd64)
```

Maybe a:

```
This container matches, but has is untagged.  Pull with simple.sif:imageid
	library://rherban/default/simple.sif
		Image ID: [5cd9efccb0c9d75347e15e13] (no tag)
```

I'm open to other ideas on how to present this, but at the moment the website just says `singularity pull library://rherban/default/simple.sif` and that results in a dead-end.

PR #4291, opened 2019-08-22
Small tweaks for encryption CLI

ikaneshiro:

### DIFF
--- a/cmd/internal/cli/action_flags.go
+++ b/cmd/internal/cli/action_flags.go
@@ -312,7 +312,6 @@ var actionPEMPathFlag = cmdline.Flag{
 	DefaultValue: "",
 	Name:         "pem-path",
 	Usage:        "Enter an path to a PEM formated RSA key for an encrypted container",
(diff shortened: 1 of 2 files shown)


ikaneshiro:
@GodloveD Can you test this with the documentation you're preparing to make sure everything is consistent?

gvallee:
And just to make sure we are all on the same page: one of the goals of this PR is to have an early failure if users specify an empty path for a PEM file, instead of, for example, failing after starting to build an image.

ikaneshiro:
@gvallee This does not do that, I was going to do that in another one. This only targets the order of flag precedence and removes the requirement for both the `--encrypt` and `--pem-path/--passphrase` flags to be present. I should have added a better description to avoid confusion

gvallee:
@ikaneshiro My bad, it is what happens when I try to be too smart without really reading the code. :)

GodloveD (approved):
This is behaving as expected from my end.  Thank you @ikaneshiro !

PR #4297, opened 2019-08-22
Fail if trying to push to a unsupported transport type

### LINKED ISSUE #4296: push to unsupported destination should raise an error
## Version of Singularity:

3.3.0

### Expected behavior

Attempts to `singularity push` to an unsupported desination should error...

e.g.

```
$ singularity push test.sif shub://test/test
[ERROR] Push to shub:// is not supported
```

### Actual behavior

There is no error, and no action is taken. Return code is 0

```
[rherban@test]:~$ singularity push alpine_latest.sif shub://rherban/alpine
[rherban@test]:~$ echo $?
0
```

Push to a destination of any protocol except `library://` or `oras://`

These are the only protocols handled by the switch statement here:

https://github.com/sylabs/singularity/blob/dbb333ff6e2ec9cbaac77cf6b9a668d6c5caf97c/cmd/internal/cli/push.go#L74

.. which does not have a default / fall-through path with an error.

WestleyK:

This PR fixes the issue when trying to push a image to a unsupported transport.

- Fixes #4296

<br>

### DIFF
--- a/cmd/internal/cli/push.go
+++ b/cmd/internal/cli/push.go
@@ -94,6 +94,8 @@ var PushCmd = &cobra.Command{
 				sylog.Fatalf("Unable to push image to oci registry: %v", err)
 			}
 			sylog.Infof("Upload complete")
+		default:
+			sylog.Fatalf("Unsupported transport type: %s", transport)
 		}
 	},
 


ikaneshiro (approved):
Thanks for the quick fix

PR #4305, opened 2019-08-23
Document attribute constructor in starter.c

### LINKED ISSUE #3387: Document usage of __attribute__((constructor)) function in starter.c
Currently, `cmd/starter/c/starter.c` includes an `__attribute__((constructor))` function which is the first entrypoint into the `starter` and `starter-suid` binaries. We should clarify with code comments our reasoning for this methodology. It would be good to document in _at least_ these locations:

- `cmd/starter/c/starter.c` above the function declaration
- `internal/app/starter` in the package comment we should clarify in what context the go runtime will be initialized in which stage of the runtime
- `cmd/starter/main_linux.go` the package comment should contain the same contextualization as `internal/app/starter`

sashayakovtseva:
Closes #3387


cclerget (approved):
LGTM, thanks !

PR #4306, opened 2019-08-23
Store image labels in its own partition

### LINKED ISSUE #2790: Singularity not honoring json specification for inspect
Singularity inspect should honor the json specification [here](https://jsonapi.org/format/), specifically, we have "data" at the top level, and then attributes below it (and the majority of metadata below to attributes). On a high level, it should look like this:

```
$ singularity inspect -e -d -l -r -H -t --json /tmp/tmpkbq7z2nb/busybox:1.30.1.simg
{
       "data": {
   	        "attributes": {
		    "deffile": "bootstrap: docker\nfrom: busybox:1.30.1",
		    "environment": "...",
		    "labels": "...",
   		    "runscript": "..."
               }
	},
	"type": "container"
}
```

However what inspect is returning skips the "data" level, and only returns "attributes"

```
$ singularity inspect -e -d -l -r -H -t --json /tmp/tmpkbq7z2nb/busybox:1.30.1.simg
...
{
	"attributes": {
		"deffile": "bootstrap: docker\nfrom: busybox:1.30.1",
		"environment": "==-z\n==\n#!/bin/sh\n#Custom environment shell code should follow",
		"labels": "{\n\t\"org.label-schema.build-date\": \"Saturday_2_March_2019_14:59:59_EST\",\n\t\"org.label-schema.schema-version\": \"1.0\",\n\t\"org.label-schema.usage.singularity.deffile.bootstrap\": \"docker\",\n\t\"org.label-schema.usage.singularity.deffile.from\": \"busybox:1.30.1\",\n\t\"org.label-schema.usage.singularity.version\": \"3.1.0-rc2.28.ga72e427\"\n}",
		"runscript": "#!/bin/sh\nOCI_ENTRYPOINT=''\nOCI_CMD='\"sh\"'\nCMDLINE_ARGS=\"\"\n# prepare command line arguments for evaluation\nfor arg in \"$@\"; do\n    CMDLINE_ARGS=\"${CMDLINE_ARGS} \\\"$arg\\\"\"\ndone\n\n# ENTRYPOINT only - run entrypoint plus args\nif [ -z \"$OCI_CMD\" ] \u0026\u0026 [ -n \"$OCI_ENTRYPOINT\" ]; then\n    if [ $# -gt 0 ]; then\n        SINGULARITY_OCI_RUN=\"${OCI_ENTRYPOINT} ${CMDLINE_ARGS}\"\n    else\n        SINGULARITY_OCI_RUN=\"${OCI_ENTRYPOINT}\"\n    fi\nfi\n\n# CMD only - run CMD or override with args\nif [ -n \"$OCI_CMD\" ] \u0026\u0026 [ -z \"$OCI_ENTRYPOINT\" ]; then\n    if [ $# -gt 0 ]; then\n        SINGULARITY_OCI_RUN=\"${CMDLINE_ARGS}\"\n    else\n        SINGULARITY_OCI_RUN=\"${OCI_CMD}\"\n    fi\nfi\n\n# ENTRYPOINT and CMD - run ENTRYPOINT with CMD as default args\n# override with user provided args\nif [ $# -gt 0 ]; then\n    SINGULARITY_OCI_RUN=\"${OCI_ENTRYPOINT} ${CMDLINE_ARGS}\"\nelse\n    SINGULARITY_OCI_RUN=\"${OCI_ENTRYPOINT} ${OCI_CMD}\"\nfi\n\n# Evaluate shell expressions first and set arguments accordingly,\n# then execute final command as first container process\neval \"set ${SINGULARITY_OCI_RUN}\"\nexec \"$@\""
	},
	"type": "container"
}
```

## Version of Singularity:

```bash
$ singularity version 3.1.0-rc2.28.ga72e427
```

### Expected behavior

The original inspect was designed to follow the json API, and the updated GoLang version should be consistent with that.

### Actual behavior

Something took a bite out of the result and the "data" label is missing.

Inspect a container with 3.* family.

### LINKED ISSUE #3050: improve error message when singularity inspect tries to inspect a bad file
## Version of Singularity:

3.1.0

* Inspect a text file
<pre>
[justin@ip-172-31-94-96 inspect]$ singularity inspect ~/2019-03-19/singularity/help/singularity.txt
FATAL:   exit status 255
</pre>

* Inspect a directory
<pre>
[justin@ip-172-31-94-96 inspect]$ singularity inspect ~/2019-03-19/singularity/help
FATAL:   exit status 255
</pre>

Should have some sort of message here

### LINKED ISSUE #3932: Inspection metadata no longer containers container size
Singularity 2.X versions used to provide the container size as part of metadata with the field `org.label-schema.build-size`, and this is no longer done. Is there a reason, or can it be added back? Thanks!

### LINKED ISSUE #4279: Image metadata should be stored in its own partition
## Version of Singularity:

```
v3.3
```

### Expected behavior

The container information (like labels) should be stored it its own partition, not just some files in `.singularity.d/`.

### Actual behavior

Right now, container information is stored in some file.

 - Related to: https://github.com/sylabs/singularity/issues/3932

Write here.

### LINKED ISSUE #4310: Labels should show what architecture the image is
## Version of Singularity:

```
v3.3
```

### Expected behavior

There should be a label the shows what architecture the container is.

```
$ singularity inspect foo.sif
...
org.label-schema.image-arch: "amd64"
...
```

This is only useful when the metadata is stored in its own partition. Otherwise, you can only inspect the image if the image, and the host are the same architecture.

### Actual behavior

Does not exist yet...

Write here.

WestleyK:

When building a container, the image metadata will now be stored in its own partition.

Things changed:
 1. Deffile data is parsed from the deffile partition when requested to inspect that section (or whole deffile).
 2. Only labels are stored in the metadata partition.
 3. Will only add the metadata partition if we are building a sif image. (not sandbox)
 4. Inspect is a lot faster! 10x faster on my system! From alpine to centos, its 10x faster!
 5. Image can be inspected if the hosts architecture and the image architecture does not match. _ie. you can inspect a arm32 image on your x86_64 machine!_

 - Addresses: #4279
 - Fixes: #3932
 - Fixes: #4310 
 - Fixes: #3050
 - Fixes: #2790

<br>


cmd/internal/cli/inspect_linux.go
```diff
@@ -264,127 +348,232 @@ var InspectCmd = &cobra.Command{
...
+					fmt.Println("==helpfile==\n" + inspectObj.Attributes.Helpfile)
+				}
+				if inspectObj.Attributes.Deffile != "" {
+					fmt.Println("==deffile==\n" + inspectObj.Attributes.Deffile)
+				}
+				if inspectObj.Attributes.Runscript != "" {
+					fmt.Println("==runscript==\n" + inspectObj.Attributes.Runscript)
+				}
+				if inspectObj.Attributes.Test != "" {
+					fmt.Println("==test==\n" + inspectObj.Attributes.Test)
```
sashayakovtseva: why not using printf?

WestleyK: Because I forgot. It will use printf :+1:

internal/pkg/build/build.go
```diff
@@ -316,21 +322,250 @@ func (b *Build) Full() error {
...
-		if err := stage.insertMetadata(); err != nil {
-			return fmt.Errorf("while inserting metadata to bundle: %v", err)
+		//		sylog.Debugf("Inserting Metadata")
+		//	if err := stage.insertMetadata(); err != nil {
+		//		return fmt.Errorf("while inserting metadata to bundle: %v", err)
+		//	}
+
+		//labels["contianer-info"] = "contianer-info"
+		labels["contianer-info"] = make(map[string]string, 1)
+		labels["contianer-info"]["org.label-schema.schema-version"] = "2.0"
```
dtrudg: typo here and in places below... `labels["contianer-info"]`

dtrudg:
Hi @WestleyK - conceptually I really like the idea of having metadata out in a separate part of the SIF. A well defined and easily accessible structure here seems much nicer for tools to deal with than pulling files from a rootfs.

This is a fairly serious change, so there are at least a few things to discuss and consider - some off the top of my head...

- When container file systems are encrypted then (most of the) metadata should likely be encrypted, or we are leaking information about the container that might be important to keep secret. Users may want an encrypted container to be completely opaque (no clue about content or purpose). There is a counter to this that it may also be useful for labels to be un-encrypted, for the case where it'd be useful to have limited public metadata about an encrypted container accessible and visible in the container library or similar.

- Are there tools out there other than Singularity that directly inspect the exsiting `.singularity.d` metadata? Users may be doing things with it and we could break their workflows if we removed it. Would we stage a change across a few releases, so we add the new encapsulation in `3.x` and remove `singularity.d` metadata in `3.x+1` or similar?

- Related to the above - how can we have `in-container introspection` i.e. it's trivial for a process inside a container to look at its own metadata in `.singularity.d`. How would this be possible if the metadata is in another SIF partition?

- How are we dealing with this for sandboxes and OCI formats? What is the flow for metadata going into and out of these folder structures? Would sandboxes keep a `.singularity.d` which is then removed if they are built into a SIF etc.?

- If the metadata goes into a partition at the end of a SIF, how easy would it be to e.g. only fetch the SIF header, and the metadata partition portions of the file? Having metadata out of the rootfs means that we could, in theory grab that metadata from a huge 10GB SIF without having to download the majority of the container's size, which is in the rootfs. Being able to do stuff like that is, to my mind, a major advantage of keeping metadata separately so would be nice to tackle.

jmstover:
If you add this, IMO, this should bump the SIF version. To me, this is enough of a change to bump the format version.

I guess in the CLI, we would look for `sif.Header.Version` to figure out how we do the inspection of metadata? >= 1.0.9 look at the header, otherwise we need to mount it.

@mem @bauerm97

ikaneshiro:
I agree that this actually ends up being a pretty serious technical change.

To add another point to @dctrud's list:
- Signatures will need to be changed to cover both the system partition and this new metadata partition since the metadata partition contains information that affects the runtime behavior of the container




To @jmstover's point:

What was our criteria for a new version of singularity? It was my understanding that it was whenever the image format was completely incompatible with other versions. Making such a change will cause any build to no longer be usable by v3.4.x and less. We could bump up the SIF header version to make it clear to older runtimes that this is a format they do not support, but this doesn't stem from a change in SIF but how singularity uses SIF. So I think that point is worth more discussion.

jmstover:
> but this doesn't stem from a change in SIF but how singularity uses SIF. So I think that point is worth more discussion.

Correct here. This is just setting aside a part of the header, and going "You're being used explicitly for this."

Once it becomes the _only_ way for metadata, then old binaries are out of luck on inspect on a newly created image. This is where I never liked the dying because of SIF version in Singularity specifically. There isn't a *technical* reason why a old binary couldn't run the image (in this case). But because of how it's being used, we're now saying that an image (the squashfs portion) is no longer the same format once there's no longer metadata in `/.singularity.d`.

WestleyK:
Yes, this will be a problem with sandbox containers. We will haft to revert back to using a `labels.json` in `/.singularity.d`.

Also, I'm already thinking on how we can do multi-partition signing, but that is not developed yet.

Regarding encrypted containers, the user may not want to encrypt the metadata. After all, the deffile partition is not encrypted, which that could be more important to encrypt then the labels.

@dctrud @ikaneshiro ^^^

dtrudg:
> Yes, this will be a problem with sandbox containers. We will haft to revert back to using a `labels.json` in `/.singularity.d`.
> 
> Also, I'm already thinking on how we can do multi-partition signing, but that is not developed yet.
> 
> Regarding encrypted containers, the user may not want to encrypt the metadata. After all, the deffile partition is not encrypted, which that could be more important to encrypt then the labels.
> 
> @dctrud @ikaneshiro ^^^

dtrudg:
Sorry for the close.... I clicked the wrong button :man_facepalming: 

> > Regarding encrypted containers, the user may not want to encrypt the metadata. After all, the deffile partition is not encrypted, which that could be more important to encrypt then the labels.

This is certainly something to also think about.... it may e.g. be bad practice to embed secrets in a def file... but I assume it does happen.

ikaneshiro:
I think it makes the most sense to just drop the definition partition when building an encrypted container instead of having it out in the open. Definitely not recommended to store secrets in definitions, but it's better to default to the safest option imo

WestleyK:
> Sorry for the close.... I clicked the wrong button 

No worries, been there, done that :smile: 

Lets move the encrypted deffile partition conversation to https://github.com/sylabs/singularity/issues/4316.

WestleyK:
@dctrud, regarding:

> * Are there tools out there other than Singularity that directly inspect the exsiting `.singularity.d` metadata? Users may be doing things with it and we could break their workflows if we removed it. Would we stage a change across a few releases, so we add the new encapsulation in `3.x` and remove `singularity.d` metadata in `3.x+1` or similar?

How about a flag to only use `/.singularity.d/labels.json`? Like `singularity build --no-metadata-partition foo.sif test.def`?

Also, we can abandon the `/singularity.d/labels.json`, because it is still required for sandbox images.

cmd/internal/cli/inspect_linux.go
```diff
@@ -264,132 +324,259 @@ var InspectCmd = &cobra.Command{
...
-		inspectObj := inspectFormat{}
-		inspectObj.Type = "container"
-		inspectObj.Attributes.Labels = make(map[string]string)
-		inspectObj.Attributes.Environment = make(map[string]string)
-
-		// Parse the command output string into sections.
-		reader := bufio.NewReader(strings.NewReader(fileContents))
-		for {
-			section, err := reader.ReadBytes('\n')
+			sifData, _, err := getMetaData(&fimg, sif.DataDeffile)
```
ikaneshiro: I think we should still run the container if we want the environment. We cant only rely on the definition in this case

cmd/internal/cli/inspect_linux.go
```diff
@@ -264,132 +324,259 @@ var InspectCmd = &cobra.Command{
...
+			for _, v := range sifData {
+				metaData := v.GetData(&fimg)
+				data := parseDeffile(string(metaData), "%help")
+				inspectDataJSON["container"]["helpfile"] = data
+				inspectData += "== helpfile ==\n" + data
+			}
 		}
 
+		// For %runscript
 		if runscript {
```
ikaneshiro: Solely relying on the runscript from the definition will cause a change for images build OCI images that contain a default entrypoint/command since that is inserted to the runscript inside the container silently during the build to facilitate this functionality. This is something that is obfuscated from the user anyway so I don't think this is much of an issue. Just commenting here so it is considered.

cmd/internal/cli/inspect_linux.go
```diff
@@ -264,132 +324,259 @@ var InspectCmd = &cobra.Command{
...
+				data := parseDeffile(string(metaData), "%environment")
+				inspectDataJSON["container"]["environment"] = data
+				inspectData += "== environment ==\n" + data
 			}
 		}
 
-		// Output the inspection results (use JSON if requested).
 		if jsonfmt {
-			jsonObj, err := json.MarshalIndent(inspectObj, "", "\t")
+			text, err := json.MarshalIndent(inspectDataJSON, "", "    ")
```
ikaneshiro: Why change to spaces instead of a tab?

internal/pkg/build/build.go
```diff
@@ -316,9 +323,68 @@ func (b *Build) Full() error {
...
-		sylog.Debugf("Inserting Metadata")
-		if err := stage.insertMetadata(); err != nil {
-			return fmt.Errorf("while inserting metadata to bundle: %v", err)
+		//		sylog.Debugf("Inserting Metadata")
+		//	if err := stage.insertMetadata(); err != nil {
+		//		return fmt.Errorf("while inserting metadata to bundle: %v", err)
+		//	}
+
+		// Only add the metadata partition if we are building a sif
+		if b.Conf.Format == "sif" {
```
ikaneshiro: Lets move this metadata insertion to `assembler_sif/sandbox.go` since they seem like a better fit.

ikaneshiro (changes requested):

cmd/internal/cli/inspect_linux.go
ikaneshiro: Marking resolved because I misread how it's currently working

cmd/internal/cli/inspect_linux.go
WestleyK: Just trying it out, do you prefer tabs? _I'm okay with either._

cmd/internal/cli/inspect_linux.go
WestleyK: I think its okay if we just list the runscript from the deffile, we can add/change the help for `inspect` to something like: 

```
--runscript    show the runscript from the deffile
```

Also, I'm trying to avoid running/shelling into the container for inspect, so we can inspect a container on a different architecture, or a encrypted container.

sashayakovtseva (changes requested):
1. Please, cleanup this PR. Commented code should be either removed, or uncommented. 
2. Why you renamed `cmd/internal/cli/inspect_linux.go` → `cmd/internal/cli/inspect.go`?

WestleyK:
@sashayakovtseva,

> 1. Please, cleanup this PR. Commented code should be either removed, or uncommented.

WIP, will clean it up some more.

> 2. Why you renamed `cmd/internal/cli/inspect_linux.go` → `cmd/internal/cli/inspect.go`?

Since `inspect` no longer `shell`/`exec` and container to get the labels/metadata (as of this PR), you can `inspect` a container thats built on a different architecture. _eg. you can `inspect` a arm32 container on a x86_64 machine._

sashayakovtseva:
As far as I know, linux is responsible for GOOS setting and has nothing to do about arch...

internal/pkg/build/assemblers/assembler_sif.go
```diff
@@ -215,9 +217,79 @@ func (a *SIFAssembler) Assemble(b *types.Bundle, path string) error {
...
+	// add new signature data object to SIF file
+	err := fimg.AddObject(siginput)
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
 
+func getDescr(fimg *sif.FileImage) ([]*sif.Descriptor, error) {
```
ikaneshiro, line 300: Is this helper really needed? It's just wrapping another function

internal/pkg/build/assemblers/assembler_sif.go
```diff
@@ -215,9 +217,79 @@ func (a *SIFAssembler) Assemble(b *types.Bundle, path string) error {
 		return fmt.Errorf("while creating SIF: %v", err)
 	}
 
+	labels := make(map[string]string, 1)
```
ikaneshiro: Why not have this as a part of `createSIF()`?

internal/pkg/build/sources/packer_sif_linux.go
```diff
@@ -45,6 +70,30 @@ func (p *SIFPacker) unpackSIF(b *types.Bundle, srcfile string) (err error) {
...
+	}
+	defer fimg.UnloadContainer()
+
+	sifData, _, err := getMetaData(&fimg, sif.DataLabels)
+	if err == ErrNoMetaData {
+		sylog.Warningf("No metadata partition found")
+	} else if err != nil {
+		sylog.Fatalf("Unable to get label metadata: %s", err)
+	} else if err == nil {
+		b.Recipe.ImageData.Labels = make(map[string]string, 1)
```
ikaneshiro: Lets not store this within the recipe, rather the `bundle.JSONObjects` map, this looks like it will overwrite labels that exist within a definition right?

ikaneshiro (changes requested):

cmd/internal/cli/inspect_linux.go
ikaneshiro: Doesn't matter, just wondered if there was a reason

ikaneshiro:
@WestleyK @sashayakovtseva WRT the `_linux` build tag, I think we should ensure that inspect works on `darwin` before removing it.

internal/pkg/build/assemblers/assembler_sif.go
WestleyK: Yes, I think it is needed. I will put in a common package with a better func name.

WestleyK:
> @WestleyK @sashayakovtseva WRT the `_linux` build tag, I think we should ensure that inspect works on `darwin` before removing it.

As long as the container has the label partition, then yes. It will work. Not just on MacOS, but also arm32, etc...

internal/pkg/build/sources/packer_sif_linux.go
WestleyK: Actually, the labels from the deffile will override these labels. Also, I tried to use the `JSONObjects` map, but I had a hard time getting that to work...

WestleyK:
Closing as this changes too much code/behavior in this PR -> https://github.com/sylabs/singularity/pull/4399

PR #4325, opened 2019-08-28
Fix cli to use pem path envvar

ikaneshiro:
Fixes issue where CLI will only use --pem-path flag variable instead of
envvar variable

**This fixes or addresses the following issue:**

```
$ sudo SINGULARITY_ENCRYPTION_PEM_PATH=ssh.pub singularity build -e test.sif library://alpine
FATAL:   Specified PEM file : does not exist.
```
to be 

```
sudo SINGULARITY_ENCRYPTION_PEM_PATH=ssh.pub singularity build -e test.sif library://alpine
INFO:    Starting build...
INFO:    Creating SIF file...
```

Cc: @GodloveD


gvallee (approved):
Tested and it fixes the problem. Many thanks!

PR #4327, opened 2019-08-28
Fix cli to use pem path envvar

GodloveD:
Fixes issue where CLI will only use --pem-path flag variable instead of
envvar variable


gvallee (approved):
Successfully tested manually

PR #4348, opened 2019-08-30
Delete image from remote library

### LINKED ISSUE #3473: remove container from library via CLI
## Version of Singularity:

3.2.0-rc1.204.gb50f8b2

### Expected behavior
```
$ singularity delete library://some/thing/misguided
Are you sure you want to delete library://some/thing/misguided? [y/N] y
Deleting library://some/thing/misguided... done!

$ singularity delete --force library://some/thing/else
Deleting library://some/thing/else... done!
```
### Actual behavior

Only exists in my fevered imagination.  

Try it... and you will see... NOTHING!!

pisarukv:
closes https://github.com/sylabs/singularity/issues/3473


cmd/internal/cli/delete.go
```diff
@@ -0,0 +1,111 @@
...
+	DefaultValue: "",
+	Name:         "arch",
+	ShortHand:    "A",
+	Required:     true,
+	Usage:        "specify requested image arch",
+	EnvKeys:      []string{"ARCH"},
+}
+
+var deleteImageTimeout int
+var deleteImageTimeoutFlag = cmdline.Flag{
```
WestleyK, line 43: Is this something that should be changed with a flag? (I think thats okay) but I do suggest making the flag hidden. Because this is something that should not be needing to be changed often (or at all). Afaik. 

I honestly think a flag like this would also be useful for the `push`.

cmd/internal/cli/delete.go
```diff
@@ -0,0 +1,111 @@
...
+		handleDeleteFlags(cmd)
+
+		imageRef := strings.TrimPrefix(args[0], "library://")
+
+		libraryConfig := &client.Config{
+			BaseURL:   deleteLibraryURI,
+			AuthToken: authToken,
+		}
+
+		y, err := interactive.AskYNQuestion("n", fmt.Sprintf("Are you sure you want to delete %s arch[%s] [Y,n]\n", imageRef, deleteImageArch))
```
WestleyK: Is there a force flag to override this? If not, thats probably okay; we dont want users to accidentally delete a image.

Also, you can skip the `Sprintf`:

```suggestion
		y, err := interactive.AskYNQuestion("n", "Are you sure you want to delete %s arch[%s] [Y,n]\n", imageRef, deleteImageArch)
```

cmd/internal/cli/delete.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+		if y == "n" {
+			return
+		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), time.Duration(deleteImageTimeout)*time.Second)
+		defer cancel()
+		err = singularity.DeleteImage(ctx, libraryConfig, imageRef, deleteImageArch)
+		if err != nil {
+			sylog.Fatalf(err.Error())
```
WestleyK: Do you want to warp these error messages? Maybe like:

```suggestion
			sylog.Fatalf("Unable to delete image from library: %s\n", err)
```

docs/content.go
```diff
@@ -325,6 +325,16 @@ Enterprise Performance Computing (EPC)`
...
 
+	// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+	// delete
+	// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+	DeleteUse   string = `delete [arch] <imageRef>`
+	DeleteShort string = `Deletes requested image from the library`
+	DeleteLong  string = `
+  The 'delete' command allows you to delete an image from a remote library.`
+	DeleteExample string = `
+  $ singularity delete --arch=amd64 library://sylabsed/examples/lolcow:1.0`
```
WestleyK: Do we want to use this image (`lolcow`) as the example for a user to try to delete it? Maybe it should be like:

```suggestion
  $ singularity delete --arch=amd64 library://username/project/image:1.0`
```

WestleyK (changes requested):
Looks good, just a couple suggestions...

cmd/internal/cli/delete.go
pisarukv: Nope, there is no flag for this :)

WestleyK: Okay, thats probably fine, imo.

cmd/internal/cli/delete.go
```diff
@@ -0,0 +1,111 @@
...
+
+var deleteImageTimeout int
+var deleteImageTimeoutFlag = cmdline.Flag{
+	ID:           "deleteImageTimeoutFlag",
+	Value:        &deleteImageTimeout,
+	DefaultValue: 15,
+	Name:         "timeout",
+	ShortHand:    "T",
+	Hidden:       true,
+	Usage:        "specify delete timeout in secods",
```
sashayakovtseva:
```suggestion
	Usage:        "specify delete timeout in seconds",
```

e2e/delete/delete.go
```diff
@@ -0,0 +1,67 @@
...
+			e2e.WithProfile(e2e.UserProfile),
+			e2e.WithCommand("delete"),
+			e2e.WithArgs(tt.args...),
+			e2e.WithStdin(bytes.NewBufferString(tt.agree)),
+			e2e.ExpectExit(tt.expectExit),
+		)
+	}
+}
+
+// E2ETests is the main func to trigger the test suite
```
sashayakovtseva:
```suggestion
// E2ETests is the main func to trigger the test suite.
```

sashayakovtseva (approved):

cmd/internal/cli/delete.go
```diff
@@ -0,0 +1,111 @@
...
+		handleDeleteFlags(cmd)
+
+		imageRef := strings.TrimPrefix(args[0], "library://")
+
+		libraryConfig := &client.Config{
+			BaseURL:   deleteLibraryURI,
+			AuthToken: authToken,
+		}
+
+		y, err := interactive.AskYNQuestion("n", "Are you sure you want to delete %s arch[%s] [Y,n]\n", imageRef, deleteImageArch)
```
WestleyK: Why print a newline here? Typical `y/n` questions are on the same line.

```suggestion
		y, err := interactive.AskYNQuestion("n", "Are you sure you want to delete %s arch[%s] [Y,n] ", imageRef, deleteImageArch)
```

And should the default be `n`? Instead of `y`?

WestleyK (commented):
One question, otherwise looks good.

cmd/internal/cli/delete.go
pisarukv: I would better leave n, since it's deleting operation. So we can prevent user occasionally delete an image.

cmd/internal/cli/delete.go
```diff
@@ -0,0 +1,111 @@
...
+		handleDeleteFlags(cmd)
+
+		imageRef := strings.TrimPrefix(args[0], "library://")
+
+		libraryConfig := &client.Config{
+			BaseURL:   deleteLibraryURI,
+			AuthToken: authToken,
+		}
+
+		y, err := interactive.AskYNQuestion("n", "Are you sure you want to delete %s arch[%s] [Y,n] ", imageRef, deleteImageArch)
```
WestleyK: Ohh, I see, the default is `n`. Then I suggest changing the question to show the user that the default is `n`

```suggestion
		y, err := interactive.AskYNQuestion("n", "Are you sure you want to delete %s arch[%s] [N/y] ", imageRef, deleteImageArch)
```

WestleyK (approved):
Looks good to me :+1:

ikaneshiro (approved):

PR #4363, opened 2019-09-03
Export many keys to one file

### LINKED ISSUE #3280: Support Multiple Exported Keys in a Single File. (binary keys only)
## Version of Singularity:

```
3.1.1-902.ge094ee448
```

### Expected behavior

You should be able to export multiple keys in a single file.

### Actual behavior

Singularity overrides any key that might be in the file.

```bash
$ singularity key export -s ./file
[...]
$ singularity key export -s ./file
[...]

$ mv ~/.singularity/sypgp/pgp-secret ~/.singularity/sypgp/pgp-secret.bck

$ singularity key import ./file
[...]

$ singularity key list -s
# you should see the two keys that you exported
```

pisarukv:
closes https://github.com/sylabs/singularity/issues/3280

### DIFF
--- a/pkg/sypgp/sypgp.go
+++ b/pkg/sypgp/sypgp.go
@@ -961,7 +961,7 @@ func (keyring *Handle) ExportPrivateKey(kpath string, armor bool) error {
 	}
 
 	// Create the file that we will be exporting to
-	file, err := os.Create(kpath)
+	file, err := os.OpenFile(kpath, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0600)
 	if err != nil {
 		return err
 	}
@@ -1003,7 +1003,7 @@ func (keyring *Handle) ExportPubKey(kpath string, armor bool) error {
 		return err
 	}
 
-	file, err := os.Create(kpath)
+	file, err := os.OpenFile(kpath, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0600)
 	if err != nil {
 		return fmt.Errorf("unable to create file: %v", err)
 	}
--- a/pkg/sypgp/sypgp.go
+++ b/pkg/sypgp/sypgp.go
@@ -961,7 +961,7 @@ func (keyring *Handle) ExportPrivateKey(kpath string, armor bool) error {
 	}
 
 	// Create the file that we will be exporting to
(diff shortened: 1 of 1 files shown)


WestleyK (changes requested):
Thanks for fixing the issue, but is this still something we still want to support?

 - First, this dose not work with ASCII keys (as expected but that means that ASCII key files, will behave differently when exporting them)
 - Second, this behaves weird with private keys; when importing it, it will ask for a password to import it, (as expected) but it does not tell you what key you are importing.
 -  Last, if you are importing a private key, and you already have that public key, that import will fail, and you cant import the second key in that file, until you remove your public key.

_I hope all of that makes sense, if not let me explain what I think it should be like, **IF** we want to support this_

```bash
# Export the first key
$ singularity key export -s ./bar
# Ask for the password and whatever...

# Export the second key
$ singularity key export -s ./bar
WARNING: export target already contains something; appending to file
# Ask for the password and exports the key 

# Importing that keyfile
$ singularity key import ./bar 
Enter your old password for XXXXXXXXXXXXXXXXXX
# <entered the password>
Enter a new password for this key : 
Retype your passphrase : 
Key with fingerprint 3742BC0172E4310034E34DDCB510DAC2A9EF882D successfully added to the private keyring
Key with fingerprint 3742BC0172E4310034E34DDCB510DAC2A9EF882D already exist in your public keyring

Enter your old password for XXXXXXXXXXXXXXXXXX
# <password for the second key>
Enter a new password for this key : 
Retype your passphrase : 
Key with fingerprint AFF8A4552FF8DA3DF89E11F4376DE194264DC22C successfully added to the private keyring
Key with fingerprint AFF8A4552FF8DA3DF89E11F4376DE194264DC22C successfully added to the public keyring
```

Thats how I think it should be like, if we want to support this. Besides, afaik gpg does not support multi-exported keys per file. (I could be wrong)

But, what I do think should change is that when importing a private key, it should not ask for a new password, unless you specify a flag, maybe like `--with-newpassword`. I'll open a issue. (https://github.com/sylabs/singularity/issues/4369)

pisarukv:
I'm not sure if we still want to support it. Actually, from my perspective, such behavior when I can export several keys in a single file could be misleading. It could be useful when I want to create a backup, but in this case I'll prefer to have a separate command for ex.`singularity keys backup`.

Issue #3764, opened 2019-06-20
sporadic issue with running singularity: loop device: Failed to set loop flags on loop device: resource temporarily unavailable


bbarker:
## Version of Singularity:

3.0.1

### Expected behavior

Being able to run singularity without occasionally getting failures. A "timeout" option to singularity might be useful, but perhaps there are some other parameters, like kernel flags or options, that could be tweaked.

### Actual behavior

```
$ singularity exec --containall   -H /tmp/.xsede-job-runner/589b9ed8-c31c-4b41-ae6f-4a7cd32dc9ff /home/ccrs/SING_IMAGES/vsoch-pokemon-master-latest.simg pwd
/tmp/.xsede-job-runner/589b9ed8-c31c-4b41-ae6f-4a7cd32dc9ff

$ singularity exec --containall   -H /tmp/.xsede-job-runner/589b9ed8-c31c-4b41-ae6f-4a7cd32dc9ff /home/ccrs/SING_IMAGES/vsoch-pokemon-master-latest.simg pwd
/tmp/.xsede-job-runner/589b9ed8-c31c-4b41-ae6f-4a7cd32dc9ff

$ singularity exec --containall   -H /tmp/.xsede-job-runner/589b9ed8-c31c-4b41-ae6f-4a7cd32dc9ff /home/ccrs/SING_IMAGES/vsoch-pokemon-master-latest.simg pwd
FATAL:   container creation failed: mount error: can't mount image /proc/self/fd/8: failed to find loop device: Failed to set loop flags on loop device: resource temporarily unavailable
```

### Steps to reproduce behavior

I don't recall seeing this issue in 2.6 but it was very clearly present in 3.0.1, which I just switched to. I also recently updated my kernel and many other system packages, so it isn't definitely the case that it is singularity 2.6 -> 3.0.1 that caused the issue.

cclerget:
Hi @bbarker , running Singularity version < 3.2 is not recommended, 3.0.X have many bugs, 3.1.X versions contains a severe security issue (https://github.com/sylabs/singularity/releases/tag/v3.2.0). I would suggest to install at least 3.2 version, ideally 3.2.1 for now, you can also give a try to 3.3.0-rc1

cclerget:
Closing, feel free to reopen it if needed

PR #4373, opened 2019-09-04
Fix ioctl errno handling for LOOP_SET_STATUS64

muffgaga:
Linux kernel commit [5db470e229e22b7eda6e23b5566e532c96fb5bc3](https://github.com/torvalds/linux/commit/5db470e229e22b7eda6e23b5566e532c96fb5bc3) ("_loop:
drop caches if offset or block_size are changed_") introduced a transient
fail (with `-EAGAIN`) possibility into the `LOOP_SET_STATUS64` `ioctl`.
This also affects longterm kernel versions 4.14 and 4.19.

Since we switched to a newer kernel version 4.19.37 (Debian Buster), we sporadically encountered errors such as:
```
container creation failed: mount error: can't mount image /proc/self/fd/29: failed to find loop device: could not attach image file to loop device: Failed to set loop flags on loop device: resource temporarily unavailable
```
As mentioned above, in recent kernel versions the syscall might fail with `errno == EAGAIN`; similar to what the [`losetup` tool](https://github.com/karelzak/util-linux/blob/dab1303287b7ebe30b57ccc78591070dad0a85ea/lib/loopdev.c#L1355) does, we have to retry when encountering EAGAIN.

- might be related to #3764

### DIFF
--- a/pkg/util/loop/loop_linux.go
+++ b/pkg/util/loop/loop_linux.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"os"
 	"syscall"
+	"time"
 	"unsafe"
 
 	"github.com/sylabs/singularity/pkg/util/fs/lock"
@@ -101,8 +102,18 @@ func (loop *Device) AttachFromFile(image *os.File, mode int, number *int) error 
 		return fmt.Errorf("failed to set close-on-exec on loop device %s: %s", path, err.Error())
 	}
 
-	if _, _, err := syscall.Syscall(syscall.SYS_IOCTL, uintptr(loopFd), CmdSetStatus64, uintptr(unsafe.Pointer(loop.Info))); err != 0 {
-		return fmt.Errorf("failed to set loop flags on loop device: %s", syscall.Errno(err))
+	maxRetries := 5
+	for i := 0; i < maxRetries; i++ {
+		if _, _, err := syscall.Syscall(syscall.SYS_IOCTL, uintptr(loopFd), CmdSetStatus64, uintptr(unsafe.Pointer(loop.Info))); err != 0 {
+			if err == syscall.EAGAIN && i < maxRetries-1 {
+				// loop_set_status() can temporarily fail with EAGAIN -> sleep and try again
+				// (cf. https://github.com/karelzak/util-linux/blob/dab1303287b7ebe30b57ccc78591070dad0a85ea/lib/loopdev.c#L1355)
+				time.Sleep(250*time.Millisecond)
+				continue
+			}
+			return fmt.Errorf("failed to set loop flags on loop device: %s", syscall.Errno(err))
+		}
+		break
 	}
 
 	return nil


sashayakovtseva:
Maybe some kind of attempts and timeouts should be added? E.g. try 5 times with small pauses.

sashayakovtseva:
Tests failed due to CI timeout.

cclerget:
Hi @muffgaga , thank you for fixing it ! At first I thought the fix could enter in infinite loop but ... I tried on a debian buster VM and looks like the fix works fine without loop break, however like @sashayakovtseva suggested it requires to add a pause between each retry otherwise it increases the chance to get another `EAGAIN` over and over as each call to SET_STATUS triggers a cache drop of the loop device, it needs to be less aggressive, the same timing as `losetup` does the job.

pkg/util/loop/loop_linux.go
```diff
@@ -101,8 +101,15 @@ func (loop *Device) AttachFromFile(image *os.File, mode int, number *int) error
...
 	}
 
-	if _, _, err := syscall.Syscall(syscall.SYS_IOCTL, uintptr(loopFd), CmdSetStatus64, uintptr(unsafe.Pointer(loop.Info))); err != 0 {
-		return fmt.Errorf("failed to set loop flags on loop device: %s", syscall.Errno(err))
+	for {
+		if _, _, err := syscall.Syscall(syscall.SYS_IOCTL, uintptr(loopFd), CmdSetStatus64, uintptr(unsafe.Pointer(loop.Info))); err != 0 {
+			if err == syscall.EAGAIN {
+				// loop_set_status() can temporarily fail with EAGAIN
+				// (cf. https://github.com/torvalds/linux/blob/d0a255e795ab976481565f6ac178314b34fbf891/drivers/block/loop.c#L1280)
+				continue
```
cclerget, line 112: Requires a pause: `time.Sleep(250000*time.Microsecond)`

cclerget (changes requested):

muffgaga:
Thanks for the comments.

* sorry, `break` was missing
* we now try up to 5 times
* we wait like `losetup` does → 25ms (@cclerget suggested 250ms?)

pkg/util/loop/loop_linux.go
```diff
@@ -101,8 +102,18 @@ func (loop *Device) AttachFromFile(image *os.File, mode int, number *int) error
 		return fmt.Errorf("failed to set close-on-exec on loop device %s: %s", path, err.Error())
 	}
 
-	if _, _, err := syscall.Syscall(syscall.SYS_IOCTL, uintptr(loopFd), CmdSetStatus64, uintptr(unsafe.Pointer(loop.Info))); err != 0 {
-		return fmt.Errorf("failed to set loop flags on loop device: %s", syscall.Errno(err))
+	maxRetries := 5
+	for i := 0; i < maxRetries; i++ {
+		if _, _, err := syscall.Syscall(syscall.SYS_IOCTL, uintptr(loopFd), CmdSetStatus64, uintptr(unsafe.Pointer(loop.Info))); err != 0 {
+			if err == syscall.EAGAIN {
```
cclerget:
```suggestion
			if err == syscall.EAGAIN && i < maxRetries-1 {
```

pkg/util/loop/loop_linux.go
```diff
@@ -101,8 +102,18 @@ func (loop *Device) AttachFromFile(image *os.File, mode int, number *int) error
...
+				// loop_set_status() can temporarily fail with EAGAIN -> sleep and try again
+				// (cf. https://github.com/karelzak/util-linux/blob/dab1303287b7ebe30b57ccc78591070dad0a85ea/lib/loopdev.c#L1355)
+				time.Sleep(25*time.Millisecond)
```
cclerget: @muffgaga Looks like `losetup` set a 250ms : `xusleep(250000)` but whatever I tested within a loop with 25 ms pause while stressing the VM and it works fine

cclerget (changes requested):
Just a last fix to trigger the error after `maxRetries` is reached, after that it's good to go. Thank you very much !

pkg/util/loop/loop_linux.go
muffgaga: Argh, sorry for that… fixing up

pkg/util/loop/loop_linux.go
muffgaga: Sorry again → I looked at some [other part](https://github.com/karelzak/util-linux/blob/dab1303287b7ebe30b57ccc78591070dad0a85ea/lib/loopdev.c#L1327) of losetup where 25ms was used.

cclerget: Really thank you very much for tackling this issue, very much appreciated. And please add yourself in https://github.com/sylabs/singularity/blob/master/CONTRIBUTORS.md :)

cclerget: @muffgaga Linter complains about formatting error, you need to execute `gofmt -w pkg/util/loop/loop_linux.go` to fix that

cclerget:
ping @muffgaga

dtrudg:
Hi @muffgaga - In the interest of getting this into a `3.4.1` release, which we hope can happen soon, @cclerget has made the final changes in #4449 - which, of course, retains credit to you for the fix in the change log and commit record.

Many thanks again.

PR #4384, opened 2019-09-05
Print ip in instance list command

### LINKED ISSUE #3751: Networking addresses should be accessible
Currently, when I create instances (even with --net arguments) I can't easily see the resulting addresses with list:

```bash
$ sudo singularity instance list
app              8756     /home/vanessa/Documents/Dropbox/Code/singularity/singularity-compose-example/app/app.sif
db               9095     /home/vanessa/Documents/Dropbox/Code/singularity/singularity-compose-example/db/db.sif
nginx            8365     /home/vanessa/Documents/Dropbox/Code/singularity/singularity-compose-example/nginx/nginx.sif
```

I have to start them, then execute commands to the container myself. Here is what works for ubuntu/debian flavored (per your example docs):

```bash
$ sudo singularity exec instance://app hostname -I
10.22.0.12
```
But that isn't a universal solution - this will work on *some* busybox:

```bash
$ sudo singularity exec instance://nginx ip addr show eth0 | grep "inet\b" | awk '{print $2}' | cut -d/ -f1
10.22.0.11
```
It's sort of hacky, and a mess. It should be possible to list instances, and have this information available, since the Singularity client creates the networking. And the same goes for hostname, although the command `hostname` does in fact work for both.

```bash
$ sudo singularity instance list --json
{
	"instances": [
		{
			"instance": "nginx",
			"pid": 8365,
                        "hostname": "nginx",
                        "ip_address": "10.22.0.11",
			"img": "/home/vanessa/Documents/Dropbox/Code/singularity/singularity-compose-example/nginx/nginx.sif"
		}
	]
}
```

Would this be possible?

pisarukv:
closes https://github.com/sylabs/singularity/issues/3751


sashayakovtseva (commented):
lgtm, but need @cclerget opinion

internal/pkg/runtime/engine/singularity/process_linux.go
```diff
@@ -567,3 +573,25 @@ func preStartProcess(e *EngineOperations) error {
 
 	return nil
 }
+
+func (e *EngineOperations) getIP() (string, error) {
+	if e.EngineConfig.Network == nil {
+		return "", errors.New("not in net ns")
```
cclerget: I would pass `nil` error to avoid to display warning message
```suggestion
		return "", nil
```

internal/pkg/runtime/engine/singularity/process_linux.go
```diff
@@ -567,3 +573,25 @@ func preStartProcess(e *EngineOperations) error {
...
+	}
+
+	net := e.EngineConfig.GetNetwork()
```
cclerget: `Network` can take a list of network separated by commas

```suggestion
	net := strings.Split(e.EngineConfig.GetNetwork(), ",")
```

internal/pkg/runtime/engine/singularity/process_linux.go
```diff
@@ -567,3 +573,25 @@ func preStartProcess(e *EngineOperations) error {
...
+
+	ip, err := e.EngineConfig.Network.GetNetworkIP(net, "4")
```
cclerget:
```suggestion
	ip, err := e.EngineConfig.Network.GetNetworkIP(net[0], "4")
```

internal/pkg/runtime/engine/singularity/process_linux.go
```diff
@@ -567,3 +573,25 @@ func preStartProcess(e *EngineOperations) error {
...
+	if err == nil {
+		return ip.String(), nil
+	}
+	sylog.Warningf("Could not get ipv4 %s", err)
+
+	ip, err = e.EngineConfig.Network.GetNetworkIP(net, "6")
```
cclerget:
```suggestion
	ip, err = e.EngineConfig.Network.GetNetworkIP(net[0], "6")
```

cclerget (changes requested):

cclerget (approved):
LGTM, thanks !

PR #4387, opened 2019-09-05
Fix fakeroot image cleanup permission denied and rework starter execution util

### LINKED ISSUE #4263: singularity shell --fakeroot leaves file permission problem
This is rather an inherent problem of fakeroot, but I hope Singularity can do something smart about this. 

## Version of Singularity:

singularity version 3.3.0

### Expected behavior
ownership of files and directories should be mapped back to the USER who starts the container

### Actual behavior
```
INFO:    Cleaning up image...
ERROR:   failed to delete container image /tmp/rootfs-569592909: unlinkat /tmp/rootfs-569592909/home/testsub/.bash_logout: permission denied
```

- start a container in shell mode with fake root: `singularity shell --fakeroot library://ubuntu`
- create a new user, say with username `testsub` inside the container with useradd command. 
- create a test directory, say `testdir` under fakeroot's home, and set permission: `chmod 777 testdir`. This is to ensure it is writetable to user `testsub`.
- switch to `testsub` user and cd into `testdir`. 
- create some test file there, for example `touch testfilebytestsub`.
- exit the container
- observe two things: `testfilebytestsub` under `testdir`  has user id and group id that are different from the current users; and files under `/home/testsub` which is actually under container's temporary directory also have the same ownerships as `testfilebytessub`. That means the HOSTUSER can't cleanup files without sudo/admin right.

cclerget:

Move internal/pkg/util/exec/pipe* to internal/pkg/util/starter and code to execute starter was refactored to remove some code redundancies disseminate in CLI and internal packages.

Also fix issue during image cleanup when using fakeroot with SUID workflow, because CleanupContainer method is called by master process when fakeroot+SUID workflow are in use this method is executed outside of the fakeroot context which means that under some circumstances the image removal may fail with a permission denied error. To fix the issue the fakeroot engine is called in order to execute /bin/rm command.

- Fixes #4263


internal/pkg/util/starter/starter.go
```diff
@@ -0,0 +1,144 @@
...
+type Command struct {
+	path   string
+	env    []string
+	stdin  io.Reader
+	stdout io.Writer
+	stderr io.Writer
+}
+
+// Exec executes the starter binary in place of the caller if
+// there is no error. This function never return on success.
```
sashayakovtseva:
```suggestion
// there is no error. This function never returns on success.
```

sashayakovtseva (approved):

PR #4366, opened 2019-09-03
Changing conveyor_oci.go to take LD_LIBRARY_PATH to fix bug #3331

Amjadhpc:
**Description of the Pull Request (PR):**
This PR tries to take all ENV having PATH in their name for docker builds. Currently env variables from docker are not being passed to singularity environment variable

Testing:
singularity pull docker://quay.io/wtsicgp/dockstore-cgpwxs:singularity_bug_report
singularity exec singularity_bug_report.sif  samtools

Program: samtools (Tools for alignments in the SAM format)
Version: 1.9 (using htslib 1.9)

Usage:   samtools <command> [options]

Commands:
  -- Indexing
     dict           create a sequence dictionary file
     faidx          index/extract FASTA
     fqidx          index/extract FASTQ
     index          index alignment

  -- Editing
     calmd          recalculate MD/NM tags and '=' bases
     fixmate        fix mate information
     reheader       replace BAM header
     targetcut      cut fosmid regions (for fosmid pool only)
     addreplacerg   adds or replaces RG tags
     markdup        mark duplicates

  -- File operations
     collate        shuffle and group alignments by name
     cat            concatenate BAMs
     merge          merge sorted alignments
     mpileup        multi-way pileup
     sort           sort alignment file
     split          splits a file by read group
     quickcheck     quickly check if SAM/BAM/CRAM file appears intact
     fastq          converts a BAM to a FASTQ
     fasta          converts a BAM to a FASTA

  -- Statistics
     bedcov         read depth per BED region
     depth          compute the depth
     flagstat       simple stats
     idxstats       BAM index stats
     phase          phase heterozygotes
     stats          generate stats (former bamcheck)

  -- Viewing
     flags          explain BAM flags
     tview          text alignment viewer
     view           SAM<->BAM<->CRAM conversion
     depad          convert padded BAM to unpadded BAM

**This fixes or addresses the following GitHub issues:**

- Fixes #
Bug #3331

**Before submitting a PR, make sure you have done the following:**

- Read the [Guidelines for Contributing](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/sylabs/singularity/blob/master/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- Added tests to validate this PR and tested this PR locally with a `make testall`
[...]

Issue #3978, opened 2019-07-23
Invalid environment variable name prevents running container


benmishkanian:
## Version of Singularity:

3.1.0-1.el7.centos

### Expected behavior

According to IEEE Std 1003.1-2001, environment variables may contain any character except '=', even though some shells such as Bash do not support certain characters (such as '.') in environment variable names. Docker allows using '.' in the name of an `ENV` variable, even if the `ENTRYPOINT` is Bash. Running such a Docker container still works, and Bash does not crash. Singularity containers built from such Docker images should work correctly even if the container's shell does not support such environment variable names.

### Actual behavior

Singularity containers built from Docker images that have an `ENV` variable with a '.' character in the name crash when starting Bash.

### Steps to reproduce behavior

Create a Singularity container from the Docker image `jsreport/jsreport:2.5.0` (available on Docker Hub) and try to run it. It fails with the following error:

```-bash: export: `chrome:launchOptions:executablePath=abc': not a valid identifier```

**Comment by sashayakovtseva** [comment]
@amjad you said this issue is fixed in https://github.com/sylabs/singularity/pull/4398#issuecomment-529466896, but I still can reproduce it by running

```
$ singularity shell docker://jsreport/jsreport:2.5.0
```

Amjadhpc:
@sashayakovtseva  Yes   i  just tested it .  Just a question can we use os.Setenv() while creating the docker2singularity.sh?   .What would be the implications of such step ? obviously export will not work ?

carterpeel:
Hello,

This is a templated response that is being sent out to all open issues.  We are working hard on 'rebuilding' the Singularity community, and a major task on the agenda is finding out what issues are still outstanding.


**Please consider the following:**

1. Is this issue a duplicate, or has it been fixed/implemented since being added?
2. Is the issue still relevant to the current state of Singularity's functionality?
3. Would you like to continue discussing this issue or feature request?


Thanks,
Carter

pedroalvesbatista:
@benmishkanian Still getting this problem ? Have this been solved already ? If yes, what work around have you followed and applied ?

We're looking into the issue carefully, soon will bring to community and discuss ways to better solve as well address this. Thankyou for keeping the interest in the subject.

benmishkanian:
I don't know. I stopped using Singularity a long time ago and no longer have access to an HPC cluster.

kmuriki:
[...]

PR #4398, opened 2019-09-06
Clean all environment variables that contain PATH

### LINKED ISSUE #3331: Docker container pulled with singularity 2.5.2 works with 3.1.1 exec, but fails if pulled with 3.1.1
## Version of Singularity:

`pull 2.5.2 + exec 3.1.1` vs `pull & exec 3.1.1`

### Expected behavior

Container pulled with 3.1.1 should exec as well (or better) than one pulled with 2.5.2

### Actual behavior

If I pull a docker image with singularity 2.5.2 it runs fine with the 3.1.1 exec:

```bash
$ /software/singularity-3.1.1/bin/singularity exec dockstore-cgpwxs_singularity_bug_report_pull_2.5.2.simg samtools view

Usage: samtools view [options] <in.bam>|<in.sam>|<in.cram> [region ...]
...
```

If I then pull the same image with 3.1.1 and attempt to run it I get problems loading shared libraries, these are ones being used within the container (not by singularity itself as in #3027):

```bash
$ /software/singularity-3.1.1/bin/singularity exec dockstore-cgpwxs_singularity_bug_report_pull_3.1.1.simg samtools view
samtools: error while loading shared libraries: libhts.so.2: cannot open shared object file: No such file or directory
```

On the container pulled with 3.1.1 I've checked the files indicated as being a problem:

```bash
$ /software/singularity-3.1.1/bin/singularity exec dockstore-cgpwxs_singularity_bug_report_pull_3.1.1.simg ls -l /opt/wtsi-cgp/lib/
total 8613
-rwxr-xr-x  1 root root  415532 Apr 16 10:33 libBigWig.a
-rwxr-xr-x  1 root root  240712 Apr 16 10:33 libBigWig.so
-rw-r--r--  1 root root 4950682 Apr 16 10:33 libhts.a
lrwxrwxrwx  1 root root      13 Apr 16 10:33 libhts.so -> libhts.so.1.9
-rw-r--r--  1 root root 3211128 Apr 16 10:33 libhts.so.1.9
lrwxrwxrwx  1 root root      13 Apr 16 10:33 libhts.so.2 -> libhts.so.1.9
drwxrwxr-x 63 root root    1186 Apr 16 07:31 perl5
drwxr-xr-x  2 root root      32 Apr 15 08:08 pkgconfig
```

This is an exact match to the same folder when pulled with 2.5.2.

I'd expect the image pulled with 3.1.1 and executed with 3.1.1 to function better than one pulled with 2.5.2.

Covered above, created specific tag on quay.io to ensure this version is retained for testing purposes, e.g.

singularity pull docker://quay.io/wtsicgp/dockstore-cgpwxs:singularity_bug_report

Related post on [stack overflow](https://stackoverflow.com/q/55703347/4460430)

sashayakovtseva:

- do not propagate any envs that contain PATH
- get special PATH envs from the passed envs, not by calling `Getenv`. Currently `SetContainerEnv` is called from two places: [here](https://github.com/sylabs/singularity/blob/master/cmd/internal/cli/actions_linux.go#L478) where calling `Getenv` doesn't make sense since they are collected right before, and [here](https://github.com/sylabs/singularity/blob/master/internal/pkg/runtime/engine/imgbuild/process_linux.go#L106) where envs are already cleaned by starter.
- fix wrong tests in the package
- use empty struct as a canonical way to represent set

- Fixes #3331
- Closes #4366

### DIFF
--- a/internal/pkg/util/env/clean.go
+++ b/internal/pkg/util/env/clean.go
@@ -6,48 +6,24 @@
 package env
 
 import (
-	"os"
 	"strings"
 
 	"github.com/opencontainers/runtime-tools/generate"
 	"github.com/sylabs/singularity/internal/pkg/sylog"
 )
 
-const (
-	envPrefix = "SINGULARITYENV_"
-)
-
-var alwaysPassKeys = map[string]bool{
-	"TERM":        true,
-	"http_proxy":  true,
-	"HTTP_PROXY":  true,
-	"https_proxy": true,
-	"HTTPS_PROXY": true,
-	"no_proxy":    true,
-	"NO_PROXY":    true,
-	"all_proxy":   true,
-	"ALL_PROXY":   true,
-	"ftp_proxy":   true,
-	"FTP_PROXY":   true,
+var alwaysPassKeys = map[string]struct{}{
+	"term":        {},
+	"http_proxy":  {},
+	"https_proxy": {},
+	"no_proxy":    {},
+	"all_proxy":   {},
+	"ftp_proxy":   {},
 }
 
-// SetContainerEnv cleans environment variables before running the container
-func SetContainerEnv(g *generate.Generator, env []string, cleanEnv bool, homeDest string) {
-	// first deal with special variables that allow user to control $PATH at
-	// runtime (meh... special cases)
-	if prependPath := os.Getenv("SINGULARITYENV_PREPEND_PATH"); prependPath != "" {
-		g.AddProcessEnv("SING_USER_DEFINED_PREPEND_PATH", prependPath)
-	}
-
-	if appendPath := os.Getenv("SINGULARITYENV_APPEND_PATH"); appendPath != "" {
-		g.AddProcessEnv("SING_USER_DEFINED_APPEND_PATH", appendPath)
-	}
-
-	if userPath := os.Getenv("SINGULARITYENV_PATH"); userPath != "" {
-		g.AddProcessEnv("SING_USER_DEFINED_PATH", userPath)
-	}
-
-	for _, env := range env {
+// SetContainerEnv cleans environment variables before running the container.
+func SetContainerEnv(g *generate.Generator, hostEnvs []string, cleanEnv bool, homeDest string) {
+	for _, env := range hostEnvs {
 		e := strings.SplitN(env, "=", 2)
 		if len(e) != 2 {
 			sylog.Verbosef("Can't process environment variable %s", env)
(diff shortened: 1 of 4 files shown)


sashayakovtseva:
Hello @Amjadhpc,
I would appreciate your feedback on this one, it should fix the issue.
Thanks!

mem (approved):

mem:
@Amjadhpc,it would be great if you can try this and report your results here. If something is missing, we can follow up.

Amjadhpc:
@sashayakovtseva 
Yes no issues works like expecte  . Also fyi this also fixes bug  #3978  and could confirm it by testing it

sashayakovtseva:
Thank you @Amjadhpc!

DrDaveD:
I haven't tried to test this yet, but the description sounds like it may have unintended side effects.  First of all, does this apply only when building containers (which was the reported problem in #3331) or does it also apply when running containers?  If it's only for building then I'd be less concerned.

sashayakovtseva:
@DrDaveD that is applied during starter execution.
So yes, when you run the container starter is called as well, and any envs with `PATH` are not propagated.

DrDaveD:
In that case this sounds like a quite significant change in the Command Line Interface.  There may have been multiple scripts depending on the functionality of passing in *PATH environment variables from the outside.  Hasn't singularity always allowed that up to this point?  Or is this fixing a difference from singularity-2?

sashayakovtseva:
@DrDaveD I joined team since 3.0, but afaik this is actually fixing the difference.

DrDaveD:
I just tested doing exec with singularity-2.6.1, singularity-3.4.0, and the master branch, on a container that had been built by a singularity-3 version with `singularity build --sandbox centos6-sandbox docker://centos:6`.  In all 3 cases PATH and LD_LIBRARY_PATH were completely set in the container and it did not import them from the host.  However GOPATH was imported in the released versions but not in the master branch.  I think this is asking for trouble when you start getting complaints why some variables are no longer being passed in when they used to be.

I think the regression you were trying to fix compared to singularity-2 was only for building images, so this fix should only apply to building images.

sashayakovtseva:
@DrDaveD in `docker://centos:6` you get `LD_LIBRARY_PATH=/.singularity.d/libs` from Singularity scripts (`99-base.sh`),  not from Docker image itself (compare envs by `docker run --rm centos:6 env`).

Also regarding your example: 
```
$ singularity version
3.4.0
$ singularity build --sandbox centos6-sandbox docker://centos:6
$ singularity exec centos6-sandbox/ env | grep LD_LIBRARY_PATH
$ export LD_LIBRARY_PATH=/foo/bar
$ singularity exec centos6-sandbox/ env | grep LD_LIBRARY_PATH
LD_LIBRARY_PATH=/foo/bar:/.singularity.d/libs
```

So as I can see host's `LD_LIBRARY_PATH` affects container.

Now let's see what happens if env was set in docker:
```
$ singularity version
3.4.0
$ singularity exec docker://quay.io/wtsicgp/dockstore-cgpwxs:singularity_bug_report env | grep LD_LIBRARY_PATH
LD_LIBRARY_PATH=/foo/bar:/.singularity.d/libs
$ docker run --rm quay.io/wtsicgp/dockstore-cgpwxs:singularity_bug_report env | grep LD_LIBRARY_PATH
LD_LIBRARY_PATH=/opt/wtsi-cgp/lib
```

The docker env is not respected, and that is what this PR fixes.

sashayakovtseva:
Now let's test with Singularity 2.6:

```
$ env | grep LD_LIBRARY_PATH
LD_LIBRARY_PATH=/foo/bar
$ singularity --version
2.6.1-vault/release-2.6.9103f0155
$ singularity exec docker://quay.io/wtsicgp/dockstore-cgpwxs:singularity_bug_report env | grep LD_LIBRARY_PATH
LD_LIBRARY_PATH=/opt/wtsi-cgp/lib:/.singularity.d/libs
```

So basically this PR returns previous behavior.

DrDaveD:
> So basically this PR returns previous behavior.

Ok it does for those examples, that's good, but I'm complaining about the change of behavior for other *PATH variables.  That is a new behavior that I do not think we want.

sashayakovtseva:
Okay, a couple e to clarify.

2.6: 
```
$ env | grep PATH
GOPATH=/home/vagrant/go
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/usr/local/go/bin:/home/vagrant/go/bin
$ singularity --version
2.6.1-vault/release-2.6.9103f0155
$ singularity exec docker://quay.io/wtsicgp/dockstore-cgpwxs:singularity_bug_report env | grep PATH
LD_LIBRARY_PATH=/opt/wtsi-cgp/lib:/.singularity.d/libs
PATH=/opt/wtsi-cgp/bin:/opt/wtsi-cgp/biobambam2/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
GOPATH=/home/vagrant/go
```

3.4:
```
$ env | grep PATH
GOPATH=/home/vagrant/go
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/usr/local/go/bin:/home/vagrant/go/bin
$ singularity version
3.4.0
$ singularity exec docker://quay.io/wtsicgp/dockstore-cgpwxs:singularity_bug_report env | grep PATH
LD_LIBRARY_PATH=/opt/wtsi-cgp/lib:/.singularity.d/libs
USER_PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/usr/local/go/bin:/home/vagrant/go/bin:/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin
PATH=/opt/wtsi-cgp/bin:/opt/wtsi-cgp/biobambam2/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
GOPATH=/home/vagrant/go
```

master: 
```
$ singularity version
3.4.0-163.g4ad9becce
$ singularity exec docker://quay.io/wtsicgp/dockstore-cgpwxs:singularity_bug_report env | grep PATH
LD_LIBRARY_PATH=/opt/wtsi-cgp/lib:/.singularity.d/libs
PATH=/opt/wtsi-cgp/bin:/opt/wtsi-cgp/biobambam2/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

@DrDaveD I see your point, but do you think it is reasonable to have any host's *PATH* envs in container? 
We might need to discuss it with @cclerget and @mem as well.

DrDaveD:
Yes I think it's quite reasonable and I think it is very likely that there are scripts that rely upon that behavior.  If people don't want the environment they use --cleanenv.

PR #4399, opened 2019-09-06
Store image labels in a label partition

### LINKED ISSUE #2790: Singularity not honoring json specification for inspect

### LINKED ISSUE #3050: improve error message when singularity inspect tries to inspect a bad file

### LINKED ISSUE #3932: Inspection metadata no longer containers container size

### LINKED ISSUE #4310: Labels should show what architecture the image is

WestleyK:

This PR changes:
 1. Adds a label partition in the container when building a SIF.
 2. Stores the labels, and app labels in the label partition.
 3. Inspect will print the whole deffile partition when the `--deffile` flag is passed.

 - Addresses: #4279
 - Fixes: #3932
 - Fixes: #4310
 - Fixes: #3050
 - Fixes: #2790

<br>


WestleyK:
This should be good for the first round of review!

sashayakovtseva:
@WestleyK I think you should cleanup this PR before anyone gets to review it.

cmd/internal/cli/inspect.go
```diff
@@ -12,20 +12,22 @@ import (
 	"io"
```
ikaneshiro: In general, I don't think we can rename this file to change the build tag since there are still paths that launch the singularity runtime which wont work on darwin

internal/pkg/build/build.go
```diff
@@ -298,7 +298,11 @@ func (b *Build) Full() error {
 			a.HandleSection(k, v)
 		}
 
-		a.HandleBundle(stage.b)
+		err := a.HandleBundle(stage.b)
+		if err != nil {
+			return err
```
ikaneshiro: Lets wrap this with something a like "while inserting app metadata:..." or something

internal/pkg/build/metadata/metadata.go
```diff
@@ -0,0 +1,169 @@
...
+	descr[0], _, err = fimg.GetPartPrimSys()
+	if err != nil {
+		return nil, fmt.Errorf("no primary partition found")
+	}
+
+	return descr, nil
+}
+
+// GetSIFData will return a dataType from a SIF.
+func GetSIFData(fimg *sif.FileImage, dataType sif.Datatype) ([]*sif.Descriptor, error) {
```
ikaneshiro: Doesn't look like it's doing that. It looks like its returning _all_ descriptors of a SIF that are of a particular `dataType`. If this is intended to be a nice util function, maybe have one specific for labels, and one specific for defintions, otherwise this is mostly just a wrapping of `GetLinkedDescrsByType()` and doesn't seem that useful. You can make a nice helper to check for a system partition since that's the only extra thing this is doing, but I have a feeling may that exist somewhere already.

ikaneshiro (changes requested):
As @sashayakovtseva said. This definitely needs to be cleaned up and ideally squashed into a lot less commits

WestleyK:
> and ideally squashed into a lot less commits

Yes, I was already planing for that :smile:

cmd/internal/cli/inspect.go
WestleyK: How about I'll split the `inspect.go` to `inspect_darwin.go`, and `inspect_linux.go`. That way we will still be able to inspect a container on darwin (only a sandbox or image with label partition).

internal/pkg/build/metadata/metadata.go
WestleyK: Okay, I'll just remove this helper function. Since I'v cleaned the code, and removed extra code, this function got a lot smaller, its not needed anymore.

cmd/internal/cli/inspect.go
ikaneshiro: Sure, but i recommend doing that in a separate pr to keep this diff from getting even larger

ikaneshiro: I see its already there, so nevermind :+1:

WestleyK:
Closing by https://github.com/sylabs/singularity/pull/4417.

PR #4405, opened 2019-09-09
Add key import flag to specify if new pass needed

### LINKED ISSUE #4369: Newpair: should mimic 'gpg --import'
## Version of Singularity:

```
v3.4
```

### Expected behavior

When importing a private key, by default it should not ask to set a new password. That should be a separate flag like `--with-newpassword`.

### Actual behavior

As of now, `import` will ask to set a newpassword when importing a key.

Write here.

pisarukv:
closes https://github.com/sylabs/singularity/issues/4369


sashayakovtseva (approved):

cmd/internal/cli/key_import.go
```diff
@@ -8,6 +8,8 @@ package cli
 import (
 	"os"
 
+	"github.com/sylabs/singularity/pkg/cmdline"
+
```
WestleyK: Extra newline

```suggestion
```

cmd/internal/cli/key_import.go
```diff
@@ -27,9 +29,18 @@ var KeyImportCmd = &cobra.Command{
 	Example: docs.KeyImportExample,
 }
 
+var keyImportWithNewPassword bool
+var keyImportWithNewPasswordFlag = cmdline.Flag{
+	ID:           "keyImportWithNewPasswordFlag",
+	Value:        &keyImportWithNewPassword,
+	DefaultValue: false,
+	Name:         "with-newpassword",
```
WestleyK: Maybe just `new-password` will be better?

```suggestion
	Name:         "new-password",
```

cmd/internal/cli/key_import.go
```diff
@@ -27,9 +29,18 @@ var KeyImportCmd = &cobra.Command{
...
+	Usage:        `if specified will ask to set a new password`,
```
WestleyK: ~Maybe?...~

```
	Usage:        `if specified, newpair will ask to set a new password`,
```

WestleyK (changes requested):
Have a couple suggestions, otherwise looks good.

cmd/internal/cli/key_import.go
WestleyK: Or better:

```suggestion
	Usage:        `set a new password to the private key`,
```

WestleyK (commented):
We will also need to update the E2E tests to not send to many passwords.

pkg/sypgp/sypgp.go
```diff
@@ -1014,38 +1018,35 @@ func (keyring *Handle) importPrivateKey(entity *openpgp.Entity) error {
...
-
-	// Make a clone of the entity
-	newEntity := &openpgp.Entity{
-		PrimaryKey:  entity.PrimaryKey,
-		PrivateKey:  entity.PrivateKey,
-		Identities:  entity.Identities,
-		Revocations: entity.Revocations,
-		Subkeys:     entity.Subkeys,
-	}
+	newEntity := *entity
```
WestleyK, line 1021: Why make a `newEntity`? can you just use `entity`?

pkg/sypgp/sypgp.go
```diff
@@ -1014,38 +1018,35 @@ func (keyring *Handle) importPrivateKey(entity *openpgp.Entity) error {
...
-		return err
+	if password != "" {
+		if err := newEntity.PrivateKey.Encrypt([]byte(password)); err != nil {
+			return err
+		}
 	}
 
 	// Store the private key
-	if err := keyring.appendPrivateKey(newEntity); err != nil {
+	if err := keyring.appendPrivateKey(&newEntity); err != nil {
```
WestleyK, line 1049: Why is the `&` here? Is it also necessary for the other calls to `keyring.appendPubKey`, and `keyring.appendPrivateKey`?

WestleyK (changes requested):
This behaves much nicer! Just a couple comment/suggestions. Also, the e2e tests need to be updated (the `import` tests only need to send the password once)

pkg/sypgp/sypgp.go
pisarukv: So for some reasons here was used a copy of entity. I decided to leave it as is. I'll investigate deeper if we can use entity directly.

pkg/sypgp/sypgp.go
sashayakovtseva: @WestleyK `&` is a pointer gian operation. It is there b/c `keyring.appendPubKey` accepts pointer, not value.

WestleyK (approved):
This looks good! We can update the E2E tests later. _Also maybe worth squashing the commits?..._

PR #4406, opened 2019-09-09
Fix truncated gecos field when reading /etc/passwd

### LINKED ISSUE #4323: getent info truncated
## Version of Singularity:

3.x

### Expected behavior

getent passwd for a user calling a container should match the record outside of the container.

### Actual behavior

gecos is truncated, so only Full Name is available, but not the email.

python -c 'import pwd,os; user=os.getenv("USER"); print(pwd.getpwnam(user))'
outside
vs
singularity exec <IMAGE> python -c 'import pwd,os; user=os.getenv("USER"); print(pwd.getpwnam(user))'
inside the container. See the pw_gecos field.

sashayakovtseva:

When Singularity engine updates `/etc/passwd` inside the container, current user is fetched with `user.Lookup`. Go's `os/user` truncates `gecos` to fetch user name, while we need full content of the filed. 

So in this PR Singularity's `user` package calls C library itself to prevent `gecos` truncating. Basically, we now have a simplified copy of Go's `os/user/cgo_lookup_unix.go` file. Note that unlike Go, we don't cache current user information. Also note that we don't use the same buildtags that Go uses (not sure we should).

@moskalenko Feel free to test this one :)
 
- Fixes #4323

Attn: @sylabs/singularity-maintainers


internal/pkg/util/user/lookup_unix.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+func buildUser(pwd *C.struct_passwd) *User {
+	return &User{
+		Name:  C.GoString(pwd.pw_name),
+		UID:   uint32(pwd.pw_uid),
+		GID:   uint32(pwd.pw_gid),
+		Gecos: C.GoString(pwd.pw_gecos),
+		Dir:   C.GoString(pwd.pw_dir),
+		Shell: "/bin/sh",
```
cclerget:
```suggestion
		Shell: C.GoString(pwd.pw_shell),
```

mem (changes requested):
Can you provide a example `/etc/passwd` file where the current implementation fails?

sashayakovtseva:
@mem sure, posted on the issue https://github.com/sylabs/singularity/issues/4323#issuecomment-529545840

cclerget (changes requested):

internal/pkg/util/user/lookup_unix.go
sashayakovtseva: Thank you!

cclerget (approved):

internal/pkg/util/user/lookup_unix.go
```diff
@@ -0,0 +1,246 @@
+// Copyright 2011 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Copyright (c) 2019, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+// This is the simplified version of Go's os/user/cgo_lookup_unix.go file.
```
mem: I would keep the original build tags (even if it's not explicit, I suspect we are building singularity with CGO enabled, otherwise we would have gotten an issue around not working on systems using NIS or LDAP), keep the original names and structure as much as possible (with the original errors, too), _not_ do the shell patching here and modify `buildUser` so that it uses `Gecos` instead of `Name` and it doesn't edit the Gecos information to extract the name.

Add a _different_ `.go` file that calls `buildUser` or `lookupUser` as necessary and do the shell patching there (look at the original `convertUser`, I get the impression we lost that in this change, if it's still there I'm not finding it).

mem:
> @mem sure, posted on the issue [#4323 (comment)](https://github.com/sylabs/singularity/issues/4323#issuecomment-529545840)

thanks, it wasn't clear to me what the issue was. Please see the comment I left in the code.

The change looks good, I would just try to keep the copied code as close to the original as possible, so that we don't have to do three-way merges in the future.

moskalenko:
To clarify, the query at container instantiation is done against LDAP and not an /etc/passwd file on the host system. It's inside the container that an /etc/passwd is generated.

sashayakovtseva:
@moskalenko yep, the C part is responsible for handling that

internal/pkg/util/user/lookup_unix.go
sashayakovtseva: After some thoughts I still think we need to keep shell patching and try to get rid of  unnecessary int/string casting. 
However, I agree that it is reasonable to keep build tags, errors and file structure alike.

sashayakovtseva:
@mem IMO diff looks pretty reasonable and as small as possible now:

```
8,15d7
< // Copyright (c) 2019, Sylabs Inc. All rights reserved.
< // This software is licensed under a 3-clause BSD license. Please consult the
< // LICENSE.md file distributed with the sources of this project regarding your
< // rights to use or distribute this software.
< 
< // This is a slightly patched version of Go's os/user/cgo_lookup_unix.go file.
< // We need full gecos content so the only way to do so is to call C ourselves.
< // User and Group types are taken from this package rather then from os/user.
20d11
< 	osuser "os/user"
21a13
> 	"strings"
84c76
< 		return nil, osuser.UnknownUserError(username)
---
> 		return nil, UnknownUserError(username)
117c109
< 		return nil, osuser.UnknownUserIdError(uid)
---
> 		return nil, UnknownUserIdError(uid)
123,129c115,127
< 	return &User{
< 		Name:  C.GoString(pwd.pw_name),
< 		UID:   uint32(pwd.pw_uid),
< 		GID:   uint32(pwd.pw_gid),
< 		Gecos: C.GoString(pwd.pw_gecos),
< 		Dir:   C.GoString(pwd.pw_dir),
< 		Shell: C.GoString(pwd.pw_shell),
---
> 	u := &User{
> 		Uid:      strconv.FormatUint(uint64(pwd.pw_uid), 10),
> 		Gid:      strconv.FormatUint(uint64(pwd.pw_gid), 10),
> 		Username: C.GoString(pwd.pw_name),
> 		Name:     C.GoString(pwd.pw_gecos),
> 		HomeDir:  C.GoString(pwd.pw_dir),
> 	}
> 	// The pw_gecos field isn't quite standardized. Some docs
> 	// say: "It is expected to be a comma separated list of
> 	// personal data where the first item is the full name of the
> 	// user."
> 	if i := strings.Index(u.Name, ","); i >= 0 {
> 		u.Name = u.Name[:i]
130a129
> 	return u
157c156
< 		return nil, osuser.UnknownGroupError(groupname)
---
> 		return nil, UnknownGroupError(groupname)
190c189
< 		return nil, osuser.UnknownGroupIdError(strconv.Itoa(gid))
---
> 		return nil, UnknownGroupIdError(strconv.Itoa(gid))
197c196
< 		GID:  uint32(grp.gr_gid),
---
> 		Gid:  strconv.Itoa(int(grp.gr_gid)),
```

mem (approved):

PR #4426, opened 2019-09-12
go mod vendor (vendor cleanup?)

WestleyK:

While on my dev branch, when I ran `go mod vendor`, some vendor stuff got deleted. Maybe that cleans some stuff up?

- Fixes #

<br>


mem:
You are using Go 1.13, which is deleting additional files wrt 1.12.

We need to find a way to support this, but we cannot commit this as is, as it would break CI.

PR #4434, opened 2019-09-13
build_runtime_suid.mk

### LINKED ISSUE #4429: allow disabling the new "make install as root" check
[`build_runtime_suid.mk`](https://github.com/sylabs/singularity/blob/b94a931eb8f095b266e368be07fa3d274d440f77/mlocal/frags/build_runtime_suid.mk#L14-L17) changed between v3.3.0 and v3.4.0, due to issue #4227, which was addressed by #4262.

I use [Spack](https://spack.readthedocs.io/en/latest/), whose raison d'etre (HPC with environment modules) is exactly like @michaelkarlcoleman described in #4227:
> We're in an HPC environment and use Lmod to manage software, which means that we install pretty much everything into custom paths. (This is likely true for a lot of Singularity users.)

Spack users also commonly do scripted installs, of suites of software, installed automatically and reproducibly from source.  All without root.  (Spack generally handles root requiring steps (like this) with post-hoc scripts.)

I was recently [helping to package](https://github.com/spack/spack/pull/12785) Singularity v3.4.0, and in order to get it to work, we now have to [patch](https://github.com/spack/spack/blob/cb3dc206dc04047d634ea902af2dc6be815baf65/var/spack/repos/builtin/packages/singularity/singularity_v3.4.0_remove_root_check.patch) the file `build_runtime_suid.mk` to remove this check.

I'd argue the "exit 1 if not root" check creates new problems in use cases like the one I describe.   Perhaps consider adding a flag to the configure step, to turn off this check?  That would allow projects like Spack to more elegantly install the software.

jmstover:
Update the starter-suid install to install the binary when ran as
a normal user. But if ran as root do the correct thing and set
the binary permission to 4755.

Fixes: #4429

### DIFF
--- a/mlocal/frags/build_runtime_suid.mk
+++ b/mlocal/frags/build_runtime_suid.mk
@@ -14,7 +14,7 @@ $(starter_suid_INSTALL): $(starter_suid)
 	@echo " INSTALL SUID" $@
 	$(V)install -d $(@D)
 	$(V)install $(starter_suid) $(starter_suid_INSTALL)
-	@if [ `id -u` -ne 0 -a -z "${RPM_BUILD_ROOT}" ] ; then \
+	@if [ `id -u` -ne 0 ] ; then \
 		echo "$(starter_suid_INSTALL) -- installed with incorrect permissions"; \
 	else \
 		$(V)chmod 4755 $(starter_suid_INSTALL); \
--- a/mlocal/frags/build_runtime_suid.mk
+++ b/mlocal/frags/build_runtime_suid.mk
@@ -11,13 +11,14 @@ $(starter_suid): $(BUILDDIR)/.clean-starter $(singularity_build_config) $(starte
 		-o $@ $(SOURCEDIR)/cmd/starter/main_linux.go
 
 $(starter_suid_INSTALL): $(starter_suid)
-	@if [ `id -u` -ne 0 -a -z "${RPM_BUILD_ROOT}" ] ; then \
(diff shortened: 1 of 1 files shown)


mlocal/frags/build_runtime_suid.mk
```diff
@@ -11,13 +11,14 @@ $(starter_suid): $(BUILDDIR)/.clean-starter $(singularity_build_config) $(starte
...
 $(starter_suid_INSTALL): $(starter_suid)
-	@if [ `id -u` -ne 0 -a -z "${RPM_BUILD_ROOT}" ] ; then \
-		echo "SUID binary requires to execute make install as root, use sudo make install to finish installation"; \
-		exit 1 ; \
-	fi
 	@echo " INSTALL SUID" $@
 	$(V)install -d $(@D)
-	$(V)install -m 4755 $(starter_suid) $(starter_suid_INSTALL)
+	$(V)install $(starter_suid) $(starter_suid_INSTALL)
+	@if [ `id -u` -ne 0 -a -z "${RPM_BUILD_ROOT}" ] ; then \
```
cclerget: We can get rid of `RPM_BUILD_ROOT` check

```suggestion
	@if [ `id -u` -ne 0 ] ; then \
```

cclerget (approved):
LGTM, just suggestion for `RPM_BUILD_ROOT` removal

sashayakovtseva (approved):
Thanks @jmstover !

PR #4451, opened 2019-09-16
Detect autofs path automatically and get rid of autofs bug path configuration

### LINKED ISSUE #2556: autofs bug fix is still incomplete
We use 2.5.2 and 2.6.0 and for us the current auto fs bug fix is not working all that well.

we have multiple storage systems and we use autofs to create a unified folder structure so the end user can be seamlessly migrated to another storage system.

This means we have about 1500 autofs mountpoints for home directories and 250 mountpoints for shared folders and all of those have to be configured in singularity .conf

If I don't have read access to another folder (like most, we are using umask 0007) Singularity sends me a warning for each and every mount point 

```
petersen@rhino3:~$ singularity run ./ubuntu-latest.simg
WARNING: Failed to open directory '/fh/fast/xyz'
WARNING: Failed to open directory '/fh/fast/abc'
WARNING: Failed to open directory '/fh/fast/klm'
WARNING: Failed to open directory '/home/xyz'
WARNING: Failed to open directory '/home/abc'
WARNING: Failed to open directory '/home/klm'
```

The effect is that we only put the folders in there that are absolutely  required which also means that now each user has to notifyc us that they want to use singularity so we can add their folder to the singularity.conf file. This leads us to not want to widely advertise singularity . Can this autofs isssue be addressed ?  It should be easy to detect which root folder is under the rule of auto fs :

```
petersen@rhino3:~$ cat /proc/mounts | grep auto | grep /fh/fast
auto.fast /fh/fast autofs rw,relatime,fd=17,pgrp=6636,timeout=300,minproto=5,maxproto=5,indirect 0 0
```

and then the actual autofs mount point below

```
petersen@rhino3:~$ cat /proc/mounts  | grep /fh/fast/_HDC
silver:/ifs/data/fast/_HDC /fh/fast/_HDC nfs rw,relatime,vers=3,rsize=131072,wsize=524288,namlen=255,hard,proto=tcp,timeo=600,retrans=2,sec=sys,mountaddr=140.107.130.81,mountvers=3,mountport=300,mountproto=udp,local_lock=none,addr=140.107.130.81 0 0
```

Those could be detected and fixed at runtime when they are used instead of globally trying to acccess all configured mount points at application startup?

cclerget:

Look at `/proc/self/mountinfo` to look for potential mount point under autofs control and open them in stage 1 in order to keep a reference in host mount namespace. Those changes make `autofs bug path` configuration directive obsolete.

- Fixes #2556

### DIFF
--- a/internal/pkg/runtime/engine/singularity/prepare_linux.go
+++ b/internal/pkg/runtime/engine/singularity/prepare_linux.go
@@ -369,60 +369,98 @@ func (e *EngineOperations) prepareRootCaps() error {
 	return nil
 }
 
-func (e *EngineOperations) prepareFd(starterConfig *starter.Config) error {
+func keepAutofsMount(source string, autoFsPoints []string) (int, error) {
+	resolved, err := filepath.EvalSymlinks(source)
+	if err != nil {
+		return -1, err
+	}
+
+	for _, p := range autoFsPoints {
+		if strings.HasPrefix(resolved, p) {
+			sylog.Debugf("Open file descriptor for %s", resolved)
+			f, err := os.Open(resolved)
+			if err != nil {
+				return -1, err
+			}
+			return int(f.Fd()), nil
+		}
+	}
+
+	return -1, fmt.Errorf("no mount point")
+}
+
+func (e *EngineOperations) prepareAutofs(starterConfig *starter.Config) error {
+	const mountInfoPath = "/proc/self/mountinfo"
+
+	entries, err := proc.GetMountInfoEntry(mountInfoPath)
+	if err != nil {
+		return fmt.Errorf("while parsing %s: %s", mountInfoPath, err)
+	}
+	autoFsPoints := make([]string, 0)
+	for _, e := range entries {
+		if e.FSType == "autofs" {
+			sylog.Debugf("Found %q as autofs mount point", e.Point)
+			autoFsPoints = append(autoFsPoints, e.Point)
+		}
+	}
+	if len(autoFsPoints) == 0 {
+		sylog.Debugf("No autofs mount point found")
+		return nil
+	}
+
 	fds := make([]int, 0)
 
 	if e.EngineConfig.File.UserBindControl {
 		for _, b := range e.EngineConfig.GetBindPath() {
 			splitted := strings.Split(b, ":")
 
-			src, err := filepath.Abs(splitted[0])
+			fd, err := keepAutofsMount(splitted[0], autoFsPoints)
 			if err != nil {
+				sylog.Debugf("Could not keep file descriptor for user bind path %s: %s", splitted[0], err)
 				continue
 			}
-
-			if !fs.IsDir(src) {
-				continue
-			}
-
-			sylog.Debugf("Open file descriptor for %s", src)
-			f, err := os.Open(src)
-			if err != nil {
-				continue
-			}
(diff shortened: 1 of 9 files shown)


sashayakovtseva (approved):

internal/pkg/runtime/engine/singularity/prepare_linux.go
```diff
@@ -369,60 +369,78 @@ func (e *EngineOperations) prepareRootCaps() error {
...
-
-			if !fs.IsDir(src) {
-				continue
-			}
-
-			sylog.Debugf("Open file descriptor for %s", src)
-			f, err := os.Open(src)
-			if err != nil {
-				continue
+			if fd, err := keepAutofsMount(splitted[0], autoFsPoints); err == nil {
```
WestleyK: Should this fail (return an error) if one occurs?

internal/pkg/runtime/engine/singularity/prepare_linux.go
```diff
@@ -369,60 +369,78 @@ func (e *EngineOperations) prepareRootCaps() error {
...
-
-			if !fs.IsDir(src) {
-				continue
-			}
 
-			sylog.Debugf("Open file descriptor for %s", src)
-			f, err := os.Open(src)
-			if err != nil {
-				continue
+			if fd, err := keepAutofsMount(splitted[0], autoFsPoints); err == nil {
```
WestleyK: Same here

internal/pkg/runtime/engine/singularity/prepare_linux.go
```diff
@@ -369,60 +369,78 @@ func (e *EngineOperations) prepareRootCaps() error {
...
+			sylog.Debugf("Found %q as autofs mount point", e.Point)
+			autoFsPoints = append(autoFsPoints, e.Point)
+		}
+	}
+	if len(autoFsPoints) == 0 {
+		sylog.Debugf("No autofs mount point found")
+		return nil
+	}
+
 	fds := make([]int, 0)
```
WestleyK, line 411: Why is the buffer here 0? Should it be at least 1?

internal/pkg/runtime/engine/singularity/prepare_linux.go
```diff
@@ -369,60 +369,78 @@ func (e *EngineOperations) prepareRootCaps() error {
...
 			}
-			fds = append(fds, int(f.Fd()))
 		}
-	}
-
-	for _, path := range e.EngineConfig.File.AutofsBugPath {
-		if !fs.IsDir(path) {
-			continue
+		// check home source directory
+		if fd, err := keepAutofsMount(e.EngineConfig.GetHomeSource(), autoFsPoints); err == nil {
```
WestleyK: Here too, should't this return a error if one occurs?

WestleyK (commented):
Couple questions about error handling, otherwise looks good.

internal/pkg/runtime/engine/singularity/prepare_linux.go
cclerget: All errors are deliberately ignored, we are in a best-effort approach regarding autofs bug, Singularity will do its best to fix that transparently for users but must not abort if it fail, if we can't open file for some reason (eg: permissions rights), I could log those errors to help further debugging

WestleyK (approved):
Looks good to me!

PR #4449, opened 2019-09-16
Fix ioctl errno handling for LOOP_SET_STATUS64

cclerget:
**Description of the Pull Request (PR):**

Based on work from @muffgaga in #4373 with minor addition and loop device cleanup in case of failure.

Fix ioctl errno handling for LOOP_SET_STATUS64 Linux kernel commit 5db470e229e22b7eda6e23b5566e532c96fb5bc3 ("loop:
drop caches if offset or block_size are changed") introduced a transient
fail (with -EAGAIN) possibility into the LOOP_SET_STATUS64 ioctl.
This also affects longterm kernel versions 4.14 and 4.19.

**This fixes or addresses the following GitHub issues:**

- Fixes #3764
- Related to #4373 

**Before submitting a PR, make sure you have done the following:**

- Read the [Guidelines for Contributing](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/sylabs/singularity/blob/master/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- Added tests to validate this PR and tested this PR locally with a `make testall`
- Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](https://github.com/sylabs/singularity/blob/master/CONTRIBUTORS.md)

Attn: @singularity-maintainers

PR #4461, opened 2019-09-16
Cherry pick #4449 to 3.4.1

dtrudg:
Cherry pick #4449 to fix ioctl errno handling on loop operations in 3.4.1

Tested with a tight loop of `singularity exec` which failed often on my Pi4 (kernel 4.19.57-v7l+) prior to this fix - and now does not fail with this fix.


ArangoGutierrez (approved):
if Circle is Happy i am happy

PR #4462, opened 2019-09-16
Cherry pick 4321 and 4375 for 3.4.1

ikaneshiro:
Cherry picks #4321 and #4375 

Tested by running `mconfig` and then adding `GO_MODFLAGS := -x` to the `Makefile` before running `make`. Without this fix, the build will fail.


ArangoGutierrez (approved):
tested on Fedora 30, works

PR #4439, opened 2019-09-13
Update to scs-library-client 0.4.4

EmmEff:
**Description of the Pull Request (PR):**

Update to `scs-library-client` 0.4.4, which includes fix for image upload issue on clients with low memory.

**This fixes or addresses the following GitHub issues:**

Closes #4393 

**Before submitting a PR, make sure you have done the following:**

- Read the [Guidelines for Contributing](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/sylabs/singularity/blob/master/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- Added tests to validate this PR and tested this PR locally with a `make testall`
- Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](https://github.com/sylabs/singularity/blob/master/CONTRIBUTORS.md)

Attn: @singularity-maintainers

PR #4466, opened 2019-09-16
Cherry pick #4439 for 3.4.1

ikaneshiro:
Apply #4439  to branch for 3.4.1 - Update to scs-library-client 0.4.4.

This required adding an additional commit to commit vendor to a new state. I think this is due to the mixing of different dependency versions between `release-3.4` and `master` at the time of merging this commit. This feels unavoidable making the commit necessary.

Tested by using `htop` and not seeing a 4 gig spike in memory usage when uploading a 4 gig container.


dtrudg (approved):
This one is working for me too :+1:

PR #4467, opened 2019-09-16
Use hkps pool for test hitting external keyserver

dtrudg:
Replace a single-server hardcoded hkps URI with the address of the hkps pool.

This should help stop PR merge being blocked by consistently failing tests where the one keyserver we depend on has an issue.

https://sks-keyservers.net/overview-of-pools.php

>hkps.pool.sks-keyservers.net
>
>This is a pool containing only servers available using hkps. Regular A and AAAA and SRV records are included for port 443 servers, and a lookup is performed for _pgpkey-https._tcp on the individual servers to determine if a hkps enabled service is listening on another port. At this point, however, servers not running on port 443 are not included.
>
>This pool only include servers that have been certified by the sks-keyservers.net CA, of which the certificate can be found at https://sks-keyservers.net/sks-keyservers.netCA.pem [OpenPGP signature] [CRL].

### DIFF
--- a/e2e/key/key.go
+++ b/e2e/key/key.go
@@ -95,12 +95,12 @@ func (c *ctx) singularityKeySearch(t *testing.T) {
 		},
 		{
 			name:   "key search --url <open key server> <name>",
-			args:   []string{"search", "--url", "https://keyserver.2ndquadrant.com", "WestleyK"},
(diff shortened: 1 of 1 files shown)


mem (approved):

dtrudg:
Well - I tried this and it worked. But then it started resolving to a different pool server with a certificate from an unknown authority... so this isn't going to fix it :-(

PR #4477, opened 2019-09-17
Sign all signable partitions

### LINKED ISSUE #4425: Sign should default to sign all sign-able partitions
## Version of Singularity:

```
v3.4
```

### Expected behavior

When you run `singularity sign foo.sif`, it should sign all sign-able partitions on that sif.

### Actual behavior

Only signed the system partition.

Write here.

WestleyK:

This PR changes the `sign` behavior to sign all singable partitions, (non-signature partitions).

I know that there will be some disusing about default `verify` and `sign` behavior, so if the default `sign` behavior is to sign all singable partitions, then `verify` should also verify all signable partitions.

 - Fixes #4425

### DIFF
--- a/pkg/signing/signing.go
+++ b/pkg/signing/signing.go
@@ -87,6 +87,41 @@ func sifAddSignature(fimg *sif.FileImage, groupid, link uint32, fingerprint [20]
 	return nil
 }
 
+// Copy-paste from sylabs/sif
+// datatypeStr returns a string representation of a datatype.
+func datatypeStr(dtype sif.Datatype) string {
+	switch dtype {
+	case sif.DataDeffile:
+		return "Def.FILE"
+	case sif.DataEnvVar:
+		return "Env.Vars"
+	case sif.DataLabels:
+		return "JSON.Labels"
+	case sif.DataPartition:
+		return "FS"
+	case sif.DataSignature:
+		return "Signature"
+	case sif.DataGenericJSON:
+		return "JSON.Generic"
+	case sif.DataGeneric:
+		return "Generic/Raw"
+	}
+	return "Unknown data-type"
+}
+
+func getDataPartitionToSign(fimg *sif.FileImage, dataType sif.Datatype) ([]*sif.Descriptor, error) {
+	sylog.Debugf("Looking for: %s partition to sign...", datatypeStr(dataType))
+	// We are using ID 0 (skipping ID), because we are looking for all Datatypes,
+	// and ID's will limit the search.
+	data, _, err := fimg.GetLinkedDescrsByType(uint32(0), dataType)
+	if err != nil && err != sif.ErrNotFound {
+		return nil, fmt.Errorf("failed to get descr for deffile: %s", err)
+	}
+	sylog.Debugf("Found %d partitions", len(data))
+
+	return data, nil
+}
+
 // descrToSign determines via argument or interactively which descriptor to sign
 func descrToSign(fimg *sif.FileImage, id uint32, isGroup bool) ([]*sif.Descriptor, error) {
 	descr := make([]*sif.Descriptor, 1)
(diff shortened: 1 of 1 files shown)


pkg/signing/signing.go
```diff
@@ -89,15 +87,64 @@ func sifAddSignature(fimg *sif.FileImage, groupid, link uint32, fingerprint [20]
 	return nil
 }
 
+// Copy-paste from sylabs/sif
+// datatypeStr returns a string representation of a datatype.
+func datatypeStr(dtype sif.Datatype) string {
```
mem, line 92: this is usually done by defining a `String()` method on the type, `sif.Datatype` in this case.

The copy/paste comment makes me guess this is already there?

pkg/signing/signing.go
```diff
@@ -89,15 +87,64 @@ func sifAddSignature(fimg *sif.FileImage, groupid, link uint32, fingerprint [20]
...
+		return "JSON.Generic"
+	case sif.DataGeneric:
+		return "Generic/Raw"
+	}
+	return "Unknown data-type"
+}
+
+func getDataPartitionToSign(fimg *sif.FileImage, dataType sif.Datatype) ([]*sif.Descriptor, error) {
+	sylog.Debugf("Looking for: %s partition to sign...", datatypeStr(dataType))
+	data, _, err := fimg.GetLinkedDescrsByType(uint32(0), dataType)
```
mem, line 116: could you please add a comment here explaning the logic behind the first argument?

It's looking for descriptors linking to ID 0 with a specific data type, so much is clear, but why 0? That's the bit that needs documenting so that other people don't get stuck on this.

pkg/signing/signing.go
```diff
@@ -89,15 +87,64 @@ func sifAddSignature(fimg *sif.FileImage, groupid, link uint32, fingerprint [20]
...
+	var err error
 
 	if id == 0 {
 		descr[0], _, err = fimg.GetPartPrimSys()
 		if err != nil {
 			return nil, fmt.Errorf("no primary partition found")
 		}
+
+		// signablePartitions is a list of all the signable datatypes.
+		signablePartitions := []sif.Datatype{
```
mem: nit picking: it might be more clear to call this "signableDatatypes" and call the loop variable "datatype" instead of s?

pkg/signing/signing.go
```diff
@@ -122,16 +169,16 @@ func descrToSign(fimg *sif.FileImage, id uint32, isGroup bool) (descr []*sif.Des
 func Sign(cpath string, id uint32, isGroup bool, keyIdx int) error {
 	keyring := sypgp.NewHandle("")
 
+	// Generate a private key usable for signing
```
mem: is it "generating" or "loading"?

mem (commented):
Some minor issues, otherwise it looks good

pkg/signing/signing.go
WestleyK: This function already exists here https://github.com/sylabs/sif/blob/master/pkg/sif/fmt.go#L61, since that non-exported (for now...) I just copy-paste here.

Since this function is not _that_ importent (it only generates a name for `sylog.Debugf()`), I can leave it out until we have the `String()` method for `Datatype`. I'm fine either way.

pkg/signing/signing.go
WestleyK: It should be `Load a private key...`

pkg/signing/signing.go
ikaneshiro: I think this is OK for now. I think it makes sense to support a `String()` method for data types within the SIF package. Lets open an issue for that in the SIF repo and integrate it the next time we cut a SIF release

ikaneshiro (approved):
lgtm

pkg/signing/signing.go
WestleyK: Issue opened -> https://github.com/sylabs/sif/issues/67

PR #4479, opened 2019-09-18
Get capabilities from /proc/self/status in E2E tests

mem:
The current implementation of testSecurityPriv and testSecurityUnpriv
uses ping to infer the status of the cap_net_raw capability, as a proxy
for testing whether --keep-privs, --no-privs and --drop-caps are doing
the right thing.

Change this to directly reading /proc/self/status to obtain the
effective set of capabilities in the container. The downside of this is
that the complete set of capabilities might change, so this test might
break in the future if something is added to the set of known
capabilities. Punt that problem for now.

This removes the need to download an image, making the test faster.

fixing or enhancing.

 - Fixes #

### DIFF
--- a/e2e/security/security.go
+++ b/e2e/security/security.go
@@ -7,7 +7,6 @@ package security
 
 import (
 	"os"
-	"path/filepath"
 	"testing"
 
 	"github.com/sylabs/singularity/e2e/internal/e2e"
@@ -16,8 +15,7 @@ import (
 )
 
 type ctx struct {
-	env     e2e.TestEnv
-	pingImg string
+	env e2e.TestEnv
 }
 
 // testSecurityUnpriv tests the security flag fuctionality for singularity exec without elevated privileges
@@ -28,6 +26,7 @@ func (c ctx) testSecurityUnpriv(t *testing.T) {
 		argv       []string
 		opts       []string
 		preFn      func(*testing.T)
+		expectOp   e2e.SingularityCmdResultOp
 		expectExit int
 	}{
 		// taget UID/GID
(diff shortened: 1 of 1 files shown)


e2e/security/security.go
```diff
@@ -63,27 +62,34 @@ func (c ctx) testSecurityUnpriv(t *testing.T) {
...
-			expectExit: 255,
+			expectExit: 255, // singularity errors out, --keep-privs needs root
 		},
 		{
-			name:       "capabilities_keep-false",
-			argv:       []string{"ping", "-c", "1", "8.8.8.8"},
-			expectExit: 2,
+			// we start without any capabilities, the
+			// expected set is empty.
+			name:     "capabilities_keep-false",
```
jscook2345, line 72: I think this is supposed to be an underscore between `keep` and `false`.

jscook2345 (changes requested):
It looks good minus the typo.

e2e/security/security.go
```diff
@@ -141,23 +147,39 @@ func (c ctx) testSecurityPriv(t *testing.T) {
...
+			// this might break if new capabilities are
+			// added
+			//
+			// TODO(mem): what we really want to test is
+			// that the set outside the container is the
+			// same set inside the container.
+			name:     "capabilities_keep",
+			argv:     []string{"grep", "^CapEff:", "/proc/self/status"},
+			opts:     []string{"--keep-privs"},
+			expectOp: e2e.ExpectOutput(e2e.RegexMatch, `CapEff:\s+0000003fffffffff\n`),
```
cclerget, line 159: Old distro like centos 6, sles 11 give `fffffffffffffffff` here. Overall is ping on `127.0.0.1` work ? or running a command like tcpdump to test if capability is dropped ?

e2e/security/security.go
mem: believe me, I held back touching that sort of thing. I would have changed so much :-\

jscook2345 (approved):

e2e/security/security.go
mem: Pinging 127.0.0.1 instead will work, yes.

But once I started down this path I questioned two things:

- What is this test testing?
- Do I really need to pull 35+ MB of additional containers to test this?

The test as it exists is testing that ping works. It is *assumed* that is happens because the cap_net_raw capability is preserved.

But that's not what --keep-privs is supposed to do. --keep-privs is supposed to preserve _all_ capabilities.

So using ping for this purpose is insufficient, at best. As noted in the comments I added, this new test is brittle. Suggestions as to how to fix this are welcomed.

PR #4499, opened 2019-09-20
Verify all partitions

### LINKED ISSUE #4491: Verify should verify all partitions
### Version of Singularity:

```
3.4
```

### Expected behavior

Verify should verify all non-signature partitions by default, unless the user specifies the `--sif-id` flag.

```
$ singularity verify foo.sif
Container is signed by 2 key(s):

Verifying partition: SystemPartition:
4C3303B43EFDA24A73BB0D9D0FC0B7217DB6DA2A
[LOCAL]            westley (bar) <foo>
[OK]               Data integrity verified

Verifying partition: DataDeffile:
4C3303B43EFDA24A73BB0D9D0FC0B7217DB6DA2A
[LOCAL]            westley (bar) <foo>
[OK]               Data integrity verified
```

<br>

 - Related to https://github.com/sylabs/singularity/pull/4477 / https://github.com/sylabs/singularity/issues/4425

### Actual behavior

Only verifies the system partition by default.

WestleyK:

This PR changes the default behavior of `verify` to verify all partitions. Now this behaves the same as `sign`.

#### Example

Verifying a alpine container:

```
$ singularity verify alpine_latest.sif 
WARNING: No signature found for Def.FILE
Container is signed by 1 key(s):

Verifying partition: FS:
8883491F4268F173C6E5DC49EDECE4F3F38D871E
[LOCAL]   Sylabs Admin <support@sylabs.io>
[OK]      Data integrity verified

INFO:    Container verified: alpine_latest.sif
```

Verifying a container that has all partitions signed: (any newly signed container after https://github.com/sylabs/singularity/pull/4477)

```
$ singularity verify foo.sif 
Container is signed by 3 key(s):

Verifying partition: Def.FILE:
C7E7C8C3635DD06930669A2283B30190FEEF8162
[REMOTE]  Westley (from gpg) <westley@sylabs.io>
[OK]      Data integrity verified

Verifying partition: FS:
C7E7C8C3635DD06930669A2283B30190FEEF8162
[REMOTE]  Westley (from gpg) <westley@sylabs.io>
[OK]      Data integrity verified

Verifying partition: JSON.Labels:
C7E7C8C3635DD06930669A2283B30190FEEF8162
[REMOTE]  Westley (from gpg) <westley@sylabs.io>
[OK]      Data integrity verified

INFO:    Container verified: foo.sif
```

<br>

**Must be merged after https://github.com/sylabs/singularity/pull/4611**

 - Fixes #4491


WestleyK:
Now should be ready for review!

pkg/signing/signing.go
```diff
@@ -274,6 +275,51 @@ func getSigsPrimPart(fimg *sif.FileImage) (sigs []*sif.Descriptor, descr []*sif.
...
+	var sigs []*sif.Descriptor
+	var err error
+
+	// Ensure theres a primary partition.
+	_, _, err = fimg.GetPartPrimSys()
+	if err != nil {
+		return nil, nil, fmt.Errorf("no primary partition found")
+	}
+
+	// Loop thrugh all the partitions, (skipping DataSignatures)
```
ikaneshiro:
```suggestion
	// Loop through all the partitions, (skipping DataSignatures)
```

pkg/signing/signing.go
```diff
@@ -274,6 +275,51 @@ func getSigsPrimPart(fimg *sif.FileImage) (sigs []*sif.Descriptor, descr []*sif.
...
+			continue
+		}
+		// No need to verify a signature.
+		if d.Datatype == sif.DataSignature {
+			continue
+		}
+		s, _, err := fimg.GetLinkedDescrsByType(d.ID, sif.DataSignature)
+		if err != nil {
+			// If a partition is not signed, print a warning.
+			// TODO: how will/should this efect the exit code?
```
ikaneshiro: The exit codes for verify are something I think we need to consider at a higher level in general and document. This is a bit off topic, but what are the current exit codes in use?

pkg/signing/signing.go
```diff
@@ -274,6 +275,51 @@ func getSigsPrimPart(fimg *sif.FileImage) (sigs []*sif.Descriptor, descr []*sif.
...
+			sylog.Warningf("No signature found for %s\n", datatypeStr(d.Datatype))
+			continue
+		}
+
+		// For every signature part, there will be a data partition.
+		for _, sp := range s {
```
ikaneshiro: I think this data duplication is ok for now since the need to do so should be eliminated with upcoming refactoring of this package

ikaneshiro (dismissed):
Can you update the `datatypestr()` to include an entry to crypto messages? Otherwise we get a confusing warning:
```
$ singularity sif list test.sif 
Container id: 767ffab1-251e-4382-bd60-d19c301257dc
Created on:   2019-08-28 16:58:49 -0400 EDT
Modified on:  2019-09-20 10:45:33 -0400 EDT
Descriptor list:
1    |1       |NONE    |32768-32801               |Def.FILE
2    |1       |NONE    |36864-19496960            |FS (Encrypted squashfs/*System/amd64)
3    |1       |2       |19496960-19497360         |Cryptographic Message (PEM/RSA-OAEP)
4    |1       |2       |19501056-19502011         |Signature (SHA384)
5    |1       |1       |19505152-19506107         |Signature (SHA384)
ian@T440s:keys$ singularity verify test.sif 
WARNING: No signature found for Unknown data-type
Container is signed by 2 key(s):

Verifying partition: Def.FILE:
54807B145686544B68C950CB76DDF06FCDB05A8B
[LOCAL]   Ian <ian@sylabs.io>
[OK]      Data integrity verified

Verifying partition: FS:
54807B145686544B68C950CB76DDF06FCDB05A8B
[LOCAL]   Ian <ian@sylabs.io>
[OK]      Data integrity verified

INFO:    Container verified: test.sif
```

Otherwise appears to work as advertised!

pkg/signing/signing.go
WestleyK: Currently, by default (`singularity verify foo.sif`), the exit code will be 0 if there is no corrupted data. If the `--local` flag is specified, them it will succeed only if all the keys are local.

Actually, I think the current behavior for this PR is fine. If a user wants to ensure a specific partition is signed, then they need to use the `--sif-id` flag.

pkg/signing/signing.go
WestleyK: Yes, this should be refactored later. But this will create a Descriptor, for every DataSignature, so if a partition is signed twice, there will be two Descriptors in that slice/array.

pkg/signing/signing.go
```diff
@@ -274,6 +275,50 @@ func getSigsPrimPart(fimg *sif.FileImage) (sigs []*sif.Descriptor, descr []*sif.
...
+			continue
+		}
+		// No need to verify a signature.
+		if d.Datatype == sif.DataSignature {
+			continue
+		}
+		s, _, err := fimg.GetLinkedDescrsByType(d.ID, sif.DataSignature)
+		if err != nil {
+			// If a partition is not signed, print a warning.
+			sylog.Warningf("No signature found for %s\n", datatypeStr(d.Datatype))
```
sashayakovtseva: Are you sure we should print a warning here? E.g. I haven't signed some JSON config, IMO that is not something to consider (maybe I am wrong here). Also when we do `singularity sign` does it sign all partitions? If not, that will lead to lots of false warnings.
p.s. no need for new line in logs (it is added automatically)

pkg/signing/signing.go
```diff
@@ -274,6 +275,50 @@ func getSigsPrimPart(fimg *sif.FileImage) (sigs []*sif.Descriptor, descr []*sif.
...
+		// For every signature part, there will be a data partition.
+		for _, sp := range s {
+			sigs = append(sigs, sp)
+
+			partDescr := d
+			dpart = append(dpart, &partDescr)
+		}
+	}
+
+	if len(dpart) == 0 || len(sigs) == 0 {
```
sashayakovtseva: No point in double check, lengths are the same here

pkg/signing/signing.go
WestleyK: Yes, sign will now sign all partitions, as of https://github.com/sylabs/singularity/pull/4477. I think there should be a warning here. If not a warning, what do you think it should be?

dtrudg: A warning for every non-signed partition generated with an earlier version might be annoying. What about at the end we say something more like...

`system partition verified.....`

`all partitions verified....`

?

WestleyK: I'm okay with `All partitions verified` after the verify output, but I kinda like the warning here, it give the user a general idea on what partitions are there, and whats signed/unsigned. But I'll change it if you prefer something else.

mem: I don't mind a warning per descriptor. I kind of mind the format of the message. I would prefer something like:

```
WARNING: Missing signature for SIF descriptor 1 (Def.FILE)
```

It will create questions from users, yes (mostly in the form of "do I care?"). The danger with this sort of message is that you are training users to ignore them, but since by default there are going to be only a handful of these lines, I think it's OK.

I would output a INFO-level message like:

```
INFO: All SIF descriptors verified
```

just to let users know that everything is OK.

Over time the ecosystem will transition from the warnings to this last message (as images are updated).

pkg/signing/signing.go
```diff
@@ -274,6 +277,50 @@ func getSigsPrimPart(fimg *sif.FileImage) (sigs []*sif.Descriptor, descr []*sif.
...
+			// If a partition is not signed, print a warning.
+			sylog.Warningf("No signature found for %s\n", datatypeStr(d.Datatype))
+			continue
+		}
+
+		// For every signature part, there will be a data partition.
+		for _, sp := range s {
+			sigs = append(sigs, sp)
+
+			partDescr := d
```
mem: I had to think for a second as to why you are doing this here...

This is Go telling you that you probably shouldn't be doing this in the first place.

As the descriptor array (fimg.DescrArr) is not going to change under you, and the elements of the array are readily available, it's probably a better idea to store indexes into the array than pointers to sif.Descriptor.

You might be better off returning a single slice, too:

```
type signatureLink {
   sigIndex  int  // the index of the descriptor with the signature
   dataIndex int  // the index of the descriptor of the signed data
}

func getAllSignatureLinks(fimg *sif.FileImage) ([]signatureLink, error) {
    // ...
    _, idxs, err := fimg.GetLinkedDescrsByType(d.ID, sif.DataSignature)
    // ...
    for sidx := range idxs {
       tbl = append(tbl, signatureLink{ sidx, didx })
    }
    // ...
    return tbl, nil
}
```

mem (approved):

PR #4487, opened 2019-09-19
Fix regression breaking --net usage

cclerget:
## Description of the Pull Request (PR):

Fix another regression introduced by #4411 when using `--net`, using `--net` result in following error :

```
singularity shell -f --net docker://alpine
FATAL:   container creation failed: mount /proc/self/ns/net->/usr/local/var/singularity/mnt/session/netns error: while mounting /proc/self/ns/net: while getting mount flags for /proc/self/ns/net: while searching parent mount point entry for /proc/self/ns/net: while resolving path /proc/self/ns/net: lstat /proc/12772/ns/net:[4026532009]: no such file or directory
```

#### Before submitting a PR, make sure you have done the following:

- Read the [Guidelines for Contributing](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/sylabs/singularity/blob/master/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- Added tests to validate this PR and tested this PR locally with a `make testall`
- Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](https://github.com/sylabs/singularity/blob/master/CONTRIBUTORS.md)

Attn: @singularity-maintainers

PR #4508, opened 2019-09-23
Add network e2e helper and few basic networking e2e tests

### LINKED ISSUE #4496: Add require helper for network
Add `require` helper for network

cclerget:

This PR adds a `require` `Network` helper to proceed with e2e network tests. Also adds few basic networking tests to e2e actions.

 - Fixes #4496
 - Related to #4487

### DIFF
--- a/internal/pkg/test/tool/require/require_linux.go
+++ b/internal/pkg/test/tool/require/require_linux.go
@@ -1,0 +1,110 @@
+// Copyright (c) 2019, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package require
+
+import (
+	"fmt"
+	"os/exec"
+	"path/filepath"
+	"sync"
+	"syscall"
+	"testing"
+
+	"github.com/sylabs/singularity/internal/pkg/buildcfg"
+	"github.com/sylabs/singularity/pkg/network"
+)
+
+var hasUserNamespace bool
+var hasUserNamespaceOnce sync.Once
+
+// UserNamespace checks that the current test could use
+// user namespace, if user namespaces are not enabled or
+// supported, the current test is skipped with a message.
+func UserNamespace(t *testing.T) {
+	// not performance critical, just save extra execution
+	// to get the same result
+	hasUserNamespaceOnce.Do(func() {
+		// user namespace is a bit special, as there is no simple
+		// way to detect if it's supported or enabled via a call
+		// on /proc/self/ns/user, the easiest and reliable way seems
+		// to directly execute a command by requesting user namespace
+		cmd := exec.Command("/bin/true")
+		cmd.SysProcAttr = &syscall.SysProcAttr{
+			Cloneflags: syscall.CLONE_NEWUSER,
+		}
+		// no error means user namespaces are enabled
+		err := cmd.Run()
+		hasUserNamespace = err == nil
+		if !hasUserNamespace {
+			t.Logf("Could not use user namespaces: %s", err)
+		}
+	})
+	if !hasUserNamespace {
+		t.Skipf("user namespaces seems not enabled or supported")
+	}
+}
+
+var supportNetwork bool
+var supportNetworkOnce sync.Once
+
+// Network check that bridge network is supported by
+// system, if not the current test is skipped with a
+// message.
+func Network(t *testing.T) {
+	supportNetworkOnce.Do(func() {
+		logFn := func(err error) {
+			t.Logf("Could not use network: %s", err)
+		}
+
+		cmd := exec.Command("/bin/cat")
+		cmd.SysProcAttr = &syscall.SysProcAttr{}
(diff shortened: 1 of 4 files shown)


internal/pkg/test/tool/require/require_linux.go
```diff
@@ -0,0 +1,99 @@
...
+	"testing"
+
+	"github.com/sylabs/singularity/internal/pkg/buildcfg"
+	"github.com/sylabs/singularity/pkg/network"
+)
+
+var hasUserNamespace bool
+var hasUserNamespaceOnce sync.Once
+
+var supportNetwork bool
```
jscook2345: Should these variables be before the `Network` function?

internal/pkg/test/tool/require/require_linux.go
```diff
@@ -0,0 +1,99 @@
...
+		}
+		// no error means user namespaces are enabled
+		hasUserNamespace = cmd.Run() == nil
+	})
+	if !hasUserNamespace {
+		t.Skipf("user namespaces seems not enabled or supported")
+	}
+}
+
+// Network check that bridge network is supported by
```
jscook2345, line 53: Why do we require that the bridge network is enabled? Just testing for CNI?

I figured this would just test for the network namespace working.

Could this be split into multiple functions? 1 for the namespace require, 1 for the network require?

internal/pkg/test/tool/require/require_linux.go
```diff
@@ -0,0 +1,99 @@
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		cmd := exec.Command(cmdPath)
+		cmd.SysProcAttr = &syscall.SysProcAttr{}
+		cmd.SysProcAttr.Cloneflags = syscall.CLONE_NEWNET
+
+		stdinPipe, err = cmd.StdinPipe()
+		if err != nil {
+			return
```
jscook2345, line 69: Could we return the error here instead of nothing and print that as the reason the bridge network is not supported?

internal/pkg/test/tool/require/require_linux.go
```diff
@@ -0,0 +1,99 @@
...
+		}
+
+		if err := cmd.Start(); err != nil {
+			return
```
jscook2345, line 75: Same as above. Return error for message

internal/pkg/test/tool/require/require_linux.go
```diff
@@ -0,0 +1,99 @@
...
+
+		nsPath := fmt.Sprintf("/proc/%d/ns/net", cmd.Process.Pid)
+
+		cniPath := new(network.CNIPath)
+		cniPath.Conf = filepath.Join(buildcfg.SYSCONFDIR, "singularity", "network")
+		cniPath.Plugin = filepath.Join(buildcfg.LIBEXECDIR, "singularity", "cni")
+
+		setup, err := network.NewSetup([]string{"bridge"}, "_test_", nsPath, cniPath)
+		if err != nil {
+			return
```
jscook2345, line 87: Same as above.

internal/pkg/test/tool/require/require_linux.go
```diff
@@ -0,0 +1,99 @@
...
+		}
+		if err := setup.AddNetworks(); err != nil {
+			return
```
jscook2345, line 95: Same as above

internal/pkg/test/tool/require/require_linux.go
```diff
@@ -0,0 +1,99 @@
...
+		stdinPipe, err = cmd.StdinPipe()
+		if err != nil {
+			return
+		}
+
+		if err := cmd.Start(); err != nil {
+			return
+		}
+
+		nsPath := fmt.Sprintf("/proc/%d/ns/net", cmd.Process.Pid)
```
jscook2345, line 78: Why do we start the `cat` command here? Could we instead use `/proc/self/ns/net`? If not, why not?

jscook2345 (changes requested):
Few comments. Please let me know if you have questions. Thanks! Great work!

internal/pkg/test/tool/require/require_linux.go
cclerget: Will log error but can't change function prototype, all functions in `require` must have the same prototype as test function where any errors -> skip current test

internal/pkg/test/tool/require/require_linux.go
cclerget: By doing that we would screw up the current network namespace which is the host namespace, running `cat` is a convenient way to pause (until we close stdin pipe) while we are manipulating the command process namespace from Go side without interfering with current namespaces.

internal/pkg/test/tool/require/require_linux.go
cclerget: What the reason of splitting it into two functions ? If we check for network namespace it means we are going to use `network` package with CNI

jscook2345: I guess I thought checking if the network namespace is an enabled feature should be a smaller function we could maybe reuse, but you're correct in that if it is enabled, we would use CNI anyways.

jscook2345 (approved):
👍

PR #4515, opened 2019-09-24
Pass go's path to gengodep

### LINKED ISSUE #3547: sudo make install UX issue with go path
## Version of Singularity:

3.2.0

### Expected behavior

If go is required to be in the root PATH (and if the GOPATH needs to be set in that environment) please specify it in the install docs or provide a clear error message. If not, please make sure 'make install' does not throw the following error.

### Actual behavior

go dependency resolver throws an error complaining that go is not in the root's PATH when running 'sudo make install' in the user environment with sudo access, go in the PATH, and GOPATH set as required for the singularity build build.

See above

mem:
When mconfig runs, it determines the absoluate path to the go compiler
to be used. The gengodep script is used in the generated Makefile in
order to generate dependencies. Because of the way this has been
implemented, it runs every time the Makefile is used, even if the
compilation is already done. Since gengodep is not using the path that
mconfig determined, when running "make install" it's possible that the
"go" program is no longer found via PATH.

Fix this by passing the absoluate path to gengodep.

To avoid weird errors, add an argument to gengodep to specify the
expected "version" of the script. If the makefile is updated without
updating the script, or if the script is updated without updating the
Makefile (the more common case), a message will be printed and an error
will be returned. The error is not enough to make the compilation fail,
but it should be visible enough to let the user know what needs to
happen.

Fixes #3547.

fixing or enhancing.

 - Fixes #


makeit/gengodep
```diff
@@ -1,10 +1,29 @@
 #!/bin/sh
 
-srcdir=$1
-gotags=$2
-shift 2
+if [ "$1" != -v2 ] ; then
```
jmstover:
```suggestion
if [ "$1" != "-v2" ] ; then
```

I would quote the `-v2` part

jmstover (changes requested):
Ack... didn't start a review. But there's the one quote change.

jmstover (approved):

mem:
code already passed CI

Issue #3880, opened 2019-07-06
Importing docker image results in incorrect file permissions


AdamSimpson:
## Version of Singularity:

v3.2.1

### Expected behavior

When importing a docker image I expect file permissions to be applied correctly.

### Actual behavior

If a Docker image is imported that was built such that a files permissions are changed in a separate layer than the file was originally created in the permissions when running under Singularity are incorrect. 

### Steps to reproduce behavior
Dockerfile
```
FROM ubuntu

RUN mkdir -m 700 /foobar

RUN chmod a+rx /foobar
```
Docker image created with correct permissions
```
$ docker build -t permission:test .
$ docker run permission:test ls -ld /foobar
drwxr-xr-x 1 root root 4096 Jul  5 21:24 /foobar
```
Running under Singularity shows the incorrect permissions
```
$ singularity build permission.simg docker-daemon://permission:test
$ singularity exec permission.simg ls -ld /foobar
drwx------ 2 root root 3 Jul  5 14:24 /foobar
```

AdamSimpson:
From what I can tell this looks to be an issue with opencontainers/image-tools which is used to create the bundle. I've created opencontainers/image-tools#218 to track this.

GodloveD:
Thanks for the update @AdamSimpson.  Looks like @cyphar thinks we should be using `umoci`.  Might be useful to do some research on what it would take to make that change.

PR #4522, opened 2019-09-25
Set workable permissions on OCI -> sandbox rootless builds

### LINKED ISSUE #4517: sandbox directory cannot be removed if original container has non-writeable directories
### Version of Singularity:

singularity-3.4.0-1.2.el7.x86_64

### Expected behavior

singularity -s build --sandbox /home/aforti/docker_centos_7/image docker://centos:7

Expect like in 3.2.1 or even 2.6.1 singularity to build the sandbox with the paermission of the callers

### Actual behavior
if I leave the cache eanbled the command fails with 

```
2019-09-24 22:37:58,127 | ERROR    | Container execution failed with errors. Error code: 255
2019-09-24 22:37:58,127 | ERROR    | FATAL:   While performing build: conveyor failed to get: Error initializing source oci::307835c385f656ec2e2fec602cf093224173c51119bbebd602c53c3653a3d6eb: Error initializing destination oci::307835c385f656ec2e2fec602cf093224173c51119bbebd602c53c3653a3d6eb: mkdir : no such file or directory
```
if I disable the cache it returns some warnings and then it builds the sandbox with root permissions whether the user napmesspaces is enabled or not.

warnings
```
2019-09-24 22:43:54,556 | WARNING  | 2019/09/24 22:43:49  info unpack layer: sha256:d8d02d45731499028db01b6fa35475f91d230628b4e25fab8e3c015594dc3261
2019-09-24 22:43:54,556 | WARNING  | 2019/09/24 22:43:49  warn rootless{usr/bin/ping} ignoring (usually) harmless EPERM on setxattr "security.capability"
2019-09-24 22:43:54,556 | WARNING  | 2019/09/24 22:43:52  warn rootless{usr/sbin/arping} ignoring (usually) harmless EPERM on setxattr "security.capability"
2019-09-24 22:43:54,556 | WARNING  | 2019/09/24 22:43:52  warn rootless{usr/sbin/clockdiff} ignoring (usually) harmless EPERM on setxattr "security.capability"
```
permissions that do not allow to delete the sandbox

```
-bash-4.2$ rm -rf docker_centos_7/
rm: cannot remove ‘docker_centos_7/image/root/.cshrc’: Permission denied
rm: cannot remove ‘docker_centos_7/image/root/anaconda-ks.cfg’: Permission denied
rm: cannot remove ‘docker_centos_7/image/root/.bashrc’: Permission denied
rm: cannot remove ‘docker_centos_7/image/root/.bash_logout’: Permission denied
rm: cannot remove ‘docker_centos_7/image/root/.bash_profile’: Permission denied
rm: cannot remove ‘docker_centos_7/image/root/.tcshrc’: Permission denied
[......]
```

with the cache
```
mkdir docker_centos_7
export SINGULARITY_TMPDIR=docker_centos_7
export SINGULARITY_CACHEDIR=$SINGULAIRTY_TMPDIR/cache
singularity -s build --sandbox /home/aforti/docker_centos_7/image docker://centos:7
````
without the cache
```
export SINGULARITY_DISABLE_CACHE=1
docker_centos_7
export SINGULARITY_TMPDIR=docker_centos_7
singularity -s build --sandbox /home/aforti/docker_centos_7/image docker://centos:7
```
```
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
ANSI_COLOR="0;31"

```

rpm from EPEL repository

### LINKED ISSUE #4524: Restrictive permissions on non-root sandbox - build failures / cannot delete
**This is one issue extracted from #4517** reported originaly by @afortiorama

### Version of Singularity:

3.4.0 or greater

### Expected behavior

When building a `docker://centos:7` sandbox as a non-root user, the build completes. When finished with the sandbox, it can be removed without issue by the non-root user.

### Actual behavior

There are 2 symptoms...

(1) *IF* the sandbox final location (e.g. `$HOME/test_centos`) is *not* on the same filesystem as `TMPDIR` then the build will fail...

```
dave@piran~> singularity build --sandbox test_centos docker://centos:7
INFO:    Starting build...
Skipping fetch of repeat blob sha256:d8d02d45731499028db01b6fa35475f91d230628b4e25fab8e3c015594dc3261
Copying config sha256:acab94af64effb1f7481666a37788e7a59465e723f0b0fe0a0f458f3f4856638
Storing signatures
2019/09/25 19:47:47  info unpack layer: sha256:d8d02d45731499028db01b6fa35475f91d230628b4e25fab8e3c015594dc3261
2019/09/25 19:47:47  warn rootless{usr/bin/ping} ignoring (usually) harmless EPERM on setxattr "security.capability"
2019/09/25 19:47:48  warn rootless{usr/sbin/arping} ignoring (usually) harmless EPERM on setxattr "security.capability"
2019/09/25 19:47:48  warn rootless{usr/sbin/clockdiff} ignoring (usually) harmless EPERM on setxattr "security.capability"
FATAL:   While performing build: sandbox assemble failed: exit status 1: mv: cannot open '/tmp/sbuild-474633384/fs/etc/gshadow' for reading: Permission denied
mv: cannot open '/tmp/sbuild-474633384/fs/etc/gshadow-' for reading: Permission denied
mv: cannot open '/tmp/sbuild-474633384/fs/etc/shadow' for reading: Permission denied
mv: cannot open '/tmp/sbuild-474633384/fs/etc/shadow-' for reading: Permission denied
```

(2) If the sandbox is built to the same filesystem as `TMPDIR` it will build okay, but it is not possible to delete it due to there being files/directories present with permissions that do not allow deletion:

```
$ singularity build --sandbox test_samefs docker://centos:7
INFO:    Starting build...
Skipping fetch of repeat blob sha256:d8d02d45731499028db01b6fa35475f91d230628b4e25fab8e3c015594dc3261
Copying config sha256:acab94af64effb1f7481666a37788e7a59465e723f0b0fe0a0f458f3f4856638
Storing signatures
2019/09/25 19:52:10  info unpack layer: sha256:d8d02d45731499028db01b6fa35475f91d230628b4e25fab8e3c015594dc3261
2019/09/25 19:52:10  warn rootless{usr/bin/ping} ignoring (usually) harmless EPERM on setxattr "security.capability"
2019/09/25 19:52:11  warn rootless{usr/sbin/arping} ignoring (usually) harmless EPERM on setxattr "security.capability"
2019/09/25 19:52:11  warn rootless{usr/sbin/clockdiff} ignoring (usually) harmless EPERM on setxattr "security.capability"
INFO:    Build complete: test_samefs

$ rm -rf test_samefs
rm: cannot remove 'test_samefs/root/.bash_profile': Permission denied
rm: cannot remove 'test_samefs/root/.bash_logout': Permission denied
rm: cannot remove 'test_samefs/root/.cshrc': Permission denied
rm: cannot remove 'test_samefs/root/.bashrc': Permission denied
rm: cannot remove 'test_samefs/root/.tcshrc': Permission denied
rm: cannot remove 'test_samefs/root/anaconda-ks.cfg': Permission denied
rm: cannot remove 'test_samefs/usr/share/licenses/device-mapper-libs-1.02.149/COPYING': Permission denied
rm: cannot remove 'test_samefs/usr/share/licenses/device-mapper-libs-1.02.149/COPYING.LIB': Permission denied
rm: cannot remove 'test_samefs/usr/bin/sqlite3': Permission denied
rm: cannot remove 'test_samefs/usr/bin/comm': Permission denied
rm: cannot remove 'test_samefs/usr/bin/testgdbm': Permission denied
rm: cannot remove 'test_samefs/usr/bin/ln': Permission denied
rm: cannot remove 'test_samefs/usr/bin/show-installed': Permission denied
rm: cannot remove 'test_samefs/usr/bin/lslocks': Permission denied
rm: cannot remove 'test_samefs/usr/bin/grep': Permission denied
rm: cannot remove 'test_samefs/usr/bin/ipcmk': Permission denied
rm: cannot remove 'test_samefs/usr/bin/repoclosure': Permission denied
rm: cannot remove 'test_samefs/usr/bin/hostid': Permission denied
rm: cannot remove 'test_samefs/usr/bin/glib-compile-schemas': Permission denied
...
```

This is caused by the change to `umoci` for unpacking OCI/docker layers into a rootfs. We previously used a patched `image-tools` for extraction that ensured minimum permissions on files / dirs so that they could be moved (tmp location -> sandbox location) or deleted by a non-root user.

See the `image-tools` patch here: https://github.com/sylabs/image-tools/commit/2814f498056809a9d5baaf76d1d82312180a5888

A similar thing was done in the `docker-extract` libarchive C routines in singularity `2.x` for the same reason.

### PR DESCRIPTION
**Pull request by dtrudg**

We previously used a patched version of `image-tools` to extract OCI layers into a rootfs. This was patched to ensure all files and dirs extracted from a layer had minimum permissions such that when performing a rootless sandbox build:

1. overwrites from subsequent layers succeeded even over files with `000` or similar permissions
2. the rootfs could be moved and deleted by the non-root user/owner

In 3.4.0 we switched to `umoci` for extractions, as it is more faithful to the content of the layers than `image-tools`. `umoci` handles (1) above. However, we need to fix (2) ourself, as `umoci` will faithfully create a rootfs with `000` files in it, which cannot be moved or trivially `rm'd` by the non-root user.

 - Fixes #4524 (which was split from #4517)

### DIFF
--- a/e2e/internal/e2e/fileutil.go
+++ b/e2e/internal/e2e/fileutil.go
@@ -88,3 +88,14 @@ func FileExists(t *testing.T, path string) bool {
 	// todo: we should check if it is a file
 	return PathExists(t, path)
 }
+
+// PathPerms return true if the path (file or directory) has specified permissions, false otherwise.
+func PathPerms(t *testing.T, path string, perms os.FileMode) bool {
+	s, err := os.Stat(path)
+
+	if err != nil {
+		t.Fatalf("While stating file: %v", err)
+	}
+
+	return s.Mode().Perm() == perms
+}
--- a/e2e/regressions/build.go
+++ b/e2e/regressions/build.go
@@ -7,6 +7,7 @@ package regressions
 
 import (
 	"os"
+	"path"
 	"path/filepath"
 	"testing"
 
@@ -56,6 +57,51 @@ func (c *regressionsTests) issue4203(t *testing.T) {
 	)
 }
 
+// This test will build a sandbox, as a non-root user from a dockerhub image
+// that contains a single folder and file with `000` permissions.
+// To verify we force files to be accessible / moveable / removable by the user
+// we check for `700` and `400` permissions on the folder and file respectively.
+func (c *regressionsTests) issue4524(t *testing.T) {
+	sandbox := filepath.Join(c.env.TestDir, "issue_4524")
+
+	c.env.RunSingularity(
+		t,
+		e2e.WithPrivileges(false),
+		e2e.WithCommand("build"),
+		e2e.WithArgs("--sandbox", sandbox, "docker://sylabsio/issue4524"),
+		e2e.PostRun(func(t *testing.T) {
+
+			// If we failed to build the sandbox completely, leave what we have for
+			// investigation.
+			if t.Failed() {
+				t.Logf("Test %s failed, not removing directory %s", t.Name(), sandbox)
+				return
+			}
+
+			if !e2e.PathPerms(t, path.Join(sandbox, "directory"), 0700) {
+				t.Error("Expected 0700 permissions on 000 test directory in rootless sandbox")
+			}
+			if !e2e.PathPerms(t, path.Join(sandbox, "file"), 0600) {
+				t.Error("Expected 0600 permissions on 000 test file in rootless sandbox")
+			}
+
+			// If the permissions aren't as we expect them to be, leave what we have for
+			// investigation.
+			if t.Failed() {
+				t.Logf("Test %s failed, not removing directory %s", t.Name(), sandbox)
+				return
+			}
+
+			err := os.RemoveAll(sandbox)
+			if err != nil {
+				t.Logf("Cannot remove sandbox directory: %#v", err)
+			}
+
+		}),
+		e2e.ExpectExit(0),
+	)
+}
+
 // RunE2ETests is the main func to trigger the test suite
 func RunE2ETests(env e2e.TestEnv) func(*testing.T) {
 	c := &regressionsTests{
@@ -63,7 +109,7 @@ func RunE2ETests(env e2e.TestEnv) func(*testing.T) {
 	}
 
 	return func(t *testing.T) {
-		// https://github.com/sylabs/singularity/issues/4203
-		t.Run("Issue4203", c.issue4203)
+		t.Run("Issue 4203", c.issue4203) // https://github.com/sylabs/singularity/issues/4203
+		t.Run("Issue 4524", c.issue4524) // https://github.com/sylabs/singularity/issues/4524
 	}
 }
--- a/internal/pkg/build/sources/oci_unpack_linux.go
+++ b/internal/pkg/build/sources/oci_unpack_linux.go
@@ -14,12 +14,15 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"path/filepath"
+	"sort"
 
 	"github.com/containers/image/types"
 	"github.com/openSUSE/umoci"
 	umocilayer "github.com/openSUSE/umoci/oci/layer"
 	"github.com/openSUSE/umoci/pkg/idtools"
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/sylabs/singularity/internal/pkg/sylog"
 	sytypes "github.com/sylabs/singularity/pkg/build/types"
 )
 
@@ -68,5 +71,119 @@ func unpackRootfs(b *sytypes.Bundle, tmpfsRef types.ImageReference, sysCtx *type
 	os.RemoveAll(b.Rootfs())
 
 	// Unpack root filesystem
-	return umocilayer.UnpackRootfs(context.Background(), engineExt, b.Rootfs(), manifest, &mapOptions)
+	err = umocilayer.UnpackRootfs(context.Background(), engineExt, b.Rootfs(), manifest, &mapOptions)
+	if err != nil {
+		return fmt.Errorf("error unpacking rootfs: %s", err)
+	}
+
+	// If this is a rootless extraction we need to mangle permissions to fix #4524. This
+	// returns to the <=3.3 permissions on the rootfs, with the exception that umoci
+	// correctly applies permission changes across layers when extracting.
+	if mapOptions.Rootless {
+		sylog.Debugf("Modifying rootless permissions on temporary rootfs")
+		return fixPermsRootless(b.Rootfs())
+	}
+
+	return nil
+}
+
+// fixPermsRootless forces permissions on the rootfs so that it can be easily
+// moved and deleted by a non-root user owner.
+func fixPermsRootless(rootfs string) (err error) {
+	errors := 0
+	err = permWalk(rootfs, func(path string, f os.FileInfo, err error) error {
+		if err != nil {
+			sylog.Errorf("Unable to access sandbox path %s: %s", path, err)
+			errors++
+			return nil
+		}
+		// Directories must have the owner 'rx' bits to allow traversal and reading on move, and the 'w' bit
+		// so their content can be deleted by the user when the rootfs/sandbox is deleted
+		switch mode := f.Mode(); {
+		case mode.IsDir():
+			if err := os.Chmod(path, f.Mode().Perm()|0700); err != nil {
+				sylog.Errorf("Error setting rootless permission for %s: %s", path, err)
+				errors++
+			}
+		case mode.IsRegular():
+			// Regular files must have the owner 'r' bit so that everything can be read in order to
+			// copy or move the rootfs/sandbox around. Also, the `w` bit as the build does write into
+			// some files (e.g. resolv.conf) in the container rootfs.
+			if err := os.Chmod(path, f.Mode().Perm()|0600); err != nil {
+				sylog.Errorf("Error setting rootless permission for %s: %s", path, err)
+				errors++
+			}
+		}
+		return nil
+	})
+
+	if errors > 0 {
+		err = fmt.Errorf("%d errors were encountered when setting permissions", errors)
+	}
+
+	return err
+}
+
+// permWalk is similar to os.Walk - but:
+//   1. The skipDir checks are removed (we never want to skip anything here)
+//   2. Our walk will call walkFn on a directory *before* attempting to look
+//      inside that directory.
+func permWalk(root string, walkFn filepath.WalkFunc) error {
+	info, err := os.Lstat(root)
+	if err != nil {
+		return fmt.Errorf("could not access rootfs %s: %s", root, err)
+	}
+	return walk(root, info, walkFn)
+}
+
+func walk(path string, info os.FileInfo, walkFn filepath.WalkFunc) error {
+	if !info.IsDir() {
+		return walkFn(path, info, nil)
+	}
+
+	// Unlike filepath.walk we call walkFn *before* trying to list the content of
+	// the directory, so that walkFn has a chance to assign perms that allow us into
+	// the directory, if we can't get in there already.
+	if err := walkFn(path, info, nil); err != nil {
+		return err
+	}
+
+	names, err := readDirNames(path)
+	if err != nil {
+		return err
+	}
+
+	for _, name := range names {
+		filename := filepath.Join(path, name)
+		fileInfo, err := os.Lstat(filename)
+		if err != nil {
+			if err := walkFn(filename, fileInfo, err); err != nil {
+				return err
+			}
+		} else {
+			err = walk(filename, fileInfo, walkFn)
+			if err != nil {
+				if !fileInfo.IsDir() {
+					return err
+				}
+			}
+		}
+	}
+	return nil
+}
+
+// readDirNames reads the directory named by dirname and returns
+// a sorted list of directory entries.
+func readDirNames(dirname string) ([]string, error) {
+	f, err := os.Open(dirname)
+	if err != nil {
+		return nil, err
+	}
+	names, err := f.Readdirnames(-1)
+	f.Close()
+	if err != nil {
+		return nil, err
+	}
+	sort.Strings(names)
+	return names, nil
 }

### CODE REVIEW

**Review by jmstover — DISMISSED** [review]
LGTM

File: e2e/imgbuild/imgbuild.go [hunk 1]
```diff
@@ -121,6 +121,12 @@ func (c *imgBuildTests) nonRootBuild(t *testing.T) {
 			buildSpec: "docker://busybox:latest",
 			sandbox:   false,
 		},
+		// #4517 - CentOS7 sandbox as non-root needs permissions forced
```
ikaneshiro: Lets move this into `e2e/regression` since with the new work to build sandboxes in place within `master`, the `mv` currently occurring will no longer exist to trigger an error in this case.

e2e/imgbuild/imgbuild.go
```diff
@@ -121,6 +121,12 @@ func (c *imgBuildTests) nonRootBuild(t *testing.T) {
...
+		{
+			name:      "docker centos7 to sandbox",
+			buildSpec: "docker://centos:7",
```
mem: is this specific image needed for this? If there's something special about this image, please document it.

Otherwise, is it possible to use a smaller image for the same test?

internal/pkg/build/sources/oci_unpack_linux.go
```diff
@@ -68,5 +70,41 @@ func unpackRootfs(b *sytypes.Bundle, tmpfsRef types.ImageReference, sysCtx *type
...
+
+}
+
+// fixPermsRootless forces permissions on the rootfs so that it can be easily
+// moved and deleted by a non-root user owner.
+func fixPermsRootless(rootfs string) (err error) {
+	err = filepath.Walk(rootfs, func(path string, f os.FileInfo, err error) error {
+		// Directories must have the owner 'rx' bits to allow traversal and reading on move, and the 'w' bit
+		// so their content can be deleted by the user when the rootfs/sandbox is deleted
+		switch mode := f.Mode(); {
```
mem, line 102: if there is an error walking into a directory, the `err` argument will contain that error, so that the walk function (this one) can decide what to do about it. In that case the f argument here will be nil (os.FileInfo is an interface) and this code will likely panic.

ikaneshiro (changes requested):
looks good. Fixes the issue when testing locally for me

internal/pkg/build/sources/oci_unpack_linux.go
```diff
@@ -68,5 +70,41 @@ func unpackRootfs(b *sytypes.Bundle, tmpfsRef types.ImageReference, sysCtx *type
...
+		case mode.IsDir():
+			if err := os.Chmod(path, f.Mode().Perm()|0700); err != nil {
+				return fmt.Errorf("error setting rootless sandbox permission for %s: %s", path, err)
```
mem: returning an error will will cause the walk to stop, so other files and directories won't be fixed, possibly leaving the directory in the undesired state. If you log the error but don't return it, Walk will continue fixing the rest of the directories.

internal/pkg/build/sources/oci_unpack_linux.go
```diff
@@ -68,5 +70,41 @@ func unpackRootfs(b *sytypes.Bundle, tmpfsRef types.ImageReference, sysCtx *type
...
+			}
+		case mode.IsRegular():
+			// Regular files must have the owner 'r' bit so that everything can be read in order to
+			// copy or move the rootfs/sandbox around
+			if err := os.Chmod(path, f.Mode().Perm()|0400); err != nil {
+				return fmt.Errorf("error setting rootless sandbox permission for %s: %s", path, err)
```
mem: same here

internal/pkg/build/sources/oci_unpack_linux.go
```diff
@@ -68,5 +70,41 @@ func unpackRootfs(b *sytypes.Bundle, tmpfsRef types.ImageReference, sysCtx *type
...
+	err = filepath.Walk(rootfs, func(path string, f os.FileInfo, err error) error {
+		// Directories must have the owner 'rx' bits to allow traversal and reading on move, and the 'w' bit
+		// so their content can be deleted by the user when the rootfs/sandbox is deleted
+		switch mode := f.Mode(); {
+		case mode.IsDir():
+			if err := os.Chmod(path, f.Mode().Perm()|0700); err != nil {
+				return fmt.Errorf("error setting rootless sandbox permission for %s: %s", path, err)
+			}
+		case mode.IsRegular():
+			// Regular files must have the owner 'r' bit so that everything can be read in order to
```
mem, line 109: nitpick: what the comment says is not strictly true for moves. In order to move a file you need to be able to read and write to a directory, but not the file itself. I guess the read bit is needed for copying (the other part of the comment), and I'm wondering where exactly we need to copy these files. Is this about the case where the rootfs is created in one filesystem and later moved to a different one? That operation is in fact a copy+delete.

e2e/imgbuild/imgbuild.go
dtrudg: I am constructing one now, per discussion with Ian K.

internal/pkg/build/sources/oci_unpack_linux.go
dtrudg: Yes, I am using 'move' here to refer to either same-device move, or inter-device `copy+delete` hence needing `r`

The rootfs is initially extracted into `TMPDIR` so it'll often be an inter-device move (copy+delete).

mem: ah, good, just making sure I wasn't missing something there...

internal/pkg/build/sources/oci_unpack_linux.go
dtrudg: If we can't walk to or chmod any single file then it's very likely we have a fundamental issue leading to a totally broken rootfs... or at least one that will be un-moveable / copyable / removable for the user. From past experience with the singularity `2.x` C code and `3.x` `image-tools` patch I wrote to manage the extraction perms, breakage is all or nothing, and stopping early seems better than spitting out huge error logs that bust out of terminal scrollback on the 1000s of files that are in big containers.

I'd personally rather abandon all hope early, and maybe end up leave something lying in `/tmp` where it has a good chance of auto cleanup on the HPC systems that are generally using rootless sandboxes.... than let the code proceed to anything else.

internal/pkg/build/sources/oci_unpack_linux.go
dctrud: ditto :-)

internal/pkg/build/sources/oci_unpack_linux.go
dtrudg: @ikaneshiro - can we page you for a third opinion?

DrDaveD:
I did a build with this patch to test it, and it works with setuid-root singularity, but I still get this error when I run with -u:
```
INFO:    Starting build...
INFO:    Building into existing container: /cloud/login/dwd/scratch/centos7-sandbox
FATAL:   While performing build: failed to retrieve path for /cloud/login/dwd/scratch/centos7-sandbox: lstat /cloud/login/dwd/scratch/centos7-sandbox: no such file or directory
```

internal/pkg/build/sources/oci_unpack_linux.go
dtrudg: Hmm - the naive attempt to fix this one doesn't work, as this is not walking as I expected... and on catching errors here that's obvious (though it didn't panic in this case).

```
 FATAL:   While performing build: packer failed to pack: while unpacking tmpfs: unable to access sandbox path: /tmp/sbuild-621659804/fs/directory
```

Thanks - I'll have to do some more thinking on this, obviously.

dtrudg:
@DrDaveD - thanks... I think that is another somewhat separate issue. I think I need to ask @cclerget to look at that . I'll try to split the different issues out into separate things tomorrow morning.

e2e/internal/e2e/fileutil.go
```diff
@@ -88,3 +88,17 @@ func FileExists(t *testing.T, path string) bool {
...
+
+// PathPerms return true if the path (file or directory) has specified permissions, false otherwise.
+func PathPerms(t *testing.T, path string, perms os.FileMode) bool {
+	s, err := os.Stat(path)
+
+	if err != nil {
+		t.Fatalf("While stating file: %v", err)
+	}
+
+	if s.Mode().Perm() == perms {
```
mem: this is looking at a boolean to return a boolean. It's equivalent to:

```
return s.Mode().Perm() == perms
```

e2e/regressions/build.go
```diff
@@ -56,6 +57,37 @@ func (c *regressionsTests) issue4203(t *testing.T) {
...
+// To verify we force files to be accessible / moveable / removable by the user
+// we check for `700` and `400` permissions on the folder and file respectively.
+func (c *regressionsTests) issue4517(t *testing.T) {
+	sandbox := filepath.Join(c.env.TestDir, "issue_4203")
+
+	c.env.RunSingularity(
+		t,
+		e2e.WithPrivileges(false),
+		e2e.WithCommand("build"),
+		e2e.WithArgs("--sandbox", sandbox, "docker://sylabsio/issue4517"),
```
mem: I don't want to derail this, but to the best of my understanding, the issue reported in #4517 can be reproduced using this definition file:

```
bootstrap: scratch

%setup
        mkdir ${SINGULARITY_ROOTFS}/d
        touch ${SINGULARITY_ROOTFS}/d/file
        chmod 000 ${SINGULARITY_ROOTFS}/d/file
        chmod a-w ${SINGULARITY_ROOTFS}/d
```

The resulting sandbox cannot be deleted:

```
$ ./builddir/singularity build --fakeroot --sandbox issue-4517.s ./issue4517.def
INFO:    Starting build...
INFO:    Running setup scriptlet
+ mkdir rootfs-120641e2-dff4-11e9-9ab5-54ee757bbcd2/d
+ touch rootfs-120641e2-dff4-11e9-9ab5-54ee757bbcd2/d/file
+ chmod 000 rootfs-120641e2-dff4-11e9-9ab5-54ee757bbcd2/d/file
+ chmod a-w rootfs-120641e2-dff4-11e9-9ab5-54ee757bbcd2/d
INFO:    Build complete: issue-4517.s

$ rm -rf issue-4517.s/
rm: cannot remove 'issue-4517.s/d/file': Permission denied
```

After fixing the permissions on the directory, the sandbox can be removed.

Is that what we are trying to fix / test in this PR or is that something different?

e2e/regressions/build.go
```diff
@@ -56,6 +57,37 @@ func (c *regressionsTests) issue4203(t *testing.T) {
...
+		e2e.PostRun(func(t *testing.T) {
+			defer os.RemoveAll(sandbox)
```
mem: for debugging purposes, I would keep this directory in case of failure.

Also, I would capture any returned errors and log them if not nil.

So:

```
if t.Failed() {
    t.Logf("Test %s failed, not removing directory %s", t.Name(), sandbox)
    return
}

defer func() {
    err := os.RemoveAll(sandbox)
    if err != nil {
        t.Logf("Cannot remove sandbox directory: %#v", err)
    }
}()
```

e2e/regressions/build.go
```diff
@@ -65,5 +97,7 @@ func RunE2ETests(env e2e.TestEnv) func(*testing.T) {
 	return func(t *testing.T) {
 		// https://github.com/sylabs/singularity/issues/4203
 		t.Run("Issue4203", c.issue4203)
+		// https://github.com/sylabs/singularity/issues/4517
+		t.Run("Issue4517", c.issue4517)
```
mem: it's not a problem to make the first argument to t.Run readable: `t.Run("issue 4517", c.issue4517)`

I believe master looks more like the suggestion right now.

e2e/regressions/build.go
dtrudg: To my mind that's something a bit different, as the user is explicitly executing `chmod 000` instructions in a file they are directly providing for the build, but it's worth considering in future.

In #4517 we have no `--fakeroot` and no def file. The user is able to get something that they cannot delete when they have not explicitly provided instructions to create `000` files / dirs - but simply built a local sandbox of a docker image.

mem: I'm not sure it's different.

My understanding of the issue is that the sandbox is created in a temporary directory, and if that directory is located in a different filesystem than the final destination, the move operation from the temporary location to the final location fails because it becomes a copy instead of a move. The reason for failing is because the original container image has a non-readable file.

Starting with the above, you can do:

```
$ singularity build --fakeroot issue-4517.sif ./issue-4517.def
...

$ singularity build --sandbox issue-4517.s issue-4517.sif
INFO:    Starting build...
FATAL:   While performing build: sandbox assemble failed: exit status 1: mv: cannot open '/tmp/sbuild-878592656/fs/d/file' for reading: Permission denied
```

that's the issue as reported, right?

Notice that master does not need this fix (it has already fixed the above issue in a different way). I'm a little concerned that with the proposed fix we would be doing things in a slightly different way in 3.4 and master (3.4 would be mucking around with permissions, master would not). And also about the fact that we are applying this fix based on the bootstrap value (the above steps are equivalent to `boostrap: localimage / from: my.sif`, correct?).

Also note that the test as proposed is not testing for this issue, as it's not making sure that the triggering condition exists (/tmp is on a different filesystem). The test as proposed is testing for the condition I wrote in the original comment (final permissions are "incorrect"), which you are saying is something different.

internal/pkg/build/sources/oci_unpack_linux.go
dtrudg: okay - I've come around to your thinking no @mem - am counting the errors but letting it try the fixes in current code (will push up in a bit).

e2e/regressions/build.go
dtrudg: We still have to mess with the permissions in master... otherwise the sandbox cannot be removed easily by the user as there could by `000` files there etc. master will still needs this fix from this PR, even if there is no `mv` of the sandbox, so it doesn't result in a build failure - so the intention is to do the same perm mangling in master.

The test is correct for verifying perms in master... and I'd argue that (albeit indirectly) it does ensure there wouldn't be a build failure if `/tmp` is on a different filesystem - with the current 3.4 code flow. If you feel very strongly about it looking directly into the `/tmp` rootfs on 3.4 then I can write a test for that - but it'll be more difficult, and that test is only useful on 3.4 - it won't work on master.

Regarding whether we do the permission mangling in the OCI extraction, or later (which would fix your def file example perms)... I'd still argue here that the user in the def file case is explicitly asking for a `chmod 000` in something they are directly providing to singularity - so it's reasonable to honor that. We *have* to do the perm changes on the OCI side though, otherwise we end up with straight builds from docker images being having perms preventing easy copy/move/delete, which isn't at all friendly for the user.

mem: Pardon my brain dump, I'm trying to narrow down what's bothering me.

Perhaps I'm confused regarding which issue we are trying to address.

There are two issues:

1. Failure to build because the move operation fails
2. Unable to remove the resulting sandbox directory easily

If we start from the .def file I provided, at some point during the build there's an intermediate location which contains a directory (d) and a file (d/file) with "funny" permissions (1. above). Somehow are are able to work with that, including reading the files to push them to the squashfs image (why? --fakeroot?) and removing the temporary directory when we are done (IIRC there's code in os.RemoveAll that's doing this, I mean trying to remove a directory, failing, and tweaking permissions so that it works... or it could be --fakeroot, too). In this case, we are happy to leave those funny permissions alone (2. above).

If instead of using the .def file as input, we use another image as input, we are saying we do have to tweak the permissions. I agree that the user is not being as explicit as in the first case, but the user is still saying "use that image as input", not "use that image as input and tweak it a little bit if necessary". That's the bit that's a little uncomfortable in my mind, as introducing this intermediate step is causing the resulting sandbox image to be different.

Thinking out loud: print a warning for every file where we are tweaking permissions, so that the user has a chance to know what happened?

Only 2. above affects master. In 3.4 (this PR) we are fixing both by tweaking the permissions.

And trying to go back to my original comment: this is completely optional, but I would rather avoid hitting the network unless it's necessary. I agree that a test for 1. above is complicated (mount tmpfs filesystem somewhere, point SINGULARITY_TEMPDIR there?) but since hitting the network is not changing this, I would _prefer_ to go with the version that starts from the .def file, creates a .sif and then creates a sandbox out of that .sif so that it can test that the resulting permissions in the sandbox have been correctly tweaked. That would be testing 2. above, right?

Anyhow, I'm not asking for any changes here, just trying to figure things out.

dtrudg:
Noting here that I chatted with @mem out of band on this, with particular reference to the comment above.

The current state of this work is that:
- This code does fix the #4524 and #4532 issues, and could be applied by someone with an urgent need (or they can roll back to 3.3).
- Due to the way that `master` incorporates #4400 we can build a centos7 sandbox as non-root on `master` without the build failing due to permission issues regardless of the location of TMPDIR. This will be much more faithful to the original docker image than if we mangle permissions - which is probably a very good thing.
- Doing the above does mean you may end up with a sandbox you have to manually `chmod` before you delete, but perhaps that is acceptable - or we could put in a more specific permissions fix for certain non-root sandbox circumstances, rather than all non-root OCI rootfs assemblies.
- Whichever way we go for OCI, let's think about how we end up with perm differences for OCI vs other routes (@mem's def file example) and consider if we should tackle anything there.

Upshot is that I don't think this should be rushed for a 3.4.2. It is probably wise there is some thought on it over the weekend, and that it is picked up in the working group meeting Sylabs developers have, and we look to comment from other developers here too.

dtrudg:
There has been some general discussion about the permissions issue, thinking about the context of changes for `3.4.2` and a future `3.5.0`. To summarize, general consensus is that:

1. Not mangling permissions would be the 'right' thing to do, to ensure that a sandbox is faithful to the upstream layer content. Switching to `umoci` was partly to pursue this aim (ref #3880). A faithful sandbox should be an ultimate goal.
2. Incorporating the non-relocation of sandboxes from `master` into a `3.4.2` would prevent build failures, create faithful sandboxes, but leave them unable to be removed without changing permissions. This is against user expectations since `2.x` and we have many users who are likely unfamiliar with the consequences of owned, but `000` files, leaving them to believe that faithful permissions are an 'error' in our extraction.

Note that with regard to the above, as @mem has pointed out above, you can create true `000` files in a sandbox via build from definition file etc. which would lead to the 'surprise' when a user tries to copy/move/delete a sandbox. Therefore we have inconsistency here, and the 'correct' way to resolve it agrees with (1)... that we should be faithful to permissions specified in any source.

With these points in mind, we plan to move forward to:

1. Bring back permission mangling to `3.4.2` to restore the previous behavior so that there is no surprise for users in the `3.4` series. This will likely be accomplished via a (revised?) version of this PR... *to the 3.4 branch only*.
2. *Do not* mangle permissions in `master`. Raise the topic with the community via all routes. Explore if we can e.g. warn users about `000` perms and consequences when building a sandbox, and/or provide a command to remove such a sandbox. Take the temperature of the community for the balance of 'faithful sandbox permissions' vs 'no surprises'.

e2e/regressions/build.go
dtrudg: Hi @mem - I am going to mark this as resolved from the other discussion that has taken place - to clean up the thread for people newly following through it. I've summarized this discussion below, please let me know if there's anything inaccurate, or you want to raise there.

dtrudg:
Setting WIP on this as need to ensure the perm change only comes in on sandbox destinations, and not in oci->sandbox->sif intermediate step to avoid regressing on #3880 in sif (via a different mechanism)

dtrudg:
Well - turns out we cant do the perm change on only sandbox destinations as we will hit #4532 then. Will revert to the full <=3.3. behavior of modifying perms directly on OCI extraction. This means that with regard to #3880 permissions changes on subsequent layers are honored properly during umoci extraction, but then mangled by us... so may not match 100% docker perms, even in a SIF.

#4532 is going to need to be addressed by different means on `master` if we are not going to mangle permissions - as the faithful container permissions after extraction prevent insertion of some files we put / modify in the container.

mem (approved):
It looks good for what we need right now.

Note that because walk is doing a depth-first traversal, and because readDirNames is reading all the directory entries at once, this is going to consume quite a bit of memory for large and deep directory structures (it's going to allocate memory for all the entries in a directory, recurse into one directory, allocate memory for all the entries there and keep going until it finds a directory without any further subdirectories; only when going back up it's going to start releasing the slices it allocated along the way). No need to fix this right now, but it's worth keeping it in mind.

ikaneshiro (approved):

DrDaveD:
Note for those looking at this in the future: this fix is only in the 3.4 branch.  At this time the master (3.5) branch does not always make directories user-writable (and therefore deletable) and the current plan is to not change the permissions there, even though it is not backward compatible, in order to make it more compatible with other OCI compliant tools.

dtrudg:
With regard to the above comment, we have discussion sought on the mailing list here: https://groups.google.com/a/lbl.gov/forum/#!topic/singularity/wScrzv7_fxc

PR #4533, opened 2019-09-26
re-download bin/golangci-lint if version changes

DrDaveD:

Change the make check rule to redownload bin/golangci-lint if the version is not the current desired version.

None, but I had a problem with v1.15.0 complaining about code problems in the master branch that the current version (v.17.1) did not.

### DIFF
--- a/mlocal/frags/Makefile.stub
+++ b/mlocal/frags/Makefile.stub
@@ -17,13 +17,18 @@ collect:
 			>> $(BUILDDIR)/mergeddeps)			\
 	done
 
+CI_LINT_VERSION=1.17.1
 .PHONY: check
 check: $(go_OBJ)
 	@echo " CHECK golangci-lint"
 	$(V)cd $(SOURCEDIR) && \
-		(PATH=$(SOURCEDIR)/bin:$$PATH; \
+		(if [ "`bin/golangci-lint --version 2>/dev/null| \
+			sed 's/.* version //;s/ .*//'`" != $(CI_LINT_VERSION) ]; then \
+		    rm -f bin/golangci-lint; \
+		 fi; \
+		 PATH=$(SOURCEDIR)/bin:$$PATH; \
 		 (command -v golangci-lint >/dev/null || \
-		  curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.17.1) && \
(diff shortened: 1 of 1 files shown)


mlocal/frags/Makefile.stub
```diff
@@ -17,13 +17,18 @@ collect:
 			>> $(BUILDDIR)/mergeddeps)			\
 	done
 
+CI_LINT_VERSION=1.17.1
 .PHONY: check
 check: $(go_OBJ)
 	@echo " CHECK golangci-lint"
 	$(V)cd $(SOURCEDIR) && \
-		(PATH=$(SOURCEDIR)/bin:$$PATH; \
+		(if [ "`bin/golangci-lint --version 2>/dev/null| \
```
mem, line 25: 1. This is too complicated to inline; you can move it to a single separate script
2. Deleting things that are already there, even if we _believe_ we put them there is not nice
3. This is backwards: this is deleting a file so that the test fails so that the download script runs

This is highly coupled to an external process we don't fully control. If you want to make sure that the download script runs, delete the binary on your own, _before_ calling the makefile. That's why there's a check for the existence of golangci-lint.

The problem with this whole process is that it should not have existed in the first place. That's were we took a wrong turn. It was added so that we could provide simple instructions as to what to do before creating a PR. IIRC the first version of this was really "command -v golangci-lint && golangci-lint ...", but that obviously has the undesired side-effect of not running the checks if golangci-lint is not available, which in turn means CI will fail in annoying ways. The current images we use to run the CI process do include golangci-lint in them, this bit here is not necessary for that.

mem (changes requested):
Please address comments

mlocal/frags/Makefile.stub
DrDaveD: @mem I disagree that it's too complicated to inline or that the implementation is backwards, but if you like I could address your comments 1 & 3.  However there's nothing I can do about comment 2 or your other comments after the numbered list.  If I do go through the trouble of moving it to a separate script, will that make you happy enough to accept this idea?  Or, I'd be happy to cancel this PR if you want to come up with an alternative.

I think of this as an automatically maintained tool, bin/golangci-lint is just a cache of the download, and this PR checks for the cache being stale.  It's one less thing for a developer to worry about maintaining.  If somebody wants to use a different version they can install it elsewhere in their PATH.  However in general everybody has to use the same version as the CI in order to pass the CI checks.

dtrudg:
Thanks for the contribution @DrDaveD - the download of golangci-lint has, I understand, been wrapped up in tidying that now happened on #4607

Issue #3653, opened 2019-05-31
Releases are tagged as pre-releases


vsoch:
## Version of Singularity:

3.2.0, 3.2.1

### Expected behavior

Parsing the version string for either of the above official releases returns the release version, not a pre-release.

### Actual behavior

Both official releases are (semver) pre-releases.

For example, here is a snippet to install 3.2.0 (you can do the same for 3.2.1):

```
...
cd ${GOPATH}/src/github.com/sylabs && \
wget https://github.com/sylabs/singularity/releases/download/v3.2.0/singularity-3.2.0.tar.gz && \
tar -xzvf singularity-3.2.0.tar.gz && \
cd singularity && \
./mconfig -p /usr/local && \
make -C builddir && \
$SUDO make -C builddir install
```

The version of Singularity installed from the above should be singularity version 3.2.0, but it's not:

```bash
$ singularity --version
singularity version 3.2.0-1
```
The issue is that when we parse with semver, and we want to do something like compare the version we find to the actual version (3, 2, 1) when it should return True, it returns False (as it's flagged as a pre-release): Here is interaction with the version info string from Singularity Python, using the semver library.

```python
from semver import VersionInfo                                          

# Here I can show what a proper version 3.2.0 looks like, note that prelease is None.
VersionInfo(3,2,0)                                                      
VersionInfo(major=3, minor=2, patch=0, prerelease=None, build=None)

# Now I can import the singularity python client, which parses the version info string from Singularity
from spython.main import Client                                         

# Notice how prelease is 1
Client.version_info()                                                   
VersionInfo(major=3, minor=2, patch=0, prerelease='1', build=None)

# And this is because the pre-release variable -1 is there:
print(Client.version_info())                                            
3.2.0-1

# Here is the issue - checking if 3.2.0 is >= 3.2.0 returns False.

Client.version_info() >= VersionInfo(3,2,0)                             
False
```

And here is the culprit - a fresh download of the .tar.gz (for either of these) has a VERSION file that still has the pre-release:

![image](https://user-images.githubusercontent.com/814322/58711986-a433db00-838d-11e9-8e59-1f47ace3f332.png)

mem:
I believe I have fixed this issue with the associated PR.

dtrudg:
This does fix the semver format. The referenced #4558 needs to be addressed, but is not directly this issue.

PR #4534, opened 2019-09-26
Add get-version script to obtain singularity's version

mem:
Instead of trying to pull this off using a one-liner, move the logic to
a separate script that can do error checking, etc.

In the script, change the format of the version from 3.4.1-1 to 3.4.1+1
to address #3653. Consider cases like 3.4.1-rc.1, where the "rc.1" part
does indicate a prerelease.

fixing or enhancing.

 - Fixes #

### DIFF
--- a/dist/rpm/singularity.spec.in
+++ b/dist/rpm/singularity.spec.in
@@ -29,12 +29,12 @@
 
 Summary: Application and environment virtualization
 Name: singularity
-Version: @PACKAGE_VERSION@
+Version: @PACKAGE_RPM_VERSION@
 Release: @PACKAGE_RELEASE@%{?dist}
 # https://spdx.org/licenses/BSD-3-Clause-LBNL.html
 License: BSD-3-Clause-LBNL
 URL: https://www.sylabs.io/singularity/
-Source: %{name}-%{version}.tar.gz
+Source: %{name}-@PACKAGE_VERSION@.tar.gz
 ExclusiveOS: linux
 # RPM_BUILD_ROOT wasn't being set ... for some reason
 %if "%{sles_version}" == "11"
--- a/mconfig
+++ b/mconfig
@@ -48,7 +48,12 @@ cflags="$CFLAGS -D_FORTIFY_SOURCE=2 -Wformat -Wformat-security -fstack-protector
 ldflags=$LDFLAGS
 
 package_name=singularity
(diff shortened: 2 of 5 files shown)


mem:
To add more info here:

The scripts/get-version script is used by mconfig to obtain the singularity version. This is the reason why it prefers to use the VERSION file if it exists. This is the opposite of the current code, where the output of git describe is preferred. The logic here is that if the VERSION file is present, we are looking at a released tarball, as the VERSION file SHOULD NOT exist in the git repository.

That file captures the tag that was used to create the release. So for tag v3.4.1 the VERSION file will contain 3.4.1.

There are two changes there:

1. The existing code strips things like -rc.1; looking at the code there isn't any indication that this is intentional, and it looks like a bug, because we end up with a release tarball for 3.4.1-rc.1 that says it's 3.4.1 (and the resulting binary will say 3.4.1, too).
2. The existing code is adding a "-1" suffix via a somewhat convoluted path. That suffix is not something that is useful to us (it's useful to a distribution like Debian or RedHat, but we are not a distribution). Issue #3653 is asking us to use a semantic version (as much as possible, we have other issues to address beyond this change), and that means that something like 3.4.1-1 is interpreted as a prerelease. The other option, 3.4.1+1 is interpreted as 3.4.1 with something on top (e.g. a particular build). Since we are not providing binaries, having singularity report its version 3.4.1+1 does not serve a useful purpose. People needed to modify singularity's version to reflect local changes can set up a local repository (and the get-version script will do the right thing in that case) or use the -V flag to set a version. This is also the reason for wanting to keep the "-rc.1" part of release candidates, as that will reported as 3.4.1-rc.1, correctly identifying it as a prerelease under semantic versioning.

dtrudg (approved):
This looks like it satisfies the requirement to me. It's also nice that is has the tarball from `make dist` matching exactly. :+1:

vsoch:
oup, I guess no review was wanted after all? :/ okay.

mem:
> oup, I guess no review was wanted after all? :/ okay.

oops, sorry, my bad. Trigger happy with the merge button :-\

@vsoch I would appreciate any feedback you'd like to provide

vsoch:
It's ok, and if you merged it, I don't think it's worth my time. Sorry.

DrDaveD:
This PR broke the connection between mconfig -V and make rpm.  A proposed fix is in #4558.

PR #4541, opened 2019-09-30
Do not export SingularityCmd

mem:
This is a minimal change to avoid exporting SingularityCmd so that we
don't have to chase its uses all over the place. In order to avoid
touching the few instances outside
github.com/sylabs/singularity/cmd/internal/cli where SingularityCmd is
being used, introduce a couple of functions in that package to provide
access to this command.

There's one used to generate bash completions, where it doesn't really
need access to the command, it just needs a way to request the cobra
bash completions to be generated.

There are a couple of other ones where they need to traverse the command
tree to access every command, and SingularityCmd is just the starting
poont. While the change is still leaving the door open for arbitrary
modifications, it makes it easier to track down these accesses.

fixing or enhancing.

 - Fixes #

### DIFF
--- a/cmd/internal/cli/singularity.go
+++ b/cmd/internal/cli/singularity.go
@@ -7,6 +7,7 @@ package cli
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"os/user"
 	"path"
@@ -24,7 +25,7 @@ import (
 	"github.com/sylabs/singularity/pkg/syfs"
 )
 
-var cmdManager = cmdline.NewCommandManager(SingularityCmd)
+var cmdManager = cmdline.NewCommandManager(singularityCmd)
 
 // CurrentUser holds the current user account information
 var CurrentUser = getCurrentUser()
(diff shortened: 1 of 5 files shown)


cmd/internal/cli/singularity.go
```diff
@@ -243,25 +249,30 @@ func ExecuteSingularity() {
...
-		SingularityCmd.Printf("Run '%s --help' for more detailed usage information.\n",
+		singularityCmd.Printf("Run '%s --help' for more detailed usage information.\n",
 			cmd.CommandPath())
 		os.Exit(1)
 	}
 }
 
+// GenBashCompletionFile
+func GenBashCompletion(w io.Writer) error {
+	return singularityCmd.GenBashCompletion(fn)
```
ikaneshiro: Looks like this should be passing `w` instead of `fn`

mem: LOL, thanks. I edited something at the last minute and forgot about it.

mem:
```suggestion
	return singularityCmd.GenBashCompletion(w)
```

ikaneshiro (approved):

PR #4556, opened 2019-10-02
Use image extractor for v2 images

### LINKED ISSUE #4550: Extremely slow building of sandbox from image: dd with bs=31, use iflag=skip_bytes instead.
I installed singularity 3.3.0-1 on a HPC cluster from source (I'm the admin) and am trying to convert a .sif to a sandbox to check how runtime performance can improve with --userns, avoiding the temporary extraction to $TMPDIR.

So I do this:
```
singularity pull shub://hariszaf/pema:v1
singularity build --sandbox pema_v1 pema_v1.sif
```
And then I wait … and wait … and notice that I am waiting for _this_ to finish:
```
dd bs=31 skip=1 if=pema_v1.sif of=/work/tmp/node002.2019-10-02-133002.rrztest.16897/sbuild-838622514/trim.squashfs667647721
```
This is excruciatingly slow. A copy in 31-byte blocks? This takes ages to copy 2G around on a parallel filesystem that easily gives you around 1G/s on a single stream.

Compare this:
```
-bash-4.2$ dd bs=1M skip=31 iflag=skip_bytes  if=pema_v1.sif  of=pema_v1.squashfs
2699+1 records in
2699+1 records out
2830450688 bytes (2.8 GB) copied, 3.69102 s, 767 MB/s
```
While still waiting for `singularity build` to finish, I created the raw squashfs
this way and ran it through `unsquashfs`. This is _a_lot_ faster. It's slower without bs=1M, but still orders of magnitude faster than working in 31-byte-blocks!

Please consider changing the inefficient `dd` call. I might end up having the users use the .sif image via
```
TMPDIR=/dev/shm singularity run --userns pema_v1.sif
```
anyway, as that only has 10 seconds overhead (obviously uses something more efficient than the build command), but maybe I want to offer them the sandbox directories, if user permissions don't cause trouble in there. And in an HPC context this wildly inefficient I/O just looks weird.

Is it maybe a bug that `singularity build` uses a different routine to extract the .sif than `singularity run --userns`? An interesting difference is that the latter breaks on SIGSTOP (interrupted system call on putting it into the backgroud in the shell), while the former is fine with that. Both could learn from each other, I suppose.

PS: While composing this report and testing things, the initial `build` advanced to 2.2G of extracted squashfs. It will finish _eventually_. In the end, it's silly having to make a copy of the whole image just to skip a header. If `unsquashfs` supported an input offset … and how did we arrive at 31 bytes to begin with? Not even 32, let alone a sane number like 256, 4K or so.

ikaneshiro:
Previously this involved using a combination of `dd` and `unsquashfs` to
perform the same task. This also includes some refactoring of the
various packers in the build package.

 - Fixes #4550 

### DIFF
--- a/internal/pkg/build/sources/conveyorPacker_local.go
+++ b/internal/pkg/build/sources/conveyorPacker_local.go
@@ -12,7 +12,6 @@ import (
 	"github.com/sylabs/singularity/internal/pkg/sylog"
 	"github.com/sylabs/singularity/pkg/build/types"
 	"github.com/sylabs/singularity/pkg/image"
-	"github.com/sylabs/singularity/pkg/util/loop"
 )
 
 // LocalConveyor only needs to hold the conveyor to have the needed data to pack
@@ -40,8 +39,6 @@ func GetLocalPacker(src string, b *types.Bundle) (LocalPacker, error) {
 		return nil, err
 	}
 
-	info := new(loop.Info64)
-
 	switch imageObject.Type {
 	case image.SIF:
 		sylog.Debugf("Packing from SIF")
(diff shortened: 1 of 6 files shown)


dtrudg (approved):
This looks good to me.

I have verified the functionality on some squashfs images from https://depot.galaxyproject.org/singularity/old/ and the speed up is indeed a nice one:

*With this patch*
```
dave@piran~/testdd> /bin/time singularity build --sandbox testout ansible:1.9.4--py27_0
INFO:    Starting build...
INFO:    Build complete: testout
0.65user 0.32system 0:00.32elapsed 298%CPU (0avgtext+0avgdata 150504maxresident)k
0inputs+410872outputs (0major+37918minor)pagefaults 0swa
```

*With 3.4.1*
```
dave@piran~/testdd> /bin/time singularity build --sandbox testout ansible:1.9.4--py27_0
INFO:    Starting build...
INFO:    Build complete: testout
0.79user 1.80system 0:01.92elapsed 134%CPU (0avgtext+0avgdata 150460maxresident)k
0inputs+410872outputs (0major+38064minor)pagefaults 0swaps
```

cclerget (approved):
LGTM

PR #4558, opened 2019-10-02
Restore connection between mconfig -V and make rpm

DrDaveD:

PR #4534 which enabled semantic versioning internally to singularity had a side effect of breaking the connection between `mconfig -V` and the version used by `make rpm`.  This PR fixes that connection while attempting to preserve the semantic versioning, where a hyphen indicates a pre-release.  Rpms however do not fully work with the official definition of semantic versioning, because for it the hyphen is reserved to separate the version number from the release number.  PR #4534 changed the internal separator to be colon, and this PR preserves that.  This PR also preserves the behavior of #4534 of converting hyphens in the `mconfig -V` version to dots for rpms only.

I haven't tested with Debian packaging.  I wouldn't be surprised if #4534 broke that as well.

 - Fixes #

Attn: @singularity-maintainers, @mem, @vsoch


mlocal/frags/Makefile.stub
```diff
@@ -73,17 +73,18 @@ testall: check test
...
 	      --define '_prefix $(RPMPREFIX)' \
 	      --define '_sysconfdir $(RPMPREFIX)/etc' \
 	      --define '_localstatedir $(RPMPREFIX)/var' \
 	      --define '_mandir $(RPMPREFIX)/share/man' \
-	      $(SOURCEDIR)/singularity-$(VERSION).tar.gz; \
+	      $(SOURCEDIR)/singularity-$$short_version.tar.gz; \
 	  else \
 	    rpmbuild $(RPMCLEAN) -ta \
-	      $(SOURCEDIR)/singularity-$(VERSION).tar.gz; \
+	      $(SOURCEDIR)/singularity-$$short_version.tar.gz; \
```
jmstover, line 86: `$$short_version` ?  Should that not just be `$short_version`? Or `${short_version}`

jmstover (changes requested):

mlocal/frags/Makefile.stub
DrDaveD: No, in make two dollars translates into one dollar passed to the shell.  It works as it is.

jmstover: Ahh... right. That's to be evaluated in the shell command there.

jmstover (approved):

mem (changes requested):
Give me a chance to check what this is doing

dtrudg:
@DrDaveD could the `-` be simply swapped for `~` per http://lists.rpm.org/pipermail/rpm-list/2014-November/001659.html?

DrDaveD:
@dctrud Good catch, I wasn't aware of that.  I changed the translation of hyphen for rpm to a tilde instead of a dot.  The article said it applied only to rpm 4.10, but I found it also worked for ordering on el6 which is rpm 4.8.  Red Hat must have backported that functionality.

I have seen tildes also in Debian versions, and I found [documentation](https://www.debian.org/doc/debian-policy/ch-controlfields.html) that says tildes are always considered to be earliest in Debian sorting.  So what do people think about using tildes in the singularity internal version as well?  That is, always use tildes as the semantic version pre-release, and go back to using hyphen as the separator between version and release?  Then there would not need to be any special translation for rpms nor a singularity-specific convention of colon as the version & release separator (or at least I personally haven't seen a colon used for that anywhere else).

A related consideration is that `singularity --version` always shows the internal version, even when the rpm package version is using the rpm conventions.

dtrudg:
I think we need to stick with hyphen internally. We want to adhere to the semver grammar here. Using a tilde would be a regression on #3653 from what I can tell.

DrDaveD:
@mem Can you please check this?

DrDaveD:
This breaks rpm devel compiles, I don't think we should put this off for so long.

jmstover (approved):

mem:
@DrDaveD can you provide an example of how this is breaking? The CI process does build RPMs, and that seems to be happy with it. If you can describe what breaks and what the fix is doing, it would be easier to move forward.

DrDaveD:
It breaks if anybody does 'mconfig -V' to set a specific version number, as I always do when I build rpms for development.  It's also done as part of the EPEL/Fedora build process, so this is an important feature to have working.  This PR makes mconfig -V work again for rpms.  It only happens to work now if the default version number is selected.

mem:
> It breaks if anybody does 'mconfig -V' to set a specific version number, as I always do when I build rpms for development. It's also done as part of the EPEL/Fedora build process, so this is an important feature to have working. This PR makes mconfig -V work again for rpms. It only happens to work now if the default version number is selected.

Can you paste an error or a log? Or what the result looks like?

DrDaveD:
For example I do `mconfig -V 3.4.2:0.0.20191021140408.dwd.497.g45e4c99git -p / -e /usr` which results in 
```
...
    - version: 3.4.2:0.0.20191021140408.dwd.497.g45e4c99git
```
then when I do a make rpm  it tries to build with the default version and fails:
```
+ make -C builddir rpm
make: Entering directory `/cloud/login/dwd/gopath/src/github.com/sylabs/singularity/builddir'
 DIST setup VERSION: 3.4.2+497-g45e4c99
 DIST create tarball: singularity-3.4.2+497-g45e4c99.tar.gz
 BUILD RPM
+ cd /cloud/login/dwd/gopath/src/github.com/sylabs/singularity
+ '[' -n '' ']'
+ rpmbuild -ta /cloud/login/dwd/gopath/src/github.com/sylabs/singularity/singularity-3.4.2:0.0.20191021140408.dwd.497.g45e4c99git.tar.gz
error: File /cloud/login/dwd/gopath/src/github.com/sylabs/singularity/singularity-3.4.2:0.0.20191021140408.dwd.497.g45e4c99git.tar.gz: No such file or directory
/usr/bin/cat: /cloud/login/dwd/gopath/src/github.com/sylabs/singularity/singularity-3.4.2:0.0.20191021140408.dwd.497.g45e4c99git.tar.gz: No such file or directory
error: File /cloud/login/dwd/gopath/src/github.com/sylabs/singularity/singularity-3.4.2:0.0.20191021140408.dwd.497.g45e4c99git.tar.gz: No such file or directory
/usr/bin/cat: /cloud/login/dwd/gopath/src/github.com/sylabs/singularity/singularity-3.4.2:0.0.20191021140408.dwd.497.g45e4c99git.tar.gz: No such file or directory
error: Failed to read spec file from /cloud/login/dwd/gopath/src/github.com/sylabs/singularity/singularity-3.4.2:0.0.20191021140408.dwd.497.g45e4c99git.tar.gz
make: *** [rpm] Error 1
make: Leaving directory `/cloud/login/dwd/gopath/src/github.com/sylabs/singularity/builddir'
```

mem:
With the provided information, this is my interpretation of what needs to be fixed:

https://github.com/sylabs/singularity/pull/4661

dtrudg:
I'm going to close this one as discussion has now moved onto #4661 - we can reopen if necessary.

PR #4562, opened 2019-10-03
Check the remote.yaml permission on runtime

WestleyK:

This PR adds a check (on runtime) to ensure the correct file permission
on `~/.singularity/remote.yaml`. Just like we check/create the
`~/.singularity` directory.

 - Fixes #


internal/pkg/util/fs/helper.go
```diff
@@ -14,9 +14,36 @@ import (
...
 	"syscall"
 
+	"github.com/sylabs/singularity/internal/pkg/sylog"
 	"golang.org/x/sys/unix"
 )
 
+// EnsureFilePrivate takes a file path, and 1. Creates it with
+// the specified permission, or 2. ensures a file path is the
+// specified permission.
+func EnsureFilePermission(fn string, mode os.FileMode) error {
```
dtrudg: I'd prefer something without the side effect of creating an empty file tbh. If this ensures a file is present with certain permission maybe `EnsureFileAndPermission` is a better name for what it's doing.

Also, please could you add a test to `helper_linux_test.go` for things added here - to keep up the nice coverage on these helpers.

internal/pkg/util/fs/helper.go
```diff
@@ -14,9 +14,36 @@ import (
...
+	defer fs.Close()
+
+	// check and fix permissions
+	fsinfo, err := fs.Stat()
+	if err != nil {
+		return err
+	}
+
+	if currentMode := fsinfo.Mode(); currentMode != mode {
+		sylog.Warningf("File mode (%o) on %s needs to be %o, fixing that...", currentMode, fn, mode)
```
dtrudg, line 38: This message is a bit generic - I think if we change the perms we should give at least a hint of why they need to be `mode` for a specific file - which needs to be handled elsewhere.

dtrudg (changes requested):
Hi @WestleyK 

On this one I don't really like the approach of creating an empty file with the correct permissions. It's not a super obvious thing to expect and opens us up to having to consider elsewhere that `remote.yaml` might be present but empty (we should handle cases when it doesn't contain what we expect, but it's still a little unusual).

I'm also more of the view that we should probably just check the permission and error out if it is too open, telling the user when it is wrong, rather than fix it. We usually create and manage the file ourselves, ensuring correct permissions. The situations the permissions could be changes that I can think of off the top of my head are:

- The user specifically `chmod`'d the file for some reason. We fix it, but don't really say why here - which is confusing for them.
- The user copied their directory somewhere in a way which didn't preserve permissions. We might fix this one file, but we're not then prompting them to investigate what else might be open too wide, e.g. a cache dir with SIFs holding private data etc.

With those points made, it'd be good to get @ikaneshiro and/or @mem 's opinions as they might differ.

WestleyK:
> On this one I don't really like the approach of creating an empty file with the correct permissions.

Before creating/checking the file/permission, we first check to see if that file exist, if so, then ensure the correct permission. If not, then ignore it.

internal/pkg/util/fs/helper.go
WestleyK: Okay, I will change it to `EnsureFileAndPermission`

dtrudg: or `EnsureFileWithPermission` ... is possibly nicer ... up to you :-)

dtrudg:
> > On this one I don't really like the approach of creating an empty file with the correct permissions.
> 
> Before creating/checking the file/permission, we first check to see if that file exist, if so, then ensure the correct permission. If not, then ignore it.

Gotcha - I misread. Apologies.

internal/pkg/util/fs/helper.go
WestleyK: I added debug message where this is called, should it more verbose?

internal/pkg/util/fs/helper_linux_test.go
```diff
@@ -16,6 +16,125 @@ import (
...
+
+	//
+	// First test: Ensure a already-existing file is the
+	// correct permssion.
+	//
+
+	existFile := filepath.Join(tmpDir, "already-exists")
+
+	// Create the test file.
+	fp, err := os.OpenFile(existFile, os.O_CREATE, 0755)
```
dtrudg, line 36: This test will fail on a system with a non-standard umask, as 0755 is before umask. Adding a separate `chmod` should make it safer.

internal/pkg/util/fs/helper.go
```diff
@@ -14,9 +14,36 @@ import (
...
 
+	"github.com/sylabs/singularity/internal/pkg/sylog"
 	"golang.org/x/sys/unix"
 )
 
+// EnsureFileWithPermission takes a file path, and 1. Creates it with
+// the specified permission, or 2. ensures a file is the specified
+// permission.
+func EnsureFileWithPermission(fn string, mode os.FileMode) error {
+	fs, err := os.OpenFile(fn, os.O_CREATE, mode)
```
dtrudg, line 25: This is not guaranteed to produce a file with the requested mode, dependent on umask. It could create a file with more restrictive permissions than requested, once the umask is applied.

Note in the old removed code the umask is being set explicitly. To guarantee mode you would need an open umask, or to do a separate `chmod` step here.

dtrudg (changes requested):

internal/pkg/util/fs/helper.go
WestleyK: I dont want to add the umask change here, because depending on the specified file mode, the umask will require a different value, (afaik) so a `Chmod` would be better in this case, imo.

Right after creating/opening the file, we check the file permission, if its incorrect, then we use chmod. Line ~39.

internal/pkg/util/fs/helper.go
dtrudg: Okay - it's not ideal as generally people we only see this Warning (they won't see debug messages), but we'll live with that.

internal/pkg/util/fs/helper.go
dtrudg: Yeah - understood. It's a little unusual to me that we would then see a message about a permission being wrong for a file we just created, but I can live with that here.

dtrudg (approved):

internal/pkg/util/fs/helper.go
WestleyK: Makes sense, but also, that unusual warning should only appear with a unusual umask :wink: 

Would you like it better if we always run `Chmod` right after opening/creating the file? _that may not be the best solution as most of the time we will not need to run Chmod..._

PR #4569, opened 2019-10-03
Remove the 'pull' overridden image at the last moment

### LINKED ISSUE #4561: If 'pull' fails, the overridden image will be deleted
### Version of Singularity:

```
$ singularity version
3.4.1+331-gcae120098
```

### Expected behavior

Image should not be deleted if the pull fails, like:

```
$ echo "Hi" > foo.sif
$ singularity pull -F foo.sif library://foo/bar/nothing:latest
FATAL:   While pulling library image: image foo/bar/nothing:latest (amd64) does not exist in the library

$ cat foo.sif
Hi
```

### Actual behavior

`foo.sif` is deleted when the pull fails:

```
$ echo "Hi" > foo.sif
$ singularity pull -F foo.sif library://foo/bar/nothing:latest
FATAL:   While pulling library image: image foo/bar/nothing:latest (amd64) does not exist in the library

$ cat foo.sif
cat: foo.sif: No such file or directory
```

See above.

```
$ cat /etc/os-release
NAME="Ubuntu"
VERSION="18.04.2 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.2 LTS"
VERSION_ID="18.04"
VERSION_CODENAME=bionic
UBUNTU_CODENAME=bionic
```

Source.

WestleyK:

This PR changes the `pull` to only remove the overridden image at the
last moment. So if the `pull` fails, the overridden image will not be
deleted/lost.

 - Fixes #4561

### DIFF
--- a/internal/app/singularity/registry.go
+++ b/internal/app/singularity/registry.go
@@ -8,7 +8,9 @@ package singularity
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 
 	scs "github.com/sylabs/scs-library-client/client"
 	"github.com/sylabs/singularity/internal/pkg/client/cache"
@@ -58,29 +60,70 @@ func (l *Library) Pull(ctx context.Context, from, to, arch string) error {
 		return fmt.Errorf("could not get image info: %v", err)
 	}
 
-	if l.cache.IsDisabled() {
-		// don't use cached image
-		err := l.pullAndVerify(ctx, imageMeta, libraryPath, to, arch)
-		if err != nil {
-			return fmt.Errorf("unable to download image: %v", err)
+	dst, tmpName, err := func() (string, string, error) {
+		dst := to
+		if !l.cache.IsDisabled() {
+			imageName := uri.GetName("library://" + libraryPath)
+			dst = l.cache.LibraryImage(imageMeta.Hash, imageName)
+
+			// here we can check if the file is already in
+			// the cache
+			if _, err := os.Stat(dst); err == nil {
+				// we have the file in the cache, return
+				// the same name for the final
+				// destination and the temporary
+				// location to signal that no rename is
+				// necessary
+				return dst, dst, nil
+			}
 		}
-	} else {
-		// check and use cached image
-		imageName := uri.GetName("library://" + libraryPath)
-		err := l.copyFromCache(imageMeta.Hash, imageName, to)
+
+		tmpHandle, err := ioutil.TempFile(filepath.Dir(dst), filepath.Base(dst)+".")
 		if err != nil {
-			if err != errNotInCache {
-				return fmt.Errorf("could not copy image from cache: %v", err)
-			}
-			imagePath := l.cache.LibraryImage(imageMeta.Hash, imageName)
-			err := l.pullAndVerify(ctx, imageMeta, libraryPath, imagePath, arch)
-			if err != nil {
-				return fmt.Errorf("could not pull image: %v", err)
-			}
-			err = l.copyFromCache(imageMeta.Hash, imageName, to)
-			if err != nil {
-				return fmt.Errorf("could not copy image from cache: %v", err)
-			}
+			return "", "", fmt.Errorf("unable to create temporary image: %w", err)
+		}
+		tmpHandle.Close()
+
+		tmpName := tmpHandle.Name()
+
+		// This is racy
+		if err := os.Remove(tmpName); err != nil {
+			return "", "", fmt.Errorf("unable to remove temporary file %s: %w", tmpName, err)
+		}
+
+		go interruptCleanup(tmpName)
+
+		return dst, tmpName, nil
+	}()
+
+	if err != nil {
+		return fmt.Errorf("unable to obtain intermediate location for %s: %w", to, err)
+	}
+
+	// if the tmpName is the same as dst, that means we are
+	// looking at a file present in the cache, so we can skip
+	// downloading it
+	if tmpName != dst {
+		sylog.Debugf("Downloading to %s for intermediate destination %s and final destination %s", tmpName, dst, to)
+		// Download the image, either to the cache, or to the
+		// temporary location
+		if err := l.pullAndVerify(ctx, imageMeta, libraryPath, tmpName, arch); err != nil {
+			return fmt.Errorf("unable to download image: %s", err)
+		}
+
+		sylog.Debugf("Renaming temporary file %s to %s", tmpName, dst)
+		os.Rename(tmpName, dst)
+	}
+
+	// now we either have the image in the correct location (dst ==
+	// to) or we have the image in the cache (dst != to). In the
+	// later case we need to copy from the cache to the final
+	// destination.
+	if dst != to {
+		os.Remove(to)
+		sylog.Debugf("Copying %s to %s", dst, to)
+		if err := fs.CopyFile(dst, to, 0777); err != nil {
+			return fmt.Errorf("cannot copy cache element %s to final destination %s: %w", dst, to, err)
 		}
 	}
 
@@ -132,6 +175,12 @@ func (l *Library) copyFromCache(hash, name, to string) error {
 
 	if !exists {
 		return errNotInCache
+	}
+
+	// Remove the 'to' image if exists, (before copying from cache).
+	err = os.Remove(to)
+	if err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("unable to remote old image to overide: %s", err)
 	}
 
 	from := l.cache.LibraryImage(hash, name)
--- a/cmd/internal/cli/pull.go
+++ b/cmd/internal/cli/pull.go
@@ -8,10 +8,8 @@ package cli
 import (
 	"context"
 	"os"
-	"os/signal"
 	"path/filepath"
 	"runtime"
-	"syscall"
 
 	"github.com/spf13/cobra"
 	"github.com/sylabs/scs-library-client/client"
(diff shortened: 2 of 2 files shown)


internal/app/singularity/registry.go
```diff
@@ -134,6 +134,15 @@ func (l *Library) copyFromCache(hash, name, to string) error {
 		return errNotInCache
 	}
 
+	sylog.Debugf("Removing %s (if exists) before copying cached image", to)
```
dtrudg: It's a bit odd to put a message saying we might or might not be doing something. Can you make it explicit only if we *actually* remove the file?

dtrudg (changes requested):
A small comment on a change to the debug message.

More generally - if the *cache is disabled* (i.e. we are downloading to `to` directly) what happens:

  (a) when there is an existing SIF in the location - it's not explicitly removed now, so is it overwritten as expected?
 (b) when I press `Ctrl-C` during the download? I think there will be an image left there that is incomplete because we only handle deletion after an interrupt in `copyFromCache` now... and not at all if the cache is disabled?

internal/app/singularity/registry.go
```diff
@@ -134,6 +134,15 @@ func (l *Library) copyFromCache(hash, name, to string) error {
...
+	err = os.Remove(to)
```
sashayakovtseva, line 181: I don't think this function is the right place for removal. Instead, you should remove image at a higher level, e.g. in `Pull`.

sashayakovtseva (changes requested):

internal/app/singularity/registry.go
WestleyK: Well, it was at a very high level before, (in the cli code). This was a issue because if a download fails, the overridden image will be lost. And that the point of this PR, to only remove the image at the very last moment. But I'm open to suggestions if you think otherwise...

WestleyK:
@dctrud, 

> More generally - if the _cache is disabled_ (i.e. we are downloading to `to` directly) what happens:
> 
> (a) when there is an existing SIF in the location - it's not explicitly removed now, so is it overwritten as expected?

Yes, as of now, the it will be overwritten when downloading the new container. So if you press crtl-c during the download, the overwritten image will be lost. ~What is should do is remove the overwritten image before downloading it, to ensure the correct permission of `755`.~ (Fixed)

> (b) when I press `Ctrl-C` during the download? I think there will be an image left there that is incomplete because we only handle deletion after an interrupt in `copyFromCache` now... and not at all if the cache is disabled?

No, the partly downloaded image will be removed by `pullAndVerify()`, and since we are downloading directly to `to`, that will be the cleanup file.

internal/app/singularity/registry.go
mem: The other day I was working on something like this but I haven't pushed a PR yet because there's some very awkward logic around this.

My approach is to download to a temporary file in the same directory as the target. Check if the target exists _before_ downloading, if it's there check the --force flag and _not_ delete anything. Proceed with the download and check _again_ after the file has been downloaded, and if the --force flag was passed, nuke the existing file and rename the temporary file to the final name. Checking twice is necessary because the whole code around this is a little awkward, in the sense that the involved code does not take an io.Writer as parameter but instead everyone is dead set on getting the filename.

If the involved code took an io.Writer, things would not be simpler, you still have to download to a temporary location and check twice if the file is there, but at least the code would be a tiny bit less racy. It would make things simpler in the sense that you would not have to check for clobbering files all over the place.

So, I tend to agree with Sasha: the overwrite check should happen elsewhere. The reason why it's here is because this code contains some UI elements, as it will output warnings and what not. When I changed the code to work with a temporary file instead of the actual destination, I started to get messages on the console about the temporary filename, and that's actually the reason for not uploading my PR yet.

(and the other half of the awkwardness is the cache code, that seems to be getting in the middle of the download, as at least in one case it seemed to be looking at the temporary filename I was passing)

dtrudg:
Hi @WestleyK - with the force push I can't easily see what has changed vs previous versions of the PR. Can you just quickly mention that / any response to @mem / @sashayakovtseva 's thoughts?

internal/app/singularity/registry.go
WestleyK: Download to a temporary file in the same directory as the target, seems reasonable, I'll work on implementing that...

WestleyK:
> Hi @WestleyK - with the force push I can't easily see what has changed vs previous versions of the PR.

Sorry, I'm trying to keep this/other PRs cleaner, with less commits. But it make sense to not force push all commits. Then after the PR is ready for review, squash the commits. _I'll try that..._

internal/app/singularity/registry.go
```diff
@@ -58,9 +58,16 @@ func (l *Library) Pull(ctx context.Context, from, to, arch string) error {
 		return fmt.Errorf("could not get image info: %v", err)
 	}
 
+	// don't use cached image
```
sashayakovtseva: Why commenting this? IMO it is obvious enough... 
Generally comments should be used in places where logic is not clear by simply reading the code. Here it is redundant, I'd prefer not to add it.

internal/app/singularity/registry.go
sashayakovtseva: And still you have this at a higher level, no need to remove here.

internal/app/singularity/registry.go
```diff
@@ -58,9 +58,16 @@ func (l *Library) Pull(ctx context.Context, from, to, arch string) error {
...
 	if l.cache.IsDisabled() {
-		// don't use cached image
-		err := l.pullAndVerify(ctx, imageMeta, libraryPath, to, arch)
+		// Since we are downloading it directly to the output image, we
+		// need to remove the overridden image if exists.
+		err := os.Remove(to)
```
sashayakovtseva: What if next pull returns an error? In that case will not get image pull but also remove the old one.

internal/app/singularity/registry.go
```diff
@@ -134,6 +141,15 @@ func (l *Library) copyFromCache(hash, name, to string) error {
...
 	}
 
+	// Remove the 'to' image if exists, (before copying from cache).
+	err = os.Remove(to)
+	if err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("unable to remote old image to overide: %s", err)
+	}
+
+	// Remove the not-fully copied image if interrupted.
+	go interruptCleanup(to)
```
sashayakovtseva: maybe remove this one from here and `pullAndVerify` and keep it in a single place?

sashayakovtseva (dismissed):

dtrudg:
Hi @WestleyK - I keep getting PR reminders for review on this one, but it's still marked WIP.

Can you advise if it is truly still WIP? If not, can you comment r.e. Sasha's comments. As she is away for a few days and we are trying to tidy up into 3.5 I'm happy to dismiss and re-review (or someone else can), as long as we know your general response there.

Thanks!

WestleyK:
Yes this is still a WIP. I'll try to fix this up in time for 3.5, if not then we dont have to include this. I'll re-request your review when ready.

internal/app/singularity/registry.go
WestleyK: So, "download to a temporary file in the same directory as the target." only seems useful if the cache is disabled
(`--disable-cache`). Unless I'm not understanding something? This seems like a good place to remove the final image (if overriding it), at the very last moment, just as the PR is suppose to fix.

In case there is some misunderstanding, heres the order-of-operation: (current behavior in this PR)

CLI:
 - check if the image already exists,
    - if force is false, abort
    - else if force is true, continue

PULL:
 - is cache disabled? if yes: pull to a temporary file in the same directory, then remove the 'to' image and rename the temp file to the 'to' image; else
    - check if the image is in cache (if not, download it)
    - remove the overwritten image (if exists)
    - copy the image from cache to the final destination

WestleyK (commented):
Now should be re-ready for review!

dtrudg:
Unfortunately this is failing on `TestE2E/CMDENVVARS/read-only_cache_directory` - which is a test to ensure we can pull if the cache directory is set to a read-only location.

mem:
Fixing this.

mem:
> Unfortunately this is failing on `TestE2E/CMDENVVARS/read-only_cache_directory` - which is a test to ensure we can pull if the cache directory is set to a read-only location.

It's doing the right thing, it seems:

    $ rm -rf /tmp/foobar
    $ install -d -m 0555 /tmp/foobar
    $ SINGULARITY_CACHEDIR=/tmp/foobar builddir/singularity --debug pull --force library://alpine
    DEBUG   [U=1001,P=9613]    persistentPreRunE()           Singularity version: 3.4.2+532-g4ce8aceef
    DEBUG   [U=1001,P=9613]    handleConfDir()               /home/mem/.singularity already exists. Not creating.
    DEBUG   [U=1001,P=9613]    handleRemoteConf()            Ensuring file permission of 0600 on /home/mem/.singularity/remote.yaml
    DEBUG   [U=1001,P=9613]    Pull()                        Downloading to alpine_latest.sif.849867644 for intermediate destination alpine_latest.sif and final destination alpine_latest.sif
    INFO    [U=1001,P=9613]    pullAndVerify()               Downloading library image
     2.59 MiB / 2.59 MiB [==================================================================================================================================================] 100.00% 4.24 MiB/s 0s
    DEBUG   [U=1001,P=9613]    Pull()                        Renaming temporary file alpine_latest.sif.849867644 to alpine_latest.sif
    DEBUG   [U=1001,P=9613]    Verify()                      Verifying hash: SIFHASH:
    daccdb1b531e41633fa5c187e9b6a7ea9a1188758db40f95c803abf6df19e29b4a92840b43a6af19f42471fc72fed7a3
    VERBOSE [U=1001,P=9613]    getSignerIdentity()           Key not found in local keyring, checking remote keystore: F38D871E
    VERBOSE [U=1001,P=9613]    getSignerIdentity()           Found key in remote keystore: F38D871E
    WARNING [U=1001,P=9613]    IsSigned()                    Container might not be trusted; run 'singularity verify alpine_latest.sif' to show who signed it
    INFO    [U=1001,P=9613]    Pull()                        Download complete: alpine_latest.sif

and the test passes, too:

    $ scripts/e2e-test -v -run TestE2E/CMDENVVARS/read-only_cache_directory
    === RUN   TestE2E
    === RUN   TestE2E/CMDENVVARS
    === RUN   TestE2E/CMDENVVARS/read-only_cache_directory
    --- PASS: TestE2E (2.82s)
        e2e.go:14: Running E2E tests for Singularity
        suite.go:121: Path to test image: /tmp/stest.831046569/test.sif
        --- PASS: TestE2E/CMDENVVARS (2.82s)
            --- PASS: TestE2E/CMDENVVARS/read-only_cache_directory (2.82s)
                cmdenvvars.go:30: Set up temporary cache-dir directory to /tmp/stest.831046569/cache-dir.507277044
                singularitycmd.go:637: Running command "/tmp/s/bin/singularity --debug pull /tmp/stest.831046569/image-cache.484577219/testImg.sif library://alpine:latest"
    PASS
    coverage: 79.1% of statements
    ok      github.com/sylabs/singularity/e2e       3.855s  coverage: 79.1% of statements

mem (approved):

dtrudg (approved):

PR #4588, opened 2019-10-08
mconfig: Allow build without cryptsetup

drhpc:
This simply deactivates the fatal error in mconfig when not finding cryptsetup.
Is this the right approach? Something I am overlooking?

The code using cryptsetup checks at runtime if the version is compatible, anyway. So, I gather we should have graceful failure. And cryptsetup is not really a prerequisite for building Singularity, or is it?

PS: And yes, I managed to create a pull request to my fork's master instead of the singularity repo before, waiting a few days and wondering why nothing happens.


jmstover (approved):
Thanks... I _thought_ that there was already a PR for this. I must have imagined it... :/

I'd like to get @mem to be the merge point in case there was a reason I don't know that it was actually erroring out at the `mconfig` stage.

mem:
> Thanks... I _thought_ that there was already a PR for this. I must have imagined it... :/
> 
> I'd like to get @mem to be the merge point in case there was a reason I don't know that it was actually erroring out at the `mconfig` stage.

cryptsetup is required for the tests to work. Some tests will check if cryptsetup is installed and skip the test if not. _That_ is the wrong way to go about it. Those tests should fail instead. This check is / was here so that instead of a failure 20 minutes down the process, the build fails quickly.

If this gets merged, we probably need to open an issue to check and fix any incorrect tests.

Also, it's probably a good idea to introduce a build tag so that if cryptsetup is not available at build time (and therefore cannot be properly tested), then encryption support is left out. I rather have a singularity binary without encryption support than a singularity binary with untested encryption support.

drhpc:
I don't feel very strongly about whether the crypto code should be available if it has not been tested. But turning this into an optional dependency is much appreciated.

So, set a flag that disables crypt stuff if cryptsetup is absent during mconfig (and also, if it's the wrong version)?

jmstover:
> Also, it's probably a good idea to introduce a build tag so that if cryptsetup is not available at build time (and therefore cannot be properly tested), then encryption support is left out. I rather have a singularity binary without encryption support than a singularity binary with untested encryption support.

This to me is bad. Testing doesn't happen at build time. We may _build_ on a system where `cryptsetup` doesn't exist, but it may exist where it will run.

PR #4589, opened 2019-10-08
Fix current working directory mount check

### LINKED ISSUE #4587: Current working directory can override --home flag
### Version of Singularity:

```
3.4.1+400-gda0e39beb
```

### Expected behavior

List an empty directory:

```
mkdir /tmp/home && cd $HOME && singularity exec -H /tmp/home:$HOME image ls -la
```

### Actual behavior

The above command actually bind the host home directory instead of the one provided by `-H` flag

cclerget:

Add a new RPC methods to Singularity RPC server to get file status for container mount so dev/ino are checked against the host one to determine if the current working is already mounted.
Also set `PWD` environment variable in `RunSingularity` to match the one set by `WithDir` operations.

 - Fixes #4587 

### DIFF
--- a/e2e/actions/regressions.go
+++ b/e2e/actions/regressions.go
@@ -9,9 +9,8 @@ import (
 	"path/filepath"
 	"testing"
 
-	"github.com/sylabs/singularity/internal/pkg/util/fs"
-
 	"github.com/sylabs/singularity/e2e/internal/e2e"
+	"github.com/sylabs/singularity/internal/pkg/util/fs"
 )
 
 // Check there is no file descriptor leaked in the container
--- a/e2e/actions/regressions.go
+++ b/e2e/actions/regressions.go
@@ -6,7 +6,10 @@
 package actions
 
 import (
+	"path/filepath"
 	"testing"
+
+	"github.com/sylabs/singularity/internal/pkg/util/fs"
 
 	"github.com/sylabs/singularity/e2e/internal/e2e"
 )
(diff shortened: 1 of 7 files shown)


e2e/actions/regressions.go
```diff
@@ -6,8 +6,11 @@
 package actions
 
 import (
+	"path/filepath"
 	"testing"
 
+	"github.com/sylabs/singularity/internal/pkg/util/fs"
+
```
WestleyK: Extra newline?

```suggestion
```

internal/pkg/runtime/engine/singularity/rpc/args.go
```diff
@@ -7,6 +7,7 @@ package rpc
 
 import (
 	"os"
+	"syscall"
```
WestleyK, line 10: Would it make more sense to use `golang.org/x/sys/unix` instead of `syscall`? Since `syscall` is deprecated.

_From the `syscall` godoc_
>Deprecated: this package is locked down. Callers should use the corresponding package in the golang.org/x/sys repository instead. That is also where updates required by new systems or versions should be applied. See https://golang.org/s/go1.4-syscall for more information.

WestleyK (commented):
Looks good to me, just a couple small suggestions...

internal/pkg/runtime/engine/singularity/rpc/args.go
cclerget: Yes but not in this PR, I will open a technical debt issue for that to remove `syscall` from Singularity codebase

WestleyK (approved):
Looks good to me! Thanks for also adding e2e tests for this issue!

Issue #4578, opened 2019-10-04
[Regression] Unable to pull SLC6 from DockerHub: packer failed to pack: while unpacking tmpfs: unpack layer: unpack entry: .: apply hdr metadata: restore xattr metadata: saw forbidden xattr "security.selinux":


olifre:
### Version of Singularity:

```
singularity version 3.4.1-1.1.el7
```

It worked fine with `3.2.1-1.1.el7`, so that's a regression. 

### Expected behavior

Using `SL6.def` containing the following:
```
BootStrap: docker
From: cern/slc6-base:latest
```
I'd expect this:
```
$ singularity build foo.sif SL6.def
```
to work. 

### Actual behavior

```
$ singularity build foo.sif SL6.def 
INFO:    Starting build...
Getting image source signatures
Skipping fetch of repeat blob sha256:5fb95acba89baff8fc06e9f83ac042b15437a69b1c8cb494846ced8f95e35080
Copying config sha256:b0486bbac8a1e7810ca22fd74038374a4320629b70cb7c431e1bcbf6eaebe94d
 375 B / 375 B [============================================================] 0s
Writing manifest to image destination
Storing signatures
2019/10/04 15:55:43  info unpack layer: sha256:5fb95acba89baff8fc06e9f83ac042b15437a69b1c8cb494846ced8f95e35080
FATAL:   While performing build: packer failed to pack: while unpacking tmpfs: unpack layer: unpack entry: .: apply hdr metadata: restore xattr metadata: saw forbidden xattr "security.selinux": .
```

### What OS/distro are you running

```
$ lsb_release -a
LSB Version:    :core-4.1-amd64:core-4.1-noarch
Distributor ID: CentOS
Description:    CentOS Linux release 7.7.1908 (Core) 
Release:        7.7.1908
Codename:       Core
```

### How did you install Singularity

EPEL

dtrudg:
I don't believe this will be fixed by what is already in `release-3.4` for the upcoming `3.4.2`. Need to investigate further.

dtrudg:
Confirmed this replicates on current `release-3.4` on my Debian host.

dtrudg:
So this is a nasty one to deal with. The `umoci` library we are using now to more faithfully extract docker layers will set permissions, xattrs more faithfully than the previous `image-tools` library. When umoci is used to create a layer tar it will ignore "security.selinux" xattrs, which are host specific:

[...]

PR #4592, opened 2019-10-08
Set apex log level to correspond with sylog log level

### LINKED ISSUE #4586: 3.4.1 warnings about xattr should be turned off
### Version of Singularity:

```
singularity version 3.4.1-1.2.el7
```
### Expected behavior

I expect singularity to be silent if I use the -s flag and to print a minor number of warning without. In 3.2.1 with -s it executes and returns
```
-bash-4.2$ /cvmfs/atlas.cern.ch/repo/containers/sw/singularity/testing/x86_64-el7/3.2.1/bin/singularity -s build --sandbox /home/aforti/docker_atlas_analysisbase/image docker://atlas/analysisbase
-bash-4.2$ 
```

### Actual behavior

when building centos images singularity prints out an enormous level of warnings about ignoring xattr on files. It is only 3 on the basic OS image (which is already annoying) but with more complex  images I have >12k warnings polluting the stdout. This should really be hidden behind debugging and even there I'm afraid it is polluting, maybe not for a dev, but certainly for an admin. 

```
export SINGULARITY_CACHEDIR=docker_atlas_analysisbase/cache
export SINGULARITY_TMPDIR=docker_atlas_analysisbase
export SINGULARITY_DISABLE_CACHE=true

singularity -s build --sandbox /home/aforti/docker_atlas_analysisbase/image docker://atlas/analysisbase > 3.4.1-test.sing.silent.txt
```
I've attached the output of this command
[3.4.1-test.sing.silent.txt](https://github.com/sylabs/singularity/files/3701702/3.4.1-test.sing.silent.txt)

```
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
ANSI_COLOR="0;31"

```

rpm from epel-testing

DrDaveD:

Make the apex log level, used by umoci, track sylog log levels.  The apex warning level is too verbose, so this only enables that when the sylog level is verbose and by default sets the level to errors only.

This PR is submitted against release-3.4 because sandbox builds on the master branch are broken (at least for docker://centos:7) at the moment, but the PR applies equally to master.

 - Fixes #4586

### DIFF
--- a/internal/pkg/sylog/sylog.go
+++ b/internal/pkg/sylog/sylog.go
@@ -9,6 +9,7 @@ package sylog
 
 import (
 	"fmt"
+	apexlog "github.com/apex/log"
 	"io"
 	"io/ioutil"
 	"os"
@@ -66,17 +67,15 @@ var colorReset = "\x1b[0m"
 var loggerLevel messageLevel
 
 func init() {
-	_level, ok := os.LookupEnv("SINGULARITY_MESSAGELEVEL")
-	if !ok {
-		loggerLevel = info
-	} else {
-		_levelint, err := strconv.Atoi(_level)
-		if err != nil {
-			loggerLevel = info
-		} else {
-			loggerLevel = messageLevel(_levelint)
+	_levelint := int(messageLevel(info))
+	_levelstr, ok := os.LookupEnv("SINGULARITY_MESSAGELEVEL")
+	if ok {
+		_leveli, err := strconv.Atoi(_levelstr)
+		if err == nil {
+			_levelint = _leveli
 		}
 	}
+	SetLevel(_levelint)
 }
 
 func prefix(level messageLevel) string {
@@ -158,6 +157,14 @@ func Debugf(format string, a ...interface{}) {
 // SetLevel explicitly sets the loggerLevel
 func SetLevel(l int) {
 	loggerLevel = messageLevel(l)
+	// the apex log (for umoci) is too noisy at warn level, avoid that
+	if loggerLevel < verbose {
+		apexlog.SetLevel(apexlog.ErrorLevel)
+	} else if loggerLevel < debug {
+		apexlog.SetLevel(apexlog.InfoLevel)
+	} else {
+		apexlog.SetLevel(apexlog.DebugLevel)
+	}
 }
 
 // DisableColor for the logger
--- a/internal/pkg/sylog/sylog.go
+++ b/internal/pkg/sylog/sylog.go
@@ -9,13 +9,14 @@ package sylog
 
 import (
 	"fmt"
-	apexlog "github.com/apex/log"
 	"io"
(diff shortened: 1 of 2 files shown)


dtrudg:
Thanks @DrDaveD - I'm afraid at present without further investigation I can't endorse hiding all apex warnings when we are expecting other warnings to show.

The `selinux` xattr warnings may be just noise to most, but there are other parts of the umoci code where Warn may display something important.

I agree the logelevel should be set to errors only for `--silent` though.

internal/pkg/sylog/sylog.go
```diff
@@ -158,6 +157,14 @@ func Debugf(format string, a ...interface{}) {
 // SetLevel explicitly sets the loggerLevel
 func SetLevel(l int) {
 	loggerLevel = messageLevel(l)
+	// the apex log (for umoci) is too noisy at warn level, avoid that
+	if loggerLevel < verbose {
```
dtrudg: I'd be happy to take this as a PR into 3.4.2 if the apex loglevel is matched to the sylog level - i.e. hide the apex warnings on `-s` but leave them otherwise. I think we'd need a lot more confidence that umoci was only ever using 'warn' for unimportant things to do what's currently in place in this PR.

dtrudg (changes requested):

internal/pkg/sylog/sylog.go
DrDaveD: Alright, I'll make that change.  It will still address the request in #4578.

DrDaveD:
The sandbox build failure I referred to in the submission is detailed in issue #4593

internal/pkg/sylog/sylog.go
DrDaveD: What about the quiet level?  There's no corresponding level in apex.  Can I have that silence apex warnings?

internal/pkg/sylog/sylog.go
```diff
@@ -158,6 +158,17 @@ func Debugf(format string, a ...interface{}) {
 // SetLevel explicitly sets the loggerLevel
 func SetLevel(l int) {
 	loggerLevel = messageLevel(l)
+	// set the apex log level, for umoci
+	if loggerLevel <= log {
+		// quiet or silent options
```
dtrudg: I don't think we want to do this for quiet - a warning isn't normal output.

`-q, --quiet     suppress normal output`

dtrudg (changes requested):

internal/pkg/sylog/sylog.go
dtrudg: I don't think so - sorry for the delay - see below

dtrudg (approved):

dtrudg:
LGTM - thanks @DrDaveD !

PR #4511, opened 2019-09-23
Merge release 3.4 into master

mem:
This is necessary to update master's version by bringing in the release-3.4.1 tag.

PR #4595, opened 2019-10-09
Cherry-pick #4592 to master

dtrudg:
This cherry-picks #4592 into master.

That PR was submitted against release-3.4 as an urgent fix to get into 3.4.2, but needs to be in master.

Opening this here to offer a choice of either:

1. Merge this cherry-pick to bring the change in straight away
2. Wait until we do the release-3.4 -> master merge after final (non RC) 3.4.2 release is made #4511


DrDaveD:
Note that the go.mod is for golang 1.12, so you probably want to put this into master before the switch to 1.13.

ikaneshiro (approved):
I think it's best to take the approach of merging this now rather than waiting

PR #4601, opened 2019-10-10
Error if '--sif-id' is 0

WestleyK:

If the user passed `--sif-id 0`, sinse that is the default
`verify`/`sign` behavior, it will seem like that flag has no effect
(witch it does not), it should fail as an invalid value.

There are probably better ways to solve this, but this is what I came up
with. Open to suggestions.

 - Fixes #


cmd/internal/cli/sign.go
```diff
@@ -104,17 +103,29 @@ var SignCmd = &cobra.Command{
...
 func doSignCmd(cpath, url string) error {
-	if sifGroupID != 0 && sifDescID != 0 {
-		return fmt.Errorf("only one of -i or -g may be set")
+	if sifGroupID != -1 && sifDescID != -1 {
+		sylog.Fatalf("only one of -i or -g may be set")
+	}
+
+	// If the user used '0' as descriptor id, then fail as
+	// 0 is invalid.
+	if sifGroupID == -1 && sifDescID == 0 {
```
cclerget: Looks like both `sifGroupID` and `sifDescID` can't be 0

cmd/internal/cli/sign.go
```diff
@@ -87,7 +87,6 @@ var SignCmd = &cobra.Command{
 	Args:                  cobra.ExactArgs(1),
 
 	Run: func(cmd *cobra.Command, args []string) {
-
 		// args[0] contains image path
 		fmt.Printf("Signing image: %s\n", args[0])
 		if err := doSignCmd(args[0], keyServerURI); err != nil {
```
cclerget: while we are, can we remove `keyServerURI` which is global and maybe replace `sylog.Errorf/os.Exit` by `sylog.Fatalf()` unless there is need to return a specific exit code ?

cmd/internal/cli/sign.go
```diff
@@ -104,17 +103,29 @@ var SignCmd = &cobra.Command{
 }
 
 func doSignCmd(cpath, url string) error {
-	if sifGroupID != 0 && sifDescID != 0 {
-		return fmt.Errorf("only one of -i or -g may be set")
+	if sifGroupID != -1 && sifDescID != -1 {
+		sylog.Fatalf("only one of -i or -g may be set")
+	}
+
```
cclerget, line 105: you could use cobra by passing `cmd *cobra.Command` to `doSignCmd` and have something like

```
if cmd.Flag(signSifGroupIDFlag.Name).Changed() && sifGroupID == 0 {
    sylog.Fatalf("invalid group id")
}
if cmd.Flag(signSifDescSifIDFlag.Name).Changed() && sifDescID == 0 {
    sylog.Fatalf("invalid descriptor id")
}
if cmd.Flag(signSifDescIDFlag.Name).Changed() && sifDescID == 0 {
    sylog.Fatalf("invalid descriptor id")
}
```

cclerget (changes requested):
Left suggestions about using cobra instead, request changes mostly for groupid must be > 0

cmd/internal/cli/sign.go
WestleyK: I just removed `keyServerURI` altogether, does not seem like it was used anywhere here.

WestleyK: Also removed the error return, to keep it consistent with `verify.go`.

cclerget (approved):
LGTM. Thanks !

PR #4604, opened 2019-10-10
Check go version at buildtime

mem:
This PR contains three commits:

* Update goversion to require Go 1.13

Checks for the correct go version at build time and errors out if Go 1.13 or later is not in use.

* Improve messaging around Go version check

Change the format of the messaging around the version check to make it more clear.

* Use correct Go path is the user provided one

Fix typo (?) in expansion of go's path.


sashayakovtseva (approved):
Thanks!

PR #4584, opened 2019-10-07
e2e binds tests and nested bind fix

cclerget:
## Description of the Pull Request (PR):

Add e2e bind tests and fix nested binds 

### This fixes or addresses the following GitHub issues:

 - Fixes #4435 

#### Before submitting a PR, make sure you have done the following:

- Read the [Guidelines for Contributing](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/sylabs/singularity/blob/master/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- Added tests to validate this PR and tested this PR locally with a `make testall`
- Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](https://github.com/sylabs/singularity/blob/master/CONTRIBUTORS.md)

Attn: @singularity-maintainers

PR #4609, opened 2019-10-11
Move e2e test image from centos to alpine

cclerget:

Move e2e test image from centos to alpine to speed up test bootstrap

### DIFF
--- a/e2e/actions/actions.go
+++ b/e2e/actions/actions.go
@@ -244,7 +244,7 @@ func (c actionTests) actionExec(t *testing.T) {
 		{
 			name: "NoHome",
 			argv: []string{"--no-home", c.env.ImagePath, "ls", "-ld", user.Dir},
-			exit: 2,
+			exit: 1,
 		},
 	}
 
--- a/e2e/testdata/Singularity
+++ b/e2e/testdata/Singularity
@@ -1,8 +1,8 @@
 bootstrap: library
-from: centos:7
+from: alpine:3.7
 
 %labels
-    maintainer "Eduardo Arango<eduardo@sylabs.io>"
+    maintainer "Sylabs Team <support@sylabs.io>"
 
 %help
     export BAD_GUY=Thanos
@@ -65,7 +65,8 @@ export HELLOTHISIS
 %post
     echo "POST"
     # Needed for startscript
-    yum install -y nmap-ncat hostname
+    apk update
+    apk add nmap-ncat
 
 %test
     exec "$@"
--- a/e2e/verify/verify.go
+++ b/e2e/verify/verify.go
@@ -72,7 +72,7 @@ func (c ctx) singularityVerifyKeyNum(t *testing.T) {
 		},
 		{
 			name:         "verify number signers success",
-			expectNumOut: 1,
+			expectNumOut: 2,
 			imageURL:     successURL,
 			imagePath:    c.successImage,
 			expectExit:   0,


mem (changes requested):
I don't get it. Rebuild time for e2e is minimal.

What problem are you actually solving by doing this?

cclerget:
@mem Well when I add tests like in #4584 and have to wait 1 minute each time I run a test alone or the array of tests by waiting than `e2e.EnsureImage` finished to build the test image ... it's annoying to wait 1 minute for a test execution of 1 second and see the test failed and so on ... So basically this change would allow when adding tests to do :

```
singularity build -f /tmp/test_image.sif e2e/testdata/Singularity
SINGULARITY_E2E_IMAGE=/tmp/test_image.sif ./scripts/e2e-test -v -run TestE2E/ACTIONS/binds ./e2e
```

The test failed ... fix it and re-run without losing 1 minute again :

mem (changes requested):
Given your previous comment, I understand why you want this sort of optimization, but you are introducing unnecessary variability into the tests. This sort of change does not generalize well: now there would be a branch where there should be none; it's changing the behavior of the tests based on a value taken from the environment; the image is not controlled by the tests (its contents is unknown).

This is the kind of thing that you keep in a local patch, but not one to be pushed to the repository. I'm sorry I cannot provide a better answer, but the nature of the problem you want to solve does not have a better answer, as you have to pick one: either you make the tests deterministic, or you make them flexible.

Taking a step back, the only reason you mentioned previously for this change is the time it takes to run a test. If the problem is that the tests are slow because they need to build this image, maybe the problem that needs fixing is that: identify a _very small_ image that can be used for this purpose and commit _that_ (along with its corresponding definition file) to the repository. We already have a couple of those in the repo, and that's a problem in itself (there's more than one). We are also downloading images from the network for cases where we are _not_ testing the download path but something else. If we can create a single _very small_ image that can be shared by all these tests, we would be fixing an actual problem that's impacting a lot of people.

cclerget:
@mem Changed for alpine resulting in a 10 minutes gain :+1: . I checked tests wasn't skipped and it looks good so far, just I noticed that encryption tests are actually skipped

scripts/e2e-test
```diff
@@ -16,5 +16,6 @@ sudo \
 	env -i \
 		PATH=$PATH \
 		HOME=$HOME \
+		SINGULARITY_E2E_TESTIMAGE=$SINGULARITY_E2E_TESTIMAGE \
```
dtrudg: @cclerget - this line is now redundant I think?

e2e/suite.go
```diff
@@ -116,7 +116,7 @@ func Run(t *testing.T) {
 	}
 	testenv.TestDir = name
 
-	// Build a base image for tests
+	// Build a base image for tests or use the provided one
```
mem:
```suggestion
	// Build a base image for tests
```

scripts/e2e-test
mem:
```suggestion
```

mem (approved):
thanks! that's good news :-)

mem (changes requested):
Just commit those changes

jscook2345:
Alpine is kinda of a specialist container OS, without glibc yes? Are there other options for a small image but a more standard OS? I feel like our users do not use alpine often.

update: This is more of a general question. For this issue I think its fine to use alpine.

mem:
> Alpine is kinda of a specialist container OS, without glibc yes? Are there other options for a small image but a more standard OS? I feel like our users do not use alpine often.

Yeah, but the point is that we need an image to test certain singularity commands. What we are testing here is that singularity does what we expect it to do for a given command (IIRC there's a test that does something like `singularity exec test.sif true`), not that CentOS provides enough jun^Wfeatures by default to make a test pass ;-)

mem (approved):
Thanks!

mem:
> > Alpine is kinda of a specialist container OS, without glibc yes? Are there other options for a small image but a more standard OS? I feel like our users do not use alpine often.
> 
> Yeah, but the point is that we need an image to test certain singularity commands. What we are testing here is that singularity does what we expect it to do for a given command (IIRC there's a test that does something like `singularity exec test.sif true`), not that CentOS provides enough jun^Wfeatures by default to make a test pass ;-)

And just to expand on that thought:

We need a series of tests that test basic singularity functionality (like the above) and then we need a series of tests that check that singularity works on a given distribution (Ubuntu, CentOS, etc) and then we need some tests that test that specific containers (or containers based on specific distros) work. While it might look like we are doing the later with the image as it is before this change, I really don't think that's the case.

Does that make sense?

PR #4612, opened 2019-10-11
Junit publication fixes

mem:
This PR contains two commits:

* Make sure that the JUnit report is published even if the test fails

When tests fail, the JUnit report is not published because the script returns too early.

* Publish a JUnit report (on demand) for E2E, too

When the other changes were merged, the unit-test publication went in, but not the e2e publication.


dtrudg (approved):
LGTM - sorry I didn't think about e2e / multiple junit output when I reviewed the last.

Issue #3944, opened 2019-07-17
RFE: singularity config command


cclerget:
Add a `singularity config` command for admin tasks in order to manage the various Singularity configuration files.

Proposal for:
- `singularity.conf`:
  ```
  singularity config global --set "directive_name" directive_value
  singularity config global --get "directive_name"
  ```

  For multi-values directives in `singularity.conf` like `bind path`:
  ```
  singularity config global --set "bind path" /etc/hosts
  singularity config global --set --remove "bind path" /etc/hosts
  ```

- fakeroot configuration:
  ```
  singularity config fakeroot --add user
  singularity config fakeroot --remove user
  singularity config fakeroot --disable user
  singularity config fakeroot --enable user
  ```

- basic network management:
  ```
  singularity config network --enable network_name
  singularity config network --disable network_name
  singularity config network --list
  singularity config network --list --disabled
  ```

- capability management:
  ```
  singularity capability -> singularity config capability
  ```

DrDaveD:
This did not make it into the 3.4 release so please change the label.

mem:
> This did not make it into the 3.4 release so please change the label.

Lacking a better place to talk about this, this is as good a place as any.

_My_ own understanding of this is that _labels_ identify the release to which this issue applies (other bug trackers will call this "found in version", "applicable version", etc). The _Milestone_ field is where you document the target milestone for resolution. The _Release_ field is where you document the target release for resolution or the release that should include this fix (in case something needs to be backported, for example).

If you think about it, it does not make sense to have the same information duplicated across the milestone field and the labels (likewise for release).

Also note that _milestone_ and _release_ are overlapping concepts but not the same thing.

Now... where's a good place to document this that doesn't sit too far away from the place where it's used? I mean, this could be added to the website somewhere, and nobody would find the information.

jeffcbecker:
Will this be in the next release(3.4.1 perhaps)? When is that?

DrDaveD:
It has not been decided yet but is being discussed in a thread on the new [singularity-devel@lbl.gov](https://groups.google.com/a/lbl.gov/forum/#!forum/singularity-devel) mailing list.

jeffcbecker:
Did this make it into 3.4.1?

dtrudg:
@jeffcbecker - no, this will be targeted for 3.5 - there was a decision not to add a new feature into a patch release.

dtrudg:
`config global` is in from #4623

PR #4614, opened 2019-10-14
Rework singularity configuration parser

cclerget:

To ease development of `singularity config global` see (#3944) the parser code was reworked and cleanup a bit :

- Splitted configuration parser into two functions: `ParseData` and `SetConfig`
- Code cleanup.
- Added unit tests for code coverage.
- Fixed bad template formatting loop for `limit container xxx` directives

 - Related to #3944 

### DIFF
--- a/internal/pkg/util/fs/squashfs/squashfs.go
+++ b/internal/pkg/util/fs/squashfs/squashfs.go
@@ -13,16 +13,15 @@ import (
 
 	"github.com/sylabs/singularity/internal/pkg/buildcfg"
 	"github.com/sylabs/singularity/pkg/runtime/engine/config"
-	singularityconfig "github.com/sylabs/singularity/pkg/runtime/engine/singularity/config"
 )
 
 // GetPath figures out where the mksquashfs binary is
 // and return an error is not available or not usable.
 func GetPath() (string, error) {
 	// Parse singularity configuration file
-	c := &singularityconfig.FileConfig{}
 	configFile := buildcfg.SINGULARITY_CONF_FILE
-	if err := config.Parser(configFile, c); err != nil {
+	c, err := config.ParseFile(configFile)
+	if err != nil {
 		return "", fmt.Errorf("unable to parse singularity.conf file: %s", err)
 	}
 
--- a/pkg/runtime/engine/config/parser.go
+++ b/pkg/runtime/engine/config/parser.go
@@ -17,159 +17,166 @@ import (
 	"text/template"
 )
 
-// Parser parses configuration found in the file with the specified path.
-func Parser(filepath string, f interface{}) error {
-	var err error
-	var c *os.File
-	var b []byte
-	directives := make(map[string][]string)
+// Directives represents the configuration directives type
+// holding directives mapped to their respective values.
+type Directives map[string][]string
 
-	if filepath != "" {
-		c, err = os.Open(filepath)
-		if err != nil {
-			return err
-		}
-		b, err = ioutil.ReadAll(c)
-		if err != nil {
-			return err
-		}
+var parserReg = regexp.MustCompile(`(?m)^\s*([a-zA-Z _]+)\s*=\s*(.*)$`)
 
-		c.Close()
+// GetDirectives parses configuration directives from reader
+// and returns a directive map with associated values.
+func GetDirectives(reader io.Reader) (Directives, error) {
+	if reader == nil {
+		return make(Directives), nil
 	}
 
-	r, err := regexp.Compile(`(?m)^\s*([a-zA-Z _]+)\s*=\s*(.*)$`)
+	data, err := ioutil.ReadAll(reader)
 	if err != nil {
-		return fmt.Errorf("regex compilation failed")
+		return nil, fmt.Errorf("while reading data: %s", err)
 	}
 
-	for _, match := range r.FindAllSubmatch(b, -1) {
+	directives := make(Directives)
+
+	for _, match := range parserReg.FindAllSubmatch(data, -1) {
 		if match != nil {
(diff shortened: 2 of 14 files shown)


pkg/runtime/engine/config/parser.go
```diff
@@ -17,147 +17,156 @@ import (
...
+	var buf bytes.Buffer
 
-		c.Close()
+	if reader != nil {
+		if _, err := io.Copy(&buf, reader); err != nil {
+			return nil, err
+		}
 	}
 
 	r, err := regexp.Compile(`(?m)^\s*([a-zA-Z _]+)\s*=\s*(.*)$`)
```
sashayakovtseva: Can be defined as a package var with `regexp.MustCompile`. Just a suggestion, this way is also fine taking into account when `Parser` is called.

pkg/runtime/engine/config/parser.go
```diff
@@ -17,147 +17,156 @@ import (
...
-			return err
-		}
+// ParseData parses data and returns a directive map with
+// associated value.
+func ParseData(reader io.Reader) (Directives, error) {
+	var buf bytes.Buffer
 
-		c.Close()
+	if reader != nil {
+		if _, err := io.Copy(&buf, reader); err != nil {
```
sashayakovtseva: You can use `ioutil.ReadAll` to get bytes.

pkg/runtime/engine/config/parser.go
```diff
@@ -17,147 +17,156 @@ import (
...
+			}
 		}
 	}
 
-	val := reflect.ValueOf(f).Elem()
+	return directives, nil
+}
+
+// SetConfig sets the corresponding interface fields associated
+// with directives
```
sashayakovtseva:
```suggestion
// with directives.
```

pkg/runtime/engine/config/parser.go
```diff
@@ -17,147 +17,156 @@ import (
...
 				}
 			}
 		}
 	}
+
+	return nil
+}
+
+// Parser parses configuration found in the file with the specified path.
+func Parser(filepath string, f interface{}) error {
```
sashayakovtseva: I went through all usages, and `*singularityconfig.FileConfig` is always passed as a second argument. 
Maybe worth accepting a concrete type then?

pkg/runtime/engine/config/parser.go
```diff
@@ -17,147 +17,156 @@ import (
...
+	var reader io.Reader
+
+	if filepath != "" {
```
sashayakovtseva: Also after searching for usages, this is usually `buildcfg.SINGULARITY_CONF_FILE`. Can it be empty?

pkg/runtime/engine/config/parser.go
sashayakovtseva: I understand, that we don't want to break package API here.
But generally changing a passed argument is not an explicit API. I'd prefer to return `*FileConfig` here (if my comment about accepting `*FileConfig` is addressed).  Otherwise that is fine since we don't know concrete type.

sashayakovtseva: Also  name `Parser` implies we return some parser, but here we parse and set values into `f`. 
Maybe rename to smth like `config.ParseFile`?

pkg/runtime/engine/config/parser.go
sashayakovtseva: For explicitness we can do smth like
```
if filepath == "" {
   makeDefault(f)
   return nil 
}
```

so that we don't parse regex and do all other checks.

sashayakovtseva (changes requested):

pkg/runtime/engine/config/parser.go
cclerget: Originally it was done like that with the eventuality to have other configuration file using the same format. I don't think it will be the case, so yes we could return `*FileConfig` instead

pkg/runtime/engine/config/parser.go
cclerget: Will look

internal/pkg/util/fs/squashfs/squashfs.go
```diff
@@ -13,16 +13,15 @@ import (
...
 )
 
 // GetPath figures out where the mksquashfs binary is
 // and return an error is not available or not usable.
 func GetPath() (string, error) {
 	// Parse singularity configuration file
-	c := &singularityconfig.FileConfig{}
 	configFile := buildcfg.SINGULARITY_CONF_FILE
-	if err := config.Parser(configFile, c); err != nil {
+	c, err := config.ParseFile(configFile)
```
sashayakovtseva, line 23: `configFile` can be omitted

pkg/runtime/engine/config/parser.go
```diff
@@ -17,159 +17,169 @@ import (
...
+}
+
+// ParseFile parses configuration file with the specified path.
+func ParseFile(filepath string) (*FileConfig, error) {
+	if filepath == "" {
+		// grab the default configuration
+		return GetConfig(nil)
+	}
+
+	var reader io.Reader
```
sashayakovtseva: Is there a reason for this reader? you can use opened file b/c it implements Reader as well

sashayakovtseva (approved):
LGTM, minor suggestions that can be optionally addressed.

internal/pkg/util/fs/squashfs/squashfs.go
cclerget: no we need to parse `singularity.conf` to get path of `mksquashfs` if defined by admin otherwise we fallback and look in default paths

Issue #4285, opened 2019-08-22
[E2E]: create tests to ensure umask is honored


WestleyK:
## Version of Singularity:

```
3.3
```

### Expected behavior

There should be some tests to ensure umask is honored.

### Actual behavior

Does not exist yet.

### Steps to reproduce behavior

Write here.

dtrudg:
Closed via #4626

Issue #2381, opened 2018-11-20
Some umask value generate bad image


cclerget:
## Version of Singularity:

3.0.1

### Expected behavior

```
singularity shell docker://busybox
Singularity>
```

### Actual behavior

```
singularity shell docker://alpine
WARNING: passwd file doesn't exist in container, not updating
WARNING: group file doesn't exist in container, not updating
FATAL:   Unable to parse singularity.conf file: open /usr/local/etc/singularity/singularity.conf: permission denied
```

### Steps to reproduce behavior

```
rm -rf $HOME/.singularity/cache/oci-tmp/*
umask 0027
singularity shell docker://alpine
```

cclerget:
Fixed by #2382

PR #4626, opened 2019-10-16
[E2E] Add some umask tests for 'pull'

WestleyK:

This PR adds some tests to ensure the umask is honored for
pulling and overriding a image.

 - Addresses #4285


e2e/pull/pull.go
```diff
@@ -464,13 +465,107 @@ func (c ctx) testPullDisableCacheCmd(t *testing.T) {
...
+			umask:      0027,
+			expectPerm: 0750,
+		},
+
+		// With the force flag.
+		{
+			name:       "0022 umask pull",
+			imagePath:  filepath.Join(c.env.TestDir, "0022-umask-pull"),
+			umask:      0022,
+			expectPerm: 0755,
```
dtrudg, line 507: These tests with force (overwriting) aren't really adding anything I don't think, since they are overwriting files in a way that the exact same permissions are expected / pretty much guaranteed.

They *would* be testing something valuable if the file they were overwriting *didn't already* have the expected permissions.

e2e/pull/pull.go
```diff
@@ -464,13 +465,107 @@ func (c ctx) testPullDisableCacheCmd(t *testing.T) {
...
+		},
+		{
+			name:       "0077 umask pull",
+			imagePath:  filepath.Join(c.env.TestDir, "0077-umask-pull"),
+			umask:      0077,
+			expectPerm: 0700,
+		},
+		{
+			name:       "0027 umask pull",
```
dtrudg, line 495: Is a 3rd test telling us anything here? If we get perms right from 2 different umasks then our code is correct and we're really just testing upstream go functions I think?

e2e/pull/pull.go
dtrudg: That is... assuming we *want* the file permissions to be changed on overwrite.

dtrudg (changes requested):

e2e/pull/pull.go
WestleyK: Good point, I'll come up with a fix...

e2e/pull/pull.go
WestleyK: I dont think so, `0027` seems to be the magic umask where things start to break/not function. Example: https://github.com/sylabs/singularity/issues/2381

e2e/pull/pull.go
```diff
@@ -464,13 +465,120 @@ func (c ctx) testPullDisableCacheCmd(t *testing.T) {
...
+}
+
 // E2ETests is the main func to trigger the test suite
 func E2ETests(env e2e.TestEnv) func(*testing.T) {
 	c := ctx{
 		env: env,
 	}
 
 	return func(t *testing.T) {
+		t.Run("pullUmaskCheck", c.testPullUmask)
```
dtrudg, line 582: Is there a reason why this is above `c.setup(t)`? If yes - please comment why here. If not, please move it to the bottom of the test runs.

WestleyK: This can run before the setup. The `setup` only seems to be for the oras test registry.

dtrudg (approved):

PR #4634, opened 2019-10-18
Fix issue with /proc hidepid option

### LINKED ISSUE #4633: fakeroot option does not work with hidepid procfs
### Version of Singularity:

```
$ singularity version
3.4.1
```

### Expected behavior

```
$ singularity shell -f -w ubuntu/
Singularity ubuntu:~>
```

### Actual behavior

```
singularity shell -f -w ubuntu/
ERROR  : Failed to change directory to /proc/21020: No such file or directory
```

This error occurs when hidepid=1 or 2 option is used in the procfs with fakeroot option.
Is it possible to use fakeroot with hidpid?

Mount procfs with hidepid=1 or 2 and execute `singularity -f`

```
$ cat /etc/os-release
NAME="SLES"
VERSION="12-SP2"
VERSION_ID="12.2"
PRETTY_NAME="SUSE Linux Enterprise Server 12 SP2"
ID="sles"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:12:sp2"

```

From the source

cclerget:

Actually `hidepid=1` or `hidepid=2` prevent Singularity from working correctly. This PR make it works correctly while `hidepid=` is set, unfortunately there is one case where `hidepid` require to be disabled, it's possible to start an instance with setuid workflow but it will join during instance join and won't stop all processes correctly during instance stop, this PR handle this case and show an error explaining to disable `hidepid`.

 - Fixes #4633


cmd/internal/cli/actions_linux.go
```diff
@@ -94,6 +95,27 @@ func convertImage(filename string, unsquashfsPath string) (string, error) {
...
+func hidepidProc() bool {
+	entries, err := proc.GetMountInfoEntry("/proc/self/mountinfo")
+	if err != nil {
+		sylog.Warningf("while reading /proc/self/mountinfo: %s", err)
+		return false
+	}
+	for _, e := range entries {
+		if e.Point == "/proc" {
+			for _, o := range e.SuperOptions {
+				if strings.HasPrefix(o, "hidepid=") {
```
sashayakovtseva, line 111: What if hidepid=0?

cclerget: `hidepid` appears only if > 0

sashayakovtseva (approved):

dtrudg:
I came across this today, in the Debian buster upgrade notes

https://www.debian.org/releases/stable/amd64/release-notes/ch-information.html

> 5.1.1. Hidepid mount option for procfs unsupported

> Using the hidepid mount option for /proc is known to cause problems with current versions of systemd, and is considered by systemd upstream to be an unsupported configuration. Users who have modified /etc/fstab to enable this option are advised to disable it before the upgrade, to ensure login sessions work on buster. (A possible route to re-enabling it is outlined on the wiki's Hardening page.) 

Given this is an unsupported systemd configuration, is it something we can anticipate seeing often?

cclerget:
@dctrud I don't think that's something we will see often for now as it can break user space applications but we may need to run in hardened environments, additionally is systemd will always consider it as an unsupported configuration ? Actually the fix is not complete as instances started with setuid workflow can't be joined.

dtrudg:
> additionally is systemd will always consider it as an unsupported configuration ?

It looks like it according to a comment from Lennart Poettering - but then lower down in that thread it looks like people may be working on things to get around this, which require kernel changes.

https://github.com/systemd/systemd/issues/12955#issuecomment-508490893
> But anyway, sorry. hidepid= the way it is is not compatible with systemd, and that's not going to change.

dtrudg (approved):

PR #4635, opened 2019-10-18
Add automatic detection of architecture during SIF assembler

### LINKED ISSUE #4627: Cross build image architecture
### Version of Singularity:

```
3.4.2+455-gfac1ca469
```

### Expected behavior

Show `s390x` and identify architecture correctly not showing `amd64`

```
singularity sif list /tmp/ubuntu-s390x.sif
Container id: b7251108-9d5c-4acf-b645-adfc5c792247
Created on:   2019-10-17 10:14:14 +0200 CEST
Modified on:  2019-10-17 10:14:14 +0200 CEST
Descriptor list:
1    |1       |NONE    |32768-32825               |Def.FILE
2    |1       |NONE    |36864-36961               |JSON.Generic
3    |1       |NONE    |40960-24965120            |FS (Squashfs/*System/s390x)
```

### Actual behavior

See steps to reproduce and run after :

```
singularity sif list /tmp/ubuntu-s390x.sif
Container id: b7251108-9d5c-4acf-b645-adfc5c792247
Created on:   2019-10-17 10:14:14 +0200 CEST
Modified on:  2019-10-17 10:14:14 +0200 CEST
Descriptor list:
1    |1       |NONE    |32768-32825               |Def.FILE
2    |1       |NONE    |36864-36961               |JSON.Generic
3    |1       |NONE    |40960-24965120            |FS (Squashfs/*System/amd64)
```

```
cat > /tmp/ubuntu-s390.def <<EOF
bootstrap: docker
from: s390x/ubuntu

%post
    uname -m
EOF

sudo singularity run docker://multiarch/qemu-user-static --reset -p yes
singularity build --fakeroot /tmp/ubuntu-s390x.sif /tmp/ubuntu-s390.def
```

```
NAME="Ubuntu"
VERSION="18.04.3 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.3 LTS"
VERSION_ID="18.04"
VERSION_CODENAME=bionic
UBUNTU_CODENAME=bionic
```

Git

cclerget:

Add automatic detection of architecture during SIF assembler process by scanning container image tree to find an ELF binary file for inspection.

FIx SIF image check to match compatible architecture and also check if there is persistent emulation installed in binfmt_misc.

 - Fixes #4627

### DIFF
--- a/internal/pkg/util/machine/machine.go
+++ b/internal/pkg/util/machine/machine.go
@@ -1,0 +1,316 @@
+// Copyright (c) 2019, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package machine
+
+import (
+	"bufio"
+	"bytes"
+	"debug/elf"
+	"encoding/binary"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"github.com/sylabs/sif/pkg/sif"
+	"github.com/sylabs/singularity/internal/pkg/sylog"
+	"github.com/sylabs/singularity/internal/pkg/util/fs"
+)
+
+// ErrUnknownArch is the error returned for unknown architecture.
+var ErrUnknownArch = errors.New("architecture not recognized")
+
+type format struct {
+	Arch       string
+	Sif        string
+	Compatible string
+	Machine    elf.Machine
+	Class      elf.Class
+	Endianness binary.ByteOrder
+	ElfMagic   []byte
+}
+
+var formats = []format{
+	{
+		Arch:       "386",
+		Sif:        sif.HdrArch386,
+		Compatible: "amd64",
+		Machine:    elf.EM_386,
+		Class:      elf.ELFCLASS32,
+		Endianness: binary.LittleEndian,
+		ElfMagic:   []byte{0x7F, 0x45, 0x4C, 0x46, 0x01, 0x01, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x02, 0x00, 0x03, 0x00},
+	},
+	{
+		Arch:       "386",
+		Sif:        sif.HdrArch386,
+		Compatible: "amd64",
+		Machine:    elf.EM_486,
+		Class:      elf.ELFCLASS32,
+		Endianness: binary.LittleEndian,
+		ElfMagic:   []byte{0x7F, 0x45, 0x4C, 0x46, 0x01, 0x01, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x02, 0x00, 0x06, 0x00},
+	},
+	{
+		Arch:       "amd64",
+		Sif:        sif.HdrArchAMD64,
+		Machine:    elf.EM_X86_64,
+		Class:      elf.ELFCLASS64,
+		Endianness: binary.LittleEndian,
+		ElfMagic:   []byte{0x7F, 0x45, 0x4C, 0x46, 0x02, 0x01, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x02, 0x00, 0x3e, 0x00},
+	},
+	{
+		Arch:       "arm",
+		Sif:        sif.HdrArchARM,
+		Compatible: "arm64",
+		Machine:    elf.EM_ARM,
+		Class:      elf.ELFCLASS32,
+		Endianness: binary.LittleEndian,
+		ElfMagic:   []byte{0x7F, 0x45, 0x4C, 0x46, 0x01, 0x01, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x02, 0x00, 0x28, 0x00},
+	},
+	{
+		Arch:       "armbe",
+		Sif:        sif.HdrArchARM, // FIXME: add HdrArchARMbe to sif package
+		Compatible: "arm64be",
+		Machine:    elf.EM_ARM,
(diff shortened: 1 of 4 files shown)


dtrudg:
I don't really understand why this complex inspection is necessary?

In the linked issue #4627 you are building from a docker image. Shouldn't it be sufficient to just ensure we carry forward the architecture from the OCI image config? `architecture` is a required field there.

If the source is not OCI/docker - but instead a SIF etc we should also have architecture known to carry forward.

cclerget:
> I don't really understand why this complex inspection is necessary?

And I'm asked the same, but as we can build from different sources like oras, shub, sandbox ,oci ... I opted for a general approach rather than relying on specifics of source formats, by example for a workflow :

```
sudo singularity run docker://multiarch/qemu-user-static --reset -p yes
singularity build --fakeroot -s /tmp/stage /tmp/ubuntu-s390.def
singularity exec --fakeroot /tmp/stage /install_script
singularity build /tmp/final.sif /tmp/stage
```

SIF will show native architecture not s390x

dtrudg:
@cclerget - ahh, right okay - all the others should have arch info, but I neglected to think of the sandbox. And with our plans to not have metadata in the rootfs I don't think we would have anything embedded in a sandbox showing the architecture in future.

:+1:

internal/pkg/util/machine/machine.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+// ArchFromContainer walks through a container filesystem until it
+// find an elf binary to read target architecture from and returns it.
+// If there is no suitable elf binary or if the architecture is not
+// recognized it will return an empty string.
+func ArchFromContainer(container string) string {
+	arch := ""
+
+	filepath.Walk(container, func(path string, info os.FileInfo, err error) error {
```
WestleyK, line 193: I'm not sure this is the best way to handle this;
 1. This process seems like it could take a while.
 2. What if the container has one or more binaries that are on a different architecture.

So, what about just looking at `/bin/sh`? Since that is the _only_ binary that we care about, should't we just check that?

WestleyK (changes requested):

internal/pkg/util/machine/machine.go
cclerget: It's true that `/bin/sh` should be the first to look at but many images especially from ngc/docker registries come without `/bin/sh` inside that's why the image is scanned but looking at `/bin/sh` first would clearly optimized the process.

This process doesn't take so long as it returns immediately when it find a suitable ELF binary and the process occurs at build step.
After thinking if there is a binary for a different architecture this is for a very specific use case which should be very rare and this potential case should be minimized by looking at `/bin/sh` first.

Another solution could be to extend use of `--arch` to local build image instead of `--remote` only but I don't like this idea too much as it's more error prone than the above because users could easily define wrong architecture and/or be annoyed by the "non standardization" of architecture flavors.

Overall this is process we can automate to make user experience better.

jmstover: > come without /bin/sh inside

Wait... wait wait wait.... What?!?

Is it that /bin/sh is supposed to be a symlink and that symlink was never created? (To dash, bash, or whatever...)

Or are they doing this as cut down images, and the binary being executed is the entrypoint ... so they remove everything else? So, you effectively have singularity 1.x images that were ldd walks of a given binary?

dtrudg: @jmstover  - very minimal container images are becoming a bit more common - especially with the uptake of Go for projects and being able to have static binaries. Eventually when we have metadata outside of the rootfs we can hopefully support these fully.

cclerget: @jmstover Yes micro-service images are often like that, they just embed the binary + libraries + conf/data files and let's go with this binary as entrypoint

dtrudg: The current code looking at the shell first seems to be workable here.

dtrudg:
Hi @WestleyK - do you still have concerns about the method to find the architecture? If not, could you approve or clear your request for changes. This would be a nice to have given some questions we've had around using Linux binfmt_misc / qemu support and so on.

Thanks.

internal/pkg/util/machine/machine.go
```diff
@@ -0,0 +1,308 @@
...
+// If there is no suitable elf binary or if the architecture is not
+// recognized it will return an empty string.
+func ArchFromContainer(container string) string {
+	// fast path if we can get architecture from shell binary
+	shell := fs.EvalRelative("/bin/sh", container)
+	arch, err := ArchFromElf(filepath.Join(shell, container))
+	if err == nil {
+		return arch
+	}
+
```
WestleyK, line 190: Would it make sense to add a warning/info/verbose here that says that we are not looking at bin/sh? Just to add a little more verbose if a user has multiple arch binaries in a container. Like:

```suggestion
	sylog.Verbosef("Cant find /bin/sh in container, looking at executable files to find architecture...")
```

Or something like that...

pkg/image/sif.go
```diff
@@ -12,6 +12,8 @@ import (
 	"runtime"
 	"syscall"
 
+	"github.com/sylabs/singularity/internal/pkg/util/machine"
+
```
WestleyK: Extra newline

```suggestion
```

internal/pkg/util/machine/machine.go
```diff
@@ -0,0 +1,308 @@
...
+	return arch
+}
+
+const binfmtMisc = "/proc/sys/fs/binfmt_misc"
+
+type binfmtEntry struct {
+	magic      string
+	enabled    bool
+	persistent bool
+}
```
WestleyK, line 233: Should this be declared at the top of the file? _Not critical..._

WestleyK (commented):
This looks good, just a couple small suggestions.

internal/pkg/util/machine/machine.go
cclerget: it doesn't hurt :) :+1:

internal/pkg/util/machine/machine.go
cclerget: As it's use only by `canEmulate` function located just after, it's more convenient for the reader to have the declaration near it

WestleyK (approved):
Looks good to me :+1:

PR #4640, opened 2019-10-18
Remove plugin compile artifacts

### LINKED ISSUE #4637: Plugin compile should remove artifacts other than SIF
Plugin compile produces .sif, but leftovers like .so and manifest.yaml are also left.
They should be removed.

sashayakovtseva:
Closes #4637


mem (approved):

tri-adam (approved):
LGTM

PR #4644, opened 2019-10-18
Check plugin path exists before compile

### LINKED ISSUE #3449: improve plug-in compile error when using an incorrect path
## Version of Singularity:

3.2.0rc2

Plug-in compile error could be improved when using incorrect path

```
$ singularity plugin compile examples/plugins/test-plugi
go: no such directory /home/vagrant/singularity/examples/plugins/test-plugi
FATAL:   Plugin compile failed with error: while building plugin .so: exit status 1
```

Perhaps when we run `go build` we should capture the output and include that in the error?

sashayakovtseva:
Closes #3449


mem (approved):

tri-adam (approved):
LGTM

PR #4658, opened 2019-10-22
Get plugin manifest in a separate process

### LINKED ISSUE #4638: Plugin Compile Fails when Plugin is Installed
### Version of Singularity:

```sh
$ singularity version
3.4.2-rc.1+477-g53ed0f10f
```

### Expected behavior

Running `singularity plugin compile` when a plugin with the same name is installed should succeed.

### Actual behavior

Running `singularity plugin compile` when a plugin with the same name is installed fails:

```sh
$ singularity plugin compile examples/plugins/cli-plugin/
$ sudo singularity plugin install examples/plugins/cli-plugin/cli-plugin.sif 
$ singularity plugin list
ENABLED  NAME
    yes  sylabs.io/cli-plugin
$ singularity plugin compile examples/plugins/cli-plugin/
FATAL:   Plugin compile failed with error: while generating plugin manifest: plugin.Open("/home/adam/src/singularity/examples/plugins/cli-plugin/plugin"): plugin already loaded
```

As above.

```
$ cat /etc/os-release
NAME="Ubuntu"
VERSION="18.04.3 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.3 LTS"
VERSION_ID="18.04"
VERSION_CODENAME=bionic
UBUNTU_CODENAME=bionic
```

From source.

pisarukv:
Fixes https://github.com/sylabs/singularity/issues/4638

### DIFF
--- a/internal/app/singularity/plugin_compile_linux.go
+++ b/internal/app/singularity/plugin_compile_linux.go
@@ -6,23 +6,26 @@
 package singularity
 
 import (
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
-	"plugin"
 	"runtime"
 	"strings"
 
 	uuid "github.com/satori/go.uuid"
 	"github.com/sylabs/sif/pkg/sif"
 	"github.com/sylabs/singularity/internal/pkg/sylog"
-	pluginapi "github.com/sylabs/singularity/pkg/plugin"
 )
+
+type buildToolchain struct {
+	goPath    string
+	workPath  string
+	buildTags string
+	envs      []string
+}
 
 // getSingularitySrcDir returns the source directory for singularity.
 func getSingularitySrcDir() (string, error) {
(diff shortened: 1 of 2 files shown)


cmd/plugin/manifest.go
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) 2018-2019, Sylabs Inc. All rights reserved.
```
mem: even if it's not the intention to produce a binary out of this (go build ...) it's still a good idea to follow the convention that programs are named after the directory where they are located. In this case the binary would be named "plugin" by default. I suspect "manifest" or "plugin-manifest" might be a more appropriate name?

mem: regarding the copyright statement, was this code copied from somewhere else to add "2018" there?

internal/app/singularity/plugin_compile_linux.go
```diff
@@ -87,69 +108,59 @@ func CompilePlugin(sourceDir, destSif, buildTags string) error {
...
 
-	manifest, err := json.Marshal(p.Manifest)
-	if err != nil {
-		return "", fmt.Errorf("could not marshal manifest to json: %v", err)
+	if err := runCmd.Run(); err != nil {
+		return "", fmt.Errorf("could not get manifest: %s", err)
 	}
 
-	if err := ioutil.WriteFile(out, manifest, 0644); err != nil {
+	if err := ioutil.WriteFile(out, manifest.Bytes(), 0644); err != nil {
```
mem: you could simply pass a second argument to the tool with the destination filename.

mem (commented):
Looks good to me.

What I'm missing is why this is necessary... could you please amend the commit message to include that information so that future developers know what this is about?

mem:
> Looks good to me.
> 
> What I'm missing is why this is necessary... could you please amend the commit message to include that information so that future developers know what this is about?

Ah, I had missed that bit of info in the issue.

I would still prefer to have the reason for this change documented in the commit's message. Thanks.

internal/app/singularity/plugin_compile_linux.go
pisarukv: Not sure what do you mean here. Could you explain a bit more? WriteFile takes bytes, not reader.

mem: I mean you could have the program write a file instead of spitting the manifest to stdout. If you do that, you don't need this at all, you just need to run the program, verify the exit code and return the filename, as you are doing now.

If you need to support writing the manifest to stdout, something line `manifest-gen input /dev/stdout` always works, or you could follow the convention that `manifest input -` means "write to stdout", or you could make the second argument optional, and fall back to stdout if it's not provided.

pisarukv: got it, will update. Thanks

dtrudg:
@cali4888 - would you be able to squash this down and put the brief explanation on the commit message long form rather than just the `-m "Get plugin manifest in a separate process"` then we can get this merged.

dtrudg (approved):

dtrudg:
Thanks @mem

PR #4667, opened 2019-10-23
Restore vendor/ for release tarballs

### LINKED ISSUE #4664: Provide `vendor/` in make dist tarball on master branch
### Version of Singularity:

```
master
```

### Expected behavior

A `vendor/` directory is included in the tarball output of `make dist` so that there is a source distribution that can be built without pulling go modules from the internet.

### Actual behavior

`vendor/` does not exist in the `make dist` tarball.

```
./mconfig
cd builddir && make dist
```

mem:
In order to be more friendly to people wanting to build singularity in
air-gapped or air-gapped-like situations (like Linux distributions),
restore the vendor directory in release tarballs (and only in that
case).

There's a simple assumption here: if there's a vendor directory, we must
pass -mod=vendor, and if there isn't, we pass -mod=readonly to make sure
that we are not unintentionally adding modules.

Since this affects the resulting binary interface, plugins need to be
aware of this. Make the corresponding adjustments there.

The vendor directory is created when running "make dist". To avoid
making larger changes, make it a requirement to use the make-dist script
from the makefile. It's still possible to use it directly if the user
runs "go mod vendor" manually _before_ calling scripts/make-dist.sh. In
the future all the scripts requiring access to the Go tool path, build
flags, etc, can move to sourcing a shell file with those settings.

Fixes: #4664

fixing or enhancing.

 - Fixes #

### DIFF
--- a/internal/app/singularity/plugin_compile_linux.go
+++ b/internal/app/singularity/plugin_compile_linux.go
@@ -93,6 +93,18 @@ func buildPlugin(sourceDir, buildTags string) (string, error) {
 		return "", errors.New("singularity source directory not found")
 	}
 
+	modFlag := "-mod=readonly"
+
+	hasVendor := func() bool {
+		vendorDir := filepath.Join(workpath, "vendor")
+		_, err := os.Stat(vendorDir)
+		return !os.IsNotExist(err)
+	}()
+
+	if hasVendor {
+		modFlag = "-mod=vendor"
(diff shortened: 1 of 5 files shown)


dtrudg (approved):
This looks as elegant/simple as it's gonna get and I confirmed `rm -r` on the extracted tarball dir works without messing around.

PR #4704, opened 2019-11-05
Fix container mount destination containing symlink

### LINKED ISSUE #4702: Fatal error binding nvidia persistenced socket
### Version of Singularity:

```
3.5.0-rc.1+4-ga7c731b22
```

### Expected behavior

On a machine with the nvidia-persistenced running, a container starts with `--nv` correctly.

This works correctly on 3.4.2:

```
dave@ythel~> singularity run --nv docker://ubuntu
dave@ythel:~
10:27 AM $ 
```

### Actual behavior

On a machine with the nvidia-persistenced running, a container does not start with `--nv` correctly

```
dave@ythel~> singularity run --nv docker://ubuntu
...
INFO:    Creating SIF file...
FATAL:   container creation failed: mount /var/run/nvidia-persistenced/socket->/var/run/nvidia-persistenced/socket error: while mounting /var/run/nvidia-persistenced/socket: destination /var/run/nvidia-persistenced/socket doesn't exist in container
```

Try using `--nv` on a machine with nvidia drivers installed *and* the `nvidia-persistenced` running

```
Debian 10.1
```

Git tree

cclerget:

Fix overlay/underlay layers to resolve destination relative to the container root filesystem, `fs.EvalRelative` was reworked by replacing `filepath.EvalSymlinks` with `os.Readlink` in order to minimize false positive and follow links relatively to the root destination.

 - Fixes #4702 

### DIFF
--- a/internal/pkg/util/fs/helper.go
+++ b/internal/pkg/util/fs/helper.go
@@ -134,34 +134,117 @@ func RootDir(path string) string {
 	return p
 }
 
-// EvalRelative evaluates symlinks in path relative to root path. This
-// function doesn't return error but always returns an evaluated path.
-func EvalRelative(path string, root string) string {
-	splitted := strings.Split(filepath.Clean(path), string(os.PathSeparator))
-	dest := string(os.PathSeparator)
+// walkSymRelative follows and resolves all symlinks found in a path
+// located in root, it ensures symlinks resolution never go past the
+// provided root path.
+func walkSymRelative(path string, root string, maxLinks uint) string {
+	// symlinks counter
+	symlinks := uint(0)
 
-	for i := 1; i < len(splitted); i++ {
-		s := splitted[i : i+1][0]
-		dest = filepath.Join(dest, s)
+	// generate clean absolute path
+	absRoot := filepath.Join("/", root)
+	absPath := filepath.Join("/", path)
 
-		if s != "" {
-			rootDestPath := filepath.Join(root, dest)
-			for {
-				target, err := filepath.EvalSymlinks(rootDestPath)
-				if err != nil {
-					break
-				}
-				if !strings.HasPrefix(target, root) {
-					rootDestPath = filepath.Join(root, target)
-					continue
-				}
-				dest = strings.Replace(target, root, "", 1)
-				break
-			}
+	sep := string(os.PathSeparator)
+
+	// get path components by skipping the first
+	// character as it will be always "/" to avoid
+	// to add an empty first element in the array
+	comp := strings.Split(absPath[1:], sep)
+
+	// start from absolute root path
+	dest := absRoot
+
+	for i := 0; i < len(comp); i++ {
+		dest = filepath.Join(dest, comp[i])
+
+		// this call can return various errors that we don't need
+		// or want to deal with like a lack of permission for a
+		// directory traversal, not a symlink, non-existent path.
+		// As this function doesn't return any error we ignore them
+		// and generate the path assuming there is no hidden symlink
+		// in the next components
+		d, err := os.Readlink(dest)
+		if err != nil {
+			continue
 		}
+		symlinks++
+
+		newDest := absRoot
+
+		if !filepath.IsAbs(d) {
+			// this is a relative target, we are taking the current
+			// parent path of dest concatenated with the target
+			parentDest := filepath.Dir(dest)
+			dest = filepath.Join(parentDest, d)
+
+			// if we are outside of root, we join the relative target
+			// with "/" to obtain an absolute path as we were at root
+			// of "/" thanks to filepath.Clean implicitly called by
+			// filepath.Join
+			if !strings.HasPrefix(dest, absRoot) {
+				d = filepath.Join("/", d)
+				dest = filepath.Join(absRoot, d)
+			} else {
+				if strings.HasPrefix(dest, parentDest) {
+					// trivial case where the resulting path is
+					// within the current path
+					d = strings.TrimPrefix(dest, parentDest)
+					newDest = parentDest
+				} else {
+					// we go back in the hierarchy and take a
+					// naive approach by trimming root prefix
+					// from path instead of finding the exact
+					// path chunk involving too much complexity
+					d = strings.TrimPrefix(dest, absRoot)
+				}
+			}
+		} else {
+			// it's an absolute path, simply concatenate root
+			// and symlink target
+			dest = filepath.Join(absRoot, d)
+		}
+
+		// too many symbolic links, stop and return the
+		// resolved path as is, should not happen with
+		// sane images
+		if symlinks == maxLinks {
+			break
+		}
+		// symlink target point to the current destination,
+		// nothing to do
+		if len(d) == 0 {
+			continue
+		}
+
+		dest = newDest
+		// either replace current path components or merge
+		// the components of the symlink target with the next
+		// components
+		if i+1 < len(comp) {
+			comp = append(strings.Split(d[1:], sep), comp[i+1:]...)
+		} else {
+			comp = strings.Split(d[1:], sep)
+		}
+		i = -1
 	}
 
-	return dest
+	// ensure the final path is absolute
+	return filepath.Join("/", strings.TrimPrefix(dest, absRoot))
+}
+
+// EvalRelative evaluates symlinks in path relative to root path, it returns
+// a path as if it was evaluated from chroot. This function always returns
+// an absolute path and is intended to be used to resolve mount points
+// destinations, it helps the runtime to not bind mount directories/files
+// outside of the container image provided by the root argument.
+func EvalRelative(path string, root string) string {
+	// return "/ if path is empty
+	if path == "" {
+		return "/"
+	}
+	// resolve path and allow up to 40 symlinks
+	return walkSymRelative(path, root, 40)
 }
 
 // Touch behaves like touch command.
--- a/internal/pkg/util/fs/helper_linux_test.go
+++ b/internal/pkg/util/fs/helper_linux_test.go
@@ -295,20 +295,38 @@ func TestEvalRelative(t *testing.T) {
 	// - /bin/sbin -> ../sbin
 	// - /sbin/sbin2 -> ../../sbin
 
+	// symlink $TMP/bin -> usr/bin
 	os.Symlink("usr/bin", filepath.Join(tmpdir, "bin"))
+	// symlink $TMP/sbin -> usr/sbin
 	os.Symlink("usr/sbin", filepath.Join(tmpdir, "sbin"))
 
+	// directory $TMP/usr/bin
 	MkdirAll(filepath.Join(tmpdir, "usr", "bin"), 0755)
+	// directory $TMP/usr/sbin
 	MkdirAll(filepath.Join(tmpdir, "usr", "sbin"), 0755)
 
+	// symlink $TMP/usr/bin/bin -> /bin
 	os.Symlink("/bin", filepath.Join(tmpdir, "bin", "bin"))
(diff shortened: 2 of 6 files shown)


internal/pkg/util/fs/layout/layer/underlay/underlay_linux.go
```diff
@@ -143,14 +155,14 @@ func (u *Underlay) createLayer(rootFsPath string, system *mount.System) error {
 
 func (u *Underlay) duplicateDir(dir string, system *mount.System, existingPath string) error {
 	binds := 0
-	path := filepath.Clean(u.session.RootFsPath() + dir)
```
dtrudg, line 146: When eliminating `Clean` here, is there any impact if `dir` is e.g. `../../../../` ?

internal/pkg/util/fs/helper_linux_test.go
```diff
@@ -311,17 +311,17 @@ func TestEvalRelative(t *testing.T) {
 	}{
 		{"/bin", "/usr/bin"},
 		{"/sbin", "/usr/sbin"},
-		{"/bin/bin", "/usr/bin"},
```
dtrudg, line 314: I'm pretty confused here. This looks like it used to resolve down to an actual directory `/usr/bin` but now the result `/bin` still includes a symlink?

dtrudg (commented):
Sorry - I'm kinda confused on this one. It seems to fix the issue - but I'm honestly unclear how this function is used entirely, and why the change is generally reasonable.

Would you be able to give:

1. An example of the old behavior
2. An example of the new behavior
3. A brief explanation of how this fixes the issue, and why it doesn't affect anything else?

internal/pkg/util/fs/layout/layer/underlay/underlay_linux.go
cclerget: Because `filepath.Join` call `filepath.Clean` on returned result

internal/pkg/util/fs/helper_linux_test.go
cclerget: Previous version was broken and resolved links outside of the relative root

internal/pkg/util/fs/layout/layer/underlay/underlay_linux.go
dtrudg: :+1: I was not aware of that, apologies.

internal/pkg/util/fs/helper_linux_test.go
dtrudg: Sorry I'm not clear what "relative root" means? Do you mean that if the `root` contains a symlink then only symlinks in `path` that are *not* in `root` are resolved?

cclerget:
> Sorry - I'm kinda confused on this one. It seems to fix the issue - but I'm honestly unclear how this function is used entirely, and why the change is generally reasonable.
> 
> Would you be able to give:
> 
>     1. An example of the old behavior
> 
>     2. An example of the new behavior
> 
>     3. A brief explanation of how this fixes the issue, and why it doesn't affect anything else?

1. With an image containing a smylink like `/image/var/run -> /run`, the old `fs.EvalRelative` resolved links to the host instead, path was valid as long as host contains the same symlink, the side effect was that overlay/underlay checked path and didn't create it because of presence or not on the host, basically the behavior was inconsistent depending of host FS layout and additionally of the bind path provided (eg: `/var/run/nvidia/...` VS `/run/nvidia/...`)

2. Change applied to `fs.EvalRelative` now ensure we stay and resolve links inside the provided relative root path, so we are not impacted by host FS layout and stay in the container root filesystem. Changes in overlay/underlay ensure that real destination directories/files are created correctly as mount follow symlinks and bind to the real path.

3. As explained in 1 and 2. It also affect OCI engine mount operations, overall this is better and the potential issues which could be introduced by those changes should be dependent of image FS layout with crazy symlinks in it.

internal/pkg/util/fs/helper_linux_test.go
cclerget: See explanations below

internal/pkg/util/fs/helper.go
```diff
@@ -134,34 +134,53 @@ func RootDir(path string) string {
...
+			d, err := os.Readlink(dest)
+			if err != nil {
+				continue
+			}
+			if filepath.IsAbs(d) {
+				dest = filepath.Join(cleanRoot, d)
+			} else {
+				dest = filepath.Join(filepath.Dir(dest), d)
+				if !strings.HasPrefix(dest, cleanRoot) {
+					dest = filepath.Join(filepath.Dir(dest), d)
```
dtrudg: Isn't this a duplicate of line 174 that's not needed?

cclerget: yep will fix that. Thanks !

internal/pkg/util/fs/helper.go
```diff
@@ -134,34 +134,53 @@ func RootDir(path string) string {
 	return p
 }
 
-// EvalRelative evaluates symlinks in path relative to root path. This
-// function doesn't return error but always returns an evaluated path.
+// EvalRelative evaluates symlinks in path relative to root path, it returns
```
mem: After going thru the code several times, I think I understand what this means, and I'm having a hard time finding a suggestion as to how to change this so that other people can understand it.

First, path is _not_ a valid path in the host, and it looks like it _must_ be a valid path in the container. That's what I guess "relative to root path" means.

The bit about "it returns a path as if it was evaluated from chroot" means that the result of this function is a valid path rooted at rootfs (not relative to it). What I mean is that "relative to rootfs" would make me thing that the result of this function is a relative path (x/y/z, relative to /a/b/c) and in fact it's an absolute path rooted at rootfs (/x/y/z, with root being /a/b/c).

The other bit of documentation that is missing here is why this is needed, or in other words, when would I need to use this? What I mean is that if path is valid in the container, why would I need to fudge with it at all? I _suspect_ that part of the confusion comes from the name of this function...

internal/pkg/util/fs/helper.go
```diff
@@ -134,34 +134,53 @@ func RootDir(path string) string {
...
-				if err != nil {
-					break
-				}
-				if !strings.HasPrefix(target, root) {
-					rootDestPath = filepath.Join(root, target)
-					continue
+	cleanRoot := filepath.Clean(root)
+	cleanPath := filepath.Clean(path)
+
+	if cleanPath == "" {
```
mem: trying to understand here...

filepath.Clean should not return an empty value, so I guess path was empty to start with?

internal/pkg/util/fs/helper.go
```diff
@@ -134,34 +134,53 @@ func RootDir(path string) string {
...
+		return cleanRoot
```
dtrudg: If `path=/1/2/3` and `root=/a/b/c` then we evaluate `/1/2/3` under `/a/b/c` and the return doesn't include `/a/b/c`?

If `path=""` then the return is `root` - this is a little odd compared to the above. Is something done with that `root` value? Is it some sort of safety measure?

internal/pkg/util/fs/helper.go
```diff
@@ -134,34 +134,53 @@ func RootDir(path string) string {
...
+	}
+
+	sep := string(os.PathSeparator)
+	splitted := strings.Split(cleanPath, sep)
+
+	i := 1
```
mem:
```
i := 0
if filepath.IsAbs(cleanPath) {
    // path is absolute, skip the first empty element
    // that resulted from calling strings.Split above.
    i = 1
}
```

note that the index itself is not really important, so this is more or less equivalent to doing something like:

```
dest := cleanRoot

components := strings.Split(cleanPath, string(os.PathSeparator))
if filepath.IsAbs(cleanPath) {
    components = components[1:]
}

for _, component := range components {
}
```

internal/pkg/util/fs/helper.go
```diff
@@ -134,34 +134,53 @@ func RootDir(path string) string {
...
+	}
+
+	dest := cleanRoot
+
+	for ; i < len(splitted); i++ {
+		dest = filepath.Join(dest, splitted[i])
+
+		fi, err := os.Lstat(dest)
+		if err != nil {
+			continue
```
mem, line 169: this continue here needs a comment, specially this this is not checking the kind of error that it's getting back. What exactly are we skipping here and why?

internal/pkg/util/fs/helper.go
```diff
@@ -134,34 +134,53 @@ func RootDir(path string) string {
...
+		}
+
+		if fi.Mode()&os.ModeSymlink != 0 {
```
mem: just to avoid indenting the code too much, I would do:

```
if fi.Mode() & os.ModeSymlink == 0 {
    // not a symlink, continue
    continue
}
```

internal/pkg/util/fs/helper.go
```diff
@@ -134,34 +134,53 @@ func RootDir(path string) string {
...
+			d, err := os.Readlink(dest)
+			if err != nil {
+				continue
```
mem: again, what are we skipping and why?

internal/pkg/util/fs/helper.go
```diff
@@ -134,34 +134,53 @@ func RootDir(path string) string {
...
+			}
+			if filepath.IsAbs(d) {
+				dest = filepath.Join(cleanRoot, d)
```
mem: this one needs a comment as well...

We got an absolute path from the symlink, and we want to make it relative to cleanRoot. I know it's obvious now, but when reading this code I had to stop and think about what this is trying to do.

internal/pkg/util/fs/helper.go
```diff
@@ -134,34 +134,53 @@ func RootDir(path string) string {
...
+			} else {
+				dest = filepath.Join(filepath.Dir(dest), d)
```
mem: this line is really tricky and needs documentation...

Someone reading this will ask why is this chopping the last element of dest?

I removed the component that was added on line 159 in order to evaluate the relative symlink relative to the directory containing the new element, in other words, I go back to the containing directory to evaluate the symlink relative to that directory.

internal/pkg/util/fs/helper.go
```diff
@@ -134,34 +134,53 @@ func RootDir(path string) string {
...
+				if !strings.HasPrefix(dest, cleanRoot) {
```
mem: Please add a comment...

if the resulting path falls outside cleanRoot (e.g. the symlink was something like ..), then we do ... what?

At this point, my intuition says "error out", so this code must be here for some reason?

internal/pkg/util/fs/helper.go
mem: dest already had the element added on line 159 chopped off. That one I can understand.

Why is this removing another element?

internal/pkg/util/fs/helper.go
```diff
@@ -134,34 +134,53 @@ func RootDir(path string) string {
...
+					dest = filepath.Join(cleanRoot, dest)
 				}
-				dest = strings.Replace(target, root, "", 1)
-				break
 			}
 		}
 	}
 
-	return dest
+	return strings.Replace(dest, cleanRoot, "", 1)
```
mem: Sorry to nag, but how is this different from `dest = strings.TrimPrefix(dest, cleanRoot)`?

If there's a special case that's being covered by this, could you please add a comment?

mem (approved):

mem (dismissed):
I only meant to post comments :-(

dtrudg (approved):
Thanks @cclerget - it's much faster to see the intention, flow here now.

dtrudg:
NB - CI failure was a library download timeout - should be okay on re-run

PR #4707, opened 2019-11-05
Move logic deciding whether to run E2E to its own script

mem:
This makes it easier to test and update the code.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,42 +64,10 @@ x-run:
   check_changes: &check_changes
     name: Check changes
     command: |-
-      # Skip e2e by default
-      require_e2e=false
-
-      pull_info_file=$(mktemp)
-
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -64,42 +64,10 @@ x-run:
...
-              ;;
-          *)
-              # The branch is not master or release, skip e2e
-              require_e2e=false
-              ;;
-      esac
-
-      if ! "$require_e2e" ; then
-        echo "No changes in critical components, skipping tests"
+      if scripts/should-e2e-run ; then
```
jscook2345, line 67: could we redirect the output to /dev/null in case some error pops-up in the future and causes the script to output something that looks like a successful return?

mem: Later, sure. At this point, I'd like to at least have some output in case something goes wrong.

jscook2345 (approved):

PR #4710, opened 2019-11-06
Move logic deciding whether to run E2E to its own script

mem:
This makes it easier to test and update the code.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,44 +64,23 @@ x-run:
   check_changes: &check_changes
     name: Check changes
     command: |-
-      # Skip e2e by default
-      require_e2e=false
+      rc=0
+      scripts/should-e2e-run "${CIRCLE_PROJECT_USERNAME}" "${CIRCLE_PROJECT_REPONAME}" "${CIRCLE_PULL_REQUEST}" || rc=$?
+      case $rc in
+        0)
+          echo "Verifying critical changes"
+          ;;
 
-      pull_info_file=$(mktemp)
+        1)
+          echo "No critical changes, skipping tests"
+          circleci step halt
+          ;;
 
-      curl -s "https://api.github.com/repos/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}/pulls/${CIRCLE_PR_NUMBER}" > "${pull_info_file}"
-
-      # Keep the matching lines instead of simply looking at the exit
-      # code for debugging purposes
-      E2E_LABELS=$(jq -r '.labels[].name' < "${pull_info_file}" | grep --line-regexp --fixed-strings ci:e2e || true)
-
-      if test -n "${E2E_LABELS}" ; then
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -64,42 +64,10 @@ x-run:
...
-              ;;
-          *)
-              # The branch is not master or release, skip e2e
-              require_e2e=false
-              ;;
-      esac
-
-      if ! "$require_e2e" ; then
-        echo "No changes in critical components, skipping tests"
+      if scripts/should-e2e-run ; then
```
dtrudg: We need to pass the user/repo/pr here, right?

```
E: Invalid usage. Abort.

Usage:

    scripts/should-e2e-run {github_username} {github_reponame} {pr_number}
+ echo 'No critical changes, skipping tests'
No critical changes, skipping tests
+ circleci step halt
```

dtrudg (changes requested):

.circleci/config.yml
dtrudg: `scripts/should-e2e-run $CIRCLE_PR_USERNAME $CIRCLE_PR_REPONAME $CIRCLE_PR_NUMBER` ?

scripts/should-e2e-run
```diff
@@ -0,0 +1,76 @@
+#!/bin/sh
```
cclerget, line 1: Could this be moved to `.circleci` directory instead to not mix CI and build related scripts ?

.circleci/config.yml
mem: Yeah, thanks for catching that. I could have sworn I had that...

scripts/should-e2e-run
mem: No, I'd prefer to have things in the same visible location than reproduce the mess that is the Travis setup.

cclerget: And reproduce the same mess in `scripts` :/ ?

mem: I'll just assume you haven't looked at how travis is set up and that's why my comment is not understood.

Thanks!

.circleci/config.yml
```diff
@@ -64,45 +64,24 @@ x-run:
   check_changes: &check_changes
     name: Check changes
     command: |-
-      # Skip e2e by default
-      require_e2e=false
+      rc=0
+      scripts/should-e2e-run "${CIRCLE_PROJECT_USERNAME}" "${CIRCLE_PROJECT_REPONAME}" "${CIRCLE_PULL_REQUEST}" || rc=$?
```
mem, line 68: For some reason the other variable is empty. Looking at the environment where _this_ check is running, that variable is not available. Only CIRCLE_PULL_REQUEST is. So use that and extract the PR number that we need from there.

dtrudg: heh - I just got it from their docs page somewhere. fair enough!

dtrudg (approved):

PR #4721, opened 2019-11-07
Fix "build_and_test" workflow failure on push

mem:
In CircleCI we (unintentionally) configured the build_and_test workflow
to run both on pull requests and pushes, and in the later case the
workflow is failing because no pull request URL is available.

Fix this by handling the case explicitly and assume that pushes MUST run
E2E tests. In that way, even commits that didn't require E2E to run will
get the benefit of testing (after being merged).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,7 +65,7 @@ x-run:
     name: Check changes
     command: |-
       rc=0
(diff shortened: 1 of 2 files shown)


dtrudg (approved):
Makes sense to me :+1:

PR #4726, opened 2019-11-12
CLI cleanup for sign/verify

WestleyK:

This PR cleanups the cli code for sign/verify.

 - Fixes #

### DIFF
--- a/cmd/internal/cli/sign.go
+++ b/cmd/internal/cli/sign.go
@@ -99,7 +99,6 @@ var SignCmd = &cobra.Command{
 
 	Run: func(cmd *cobra.Command, args []string) {
 		// args[0] contains image path
-		fmt.Printf("Signing image: %s\n", args[0])
 		doSignCmd(cmd, args[0])
 	},
 
@@ -110,36 +109,12 @@ var SignCmd = &cobra.Command{
 }
 
 func doSignCmd(cmd *cobra.Command, cpath string) {
-	// Group id should start at 1.
-	if cmd.Flag(verifySifGroupIDFlag.Name).Changed && sifGroupID == 0 {
-		sylog.Fatalf("invalid group id")
+	id, isGroup, err := ensureImageFlags(cmd, cpath, sifDescID, sifGroupID, signAll)
+	if err != nil {
+		sylog.Fatalf("%s", err)
 	}
 
-	// Descriptor id should start at 1.
-	if cmd.Flag(verifySifDescSifIDFlag.Name).Changed && sifDescID == 0 {
-		sylog.Fatalf("invalid descriptor id")
-	}
-	if cmd.Flag(verifySifDescIDFlag.Name).Changed && sifDescID == 0 {
-		sylog.Fatalf("invalid descriptor id")
-	}
-
-	if sifGroupID != 0 && sifDescID != 0 {
-		sylog.Fatalf("only one of -i or -g may be set")
-	}
-
-	var isGroup bool
-	var id uint32
-	if sifGroupID != 0 {
-		isGroup = true
-		id = sifGroupID
-	} else {
-		id = sifDescID
-	}
-
-	if (id != 0 || isGroup) && signAll {
-		sylog.Fatalf("'--all' not compatible with '--sif-id' or '--groupid'")
-	}
-
+	fmt.Printf("Signing image: %s\n", cpath)
 	if err := signing.Sign(cpath, id, isGroup, signAll, privKey); err != nil {
 		sylog.Fatalf("Failed to sign container: %s", err)
 	}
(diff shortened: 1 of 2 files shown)


cmd/internal/cli/verify.go
```diff
@@ -141,44 +136,64 @@ var VerifyCmd = &cobra.Command{
...
+	if err == signing.ErrVerificationFail {
+		sylog.Fatalf("Failed to verify: %s", cpath)
+	} else if err != nil {
+		sylog.Fatalf("Failed to verify: %s: %s", cpath, err)
+	}
+	sylog.Infof("Container verified: %s", cpath)
+}
+
+// ensureImageFlags takes a container path, sif id and group id, and formats to a more usable output.
+func ensureImageFlags(cmd *cobra.Command, cpath string, descrID, groupID uint32, all bool) (uint32, bool, error) {
```
dtrudg: Can you comment a little more on the expected output here? It should e.g. be very obvious what the `all` boolean will return - if that's a special number that means 'everything'. Named returns would probably be helpful here too to clarify things.

cmd/internal/cli/sign.go
```diff
@@ -110,36 +109,12 @@ var SignCmd = &cobra.Command{
...
-	}
-
-	if (id != 0 || isGroup) && signAll {
-		sylog.Fatalf("'--all' not compatible with '--sif-id' or '--groupid'")
+	id, isGroup, err := ensureImageFlags(cmd, cpath, sifDescID, sifGroupID, signAll)
+	if err != nil {
+		sylog.Fatalf("%s", err)
 	}
 
+	fmt.Printf("Signing image: %s\n", cpath)
```
dtrudg, line 128: Do we ever need the `-s` silent flag to work on sign? I'm not sure if it is needed.

cmd/internal/cli/verify.go
```diff
@@ -141,44 +136,64 @@ var VerifyCmd = &cobra.Command{
...
+	if err != nil {
+		sylog.Fatalf("%s", err)
+	}
+
+	// Dont need to resolve remote endpoint.
+	if !localVerify {
+		handleVerifyFlags(cmd)
+	}
+
+	author, _, err := signing.Verify(ctx, cpath, url, id, isGroup, verifyAll, authToken, localVerify, jsonVerify)
```
dtrudg, line 160: We pass `verifyAll` here - but we are also using it in the call to `ensureImageFlags` which returns `id, isGroup`. This is a bit confusing - as it is not immediately clear if `id, isGroup` are ignored, or affected when `verifyAll` is true.

cmd/internal/cli/verify.go
dtrudg: Would `ensureImageFlags` be better called something like `sifIDFromFlags` or similar? `ImageFlags` is a little generic to me.

dtrudg (changes requested):

cmd/internal/cli/sign.go
WestleyK: The `-s` flag probably will not be very effective on `sign`, thats not something that will be fixed in this PR. And this `fmt.Printf(...)` was already there, I just moved it to a better location.

cmd/internal/cli/verify.go
WestleyK: We pass `verifyAll` to `ensureImageFlags()` only to check that the `--sif-id`/`--group-id` are not used at the same time. Since we will need to check that for sign and verify, I just added that check to that function, just so theres not duplicate code, witch is what this PR is supposed to improve.

cmd/internal/cli/verify.go
WestleyK: Yes, I will fix this function name, and comment to be more appropriate with what this function does. In this case, we pass the `all` here just to check to make sure that we are not using the `--sif-id`/`--group-id` flag at the same time.

WestleyK: I changed it to `checkImageAndFlags`, and updated the comment. Does that sound a little better?

dtrudg: >  In this case, we pass the all here just to check to make sure that we are not using the --sif-id/--group-id flag at the same time.

this is the type of info which would be good in the comment really. The updated comment still doesn't really tell us what the function is/should be doing. Maybe something a bit more like...

```
// checkImageAndFlags verifies that the SIF image and SIF descriptor / group ID flags provided are usable, and returns which SIF ID to sign or verify:
//  - checks that a descrID and groupID are not being used together
//  - checks that a descrID or groupID are not being used with a request to verify all
// -  returns an error if flags are not usable
//  - returns an ID to sign or verify, and true if it is a group ID / false if it is a descriptor ID
```

I think named returns would be good here to. Then in a nice editor we can easily see that the `bool` means `isGroup` from the function definition that pops up.

dtrudg (approved):

PR #4737, opened 2019-11-13
Fix disk space leak when building from docker-archive

### LINKED ISSUE #4560: docker-daemon:// builder does not clean up all of its tempfiles.
### Version of Singularity:

What version of Singularity are you using? 
```
$ singularity version
3.4.0-1
```

```
$docker version
docker version
Password: 
Client:
 Version:           18.09.9
 API version:       1.39
 Go version:        go1.12.9
 Git commit:        039a7df
 Built:             Tue Sep 24 13:23:49 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          18.09.9
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.12.9
  Git commit:       039a7df
  Built:            Tue Sep 24 13:23:15 2019
  OS/Arch:          linux/amd64
  Experimental:     false
```
### Expected behavior

When you build an image using the `docker-daemon://` builder it should at a minimum temp files created by this process should be cleaned up by `singularity cache clean` or a similar command, but I can't seem to find one.  Preferably it would just get cleaned up when the build finishes without an additional command

### Actual behavior

`singularity build $img docker-daemon://$docker_container_name` leave a potentially large file(s) in `/var/tmp/docker-tar-${some_hash}` 

build any container using the `docker-daemon://` builder

```
$ cat /etc/os-release 
NAME=Gentoo
ID=gentoo
PRETTY_NAME="Gentoo/Linux"
ANSI_COLOR="1;32"
HOME_URL="https://www.gentoo.org/"
SUPPORT_URL="https://www.gentoo.org/support/"
BUG_REPORT_URL="https://bugs.gentoo.org/"
```

Installed from source using an [ebuild that I wrote](https://bugs.gentoo.org/694048) since the upstream one hasn't been updated in a while.  I'm trying to get it merged.

C-Elegans:

When building a singularity .sif from a docker archive, singularity would
create a temporary file named docker-tarXXX in /var/tmp, and never clean it up.
This adds a line to calculateRefHash (which was creating the file via
NewImageSource()) to close the image source it created, cleaning up the
temporary file.

 - Fixes #4560 

### DIFF
--- a/internal/pkg/client/oci/oci.go
+++ b/internal/pkg/client/oci/oci.go
@@ -18,6 +18,7 @@ import (
 	"github.com/containers/image/signature"
 	"github.com/containers/image/transports"
 	"github.com/containers/image/types"
+	"github.com/pkg/errors"
 	"github.com/sylabs/singularity/internal/pkg/client/cache"
 	"github.com/sylabs/singularity/internal/pkg/sylog"
 )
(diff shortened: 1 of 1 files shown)


internal/pkg/client/oci/oci.go
```diff
@@ -114,11 +115,16 @@ func ImageSHA(ctx context.Context, uri string, sys *types.SystemContext) (string
 	return calculateRefHash(ctx, ref, sys)
 }
 
-func calculateRefHash(ctx context.Context, ref types.ImageReference, sys *types.SystemContext) (string, error) {
+func calculateRefHash(ctx context.Context, ref types.ImageReference, sys *types.SystemContext) (retStr string, retErr error) {
```
dtrudg: Please could you use `err` as we are using elsewhere in the code.

Also, instead of retStr the return can be called `hash` as it is in the function body.

dtrudg (changes requested):
Many thanks for this!

Minor request on the named returns from the function.

internal/pkg/client/oci/oci.go
C-Elegans: Sure!

dtrudg (approved):

PR #4744, opened 2019-11-15
Save/restore stack size limit with setuid workflow

### LINKED ISSUE #4701: HostOS Stack size limit not honored inside the container
### Version of Singularity:

3.2.1

### Expected behavior

Have the same stack size limit inside and outside the container

### Actual behavior

The stack size limit is resetted to the default value of 8192

singularity pull myalpine docker://alpine
singularity pull mydebian docker://debian

ulimit -a | grep stack
stack size              (kbytes, -s) unlimited

singularity shell myalpine 
Singularity> ulimit -a | grep stack
-s: stack size (kb)                8192

singularity shell mydebian
Singularity> ulimit -a | grep stack
stack size              (kbytes, -s) 8192

debian9

Compiled from source

cclerget:

The stack size limit is reset to its default value when setuid workflow is used, so the original value is saved and restored before executing the container process for Singularity engine.

 - Fixes #4701

### DIFF
--- a/cmd/internal/cli/actions_linux.go
+++ b/cmd/internal/cli/actions_linux.go
@@ -35,6 +35,7 @@ import (
 	"github.com/sylabs/singularity/pkg/util/crypt"
 	"github.com/sylabs/singularity/pkg/util/gpu"
 	"github.com/sylabs/singularity/pkg/util/namespaces"
+	"github.com/sylabs/singularity/pkg/util/rlimit"
 )
 
 // EnsureRootPriv ensures that a command is executed with root privileges.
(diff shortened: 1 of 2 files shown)


dtrudg:
Hi @cclerget - couple of thoughts about this...

- I think that carrying limits like this from the host makes containers less portable? If an uncommon limit is needed, then setting in the container should give better portability, and also raise an error if the stack limit can't be set high enough for what's going to be run.
- There are a lot of other limits.... what else is/isn't reset by the setuid workflow?

cclerget:
Hi @dctrud ,

> I think that carrying limits like this from the host makes containers less portable?

Singularity just read soft/hard limit and restore them as is for the container process which is then free to set it to any (or authorized values if enforced in `/etc/security/limits.conf`), so it doesn't impact the portability, an admin could impact some container requiring to run with a higher limit that the one enforced on the system by him.

Other limits seems to be not impacted by the setuid workflow

dtrudg (approved):

PR #4760, opened 2019-11-19
Add e2e tests for various configuration directives

### LINKED ISSUE #4061: Add e2e tests for various singularity.conf values
Add a test for each configuration directives in `singularity.conf`

cclerget:

Add e2e tests to execute singularity with various configuration directives value

 - Fixes #4061

### DIFF
--- a/e2e/suite.go
+++ b/e2e/suite.go
@@ -22,6 +22,7 @@ import (
 	e2ebuildcfg "github.com/sylabs/singularity/e2e/buildcfg"
 	"github.com/sylabs/singularity/e2e/cache"
 	"github.com/sylabs/singularity/e2e/cmdenvvars"
+	"github.com/sylabs/singularity/e2e/config"
 	"github.com/sylabs/singularity/e2e/delete"
 	"github.com/sylabs/singularity/e2e/docker"
 	singularityenv "github.com/sylabs/singularity/e2e/env"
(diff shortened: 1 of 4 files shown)


e2e/suite.go
```diff
@@ -73,12 +74,34 @@ func Run(t *testing.T) {
...
+	testenv.TestDir = name
+
 	// e2e tests need to run in a somehow agnostic environment, so we
 	// don't use environment of user executing tests in order to not
 	// wrongly interfering with cache stuff, sylabs library tokens,
 	// PGP keys
 	e2e.SetupHomeDirectories(t)
 
+	// generate singularity.conf with default values
+	e2e.SetupDefaultConfig(t, filepath.Join(testenv.TestDir, "singularity.conf"))
```
dtrudg, line 103: How does Singularity use this config file in TestDir at all? If we are calling an installed `singularity` executable then isn't it using the one in `sysconfdir`?

Can you explain why it's necessary to do this SetupDefaultConfig into TestDir?

dtrudg (commented):
Question on the singularity.conf in TestDir.

e2e/suite.go
cclerget: The idea is to not let the current `singularity.conf` influence tests and use the default values, that's why a config is generated as if it was the first installation of singularity and the file is bind mount on top of the current : https://github.com/sylabs/singularity/pull/4760/files#diff-c95326473ca3935b0c44e93cde3b7059R35

This way current configuration set by the dev won't conflicts with e2e tests, actually if you put `bind path = /my/nfs/share` it's bound into container during e2e run because install procedure doesn't reset configuration file but instead update it if necessary (like new directive) and reuse current value to generate the new one.

dtrudg (approved):

PR #4776, opened 2019-11-22
Small Hotfix for Issue #2860 - All files in container owned by root

### LINKED ISSUE #2860: All files in container owned by root
## Version of Singularity:

`3.0.3+ds`

### Expected behavior

When creating an image from a sandbox as root, files inside the container should retain their owner.
This was the case with singularity 2.x.

### Actual behavior

All files in the container are owned by root even though the container was build via root (i.e., `-all-root` does not get set for `mksquashfs`).

1. Create sandbox from recipe.
```
sudo -E singularity build --sandbox "${TARGET_FOLDER}" "${RECIPE}"
```
2. Create image from sandbox.
```
sudo singularity build "${IMAGE_NAME}" "${TARGET_FOLDER}"
```

### Notes

I noticed that instead of creating the image in-place (as singularity 2.x used to do), the whole sandbox gets (needlessly?) copied to `/tmp/sbuild-XXXXXXXXXX` via `cp -r`. This changes the owner of all files to root (as well as elongating the build process). 
It would be great if this behavior could be toggled-off (to resemble singularity 2.x).

kadeng:

Replaces "cp -r" commands with "cp -ar" in order to ensure permissions and ownerships are preserved when building images.

 - Fixes #2860

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -5,9 +5,8 @@
 # Maintainers:
 
     - Adam Hughes <adam@sylabs.io>, <stickmanica@gmail.com>
-    - Dave Godlove <d@sylabs.io>, <davidgodlove@gmail.com>
-    - Michael Bauer <m@sylabs.io>, <bauerm@umich.edu>
-    - Mike Gray <mike@sylabs.io>
+    - Cedric Clerget <cedric@sylabs.io>, <cedric.clerget@univ-fcomte.fr>
+    - David Trudgian <david.trudgian@sylabs.io>, <dave@trudgian.net>
 
 # Contributors:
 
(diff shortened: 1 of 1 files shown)


internal/pkg/build/sources/packer_ext3_linux.go
```diff
@@ -65,7 +65,7 @@ func unpackExt3(b *types.Bundle, img *image.Image) error {
 	// copy filesystem into bundle rootfs
 	sylog.Debugf("Copying filesystem from %s to %s in Bundle\n", tmpmnt, b.RootfsPath)
 	var stderr bytes.Buffer
-	cmd := exec.Command("cp", "-r", tmpmnt+`/.`, b.RootfsPath)
+	cmd := exec.Command("cp", "-ar", tmpmnt+`/.`, b.RootfsPath)
```
cclerget: `-a` implies `-r` so `-a` is enough

internal/pkg/build/sources/packer_sandbox.go
```diff
@@ -29,7 +29,7 @@ func (p *SandboxPacker) Pack(context.Context) (*types.Bundle, error) {
 	// copy filesystem into bundle rootfs
 	sylog.Debugf("Copying file system from %s to %s in Bundle\n", rootfs, p.b.RootfsPath)
 	var stderr bytes.Buffer
-	cmd := exec.Command("cp", "-r", rootfs+`/.`, p.b.RootfsPath)
+	cmd := exec.Command("cp", "-ar", rootfs+`/.`, p.b.RootfsPath)
```
cclerget: same

cclerget (commented):
Thanks for this PR. I will just be doing some testing with `--fakeroot` before approving. Also ping @ikaneshiro for review.

cclerget (changes requested):
@kadeng Test ok with `--fakeroot`. Once requested changes are applied we are good to go with. Thanks !

PR #4787, opened 2019-11-25
Cherry pick #4737 to release-3.5

### LINKED ISSUE #4560: docker-daemon:// builder does not clean up all of its tempfiles.

dtrudg:
Cherry pick #4737 to release-3.5 for 3.5.1

Fixes #4560

When building a singularity .sif from a docker archive, singularity would
create a temporary file named docker-tarXXX in /var/tmp, and never clean it up.
This adds a line to calculateRefHash (which was creating the file via
NewImageSource()) to close the image source it created, cleaning up the
temporary file.

### DIFF
--- a/internal/pkg/client/oci/oci.go
+++ b/internal/pkg/client/oci/oci.go
@@ -18,6 +18,7 @@ import (
 	"github.com/containers/image/signature"
 	"github.com/containers/image/transports"
 	"github.com/containers/image/types"
+	"github.com/pkg/errors"
 	"github.com/sylabs/singularity/internal/pkg/client/cache"
 	"github.com/sylabs/singularity/internal/pkg/sylog"
 )
@@ -114,17 +115,22 @@ func ImageSHA(ctx context.Context, uri string, sys *types.SystemContext) (string
 	return calculateRefHash(ctx, ref, sys)
 }
 
-func calculateRefHash(ctx context.Context, ref types.ImageReference, sys *types.SystemContext) (string, error) {
+func calculateRefHash(ctx context.Context, ref types.ImageReference, sys *types.SystemContext) (hash string, err error) {
 	source, err := ref.NewImageSource(ctx, sys)
 	if err != nil {
 		return "", err
 	}
+	defer func() {
+		if closeErr := source.Close(); closeErr != nil {
+			err = errors.Wrapf(err, " (src: %v)", closeErr)
+		}
+	}()
 
 	man, _, err := source.GetManifest(ctx, nil)
 	if err != nil {
 		return "", err
 	}
 
-	hash := fmt.Sprintf("%x", sha256.Sum256(man))
+	hash = fmt.Sprintf("%x", sha256.Sum256(man))
 	return hash, nil
 }


EmmEff (approved):
lgtm

internal/pkg/client/oci/oci.go
```diff
@@ -114,17 +115,22 @@ func ImageSHA(ctx context.Context, uri string, sys *types.SystemContext) (string
...
 }
 
-func calculateRefHash(ctx context.Context, ref types.ImageReference, sys *types.SystemContext) (string, error) {
+func calculateRefHash(ctx context.Context, ref types.ImageReference, sys *types.SystemContext) (hash string, err error) {
 	source, err := ref.NewImageSource(ctx, sys)
 	if err != nil {
 		return "", err
 	}
+	defer func() {
+		if closeErr := source.Close(); closeErr != nil {
```
mem, line 124: I'm going to approve this because it's a cherry pick, but this construction is really strange.

In principle the idea with errors.Wrapf is that you have an error that you wish you annotate with additional information (like what was happening when the error occurred). Here you are annotating the original error err with an additional error. The wrapping should have been done on line ~ 120.

Also note that you have a variable error on the function's signature and on line 119 you have a short variable declaration, except that the only variable being declared is "source", not "err". This makes for hard-to-read code, as quickly glancing at the code might look like the anonymous function on line 123 is a closure on the variable from line 119.

internal/pkg/client/oci/oci.go
```diff
@@ -114,17 +115,22 @@ func ImageSHA(ctx context.Context, uri string, sys *types.SystemContext) (string
...
+			err = errors.Wrapf(err, " (src: %v)", closeErr)
```
mem, line 125: I'm trying to picture this from the point of view of the user. The user will see:

    Something something something (src: something something): something something

The extra space before ` (src: %v)` is something that the upper layers should be dealing with (as part of the leading "Something something something"). The additional ": something something" is the result of calling Error() on this new error (and the "soemthing something" will be the from the error returned by `ref.NewImageSource`.

mem (approved):

PR #4796, opened 2019-11-26
Small Hotfix for Issue 2860 - All files in container owned by root

cclerget:
## Description of the Pull Request (PR):

I closed #4776 by mistake while force-pushed changes. This PR cherry-picked original patch from #4776 by @kadeng (sorry about that)

### This fixes or addresses the following GitHub issues:

- Fixes #2860 
- Related to #4776

#### Before submitting a PR, make sure you have done the following:

- Read the [Guidelines for Contributing](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/sylabs/singularity/blob/master/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- Added tests to validate this PR and tested this PR locally with a `make testall`
- Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](https://github.com/sylabs/singularity/blob/master/CONTRIBUTORS.md)

Attn: @singularity-maintainers

PR #4798, opened 2019-11-26
Cherry pick #4796 into release-3.5

dtrudg:
Cherry picks #4796, which stops ownership being reset to root when building from a sandbox into a SIF.

Cedric confirms this was tested okay with fakeroot, which was the only real concern for the change.


WestleyK (approved):
Looks good to me :+1:

PR #4802, opened 2019-11-27
Use errors=remount-ro mount option with ext3 filesystems only

### LINKED ISSUE #4801: Unable to mount squashfs via `singularity exec` with kernel 5.4.0
### Version of Singularity:
```
$ singularity version
3.5.0
```

### Expected behavior

Success when running:
```
$ singularity exec  build/dev.simg bash
```

### Actual behavior

On Linux kernel 5.3.12, singularity works as intended for some reason. on 5.4.0, I get the following error:
```
$ singularity exec  build/dev.simg bash
FATAL:   container creation failed: mount /proc/self/fd/3->/var/singularity/mnt/session/rootfs error: while mounting image /proc/self/fd/3: kernel reported a bad superblock for squashfs image partition, possible causes are that your kernel doesn't support the compression algorithm or the image is corrupted
```

Furthermore, dmesg logs the following (once per attempt):
```
[ 3878.270562] squashfs: Unknown parameter 'errors'
```

Change kernel to 5.4.0 and reboot with said kernel, then attempt to call 'singularity exec' with a prebuilt image.

Running Gentoo Linux unstable (~amd64), Linux kernel 5.4.0. Desktop environment is Plasma.

Built from source using ebuild from https://bugs.gentoo.org/694048 modified to use version 3.5.0 instead of 3.4.0 (literally renaming the ebuild from `singularity-3.4.0.ebuild` to `singularity-3.5.0.ebuild`). After installing the ebuild in my local portage repository, all I had to do was
```
# emerge singularity
```
and wait until compilation and installation completed.

### Additional information/comments

I will stress that the exact same binary works with kernel version 5.3.12, but not with 5.4.0 (using GRUB to switch between kernels on reboot for testing). After diving through the kernel source, I'm not sure why 5.3.12 even works, honestly, but the reason it doesn't work on 5.4.0 is that in `fs/fs_context.c` in the kernel tree squashfs does not provide an argument parser, so the `errors` argument cannot be parsed and the function `vfs_parse_fs_param` eventually returns an EINVAL as a result, along with the error message being reported in the dmesg.

Arguably both the kernel and singularity are wrong here (I think)-- the kernel is wrong because this seems like a userspace regression, and singularity is wrong by passing an argument for mounting that is not supported or understood by squashfs.

I will be away on travel for about a week, but let me know if you need anything else from me to debug this.

cclerget:

Since 5.4.0 kernel passing `errors=remount-ro` with `squashfs` mount options returns an error, fix that by conditionally set this option for `ext3` filesystems only.

 - Fixes #4801 

### DIFF
--- a/internal/pkg/runtime/engine/imgbuild/create_linux.go
+++ b/internal/pkg/runtime/engine/imgbuild/create_linux.go
@@ -96,7 +96,7 @@ func (e *EngineOperations) CreateContainer(ctx context.Context, pid int, rpcConn
 	}
 
 	sylog.Debugf("Mounting image directory %s\n", rootfs)
-	if err := rpcOps.Mount(rootfs, sessionRootFs, "", syscall.MS_BIND, "errors=remount-ro"); err != nil {
+	if err := rpcOps.Mount(rootfs, sessionRootFs, "", syscall.MS_BIND, ""); err != nil {
 		return fmt.Errorf("failed to mount directory filesystem %s: %s", rootfs, err)
 	}
 
--- a/internal/pkg/util/fs/mount/mount_linux.go
+++ b/internal/pkg/util/fs/mount/mount_linux.go
@@ -610,7 +610,10 @@ func (p *Points) AddImage(tag AuthorizedTag, source string, dest string, fstype 
 		return fmt.Errorf("invalid image size, zero length")
 	}
 	keyB64 := base64.StdEncoding.EncodeToString(key)
-	options = fmt.Sprintf("loop,offset=%d,sizelimit=%d,key=%s,errors=remount-ro", offset, sizelimit, keyB64)
+	options = fmt.Sprintf("loop,offset=%d,sizelimit=%d,key=%s", offset, sizelimit, keyB64)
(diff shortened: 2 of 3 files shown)


jmstover (changes requested):
There's a `syscall.Mount` in `pkg/ocibundle/sif/bundle_linux.go` ... Does that need modified as well?

**Edit:**
And in `internal/pkg/runtime/engine/imgbuild/create_linux.go` there's a `rpcOpt.Mount` that's passing `errors=remount-ro`  ... not sure if it's relevant

cclerget:
> There's a `syscall.Mount` in `pkg/ocibundle/sif/bundle_linux.go` ... Does that need modified as well?
> 
> **Edit:**
> And in `internal/pkg/runtime/engine/imgbuild/create_linux.go` there's a `rpcOpt.Mount` that's passing `errors=remount-ro` ... not sure if it's relevant

Good catch both fixed. Thanks !

jmstover (approved):

jmstover:
Does this need to merge back into `master` as well?

PR #4818, opened 2019-12-02
Basic support for providing info on a image with search

### LINKED ISSUE #2949: `singularity search` should provide info about a specific container
## Version of Singularity:

3.0.1

Lets say I found a container with `singularity search lolcow`.
<pre>
        library://ilumb/default/lolcow
                Tags: latest
</pre>

It would be good if I did `singularity search library://ilumb/default/lolcow:latest` to get any information on that container, including what the definition file used was, how to `pull` it, when it was updated, if it has vulnerabilities, etc.

More data is good.

WestleyK:

This PR adds basic support for providing info on a specific image

There is still some TODOs before this can be merged, but it is ready for initial review/testing.

 - Fixes #2949

### DIFF
--- a/internal/pkg/library/library.go
+++ b/internal/pkg/library/library.go
@@ -7,12 +7,16 @@ package library
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
+	"net/http"
 	"os"
 	"strings"
 
 	"github.com/sylabs/scs-library-client/client"
+	"github.com/sylabs/singularity/internal/pkg/sylog"
 )
 
 const defaultTag = "latest"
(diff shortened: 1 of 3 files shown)


dtrudg (changes requested):
Hi Westley - I think that things which deal with the library API endpoints directly like this should be in https://github.com/sylabs/scs-library-client/ and not directly in the Singularity code.

Take a look at the functions already available there. e.g. there is already a:

`func (c *Client) getTags(ctx context.Context, containerID string) (TagMap, error) {`

and a...

`func (c *Client) getContainer(ctx context.Context, containerRef string) (*Container, error) {`

These are private - but something higher-level can be added to fetch the info you need - or maybe the existing search API needs re-thinking?

It would be best to spec out exactly what you want to show, and then look at the `scs-library-client` to see the best way to retrieve it efficiently.

WestleyK:
@dctrud, I got this working by only exporting `GetContainer` from scs-library-client. PR => https://github.com/sylabs/scs-library-client/pull/67, which will probably need to be merged first.

internal/pkg/library/library.go
```diff
@@ -70,12 +71,43 @@ func DownloadImageNoProgress(ctx context.Context, c *client.Client, imagePath, a
...
+	if ref := strings.Split(value, "/"); len(ref) > 2 {
+		searchImage := value
+		if strings.HasPrefix(searchImage, "library://") {
+			searchImage = strings.TrimPrefix(searchImage, "library://")
+		}
+		sylog.Debugf("Attempting to get image info for: %s", searchImage)
+		cont, err := c.GetContainer(ctx, searchImage)
+		if err == nil {
+			fmt.Printf("Image:           %s\n", "library://"+searchImage)
+			fmt.Printf("Tags:            %s\n", getImageTags(cont.ImageTags))
```
dtrudg, line 100: ImageTags isn't really what we want to use here... they may not be the correct architecture. There is a two-level map of ArchTags where you can pull out by architecture->tag

internal/pkg/library/library.go
```diff
@@ -70,12 +71,43 @@ func DownloadImageNoProgress(ctx context.Context, c *client.Client, imagePath, a
...
+
 // SearchLibrary searches the library and outputs results to stdout
 func SearchLibrary(ctx context.Context, c *client.Client, value string) error {
 	if len(value) < 3 {
 		return fmt.Errorf("bad query '%s'. You must search for at least 3 characters", value)
 	}
 
+	// If the user is searched for a container uri (eg. library/default/alpine) then
+	// try to get the image infomation, if unsuccessful, then search as usual.
+	if ref := strings.Split(value, "/"); len(ref) > 2 {
```
dtrudg, line 91: An issue here is that the second 'collection' level is pseudo-deprecated, so looking for >2 doesn't fit with the shortening that infers `default` as the collection, and optionally `library` as the default entity.

E.g. `library://alpine` is a container, as is `library://library/alpine`

internal/pkg/library/library.go
```diff
@@ -70,12 +71,43 @@ func DownloadImageNoProgress(ctx context.Context, c *client.Client, imagePath, a
...
+}
+
 // SearchLibrary searches the library and outputs results to stdout
 func SearchLibrary(ctx context.Context, c *client.Client, value string) error {
 	if len(value) < 3 {
 		return fmt.Errorf("bad query '%s'. You must search for at least 3 characters", value)
 	}
 
+	// If the user is searched for a container uri (eg. library/default/alpine) then
+	// try to get the image infomation, if unsuccessful, then search as usual.
```
dtrudg, line 90: The terminology is a bit mixed here. You are actually getting information about a container - including all the tagged images for that container.

internal/pkg/library/library.go
```diff
@@ -70,12 +71,43 @@ func DownloadImageNoProgress(ctx context.Context, c *client.Client, imagePath, a
...
+		if err == nil {
+			fmt.Printf("Image:           %s\n", "library://"+searchImage)
+			fmt.Printf("Tags:            %s\n", getImageTags(cont.ImageTags))
+			fmt.Printf("Description:     %s\n", cont.Description)
+			// TODO: print if the image is signed or not
+			fmt.Printf("Total downloads: %d\n", cont.DownloadCount)
+			fmt.Printf("Stars:           %d\n", cont.Stars)
+			return nil
+		} else {
+			sylog.Verbosef("Failed to search container info: %s", err)
```
dtrudg, line 107: I'm not sure we need to do this if we are going to fall back to the general search - will that always give output?

dtrudg (changes requested):
Hi @WestleyK - I've left a few comments here. I'm not 100% sure how to tackle them yet though - particularly the issue about the 1 and 2 level URIs for containers.

More generally, though... this seems a bit like an overloading of `search` as a term, if we are only going to display information about exactly one container. Maybe we think about another verb, or at least the help text needs modifying to reflect the double behaviour.

dtrudg:
Hi @WestleyK - to move this along I think it might be good to consider...

1) Using a different verb like `describe` for detail on a single container, so that we don't overload search, and don't hit the issue where the length of a library reference (number of `/`) doesn't unambiguously determine if it refers to a container or not.

2) Use a flag to `search` if we want to show more detail for each search result. The detail is already provided to us in the searchResults structure.

Maybe a good thing to do here, at this point, would be to just write out in text form how you see `search` and/or `describe` being invoked, and what format output they give. We can then ask for input from @EmmEff and @tri-adam if needed, as arbiters of consistency across the client / cloud services.

I believe your linked https://github.com/sylabs/scs-library-client/pull/67 is going to be necessary for a `describe`, but lets define exactly what the end goal looks like first.

dtrudg:
Closing as this has gone stale, and would be better implemented coupled with changes in the sylabs/scs-library-client

PR #4817, opened 2019-12-02
Fix sandbox overlay incompatibility fallback to underlay

cclerget:
## Description of the Pull Request (PR):

The underlay fallback code didn't registered the image in stage one resulting in `no root filesystem found` later in container creation. This PR fix it and will display warning messages if the image is used with `--writable-tmpfs` and/or `--overlay` options.

### This fixes or addresses the following GitHub issues:

 - Fixes #4811 

#### Before submitting a PR, make sure you have done the following:

- Read the [Guidelines for Contributing](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/sylabs/singularity/blob/master/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- Added tests to validate this PR and tested this PR locally with a `make testall`
- Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](https://github.com/sylabs/singularity/blob/master/CONTRIBUTORS.md)

Attn: @singularity-maintainers

PR #4816, opened 2019-12-02
Cast Stat_t.Dev to uint64 for MIPS oddities

cclerget:
### This fixes or addresses the following GitHub issues:

 - Fixes #4814 

#### Before submitting a PR, make sure you have done the following:

- Read the [Guidelines for Contributing](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/sylabs/singularity/blob/master/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- Added tests to validate this PR and tested this PR locally with a `make testall`
- Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](https://github.com/sylabs/singularity/blob/master/CONTRIBUTORS.md)

Attn: @singularity-maintainers

PR #4810, opened 2019-11-29
Replace NO_SETUID_FIXUP with KEEP_CAPS

cclerget:
## Description of the Pull Request (PR):

`NO_SETUID_FIXUP` is propagated to container processes leading them to retain their capabilities when they switch to a non-zero UID/GID, it causes issues with some services like openssh server (see #4319)

### This fixes or addresses the following GitHub issues:

 - Fixes #4319 

#### Before submitting a PR, make sure you have done the following:

- Read the [Guidelines for Contributing](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/sylabs/singularity/blob/master/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- Added tests to validate this PR and tested this PR locally with a `make testall`
- Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](https://github.com/sylabs/singularity/blob/master/CONTRIBUTORS.md)

Attn: @singularity-maintainers

PR #4822, opened 2019-12-03
Merge from 3.5.1-rc.2 tag into master

dtrudg:
Fold in the #4810 KEEP_CAPS, #4817 lustre/underlay fix and #4816 MIPS build fixes that were against release-3.5 into master.


cclerget (approved):
LGTM

PR #4832, opened 2019-12-05
Add a travis branch safelist

mem:
Since travis-ci is configured to build both on pushes and pull requests,
pull requests created using branches on the repository (as done by
dependabot) will trigger two travis-ci runs: one for the pull request
itself and one for the push to the branch.

With this change that should be fixed and only one run will start (for
the pull requests). After merging the pull request to the target branch
this triggers another run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,9 @@
+branches:
+  only:
+    - master
+    - /^release-\d+\..+$/
+    - /^v\d+\..+$/
+
 language: go
 
 sudo: required


.travis.yml
```diff
@@ -1,3 +1,9 @@
+branches:
+  only:
+    - master
+    - /^release-\d+\..+$/
+    - /^v\d+\..+$/
```
jscook2345, line 5: What branch is signified with `^v...` here? The branches I see in our repo only have `master` or `release-*`.

dtrudg: I believe this is to match tags - Travis will fire on a new tag.

dtrudg (approved):
LGTM

.travis.yml
mem: yaml and comments :disappointed: 

According to the documentation, if I add a safelist (branch -> only) and I tag a commit on one of those branches, the process won't trigger (because it will only look at the tag and not the branch name?) so it's necessary to a pattern for the tags that would be applied to the corresponding branches. This obviously has the side effect of triggering the process for tags on any branch...

PR #4839, opened 2019-12-06
Fix fakeroot sandbox build chown error

### LINKED ISSUE #4829: Need to fallback to sandbox TMPDIR + copy for non-fakeroot supporting filesystems
### Version of Singularity:

```
3.5.0
```

### Actual behavior

In 3.5.0 it is not possible to directly build a sandbox to GPFS, and _most_ sandboxes to lustre

On filesystems with no subuid mapping support (GPFS), or that do not seem to support it completely (e.g. lustre works for root->user, but not other subgids) we need to revert to the old behaviour of <=3.4, build to `TMPDIR` and the copy over. This means the sandbox at the end of the build goes onto GPFS or lustre with the user's own UID/GID.

Note that may conflict in behavior with the change to `cp -a` in https://github.com/sylabs/singularity/pull/4796 in the different but related sandbox -> SIF flow. We will lose some amount of ownership / permission information.

cclerget:

Fakeroot may fail to set file/directory ownership when building a sandbox on some network filesystems which are not aware of user namespace, to address the issue the build move to a temporary rootfs inside $TMPDIR and copy instead of moving the final rootfs sandbox to the final destination.

 - Fixes #4829

### DIFF
--- a/internal/pkg/build/assemblers/sandbox.go
+++ b/internal/pkg/build/assemblers/sandbox.go
@@ -6,28 +6,43 @@
 package assemblers
 
 import (
+	"bytes"
 	"fmt"
 	"os"
+	"os/exec"
 
 	"github.com/sylabs/singularity/internal/pkg/sylog"
 	"github.com/sylabs/singularity/pkg/build/types"
 )
 
 // SandboxAssembler assembles a sandbox image.
-type SandboxAssembler struct{}
+type SandboxAssembler struct {
+	Copy bool
+}
 
 // Assemble creates a Sandbox image from a Bundle.
 func (a *SandboxAssembler) Assemble(b *types.Bundle, path string) (err error) {
 	sylog.Infof("Creating sandbox directory...")
 
-	sylog.Debugf("Moving sandbox from %v to %v", b.RootfsPath, path)
 	if _, err := os.Stat(path); err == nil {
 		os.RemoveAll(path)
 	}
 
-	err = os.Rename(b.RootfsPath, path)
-	if err != nil {
-		return fmt.Errorf("sandbox assemble failed: %v", err)
+	if a.Copy {
+		sylog.Debugf("Copying sandbox from %v to %v", b.RootfsPath, path)
+		var stderr bytes.Buffer
+		cmd := exec.Command("cp", "-r", b.RootfsPath+`/.`, path)
+		cmd.Stderr = &stderr
+		if err := cmd.Run(); err != nil {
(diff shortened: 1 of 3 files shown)


dtrudg (changes requested):
This seems to work for me to fix the reported issue. However, I think things could still fail in an ugly way *If* someone sets their `TMPDIR` to Lustre (I envisage people doing this when building large images on nodes with small local /tmp, e.g. tmpfs)

Can we...
- Factor out the chmod check into a function
- Call it against the intended rootfs dir... if fails fall back to TMPDIR rootfs...
- Check it against the TMPDIR rootfs... if fails raise an error that the TMPDIR must be set to a filesystem that supports chown inside user namespace?

dtrudg (approved):
Thanks @cclerget - this fixes / provides helpful info for various situations.

```
# build to lustre sandbox
[vagrant@master lustre]$ singularity build -f -s mysandbox docker://alpine
WARNING: The underlying filesystem on which resides "/lustre/mysandbox" won't allow to set ownership, as a consequence the sandbox could not preserve image's files/directories ownerships
INFO:    Starting build...
Copying blob 89d9c30c1d48 skipped: already exists
Copying config 759e71f0d3 done
Storing signatures
2019/12/09 14:27:27  info unpack layer: sha256:89d9c30c1d48bac627e5c6cb0d1ed1eec28e7dbdfbcc04712e4c79c0f83faf17
WARNING: Permission handling has changed in Singularity 3.5 for improved OCI compatibility
WARNING: The sandbox will contain files/dirs that cannot be removed until permissions are modified
WARNING: Use 'chmod -R u+rwX' to set permissions that allow removal
WARNING: Use the '--fix-perms' option to 'singularity build' to modify permissions at build time
WARNING: You can provide feedback about this change at https://github.com/sylabs/singularity/issues/4671
INFO:    Build complete: mysandbox


# build to lustre sandbox, TMPDIR is lustre
[vagrant@master lustre]$ singularity build -f -s mysandbox_lustretmp docker://alpine
ERROR:   Could not set files/directories ownership, if /lustre/tmp/rootfs-3f17bb6d-1a90-11ea-b6b4-5254008afee6 is on a network filesystem, you must set TMPDIR to a local path (eg: TMPDIR=/var/tmp singularity build ...)
FATAL:   Unable to create build: ownership change not allowed in /lustre/tmp, aborting

# noexec
[vagrant@master lustre]$ singularity build -s mysandbox_noexec docker://alpine
FATAL:   Unable to create build: 'noexec' mount option set on /lustre, sandbox /lustre/mysandbox_noexec won't be usable at this location
```

PR #4847, opened 2019-12-09
Cherry pick #4839 to release-3.5 for 3.5.2

dtrudg:
Cherry pick #4839 to release-3.5 for 3.5.2

Fakeroot may fail to set file/directory ownership when
building a sandbox on some network filesystems which are
not aware of user namespace, to address the issue the build
move to a temporary rootfs inside $TMPDIR and copy instead of
moving the final rootfs sandbox to the final destination.


cclerget (approved):
LGTM

PR #4850, opened 2019-12-09
Use a sensible hash for http(s) source caching

### LINKED ISSUE #4823: Net cache leads to running incorrect container (not using a good hash)
### Version of Singularity:

```
Reported on 3.3.0 - verified on all above
```

### Actual behavior

If you `singularity shell http://example.com/container1/v1.sif` it will be cached in `net/hash/v1.sif`.

If you then `singularity shell http://example.com/container2/v1.sif` it will run the `container1/v1.sif` cached image, not the `container2/v1.sif` image that you expect.

The net cache introduced in #2346 is basing the cache filename on the final part of the URI only.

To avoid clashes, the obvious thing that comes to mind is taking a sha256 over:

`Full URL` + `HTTP date response header`

Then the code doing the GET can compute that to check if there's something already in the cache without downloading the whole file. Some endpoints (like object stores) might have a way of getting a real content hash, or always return one in headers - but we can't assume this is the case for hosting on a standard http server... or considering all the object store protocols out there. The HTTP Date header is the closest thing I can come up with to a proxy for 'is the the same thing at this URL'.

If there's no date header available then always download again.

dtrudg:

The existing caching for `http` and `https` sources uses the final part of the URL as the key for the image within the cache. This means that `http://example.com/container1/v1.sif` and `http://example.com/container2/v1.sif` result in the same cache key, and unexpected behaviour as in #4823.

We will cache using a sha256 over the URL and the date of the file that is to be
fetched, as returned by an HTTP HEAD call. If no date is available, use the current
date-time, which will effectively result in no caching.

 - Fixes: #4823

### DIFF
--- a/cmd/internal/cli/actions.go
+++ b/cmd/internal/cli/actions.go
@@ -7,11 +7,15 @@ package cli
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"runtime"
 	"strings"
+	"time"
 
 	ocitypes "github.com/containers/image/types"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


cmd/internal/cli/actions.go
```diff
@@ -290,11 +294,35 @@ func handleShub(imgCache *cache.Handle, u string) (string, error) {
...
-	imageName := refParts[len(refParts)-1]
-	imagePath := imgCache.NetImage("hash", imageName)
+	// We will cache using a sha256 over the URL and the date of the file that
+	// is to be fetched, as returned by an HTTP HEAD call and the Last-Modified
+	// header. If no date is available, use the current date-time, which will
+	// effectively result in no caching.
+	imageDate := time.Now().String()
 
-	exists, err := imgCache.NetImageExists("hash", imageName)
+	req, err := http.NewRequest("HEAD", u, nil)
```
cclerget, line 303: Maybe trim `http://`, `https://` from the hash as some server allow access via http and/or redirect http to https

dtrudg: That's true - but equally a crazy setup could serve different things on http and http(s). That would be really stupid and unlikely, but I'd rather err on the side of re-downloading an image unless you feel strongly about this?

cclerget: Not really it was just a suggestion

e2e/actions/regressions.go
```diff
@@ -213,3 +218,100 @@ func (c actionTests) issue4836(t *testing.T) {
...
+		{
+			name:         "http_c2v1.sif_touched_uncached",
+			urlPath:      "/c1/v1.sif",
+			touchFile:    true,
+			expectCached: false,
+		},
+		// Different URL - not cached
+		{
+			name:         "http_c2v2.sif_uncached",
+			urlPath:      "/c1/v2.sif",
```
ikaneshiro: Should this url be `/c2/v1.sif` instead? If the cache was keyed with just the last part of the url, as it was previously, then wouldn't cause a collision if I understand the issue correctly

dtrudg:
CI failure seems to be a network issue - rerunning:
```
Error initializing source oci:/tmp/stest.689162124/e2e-imgcache-388157803/cache/oci:c19173c5ada610a5989151111163d28a67368362762534d8a8121ce95cf2bd5a: Error writing blob: read tcp 10.142.0.47:59836->104.18.122.25:443: read: connection reset by peer
```

ikaneshiro (commented):
Caching approach lgtm

e2e/actions/regressions.go
dtrudg: Yes - you are correct. Good spot, and :man_facepalming: from me seeing as this is what the test is meant to be for

ikaneshiro (approved):

PR #4852, opened 2019-12-09
Cherry pick #4850 to release-3.5 for 3.5.2

dtrudg:
Cherry pick #4850 to release-3.5 for 3.5.2

Use a sensible hash for http(s) source caching

Fixes: #4823

We will cache using a sha256 over the URL and the date of the file that is to be
fetched, as returned by an HTTP HEAD call. If no date is available, use the current
date-time, which will effectively result in no caching.

### DIFF
--- a/cmd/internal/cli/actions.go
+++ b/cmd/internal/cli/actions.go
@@ -7,11 +7,15 @@ package cli
 
 import (
 	"context"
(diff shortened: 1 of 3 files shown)


cclerget (approved):
LGTM

PR #4875, opened 2019-12-18
Replace +dirty by -dirty to be compatible with semantic version

cclerget:

Replace `+dirty` by `-dirty` to be compatible with semantic version. @mem do you see any side effects with this change ?

### DIFF
--- a/scripts/get-version
+++ b/scripts/get-version
@@ -38,7 +38,7 @@ if test -n "${head}" ; then
 	# cannot pass --dirty when using commit-ish
 	dirty_info=${current_commit}
 else
-	dirty_info=$(gitdesc --dirty=+dirty)
+	dirty_info=$(gitdesc --dirty)
 fi
 
 # First figure out if the current commit matches a tag:


dtrudg (commented):
@cclerget - can you give an example of when it causes a problem to have `+dirty`? `+dirty` should be valid from my reading of https://semver.org/

cclerget:
@dctrud Typically while developping and when you didn't commit change, when you run e2e tests and hit the version test, if singularity has a version like `3.5.2+152-g75f4433c0+dirty`, the test fail as two consecutives `+` is not valid semantic version and means we have two build part, so changing it to `3.5.2+152-g75f4433c0-dirty` just make it valid with `152-g75f4433c0-dirty` as the build part

dtrudg:
> @dctrud Typically while developping and when you didn't commit change, when you run e2e tests and hit the version test, if singularity has a version like `3.5.2+152-g75f4433c0+dirty`, the test fail as two consecutives `+` is not valid semantic version and means we have two build part, so changing it to `3.5.2+152-g75f4433c0-dirty` just make it valid with `152-g75f4433c0-dirty` as the build part

This looks good to me for that situation. However, if we have checkout a tag, e.g. `v3.5.2` and then make it dirty would we have then `3.5.2-dirty` instead of `3.5.2+dirty`?

cclerget:
> This looks good to me for that situation. However, if we have checkout a tag, e.g. v3.5.2 and then make it dirty would we have then 3.5.2-dirty instead of 3.5.2+dirty?

Yes ... By adding a check to conditionally set the dirty part to either `-dirty` or `+dirty` depending of a previous `+` presence that should generate a usable version in all cases right ?

dtrudg:
> > This looks good to me for that situation. However, if we have checkout a tag, e.g. v3.5.2 and then make it dirty would we have then 3.5.2-dirty instead of 3.5.2+dirty?
> 
> Yes ... By adding a check to conditionally set the dirty part to either `-dirty` or `+dirty` depending of a previous `+` presence that should generate a usable version in all cases right ?

Yes - that sounds good to me :+1:

dtrudg (approved):
This seems to work fine for the situations I can see. Off a tag we have e.g. `3.5.2-dirty` and not `3.5.2+dirty` but I don't see any issue with that.

PR #4898, opened 2019-12-22
Fix network_linux_test where bridge not available / setup fails on RHEL6

### LINKED ISSUE #4897: network_linux_test.go should not try impossible network configs on RHEL6
### Version of Singularity:

```
$ 3.5.2
```

### Actual behavior

On RHEL6, tests in `network_linux_test.go` will fail as we cannot actually create the CNI bridge network configurations attempted there. We need some kind of alternative tests, or to avoid these tests.

In addition, as a side-effect of `https://github.com/sylabs/singularity/pull/4895` which now causes `nc` on RHEL6.10 to start properly, things can get stuck. We need to ensure that processes spawned are explicitly killed if the network configuration fails. At present we are doing a `wait` on the `nc` process in `TestAddDelNetworks/TestHTTPPort` that will never complete, because if the network configuration fails, then the function that connects/closes and causes `nc` to exit is never run.

```
RHEL6

```

dtrudg:

- Kill helper process in network_linux_test on error

If the network setup fails with an error, our helper process such as nc,
may be stuck. E.g. in the tests using nc we are expecting it to end once
a connection is made and closed. If the network setup failed we need to
kill it explicitly to avoid a freeze.

- Protect tests needing CNI bridge on RHEL 6 as well as CentOS 6

 - Fixes #4897 

### DIFF
--- a/pkg/network/network_linux_test.go
+++ b/pkg/network/network_linux_test.go
@@ -597,6 +597,9 @@ func TestAddDelNetworks(t *testing.T) {
 		nsPath := fmt.Sprintf("/proc/%d/ns/net", cmd.Process.Pid)
 		if err := c.runFunc(nsPath, cniPath, stdinPipe, stdoutPipe); err != nil {
 			t.Errorf("unexpected failure for %q: %s", c.name, err)
+			if err := cmd.Process.Kill(); err != nil {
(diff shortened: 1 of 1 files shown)


jscook2345 (approved):
You might want to have the skip note why it is skipping so the test output is easier to understand.

PR #4911, opened 2019-12-23
Plugin rework

cclerget:

- [x] Compile plugin outside of Singularity tree
- [x] Enforce check during plugin compilation (match Go runtime requirement)
- [x] Add more hook points in CLI/Runtime
- [x] Implement an autoload mechanism
- [x] Add generic install callback (eg: to install configuration files)
- [x] Add `singularity plugin create` to create a skeleton plugin directory
- [ ] Install Singularity source code assets to compile plugin without keeping source repo around and use source matching the currently installed version
- [x] Add e2e plugin tests

 - Fixes #


internal/pkg/plugin/create.go
```diff
@@ -0,0 +1,110 @@
...
+	if err := ioutil.WriteFile(filename, []byte(gitIgnore), 0644); err != nil {
+		return fmt.Errorf("while creating plugin %s: %s", filename, err)
+	}
+
+	// create symlink to singularity source directory
+	source := filepath.Join(dir, SingularitySource)
+
+	if _, err := os.Stat(buildcfg.SOURCEDIR); os.IsNotExist(err) {
+		ls := fmt.Sprintf("ln -s /path/to/singularity/source %s", source)
+		sylog.Warningf("Singularity source %s doesn't exist, you would have to execute manually %q", buildcfg.SOURCEDIR, ls)
```
ikaneshiro, line 100: Should we return here so we don't try to create the symlink when sourcedir does not exist?

pkg/plugin/callback/runtime/engine/singularity/singularity.go
```diff
@@ -0,0 +1,28 @@
...
+// uses this callback the runtime aborts its execution.
+// This callback is called in:
+// - internal/pkg/runtime/engine/singularity/monitor_linux.go
+type MonitorContainer func(config *config.Common, pid int, signals chan os.Signal) (syscall.WaitStatus, error)
+
+// PostStartProcess callback is called after the container process
+// started. It's a good place to add custom logger and/or notifier.
+// This callback is called in:
+// - internal/pkg/runtime/engine/singularity/process_linux.go
+type PostStartProcess func(config *config.Common, pid int) error
```
ikaneshiro, line 28: Should there be a PreStartProcess hook for things like the the fusemount plugins? Not something that needs to be added before this is merged if that should even be added at all.

ikaneshiro:
I think the move to not expose the `/internal` packages to plugins is a good change given singularity development should be free to change those APIs without fear of breaking compatibility of externally managed software

internal/pkg/plugin/create.go
cclerget: Yep thanks !

pkg/plugin/callback/runtime/engine/singularity/singularity.go
cclerget: Actually I don't know what is the best approach to load plugin on host filesystem once we are in the container chroot where library and plugins are not (easily) accessible

ikaneshiro: Given #4953 we dont need to address this atm and it may not make sense to have a hook there for most plugin use cases

ikaneshiro (approved):

internal/pkg/plugin/module.go
```diff
@@ -0,0 +1,202 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+// PrepareGoModules returns a byte array containing a generated go.mod matching
+// Singularity modules in use in order to compile/load the plugin with same version
+// of dependencies.
+func PrepareGoModules(pluginDir string) ([]byte, error) {
```
tri-adam: I wonder if this should consider semantic versioning, and error out under certain circumstances. Just to run through a few scenarios, and maybe we can discuss what the _expected_ behaviour would be:

1. Singularity and the plugin both require the **same major version, but Singularity requires an older minor version**. In this case, downgrading the package could cause issues:
 a. Say the plugin developer has chosen to use a newer version because it contains a critical security fix. The version is downgraded, the plugin builds, and now has a vulnerability.
 b. Say the plugin developer has chosen to use a newer version because it contains a new API. The `require` is modified, the plugin fails to build.
1. Singularity and the plugin both require the **same major version, but Singularity requires a newer minor version**. This substitution should be fine, assuming the package sticks to semantic versioning.
1. Singularity and the plugin require **different major versions** of the package. In this case, `plugin compile` should probably outright fail with a clear error about incompatible package versions being required?

tri-adam (commented):
Looks OK to me. One question from a plugin developer point of view, @cclerget curious on your thoughts there.

internal/pkg/plugin/module.go
cclerget: - 1a, I would say that Singularity still remains vulnerable by using the old version, this is a risk by growing the attack surface with additional code from plugins but Singularity is also a big attack surface in case of vulnerable package.

- 1b, yes there is more chance that build fail, maybe adding some warning about a non backwards compatibility in this case.

- 2, should be OK assuming package respect backwards compatibility.

- 3, in this case yes it may be safer to fail with explicit error if the package didn't use import semantic version path.

Otherwise there is the heavy approach to reduce failure by rewriting dynamically in plugin code the import paths with semantic version path to take the expected versions of package (excluding standard packages of course) and "unshare" them from Singularity, thoughts ?

tri-adam: I'm tempted to say that we should explicitly require plugins to use the same major version, and <= minor version (case 2 above.) Otherwise, fail with a clear error message stating the package that is incompatible.

The alternative might end up with "my plugin builds with Singularity version X+1 but not X, something is wrong with Singularity." Or worse, "`plugin compile` re-introduced a vulnerability that I explicitly fixed!". If it's an explicit error, the plugin developer is forced to make the decision of whether to support that version of Singularity or not.

cclerget: Fix applied. Also I'm wondering if we should provide an extra flag to at least disable enforcement on plugin/singularity package minor version during `plugin compile` ? as there is many packages versioned like `v0.0.0-20XXXXX...` which may only differ from few commits with just bug fixes

tri-adam: Sorry for the slow reply... I agree, sounds like a good idea.

tri-adam (approved):

dtrudg (approved):
Gave this a read through too, and it looks sensible to me also.

Many thanks to @tri-adam for thinking around the versioning etc.

PR #4926, opened 2020-01-03
Bind /etc/hosts or create a default one with --contain flag

### LINKED ISSUE #1707: singularity --contain without --net stops sharing /etc/hosts
## Version of Singularity:

2.5.1

### Expected behavior

Without using --net, I expect /etc/hosts to be mapped into the container because the network is shared with the host.

### Actual behavior

Since /etc/hosts is only listed as a bind path in singularity.conf, and those are ignored with --contain (#1212), then /etc/hosts is not bind mounted into the container when --contain is used.

Start singularity with --contain and look at /etc/hosts.

A workaround is to set the environment variable SINGULARITY_BINDPATH=/etc/hosts, but that shouldn't be necessary.

cclerget:

Bind `/etc/hosts` with `--contain` flag or create a minimal staging hosts file if a network namespace is requested with `--contain`.

 - Fixes #1707

### DIFF
--- a/internal/pkg/runtime/engine/singularity/container_linux.go
+++ b/internal/pkg/runtime/engine/singularity/container_linux.go
@@ -1248,7 +1248,40 @@ func (c *container) addBindsMount(system *mount.System) error {
 	flags := uintptr(syscall.MS_BIND | c.suidFlag | syscall.MS_NODEV | syscall.MS_REC)
 
 	if c.engine.EngineConfig.GetContain() {
-		sylog.Debugf("Skipping bind mounts as contain was requested")
+		const (
+			hostsPath     = "/etc/hosts"
+			localtimePath = "/etc/localtime"
+		)
+		hosts := hostsPath
+
+		// handle special case for /etc/hosts as it is required,
+		// if no network namespace was requested we simply bind
+		// /etc/hosts from host, if network namespace is requested
+		// we create a minimal default hosts for localhost resolution
+		if !c.netNS {
+			sylog.Debugf("Binding /etc/hosts and /etc/localtime only with contain")
+		} else {
+			sylog.Debugf("Skipping bind mounts as contain was requested")
+
+			sylog.Verbosef("Binding staging /etc/hosts as contain is set")
+			if err := c.session.AddFile(hostsPath, files.DefaultHosts()); err != nil {
+				return fmt.Errorf("while adding /etc/hosts staging file: %s", err)
+			}
+			hosts, _ = c.session.GetPath(hostsPath)
+		}
+
+		if err := system.Points.AddBind(mount.BindsTag, hosts, hostsPath, flags); err != nil {
+			return fmt.Errorf("unable to add %s to mount list: %s", hosts, err)
+		}
+		if err := system.Points.AddRemount(mount.BindsTag, hostsPath, flags); err != nil {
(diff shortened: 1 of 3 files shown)


internal/pkg/runtime/engine/singularity/container_linux.go
```diff
@@ -1248,8 +1248,27 @@ func (c *container) addBindsMount(system *mount.System) error {
...
+			sylog.Debugf("Skipping bind mounts as contain was requested")
+
+			sylog.Verbosef("Binding staging /etc/hosts as contain is set")
+			if err := c.session.AddFile(hostsPath, files.DefaultHosts()); err != nil {
+				return fmt.Errorf("while adding /etc/hosts staging file: %s", err)
+			}
+			hosts, _ = c.session.GetPath(hostsPath)
+		}
+
+		system.Points.AddBind(mount.BindsTag, hosts, hostsPath, flags)
```
dtrudg: In the loop below, you catch the error on `AddBind` is that not necessary here?

internal/pkg/runtime/engine/singularity/container_linux.go
```diff
@@ -1267,6 +1286,7 @@ func (c *container) addBindsMount(system *mount.System) error {
 		if err != nil {
 			return fmt.Errorf("unable to add %s to mount list: %s", src, err)
 		}
+		system.Points.AddRemount(mount.BindsTag, dst, flags)
```
dtrudg: Is the addition here fixing a separate issue?

Also, in the contain case above you are returning any error from `AddRemount`, but here it is not caught / returned. Does it need to be?

dtrudg (commented):
Couple of questions.

internal/pkg/runtime/engine/singularity/container_linux.go
cclerget: Yes this is a different issue but while we are in this code part it's worth to fix it. Fixed the return.

internal/pkg/runtime/engine/singularity/container_linux.go
cclerget: Fixed

DrDaveD:
How about also including /etc/localtime and /etc/nsswitch.conf as I suggested in #1707?  Unless of course someone had already bind mounted them another way.

dtrudg:
Not sure about `nsswitch.conf`. Wouldn't that be quite likely to have `ldap` directory or `sss` caching stuff configured in it, which is not what we want in a `--contain`ed container which is meant to be as independent from the host as we can get?

cclerget:
`/etc/localtime` has been added with `--contain` and `singularity.conf` note updated to mention this behavior. @DrDaveD I don't think it make sense to add `nsswitch.conf` for many reasons :
- it should be specific to container images and nss services present in container image
- with this PR if we bind it by default in all cases, singularity doesn't provide a way to remove this bind if a user want to use the image `nsswitch.conf`
- OS images like alpine/busybox doesn't use NSS because of musl

For those reasons it's preferable to be explicit and let users to choose by using `-B /etc/nsswitch` if necessary and make things easier for all.

dtrudg (approved):
This LGTM

DrDaveD:
Ok on leaving out nsswitch.conf, looks good to me too.

PR #4938, opened 2020-01-10
Embed action scripts in runtime and create them on-the-fly

### LINKED ISSUE #4887: /.singularity.d/env not loaded when singularity is executed via srun/sbatch.
I am using singularity version 3.4.2

When I run the following command on our campus HPC system (Bigred3 at Indiana University), I get the following output.

```
singularity  exec -e docker://brainlife/freesurfer_on_mcr:6.0.2 /usr/bin/env
WARNING: skipping mount of /N/project: permission denied
TERM=xterm-256color
PERL5LIB=/usr/local/freesurfer/mni/share/perl5
SINGULARITY_APPNAME=
LOCAL_DIR=/usr/local/freesurfer/local
FSFAST_HOME=/usr/local/freesurfer/fsfast
MNI_PERL5LIB=/usr/local/freesurfer/mni/share/perl5
FMRI_ANALYSIS_DIR=/usr/local/freesurfer/fsfast
SINGULARITY_NAME=freesurfer_on_mcr_6.0.2.sif
MINC_BIN_DIR=/usr/local/freesurfer/mni/bin
SUBJECTS_DIR=/usr/local/freesurfer/subjects
PATH=/usr/local/freesurfer/bin:/usr/local/freesurfer/fsfast/bin:/usr/local/freesurfer/tktools:/usr/local/freesurfer/mni/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:\/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin
PWD=/gpfs/home/b/r/brlife/BigRed3/test/sbatch
FUNCTIONALS_DIR=/usr/local/freesurfer/sessions
LANG=C
MINC_LIB_DIR=/usr/local/freesurfer/mni/lib
PS1=Singularity> 
SHLVL=0
HOME=/N/u/brlife/BigRed3
MNI_DIR=/usr/local/freesurfer/mni
FREESURFER_HOME=/usr/local/freesurfer
SINGULARITY_CONTAINER=/N/dc2/scratch/brlife/singularity-cache-br3/cache/oci-tmp/34f96d505677bb18d831dbc2baae1986de1a8905ec6fadf998c1e2871f0ed741/freesurfer_on_mcr_6.0.2.sif
MNI_DATAPATH=/usr/local/freesurfer/mni/data
```

Please note that I see a lot of "freesurfer" related ENV parameters set when this container was built.

However, when I run the same command via srun, I get the following output instead.

```
srun singularity  exec -e docker://brainlife/freesurfer_on_mcr:6.0.2 /usr/bin/env
srun: job 103268 queued and waiting for resources
srun: job 103268 has been allocated resources
time="2019-12-19T16:47:43-05:00" level=warning msg="\"/run/user/1589653\" directory set by $XDG_RUNTIME_DIR does not exist. Either create the directory or unset $XDG_RUNTIME_DIR.: stat /run/user/1589653: no such file or directory: Trying to pull image in the event that it is a public image."
WARNING: skipping mount of /N/project: permission denied
WARNING: container does not have /.singularity.d/actions/exec, calling /usr/bin/env directly
SINGULARITY_CONTAINER=/N/dc2/scratch/brlife/singularity-cache-br3/cache/oci-tmp/34f96d505677bb18d831dbc2baae1986de1a8905ec6fadf998c1e2871f0ed741/freesurfer_on_mcr_6.0.2.sif
SINGULARITY_NAME=freesurfer_on_mcr_6.0.2.sif
TERM=xterm-256color
HOME=/N/u/brlife/BigRed3
PATH=/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin
LANG=C
SINGULARITY_APPNAME=
```

Please note the warning message("container does not have /.singularity.d/actions/exec, calling /usr/bin/env directly") and none of the fresurfer related ENVs are present.

I've already checked with our sysadmin to make sure that the same version of singularity / configuration is installed on both login node and all CEs on our cluster. They suggested contacting singularity team to further troubleshoot this problem.

I also run into the same problem if I run singularity via bash script submitted by sbatch. 

We are running SLES12 on our cluster.

```
cat /etc/os-release
NAME="SLES"
VERSION="12-SP3"
VERSION_ID="12.3"
PRETTY_NAME="SUSE Linux Enterprise Server 12 SP3"
ID="sles"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:12:sp3"
```

Singularity is installed by our sysadmin on the shared file system under the following directory.

```
/N/soft/cle6/singularity/3.4.2/bin/singularity
```

Here is the content of the singularity.conf

```
# SINGULARITY.CONF
# This is the global configuration file for Singularity. This file controls
# what the container is allowed to do on a particular host, and as a result
# this file must be owned by root.

# ALLOW SETUID: [BOOL]
# DEFAULT: yes
# Should we allow users to utilize the setuid program flow within Singularity?
# note1: This is the default mode, and to utilize all features, this option
# must be enabled.  For example, without this option loop mounts of image 
# files will not work; only sandbox image directories, which do not need loop
# mounts, will work (subject to note 2).
# note2: If this option is disabled, it will rely on unprivileged user
# namespaces which have not been integrated equally between different Linux
# distributions.
allow setuid = yes

# MAX LOOP DEVICES: [INT]
# DEFAULT: 256
# Set the maximum number of loop devices that Singularity should ever attempt
# to utilize.
max loop devices = 256

# ALLOW PID NS: [BOOL]
# DEFAULT: yes
# Should we allow users to request the PID namespace? Note that for some HPC
# resources, the PID namespace may confuse the resource manager and break how
# some MPI implementations utilize shared memory. (note, on some older
# systems, the PID namespace is always used)
allow pid ns = yes

# CONFIG PASSWD: [BOOL]
# DEFAULT: yes
# If /etc/passwd exists within the container, this will automatically append
# an entry for the calling user.
config passwd = yes

# CONFIG GROUP: [BOOL]
# DEFAULT: yes
# If /etc/group exists within the container, this will automatically append
# group entries for the calling user.
config group = yes

# CONFIG RESOLV_CONF: [BOOL]
# DEFAULT: yes
# If there is a bind point within the container, use the host's
# /etc/resolv.conf.
config resolv_conf = yes

# MOUNT PROC: [BOOL]
# DEFAULT: yes
# Should we automatically bind mount /proc within the container?
mount proc = yes

# MOUNT SYS: [BOOL]
# DEFAULT: yes
# Should we automatically bind mount /sys within the container?
mount sys = yes

# MOUNT DEV: [yes/no/minimal]
# DEFAULT: yes
# Should we automatically bind mount /dev within the container? If 'minimal'
# is chosen, then only 'null', 'zero', 'random', 'urandom', and 'shm' will
# be included (the same effect as the --contain options)
mount dev = yes

# MOUNT DEVPTS: [BOOL]
# DEFAULT: yes
# Should we mount a new instance of devpts if there is a 'minimal'
# /dev, or -C is passed?  Note, this requires that your kernel was
# configured with CONFIG_DEVPTS_MULTIPLE_INSTANCES=y, or that you're
# running kernel 4.7 or newer.
mount devpts = yes

# MOUNT HOME: [BOOL]
# DEFAULT: yes
# Should we automatically determine the calling user's home directory and
# attempt to mount it's base path into the container? If the --contain option
# is used, the home directory will be created within the session directory or
# can be overridden with the SINGULARITY_HOME or SINGULARITY_WORKDIR
# environment variables (or their corresponding command line options).
mount home = yes 

# MOUNT TMP: [BOOL]
# DEFAULT: yes
# Should we automatically bind mount /tmp and /var/tmp into the container? If
# the --contain option is used, both tmp locations will be created in the
# session directory or can be specified via the  SINGULARITY_WORKDIR
# environment variable (or the --workingdir command line option).
mount tmp = yes

# MOUNT HOSTFS: [BOOL]
# DEFAULT: no
# Probe for all mounted file systems that are mounted on the host, and bind
# those into the container?
mount hostfs = no

# BIND PATH: [STRING]
# DEFAULT: Undefined
# Define a list of files/directories that should be made available from within
# the container. The file or directory must exist within the container on
# which to attach to. you can specify a different source and destination
# path (respectively) with a colon; otherwise source and dest are the same.
# NOTE: these are ignored if singularity is invoked with --contain.
#bind path = /etc/singularity/default-nsswitch.conf:/etc/nsswitch.conf
#bind path = /opt
#bind path = /scratch
bind path = /etc/localtime
bind path = /etc/hosts
bind path = /N/home
bind path = /N/u
bind path = /N/dc2
bind path = /N/soft
bind path = /N/dcwan
bind path = /N/slate
bind path = /N/project

# USER BIND CONTROL: [BOOL]
# DEFAULT: yes
# Allow users to influence and/or define bind points at runtime? This will allow
# users to specify bind points, scratch and tmp locations. (note: User bind
# control is only allowed if the host also supports PR_SET_NO_NEW_PRIVS)
user bind control = yes

# ENABLE FUSEMOUNT: [BOOL]
# DEFAULT: yes
# Allow users to mount fuse filesystems inside containers with the --fusemount
# command line option.
enable fusemount = yes

# ENABLE OVERLAY: [yes/no/try]
# DEFAULT: try
# Enabling this option will make it possible to specify bind paths to locations
# that do not currently exist within the container.  If 'try' is chosen,
# overlayfs will be tried but if it is unavailable it will be silently ignored.
enable overlay = try

# ENABLE UNDERLAY: [yes/no]
# DEFAULT: yes
# Enabling this option will make it possible to specify bind paths to locations
# that do not currently exist within the container even if overlay is not
# working.  If overlay is available, it will be tried first.
enable underlay = yes

# MOUNT SLAVE: [BOOL]
# DEFAULT: yes
# Should we automatically propagate file-system changes from the host?
# This should be set to 'yes' when autofs mounts in the system should
# show up in the container.
mount slave = yes

# SESSIONDIR MAXSIZE: [STRING]
# DEFAULT: 16
# This specifies how large the default sessiondir should be (in MB) and it will
# only affect users who use the "--contain" options and don't also specify a
# location to do default read/writes to (e.g. "--workdir" or "--home").
sessiondir max size = 16

# LIMIT CONTAINER OWNERS: [STRING]
# DEFAULT: NULL
# Only allow containers to be used that are owned by a given user. If this
# configuration is undefined (commented or set to NULL), all containers are
# allowed to be used. This feature only applies when Singularity is running in
# SUID mode and the user is non-root.
#limit container owners = gmk, singularity, nobody

# LIMIT CONTAINER GROUPS: [STRING]
# DEFAULT: NULL
# Only allow containers to be used that are owned by a given group. If this
# configuration is undefined (commented or set to NULL), all containers are
# allowed to be used. This feature only applies when Singularity is running in
# SUID mode and the user is non-root.
#limit container groups = group1, singularity, nobody

# LIMIT CONTAINER PATHS: [STRING]
# DEFAULT: NULL
# Only allow containers to be used that are located within an allowed path
# prefix. If this configuration is undefined (commented or set to NULL),
# containers will be allowed to run from anywhere on the file system. This
# feature only applies when Singularity is running in SUID mode and the user is
# non-root.
#limit container paths = /scratch, /tmp, /global

# ALLOW CONTAINER ${TYPE}: [BOOL]
# DEFAULT: yes
# This feature limits what kind of containers that Singularity will allow
# users to use (note this does not apply for root).
allow container squashfs = yes
allow container extfs = yes
allow container dir = yes

# AUTOFS BUG PATH: [STRING]
# DEFAULT: Undefined
# Define list of autofs directories which produces "Too many levels of symbolink links"
# errors when accessed from container (typically bind mounts)
#autofs bug path = /nfs
#autofs bug path = /cifs-share

# ALWAYS USE NV ${TYPE}: [BOOL]
# DEFAULT: no
# This feature allows an administrator to determine that every action command
# should be executed implicitly with the --nv option (useful for GPU only 
# environments). 
always use nv = no

# ROOT DEFAULT CAPABILITIES: [full/file/no]
# DEFAULT: full
# Define default root capability set kept during runtime
# - full: keep all capabilities (same as --keep-privs)
# - file: keep capabilities configured in ${prefix}/etc/singularity/capabilities/user.root
# - no: no capabilities (same as --no-privs)
root default capabilities = no

# MEMORY FS TYPE: [tmpfs/ramfs]
# DEFAULT: tmpfs
# This feature allow to choose temporary filesystem type used by Singularity.
# Cray CLE 5 and 6 up to CLE 6.0.UP05 there is an issue (kernel panic) when Singularity
# use tmpfs, so on affected version it's recommended to set this value to ramfs to avoid
# kernel panic
memory fs type = tmpfs

# CNI CONFIGURATION PATH: [STRING]
# DEFAULT: Undefined
# Defines path from where CNI configuration files are stored
#cni configuration path =

# CNI PLUGIN PATH: [STRING]
# DEFAULT: Undefined
# Defines path from where CNI executable plugins are stored
#cni plugin path =

# MKSQUASHFS PATH: [STRING]
# DEFAULT: Undefined
# This allows the administrator to specify the location for mksquashfs if it is not
# installed in a standard system location
# mksquashfs path =

# CRYPTSETUP PATH: [STRING]
# DEFAULT: Undefined
# This allows the administrator to specify the location of cryptsetup if
# they wish to use custom location for this installation. If this value
# is undefined, at runtime singularity falls back to the value that was
# recorded at build time.
# cryptsetup path =

# SHARED LOOP DEVICES: [BOOL]
# DEFAULT: no
# Allow to share same images associated with loop devices to minimize loop
# usage and optimize kernel cache (useful for MPI)
shared loop devices = no

```

Please help us troubleshoot this problem!

cclerget:

Embed action scripts in runtime and create them on-the-fly in session directory. We don't bind action scripts directory anymore into container to avoid issue on some CRAY systems reporting 'srmount error' when accessing action scripts from container and making singularity unusable.

 - Fixes #4887 


dtrudg (commented):
This LGTM - I'm holding back from ✅ until we have it confirmed by someone on a CRAY if possible.

jmstover:
@cclerget @dctrud 
Just thought to mention... Would it be better to *not* have this in the Go code? Maybe just use LIBEXECDIR and read/copy the file in on execution instead of bind mounting? Could still use SYSCONFDIR in that case ... but just have the scripts outside of the Go code.

cclerget:
@jmstover Is it related to give the possibility to customize them ?

jmstover:
Yes. Mostly the ability to modify the scripts.

cclerget:
I'm a bit worried with copy that it won't fix issue #4887 and we could potentially get same `srmount error` during copy but we would need confirmation.

jmstover:
What is the difference between reading one file in and writing out a copy, or writing it out from a variable?

If it's the location, wouldn't anything under that location have the chance of getting that error when accessed then? Like `singularity.conf`, or CNI binaries, or wherever the location is decided to keep the source files.

cclerget:
`singularity.conf` is parsed in the process residing in host namespace not in the process creating container environment where the copy would occur if we go this way, regarding CNI plugins we actually don't know if it works on those CRAY systems but network stuff is not really relevant in HPC world.

I don't have really any preference for one or other solution, I'm just concerned if the process doing the copy of those files would get this error because we don't really know if it comes from using a dedicated mount namespace, pivot_root or another thing involved in the container creation.

We could test your suggestion first and see ...

jmstover:
Hrmm... I thought that they were added in somewhere in the same host namespace before a mount in the new namespace was created. Meh.

Would underlay have the same possibility of failure then? Or is there some behaviour difference that would mitigate the `srmount` error in the underlay case?

dtrudg:
Having thought about this more, I am comfortable with not having these scripts customizable. On the mailing list, @cclerget detailed how you can inject an additional script that will be sourced, and use it to modify the environment. Custom content in the driver scripts that I've seen seems to be used to set LD_LIBRARY_PATH pointing to host binaries, and other things that often cause issues elsewhere - and lead to containers being created on a system that are not really portable.

dtrudg (approved):

PR #4944, opened 2020-01-13
Ensure build rootPath has owner write permission

### LINKED ISSUE #4943: build: failed to make environment dirs
### Version of Singularity:

```
version 3.5.2-1.1.el7
```

Some Docker container images do not enable user write permission on /, which causes Singularity to fail to insert the /.singularity.d directotory into the image during the SIF conversion/build process.  One such container is CERN's CC7 base image:
```
$ singularity shell docker://gitlab-registry.cern.ch/linuxsupport/cc7-base:latest
ERROR:   build: failed to make environment dirs: mkdir /home/tmp/rootfs-f985df87-3624-11ea-a275-0cc47afd3556/.singularity.d: permission denied
FATAL:   Unable to handle docker://gitlab-registry.cern.ch/linuxsupport/cc7-base:latest uri: unable to build: packer failed to pack: while inserting base environment: build: failed to make environment dirs: mkdir /home/tmp/rootfs-f985df87-3624-11ea-a275-0cc47afd3556/.singularity.d: permission denied

I've reproduced the problem with both the EPEL 3.5.2 RPM on SL7, and via the 3.5.2 source on Fedora 31.

### PR DESCRIPTION
**Pull request by hollowec**

Ensure we can write the /.singularity.d directory during builds as a potential resolution for #4943.

 - Fixes #4943

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -64,7 +64,7 @@ require (
 	github.com/opencontainers/selinux v1.3.0
 	github.com/pelletier/go-toml v1.6.0
 	github.com/phayes/freeport v0.0.0-20180830031419-95f893ade6f2 // indirect
-	github.com/pkg/errors v0.8.1
+	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v0.0.0-20180607123607-faf4ec335fe0 // indirect
 	github.com/prometheus/client_model v0.0.0-20171117100541-99fa1f4be8e5 // indirect
 	github.com/prometheus/common v0.0.0-20180518154759-7600349dcfe1 // indirect
--- a/go.sum
+++ b/go.sum
@@ -252,6 +252,8 @@ github.com/phayes/freeport v0.0.0-20180830031419-95f893ade6f2/go.mod h1:iIss55rK
 github.com/pkg/errors v0.8.0/go.mod h1:bwawxfHBFNV+L2hUp1rHADufV3IMtnDRdf1r5NINEl0=
 github.com/pkg/errors v0.8.1 h1:iURUrRGxPUNPdy5/HRSm+Yj6okJ6UtLINN0Q9M4+h3I=
 github.com/pkg/errors v0.8.1/go.mod h1:bwawxfHBFNV+L2hUp1rHADufV3IMtnDRdf1r5NINEl0=
+github.com/pkg/errors v0.9.1 h1:FEBLx1zS214owpjy7qsBeixbURkuhQAwrK5UwLGTwt4=
+github.com/pkg/errors v0.9.1/go.mod h1:bwawxfHBFNV+L2hUp1rHADufV3IMtnDRdf1r5NINEl0=
 github.com/pmezard/go-difflib v1.0.0 h1:4DBwDE0NGyQoBHbLQYPwSUPoCMWR5BEzIk/f1lZbAQM=
 github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
 github.com/pquerna/ffjson v0.0.0-20181028064349-e517b90714f7 h1:gGBSHPOU7g8YjTbhwn+lvFm2VDEhhA+PwDIlstkgSxE=

### CODE REVIEW

File: internal/pkg/build/sources/base_environment.go [hunk 1]
```diff
@@ -416,6 +417,22 @@ func makeFiles(rootPath string) error {
 }
 
 func makeBaseEnv(rootPath string) (err error) {
+
+	var info os.FileInfo
+
+	// Ensure we can write into the root of rootPath
+	if info, err = os.Stat(rootPath); err != nil {
```
dctrud: should be able to `if info, err :=` here, and drop the var above.

internal/pkg/build/sources/base_environment.go
```diff
@@ -123,26 +124,26 @@ fi
 `
 	// Contents of /.singularity.d/env/01-base.sh
 	baseShFileContent = `#!/bin/sh
-#
```
dtrudg: Even if the trailing spaces are incorrect, please try to avoid changing it in this PR, to keep the diffs clean.

dtrudg (changes requested):
Hi @hollowec 

Thanks for the contribution. Have left a couple of comments, and here is the output of the failing CI:

```
INFO [runner] linters took 14.091142577s with stages: goanalysis_metalinter: 14.083234869s 
internal/pkg/build/sources/base_environment.go:13: File is not `gofmt`-ed with `-s` (gofmt)
	"github.com/sylabs/singularity/internal/pkg/util/fs"
internal/pkg/build/sources/base_environment.go:449:2: naked return in func `makeBaseEnv` with 31 lines of code (nakedret)
	return
	^
internal/pkg/build/sources/base_environment.go:425:13: unusedresult: result of fmt.Errorf call not used (govet)
		fmt.Errorf("build: failed to stat rootPath: %v", err)
		          ^
internal/pkg/build/sources/base_environment.go:431:14: unusedresult: result of fmt.Errorf call not used (govet)
			fmt.Errorf("build: failed to make rootPath writable: %v", err)
			          ^
```

For this type of issue we would also like to have a test in place our e2e testing suite:

https://github.com/sylabs/singularity/blob/master/e2e/regressions/build.go

Myself or @cclerget can add this, if it's not something you'd like to look into.

In addition @cclerget - do we want to include the code as is here, or would generalizing a 'force creation' option to `MakeDirs` or, adding a separate `ForceMakeDirs` be preferable?

internal/pkg/build/sources/base_environment.go
hollowec: Sorry, my editor just corrected the whitespace automatically.  I'll revert the change.

cclerget:
> do we want to include the code as is here, or would generalizing a 'force creation' option to MakeDirs or, adding a separate ForceMakeDirs be preferable?

@dctrud I think it's fine as is, except `.singularity.d` requiring creation of sub directory ... but looks like it's not the issue, let's go with that and if we encounter issues for few other directories we could refactor the code a bit for generalization.

internal/pkg/build/sources/base_environment.go
hollowec: OK, I've fixed the CI/formatting and whitespace issues.  Let me know if you'd like me to squash the commits.

hollowec:
Hi @dctrud,

If you or @cclerget have time to work on the E2E test that would be great.  If you're busy, I should have time later in the week: let me know.  Thanks.

dtrudg:
@hollowec - we are probably busy until early next week to work on this. If you don't have the opportunity to look at the test we can pick it up then.

internal/pkg/build/sources/base_environment.go
hollowec: I've just squashed the commits as it's cleaner in my opinion.

hollowec (commented):
OK thanks, I started working on the E2E test... I should have something later today or tomorrow.

hollowec:
Hi @dctrud,

I've added an e2e test as requested.  However, it is currently using the versioned CERN image directly from their gitlab docker registry, and I'm not sure if that is acceptable.  Would you prefer to push this image to a sylabsio dockerhub area to eliminate the external dependency?

e2e/regressions/build.go
```diff
@@ -176,6 +176,17 @@ func (c *regressionsTests) issue4583(t *testing.T) {
 	)
 }
 
+func (c *regressionsTests) issue4943(t *testing.T) {
+	c.env.RunSingularity(
+		t,
+		e2e.WithProfile(e2e.UserProfile),
```
cclerget: Could you use build instead ? As we are in `regressions/build.go`

```
e2e.WithProfile(e2e.UserProfile),
e2e.WithCommand("build"),
e2e.WithArgs("--force", "/dev/null", "docker://gitlab-registry.cern.ch/linuxsupport/cc7-base:20191107"),
e2e.ExpectExit(0),
```

hollowec: OK done.

cclerget:
For the second time I mistakenly force pushed changes into a PR implying to close it, really sorry for that @hollowec , wanted just to cleanup e2e tests for a last review by @dctrud . Restored in #4958.

PR #4948, opened 2020-01-14
Add image bind support

### LINKED ISSUE #4862: Allow to specify SIF as separate overlay
Loading a SIF image as overlay actually returned error : `SIF image not supported as overlay image`. Should be able to load squashfs/ext3 SIF partition in a first time.

### LINKED ISSUE #4931: RFE: Use images as bind source
### Proposal
The idea is to allow the use of images as a bind source in order to bind image content into container.

We currently have the following bind syntax :

`--bind source:destination[:options]`

`options` recognized values are `ro` and `rw`:
* `ro` to bind source read-only into the container
* `rw` is set by default and doesn't need to be specified

We could reuse this syntax by adding options to control image bind, we would have :

* `image-src=/` as **required** option to tell we are binding an image as bind source, the parameter would allow to specify the directory (or file) in image to bind as the source into the container
* `sif-id=X` as **optional** option where `X` would be the SIF descriptor ID to use as partition

#### Supported image formats :
- SIF
- Squash
- Ext3

### Example

Package input data:

```
mkdir -p /tmp/data/input
echo "data 1" > /tmp/data/input/data.1
echo "data 2" > /tmp/data/input/data.2
mksquashfs /tmp/data /tmp/input.sqs -all-root -comp gzip
unsquashfs -ls -d "" /tmp/input.sqs 
/input
/input/data.1
/input/data.2
```

Create SIF image and add squashfs data :
```
singularity sif new /tmp/input.sif
singularity sif add --datatype 4 --partarch 2 --partfs 1 --parttype 3 /tmp/input.sif /tmp/input.sqs
singularity sif list /tmp/input.sif
Container id: d9411a2a-afbc-439e-8268-6443af300f61
Created on:   2020-01-08 14:25:31 +0100 CET
Modified on:  2020-01-08 14:25:38 +0100 CET
Descriptor list:
1    |NONE    |NONE    |32768-36864               |FS (Squashfs/Data/amd64)
```

Now to mount `/input` directory stored in `/tmp/input.sif` image to `/img_input` in container, we would specify the following bind :

`--bind /tmp/input.sif:/img_input:image-src=/input`
or
`--bind /tmp/input.sif:/img_input:image-src=/input,sif-id=1`

We could also mount the entire image in container with `image-src=/` like this :

`--bind /tmp/input.sif:/image:image-src=/`

where `/image` in container would contain `/image/input` directory

cclerget:

Allow to specify image as bind mount point like:

`-B /tmp/image:/input:image-src=/`

 - Fixes #4931 
 - Fixes #4862 

### DIFF
--- a/internal/pkg/runtime/engine/singularity/container_linux.go
+++ b/internal/pkg/runtime/engine/singularity/container_linux.go
@@ -35,6 +35,7 @@ import (
 	"github.com/sylabs/singularity/pkg/image"
 	"github.com/sylabs/singularity/pkg/network"
 	singularity "github.com/sylabs/singularity/pkg/runtime/engine/singularity/config"
+	singularityConfig "github.com/sylabs/singularity/pkg/runtime/engine/singularity/config"
 	"github.com/sylabs/singularity/pkg/util/fs/proc"
 	"github.com/sylabs/singularity/pkg/util/gpu"
 	"github.com/sylabs/singularity/pkg/util/loop"
@@ -82,6 +83,10 @@ type container struct {
 
 func create(ctx context.Context, engine *EngineOperations, rpcOps *client.RPC, pid int) error {
 	var err error
+
+	if len(engine.EngineConfig.GetImageList()) == 0 {
+		return fmt.Errorf("no root filesystem image provided")
+	}
 
 	engine.EngineConfig.File, err = singularityconf.Parse(buildcfg.SINGULARITY_CONF_FILE)
 	if err != nil {
@@ -158,6 +163,9 @@ func create(ctx context.Context, engine *EngineOperations, rpcOps *client.RPC, p
 	}
 
 	if err := c.addRootfsMount(system); err != nil {
+		return err
+	}
+	if err := c.addImageBindMount(system); err != nil {
 		return err
 	}
 	if err := c.addKernelMount(system); err != nil {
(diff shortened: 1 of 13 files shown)


polaya07 (approved):

internal/pkg/runtime/engine/singularity/prepare_linux.go
```diff
@@ -480,9 +499,106 @@ func (e *EngineOperations) prepareAutofs(starterConfig *starter.Config) error {
...
+	return binds, nil
+}
+
 // prepareContainerConfig is responsible for getting and applying
 // user supplied configuration for container creation.
 func (e *EngineOperations) prepareContainerConfig(starterConfig *starter.Config) error {
+	// fix bind paths as CLI has splitted binds based on the
+	// comma separator and we need to support various options
+	// for bind path so we can pass as bind argument:
+	// -B "/etc:/host, /tmp/img.sif:/input:image-src=/,ro"
```
dtrudg: Is there any way we can sort out the splitting in the CLI? This is buried pretty deeply down here, so that it's not going to obvious what's going on if you are working on the CLI code.

internal/pkg/runtime/engine/singularity/prepare_linux.go
```diff
@@ -480,9 +499,106 @@ func (e *EngineOperations) prepareAutofs(starterConfig *starter.Config) error {
...
+	// we need to parse various syntax:
+	// source1
+	// source1:destination1
+	// source1:destination1:option1
+	// source1:destination1:option1,option2
+	// source1,source2
+	// source1:destination1:option1,source2
+	re := regexp.MustCompile(`([^,^:]+:?)`)
+
+	for _, m := range re.FindAllString(strings.Join(bindPath, ","), -1) {
```
dtrudg: The complexity here really makes me think we need to handle it better in the CLI, or have a different syntax without comma used in two ways?

dtrudg (changes requested):
This works when I test it.

Needs to be done...

* The help message for the -B option should be updated to reflect the syntax

General thoughts...

* The function `loadImages` continues to get longer and more complex. It's really hard for me to follow it properly when reading through, given the length and the depth of nesting. Could this be re-factored to make it more easy to follow?

* The complex code to fix the bindpaths because of `,` separating either bind paths or options feels like it should be elsewhere. It's really quite buried here, so that if you are working further up in the codebase you would see the split on `,` but not be aware of what's going on here.

internal/pkg/runtime/engine/singularity/prepare_linux.go
cclerget: cobra and pflag have two type `StringSlice` (commas separated) and `StringArray` (without commas), I made a first attempt with `StringArray` but it breaks CLI as users won't be able to specify binds like `-B /var/log,/var/run` and instead to specify them like `-B /var/log -B /var/run`. And even with `StringArray` there is also the case of environment variables `SINGULARITY_BIND` and `SINGULARITY_BINDPATH` to take into account because there is no other ways to specify multiple binds other than with `,` without again breaking CLI.
I'm all open to suggestions

dtrudg: Thanks for explaining that... makes sense that is limiting the options here.

internal/pkg/runtime/engine/singularity/container_linux.go
```diff
@@ -823,108 +791,234 @@ func (c *container) addOverlayMount(system *mount.System) error {
...
-	if hasUpper {
-		if err := system.RunAfterTag(mount.PreLayerTag, c.overlayUpperWork); err != nil {
-			return err
+			datas, err := img.GetDataPartitions()
+			if err != nil {
+				return fmt.Errorf("while getting data partition for %s: %s", img.Path, err)
+			}
+			if len(datas) == 0 && id <= 0 {
+				return fmt.Errorf("no bindable data partition found in %s", img.Path)
+			} else if len(datas) > 0 {
```
ikaneshiro: So if we have multiple data partitions to choose from in a SIF, and a valid `id` specified, we still just take the first data partition we find? Unless i'm misunderstanding the role of `id`, it seems to me that this should be the last conditional statement in this flow.

ikaneshiro (commented):
I agree with @dctrud that the bind path handling is getting a bit confusing. Imho, it would be great if we could parse these paths and options in the cli and set options in the engine configuration to make some of the engine logic less reliant on string parsing. It's hard for me to gauge how much refactoring this would require though.

dtrudg (commented):
@cclerget - this looks like a nice step to making it more readable for me :-) ready for final review now?

cclerget:
> this looks like a nice step to making it more readable for me :-) ready for final review now?

Yes it does :)

dtrudg (approved):
:+1: from me, this is nicer to follow. @ikaneshiro do you want / are you able to take another read?

ikaneshiro (approved):
LGTM

PR #4953, opened 2020-01-16
Cleanup FUSE code

cclerget:

Cleanup FUSE code and remove "plugin" term to make it part of core, `--fusemount` actually allows to run different FUSE program without the need of plugins.

### DIFF
--- a/cmd/starter/c/starter.c
+++ b/cmd/starter/c/starter.c
@@ -36,6 +36,7 @@
 #include <sys/syscall.h>
 #include <net/if.h>
 #include <sys/eventfd.h>
+#include <sys/sysmacros.h>
 #include <linux/magic.h>
 
 #ifdef SINGULARITY_SECUREBITS
@@ -852,6 +853,29 @@ static void fix_streams(void) {
     }
 }
 
+/* fix_userns_devfuse_fd reopens /dev/fuse file descriptors in a user namespace */
+static void fix_userns_devfuse_fd(struct starter *starter) {
+    struct stat st;
+    int i, newfd, oldfd;
+
+    for ( i = 0; i < starter->numfds; i++ ) {
+        oldfd = starter->fds[i];
+        if ( fstat(oldfd, &st) < 0 ) {
+            fatalf("Failed to get file information for file descriptor %d: %s\n", oldfd, strerror(errno));
+        }
+        if ( major(st.st_rdev) == 10 && minor(st.st_rdev) == 229 ) {
+            newfd = open("/dev/fuse", O_RDWR);
+            if ( newfd < 0 ) {
+                fatalf("Failed to open /dev/fuse: %s\n", strerror(errno));
+            }
+            if ( dup3(newfd, oldfd, O_CLOEXEC) < 0 ) {
+                fatalf("Failed to duplicate file descriptor: %s\n", strerror(errno));
+            }
+            close(newfd);
+        }
+    }
+}
+
 static void wait_child(const char *name, pid_t child_pid, bool noreturn) {
     int status;
     int exit_status = 0;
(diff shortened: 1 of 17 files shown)


ikaneshiro (dismissed):

cmd/starter/c/starter.c
```diff
@@ -852,6 +853,28 @@ static void fix_streams(void) {
     }
 }
 
+static void fix_userns_devfuse_fd(struct starter *starter) {
```
dtrudg, line 857: I think I understand why this is here to duplicate the fd - but could you write a short comment on why it's needed so it's easier to follow?

-edit- maybe not needed, I see a comment where it is called.

e2e/actions/actions.go
```diff
@@ -14,6 +14,8 @@ import (
 	"strconv"
 	"testing"
 
+	uuid "github.com/satori/go.uuid"
+
```
dtrudg: can you drop the empty line, so gofmt sorts them properly?

internal/pkg/runtime/engine/singularity/container_linux.go
```diff
@@ -2069,17 +2074,84 @@ func (c *container) prepareNetworkSetup(system *mount.System, pid int) (func(con
...
+		if err != nil {
+			return usernsFd, fmt.Errorf("while receiving file descriptors: %s", err)
+		}
+
+		msgs, err := unix.ParseSocketControlMessage(buf)
+		if err != nil {
+			return usernsFd, fmt.Errorf("while parsing socket control message: %s", err)
+		}
+
+		newfds := make([]int, 0)
```
dtrudg: should we make with size at least 1? If we don't get any FDs we won't do anything with this.... but otherwise we have at least 1 append.

internal/pkg/runtime/engine/singularity/container_linux.go
```diff
@@ -2069,17 +2074,84 @@ func (c *container) prepareNetworkSetup(system *mount.System, pid int) (func(con
...
+
+		if len(newfds) == 0 {
+			return usernsFd, fmt.Errorf("no file descriptor received")
 		}
 
+		usernsFd = newfds[len(newfds)-1]
+		newfds = newfds[0 : len(newfds)-1]
+
+		if len(newfds) != len(fds) {
+			return usernsFd, fmt.Errorf("get %d file descriptors instead of %d", len(newfds), len(fds))
```
dtrudg:
```suggestion
			return usernsFd, fmt.Errorf("got %d file descriptors instead of %d", len(newfds), len(fds))
```

internal/pkg/runtime/engine/singularity/container_linux.go
```diff
@@ -2094,39 +2166,49 @@ func (c *container) addFuseMount(system *mount.System) error {
...
+			opts,
+		)
+		if err != nil {
 			sylog.Debugf("Calling AddFS: %+v\n", err)
-			return err
+			return usernsFd, err
+		}
+
+		if !fuseMounts[i].FromContainer && fakerootHybrid {
+			nsenter := []string{"nsenter", "--user=/dev/fd/4", "-F", "--preserve-credentials"}
```
dtrudg, line 2209: What is `/dev/fd/4` here can you put a comment so it's clear?

stdin, stdout and stderr are 0, 1, and 2... first extra is 3, second extra is 4... but why is it always what we want?

dtrudg (commented):
Generally LGTM but I'd like to know about the `fd4` bit for certain before :white_check_mark:

internal/pkg/runtime/engine/singularity/container_linux.go
cclerget: the check after will return appropriate error if there is no file descriptor returned

dtrudg (approved):

Issue #4993, opened 2020-01-30
Find an alternative for github.com/opencontainers/runtime-tools and OCI configuration


cclerget:
This package has a huge impact on the resulting binary size, it add almost 13Mb by binary which is just ridiculous while adding zero value for Singularity in general.

PR #4998, opened 2020-01-30
Reorganize singularity engine configuration to reduce plugin binary size

cclerget:

Reorganize singularity engine configuration to get rid of non public struct fields and remove package dependencies which grow plugin binary size uselessly by almost 7 Mbytes.

 - Related to #4993 


internal/app/singularity/plugin_compile_linux.go
```diff
@@ -223,6 +223,7 @@ func buildPlugin(sourceDir string, bTool buildToolchain) (string, error) {
 		"-o", out,
 		"-mod=readonly",
 		"-trimpath",
+		"-ldflags=-s -w",
```
dtrudg: Does `-s` strip-all retain nice things in any Go stack traces? Also - what's the procedure for someone who wants to build a plugin *with* any debugging symbols retained?

pkg/runtime/engine/singularity/config/config.go
```diff
@@ -25,6 +26,21 @@ const (
 	UnderlayLayer = "underlay"
 )
 
+// EngineConfig stores both the JSONConfig and the FileConfig.
```
dtrudg: No longer has FileConfig?

pkg/runtime/engine/singularity/config/config.go
```diff
@@ -25,6 +26,21 @@ const (
...
+type EngineConfig struct {
+	JSON      *JSONConfig `json:"jsonConfig"`
+	OciConfig *oci.Config `json:"ociConfig"`
+}
+
+// NewConfig returns singularity.EngineConfig with a parsed FileConfig
+func NewConfig() *EngineConfig {
```
dtrudg, line 41: No longer has Fileconfig?

internal/pkg/runtime/engine/singularity/prepare_linux.go
```diff
@@ -163,6 +163,12 @@ func (e *EngineOperations) PrepareConfig(starterConfig *starter.Config) error {
 		return err
 	}
 
+	// fix for RHEL 6 hosts where ttyname returns a wrong terminal name
+	// (see process_linux.go for details)
+	if e.FileConfig.MountDev == "minimal" || e.EngineConfig.GetContain() {
```
dtrudg: If this is a fix, and not just a reorganization, could it be made as a separate PR, so we would be able to confirm the reorganization  maintains exactly the same behaviour before/after?

dtrudg (commented):
Hi @cclerget - left a few comments.

Could you say how much of the reduction in size is from the ldflags stripping the output, and how much is from the configuration?

I think it would be nice if...
* reorganization
* ldflags
* devconsole fix

... were in separate commits, so they can be tried in stages to see the effect easily.

internal/app/singularity/plugin_compile_linux.go
cclerget: Stack trace remains but with less information. There two approaches:
- don't use it by default and let users defined it via `GOFLAGS=-ldflags="-s -w" singularity plugin compile`
- don't use it by default and add a `--strip` flag
- use it by default and add a `--disable-strip` flag

For the second round of plugin the idea is to also generate a basic `Makefile` with `singularity plugin create` to have something like:
- `make`
- `sudo make install`
...
Making the first approach above more appropriate to have something like:
`make debug && sudo make install`

internal/pkg/runtime/engine/singularity/prepare_linux.go
cclerget: Will create a separate PR. This is a fix because actually the minimal check https://github.com/sylabs/singularity/blob/dbe15bdc62d81cbf6a45504aefefb3beddd25da0/internal/pkg/runtime/engine/singularity/process_linux.go#L71 has no effect as `singularity.conf` can't be parsed from container and the configuration is not passed between stage 1 and 2, so `e.EngineConfig.File.MountDev` is always equal to `yes` in stage 2

internal/app/singularity/plugin_compile_linux.go
dtrudg: Understood :+1:

dtrudg (approved):

PR #4999, opened 2020-01-30
Add build command sanity check to close #3502

### LINKED ISSUE #3502: add sanity check to build args
## Version of Singularity:

3.2

### Expected behavior

If someone specifies the args backwards to a build, the build should abort with an obvious warning.  While we don't mandate the definition file having a particular name or filetype, one option is to see if destination is a text file and input is a sif.  

Suggestion:
[rherban@test]:~$ sudo singularity build stage.def stage.sif
Build target appears to be a definition file, did you mean "singularity build stage.sif stage.def"
Aborting build

### Actual behavior

[rherban@test]:~$ sudo singularity build stage.def stage.sif
INFO:    Starting build...
INFO:    Creating SIF file...
INFO:    Build complete: stage.def

The definition file is overwritten by stage.sif and any changes to the definition are lost. 

Specify the image path and build spec backwards.
[rherban@test]:~$ sudo singularity build stage.def stage.sif

sudosays:

Addresses #3502
Modified the `checkBuildTarget` function to first try and parse the target as a def file if build target already exists and force overwrite flag is false. This method is insensitive to filenaming and is based on the content of the file itself.

 - Closes #3502

### DIFF
--- a/cmd/internal/cli/build.go
+++ b/cmd/internal/cli/build.go
@@ -266,14 +266,10 @@ func checkBuildTarget(path string) error {
 		}
 		if !buildArgs.update && !forceOverwrite {
 
-			var question string
+			question := fmt.Sprintf("Build target '%s' already exists. Do you want to overwrite? [N/y]", f.Name())
 
-			isDefFile, _ := parser.IsValidDefinition(path)
-			if isDefFile {
-				question = "Build target '" + f.Name() + "' is a definition file that will be overwritten. Do you still want to overwrite? [N/y]"
-			} else {
-
-				question = "Build target '" + f.Name() + "' already exists. Do you want to overwrite? [N/y]"
+			if isDefFile, _ := parser.IsValidDefinition(path); isDefFile {
+				question = fmt.Sprintf("Build target '%s' is a definition file that will be overwritten. Do you still want to overwrite? [N/y]", f.Name())
 			}
 
 			input, err := interactive.AskYNQuestion("n", question)
--- a/cmd/internal/cli/build.go
+++ b/cmd/internal/cli/build.go
@@ -265,7 +265,17 @@ func checkBuildTarget(path string) error {
 			return fmt.Errorf("only sandbox update is supported: %s is not a directory", path)
 		}
 		if !buildArgs.update && !forceOverwrite {
-			question := "Build target already exists. Do you want to overwrite? [N/y] "
+
+			var question string
+
+			isDefFile, _ := parser.IsValidDefinition(path)
+			if isDefFile {
+				question = "Build target is a valid definition file that will be overwritten. Do you still want to overwrite? [N/y]"
+			} else {
+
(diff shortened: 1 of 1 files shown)


cmd/internal/cli/build.go
```diff
@@ -265,7 +265,17 @@ func checkBuildTarget(path string) error {
 			return fmt.Errorf("only sandbox update is supported: %s is not a directory", path)
 		}
 		if !buildArgs.update && !forceOverwrite {
-			question := "Build target already exists. Do you want to overwrite? [N/y] "
+
+			var question string
+
+			isDefFile, _ := parser.IsValidDefinition(path)
+			if isDefFile {
+				question = "Build target is a valid definition file that will be overwritten. Do you still want to overwrite? [N/y]"
```
dtrudg: I think `valid` is unnecessary here

dtrudg (commented):
Thanks @sudosays !

While looking at this, I wondered what if we make this even more explicit, and put the basename of the target file in the question messages in both cases? That should maximally highlight to the user what they are about to overwrite.

`Build target 'mydef.def' is a definition file that will be overwritten. Do you still want to overwrite? [N/y]`

`Build target 'myoldsif.sif' already exists. Do you want to overwrite? [N/y]`

sudosays:
@dctrud I agree that is much better. I also dropped the "valid" in the new commit. Now it specifies the base name of the file as you suggested.

cmd/internal/cli/build.go
```diff
@@ -265,7 +265,17 @@ func checkBuildTarget(path string) error {
 			return fmt.Errorf("only sandbox update is supported: %s is not a directory", path)
 		}
 		if !buildArgs.update && !forceOverwrite {
-			question := "Build target already exists. Do you want to overwrite? [N/y] "
+
+			var question string
```
dtrudg: Apologies - I should have made this comment earlier. Please let me know if you'd rather I commit this change to the branch, rather than you needing to do it, and I can take care of that to merge this.


It would probably be more common to write something like this...

```
question := fmt.Sprintf("Build target '%s' already exists. Do you want to overwrite? [N/y]", f.Name())

if isDefFile, _ := parser.IsValidDefinition(path); isDefFile {
    question = fmt.Sprintf("Build target '%s' is a definition file that will be overwritten. Do you still want to overwrite? [N/y]", f.Name())	
}
```

The pattern of setting a default value with `:=` and overriding for the special case, rather than using `var` and `if/else` blocks is generally used in our code, and in Go code in general.

dtrudg (changes requested):

cmd/internal/cli/build.go
sudosays: I have fixed it on my side with the new commit. Thanks for the advice, I'll make sure to conform to that style in future.

dtrudg (approved):

Issue #1228, opened 2017-12-21
mksquashfs takes up all cores on build system


gnguy:
## Version of Singularity:

2.4.2

### Expected behavior
 
`singularity build` utilizes one core when calling mksquashfs, unless otherwise specified.

### Actual behavior

`singularity build` utilizes all cores available on machine

### Steps to reproduce behavior

`singularity build` with a large (>10 gb gb) image. On a 40+ core machine, this operation takes up all cores available on the machine. I believe this is due to the [default settings for the processors option in mksquashfs](http://manpages.ubuntu.com/manpages/trusty/man1/mksquashfs.1.html):

> -processors NUMBER
>            Use NUMBER processors. By default will  use  number  of  processors
>            available.

Would it be possible to add an option to Singularity build to specify the number of processors used, to pass to mksquashfs [when it is called](https://github.com/singularityware/singularity/blob/f2d7795d04bad5af371c3a4a0a6f372ba07ff810/libexec/cli/build.exec#L367)? Otherwise, I'm not sure about straightforward ways to control processor usage -- I'm looking into using [numactl](https://www.systutorials.com/docs/linux/man/8-numactl/) as a workaround, but it would be nice to be able to control the cpu usage either via an environmental variable (similar to the SINGULARITY_PYTHREADS environmental variable) or as an explicit option to singularity build.

standage:
2+ years later, this is still an issue with Singularity 3.4.1. Being unable to configure the number of processors used for `singularity build` is problematic in a lot of contexts.

standage:
Would it be as straightforward as updating this code block and appending `-processors 1` (or an alternative user-supplied value)?

https://github.com/sylabs/singularity/blob/d642f2bd3e96d77d0fa18611059407fb4f697aa6/pkg/image/packer/squashfs.go#L38-L41

**UPDATE**: as subsequent discussion has shown, no it is NOT that simple! 🤔

jmstover:
About ... would probably want to instead do it in `internal/pkg/build/assemblers/sif.go` in the `Assemble` function. See where it's doing things like:
```
flags = append(flags, "-comp", "gzip")
```

So, we'd probably want to add in a `--cpu` or `--processors` build flag to set the `-processors` option if passed, or just used the default (which is all CPUs).

@cclerget @dctrud  Thoughts on this approach?

dtrudg:
It would be nice to have this in the next release, agreed.

[...]

PR #5003, opened 2020-02-03
Add options to the config file to allow admins to set limits on CPU and memory for mksquashfs

dmbeeson:

By default mksquashfs uses all CPU cores and does not limit RAM usage.  This change is to add two new options to singularity.conf to allow the admin to set number of CPUs to use and max RAM usage when mksquashfs is called to create the filesystem  By default they are not set and the behavior doesn't change.  Only if the admin sets them will there be different behavior.


jmstover (commented):
LGTM ... This probably shouldn't be the overall final form (probably want a way for user to set it up to the value set in the `singularity.conf` file). But functionally it looks good and IMO would be good to finally get something in place.

@ikaneshiro  .... May want to look this over a bit.

Leaving for others to look over though. ;)

internal/pkg/build/assemblers/sif.go
```diff
@@ -177,7 +179,12 @@ func (a *SIFAssembler) Assemble(b *types.Bundle, path string) error {
 	if a.GzipFlag {
 		flags = append(flags, "-comp", "gzip")
 	}
-
+	if a.MksquashfsMem != "" {
+		flags = append(flags, "-mem", a.MksquashfsMem)
```
cclerget, line 183: Looks like `-mem` is not portable enough, old `mksquashfs` version doesn't have this option

pkg/util/singularityconf/config.go
```diff
@@ -42,6 +42,8 @@ type File struct {
 	CniConfPath             string   `directive:"cni configuration path"`
 	CniPluginPath           string   `directive:"cni plugin path"`
 	MksquashfsPath          string   `directive:"mksquashfs path"`
+	MksquashfsProcs         string   `directive:"mksquashfs procs"`
```
cclerget: a `uint` instead would take care of the conversion and in a first time ensure that the value is greater or equal to 0

dmbeeson: Good point! I just added a commit to convert MksquashfsProcs to uint.

internal/pkg/build/assemblers/sif.go
dmbeeson: From the changelog it appears that this option was added to mksquashfs in May of 2014.  I think this is a useful configurable option (I know I need it).  Any ideas as to how to get this available without breaking people?  I can change it so that if the value is not set in the config file it doesn't pass the command line argument at all and add a note about version needed in the singularity.conf file.  Could we add a version requirement for mksquasfs to ensure that it is a new enough version that it is supported?  Perhaps, if not, add it as a configurable value for mconfig that will be required to enable it to show in the config file?  Other ideas?

jmstover: IIRC, RHEL 6 and SLES 11 don't have a new enough version. RHEL 6 is EOL 2024, and though SLES 11 is EOL now... it's still in use quite a bit.

*Edit:*
Honestly... RHEL 5 is still supported 'til November. :/

cclerget (commented):
Could you revert changes done in plugins go.sum please ? as it's not related to this PR

internal/pkg/build/assemblers/sif.go
cclerget: Sounds reasonable, if you could add a note for `mksquashfs mem` directive in template ? to mention that it may not work with older version of mksquashfs and in this case it must be left blank/empty

internal/pkg/util/fs/squashfs/squashfs.go
```diff
@@ -36,3 +36,27 @@ func GetPath() (string, error) {
...
+
+func GetProcs() (uint, error) {
+	configFile := buildcfg.SINGULARITY_CONF_FILE
+	c, err := singularityconf.Parse(configFile)
+	if err != nil {
+		return 0, fmt.Errorf("unable to parse singularity.conf file: %s", err)
+	}
+	// proc is either "" or the string value in the conf file
+	proc := c.MksquashfsProcs
+
```
cclerget, line 48: Just for information, ideally it would be great to follow the logic commented here https://github.com/sylabs/singularity/issues/1228#issuecomment-575715043 for cpus. Not a change request just to let you know what changes will come after merging this PR

cclerget (changes requested):
Overall LGTM, just the `go.sum` cleanup with a note in template for `mksquashfs mem`

cclerget (approved):
Thank you @dmbeeson !

soichih:
It would be nice if we could set the max -mem via ENV options as well as singularity.conf.

nathanweeks:
> It would be nice if we could set the max -mem via ENV options as well as singularity.conf.

It would be beneficial if max CPU were configurable via environment variable as well. Then on HPC clusters that make singularity available only on compute nodes (not login nodes), requiring a user to submit a batch or interactive job to use Singularity, the max number of CPUs could be made equal to the number allocated to the batch/interactive job, e.g., in a Slurm prolog script (`export SINGULARITY_MKSQUASHFS_OPTS="-processors ${SLURM_JOB_CPUS_PER_NODE}"`)

dtrudg:
@soichih @nathanweeks - we'd likely welcome a patch for that.

soichih:
Is anyone still working on this? I am in desperate need to set mksquashfs mem based on max memory requested via slurm batch scheduler.. If the job is requested with `mem=8g` on #SBATCH directive, then I need to make singularity/mksquashfs to use up to 8gb. Should I try hacking around with the singularity code?

DrDaveD:
Hi Soichi, it would be good if you could complete #5805 in time for the 3.9.0 release if @kaczmarj doesn't do it.

kaczmarj:
@DrDaveD @soichih - i won't have enough bandwidth to complete #5805 -- sorry :/ but just wanted to mention that i really enjoy using singularity!!! it makes working on HPCs so much easier.

soichih:
I think I "finished" what @kaczmarj (see PR 6186) started. It's my very first PR for singularity (and very first with go-lang!) so I hope I did everything right.. I didn't bother with test script because the current master branch isn't passing checks either.. not sure what's going on there.

Issue #4827, opened 2019-12-04
Singularity ignores SINGULARITYENV_ for variables defined in host ENV


zyndagj:
### Version of Singularity:

```
$ singularity version
3.5.1-rc.2
```

### Expected behavior

- We expect `SINGULARITYENV_` to take precedence over host environment variables, so those values can be changed without affecting the host stability. While the `-e` flag does offer a way forward, removes ALL host environment variables which could be necessary for execution.
- We expect environments to be inherited the same between docker and singularity sources

#### 1. When environment variables are defined in a container recipe
  - singularity - variables can never be overridden
  - docker - variables can be overridden with
    - export VAR=new; singularity exec
    - `SINGULARITYENV_VAR=new singularity exec`
    - `export SINGULARITYENV_VAR=new; singularity exec`
       - **NOTE** - `SINGULARITYENV_VAR` is ignored if `VAR` is exported to the host ENV. The `-e` flag can ignore the host env, allowing `SINGULARITYENV_` to work, but not ideal for making small changes when host env contains things like filesystem paths.

#### 2. When environment variables are not defined in container recipe, but exist in host ENV
  - singularity - host variables can be overridden by
     - `export VAR=new; singularity exec`
     - `export SINGULARITYENV_VAR=new; singularity exec -e`
     - `SINGULARITYENV_VAR=new singularity exec -e`
  - docker - host variables can be overridden by
     - `export VAR=new; singularity exec`
     - `export SINGULARITYENV_VAR=new; singularity exec -e`
     - `SINGULARITYENV_VAR=new singularity exec -e`

### Steps to reproduce this behavior

1. Create `example.sif`
```
$ cat << 'EOF' > example.def
BootStrap: docker
From: ubuntu:xenial

%environment
    export LC_ALL=C
    export INTERNAL=1
EOF

singularity build example.sif example.def
```

2. Pull docker example
```
singularity pull docker://gzynda/sing_test:latest
```

3. Create `test.sh` script
```
#!/bin/bash

#!/bin/bash

function sing {
        singularity exec example.sif bash -c 'echo "Singularity |    | Internal: $INTERNAL External: $EXTERNAL"'
        singularity exec -e example.sif bash -c 'echo "Singularity | -e | Internal: $INTERNAL External: $EXTERNAL"'
}
function docker {
        singularity exec sing_test_latest.sif bash -c 'echo "Docker      |    | Internal: $INTERNAL External: $EXTERNAL"'
[...]

PR #5021, opened 2020-02-06
Fix SINGULARITYENV_ variables to always take precedence over host variables

cclerget:

Fix SINGULARITYENV_ variables to always take precedence over host variables because a SINGULARITYENV_ prefixed variable can be placed before a host variable which is supposed to be overriden (eg: SINGULARITYENV_FOO must take precedence over FOO) and this behavior leads to some inconsistencies.

- Related to #4827 (not a complete fix yet)

### DIFF
--- a/internal/pkg/util/env/clean.go
+++ b/internal/pkg/util/env/clean.go
@@ -15,76 +15,100 @@ import (
 const (
 	// DefaultPath defines default value for PATH environment variable.
 	DefaultPath = "/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin"
+	// SingularityPrefix defines the environment variable prefix SINGULARITY_.
+	SingularityPrefix = "SINGULARITY_"
+	// SingularityEnvPrefix defines the environment variable prefix SINGULARITYENV_.
+	SingularityEnvPrefix = "SINGULARITYENV_"
 )
 
 var alwaysPassKeys = map[string]struct{}{
-	"term":        {},
+	"TERM":        {},
 	"http_proxy":  {},
+	"HTTP_PROXY":  {},
 	"https_proxy": {},
+	"HTTPS_PROXY": {},
 	"no_proxy":    {},
+	"NO_PROXY":    {},
 	"all_proxy":   {},
+	"ALL_PROXY":   {},
 	"ftp_proxy":   {},
+	"FTP_PROXY":   {},
 }
 
-var alwaysOmitKeys = map[string]struct{}{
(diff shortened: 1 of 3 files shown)


internal/pkg/util/env/clean.go
```diff
@@ -15,76 +15,100 @@ import (
...
+	"LD_LIBRARY_PATH": true,
 }
 
 // SetContainerEnv cleans environment variables before running the container.
 func SetContainerEnv(g *generate.Generator, hostEnvs []string, cleanEnv bool, homeDest string) {
+	singEnvKeys := make(map[string]string)
+
+	// allow override with SINGULARITYENV_LANG
+	if cleanEnv {
+		g.AddProcessEnv("LANG", "C")
```
dtrudg, line 52: Is it always reasonable to force `LANG` to `C` when  a UTF-8 locale is the default almost everywhere now?

dtrudg (changes requested):
This looks reasonable to me.

One question about forcing `LANG` for cleanenv.... I'm really not sure we should force to `C` here.

Also, could you add a note about the behaviour change to the CHANGELOG.md? Although the current behaviour is definitely something that could be considered a bug this change might catch people out if they have some environment that hits it. I don't want to forget to call it out as a change.

internal/pkg/util/env/clean.go
cclerget: > UTF-8 locale is the default almost everywhere now

I don't know, it was always done like that, just moved the code before looping over environment variables to allow `SINGULARITYENV_LANG` to be passed in conjunction with `-e` like any other variables that a user would like to forward to container when using clean environment.

I think `C` was chosen because many images don't have locales (especially from docker) installed and C is de-facto the default here but in the meantime this is a bit weird to do that only with a cleaned environment and not all the time :

> Also, could you add a note about the behaviour change to the CHANGELOG.md?

Yep, do we integrate it for 3.5.3 ?

dtrudg: Okay - let's leave it as C here for now.

> Yep, do we integrate it for 3.5.3 ?

I think elsewhere I promised we wouldn't change behaviour in 3.5 and that we would look at addressing all the inconsistency including docker env var stuff: https://github.com/sylabs/singularity/issues/4827

I still think it'd be better to do this and make all behaviour changes for consistency in a `3.6.0` release if we can. Do you think that is reasonable?

cclerget: Sounds good

dtrudg (approved):

PR #5028, opened 2020-02-10
Add embedded shell interpreter

### LINKED ISSUE #4827: Singularity ignores SINGULARITYENV_ for variables defined in host ENV

### LINKED ISSUE #5040: Singularity and environment variables and 3.6.0 proposal
## To resume how things works actually

Singularity forward host user environment variables to container process in various inconsistent and not reproducible ways.

Variables always forwarded from host user environment to container process no matter if `--cleanenv` is specified or not:
* `term` (case insensitive)
* `http_proxy` (case insensitive)
* `https_proxy` (case insensitive)
* `no_proxy` (case insensitive)
* `ftp_proxy` (case insensitive)

Variables never forwarded from host user environment to container process:
* `path` (case insensitive)
* `ld_library_path` (case insensitive)

Default variables always forwarded but impossible to override from CLI:
* `HOME` by default uses home directory defined in user database or the source part from `--home` option
* `PATH` always set to `"/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin"`
* `LANG` set to "C" when `--cleanenv` is specified

Variables prefixed with `SINGULARITYENV_` see their prefix trimmed and are forwarded to container process no matter if `--cleanenv` is specified or not:
* `SINGULARITYENV_FOO` become `FOO` in container process :
  * **Important consideration**: before fix https://github.com/sylabs/singularity/pull/5021 those variables may or may not take precedence over host user environment variables when `--cleanenv` is not set leading to inconsistencies

There is few exceptions for `SINGULARITYENV_` variables :
  * `SINGULARITYENV_PREPEND_PATH` become `SING_USER_DEFINED_PREPEND_PATH`
  * `SINGULARITYENV_APPEND_PATH` become `SING_USER_DEFINED_APPEND_PATH`
  * `SINGULARITYENV_PATH` become `SING_USER_DEFINED_PATH`

All variables forwarded to container process goes through an environment setup before executing the real container process and it's done with sourced scripts located in `/.singularity.d/env` within the container image:
* `/.singularity.d/env/01-base.sh` (nothing done in this script)
* `/.singularity.d/env/10-docker2singularity.sh` (contains environment variables defined in Dockerfile if any)
* `/.singularity.d/env/90-environment.sh` (content of `%environment` section if build from a definition file)
* `/.singularity.d/env/91-environment.sh` (optional and depend of `$SINGULARITY_ENVIRONMENT` use)
* `/.singularity.d/env/94-appsbase.sh` (contains related app environment variables)
* `/.singularity.d/env/95-apps.sh` (append bin/lib apps path in `PATH` and `LD_LIBRARY_PATH`)
  * `/scif/apps/${SINGULARITY_APPNAME}/scif/env/01-base.sh`
  * `/scif/apps/${SINGULARITY_APPNAME}/scif/env/90-environment.sh` (content of app `%appenv` section)
* `/.singularity.d/env/99-base.sh` (adds `/singularity.d/libs` to `LD_LIBRARY_PATH` and force `PS1="Singularity> "`)
* `/.singularity.d/env/99-runtimevars.sh` (handle special `SING_USER_DEFINED_` path variables)

**Important consideration**: until Singularity 3.1.0 all images bootstrapped from docker translated `ENV KEY val` Dockerfile lines to the export form `export KEY=VAL`, since 3.1.0 they are translated to `export KEY=${KEY:-"VAL"}` to be overridable, those exported variables are located in container image at:
  * `/.singularity.d/env/10-docker2singularity.sh` for v3
  * `/.singularity.d/env/10-docker.sh` for v2

With all the above in mind let's take an example with an image run from `docker://golang:1.13`, the
the Dockerfile of this image has defined environment variables:

```
GOLANG_VERSION=1.14rc1
GOPATH=/go
PATH=/go/bin:/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

When running with Singularity prior to 3.1.0:

```
$ GOPATH=/fake singularity exec docker://golang:1.13 go env | grep GOPATH
GOPATH="/go"
$ SINGULARITYENV_GOPATH=/fake singularity exec --cleanenv docker://golang:1.13 go env | grep GOPATH
GOPATH="/go"
```

Because it's always forced by `/.singularity.d/env/10-docker2singularity.sh` containing `export GOPATH=/go`

When running with Singularity >= 3.1.0:

```
$ GOPATH=/fake singularity exec docker://golang:1.13 go env | grep GOPATH
GOPATH="/fake"
$ SINGULARITYENV_GOPATH=/fake singularity exec --cleanenv docker://golang:1.13 go env | grep GOPATH
GOPATH="/fake"
```

Because `/.singularity.d/env/10-docker2singularity.sh` contains `export GOPATH=${GOPATH:-"/go"}` allowing to override `GOPATH` from the forwarded variable.

Now if we build the image with the following definition file:

```
cat > /tmp/docker.def <<EOF
bootstrap: docker
from: docker://golang:1.13

%environment
	# override docker GOPATH and set it to a writable location
	export GOPATH="/tmp/go"

EOF
```

The container image now contains a `/.singularity.d/env/90-environment.sh` with the content:

```
export GOPATH="/tmp/go"
```

Obviously with any Singularity version `GOPATH` is not overridable anymore

```
$ singularity build -f /tmp/go.sif /tmp/docker.def
$ GOPATH=/fake singularity exec /tmp/go.sif go env | grep GOPATH
GOPATH="/tmp/go"
$ SINGULARITYENV_GOPATH=/fake singularity exec --cleanenv /tmp/go.sif go env | grep GOPATH
GOPATH="/tmp/go"
```

## Proposal for consistency and rules in 3.6.0

No changes regarding forwarding of user host environment variables

1. Only variables prefixed with `SINGULARITYENV_` are passed to:
  * `/.singularity.d/env/10-docker2singularity.sh` or `/.singularity.d/env/10-docker.sh`
  * `/.singularity.d/env/90-environment.sh`
  * `/.singularity.d/env/91-environment.sh`
2. Add `--env` option to pass environment variables by simply prefixing them with `SINGULARITYENV_`
  * and allow to pass environment variable as literal like this:
    `--env PYTHONPATH=/python/path:\${PYTHONPATH}`

Rules:
1. Docker inherited environment variables are always overridable (stay aligned with docker behaviour)
2. Environment variables defined in `%environment` section are overridable only if the image author allow it:
 * Example for an overridable variable:
   ```
   %environment
       # if PYTHONPATH is already set use it otherwise use
       # default value "/usr/local/python3"
       export PYTHONPATH=${PYTHONPATH:-"/usr/local/python3"}
   ```
 * Example for not overridable variable:
   ```
   %environment
       export PYTHONPATH="/usr/local/python3"
   ```

cclerget:

- Add shell interpreter able to interpret /.singularity.d/env files and execute action scripts without the need for `/bin/sh` in container image. 
- Add `--env` option to specify environment variables to pass for container environment setup:
  - allow literal for evaluation during container environment setup:
    - `--env PATH=/a/bind/path:\$PATH` where `$PATH` is replaced by `PATH` during container setup allowing to append/prepend value to the final path environment variables
    - `--env PATH=/a/bind/path:$PATH` take path from current environment
- Add `--env-file` option to specify environment variables from file:
  - allow literal for evaluation during container environment setup
  - command execution and file creation/modification are disabled during variable evaluation, but shell builtins remains available

- Fixes #5040 
- Fixes #4827

### DIFF
--- a/cmd/internal/cli/action_flags.go
+++ b/cmd/internal/cli/action_flags.go
@@ -13,25 +13,27 @@ import (
 
 // actionflags.go contains flag variables for action-like commands to draw from
 var (
-	AppName         string
-	BindPaths       []string
-	HomePath        string
-	OverlayPath     []string
-	ScratchPath     []string
-	WorkdirPath     string
-	PwdPath         string
-	ShellPath       string
-	Hostname        string
-	Network         string
-	NetworkArgs     []string
-	DNS             string
-	Security        []string
-	CgroupsPath     string
-	VMRAM           string
-	VMCPU           string
-	VMIP            string
-	ContainLibsPath []string
-	FuseMount       []string
+	AppName            string
+	BindPaths          []string
+	HomePath           string
+	OverlayPath        []string
+	ScratchPath        []string
+	WorkdirPath        string
+	PwdPath            string
+	ShellPath          string
+	Hostname           string
+	Network            string
+	NetworkArgs        []string
+	DNS                string
+	Security           []string
+	CgroupsPath        string
+	VMRAM              string
+	VMCPU              string
+	VMIP               string
+	ContainLibsPath    []string
+	FuseMount          []string
+	SingularityEnv     []string
+	SingularityEnvFile string
 
 	IsBoot          bool
 	IsFakeroot      bool
@@ -613,6 +615,27 @@ var actionAllowSetuidFlag = cmdline.Flag{
 	ExcludedOS:   []string{cmdline.Darwin},
 }
 
+// --env
+var actionEnvFlag = cmdline.Flag{
+	ID:           "actionEnvFlag",
+	Value:        &SingularityEnv,
+	DefaultValue: []string{},
+	Name:         "env",
+	Usage:        "pass environment variable to contained process",
+	ExcludedOS:   []string{cmdline.Darwin},
+}
+
+// --env-file
+var actionEnvFileFlag = cmdline.Flag{
+	ID:           "actionEnvFileFlag",
+	Value:        &SingularityEnvFile,
+	DefaultValue: "",
+	Name:         "env-file",
+	Usage:        "pass environment variables from file to contained process",
+	EnvKeys:      []string{"ENV_FILE"},
+	ExcludedOS:   []string{cmdline.Darwin},
+}
+
 func init() {
 	addCmdInit(func(cmdManager *cmdline.CommandManager) {
 		cmdManager.RegisterCmd(ExecCmd)
@@ -686,5 +709,7 @@ func init() {
 		cmdManager.RegisterFlagForCmd(&dockerLoginFlag, actionsInstanceCmd...)
 		cmdManager.RegisterFlagForCmd(&dockerPasswordFlag, actionsInstanceCmd...)
 		cmdManager.RegisterFlagForCmd(&dockerUsernameFlag, actionsInstanceCmd...)
+		cmdManager.RegisterFlagForCmd(&actionEnvFlag, actionsInstanceCmd...)
+		cmdManager.RegisterFlagForCmd(&actionEnvFileFlag, actionsInstanceCmd...)
 	})
 }
--- a/internal/pkg/util/shell/interpreter/interpreter.go
+++ b/internal/pkg/util/shell/interpreter/interpreter.go
@@ -1,0 +1,301 @@
+// Copyright (c) 2020, Sylabs, Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license.  Please
+// consult LICENSE.md file distributed with the sources of this project regarding
+// your rights to use or distribute this software.
+
+package interpreter
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"sort"
+	"strings"
+	"syscall"
+	"time"
+
+	"mvdan.cc/sh/v3/expand"
+	"mvdan.cc/sh/v3/interp"
+	"mvdan.cc/sh/v3/syntax"
+)
+
+// ShellBuiltin defines function prototype for shell interpreter builtin registration.
+type ShellBuiltin func(ctx context.Context, args []string) error
+
+// OpenHandler defines function prototype for shell interpreter file handler registration.
+type OpenHandler func(path string, flag int, perm os.FileMode) (io.ReadWriteCloser, error)
+
+// Shell defines the shell interpreter.
+type Shell struct {
+	shellBuiltins map[string]ShellBuiltin
+	openHandlers  map[string]OpenHandler
+	name          string
+	status        uint8
+	reader        io.Reader
+	runner        *interp.Runner
+}
+
+// execTimeout defines the execution timeout for commands executed by the
+// shell interpreter.
+var execTimeout = 5 * time.Second
+
+// defaultExecHandler is the default command execution handler if there is
+// no registered shell builtin.
+func defaultExecHandler(ctx context.Context, args []string) error {
+	hc := interp.HandlerCtx(ctx)
+	path, err := interp.LookPath(hc.Env, args[0])
+	if err != nil {
+		fmt.Fprintln(hc.Stderr, err)
+		return interp.NewExitStatus(127)
+	}
+
+	ectx, cancel := context.WithTimeout(ctx, execTimeout)
+	defer cancel()
+
+	cmd := exec.CommandContext(ectx, path, args[1:]...)
+	cmd.Env = append([]string{"PWD=" + hc.Dir}, GetEnv(hc)...)
+	cmd.Dir = hc.Dir
+	cmd.Stdin = hc.Stdin
+	cmd.Stdout = hc.Stdout
+	cmd.Stderr = hc.Stderr
+
+	err = cmd.Run()
+
+	switch x := err.(type) {
+	case *exec.ExitError:
+		if status, ok := x.Sys().(syscall.WaitStatus); ok {
+			if status.Signaled() && ectx.Err() != nil {
+				c := strings.Join(args, " ")
+				return fmt.Errorf("command %q was killed after %s timeout", c, execTimeout)
+			}
+			return interp.NewExitStatus(uint8(status.ExitStatus()))
+		}
+		return interp.NewExitStatus(1)
+	case *exec.Error:
+		c := strings.Join(args, " ")
+		return fmt.Errorf("command %q execution failed: %s", c, err)
+	}
+
+	return err
+}
+
+// GetEnv returns an the list of all exported environment variables within
+// the context of the shell interpreter.
+func GetEnv(hc interp.HandlerContext) []string {
+	envMap := make(map[string]expand.Variable)
+
+	// use of a map to remove duplicated variable, the latest wins
+	hc.Env.Each(func(name string, vr expand.Variable) bool {
+		envMap[name] = vr
+		return true
+	})
+
+	environ := make([]string, 0, len(envMap))
+	for k, v := range envMap {
+		if v.Exported && v.Kind == expand.String {
+			environ = append(environ, k+"="+v.Str)
+		}
+	}
+
+	sort.Strings(environ)
+
+	return environ
+}
+
+// New returns a shell interpreter instance.
+func New(r io.Reader, name string, args []string, envs []string, runnerOptions ...interp.RunnerOption) (s *Shell, err error) {
+	if r == nil {
+		return nil, fmt.Errorf("nil reader")
+	}
+
+	s = &Shell{
+		reader: r,
+		name:   name,
+	}
+
+	opts := []interp.RunnerOption{
+		interp.StdIO(os.Stdin, os.Stdout, os.Stderr),
+		interp.ExecHandler(s.internalExecHandler()),
+		interp.OpenHandler(s.internalOpenHandler()),
+		interp.Params("--"),
+		interp.Env(expand.ListEnviron(envs...)),
+	}
+	opts = append(opts, runnerOptions...)
+	s.runner, err = interp.New(opts...)
+
+	if err != nil {
+		return nil, fmt.Errorf("while creating shell interpreter: %s", err)
+	}
+
+	s.runner.Dir, err = os.Getwd()
+	if err != nil {
+		s.runner.Dir = "/"
+	}
+
+	s.runner.Params = append(s.runner.Params, args...)
+
+	return s, err
+}
+
+// internalOpenHandler returns an ExecHandlerFunc used by default.
+func (s *Shell) internalExecHandler() interp.ExecHandlerFunc {
+	return func(ctx context.Context, args []string) error {
+		if s.runner.Exited() {
+			// special path for exec builtin keyword
+			if builtin, ok := s.shellBuiltins["exec"]; ok {
+				return builtin(ctx, args)
+			}
+		} else if builtin, ok := s.shellBuiltins[args[0]]; ok {
+			return builtin(ctx, args[1:])
+		}
+		return defaultExecHandler(ctx, args)
+	}
+}
+
+// internalOpenHandler returns an OpenHandlerFunc used by default.
+func (s *Shell) internalOpenHandler() interp.OpenHandlerFunc {
+	return func(ctx context.Context, path string, flag int, perm os.FileMode) (io.ReadWriteCloser, error) {
+		mc := interp.HandlerCtx(ctx)
+		if !filepath.IsAbs(path) {
+			path = filepath.Join(mc.Dir, path)
+		}
+		if handler, ok := s.openHandlers[path]; ok {
+			return handler(path, flag, perm)
+		}
+		return os.OpenFile(path, flag, perm)
+	}
+}
+
+// RegisterShellBuiltin registers a shell interpreter builtin.
+func (s *Shell) RegisterShellBuiltin(name string, builtin ShellBuiltin) {
+	if s.shellBuiltins == nil {
+		s.shellBuiltins = make(map[string]ShellBuiltin)
+	}
+	s.shellBuiltins[name] = builtin
+}
+
+// RegisterOpenHandler registers a shell interpreter file handler.
+func (s *Shell) RegisterOpenHandler(path string, handler OpenHandler) {
+	if s.openHandlers == nil {
+		s.openHandlers = make(map[string]OpenHandler)
+	}
+	s.openHandlers[path] = handler
+}
+
+// LookPath returns the absolute path for the command passed in argument
+// within the context of the shell interpreter.
+func (s *Shell) LookPath(ctx context.Context, cmd string) (string, error) {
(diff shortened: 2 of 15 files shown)


vsoch (commented):
Okay, reviewing the points mentioned in #5040. First I installed the branch from a fresh clone of your base:
```
git checkout master-shellvm 
./mconfig 
./mconfig 
cd builddir/
make
sudo make install
```

Okay here is a testing recipe

```
Bootstrap: docker
From: busybox

%environment
   export NOCHANGEME=22
   export CHANGEME="${CHANGEME:-22}"
```
Building the image and looking at the variables, I expect them both to be 22.

```
sudo singularity build busybox.sif Singularity
$ singularity exec --containall busybox.sif env | grep CHANGE
CHANGEME=22
NOCHANGEME=22
```

Testing setting SINGULARITYENV_ for each - note that I tested each of the commands below with `--containall` and/or `--cleanenv` and got the same result.

```bash
$ SINGULARITYENV_CHANGEME=44 singularity exec busybox.sif env | grep CHANGE
NOCHANGEME=22
CHANGEME=44
$ SINGULARITYENV_NOCHANGEME=44 singularity exec busybox.sif env | grep CHANGE
NOCHANGEME=44
CHANGEME=22
```
We had talked about how a variable would only be changeable if the user allowed it (e.g., the CHANGEME variable in the recipe) but it looks like this isn't required - any use of SINGULARITYENV, with or without contain all, will override (NOCHANGEME above is changed!), and the user that designed the recipe didn't need to make it a special variable to do this. This is correct? It definitely makes sense, but it's different than what was discussed.

Now let's test using --env or --env-file, same as above. Note that the below is also consistent for --containall

```bash
$ singularity exec --env CHANGEME=44 busybox.sif env | grep CHANGE
CHANGEME=44
NOCHANGEME=22
$ singularity exec --env NOCHANGEME=44 busybox.sif env | grep CHANGE
NOCHANGEME=44
CHANGEME=22
```

One thing I noticed is that the order of the env command is different depending on using or not using a flag like --containall. Is this the env command sorting, or something Singularity is doing to change the order?

```bash
$ SINGULARITYENV_CHANGEME=44 singularity exec --containall busybox.sif env | grep CHANGE
CHANGEME=44
NOCHANGEME=22
$ SINGULARITYENV_CHANGEME=44 singularity exec busybox.sif env | grep CHANGE
NOCHANGEME=22
CHANGEME=44
```

It might be the case that a user doesn't add the = for the envar, and this would result in a weird error:

```bash
$ singularity exec --env NOCHANGEME 44 busybox.sif env | grep CHANGE
FATAL:   could not open image /tmp/44: failed to retrieve path for /tmp/44: lstat /tmp/44: no such file or directory
```

I think it's okay to not allow it, but is there a way to catch it more elegantly?

Now here is my little --envfile - note that I am testing using an export and not using.
```
export CHANGEME=55
NOCHANGEME=66
```
It seems to honor both! Note that this also is the same with --cleanenv. And I again thinks that makes sense - if the user explicitly is defining a file, --env flag, or SINGULARITYENV_ they want to override something.

And I can't do any changes without the prefix, flag, etc.

```bash
$ CHANGEME=44 singularity exec --cleanenv busybox.sif env | grep CHANGE
NOCHANGEME=22
CHANGEME=22
$ CHANGEME=44 singularity exec busybox.sif env | grep CHANGE
NOCHANGEME=22
CHANGEME=22
```

Here is an interesting use case - I can take the same env-file and bind directly to the env folder, in which case the exported variable is honored, non-exported is not.

```bash
$ singularity exec --cleanenv --bind env-file:/.singularity.d/env/100-envfile.sh busybox.sif env | grep CHANGE
CHANGEME=55
NOCHANGEME=22
```
I'm not sure when that might be used, but wanted to point it out. It looks like I can easily edit the PATH like that (and mess it up!), and it also works using `SINGULARITY_ENV` and --env (and I'm aware of the prepend variables, just wanted to test it out, I think a lot of folks will want to edit it anyway).

So - I can't test the more complex use cases, but superficially this looks good!

cclerget:
> any use of SINGULARITYENV, with or without contain all, will override (NOCHANGEME above is changed!), and the user that designed the recipe didn't need to make it a special variable to do this. This is correct? It definitely makes sense, but it's different than what was discussed.

It's correct because finally decided it was better to use an explicit syntax for this case, so if image authors want to avoid a variable override, they could use the portable syntax :

```
%environment
     readonly DONT_TOUCH="default"
     export DONT_TOUCH
```

This way if user set `--env DONT_TOUCH="really"` a message telling `DONT_TOUCH` is a readonly variable instead of silently ignoring the value

Also note that the following won't work either :

```
%environment
    unset DONT_TOUCH
    export DONT_TOUCH="default"
```

As `DONT_TOUCH` is not defined because environment variables has not been passed yet, the `unset` call is a no-op.

> Is this the env command sorting, or something Singularity is doing to change the order?

That's something to be fixed as a golang map is used for that the order is arbitrary

> It might be the case that a user doesn't add the = for the envar, and this would result in a weird error:
> 
>  $ singularity exec --env NOCHANGEME 44 busybox.sif env | grep CHANGE
> FATAL:   could not open image /tmp/44: failed to retrieve path for /tmp/44: lstat /tmp/44: no such file or directory
> 
>  I think it's okay to not allow it, but is there a way to catch it more elegantly?

I'm afraid not because 44 is considered as a the first positional argument so it takes it as the image argument and `busybox.sif env` as the exec arguments. I will add check for bad env but this case is definitively outside of control

> Now here is my little --envfile - note that I am testing using an export and not using.
> 
> export CHANGEME=55
> NOCHANGEME=66
> 
> It seems to honor both! Note that this also is the same with --cleanenv. And I again thinks that makes sense - if the user explicitly is defining a file, --env flag, or SINGULARITYENV_ they want to override something.

Yes I wanted to simplify thing with `--env-file` to be aligned with `--env` so all variables in the env file are automatically exported (`allexport` on by default)

> Here is an interesting use case - I can take the same env-file and bind directly to the env folder, in which case the exported variable is honored, non-exported is not.
> 
> $ singularity exec --cleanenv --bind env-file:/.singularity.d/env/100-envfile.sh busybox.sif env | grep CHANGE
> CHANGEME=55
> NOCHANGEME=22
> 
> I'm not sure when that might be used, but wanted to point it out. It looks like I can easily edit the PATH like that (and mess it up!), and it also works using SINGULARITY_ENV and --env (and I'm aware of the prepend variables, just wanted to test it out, I think a lot of folks will want to edit it anyway).

With `--env-file` this use case is not relevant anymore but if someone want to "hack" environment source process he can, of course we could deny bind at this location but is it really required ?

Additionally `--env-file` is evaluated in host user environment, so expect different behavior between `--env-file` and bind in container.

The shell interpreter allow to do interesting things which should address many issues especially with all kind of `*PATH*` variables, taking the `PYTHONPATH` example with someone wanting to add module from the home directory inside container with modules installed at a specific location in the image :

```
BIND_DEST=/python/modules
singularity exec -B $BIND_DEST --env PYTHONPATH=$BIND_TEST:\$PYTHONPATH ...
```

The literal form `\$` defer evaluation inside container instead of the host environment which means we can actually get rid of `PREPEND_PATH` and `APPEND_PATH` by syntax :
`PATH="/mypath:\$PATH"` or `PATH="\$PATH:/mypath"`

cmd/internal/cli/action_flags.go
```diff
@@ -613,6 +615,27 @@ var actionAllowSetuidFlag = cmdline.Flag{
 	ExcludedOS:   []string{cmdline.Darwin},
 }
 
+// --env
+var actionEnvFlag = cmdline.Flag{
+	ID:           "actionEnvFlag",
+	Value:        &SingularityEnv,
+	DefaultValue: []string{},
+	Name:         "env",
+	Usage:        "pass environment variable to container process",
```
dtrudg:
```suggestion
	Usage:        "pass environment variable to contained process",
```

cmd/internal/cli/action_flags.go
```diff
@@ -613,6 +615,27 @@ var actionAllowSetuidFlag = cmdline.Flag{
...
+	ExcludedOS:   []string{cmdline.Darwin},
+}
+
+// --env-file
+var actionEnvFileFlag = cmdline.Flag{
+	ID:           "actionEnvFileFlag",
+	Value:        &SingularityEnvFile,
+	DefaultValue: "",
+	Name:         "env-file",
+	Usage:        "pass environment variable from file to container process",
```
dtrudg:
```suggestion
	Usage:        "pass environment variables from file to contained process",
```

internal/pkg/util/shell/interpreter/interpreter.go
```diff
@@ -0,0 +1,301 @@
...
+func (e nonExportedEnv) Each(fn func(name string, vr expand.Variable) bool) {
+	for name, vr := range e.envs {
+		if !fn(name, vr) {
+			return
+		}
+	}
+}
+
+// EvaluateEnv evaluates the environment variable script and returns
+// the list of variables set in the script. Command execution is disabled
```
dtrudg, line 259: So this will break any container definition files where people rely on old behavior where `%environment` *can* execute things inside the container... 

```
%environment
	export STARTED=$(date)
```

.. not opposed to that really, but do we know how many people use it?

dtrudg (approved):
I think this is a nice place to have this merged, so that we can push forward discussion docs, etc.

My main concern is how many people are using execution in the container context within `%environment` parts of their definition files. We should try and get some input around that.

internal/pkg/util/shell/interpreter/interpreter.go
cclerget: `EvaluateEnv` is used only for `--env-file`, I choose to disable command execution, sourcing and redirection to reduce potential abuse, but everything in `%environment` section is running as usual within the shell interpreter without any restriction

Issue #5002, opened 2020-01-31
--nv --rocm won't find correct ldconfig when run from Nix userspace


mboisson:
### Version of Singularity:

What version of Singularity are you using? Run:

```
$ singularity --version
singularity version 3.5.2-1.el7
```

```
$ singularity exec --nv pytorch.simg bash
WARNING: Unable to capture nv bind points: could not execute ldconfig: exit status 1
``` 
### Expected behavior

Not an error with ldconfig.

### Actual behavior

Error with ldconfig, because it takes the wrong ldconfig. 

### Explanation
I believe this is a regression from 
https://github.com/sylabs/singularity/issues/2445

when the util/nvidia/paths.go became 
https://github.com/sylabs/singularity/blob/master/pkg/util/gpu/paths.go

mboisson:
For comparison, this works fine : 
```
$ PATH=/usr/sbin:$PATH singularity exec --nv pytorch.simg bash
Singularity> exit
``` 

So, having a different command called `ldconfig`, somewhere in the `PATH`, produces the error.

dtrudg:
Can you let us know what type of environment is being used that provides its own `ldconfig` - to better understand what workflow this is arising in? 

Singularity did begin sanitizing `PATH`, in `3.1` this was further adjusted in `3.2` via https://github.com/sylabs/singularity/commit/5646ae764dddce5a1493de584884f70653f71bf2 as blanket sanitizing of the PATH caused issues for the plugin support. The GPU code has also been modified since then.

Because we bind in CUDA and ROCm related binaries that are searched on `PATH` we can't use a sanitized `PATH` for all of the GPU bind search. The CUDA/ROCm binaries may be in non-standard locations or provided by modules in some setups, so we need to find them in the user `PATH`, not a minimal `PATH`.

We would have to santize `PATH` for only the `ldconfig` call if we want address this issue.

mboisson:
Hi,
We provide an OS-independent Linux layer provided by Nix OS. https://github.com/ComputeCanada/nixpkgs

This effectively means that the user-space of our system runs Nix OS, in which ldconfig does not quite work, and is not usually required to work, because NixOS sets RPATH for all binaries that are compiled within. 

```
$ which ldconfig
/cvmfs/soft.computecanada.ca/nix/var/nix/profiles/16.09/bin/ldconfig

$ ldconfig -p
[...]

PR #5046, opened 2020-02-15
Handle non-standard awk location for genmod.awk (Nix/Guix)

dtrudg:

:dragon: Here be dragons! :fire: 

While looking at #5002 I decided to try and also make sure we can build Singularity on nix & guix as base OS - as well as trying to solve the profile vs host `ldconfig` issue. Nix and Guix only provide `/bin/sh` and `/usr/bin/env` and *not* `/usr/bin/awk` that we need in `mconfig`'s `genmod.awk`. Unfortunately `genmod.awk` takes multiple args so portable use of `/usr/bin/env` is not possible....

---

Under nix, guix, and maybe some other environments there is no
`/usr/bin/awk` - it has to be found on `PATH`.

The GNU `-S` option to `/usr/bin/env` allows use of shebangs with multiple
arguments passed to the interpreter (as is the case for genmod.awk),
but this is non POSIX so won't work under busybox etc.

To support bare POSIX environments as well as nix type environments the
`genmod.awk` can be made into a polygot program where some valid `sh`
calls the script again using `awk` on `PATH`.

Ref: https://unix.stackexchange.com/a/149801

### DIFF
--- a/makeit/genmod.awk
+++ b/makeit/genmod.awk
@@ -1,4 +1,7 @@
-#!/usr/bin/awk -f
+#!/bin/sh
+true + /; exec -a "$0" awk -f "$0" -- "$@"; / {}
+# awk script starts here
+
 # Copyright (c) 2015-2018, Yannick Cote <yhcote@gmail.com>. All rights reserved.
 # Use of this source code is governed by a BSD-style license that can be found
 # in the LICENSE file.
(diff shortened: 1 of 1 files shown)


dtrudg:
Well - this crazyness works on Linux across various shells, awks, but seems to be impossible if we have to take into account the OSX version of `awk`.

Mac is more common the NixOS or GuixSD... so I am going to close this. `genmod.awk` needs a patch in any packaging / build process for those distros.

PR #5055, opened 2020-02-21
Provide a way to determine plugin installation path

cclerget:

- Add new symbol `PluginRootDir` variable for plugin API allowing plugins to retrieve plugin root directory in order to concatenate with the plugin manifest name to determine the current plugin installation directory.
- Remove --name option from `singularity plugin install` for now, at best if it needs to be changed this should be done at compilation time but not at installation because manifest name differ from the one used by Singularity internals and could lead to confusion during debugging by example.

### DIFF
--- a/cmd/internal/cli/plugin_install_linux.go
+++ b/cmd/internal/cli/plugin_install_linux.go
@@ -10,34 +10,16 @@ import (
 	"github.com/sylabs/singularity/docs"
 	"github.com/sylabs/singularity/internal/app/singularity"
 	"github.com/sylabs/singularity/internal/pkg/sylog"
-	"github.com/sylabs/singularity/pkg/cmdline"
 )
-
-// -n|--name
-var pluginName string
-var pluginInstallNameFlag = cmdline.Flag{
-	ID:           "pluginInstallNameFlag",
-	Value:        &pluginName,
-	DefaultValue: "",
-	Name:         "name",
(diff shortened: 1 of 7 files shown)


dtrudg (approved):
This seems fair to me :+1:

If we are deriving plugin names only from the manifest and not providing an override, should we establish a namespaced convention to prevent confusion about plugins with similar function coming from different authors?

e.g. we should have `io.sylabs.syslog` instead of `syslog`

Or... we could use the go URI style naming.. but that's a bit more verbose and maybe implies public access to a source that might be private.

cclerget:
> Or... we could use the go URI style naming.. but that's a bit more verbose and maybe implies public access to a source that might be private.

My preference goes for go URI style with something like `github.com/author/myplugin`, ideally the same name as `module` line in go.mod (it could be even not required anymore as we parse `go.mod` at compilation so we could extract here maybe ?). I didn't capture issue with private repo ?

dtrudg:
> > Or... we could use the go URI style naming.. but that's a bit more verbose and maybe implies public access to a source that might be private.
> 
> My preference goes for go URI style with something like `github.com/author/myplugin`, ideally the same name as `module` line in go.mod (it could be even not required anymore as we parse `go.mod` at compilation so we could extract here maybe ?). I didn't capture issue with private repo ?

Sounds good to me. As long as we have a plan.

PR #5070, opened 2020-02-27
Image driver fixes

cclerget:

- Fix image driver start where a driver is started too early before the session is created.
- Rework privileges handling in starter, RPC and master process
now keep the original user/group ID and keep only required capabilities,
it's necessary because when we are using FUSE mount points the access
control in the kernel check if the real uid/gid, effective uid/gid and
saved uid/gid are identical to user_id/group_id FUSE mount options, so
we can't use setresuid/setresgid with setuid workflow because the saved
uid/gid for the master process is 0 and the effective uid is 0 for the
RPC process.
- Fix condition by adding image driver check on image conversion when using
unprivileged workflow.

 - Fixes #

### DIFF
--- a/cmd/starter/c/starter.c
+++ b/cmd/starter/c/starter.c
@@ -56,6 +56,8 @@
 #define SELF_MNT_NS     "/proc/self/ns/mnt"
 #define SELF_CGROUP_NS  "/proc/self/ns/cgroup"
 
+#define capflag(x)  (1ULL << x)
+
 /* current starter configuration */
 struct starterConfig *sconfig;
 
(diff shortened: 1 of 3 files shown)


cmd/starter/c/starter.c
```diff
@@ -286,6 +292,52 @@ static int apply_container_privileges(struct privileges *privileges) {
...
+    if ( priv == NULL ) {
+        fatalf("could not allocate memory\n");
+    }
+
+    memset(priv, 0, sizeof(struct privileges));
+
+    priv->capabilities.effective |= capflag(CAP_SETGID);
+    priv->capabilities.effective |= capflag(CAP_SETUID);
+
+    priv->capabilities.permitted = 0xffffffffff;
```
dtrudg, line 330: What's the a reason we are masking like this rather than setting equal to effective?

Is there a smaller set of capabilities we would need later on?

cmd/starter/c/starter.c
```diff
@@ -1417,8 +1472,8 @@ __attribute__((constructor)) static void init(void) {
              * escalation from master process, because container network requires
              * privileges
              */
-            if ( sconfig->starter.isSuid  ) {
-                priv_drop(false);
+            if ( sconfig->starter.isSuid && geteuid() == 0 ) {
```
dtrudg, line 1475: Please could you update the comment above on 1471 to reflect that we are not dropping priv temporarily in the same way.. but instead we are setting capabilities instead

cmd/starter/c/starter.c
cclerget: We may need arbitrary capabilities (CNI), so we just keep setuid/setgid as effective and the rest will be raised automatically during an execve as root user by example, if we set all capabilities as effective spawned process couldn't raise those capabilities

cmd/starter/c/starter.c
cclerget: Will do thanks !

dtrudg (approved):

PR #5085, opened 2020-03-03
Use base metapackage for arch bootstrap

### LINKED ISSUE #5077: Arch boostrap module failure
### Version of Singularity:
```
singularity version bf6bdb5
```

### Expected behavior
It actually builds the container.

### Actual behavior
```console
$ sudo singularity build -s ./rootfs Singularity.def
WARNING: 'nodev' mount option set on /tmp, it could be a source of failure during build process
INFO:    Starting build...
FATAL:   While performing build: conveyor failed to get: while generating the installation list: exit status 1:
```

Content of `Singularity.def`:
```
Bootstrap: arch
```
Just `sudo singularity build anything.sif Singularity.def`.

Up-to-date Arch Linux

Via pacman

Am I doing something wrong?

dtrudg:

We used to pull the 'base' group for arch, and then filter out some
packages listed in it, installing the remainder via a pacstrap
bootstrap. This no longer works as the 'base' group has been removed
and there is now a 'base' metapackage:

https://www.archlinux.org/news/base-group-replaced-by-mandatory-base-package-manual-intervention-required/

Because this is a single package install for pacstrap there isn't a
nice way to filter out the things that we weren't installing into the
container with the previous method. This gives larger container
images, but comparable to the dockerhub official arch images.

The alternative would be to install things listed as dependencies of
'base' by keeping track of that manually here, or writing complex
query / exclude code. I feel these are probably a bit too brittle, and
risk breakage later on while simply using the `base` metapackage
should be okay. Have left the package list as a string slice in case
we want to go back to a list of packages to install in future.

Fixes: #5077

### DIFF
--- a/internal/pkg/build/sources/conveyorPacker_arch.go
+++ b/internal/pkg/build/sources/conveyorPacker_arch.go
@@ -6,8 +6,6 @@
 package sources
 
 import (
-	"bufio"
-	"bytes"
 	"context"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


dtrudg:
Note that e.g. `kernel` is an /optional/ dependency of base we won't install - so it's not quite like installing the base group without filtering. 

There isn't a whole lot of stuff that isn't relevant in a container:

https://www.archlinux.org/packages/core/any/base/

ikaneshiro (approved):
Lgtm

PR #5086, opened 2020-03-03
Add --no-hostfs actions flag

### LINKED ISSUE #5081: Ability to disable "mount hostfs" in command line
### Version of Singularity:

```
$ singularity version
3.5.3-1.1.fc31
```

### Expected behavior

On a machine with multiple users, some users need mount hostfs on, some need it off. 

Since a security policy may require it off, but probably won't require it on, please implement --no-hostfs as a command line option.

### Actual behavior

This can only be set in /etc/singularity/singularity.conf which means disturbing the sysadmin.

```
$ cat /etc/os-release
NAME=Fedora
VERSION="31 (Thirty One)"
```

From the Fedora repository.

dtrudg:
## Description of the Pull Request (PR)
    
Add a --no-hostfs so that if the global config specifies "mount hostfs
= yes" it is still possible to run a container without all the host
filesystems mounted.
    
An example of how this is useful - for containers that have
directories with generic names like `/data` inside the
container. `/data` is a mount point on my host so it's useful to be
able to turn off `mount hostfs` for these containers.

 -   Closes: #5081

### DIFF
--- a/cmd/internal/cli/action_flags.go
+++ b/cmd/internal/cli/action_flags.go
@@ -45,6 +45,7 @@ var (
 	Nvidia          bool
 	Rocm            bool
 	NoHome          bool
+	NoHostfs        bool
 	NoInit          bool
 	NoNvidia        bool
 	NoRocm          bool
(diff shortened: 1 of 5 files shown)


dtrudg:
@cclerget - I'm wondering how to test this nicely in the e2e structure we have? Is there any place that you think is natural to put a test which...

- bind mounts an extra dir on the host
- sets `mount hostfs = yes`
- checks execution with and without `--no-hostfs` 

? - I'm not sure whether this is something I should write into e2e/actions? - I guess it should be a stand-alone test function there?

cclerget:
> I'm wondering how to test this nicely in the e2e structure we have? Is there any place that you think is natural to put a test which...

`e2e/config` could be a good place.

On the PR change itself, I'm not a huge fan of adding more `--no-xxxx` option, if we continue to do that for `hostfs` (like we did with `--no-home`) why we wouldn't do also for all mount configuration directives ?
We could add a `--mount` option which could take as value sys, proc, home, hostfs, tmp, dev, devpts and allow to disable a mount like `--mount ^home` to obtain something equivalent to `--no-home`

dtrudg:
Agreed that this is worth talking about. I'm personally unsure about a `--mount` option as I feel the `mount` as a term will easily be confused by users with the `-B / --bind`. We'd need to be very careful about distinguishing what a `--mount` is vs a `--bind` (mount).

Let's leave this on ice for a bit until there's a good chance to discuss it.

PR #5088, opened 2020-03-03
Cache rework Stage I

### LINKED ISSUE #4015: Singularity pull doesn't respect cache
## Version of Singularity:

3.3.0-rc.1.438.gc5e243fc4

### Expected behavior

Image is not pulled when image with same sha256 is already present in cache.

### Actual behavior

Image is pulled.

```
rm -rf $HOME/.singularity/cache/library
singularity pull -U library://sashayakovtseva/test/image-server:sha256.d50278eebfe4ca5655cc28503983f7c947914a34fbbb805481657d39e98f33f0
ls -la  $HOME/.singularity/cache/library/sha256.d50278eebfe4ca5655cc28503983f7c947914a34fbbb805481657d39e98f33f0
singularity pull -U library://sashayakovtseva/test/image-server:test
ls -la  $HOME/.singularity/cache/library/sha256.d50278eebfe4ca5655cc28503983f7c947914a34fbbb805481657d39e98f33f0
```

### LINKED ISSUE #5020: Concurrent pullls fail (race condition?)
### Version of Singularity:

```
$ singularity version
singularity version 3.5.2+266-gd3586d1de
# current master
```

### Expected behavior

```
singularity run library://pipeline/samtools:v1.9 --help
# help output
# concurrently in a second shell
singularity run library://pipeline/samtools:v1.9 --help
# help output
```

### Actual behavior

```
# second concurrently started run:
FATAL:   Unable to handle library://pipeline/samtools:v1.9 uri: unable to check if /.../.singularity/cache/library/sha256.721a1b94e24ff67342f7531b240fca6d527fe4e25ebb40eef658ea1557a30ba1/samtools_v1.9.sif exists: hash does not match
```

I think what happens is that the first run creates the cache directory and starts downloading directly to `cache/library/sha256.721a1b94e24ff67342f7531b240fca6d527fe4e25ebb40eef658ea1557a30ba1/samtools_v1.9.sif`. The second run gets the image hash, sees that the cache directory already exists and calculates the hash from `samtools_v1.9.sif` that it finds there. But this is still incomplete, so the hash check fails.

Normally I wouldn't consider this a drastic problem, but I'm using singularity in nextflow pipelines. If I use the same image in multiple processes that are started in parallel, the pipeline fails when starting for the first time (possibly first few times if there are multiple instances of images in parallel processes). 

- open two terminals
- prepare two singularity run library://.... commands with the same image URI
- start first
- while the first is still downloading, start second

source

dtrudg:
This is the 1st step toward reworking the cache / image source client code to reduce duplication and make operations that fetch things into the cache safe where multiple operations are run concurrently on a file system that supports atomic rename.

Fixes: #4015 
Fixes: #5020 

**This PR does...**

- Move `pkg/internal/client/cache` to `pkg/internal/cache` as it's used by the clients, but also other code, and is not a client itself.
- Move all the image source client code (oras, shub, net, etc.) to `pkg/internal/client/*`
- Make downloads into the cache happen to a temporary file, which is then renamed, for safety with concurrent runs.
- Remove duplication of image client & cache handling across `pull`, `actions` and in `build/sources`
- Ensures all sources consistently use the cache across `pull` `actions` `build` - prior to this some things like net weren't caching in some of these.
- Add a common ProgressBar callback in sylog, that respects `--quiet` and is used across the client code.

**Future work**

This PR is getting rather large, so instead of making it ever bigger I'd like to put up additional PRs over the next week to address the following:

 - The `pkg/image/client/xxx` code is just a collection of functions per client, mostly duplicated and named consistently, but with varying and long function sigs. Refactor so there is a common `client.Config` struct, and specific clients implement a common `client.Puller` and `client.Pusher` interfaces.

 - *REGRESSION* - CTRL-C handlers have been stripped out. They need to be added back in, but using contexts with cancellation so that actions to cleanup happen close to the code creating files.

 - The `cache list` and `cache clean` code needs to be properly refactored so that things directly acting on cache directories are in `pkg/internal/cache` and not throughout the CLI code.


internal/app/singularity/cache_clean_linux.go
```diff
@@ -105,64 +55,50 @@ func removeCacheEntry(name, cacheType, cacheDir string, op func(string) error) (
...
-			for _, cacheType := range cacheTypes {
-				cacheDir, _ := cacheTypeToDir(imgCache, cacheType)
-				sylog.Debugf("Removing cache type %q with name %q from directory %q ...", cacheType, name, cacheDir)
-				foundMatch, err := removeCacheEntry(name, cacheType, cacheDir, op)
-				if err != nil {
-					return err
-				}
-				if foundMatch {
-					matches++
+	/*
```
cclerget, line 71: Is it something to remove or pending for the next PR ?

internal/app/singularity/cache_list_linux.go
```diff
@@ -50,53 +49,32 @@ func listTypeCache(printList bool, name, cachePath string) (int, int64, error) {
...
 
 	var (
 		totalSize int64
-		count     int
 	)
 
-	for _, dir := range cacheDirs {
-		checkStat, err := os.Stat(filepath.Join(cachePath, dir.Name()))
+	for _, entry := range cacheEntries {
+		fileInfo, err := os.Stat(filepath.Join(cachePath, entry.Name()))
```
cclerget: This `Stat` call is not required as entry is `os.FileInfo`

internal/pkg/sylog/sylog.go
```diff
@@ -205,3 +207,39 @@ func Writer() io.Writer {
 
 	return os.Stderr
 }
+
+// ProgressCallback is a function that provides progress information copying from a Reader to a Writer
+type ProgressCallback func(int64, io.Reader, io.Writer) error
+
+// ProgressBarCallback returns a progress bar callback unless e.g. --quiet or lower loglevel is set
+func ProgressBarCallback() ProgressCallback {
```
cclerget: `ProgressBarCallback` could be in `internal/pkg/client` instead and use sylog.GetLevel() to replace :

```
if loggerLevel <= -1 {
    return nil
}
```

cclerget (changes requested):
Overall this is a nice cleanup :+1:

internal/app/singularity/cache_clean_linux.go
dtrudg: I wanted to return this when I refactor the clean and list stuff so that the operations are in the cache code, not CLI, in a following cleanup.

internal/pkg/sylog/sylog.go
dtrudg: Okay - I thought about that first, but put it in sylog in case it will be used somewhere else. Having checked I don't think it will - so I'll move it and revert the sylog changes which will make the diff smaller for sure.

cclerget (approved):
LGTM

internal/app/singularity/cache_clean_linux.go
dtrudg: I'll make an issue so this is note missed - https://github.com/sylabs/singularity/issues/5095

PR #5124, opened 2020-03-17
Allow to specify another configuration file

### LINKED ISSUE #4403: Configurage path to conf file for non-setuid build
Baking the singularity.conf path into the binary is obviously necessary when built with, and using, setuid. I don`t think it is necessary when using user namespaces to run without privilege. In this case, any user can build and install the binary, with any path to the config file so there is not point restricting it.  
One needs a build machine with the dependencies, and some fake path to mimic the filesystem available where you want to run it (eg. HPC), so in practice it is just awkward to do.

Could we have an argument and env to set the conf path, but only active when running in user mode? This way we could use a standard binary (copied from cvmfs) without needing the conf in that specific path.

Cheers,
Rod.

cclerget:

This PR allows to specify another `singularity.conf` file via a global flag `-c`/`--config` and can only be used by root user (or fakeroot by nesting execution) or for unprivileged installation.

This flag would be useful to execute build `post` section with the help of `exec` command without being dependent of the current configuration in order to have consistent execution of the `post` section.
In the meantime it also addresses the issue #4403.

 - Fixes #4403

### DIFF
--- a/cmd/internal/cli/actions_linux.go
+++ b/cmd/internal/cli/actions_linux.go
@@ -148,9 +148,9 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 
 	engineConfig := singularityConfig.NewConfig()
 
-	engineConfig.File, err = singularityconf.Parse(buildcfg.SINGULARITY_CONF_FILE)
-	if err != nil {
-		sylog.Fatalf("Unable to parse singularity.conf file: %s", err)
+	engineConfig.File = singularityconf.GetCurrentConfig()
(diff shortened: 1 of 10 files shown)


dtrudg (changes requested):
As this is a security sensitive issue please could you add a test that ensures it absolutely does not work for a normal user in the setuid install. Thanks.

cclerget:
Yes it's planned, just need to make another PR for a e2e addition, once merged I could write tests

dtrudg (approved):

PR #5136, opened 2020-03-23
Grab all inspect metadata and add them as a SIF descriptor.

### LINKED ISSUE #4279: Image metadata should be stored in its own partition

cclerget:

Grab all inspect metadata and add them as a SIF descriptor.

 - Fixes #4279

### DIFF
--- a/e2e/actions/actions.go
+++ b/e2e/actions/actions.go
@@ -2018,8 +2018,8 @@ func (c actionTests) bindImage(t *testing.T) {
 			name:    "SifWithID",
 			profile: e2e.UserProfile,
 			args: []string{
-				// rootfs ID is now '3'
-				"--bind", c.env.ImagePath + ":/rootfs:id=3",
+				// rootfs ID is now '4'
+				"--bind", c.env.ImagePath + ":/rootfs:id=4",
 				c.env.ImagePath,
 				"test", "-d", "/rootfs/etc",
 			},
--- a/e2e/imgbuild/imgbuild.go
+++ b/e2e/imgbuild/imgbuild.go
@@ -827,7 +827,7 @@ func (c imgBuildTests) buildDefinition(t *testing.T) {
 }
 
 func (c *imgBuildTests) ensureImageIsEncrypted(t *testing.T, imgPath string) {
-	sifID := "3" // Which SIF descriptor slot contains the (encrypted) rootfs
+	sifID := "4" // Which SIF descriptor slot contains the (encrypted) rootfs
 	cmdArgs := []string{"info", sifID, imgPath}
 	c.env.RunSingularity(
 		t,
--- a/internal/pkg/build/assemblers/sif.go
+++ b/internal/pkg/build/assemblers/sif.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"regexp"
 	"runtime"
+	"sort"
 	"strconv"
 	"syscall"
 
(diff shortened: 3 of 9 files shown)


dtrudg (commented):
Hi @cclerget. In principle I'm okay with brining the existing style of script metadata into the SIF descriptor like this as a first stepping stone toward getting things out of the rootfs.

What I'd like to do, though is wait on this a little bit until @EmmEff has some available to discuss how the library might consume metadata to enhance the UX, and @tri-adam has some valuable time to think about the road toward more OCI compatibility. I'm not sure whether we migrate slowly from here towards an `oci-config.json` with some implementation specific additions, or something else. Would be good to have at least a bit more thought on this I think.

dtrudg:
@cclerget - probably a good time to circle back on this one now. I'm definitely still of the opinion that capturing the metadata like this in the SIF is a win for inspecting without needing to mount the rootfs. That's good for Singularity and any other tools.

Looking forward though, are there any plans on your side to be moving more toward OCI driven metadata format etc? My only real concern is that if this is inspect JSON format is a stepping stone, and would be short lived, then we need to consider a path for transition of anything that might start consuming it.

cclerget:
@dctrud Regarding OCI compatibility, you are thinking about a fit with https://github.com/opencontainers/image-spec/blob/79b036d80240ae530a8de15e1d21c7ab9292c693/specs-go/v1/config.go#L24 ?

dtrudg:
@cclerget - yeah, something like that. I don't know if that's a long term plan though? On the roadmap draft doc I've left a space for @gmkurtzer to indicate long-term aims. It would be good if we could get an idea of if there is an intent to do more OCI work so we can think about metadata in that context.

There are probably good reasons that we can't be 100% OCI compatible by default, due to the security model etc, and that's fine. It'd just be good to have a bit of an idea of where GMK thinks the project should be aiming over coming releases.

dtrudg:
Hi @cclerget - this is on the 3.7 roadmap still I believe. Do you want to rebase and we can all go over the PR now?

dtrudg (commented):
I'm good with this code in theory - but can we still get a little bit more information about the direction embedded metadata might take before we merge this? I'm keen to know where we are going if this is an intermediate step.

Ultimately... are we going to have the descriptor as a primary source of metadata, and inject files it in the container at runtime if needed?

cclerget:
> Ultimately... are we going to have the descriptor as a primary source of metadata, and inject files it in the container at runtime if needed?

Like this is done actually injection is not possible, the descriptor just reflect the image content to ease inspection without extraction steps, the primary source during runtime remains the image. Is injection something guys want/need ?

dtrudg:
> > Ultimately... are we going to have the descriptor as a primary source of metadata, and inject files it in the container at runtime if needed?
> 
> Like this is done actually injection is not possible, the descriptor just reflect the image content to ease inspection without extraction steps, the primary source during runtime remains the image. Is injection something guys want/need ?

It's not something I thought was urgent. I just remember that @gmkurtzer was keen to move metadata out of the container in past discussion - so wanted to check if it's something that you or other Ctrl-Cmd folks were thinking about pushing forward with.

Given there isn't a strong path forward, but this enables reasoning about a container without having to mount an fs it's likely quite useful for tools that might work with singularity containers. I think we can merge at this point.

dtrudg (approved):

PR #5138, opened 2020-03-23
Remove go report badge (service misbehaving)

dtrudg:
Fixes: #5315

Our go report badge on the README is showing an 'F' recently, instead of the previous 'A'. We haven't done anything to go from A -> F - the issue is that the go report checks are running out of memory on their server, and failing:

https://github.com/gojp/goreportcard/issues/313

```
An error occurred while running this test (AddError: could not parse "_repos/src/github.com/sylabs/singularity/etc/conf/fatal error: runtime: out of memory:1::warning: file is not gofmted with -s (gofmt)" - strconv.Atoi: parsing " runtime": invalid syntax)
```

We should probably disable the badge until this is fixed, and contact go report.


cclerget (approved):
LGTM

PR #5142, opened 2020-03-24
Log scs-library-client to sylog.debug via DebugLogger

### LINKED ISSUE #5141: Log scs-library-client debug ouput through sylog
The debug logs that scs-library-client is producing are going to `/dev/null` when called from Singularity. We need to pass the client a `log.Logger` instance that will write messages to Sylog debug messages.

dtrudg:

Add a DebugLogger to sylog, which implements the go-log/log Logger
interface that scs-library client expects to log to. Write all messages
out via sylog.Debugf

This allows scs-library-client log messages to be seen with the `-d` /
`--debug` option to the CLI.

 - Fixes: #5141

### DIFF
--- a/cmd/internal/cli/actions.go
+++ b/cmd/internal/cli/actions.go
@@ -12,6 +12,7 @@ import (
 	"runtime"
 	"strings"
 
+	golog "github.com/go-log/log"
 	"github.com/spf13/cobra"
 	scslibrary "github.com/sylabs/scs-library-client/client"
 	"github.com/sylabs/singularity/docs"
@@ -81,6 +82,7 @@ func handleLibrary(ctx context.Context, imgCache *cache.Handle, pullFrom, librar
 	c := &scslibrary.Config{
 		AuthToken: authToken,
 		BaseURL:   libraryURL,
+		Logger:    (golog.Logger)(sylog.DebugLogger{}),
 	}
 	return library.Pull(ctx, imgCache, pullFrom, runtime.GOARCH, tmpDir, c, keyServerURL)
 }
--- a/internal/pkg/sylog/sylog.go
+++ b/internal/pkg/sylog/sylog.go
@@ -205,3 +205,18 @@ func Writer() io.Writer {
 
 	return os.Stderr
 }
+
+// DebugLogger is an implementation of the go-log/log Logger interface that will
+// output log messages via sylog.debug when required by external packages such
+// as the scs-library-client
+type DebugLogger struct{}
+
+// Output a log message via sylog.Debugf
+func (t DebugLogger) Log(v ...interface{}) {
+	writef(os.Stderr, debug, "%s", fmt.Sprint(v...))
+}
+
+// Output a formatted log message via sylog.Debugf
+func (t DebugLogger) Logf(format string, v ...interface{}) {
+	writef(os.Stderr, debug, format, v...)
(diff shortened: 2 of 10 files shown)


internal/pkg/sylog/sylog_dummy.go
```diff
@@ -12,6 +12,8 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+
```
cclerget: I'm wondering if it's required to be included in `sylog_dummy` to avoid the extra dependency ?

internal/pkg/sylog/sylog.go
```diff
@@ -205,3 +206,21 @@ func Writer() io.Writer {
...
+// output log messages via sylog.debug when required by external packages such
+// as the scs-library-client
+type DebugLogger struct {
+}
+
+var _ golog.Logger = (*DebugLogger)(nil)
+
+// Output a log message via sylog.Debugf
+func (t DebugLogger) Log(v ...interface{}) {
+	Debugf("%s", fmt.Sprint(v...))
```
cclerget: By using `Debugf` here sylog will always report `Log`/`Logf` as the caller in trace because of https://github.com/sylabs/singularity/blob/2206b6d1fcacf25ad6a6c82bd54166ab5bc2d8e7/internal/pkg/sylog/sylog.go#L93
Replacing by `writef(os.Stderr, debug, format, a...)` should fix that

cclerget (changes requested):

cmd/internal/cli/actions.go
```diff
@@ -81,6 +81,7 @@ func handleLibrary(ctx context.Context, imgCache *cache.Handle, pullFrom, librar
 	c := &scslibrary.Config{
 		AuthToken: authToken,
 		BaseURL:   libraryURL,
+		Logger:    sylog.DebugLogger{},
```
cclerget: Could we use `(golog.Logger)(sylog.DebugLogger{})` and add `golog` dependency in package where it's used rather than in `sylog` because it would be nice for plugin to have `sylog` as a public package

internal/pkg/sylog/sylog_dummy.go
cclerget: By using `(golog.Logger)(sylog.DebugLogger{})` we could avoid the extra dependency in `sylog` and end with something like in the files :

In `sylog_dummy.go` :

```
type DebugLogger struct {}

// Output a log message via sylog.Debugf
func (DebugLogger) Log(v ...interface{}) {}

// Output a formatted log message via sylog.Debugf
func (DebugLogger) Logf(format string, v ...interface{}) {}
```

In `sylog.go` :

```
type DebugLogger struct {}

// Output a log message via sylog.Debugf
func (DebugLogger) Log(v ...interface{}) {
	writef(os.Stderr, debug, "%s", fmt.Sprint(v...))
}

// Output a formatted log message via sylog.Debugf
func (DebugLogger) Logf(format string, v ...interface{}) {
	writef(os.Stderr, debug, format, v...)
}
```

internal/pkg/sylog/sylog.go
dtrudg: Yeah - I just noticed that, will fix.

cclerget (approved):
LGTM

PR #5145, opened 2020-03-24
Fix for bug #5144

### LINKED ISSUE #5144: Version 2.6.1: Failed to mount squashfs: Invalid argument
### Version of Singularity: 2.6.1

For backwards compatibility reasons, I am running 2.6.1, built myself; I get the following error:

```
$ sudo singularity shell -c my-container.sif
ERROR  : Failed to mount squashfs image in (read only): Invalid argument
ABORT  : Retval = 255
$ dmesg
[317832.963914] squashfs: Unknown parameter 'errors'
```

### Expected behavior

I expect a shell.

### Actual behavior

Got the error above.

### Fix

The error is caused by a call in ```lib/image/squashfs/mount.c``` to ```singularity_mount```, which in turn calls system ```mount```.

It passes the option ```error=...``` to mount, which is not supported for squashfs file systems. I am assuming it was copied from another call, where it is valid.

The fix is to remove the option.

lyklev:
## Description of the Pull Request:

Corrected a call to sys/mount(...) with invalid options for the squashfs file system type.

 - Fixes #5144 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ and changes prior to that are (unfortunately) done retrospectively. Critical ite
  - migration guidance (how to convert images?)
  - changed behaviour (recipe sections work differently)
 
+## [v2.6.1]
+
+### [Security related fixes](https://cve.mitre.org/cgi-bin/cvename.cgi?name=2018-1929)
+ - disables instance features for mount commands, disables instance join for 
+   start command, and disables daemon start for action commands
+
 ## [v2.6.0]
 
 ### Bug fixes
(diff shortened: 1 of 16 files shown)


dtrudg:
Hi @lyklev - I'm afraid that we do not support the 2.x series open-source version of Singularity any more. The branches under `vault/` are archived, and no changes will be applied to them.

dtrudg:
Noting for the benefit of anyone finding this from search results that it appears the bug is triggered under Kernel 5.1 and above. See #5144 for more discussion, and a link to a the single commit diff from @lyklev

PR #5233, opened 2020-04-21
Fix SIGURG Handling

### LINKED ISSUE #5231: SIGURG Handling with Go >= 1.14
### Version of Singularity

```sh
$ singularity version
3.5.2+470-gd5c3f1347
```

I believe the Go version is also relevant:

```sh
$ go version
go version go1.14.2 linux/amd64
```

### Expected behavior

Expected instance to start when running:

```sh
singularity instance start docker://busybox test_from_docker
```

### Actual behavior

Instance did not start (`singularity instance list` was empty.)

Adding `--debug` shed some light on the problem:

```sh
...
DEBUG   [U=1000,P=1]       StartProcess()                Received signal urgent I/O condition
DEBUG   [U=1000,P=1]       StartProcess()                No child process, exiting ...
DEBUG   [U=1000,P=140799]  Master()                      Child exited with exit status 151
...
```

This appears to be related to golang/go/issues/37942. We will need to catch and ignore `SIGURG` to avoid this type of issue.

Install Go >= 1.14.

```
$ cat /etc/os-release
NAME="Ubuntu"
VERSION="20.04 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04 LTS"
VERSION_ID="20.04"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
```

From source

tri-adam:
Ignore `SIGURG`, which is used for non-cooperative goroutine preemption starting with Go 1.14 (more info available [here](https://github.com/golang/go/issues/24543)).

Fixes #5231

### DIFF
--- a/internal/pkg/runtime/engine/singularity/process_linux.go
+++ b/internal/pkg/runtime/engine/singularity/process_linux.go
@@ -56,7 +56,9 @@ const defaultShell = "/bin/sh"
 func (e *EngineOperations) StartProcess(masterConn net.Conn) error {
 	// Manage all signals.
 	// Queue them until they're ready to be handled below.
-	signals := make(chan os.Signal, 1)
+	// Use a channel size of two here, since we may receive SIGURG, which is
+	// used for non-cooperative goroutine preemption starting with Go 1.14.
+	signals := make(chan os.Signal, 2)
 	signal.Notify(signals)
 
 	if err := e.runFuseDrivers(true, -1); err != nil {
@@ -244,6 +246,11 @@ func (e *EngineOperations) StartProcess(masterConn net.Conn) error {
 						statusChan <- status
 					}
 				}
+			case syscall.SIGURG:
+				// Ignore SIGURG, which is used for non-cooperative goroutine
+				// preemption starting with Go 1.14. For more information, see
+				// https://github.com/golang/go/issues/24543.
+				break
 			default:
 				signal := s.(syscall.Signal)
 				// EPERM and EINVAL are deliberately ignored because they can't be
--- a/internal/app/starter/master_linux.go
+++ b/internal/app/starter/master_linux.go
@@ -123,7 +123,9 @@ func Master(rpcSocket, masterSocket int, containerPid int, e *engine.Engine) {
 	// we could receive signal from child with CreateContainer call so we
 	// set the signal handler earlier to queue signals until MonitorContainer
 	// is called to handle them
-	signals := make(chan os.Signal, 1)
+	// Use a channel size of two here, since we may receive SIGURG, which is
+	// used for non-cooperative goroutine preemption starting with Go 1.14.
+	signals := make(chan os.Signal, 2)
 	signal.Notify(signals)
 
 	ctx := context.TODO()
--- a/internal/pkg/runtime/engine/fakeroot/engine_linux.go
+++ b/internal/pkg/runtime/engine/fakeroot/engine_linux.go
@@ -261,6 +261,11 @@ func (e *EngineOperations) MonitorContainer(pid int, signals chan os.Signal) (sy
 				continue
 			}
 			return status, nil
+		case syscall.SIGURG:
+			// Ignore SIGURG, which is used for non-cooperative goroutine
+			// preemption starting with Go 1.14. For more information, see
+			// https://github.com/golang/go/issues/24543.
+			break
 		default:
 			if err := syscall.Kill(pid, s.(syscall.Signal)); err != nil {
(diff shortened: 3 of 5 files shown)


internal/pkg/runtime/engine/singularity/process_linux.go
```diff
@@ -244,6 +244,11 @@ func (e *EngineOperations) StartProcess(masterConn net.Conn) error {
 						statusChan <- status
 					}
 				}
+			case syscall.SIGURG:
+				// Ignore SIGURG, which is used for non-cooperative goroutine
+				// preemption starting with Go 1.14. For more information, see
+				// https://github.com/golang/go/issues/24543.
+				sylog.Debugf("Ignoring SIGURG signal")
```
cclerget: Do we really care about this debug message as it's for golang runtime internal purpose ? And also a risk of flooding

cclerget (commented):
I think all engine operation `MonitorContainer` for singularity/fakeroot/oci engines should also ignore this signal to not send it to container processes

EDIT: simplest is to ignore it in master process https://github.com/sylabs/singularity/blob/c4a0dc2fbcd0793d2b2bab6f23af1e639cdc71ba/internal/app/starter/master_linux.go#L127 with a call to `signal.Ignore`

internal/pkg/runtime/engine/singularity/process_linux.go
jmstover: Wouldn't you want to log it at least once? Just to log the... "We got this signal sent to us and we're ignoring it."

tri-adam:
> I think all engine operation `MonitorContainer` for singularity/fakeroot/oci engines should also ignore this signal to not send it to container processes

Makes sense.

> 
> EDIT: simplest is to ignore it in master process
> 
> https://github.com/sylabs/singularity/blob/c4a0dc2fbcd0793d2b2bab6f23af1e639cdc71ba/internal/app/starter/master_linux.go#L127
> 
> with a call to `signal.Ignore`

I believe this solution would still leave a small window of time whereby a `SIGURG` could be queued in the `signals chan`, between calls to `signal.Notify()` and `signal.Ignore()`.

To ensure `MonitorContainer` doesn't pass along `SIGURG`, two options I can see are:

1. Call `signal.Notify()` with a specific list of signals to send to `signals`, which would ensure that `signals` never receives a `SIGURG`.
2. Ensure each implementation of `(engine.Operations).MonitorContainer()` ignores `SIGURG`.

It's also worth noting that this new `SIGURG` behaviour also increases the risk that legitimate signals may be dropped. We are using a channel size of one, and according to the [documentation](https://golang.org/pkg/os/signal/#Notify), signals are dropped once the channel buffer is filled. Experimentation shows that a busy Go program can generate many `SIGURG` signals per second, but only queues a single `SIGURG` at a time in a channel registered with `signal.Notify()`.

In light of all of that, I propose:

- Update `signals chan` to be of size 2, to account for the likelyhood that `SIGURG` is queued.
- Ensure each implementation of `(engine.Operations).MonitorContainer()` ignores `SIGURG`.

internal/pkg/runtime/engine/singularity/process_linux.go
tri-adam: We're actually logging _all_ signals received in `StartProcess()`:

https://github.com/sylabs/singularity/blob/c4a0dc2fbcd0793d2b2bab6f23af1e639cdc71ba/internal/pkg/runtime/engine/singularity/process_linux.go#L230

So this `Debugf` simply results in twice the noise.

As we're not (to my knowledge) leveraging out-of-band data, the chances of a `SIGURG` that isn't related to non-cooperative goroutine preemption seems very low. And since we _are_ logging all signals, there is still a debug hint if ignoring this signal ever becomes an issue.

With that in mind, I'll remove the `Ignoring SIGURG signal` debug print unless there are any strong objections.

dtrudg (approved):
LGTM @tri-adam  - Am happy that this works, and has the effect of the point @cclerget was making about filtering more generally, but without the potential for a signal to slip through that you brought up.

tri-adam:
@cclerget can confirm you're good with this as implemented? Will hold off merging until you can weigh in. Thanks!

cclerget:
> @cclerget can confirm you're good with this as implemented? Will hold off merging until you can weigh in. Thanks!

@tri-adam LGTM

> It's also worth noting that this new SIGURG behaviour also increases the risk that legitimate signals may be dropped

I still not understand why Go runtime internal didn't handle it internally, they own the signal handler there is `siginfo` available and they can use sigqueue to pass a specific value to distinguish this signal ...

PR #5237, opened 2020-04-22
Build: use buffered chan for signal notification

tri-adam:
While working on #5233, I noticed an un-buffered channel being used with `signal.Notify()` in another part of the code. Strictly speaking, this could cause a `SIGTERM` to be missed. Submitting in a separate PR as this is not related to the bug documented in #5231.


dtrudg (approved):
LGTM

tri-adam:
Blocked until E2E tests are fixed. Working on that in #5238...

PR #5238, opened 2020-04-22
Testing: fix E2E test suite

tri-adam:
The E2E test suite references [library://alpine:latest](https://cloud.sylabs.io/library/library/default/alpine), which was recently updated. The contents of `90-environment.sh` was changed in #3746, and this appears to be causing a test failure.


dtrudg (approved):
Nice catch. LGTM

PR #5239, opened 2020-04-22
CI Improvements

tri-adam:
Modify CircleCI configuration with the aim of reducing time-to-finish.

### TLDR;

Before: ~11.5 minutes without `e2e_tests`, ~19 minutes with `e2e_tests`.

Screen Shot 2020-04-23 at 8 12 16 AM

After:  ~6 minutes without `e2e_tests`, ~14 minutes with `e2e_tests`.

Screen Shot 2020-04-23 at 8 11 49 AM

## Full Description

Detailed modifications are as follows:

- Replace `get_source` job with independent `checkout` steps in each job. The `get_source` job did a `checkout` and then `persist_to_workspace`. In practice, `checkout` takes around the same amount of time as a `persist_to_workspace`/`attach_workspace` round trip, but removing the `get_source` job entirely saves approximately one minute in overall workflow runtime.
- Rename `cache_go_mod` job to `check_go_mod`. `cache_go_mod` ran `scripts/check-go.mod` and then used `save_cache` to save `$GOPATH/pkg/mod`. By removing CircleCI cache usage, each job will instead pull Go modules from the [Go Module Mirror](https://proxy.golang.org/). This allows for jobs that previously relied on `cache_go_mod` to start roughly one minute earlier.
- Remove `build_check_and_install` job, which built Singularity and then ran `make check`. In theory, building Singularity once before fanning out to `e2e_tests`/`unit_tests`/`short_integration_tests` is a good idea. In practice however, `make install` currently requires a populated `$GOPATH/pkg/mod` to install Singularity, and it is overall faster to simply build Singularity in `e2e_tests`/`unit_tests`/`short_integration_tests` jobs. `make check` is run in both `go113-alpine` and `go113-stretch`, so removing `build_check_and_install` reduces duplication of effort there. Overall, this change adds approximately one minute to `e2e_tests`/`unit_tests`/`short_integration_tests` jobs, but allows them to start approximately four and a half minutes faster.
- Replace usage of `sylabsio/golang*` with official `golang` images, to leverage updates and simplify the upgrade path to future versions of Go. This seems to result in higher cache hit rates on the image layers, resulting in faster startup. When combined with extra steps necessary to install software, the total time per job is roughly equivalent.
- Remove some unused YAML anchor cruft.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,22 +86,12 @@ aliases:
   - &attach_workspace
     attach_workspace:
       at: ~/go
-  - &restore_cache
-    restore_cache:
-      keys:
-        - go-vm-mod-{{ checksum "go.sum" }}
-  - &save_cache
-    save_cache:
-      key: go-vm-mod-{{ checksum "go.sum" }}
-      paths:
-        - ~/go/pkg
 
 jobs:
-  cache_go_mod:
+  check_go_mod:
     <<: *vm_defaults
     steps:
       - checkout
-      - *restore_cache
       - run:
           <<: *setup_environment
       - run:
@@ -109,7 +99,6 @@ jobs:
       - run:
           name: Check go.mod
           command: scripts/check-go.mod
-      - *save_cache
 
   go113-stretch:
     <<: *defaults
@@ -167,7 +156,6 @@ jobs:
     <<: *vm_defaults
     steps:
       - checkout
-      - *restore_cache
       - run:
           <<: *setup_environment
       - run:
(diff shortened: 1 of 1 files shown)


dtrudg (approved):
Agreed this is useful on our current (older style) CircleCI plan where we are paying by the container for extra parallelism in our org and not by build minutes used.

If we get shunted over to a credits based plan at any point in future then we might want to revisit duplication of effort per job vs overall run time.

tri-adam:
Agreed, there's certainly more room for improvement in both time-to-completion and total resource utilization. This approach nudges both of those needles in the right direction, aside from cases where one job failing would prevent others from starting entirely.

PR #5279, opened 2020-05-06
Mark GPFS as incompatible for overlay lower and upper

dtrudg:
## Description of the Pull Request (PR):

When used as the lowerdir for an overlay in a mount namespace as we do, GPFS gives an NFS stale file handle error. It does not trigger the code in the kernel overlayfs source that gives a more specific error and directly denies overlay mount for other filesystems such as lustre.

We need to blacklist GPFS from overlay lower and upper usage, until we can confirm exactly what is intended to be supported by IBM.

### This fixes or addresses the following GitHub issues:

 - Fixes #5274

#### Before submitting a PR, make sure you have done the following:

- Read the [Guidelines for Contributing](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/sylabs/singularity/blob/master/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- Added tests to validate this PR and tested this PR locally with a `make testall`
- Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](https://github.com/sylabs/singularity/blob/master/CONTRIBUTORS.md)

Attn: @singularity-maintainers

PR #5293, opened 2020-05-14
Merge additional master fixes for 3.6.0-rc.3

dtrudg:
This merges 1 important fix:

#5279 - allows use of sandboxes on GPFS (via underlay... instead of failing overlays)

However, doing a simple branch merge here rather than cherry-picking as the fixes on the CNI dependency, and the containers/image are worth having (e.g. recognizing arm *v8*)


cclerget (approved):
LGTM

PR #5330, opened 2020-06-01
Detect current PID user-namespace in overlay checks #5315

### LINKED ISSUE #5315: %test fails in a fakeroot image build
### Version of Singularity:

```
$ singularity version

```
```
3.6.0~rc.3.1-1.el7
```

### Expected behavior
building a image file
### Actual behavior
```
INFO:    Adding environment to container
INFO:    Adding runscript
INFO:    Adding testscript
INFO:    Running testscript
FATAL:   container creation failed: mount overlay->/var/singularity/mnt/session/final error: while mounting overlay: can't mount overlay filesystem to /var/singularity/mnt/session/final: operation not permitted
FATAL:   While performing build: failed to execute %test script: exit status 255
```

Use lolcow.def from the documentation and append this %test section:
```
%test
    echo "space left:"
    df -h
```
Then run ```singularity build --fakeroot lolcow.sif lolcow.def```. It works without the %test section and fails with the section.

```
$ cat /etc/os-release


```
```
NAME="Scientific Linux"
VERSION="7.7 (Nitrogen)"
ID="scientific"
ID_LIKE="rhel centos fedora"
VERSION_ID="7.7"
PRETTY_NAME="Scientific Linux 7.7 (Nitrogen)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:scientificlinux:scientificlinux:7.7:GA"
HOME_URL="http://www.scientificlinux.org//"
BUG_REPORT_URL="mailto:scientific-linux-devel@listserv.fnal.gov"

REDHAT_BUGZILLA_PRODUCT="Scientific Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.7
REDHAT_SUPPORT_PRODUCT="Scientific Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="7.7"
```

rpm build from source of 3.6.0 rc.3 with the patch from pr #5313 applied.

dtrudg:

When considering overlay support in `internal/pkg/runtime/engine/singularity/prepare_linux.go` detect whether the current process is running in a user namespace, rather than looking only at OCIConfig.

This is required with the changes in 3.6 that mean the build engine is not separate, and we are running builds, and any contained tests, under the singularity engine. Failure to do this means that `%test` blocks in a build will fail due to attempting to use overlay.

 - Fixes #5315

### DIFF
--- a/internal/pkg/runtime/engine/singularity/prepare_linux.go
+++ b/internal/pkg/runtime/engine/singularity/prepare_linux.go
@@ -37,6 +37,7 @@ import (
 	"github.com/sylabs/singularity/pkg/sylog"
 	"github.com/sylabs/singularity/pkg/util/capabilities"
 	"github.com/sylabs/singularity/pkg/util/fs/proc"
+	"github.com/sylabs/singularity/pkg/util/namespaces"
 	"github.com/sylabs/singularity/pkg/util/singularityconf"
 	"golang.org/x/sys/unix"
 )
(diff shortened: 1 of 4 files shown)


internal/pkg/runtime/engine/singularity/prepare_linux.go
```diff
@@ -1009,22 +1010,31 @@ func (e *EngineOperations) setSessionLayer(img *image.Image) error {
...
-				return nil
-			}
-			if !writableImage {
-				sylog.Debugf("Using underlay layer: user namespace requested")
-				e.EngineConfig.SetSessionLayer(singularityConfig.UnderlayLayer)
-				return nil
-			}
-			sylog.Debugf("Not attempting to use overlay or underlay: writable flag requested")
+			userNS = true
+			break
```
tri-adam: If `userNS` is `true`, we don't need to iterate in this loop. To kill two birds with one stone, how about:

```
	for _, ns := range e.EngineConfig.OciConfig.Linux.Namespaces; !userNS {
		if ns.Type == specs.UserNamespace {
			userNS = true
		}
	}
```

tri-adam (commented):
LGTM, just one suggestion to simplify the code slightly.

cclerget (approved):
LGTM

internal/pkg/runtime/engine/singularity/prepare_linux.go
dtrudg: Adding `; !userNS` isn't valid on a `for xxx := range xxx`

Will add the extra if. Didn't as the loop is going to be over one item (mount namespace) in most cases, so not much perf benefit, but I agree it shows the intent.

tri-adam (approved):

PR #5383, opened 2020-06-17
Fix build instructions

DimitriPapadopoulos:

At least with Go 3.5.3, the expected build directory is `buildtree`, not `builddir`.

Do I really need to add myself to `CONTRIBUTORS.md`? This fixes a minor typo.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -99,7 +99,7 @@ You can build Singularity using the following commands:
 ```
 $ cd ${GOPATH}/src/github.com/sylabs/singularity && \
   ./mconfig && \
-  cd ./builddir && \
+  cd ./buildtree && \
   make && \
   sudo make install
 ```


dtrudg:
@DimitriPapadopoulos thanks for the PR.

In this case,  `builddir` is the default build directory, so the existing text is correct.

`buildtree` is only relevant if you set a _different_ build directory with the `-b` option to `mconfig`, as explained in the INSTALL.md text that comes after the standard `builddir` usage:

```
To build in a different folder and to set the install prefix to a different path:

$ ./mconfig -p /usr/local -b ./buildtree
```

DimitriPapadopoulos:
Not sure what you mean by _`builddir` is the default build directory, so the existing text is correct_. It didn't work for me with Go 1.14.4. I had to `cd buildtree`, `cd builddir` doesn't work.

PR #5398, opened 2020-06-25
Added log (.err, .out) paths to instance list

nathanchou4:

This PR does the following: 
- Adds -l, --logs which displays instance name, pid, and relevant log paths
- Adds log paths to instance list JSON format
- Adds myself as a Contributor!

**Examples:**
```
$ singularity instance list -l
INSTANCE NAME    PID      LOGS
test-lolcow      14002    /home/nchou/.singularity/instances/logs/localhost.localdomain/nchou/lolcow.err
                          /home/nchou/.singularity/instances/logs/localhost.localdomain/nchou/lolcow.out
test-alpine      13962    /home/nchou/.singularity/instances/logs/localhost.localdomain/nchou/test-alpine.err
                          /home/nchou/.singularity/instances/logs/localhost.localdomain/nchou/test-alpine.out
```

```
$ singularity instance list -j
{
        "instances": [
                {
                        "instance": "test-lolcow",
                        "pid": 14002,
                        "img": "/home/nchou/.singularity/cache/library/sha256.9b0dafaadb1cd1d14e4fbad23798730a938bafb9e6ad5f86e00ffb5b4541e576",
                        "ip": "",
                        "logErrPath": "/home/nchou/.singularity/instances/logs/localhost.localdomain/nchou/lolcow.err",
                        "logOutPath": "/home/nchou/.singularity/instances/logs/localhost.localdomain/nchou/lolcow.out"
                },
                {
                        "instance": "test-alpine",
                        "pid": 13962,
                        "img": "/home/nchou/.singularity/cache/library/sha256.03883ca565b32e58fa0a496316d69de35741f2ef34b5b4658a6fec04ed8149a8",
                        "ip": "",
                        "logErrPath": "/home/nchou/.singularity/instances/logs/localhost.localdomain/nchou/test-alpine.err",
                        "logOutPath": "/home/nchou/.singularity/instances/logs/localhost.localdomain/nchou/test-alpine.out"
                }
        ]
}
```

 - Fixes #[4819](https://github.com/hpcng/singularity/issues/4819)


dtrudg (commented):
Hey @nathanchou4 - this looks nice to me. 

Because I helped you directly on this I'm going to ask another person on the project to review.

cmd/internal/cli/instance_list_linux.go
```diff
@@ -47,6 +48,18 @@ var instanceListJSONFlag = cmdline.Flag{
...
 
+// -l|--logs
+var instanceListLogs bool
+var instanceListLogsFlag = cmdline.Flag{
+	ID:           "instanceListLogsFlag",
+	Value:        &instanceListLogs,
+	DefaultValue: false,
+	Name:         "logs",
+	ShortHand:    "l",
+	Usage:        "print logErr and logOut paths of instances",
```
dtrudg:
```suggestion
	Usage:        "display location of stdout and sterr log files for instances",
```

internal/app/singularity/instance_linux.go
```diff
@@ -23,17 +23,38 @@ type instanceInfo struct {
...
 		return fmt.Errorf("could not retrieve instance list: %v", err)
 	}
 
+	if showLogs {
+		_, err := fmt.Fprintf(w, "%-16s %-8s %s\n", "INSTANCE NAME", "PID", "LOGS")
+		if err != nil {
+			return fmt.Errorf("could not write list header: %v", err)
+		}
+		for _, i := range ii {
+			_, err := fmt.Fprintf(w, "%-16s %-8d %s\n%-16s %-8s %s\n", i.Name, i.Pid, i.LogErr, "", "", i.LogOut)
```
ikaneshiro:
```suggestion
			_, err := fmt.Fprintf(w, "%-16s %-8d %s\n%-26s %s\n", i.Name, i.Pid, i.LogErr, "", i.LogOut)
```
I would suggest combining the white space padding under one formatting directive so that you don't need to pass two empty string args. 

It may be worth looking at a package like [this](https://golang.org/pkg/text/tabwriter). It gives you tab alignment that could come in handy here. You could define a formatting string like `"%s\t%d\t%s\n\t\t%s\n"` preventing the need to pass an empty string at all.

internal/pkg/instance/instance_linux.go
```diff
@@ -47,6 +47,8 @@ type File struct {
 	Config []byte `json:"config"`
 	UserNs bool   `json:"userns"`
 	IP     string `json:"ip"`
+	LogErr string `json:"logErr"`
```
ikaneshiro: I don't think we need to store these log paths within the instance configuration. Instead I would suggest calling `instance.GetLogFilePaths()` directly from `internal/app/singularity/instance_linux.go`

ikaneshiro (commented):
Hi @nathanchou4, welcome to the contributors list!

I have left a couple suggestions for you.

internal/pkg/instance/instance_linux.go
nathanchou4: Thanks for the welcome! 

If I remove the log paths from within the instance configuration, I would have to call `instance.GetLogFilePaths()` twice in the `PrintInstanceList` function in `internal/app/singularity/instance_linux.go` which seems redundant: 
- Printing instance logs from CLI using the newly created -l (--logs) flag 
- Populating InstanceInfo for displaying instance list JSON format using the -j flag

To me, it seems that storing the log paths within the instance configuration fits the organization of the code. I'm more than willing to make the change, but I just want to ask why shouldn't we store the log paths within the instance configuration?

internal/app/singularity/instance_linux.go
nathanchou4: I checked out the go TabWriter package and it works great.

Currently, the `singularity instance list` uses fixed alignment `"%-16s" `formatting and when there exists `len(instanceName) > 16`, the name bleeds into the next column and the alignment is off. However, with TabWriter, the width of each column is dynamically adjusted so the bug goes away. 

Should I also change `singularity instance list` to use TabWriter, or would it be better to log an Issue and make a separate PR to take care of that?

internal/pkg/instance/instance_linux.go
ikaneshiro: The main reason behind the suggestion is because these paths don't _need_ to be stored on disk within the instance configuration json. They can be deduced directly from the directory structure within `~/.singularity/instances`.

After thinking about this more, I think your current approach is the best option as it does make sense to record all instance configuration data(like log paths) within the configuration json.

internal/app/singularity/instance_linux.go
ikaneshiro: Since changing that formatting will be small and will make the formatting in `PrintInstanceList()` uniform for both options, I think it's fine to go ahead and do that in this PR.

dtrudg: Just a note thatn if we are changing forwarding for an existing instance list it needs to be noted in the CHANGELOG in case anyone is scraping that output.

internal/pkg/runtime/engine/singularity/process_linux.go
```diff
@@ -341,10 +341,18 @@ func (e *EngineOperations) PostStartProcess(ctx context.Context, pid int) error
 		if err != nil {
 			return err
 		}
+
+		logErrPath, logOutPath, err := instance.GetLogFilePaths(name, instance.LogSubDir)
+		if err != nil {
+			return nil
```
tri-adam: Should this be `return err`?

CONTRIBUTORS.md
```diff
@@ -63,6 +63,7 @@
     - Michael Moore <michael.moore@nuance.com>
     - Mike Frisch <michael.frisch@sylabs.io>
     - Mike Gray <mike@sylabs.io>
+    - Nathan Chou <nathan.chou@sylabs.io> <choun@berkeley.edu>
```
tri-adam: Nitpick: looks like the `<emails>` in this list are comma separated?

internal/app/singularity/instance_linux.go
```diff
@@ -23,28 +24,59 @@ type instanceInfo struct {
...
-// formatJSON is true) to the passed writer.
-func PrintInstanceList(w io.Writer, name, user string, formatJSON bool) error {
+// formatJSON is true) to the passed writer. Additionally, fetches
+// log paths (if showLogs is true).
+func PrintInstanceList(w io.Writer, name, user string, formatJSON bool, showLogs bool) error {
+	if formatJSON && showLogs {
+		sylog.Fatalf("more than one flags have been set")
+	}
+
+	tabWriter := tabwriter.NewWriter(os.Stdout, 0, 8, 2, ' ', 0)
```
tri-adam: I'd encourage you to use a `defer tabWriter.Flush()` immediately following `tabwriter.NewWriter()` (and remove the two calls below). This ensures the output is always flushed, regardless of which `return` is hit, and [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)s the code.

tri-adam: I think this should probably be `tabwriter.NewWriter(w, ...`? According to the function documentation, it writes to the passed `w io.Writer` rather than `os.Stdout`.

tri-adam (commented):
Nice work @nathanchou4, just a few questions/suggestions. Thanks!

internal/pkg/runtime/engine/singularity/process_linux.go
tri-adam: @nathanchou4, I think this one still needs to be addressed?

tri-adam (approved):
Nice work @nathanchou4. LGTM!

PR #5428, opened 2020-07-14
Load command tree in RootCmd() used by doc gen code

### LINKED ISSUE #5427: v3.6.0 docs.go doesn't generate full command tree
### Version of Singularity:

```
3.6.0
```

### Actual behavior

With changes made in https://github.com/hpcng/singularity/pull/4924 the full command tree isn't available to the `docs.go` CLI docs generation at the time it calls Cobra functions to generate docs. We need to call `cli.Init` first.

dtrudg:

With changes in #4924 the command tree loading is deferred until
Init() is called - so we need to call it in RootCmd() which is
used for docs and coverage generation.

 - Fixes #5427 

### DIFF
--- a/cmd/docs/docs.go
+++ b/cmd/docs/docs.go
@@ -63,6 +63,10 @@ func main() {
 		Use:       "makeDocs {markdown | man | rst}",
 		Short:     "Generates Singularity documentation",
 		Run: func(cmd *cobra.Command, args []string) {
+			// We must Init() as loading commands etc. is deferred until this is called.
+			// Using true here will result in local docs including any content for installed
+			// plugins.
+			cli.Init(true)
 			rootCmd := cli.RootCmd()
 			switch args[0] {
 			case "markdown":


cmd/internal/cli/singularity.go
```diff
@@ -413,6 +413,10 @@ var singularityCmd = &cobra.Command{
 
 // RootCmd returns the root singularity cobra command.
 func RootCmd() *cobra.Command {
+	// We must Init() as loading commands etc. is deferred until this is called.
+	// Using true here will result in local docs including any content for installed
+	// plugins.
+	Init(true)
```
cclerget: Looks like putting `Init` here is causing to be called multiple times and would fail if there is a plugin installed because it would be loaded multiple times : https://github.com/hpcng/singularity/blob/4363ec8f9f1e81b51602362d938559ff9dffeafe/cmd/docs/cmds/cmdtree.go#L234-L238

May be simpler to avoid side effects to call `Init` just before this line: https://github.com/hpcng/singularity/blob/4363ec8f9f1e81b51602362d938559ff9dffeafe/cmd/docs/docs.go#L66

cclerget (changes requested):

cmd/internal/cli/singularity.go
dtrudg: @cclerget - moved this now. Note that in the `docs.go` I am calling `Init(true)` as I we may want to have the ability to generate docs that include any plugin related modifications?

In `cmdtree.go` it's still `Init(false)` because that's used for e2e command coverage stuff... which wouldn't cover any plugins etc.

Does that sound fair?

cclerget: @dctrud Sounds good :+1:

cclerget (approved):

PR #5445, opened 2020-07-21
Provide login/logout command to manage Docker/OCI registry credentials

cclerget:

This PR adds:
- `singularity login` command to login to a registry:
  - `singularity login -u test -p test localhost:5000`
- `singularity logout` command to logout from a logged registry
  - `singularity logout localhost:5000`

Credentials are stored in `~/.singularity/docker-config.json` and are used automatically when pulling/pushing with OCI/Docker/ORAS registries.
It would also be possible to configure docker credential helpers in this file for registries like GCR, ECR ...

Follow-up about licensing of docker auth server code : https://github.com/adigunhammedolalekan/registry-auth/issues/1

 - Fixes #

### DIFF
--- a/cmd/internal/cli/login.go
+++ b/cmd/internal/cli/login.go
@@ -1,0 +1,122 @@
+// Copyright (c) 2020, Ctrl-Cmd Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package cli
+
+import (
+	"context"
+	"io/ioutil"
+	"os"
+	"strings"
+
+	auth "github.com/deislabs/oras/pkg/auth/docker"
+	"github.com/spf13/cobra"
+	"github.com/sylabs/singularity/docs"
+	"github.com/sylabs/singularity/internal/pkg/util/interactive"
+	"github.com/sylabs/singularity/pkg/cmdline"
+)
+
+var (
+	loginUsername      string
+	loginPassword      string
+	loginPasswordStdin bool
+	loginInsecure      bool
+)
+
+// -u|--username
+var loginUsernameFlag = cmdline.Flag{
+	ID:           "loginUsernameFlag",
+	Value:        &loginUsername,
+	DefaultValue: "",
+	Name:         "username",
+	ShortHand:    "u",
+	Usage:        "username to authenticate with (leave it empty for token authentication)",
+	EnvKeys:      []string{"LOGIN_USERNAME"},
+}
+
+// -p|--password
+var loginPasswordFlag = cmdline.Flag{
+	ID:           "loginPasswordFlag",
+	Value:        &loginPassword,
+	DefaultValue: "",
+	Name:         "password",
+	ShortHand:    "p",
+	Usage:        "password to authenticate with",
+	EnvKeys:      []string{"LOGIN_PASSWORD"},
+}
+
+// --password-stdin
+var loginPasswordStdinFlag = cmdline.Flag{
+	ID:           "loginPasswordStdinFlag",
+	Value:        &loginPasswordStdin,
+	DefaultValue: false,
+	Name:         "password-stdin",
+	Usage:        "take password from standard input",
+}
+
+// -i|--insecure
+var loginInsecureFlag = cmdline.Flag{
+	ID:           "loginInsecureFlag",
+	Value:        &loginInsecure,
+	DefaultValue: false,
+	Name:         "insecure",
+	ShortHand:    "i",
+	Usage:        "allow insecure login",
+	EnvKeys:      []string{"LOGIN_INSECURE"},
+}
+
+func init() {
+	addCmdInit(func(cmdManager *cmdline.CommandManager) {
+		cmdManager.RegisterCmd(LoginCmd)
+
+		cmdManager.RegisterFlagForCmd(&loginUsernameFlag, LoginCmd)
+		cmdManager.RegisterFlagForCmd(&loginPasswordFlag, LoginCmd)
+		cmdManager.RegisterFlagForCmd(&loginPasswordStdinFlag, LoginCmd)
+		cmdManager.RegisterFlagForCmd(&loginInsecureFlag, LoginCmd)
+	})
+}
+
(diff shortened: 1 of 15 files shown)


e2e/internal/e2e/docker_auth_server.go
```diff
@@ -0,0 +1,247 @@
+// Code taken from https://github.com/adigunhammedolalekan/registry-auth
```
dtrudg: It's a bit awkward that there is no license specified at all in the repo for that code. Could you open an issue on the repo, or otherwise ask the author to clarify that re-use is okay?

cmd/internal/cli/login.go
```diff
@@ -0,0 +1,127 @@
...
+		cmdManager.RegisterFlagForCmd(&loginInsecureFlag, LoginCmd)
+	})
+}
+
+// LoginCmd is the 'login' command that allows user to login to service.
+var LoginCmd = &cobra.Command{
+	Args: cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		hostname := args[0]
+		username := loginUsername
```
dtrudg, line 88: If `loginUserName` is "" and we are interactive should we note that we are asking for a token rather than a password...?

I see:

```
dave@piran~/S/G/singularity> singularity login docker.io
Password:
```

cmd/internal/cli/login.go
```diff
@@ -0,0 +1,127 @@
...
+			if err != nil {
+				sylog.Fatalf("Failed to read password from stdin: %s", err)
+			}
+			password = strings.TrimSuffix(string(p), "\n")
+			password = strings.TrimSuffix(password, "\r")
+		} else if password == "" {
+			var err error
+
+			password, err = interactive.AskQuestionNoEcho("Password: ")
+			if err != nil {
```
dtrudg, line 103: Should we check for an empty password? If I do a default login with an empty password *and* no username it tells me...

```
dave@piran~/S/G/singularity> singularity login docker.io
Password:
INFO:    Login succeeded
```

... but I haven't really logged in at all here?

dtrudg (commented):
Hi @cclerget - sorry for the delay on this. Functionality wise this is very nice to have. I can login and pull private stuff from docker hub or elsewhere with much less friction now.

I've left a couple of comments left at this stage. One fairly important thing is that if we are including code from that GitHub repo (https://github.com/adigunhammedolalekan/registry-auth) we do need to have some kind of verification that it's okay to distribute as we are doing. It's unfortunate there is no license on the repo.

A more general thought, that I'd like to discuss a bit here....

We now have `singularity login` in this PR which is for OCI/docker registries and also `singularity remote login` which deals with the Sylabs Cloud / Enterprise stuff. I think this is going to be quite confusing to people, and note that we also have a non-Sylabs implementation of the library protocol from @vsoch out in the wild now... and I'm not sure whether that will involve auth in the future.

I was trying to think about some way where perhaps....

- this oci registry stuff might sit under `singularity oci login`
- there is a top level `singularity login` that can redirect to the 'best' place if a remote name, or a host name is provided?

I'm not really sure what I feel here. I do think it's worth a bit of discussion though and would be good to have input from @ikaneshiro @EmmEff @vsoch @tri-adam etc.

vsoch:
Thanks for pinging me @dctrud ! For Singularity Registry server, [singularity remote login](https://singularityhub.github.io/sregistry/docs/client#singularity-push) is indeed the method that we use, and the user provides a token from the web interface (associated with their account after they log in) to authenticate the pushes, and any private pulls. I agree that it could be confusing to have `singularity login/logout`, especially since there is already a precedence for `singularity login <resource>`. Then we would also run into trouble if (at some future point) another login method is needed beyond oci, or if for some reason `singularity login` doesn't make sense anymore for OCI to be the default. What about just doing `singularity oci login` or even `singularity docker login` if most user's are comfortable with understanding docker registries but may not know it's actually OCI? I would hesitate to add `singularity login` without a direct name because it could indeed be the case that a registry serves OCI and a library endpoint. For example, it's been requested for Singularity Registry server but I haven't implemented anything yet.

ikaneshiro:
I think it makes sense to add some additional context to the `login/logout` commands being added via something like `singularity [oci/docker] login`. I prefer `singularity oci login` since we are talking about `oci` registries in general and not just docker registries. My one concern is if this will make the `singularity oci` command group a little overloaded since this behavior is orthogonal to the current commands that manage containers created by the OCI compliant runtime engine within Singularity. If we think that will cause confusion, then I think `singularity docker login` is a sufficient alternative given these commands are mimicking the login behavior of the Docker CLI.

e2e/internal/e2e/docker_auth_server.go
cclerget: The obvious first step was to import this package but there is an issue between go module name not reflecting the github path and I can't figure out how to import it, if you have any idea ?
Will ask unless you know a way to import it ?

cclerget:
I think that `singularity oci login` is out of scope with the oci sub-commands where there is no interaction with registries. 
`singularity docker login` is confusing too. Maybe something like :
- `singularity registry login`
- `singularity oci-registry login`

vsoch:
@cclerget `singularity registry` is the exact name of the Singularity Registry Server, which uses the library endpoint login, so that would be hugely confusing!

cclerget:
@vsoch Where `singularity registry` came from ? Is it open sourced ?

EDIT: nevermind I thought you mentioned something managed by CLI plugin

e2e/internal/e2e/docker_auth_server.go
cclerget: The package is now under MIT license, a PR was submitted to fix the go module issue. I reworked this part to use the package interfaces instead which make this PR a bit lighter now.

cclerget:
My preference goes for `singularity oci-registry login/logout`

dtrudg:
> My preference goes for `singularity oci-registry login/logout`

Prefer this to `docker` as the world moves away from docker === oci

I personally don't think it'd be awful to add it into the `oci` command, but `oci-registry` seems reasonable.

vsoch:
@cclerget rhis has been around for years https://singularityhub.github.io/sregistry/

cclerget:
Closing, follow-up in #5593

PR #5464, opened 2020-07-30
Add configurable default keyserver in singularity.conf

cclerget:

This PR adds two configuration directives :
- `default keyserver` allowing an administrator to set the default keyserver used to pull, push and search for public keys, this server is also used during signature verification.
- `keyserver verify default only` allow the administrator to control during the image verification whether the 'default keyserver' is the only one to be used or not. When set to 'no' the verify command will also reach the keyserver associated with the active remote endpoint of users as the secondary keyserver to use if no key is found on the 'default keyserver'.

### DIFF
--- a/pkg/sypgp/keyring.go
+++ b/pkg/sypgp/keyring.go
@@ -26,30 +26,35 @@ func PublicKeyRing() (openpgp.KeyRing, error) {
 // hybridKeyRing is keyring made up of a local keyring as well as a keyserver. The type satisfies
 // the openpgp.KeyRing interface.
 type hybridKeyRing struct {
-	local openpgp.KeyRing // Local keyring.
-	ctx   context.Context // Context, for use when retrieving keys remotely.
-	c     *client.Client  // Keyserver client.
+	local   openpgp.KeyRing  // Local keyring.
+	ctx     context.Context  // Context, for use when retrieving keys remotely.
+	clients []*client.Client // Keyserver client.
 }
 
 // NewHybridKeyRing returns a keyring backed by both the local public keyring and the configured
 // keyserver.
-func NewHybridKeyRing(ctx context.Context, cfg *client.Config) (openpgp.KeyRing, error) {
+func NewHybridKeyRing(ctx context.Context, cfg []*client.Config) (openpgp.KeyRing, error) {
 	// Get local keyring.
 	kr, err := PublicKeyRing()
 	if err != nil {
 		return nil, err
 	}
 
-	// Set up client to retrieve keys from keyserver.
-	c, err := client.NewClient(cfg)
-	if err != nil {
-		return nil, err
+	var clients []*client.Client
+
+	// Set up clients to retrieve keys from keyserver.
+	for _, clientConfig := range cfg {
+		c, err := client.NewClient(clientConfig)
+		if err != nil {
+			return nil, err
+		}
+		clients = append(clients, c)
 	}
 
 	return &hybridKeyRing{
-		local: kr,
-		ctx:   ctx,
-		c:     c,
+		local:   kr,
+		ctx:     ctx,
+		clients: clients,
 	}, nil
 }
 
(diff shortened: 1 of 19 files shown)


pkg/sypgp/sypgp.go
```diff
@@ -1110,13 +1119,17 @@ func PushPubkey(ctx context.Context, httpClient *http.Client, e *openpgp.Entity,
...
-			return fmt.Errorf("unauthorized or missing token")
+		if jerr, ok := err.(*jsonresp.Error); ok {
+			switch jerr.Code {
+			case http.StatusUnauthorized:
+				// The request failed with HTTP code unauthorized. Guide user to fix that.
+				sylog.Infof(helpAuth+helpPush, e.PrimaryKey.Fingerprint)
+				return fmt.Errorf("unauthorized or missing token")
+			case http.StatusAccepted:
+				return &ErrPushAccepted{jerr.Message}
+			}
```
tri-adam, line 1130: It's quite confusing to have `PushPubkey` return non-`nil` error when pushing a key succeeds.

If I'm reading this right, you are seeing HTTP 202 returned as a success condition from an HKP server? Please submit a bug report against [github.com/sylabs/scs-key-client](https://github.com/sylabs/scs-key-client) to have this fixed properly, rather than hacking around the issue here.

pkg/sypgp/keyring.go
```diff
@@ -92,8 +111,8 @@ func (kr *hybridKeyRing) DecryptionKeys() []openpgp.Key {
 }
 
 // remoteEntitiesByID returns the set of entities from the keyserver that have the given key id.
-func (kr *hybridKeyRing) remoteEntitiesByID(id uint64) (openpgp.EntityList, error) {
-	kt, err := kr.c.PKSLookup(kr.ctx, nil, fmt.Sprintf("%#x", id), client.OperationGet, false, true, nil)
+func (kr *hybridKeyRing) remoteEntitiesByID(c *client.Client, id uint64) (openpgp.EntityList, error) {
```
tri-adam, line 114: It's confusing to have both a slice of `*client.Client` in `kr`, but then call this multiple times with an external `range` over `kr.clients`. It would be cleaner to loop on `kr.clients` here, rather than having to loop in all callers (currently `KeysById` and `KeysByIdUsage`.)

tri-adam (commented):
I have some concerns about this change. This introduces two distinct ways to configure key servers, one through `remote`s, and another through `default keyserver`. This type of duplicity will likely result in confusion, for example:

- Users who are used to using `singularity remote use` will experience breaking changes if their administrator configures a `default keyserver`. 
- Administrators setting a `default keyserver` may not realize they have inadvertently disabled an authenticated key server they configured via `singularity remote add -g`.

In general, I believe we should aim to have a single way to configure a given set of functionality, which will make for a simpler experience for both administrators and users.

I’d suggest this could be accomplished by expanding the `remote` configuration with:

- The ability to override the `remote` configuration to leverage one or more key servers.
- The ability to disable authentication for a given key server.

If/when an administrator configures an unauthenticated remote key server, I think we may need to consider changing the default behaviour of `singularity verify`. At the moment, unless `-l` is specified, Singularity will automatically pull a key from the key server. The Sylabs key server verifies email addresses when keys are uploaded. Many key servers do not do this, however, which increases the chances an attacker could impersonate a user. I would suggest either enabling `verify -l` mode in this case, or printing a warning that makes it clear the user must independently verify the fingerprint with the expected source.

I've added a few comments on the code itself. Given the above, I'll hold off on reviewing the full text of the changes until we're in agreement about the admin/user experience, as I expect the code may change as a result.

dtrudg:
I'll jump in here as well a bit, to emphasize that we absolutely don't want to block the ability to use a different keyserver, but I do agree with @tri-adam that:

- If a keyserver doesn't validate the identity of keys that are published to it in some way, then we don't want singularity to pull keys automatically and unconditionally. Like it or not, people are going to look at the email displayed for a key, and not really go through and check fingerprints out of band.
- We need to clearly define a precedence for keyservers (including any overridden default) that can be configured in a unified manner, and make it apparent to the user what is going on.... which keyserver is being used, if it's not the straightforward case that the remote they have configured is in use. e.g. they are pulling from library:// but an overridden keyserver has been specified that doesn't include the keys for those public images.

I'd also like to suggest that this is the type of change, impacting the UX, which would probably benefit from a simple write up in an issue to briefly describe (1) the use case (2) the proposed change to the UX before implementation. Now we aren't all working directly together in a single company it would help a lot to get on the same page ahead of code review, potentially preventing a bunch of extra work. I'll make an earnest effort to start doing this myself.

cclerget:
Closing, follow-up in #5593

PR #5506, opened 2020-08-18
Pass engine configuration through environment variables instead of a unix socket pair

### LINKED ISSUE #5057: Long command causes FATAL:   Failed to initialize runtime engine: engine "" is not found error
### Version of Singularity:

`singularity version 3.5.2-1.1.el7`

### Expected behavior

Command doesn't throw error

### Actual behavior

Immediate error is thrown upon executing a 125.000 characters long command:

` FATAL:   Failed to initialize runtime engine: engine "" is not found error`

Execute something like this:
singularity exec  bash -c \' long command (>120k chars)\' . If I remove the absolute path from the command, that makes the command so long in the first place, it works, so it must be related to the total command length.

```
NAME="CentOS Linux"
VERSION="7 (Core)"
```

No idea, was done by IT.

cclerget:

- Pass engine configuration through environment variables instead of a unix socket pair to allow bigger configuration to be passed between CLI and runtime. This implementation depends of RLIMIT_STACK current limit and also has a maximum hardcoded size of 1Mb for the engine configuration.

- Fixes missing memory release when joining container.

 - Fixes #5057 

### DIFF
--- a/cmd/starter/c/starter.c
+++ b/cmd/starter/c/starter.c
@@ -1135,30 +1135,6 @@ static void cleanenv(void) {
     }
 }
 
-/*
- * get_pipe_exec_fd returns the pipe file descriptor stored in
- * the PIPE_EXEC_FD environment variable. The returned pipe contains
- * JSON configuration for an engine to run a container.
- */
-static int get_pipe_exec_fd(void) {
-    int pipe_fd;
-    char *pipe_fd_env = getenv("PIPE_EXEC_FD");
-
-    if ( pipe_fd_env != NULL ) {
-        if ( sscanf(pipe_fd_env, "%d", &pipe_fd) != 1 ) {
-            fatalf("Failed to parse PIPE_EXEC_FD environment variable: %s\n", strerror(errno));
-        }
-        debugf("PIPE_EXEC_FD value: %d\n", pipe_fd);
-        if ( pipe_fd < 0 || pipe_fd >= sysconf(_SC_OPEN_MAX) ) {
-            fatalf("Bad PIPE_EXEC_FD file descriptor value\n");
-        }
-    } else {
-        fatalf("PIPE_EXEC_FD environment variable isn't set\n");
-    }
-
-    return pipe_fd;
-}
-
 /* "noop" mount operation to force kernel to load overlay module */
 void load_overlay_module(void) {
     if ( geteuid() == 0 && getenv("LOAD_OVERLAY_MODULE") != NULL ) {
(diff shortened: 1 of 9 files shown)


dtrudg (commented):
Hi @cclerget - Is there a specific reason here to chunk it into environment variables, rather than split into multiple messages sent across the socket?

I was thinking along the lines of something simple like sending a size in the first thing over the socket pair, then the receiver continuing to recv messages until it reaches that... just chunking over the socket instead?

cclerget:
@dctrud Actually CLI process is replaced by runtime process after execve, we can't have both CLI and runtime process communicating through the socket, the code actually take advantage of the kernel socket buffer to temporary store the engine configuration which is consumed by C runtime code using the read side of socket, but the socket buffer is limited in size by `/proc/sys/net/core/wmem_default`, additionally you can't write multiple times because subsequent write blocks or fails with EBUSY until there a process is consuming data on the other end which is impossible in our context.

That's why using environment variables as a storage vector allows to pass potentially big engine configuration as mentioned in #5057, where each environment variable can store up to 131072 bytes (less the bytes used by the environment key and the null byte).

dtrudg:
Thanks - I'll sit down and comb through this tomorrow morning.

mlocal/checks/project-post.chk
```diff
@@ -45,6 +45,12 @@ config_add_def NVIDIALIBS_FILE SINGULARITY_CONFDIR \"/nvliblist.conf\"
 config_add_def SESSIONDIR LOCALSTATEDIR \"/singularity/mnt/session\"
 config_add_def SINGULARITY_SUID_INSTALL $with_suid
 config_add_def PLUGIN_ROOTDIR LIBEXECDIR \"/singularity/plugin\"
+config_add_def ENGINE_CONFIG_ENV \"C\"
+config_add_def ENGINE_CONFIG_CHUNK_ENV \"CC\"
```
dtrudg: Not really a fan of this naming, as `CC=` in the envirionment has another common meaning, even if this is something not exposed to the user.

We lose a few bytes of space, but wouldn't `SINGULARITY_CONFIG` and `SINGULARITY_CONFIG_CHUNKS` be much more obvious for people who might end up tracing things?

cmd/starter/c/starter.c
```diff
@@ -1169,7 +1145,68 @@ void load_overlay_module(void) {
...
+    }
+    nchunk = strtoul(engine_chunk, NULL, 10);
+    if ( nchunk == 0 || nchunk > MAX_ENGINE_CONFIG_CHUNK ) {
+        fatalf("Bad number of engine config chunk provided '%s': 0 or > %d\n", engine_chunk, MAX_ENGINE_CONFIG_CHUNK);
+    }
+
+    for ( i = 0; i < nchunk; i++ ) {
+        /* safety check for env_key in case the hardcoded value is modified */
+        if ( snprintf(env_key, sizeof(env_key), ENGINE_CONFIG_ENV"%d", i+1) != ENGINE_CONFIG_ENV_PADDING-2 ) {
+            fatalf("Maximum engine config chunk >= 10\n");
```
dtrudg: This error message / flow is a little confusing to me. If we have a hard limit on 9 as the `MAX_ENGINE_CONFIG_CHUNK`, and our `ENV_PADDING` needs to match  the env couldn't we instead...

- Compute the padding, rather than rely on it being hard coded
- Check earlier explicitly that `MAX_ENGINE_CONFIG_CHUNK <=9` 
- Check that `nchunk <= MAX_ENGINE_CONFIG_CHUNK`

I _think_ this would have the same effect, and it might make it easier to read the intent of the code?

dtrudg (commented):
Hi Cedric - thanks for the prior explanation. This works well, but I have a couple of questions / thoughts on this r.e. the environment naming and enforcement that I've added. Would be great if you could point out my naivety or otherwise comment on those :-) Thanks.

mlocal/checks/project-post.chk
cclerget: Fixed

cmd/starter/c/starter.c
cclerget: Fixed

dtrudg (approved):
LGTM

ikaneshiro (approved):

PR #5507, opened 2020-08-18
Fix environment variable duplication within singularity engine configuration

cclerget:

Fix environment variable duplication within singularity engine configuration between host variables and SINGULARITYENV_ prefixed variables.

### DIFF
--- a/internal/pkg/runtime/engine/config/oci/generate/generate.go
+++ b/internal/pkg/runtime/engine/config/oci/generate/generate.go
@@ -107,6 +107,21 @@ func (g *Generator) AddProcessEnv(env, value string) {
 	g.Config.Process.Env = append(g.Config.Process.Env, kenv+value)
 }
 
+// RemoveProcessEnv removes a container process environment variable.
+func (g *Generator) RemoveProcessEnv(env string) {
+	g.initProcess()
+
+	kenv := fmt.Sprintf("%s=", env)
+	l := len(kenv)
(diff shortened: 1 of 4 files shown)


dtrudg:
Hi @cclerget - could you explain the issue a bit here? I'm confused looking at the changes to the tests. What is the wrong / correct behavior here? I'm wary of things that further change environment handling beyond the 3.6 release.

Also - if this is a behavior change it will need to go in the CHANGELOG.md. Cheers.

cclerget:
Hi @dctrud , since environment variable rework, there is an additional engine configuration holding all `SINGULARITYENV_` prefixed variables along with the variable passed to container process (basically the one inherited from host), and we have established that `SINGULARITYENV_` variables take precedence over the host variables, so without the fix by example if `FOO=BAR` is defined on host and user set it to a new value with `--env FOO=FOO` or with `SINGULARITYENV_FOO=FOO`, the host is passed through the engine configuration and take some useless space because in any case it will be overridden by `--env FOO=FOO` also embedded in the engine configuration.
The fix just remove the useless environment variable and there is no behavior changes, this is just an internal optimization (which is related to work in #5506) and doesn't impact users.

dtrudg:
Thanks for explaining @cclerget!

dtrudg (approved):
LGTM - I can see how this will save space with envs that could be long & duplicate.

PR #5511, opened 2020-08-18
fix: Correct missed alpine:3.8 pulls to alpine:3.11.5

### LINKED ISSUE #1150: No valid /bin/sh in container
## Version of Singularity:

2.3.1

### Expected behavior

I am trying to set up a ubuntu image to run on a cluster. However, once I try to work within the created image, I get the below listed errors. I tried different ways of reproducing the error and in all ends up with complaining in some way about that /bin/sh is missing...

### Actual behavior

> ERROR  : No valid /bin/sh in container
> ABORT  : Retval = 255

> sudo singularity create -s 4000 im.img
> sudo singularity import ubuntu-xenial-core-cloudimg-amd64-root.tar.gz im.img
> sudo singularity exec --writable im.img apt update

dtrudg:

In previous work to replace test image pulls so that they use sources that are compatible with ARM64, PP64LE architectures we switched from `library://alpine:3.8` -> `library://alpine:3.11.5`.

Unfortunately missed 2 pulls of `alpine:3.8` that did not have the library prefix.

Now that these are gone, also remove a vestigial key pull from the tests that doesn't match with the containers in use.

 - Fixes #1150

### DIFF
--- a/cmd/singularity/pull_test.go
+++ b/cmd/singularity/pull_test.go
@@ -235,7 +235,7 @@ func TestPull(t *testing.T) {
 			},*/
 		{
 			name:            "PullWithHash",
-			sourceSpec:      "library://sylabs/tests/signed:sha256.5c439fd262095766693dae95fb81334c3a02a7f0e4dc6291e0648ed4ddc61c6c",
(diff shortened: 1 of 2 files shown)


tri-adam (approved):
All clear on Ubuntu 20.04. Thanks @dctrud!

PR #5515, opened 2020-08-19
Merge master to 3.6 for 3.6.2

dtrudg:
The current state of the `master` branch is such that it only contains fixes that we want for 3.6.2

Other more major changes slated for 3.7 are currently in review as PRs, so it'd be easier to do a  `master` -> `release-3.6` merge at this point, rather than cherry-pick multiple things.

Are people okay with this? @cclerget @ikaneshiro @tri-adam 

I still have a couple of things to try and mop up for 3.6.2 - but if we can do this merge now before any 3.7 only pieces get into master I'd appreciate it.


ikaneshiro (approved):
LGTM. The only item I see as potentially worth discussion is the inclusion of a new flag in a point release. 

I think this flag is appropriate to include for `3.6.2` as it is an addition and enables automation.

dtrudg:
> LGTM. The only item I see as potentially worth discussion is the inclusion of a new flag in a point release.
> 
> I think this flag is appropriate to include for `3.6.2` as it is an addition and enables automation.

Thanks @ikaneshiro - I'll leave this open for a bit in case anyone has an alternative view on the flag (which is adding a `--force` flag for `singularity delete` so that a library container can be deleted in a scripted manner).

dtrudg:
No other comments - merging.

PR #5518, opened 2020-08-21
fix: Use -user-xattrs in rootless unsquashfs when possible

### LINKED ISSUE #5399: Filesystem extraction fails because of certain packages, no way to pass command line options to unsquashfs
### Version of Singularity: 3.6.0 rc2

```
$ singularity version
3.6.0-rc.2+46-g971f9116b
```

### Actual behavior
I created a container that has libgstreamer installed. When I run
```singularity shell --fakeroot container.sif```
The following error is given:
```
FATAL:   while extracting container-fakeroot.sif: root filesystem extraction failed: extract command failed: 
write_xattr: could not write xattr security.capability for file /tmp/rootfs-319557469/usr/lib/x86_64-linux-gnu/gstreamer1.0/gstreamer-1.0/gst-ptp-helper because you're not superuser!

write_xattr: to avoid this error message, either specify -user-xattrs, -no-xattrs, or run as superuser!
```
I'm guessing there's other software besides gstreamer that will cause this issue, this is just the example that I stumbled upon.

Note that the container works properly *without* `--fakeroot`, with or without `sudo`.

The error recommends using `-user-xattrs` or `-no-xattrs` to solve the problem, but they can't be placed in the `singularity shell` command, and there doesn't seem to be an equivalent.

### Expected behavior
Optimally, Singularity would handle this situation in such a way that it's completely hidden from the user. Running `singularity shell` would just drop you at the shell like it's supposed to. Due to inexperience, I don't know if this is technically possible or practical. The next best solution would be for Singularity to display an error that clearly specifies what has to be added to the shell command for it to work.

Create a recipe file with the following contents:
```
BootStrap: debootstrap
OSVersion: focal
MirrorURL: http://us.archive.ubuntu.com/ubuntu/

%post
DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends libgstreamer1.0-0
 ```
Run `singularity build --fakeroot container.sif recipe`
Then `singularity shell --fakeroot container.sif`

### OS/distro
Ubuntu 20.04 LTS (Focal Fossa)

From source, using the guide detailed here: https://sylabs.io/guides/3.5/admin-guide/installation.html#installation-on-linux
In order to get an unrelated bugfix, I used `master` instead of a specific release. The build was done on June 6, 2020. As far as I know, being 21 days out of date shouldn't affect this issue, but I can try on a more recent build if necessary. 

@dctrud recommended that I create this issue, tagging him to let him know it's up

dtrudg:

In `unsquashfs` 4.4 which is present in Ubuntu 20.04, a return code of `1` is given when it was not possible to extract system xattrs due to running as a non-root user. In 4.2, 4.3 warnings were printed but the return code was `0` still, so Singularity continued without error. In #5399 is a documented way of producing a fatal error on Ubuntu 20.04 due to the 4.4 version of `unsquashfs`. Any `security.selinux` attrs within a SIF file extracted as non-root on a Debian based (non-selinux) system using `unsquashfs` 4.4 will cause this:

```
write_xattr: could not write xattr security.selinux for file bob/sys-xattr because you're not superuser!
write_xattr: to avoid this error message, either specify -user-xattrs, -no-xattrs, or run as superuser
```
    
When we can, use the `-user-xattrs` option if `unsquashfs` is invoked as a non-root user. This is required to avoid a failure when there are system xattrs set on a file that will be extracted.

We have to check to see if we can use it, because `unsquashfs` is 4.0 in RHEL6 and does not include this flag. The usage output that results from the call with the missing flag starts with `SYNTAX`.

Also, provide a file with a security xattr in an e2e test image.. build.. and verify we can run it with `--fakeroot`. This test would fail on Ubuntu 20.04 without the fix.

Note that skipping over system xattrs is already done when extracting docker layer. Umoci will not attempt to set xattrs that are system ones when it is running rootless. Here we ensure SIF -> sandbox doesn't hit issues.

 - Fixes #5399 

### DIFF
--- a/pkg/image/unpacker/squashfs.go
+++ b/pkg/image/unpacker/squashfs.go
@@ -6,12 +6,15 @@
 package unpacker
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
+
+	"github.com/sylabs/singularity/pkg/sylog"
 )
 
 // Squashfs represents a squashfs unpacker.
@@ -60,6 +63,33 @@ func (s *Squashfs) extract(files []string, reader io.Reader, dest string) error 
 		}
 		if err := tmp.Close(); err != nil {
 			return fmt.Errorf("failed to close staging file: %s", err)
+		}
+	}
+
+	// If we are running as non-root, first try with `-user-xattrs` so we won't fail trying
+	// to set system xatts. This isn't supported on unsquashfs 4.0 in RHEL6 so we
+	//  have to fall back to not using that option on failure.
+	if os.Geteuid() != 0 {
+		sylog.Debugf("Rootless extraction. Trying -user-xattrs for unsquashfs")
+		args := []string{"-user-xattrs", "-f", "-d", dest, filename}
+		args = append(args, files...)
+		cmd := exec.Command(s.UnsquashfsPath, args...)
+		if stdin {
+			cmd.Stdin = reader
+		}
+
+		o, err := cmd.CombinedOutput()
+		if err == nil {
+			return nil
+		}
+
+		// Invalid options give output...
+		// SYNTAX: unsquashfs [options] filesystem [directories or files to extract]
+		if bytes.HasPrefix(o, []byte("SYNTAX")) {
+			sylog.Warningf("unsquashfs does not support -user-xattrs. Images with system xattrs my fail to extract")
+		} else {
+			// A different error is fatal
+			return fmt.Errorf("extract command failed: %s: %s", string(o), err)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/image/unpacker/squashfs.go
```diff
@@ -63,6 +66,33 @@ func (s *Squashfs) extract(files []string, reader io.Reader, dest string) error
...
+
+		o, err := cmd.CombinedOutput()
+		if err == nil {
+			return nil
+		}
+
+		// Invalid options give output...
+		// SYNTAX: unsquashfs [options] filesystem [directories or files to extract]
+		if bytes.HasPrefix(o, []byte("SYNTAX")) {
+			sylog.Warningf("unsquashfs does not support -user-xattrs. Images with system xattrs my fail to extract")
```
EmmEff:
```suggestion
			sylog.Warningf("unsquashfs does not support -user-xattrs. Images with system xattrs may fail to extract")
```

EmmEff (changes requested):
Minor typo in error message

EmmEff (approved):
LGTM

ikaneshiro (approved):
lgtm

Issue #5350, opened 2020-06-08
Cannot run singularity sandbox container with inaccessible home directory


DrDaveD:
### Version of Singularity:

What version of Singularity are you using? 

It happens on 3.5.3 and on 3.6.0~rc.4

### Expected behavior

I expect to be able to execute a sandbox image without a working home directory, because it should not have to do any caching.

### Actual behavior

It works when the home directory is missing or is only non-writable, but if the the user has no access permission to the home directory or one of its parent directories, it fails with an error
```
FATAL:   Failed to create an image cache handle: failed to get first existing parent of cache directory: stat /scratch/noperm/dwdtest/.singularity: permission denied
```

### Steps to reproduce this behavior

Change the owner of a home directory to root and make it mode 700, then try to exec a sandbox image within an account of that home directory.

### What OS/distro are you running

CentOS 7.8

### How did you install Singularity

From EPEL.

dtrudg:
@DrDaveD - I can give a partial fix for this... but not a total one at this point.

If we permit a non-accessible home in our cache code then it's okay for non docker/OCI pulls and builds.

Unfortunately if you are building from a docker source, the `containers/image` library goes off to look for the docker config.json in your home directory. When it does that it only handles a non existing file - not a home dir perm error too.

```
[noperms@dev-fedora tmp]$ singularity build -s test docker://alpine
WARNING: 'nodev' mount option set on /tmp, it could be a source of failure during build process
INFO:    Starting build...
FATAL:   While performing build: conveyor failed to get: error getting username and password: error reading JSON file "/home/noperms/.docker/config.json": open /home/noperms/.docker/config.json: permission denied
```

https://github.com/containers/image/blob/22129f19a77890293540e67843f28860f5a77895/pkg/docker/config/config.go#L289

The code that ultimately using `readJSONFile` is called internally in the flow of creating a new docker client in `containers/image` so I don't think it's something we can work around easily. Would need further investigation and probably opening an issue upstream.

Is having things working *except* for docker/OCI pulls any use to you?

DrDaveD:
Yes that would be very helpful.  The problem is when we execute a sandbox (out of cvmfs).  I'm not surprised that a docker pull requires a home directory for the cache.

dtrudg:
Okay - #5519 should be a fix... I'll mark it ready for review when it passes CI tests.

dtrudg:
#5519 is merged.

PR #5519, opened 2020-08-21
fix: Don't fail if parent of cache dir not accessible

dtrudg:

Instead of failing if the parent of the cache dir is not accessible,
just don't use caching.

Fixes part of #5350 - but not for docker / OCI sources as
containers/image will still fail on an inaccessible home dir when trying
to look for the docker config.json for auth.

### DIFF
--- a/internal/pkg/cache/cache.go
+++ b/internal/pkg/cache/cache.go
@@ -256,14 +256,19 @@ func New(cfg Config) (h *Handle, err error) {
 	}
 	h.parentDir = parentDir
 
+	// If we can't access the parent of the cache directory then don't use the
+	// cache.
 	ep, err := fs.FirstExistingParent(parentDir)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get first existing parent of cache directory: %v", err)
+		sylog.Warningf("Cache disabled - cannot access parent directory of cache: %s.", err)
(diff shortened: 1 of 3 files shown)


internal/pkg/cache/cache.go
```diff
@@ -256,9 +256,12 @@ func New(cfg Config) (h *Handle, err error) {
 	}
 	h.parentDir = parentDir
 
+	// If we can't access the parent of the cache directory then don't use the
+	// cache.
 	ep, err := fs.FirstExistingParent(parentDir)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get first existing parent of cache directory: %v", err)
+		h.disabled = true
```
cclerget, line 264: Should we display a warning here or a warning when `SINGULARITY_CACHEDIR` is set to something different than the default ?

dtrudg: I think we've been to and fro over warnings here... some users don't like them, some do. I think a warning would be good and will add one.

dtrudg:
@cclerget - added warnings to the case handles here, and also the non-writable case below. I agree we should warn if the cache is disabled, as the normal behavior is caching.

cclerget (approved):
LGTM

PR #5526, opened 2020-08-24
Merge additional master fixes for 3.6.2

dtrudg:
Last PR to mop up the final fixes for 3.6.2 while we can still do a direct merge.

After this, master will move on, as #5506 is approved and there are behavior change / new features coming along.


cclerget (approved):
LGTM

PR #5555, opened 2020-09-04
Fixes #5554

### LINKED ISSUE #5554: Easy install
### Version of Singularity:

always the newest version mentioned in https://raw.githubusercontent.com/hpcng/singularity/master/CHANGELOG.md

### Expected behavior

```bash
sudo install.sh
```
with install.sh from https://github.com/JoKalliauer/singularity/commit/0a7341aedd675f471f5ef2cf40badfcfa9348ef6

### Actual behavior
```bash

if [ -f "/etc/debian_version" ]; then
 sudo apt-get update && sudo apt-get install -y build-essential libssl-dev uuid-dev libgpgme11-dev squashfs-tools libseccomp-dev wget pkg-config git cryptsetup
fi
if [ "$(grep -Ei 'fedora|redhat' /etc/*release)" ]; then
 sudo dnf update && sudo dnf install -y squashfs-tools wget pkg-config git cryptsetup gcc-go golang-bin
fi

rm -r /usr/local/go
wget https://dl.google.com/go/go1.15.1.linux-amd64.tar.gz #https://golang.org/doc/install
sudo tar -C /usr/local -xzf go1.15.1.linux-amd64.tar.gz
echo 'export PATH=/usr/local/go/bin:$PATH' >> ~/.bashrc && source ~/.bashrc

wget https://raw.githubusercontent.com/hpcng/singularity/master/CHANGELOG.md
pid=$(grep "^# v" ./CHANGELOG.md)
export VERSION=${pid:3:5}

wget https://github.com/sylabs/singularity/releases/download/v${VERSION}/singularity-${VERSION}.tar.gz
tar -xzf singularity-${VERSION}.tar.gz
cd singularity/
./mconfig
make -C ./builddir
sudo make -C ./builddir install
which singularity
singularity --version
cd ..
```

Ask someone unfamiliar to singularity to install the singularity, without telling that the documentation is outdated and installs and old version. 
On Fedroa 32 the person will fail see #5552 , because installing an old version, because the newest available documentation is for 3.6.0.
On Ubuntu 20.04 the person will most likely fail see #5482, because Google shows as a first result https://singularity.lbl.gov/install-linux  which installs 2.5.2

an average user should be able to install it on Ubuntu and Fedora

too complicated to explain

JoKalliauer:
wget https://raw.githubusercontent.com/hpcng/singularity/master/CHANGELOG.md
pid=$(grep "^# v" ./CHANGELOG.md)
export VERSION=${pid:3:5}

 - Fixes #5554

 - Fixes #5554

I did not add myself as a contributor to the [Contributors File](https://github.com/sylabs/singularity/blob/master/CONTRIBUTORS.md) since it is a minor fix.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,27 @@
+#!/bin/sh
+
+if [ -f "/etc/debian_version" ]; then
+ sudo apt-get update && sudo apt-get install -y build-essential libssl-dev uuid-dev libgpgme11-dev squashfs-tools libseccomp-dev wget pkg-config git cryptsetup
+fi
+if [ "$(grep -Ei 'fedora|redhat' /etc/*release)" ]; then
+ sudo dnf update && sudo dnf install -y squashfs-tools wget pkg-config git cryptsetup gcc-go golang-bin
+fi
+
+rm -r /usr/local/go
+wget https://dl.google.com/go/go1.15.1.linux-amd64.tar.gz #https://golang.org/doc/install
+sudo tar -C /usr/local -xzf go1.15.1.linux-amd64.tar.gz
+echo 'export PATH=/usr/local/go/bin:$PATH' >> ~/.bashrc && source ~/.bashrc
+
+wget https://raw.githubusercontent.com/hpcng/singularity/master/CHANGELOG.md
+pid=$(grep "^# v" ./CHANGELOG.md)
+export VERSION=${pid:3:5}
+
+wget https://github.com/sylabs/singularity/releases/download/v${VERSION}/singularity-${VERSION}.tar.gz
+tar -xzf singularity-${VERSION}.tar.gz
+cd singularity/
+./mconfig
+make -C ./builddir
+sudo make -C ./builddir install
+which singularity
+singularity --version
+cd ..


dtrudg (commented):
Hi @JoKalliauer - I appreciate the intent behind this, but I don't think it's something we'd be likely to merge.

`master` is a development branch, and is not the current source for the latest version. The `master` branch `CHANGELOG.md` isn't guaranteed to contain the latest stable version at all times. We branch off released versions into `release-3.x` where x is the current number, and patch release activity might happen there and only be merged into master later. This might occur when we are working in `master` on say the future 3.7 - but we need to do a fix for 3.6 only.

In addition I don't think we want to provide any script like this which is going to `sudo rm -r /usr/local/go` and replace it without warning, or override distro Go through `$PATH` etc. These are likely pretty disruptive things for a lot of people that I don't think should be in a general `install.sh` script that people may run without much consideration of its effects.

I'm generally of the opinion that we should instead...

1. Work to get the user-guide / admin-guide built and deployed automatically so that instructions to get the latest version there are trivial to update immediately on *every* release. This is in our plans.

2. Look at the README.md and INSTALL.md in the repo, and consider changing those so they indicate how to get the latest version, regardless of the state of `master` here etc.

JoKalliauer:
@dctrud  Thanks.

The 3.7 and 3.6-issue could be solved by
```bash
wget https://raw.githubusercontent.com/hpcng/singularity/master/CHANGELOG.md
pid=$(grep "^# v3.6." ./CHANGELOG.md)
export VERSION=${pid:3:5}
```
But yes it is not as easy as I thought, also I knew that things like rm and adding something to .bashrc is problematic and should not be done in this way (maybe check if go is installed and if not (and no folder exitsts) install it).

PS. I'm sure you find a better solution than downloading and using an unneccesarry temporary variable.

PR #5561, opened 2020-09-11
Move CentOS rpm builds from TravisCI to CircleCI

DrDaveD:

This moves the centos6 and 7 rpm tests on TravisCI to centos7 & 8 tests on CircleCI.   I couldn't use the standard docker machine type in CircleCI because it was not given enough privileges to run even the short integration tests, so instead I ported the docker startup script from travis.  I decided to have these run the short integration tests rather than the full set, because the full set was taking too long and also because I was still seeing a couple of the tests fail due to what appeared to be docker restrictions.

None


dtrudg:
@DrDaveD - what tests were failing? The short integration tests don't really check very much at all, so it's not really saying much about the state of the packages.

DrDaveD:
It looks like they are tests having to do with using overlayfs.   Here is a [failed circle build](https://app.circleci.com/pipelines/github/DrDaveD/singularity/200/workflows/c1d033af-170b-48c8-bc3f-ac467b270bff/jobs/738).

dtrudg (approved):
I think I'm happy with this @DrDaveD if you are okay with the short-integration-test not really testing that the rpm is very functional, for the time being. It would be less of a headache with our tests to get away from the sporadic TravisCI failures that are all too common.

Pinging @ikaneshiro and @cclerget for an opinion though, before anything is merged here.

DrDaveD:
I'm ok with the short integration tests not testing everything.  The main thing is to make sure it builds and is able to at least execute something in a container.  Even if we got the overlay-dependent tests working, the tests run for a long time which is another pain when waiting to merge a PR.

dtrudg:
> I'm ok with the short integration tests not testing everything. The main thing is to make sure it builds and is able to at least execute something in a container. Even if we got the overlay-dependent tests working, the tests run for a long time which is another pain when waiting to merge a PR.

I'm not really sure the short-integration tests do that, or can be relied on to keep doing that to be honest. The general intention is that only `e2e` tests should call out to `singularity`. unit-tests and integration-tests should be more contained and not dependent on an installed binary. We aren't there yet, but that has been the stated intent on various testing issues that are open on the repo.

If the need is to ensure a simple container can run, it may be better to just explicitly do that.

dtrudg:
Bit more precision here... the `short-integration` tests are only covering `pkg/network`

```
.PHONY: short-integration-test
short-integration-test:
        @echo " TEST sudo go test [short-integration]"
        $(V)cd $(SOURCEDIR) && \
                scripts/go-test -sudo -v -tags 'integration_test' \
                ./pkg/network
        @echo "       PASS"
```

Take a look at `pkg/network/network_linux_test.go` to see what they run.

DrDaveD:
@dctrud Good point.  I added another commit to add a new 'action-test' make target which runs the action tests without the Persistent Overlay tests.  Those run in just under a minute.   

I was able to reproduce the overlay tests problem under docker on my centos7 VM, but wasn't able to figure out the root cause or a fix for it.  The problem does not happen outside of docker.  I was able to do an overlay mount with mount -t overlay inside the same docker session successfully.  I gave up and decided to work around it with a -skip-overlay option to the action tests.

dtrudg:
Re-pinging @cclerget and @ikaneshiro - do you have any time to look at this briefly, given it is a significant change and HPCng/Ctrl-Cmd had admin control of the CI infrastructure.

ikaneshiro (approved):
LGTM

cclerget (approved):

Issue #5516, opened 2020-08-20
SINGULARITY_SYPGPDIR not being honored


khurtado:
### Version of Singularity:

What version of Singularity are you using? Run:

```
$ singularity version
3.6.0-1.el7

```

### Expected behavior

When setting the `SINGULARITY_SYPGPDIR` environment variable, pgp related files should be writted there, rather than `~/.singularity/sypgp`. 

### Actual behavior

`~/.singularity/sypgp` is created rather than `$SINGULARITY_SYPGPDIR`.

### Steps to reproduce this behavior
Use a .sif image that writes PGP signatures, then:

```
[khurtado@qa-rtx6k-034 ~]$ rm -rf ~/.singularity
[khurtado@qa-rtx6k-034 ~]$ echo $SINGULARITY_CACHEDIR
/tmp/.singularity_khurtado
[khurtado@qa-rtx6k-034 ~]$ echo $SINGULARITY_SYPGPDIR
/tmp/.singularity_khurtado/sypgp
[khurtado@qa-rtx6k-034 ~]$ mkdir -p /tmp/.singularity_khurtado/sypgp
[khurtado@qa-rtx6k-034 ~]$ singularity shell /afs/crc.nd.edu/x86_64_linux/s/singularity/images/pytorch-1.7.sif
Singularity> exit
exit
[khurtado@qa-rtx6k-034 ~]$ ls -l ~/.singularity
total 2
drwx------ 2 khurtado campus 2048 Aug 20 11:44 sypgp
[khurtado@qa-rtx6k-034 ~]$ ls -l $SINGULARITY_SYPGPDIR
total 0
[khurtado@qa-rtx6k-034 ~]$
```

### What OS/distro are you running

```
$ cat /etc/os-release
NAME="Red Hat Enterprise Linux Server"
VERSION="7.8 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VARIANT="Server"
VARIANT_ID="server"
VERSION_ID="7.8"
PRETTY_NAME="Red Hat Enterprise Linux"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:7.8:GA:server"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.8
[...]

PR #5562, opened 2020-09-11
fix: Do not load keyring in prepare_linux if ECL not enabled

### LINKED ISSUE #5558: ECL does not handle missing /home nicely
### Version of Singularity:

```
singularity version 3.6.1-1.el7
```

### Expected behavior

Launch Singularity with ``--no-home`` option and be able to execute the container on a host
where ``$HOME`` is not mounted:

```
~$ singularity pull /tmp/python.sif docker://python:latest
~$ cd /tmp
~$ singularity --verbose exec --no-home /tmp/python.sif python --version
VERBOSE: HOME=/home/mtds                                          
VERBOSE: Set messagelevel to: 4
VERBOSE: Starter initialization
VERBOSE: Check if we are running as setuid
VERBOSE: Drop root privileges
VERBOSE: Drop root privileges permanently      
VERBOSE: Spawn stage 1                          
VERBOSE: Execute stage 1                                  
VERBOSE: stage 1 exited with status 0                                                                          
VERBOSE: Get root privileges     
VERBOSE: Change filesystem uid to 1001                              
VERBOSE: Spawn master process  
VERBOSE: Create mount namespace
VERBOSE: Entering in mount namespace      
VERBOSE: Drop root privileges
VERBOSE: Create mount namespace          
VERBOSE: Spawn RPC server
VERBOSE: Execute master process
VERBOSE: Serve RPC requests                                          
VERBOSE: Default mount: /proc:/proc    
VERBOSE: Default mount: /sys:/sys
VERBOSE: Default mount: /dev:/dev
VERBOSE: Found 'bind path' = /etc/hosts, /etc/hosts
VERBOSE: Found 'bind path' = /etc/localtime, /etc/localtime
VERBOSE: Default mount: /tmp:/tmp
VERBOSE: Default mount: /var/tmp:/var/tmp
VERBOSE: Default mount: /etc/resolv.conf:/etc/resolv.conf
VERBOSE: Default mount: /tmp: to the container
VERBOSE: Checking for template passwd file: /var/singularity/mnt/session/rootfs/etc/passwd
VERBOSE: Creating passwd content
VERBOSE: Creating template passwd file and appending user data: /var/singularity/mnt/session/rootfs/etc/passwd
VERBOSE: Default mount: /etc/passwd:/etc/passwd
VERBOSE: Checking for template group file: /var/singularity/mnt/session/rootfs/etc/group
VERBOSE: Creating group content
VERBOSE: Default mount: /etc/group:/etc/group
VERBOSE: Skipping mount /tmp, /var/singularity/mnt/session/final/tmp/mtds already mounted
VERBOSE: rpc server exited with status 0
VERBOSE: Execute stage 2
Python 3.8.5
```
The previous command worked perfectly till version _3.5.3_.

### Actual behavior

```
~$ cd /tmp
~$ singularity --verbose exec --no-home /tmp/python.sif uname -a
VERBOSE: Setting HOME=/home/mtds
VERBOSE: Setting PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
VERBOSE: Set messagelevel to: 4
VERBOSE: Starter initialization
VERBOSE: Check if we are running as setuid
VERBOSE: Drop root privileges
VERBOSE: Drop root privileges permanently
VERBOSE: Spawn stage 1
VERBOSE: Execute stage 1
FATAL:   while obtaining keyring for ECL: mkdir /home: permission denied
VERBOSE: stage 1 exited with status 255
```

It looks like Singularity is still looking for my home directory (set to be ``/home/mtds``) in any case, even when
the ``--no-home`` parameter is specified. This will not allow us to run containers as batch jobs on nodes which
by default do not mount the home directories of our users.

```bash
[root@lxdev01 ~]# singularity pull /tmp/python.sif docker://python:latest
[root@lxdev01 ~]# adduser --no-create-home foo
[root@lxdev01 ~]# su foo
bash-4.2$ ls $HOME
ls: cannot access /home/foo: No such file or directory
bash-4.2$ cd /tmp
bash-4.2$ singularity -v exec --no-home /tmp/python.sif python --version
VERBOSE: Setting HOME=/tmp
VERBOSE: Setting PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
WARNING: can't determine current working directory: stat .: permission denied
VERBOSE: Set messagelevel to: 4
VERBOSE: Starter initialization
VERBOSE: Check if we are running as setuid
VERBOSE: Drop root privileges
VERBOSE: Drop root privileges permanently
VERBOSE: Spawn stage 1
VERBOSE: Execute stage 1
WARNING: can't determine current working directory
FATAL:   while obtaining keyring for ECL: mkdir /home/foo: permission denied
VERBOSE: stage 1 exited with status 255
```
Same result is obtained also when using the ``--home`` parameter and pointing it to ``/tmp``.

```
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
ANSI_COLOR="0;31"

```
```
~$ uname -r
3.10.0-1127.13.1.el7.x86_64
```

We did build an RPM package from upstream directly.

dtrudg:

Avoid attempting to load the public keyring for ECL handling when the
ECL is not enabled. Take the approach of passing through a null keyring
so ECL apply/ignore logic remains in the ECL package and is not
duplicated in the runtime code.

This is an effective fix for #5558 when ECL is not enabled.

Note: this does not address:
- Deeper issues in the way sypgp code handles paths - such as #5516. We need to fix env var handling, and ideally be able to differentiate between a non-fatal failure for certain read-only operations on a missing keyring, vs fatal failures.
- An e2e test for missing/inaccessible home dir. Because of the way the e2e framework sets up a homedir tmpfs mount in a namespace that is shared by things in each suite.Run() call it's not trivial to test home dir issues by manipulating perms / deleting the home dir etc. We need a way to do this to avoid regressions and I'll open a tech debt issue for it.

 - Fixes #5558

### DIFF
--- a/internal/pkg/runtime/engine/singularity/prepare_linux.go
+++ b/internal/pkg/runtime/engine/singularity/prepare_linux.go
@@ -41,6 +41,7 @@ import (
 	"github.com/sylabs/singularity/pkg/util/fs/proc"
 	"github.com/sylabs/singularity/pkg/util/namespaces"
 	"github.com/sylabs/singularity/pkg/util/singularityconf"
+	"golang.org/x/crypto/openpgp"
 	"golang.org/x/sys/unix"
 )
 
@@ -1166,9 +1167,16 @@ func (e *EngineOperations) loadImages(starterConfig *starter.Config) error {
 				return fmt.Errorf("while validating ECL configuration: %s", err)
 			}
 
-			kr, err := sypgp.PublicKeyRing()
-			if err != nil {
-				return fmt.Errorf("while obtaining keyring for ECL: %s", err)
+			// Only try to load the keyring here if the ECL is active.
+			// Otherwise pass through an empty keyring rather than avoiding calling
+			// the ECL functions as this keeps the logic for applying / ignoring ECL in a
+			// single location.
+			var kr openpgp.KeyRing = openpgp.EntityList{}
+			if ecl.Activated {
+				kr, err = sypgp.PublicKeyRing()
+				if err != nil {
+					return fmt.Errorf("while obtaining keyring for ECL: %s", err)
+				}
 			}
 
 			if ok, err := ecl.ShouldRunFp(img.File, kr); err != nil {


internal/pkg/runtime/engine/singularity/prepare_linux.go
```diff
@@ -1166,9 +1166,16 @@ func (e *EngineOperations) loadImages(starterConfig *starter.Config) error {
...
 
-			kr, err := sypgp.PublicKeyRing()
-			if err != nil {
-				return fmt.Errorf("while obtaining keyring for ECL: %s", err)
+			// Only try to load the keyring here if the ECL is active.
+			// Otherwise pass through an empty keyring rather than avoiding calling
+			// the ECL functions as this keeps the logic for applying / ignoring ECL in a
+			// single location.
+			kr := sypgp.NullKeyRing()
+			if ecl.Activated {
```
tri-adam, line 1175: Arguably, if this is `false`, we don't need to call `ecl.ShouldRunFp` at all, and thus wouldn't need a keyring? What about something like: 

```
			if ecl.Activated {
				kr, err := sypgp.PublicKeyRing()
				if err != nil {
					return fmt.Errorf("while obtaining keyring for ECL: %s", err)
				}

				if ok, err := ecl.ShouldRunFp(img.File, kr); err != nil {
					return fmt.Errorf("while checking container image with ECL: %s", err)
				} else if !ok {
					return errors.New("image prohibited by ECL")
				}
			}
```

tri-adam (changes requested):
One suggestion, thank you.

internal/pkg/runtime/engine/singularity/prepare_linux.go
dtrudg: Now modified per discussion elsewhere,

tri-adam (approved):

PR #5574, opened 2020-09-15
Improved nvidia driver selection consistency when using nvidia-container-cli

ajohannesson:

Libraries returned by nvidia-container-cli always point (full path) to the correct drivers.
Therefore it should be completely unnecessary to try and match the driver basename to what has been cached by ldconfig, and instead it should be fine to just go straight for the driver returned by the container-cli

By avoiding the ld cache, this approach can fix an issue with a platform containing multiple active drivers with the same name, for example two implementations of the nvidia-tls drivers, one in /usr/lib64/libnvidia-tls.so and another in /usr/lib64/tls/libnvidia-tls.so
Needlessly leaving it to the ldcache (which is stored as an unordered map in runtime) will result in indeterminate selection of conflicting drivers, leading to random launch-time segfaults of gpu-driven applications

Also, by explicitly locating all symlinks to each library (within the same dir as fully qualified lib), we ensure the runtime environment inside the singularity runtime is as representative of the gpu driver topology as reasonably possible

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -16,6 +16,7 @@
     - Amanda Duffy <aduffy@lenovo.com>
     - Ana Guerrero Lopez <aguerrero@suse.com>
     - Ángel Bejarano <abejarano@ontropos.com>
+    - Aron Öfjörð Jóhannesson <aron1991@gmail.com>
     - Bernard Li <bernardli@lbl.gov>
     - Brian Bockelman <bbockelm@cse.unl.edu>
     - Carl Madison <carl@sylabs.io>
--- a/pkg/util/gpu/paths.go
+++ b/pkg/util/gpu/paths.go
@@ -47,9 +47,21 @@ func nvidiaContainerCli(args ...string) ([]string, error) {
 			// this is a library -> add full path to library
 			libs = append(libs, line)
 
-			// also add full path to library without .xxx.xx suffix
+			// also add full path to all linked libraries of similar name
 			bareLibPath := strings.SplitAfter(line, ".so")[0]
-			libs = append(libs, bareLibPath)
+			linkedLibCandidates, _ := filepath.Glob(fmt.Sprintf("%s*", bareLibPath))
+			for _, linkedLibCandidate := range linkedLibCandidates {
+				if fi, err := os.Lstat(linkedLibCandidate); err == nil {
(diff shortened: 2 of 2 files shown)


dtrudg:
Hi @ajohannesson - thank you for this contribution. We'll try to get to a proper review of the code next week. From a first quick look though, it would be good to refactor some of the deeply nested `if` statements, as I think some of the code is now nested 8 brackets deep.

Also there are some cases where errors are tested, but there is no log message or other handling in the case that there is an errors. Generally if we can't process a glob pattern or lstat a file etc. then there's probably something wrong with permissions or the FS and we'd want to at least log a warning so that it's apparent something unusual has happened that may result in the container bound libs not matching what was intended.

Thanks again - we'll come back to this soon.

ajohannesson:
Hi @dctrud those are good and valid points, it could have been written up in an easier to follow manner and closer to the overall style.
[9d8fdc8](https://github.com/hpcng/singularity/pull/5574/commits/9d8fdc8ccb58f4b805cdcdabaa483187bcf3755a) both splits out the deep conditional chain and adds warnings in the event of unexpected errors when looking up file info

dtrudg:
something strange happening with the CI here that it didn't report properly. Will force it to re-run.

dtrudg:
@ajohannesson - apologies. Our CircleCI is only seeing the old 4248abd commit. It seems there was some issue with GitHub <-> CircleCI communication when you made further changes. Could you make a new trivial commit to try to trigger the CI again?

ajohannesson:
@dctrud how odd, but no problem, pushed a non-change and looks like circleci is picking up the correct commit now

dtrudg:
Thanks - now the CircleCI e2e-tests failed with a networking issue reaching dokerhub, so re-running! We'll get there soon :-)

dtrudg:
@ajohannesson - thanks again for this contribution. Apologies for the delay - I'll schedule some time to get on both NVIDIA and ROCm machines this week to okay it for merge.

dtrudg (approved):
Okay - decided to grab the time now instead of doing something else  :-)

This is working for me across NVIDIA with/without the container-cli and ROCm. I think we'll re-factor the code a bit more before the 3.7 release, but that need is from the existing structure, not what's added here.

Thanks @ajohannesson !

PR #5609, opened 2020-10-12
Update golangci-lint and make version check more permissive

### LINKED ISSUE #5597: Make golangci-lint version in make check a bit more permissive
When `make` check is run, it will only proceed with an exact `golangci-lint` version, which it will install if not present. It wants `1.21.0` at present, which is quite outdated, and unlikely to be installed except for Singularity development.

This is very inconvenient when working on other code and using a newer `golangci-lint`. We should enforce the strict version in CI, and install it if `golangci-lint` is not already present.... but how about we allow the `make check` to run if a version already present is newer than what we specify?

dtrudg:

The restriction to using only the old 1.21.0 version of golangci-lint
for make check is inconvenient, as developers are unlikely to have that
version on their system.

Update to the latest version, and make the check for golangci-lint
version permissive. Warn instead of error if the version doesn't match
the one we will install, and is used in CI.
    
New golangci-lint errors instead of warns on deadcode analysis failure. On MacOS we were trying to use undefined things / incompatible types, paritcularly in e2e. I've made various minimal changes so that the linting passes. This does not mean that tests work other than on Linux.

 - Fixes #5597


ikaneshiro (approved):

dtrudg:
Thanks @ikaneshiro - looking at the commit log here... I'll squash the fixes down a bit into a couple of related commits before commiting

PR #5614, opened 2020-10-14
Loopback changes

locula:

This PR mainly changes the way that a new loopback device is found. Instead of going through all of the loopback devices one by one and checking if each one is free, we now use the function LOOP_CTL_GET_FREE to find a new loopback device. If the device found can't be opened, then we will repeat the process to find another free loop device and continue until we find one that we can use. 
We also change the maximum number of loop devices from 256 to 10,000 and fix a bug that makes the host become unresponsive after allocating a large number of loop devices.

 - Fixes #

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,12 @@ Singularity 3.6.3 addresses the following security issues.
     enable arbitrary code execution during the build and/or when the
     built container is run.
 
+## Change defaults / behaviours
+
+    - Change the default number of loop devices from 256 to 10000
+    - Use LOOP_CTL_GET_FREE to get a new free loop device instead of checking each
+      loop device to see if it is free
+
 ## Bug Fixes
 
   - Add CAP_MKNOD in capability bounding set of RPC to fix issue with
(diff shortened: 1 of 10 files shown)


dtrudg:
Hi @locula - I just saw this PR appear and closed in a short timescale. If you would like to pursue this could you please open an issue for discussion of changes to the loop device code?

We don't use `LOOP_CTL_GET_FREE` because we still work on distributions with older kernels than `3.1`, when it appeared.

Also, the shared loop device functionality isn't something that would be removed without discussion, as it was added as a useful feature for people running a lot of instances of the same container on large machines.

Thanks!

dtrudg (changes requested):
Hi @locula - please could you open an issue, or issues, describing how the current behaviour is limiting, and to provide a venue for discussion. As in the comment above there are concerns around kernel version compatibility, utility of the shared loop device code etc.

locula:
The tests are currently failing because this fix does not take into account shared loop devices. We will revisit this PR once we have updated the code to allow for shared loop devices.

DrDaveD:
At this point this is quite stale.  Please open a new PR when you're ready.

PR #5618, opened 2020-10-15
Restore umask in container, actions, not fakeroot

### LINKED ISSUE #5214: Containers do not inherit user umask into running process
Singularity version 3.5.3, running on Ubuntu 18.04LTS with go 1.13.8; Installed from source (I can give detailed instructions if necessary).

When creating a directory, default permissions are different inside and outside the container. The used container is a basic debian container one additional package installed.

See below for example:
```
mark@turing:/BIP/_Mark$ mkdir test_outside
mark@turing:/BIP/_Mark$ singularity exec /containers/elastix.sif mkdir test_inside
mark@turing:/BIP/_Mark$ ls -lh | grep test
drwxr-xr-x  2 mark BIP    0 Apr 16 11:41 test_inside
drwxrwx---  2 mark BIP    0 Apr 16 11:40 test_outside
```

Further investigation shows that the umask inside and outside the container environment is different: outside I explicitly set it to 007, inside it is 022.

Unfortunately for my users this is a bit confusing, as they did not expect this different behavior.

### Expected/actual behavior
Expected: Container inherits umask from the outside world.

Actual: umask is set to 022 inside container.

I saw some discussion about this in #1465 with the suggestion to open a new issue, so here it is :)

Edit: I saw some workaround involving changing the environment scripts inside the containers, but I'd like to avoid this if possible.

dtrudg:

Restore umask in container, actions, not fakeroot
    
When we are running through an actions flow, and are *not* using
fakeroot then by default save the umask from the calling environment and
set it within the container.
    
This ensures that when a user works inside a container then files they
create on host filesystems have the permissions that they expect, given
their umask.
    
Add a flag `--no-umask` to disable this propagation.

 - Fixes #5214

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,17 @@ _With the release of `v3.0.0`, we're introducing a new changelog format in an at
 
 _The old changelog can be found in the `release-2.6` branch_
 
+# Changes since v3.6.4
+
+## Changed defaults / behaviours
+
+  - When actions (run/shell/exec...) are used without --fakeroot the
+    umask from the calling environment will be propagated into the
+    container, so that files are created with expected permissions.
+    Use the `--no-umask` flag to return to the previous behaviour of
+    setting a default 0022 umask.
+
+
 # v3.6.3 - [2020-09-15]
 
 ## Security related fixes
--- a/cmd/internal/cli/action_flags.go
+++ b/cmd/internal/cli/action_flags.go
@@ -48,6 +48,7 @@ var (
 	NoInit          bool
 	NoNvidia        bool
 	NoRocm          bool
+	NoUmask         bool
 	VM              bool
 	VMErr           bool
 	NoNet           bool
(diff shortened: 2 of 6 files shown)


cclerget (approved):
LGTM, just asking why fakeroot behaves differently here ?

dtrudg:
> LGTM, just asking why fakeroot behaves differently here ?

Hi @cclerget - my reasoning for not restoring the umask in the fakeroot engine, was that when you start a container as the `fake root` user you are explicitly not trying to act in the same way as your normal user account on the host system. You are becoming a 'different user' at least in the container view. Therefore, a standard `0022` umask makes more sense.

Very happy to consider circumstances this doesn't make sense, and change it if you think it makes sense for fakeroot to propagate the user umask too. Might be something I can bring up on the slack / mailing list / original issue(s).

cclerget:
Regarding fakeroot my first guess was to also propagate umask because user/group ID are same at filesystem level, but @dctrud mentioned in private the case where a umask of `0027` in a container as fake root user could cause issues when creating files or building image in locations which used by other users making them unusable.

dtrudg:
@cclerget - Going to merge this but leave the issue held open for comment for a while. If there's a call to have same behavior in fakeroot there it's easy to implement.

PR #5659, opened 2020-10-26
Verify SIF sources at build, optional Fingerprint check

### LINKED ISSUE #2748: Check if base container is signed when building from def file

### LINKED ISSUE #5150: verify base image in a def file
When building a container, I'd like to be able to do a `singularity verify` and then bomb out if the base container is untrusted.  Here is an extremely ugly and brittle way to do it in a def file:
```
Bootstrap: library
From: alpine:sha256.63259fd0a2acb88bb652702c08c1460b071df51149ff85dc88db5034532a14a0

%pre
    # this is going to delete your cache to make sure you are getting a minty
    # fresh image from the library
    rm -rf  /root/.singularity/cache/library

%setup
    # insert the fingerprint here
    KEY_FP=8883491F4268F173C6E5DC49EDECE4F3F38D871E
    MESSG="Data integrity verified"
    singularity verify /root/.singularity/cache/library/sha256*/* | \
        grep -q $KEY_FP
    singularity verify /root/.singularity/cache/library/sha256*/* | \
        grep -q "$MESSG"

%post
    echo "build here"
```
It works.  But it makes the baby Jesus cry.  👶 ✝️ 😢 

In the long term, it would be great if this functionality were built into Singularity.  Maybe a new keyword in the def file called `Fingerprint:`.
```
Bootstrap: library
From: my_signed_container
Fingerprint: 8883491F4268F173C6E5DC49EDECE4F3F38D871E
```
^Importantly this would cause the build to fail if the container was unsigned **OR** if it was signed but had an incorrect fingerprint.  

In the near term it would just be nice to have an env var or something available at build time that would point to the location of the downloaded SIF file so I could dispense with deleting the cache and whatnot.  

**BONUS:**
It would be cool if the `verify` command had a `--fingerprint` option and if verify would just return a 0 or 1 exit code if the fingerprint does or doesn't match.  Something like this:
```
$ singularity verify --fingerprint GOODPRINT123 my.sif

$ echo $?
0

$ singularity verify --fingerprint BADPRINT321 my.sif

$ echo $?
1
```

dtrudg:
### Description of Pull Request

When we run a build, and a SIF image is the bootstrap source, run a verify on it and warn for failures similar to how we do with `push`.
    
Add a new definition file header `Fingerprint:` that enforces the source image must verify, with a signature that matches the supplied fingerprint, for the build to proceed.

 - Fixes #2748
 - Fixes #5150


dtrudg:
Example of WIP...
```
dave@dev-fedora~> cat test_invalid.def 
Bootstrap: localimage
From: test.sif
Fingerprint: AA5D629F4C2A7602963A2F5766D8269BB5014355

dave@dev-fedora~> singularity build -F -f test_invalid.sif test_invalid.def
INFO:    Starting build...
INFO:    Checking bootstrap image verifies with fingerprint AA5D629F4C2A7602963A2F5766D8269BB5014355
FATAL:   While performing build: conveyor failed to get: while checking fingerprint: image not signed by required entities

dave@dev-fedora~> cat test_valid.def 
Bootstrap: localimage
From: test.sif
Fingerprint: 325D629F4C2A7602963A2F5766D8269BB5014355

dave@dev-fedora~> singularity build -F -f test_valid.sif test_valid.def
INFO:    Starting build...
INFO:    Checking bootstrap image verifies with fingerprint 325D629F4C2A7602963A2F5766D8269BB5014355
INFO:    Creating SIF file...
INFO:    Build complete: test_valid.sif
```

dtrudg:
Ready for a first pass review. Would be grateful for your input @GodloveD if you have a chance to try `Fingerprint:` in the def file.

cclerget:
`Fingerprint` keyword addition while convenient for users removes some flexibility. A keyword like `Verify` to specify verify command options could also bring `--fingerprint` option in `singularity verify` command which would be useful for scripting purpose. We would end with something like :

```
bootstrap: library
from: image
verify: --url https://keys.example.com --fingerprint 3C9407D0E9B4BC7D3090A367E099A5F2EB715B78
```

The build code could just call singularity executable with those options.

It would also be useful to allow injection of public keys from the definition file, we could start building with an empty local keyring dedicated to build without interference with the user local keyring by using `SINGULARITY_SYPGPDIR`, the approach of always using an empty keyring would structure `verify` header options, if an author set the `--local` option, it would force him to import the public key from the definition file, so a build could be reproduced by other people without worrying about the local keyring.
Globally for a reproducible build, it would be recommended to either provide the public key or specify the keyserver URL.

The build code could automatically inject a shell function as part of the `%pre` section:

```
import_pubkey() {
   singularity key import /dev/stdin
}
```

So a user could import a public key directly from the definition file to provision the temporary keyring before verify step

```
%pre
    import_pubkey <<EOF
-----BEGIN PGP PUBLIC KEY BLOCK-----

xsFNBF8qhY0BEAC72vrtkQwa0GKOteDUiGrug1VfCfnkoVdWoKlLuUaz35QkphKr
11LKhVtnwzgdx6SRTxdwJP++F9kDFOZ4wH4NqQMqLcDgpmDqUN8bljuACSkwdA/R
TtAQ2x6GNcMuIJIxnw4dbD1KQhy+LTwLECZWlsLKHFYQ9QadNVTlblNKecwEz7QM
h0MjDlBiNohK+/SC80cADX3EBOWomGoPPVLLCwnfpthRSRag5SmUWBmFFdtb5gJY
cItC81ECd6foERHaSvYmPp72Bq42eL8QvilfPsvPAJePYFAkZsfw/pbdgo8gRDhB
lQEdslyLjnlqZzO5WKj3HXxSBNOuRT/n+nszhy5Pl90lbDY63zWCyu1aYG7uOP5a
jcn4Rfe+e1T73B4fW+OEWU8YV7ctKK2eiLIlSWfCyQx+GfcMabFSd+mGbbCyMlO3
84wS6GvwLTlvpbRGGFwfvmLfjwUpD9jAUGwZ819vxhy2tGhBkr59YemIg3Nq8kgl
tSctqD1U6OaAwnmla+yBtTgCbUWq9R+zTsiqC5JMmC9nauGc8n9P+umUFsdXRlrG
ZBHZm3NSoowHS0EI37FoYJSicUxJiTdCWSokGU6UClw/v6kaFdVuXk6s47xxfiL8
C0BwIVL2+N3JnoUpvu9+ReHnOlgkqkpoJst3UPb5gE5n2AiDtMSKGqAsOQARAQAB
zRJKb2huIERvZSA8Sm9obkRvZT7CwWUEEwEJABkFAl8qhY0JEOCZpfLrcVt4AhsD
AhkBAhUJAAAsSBAAJSxn9qAGHGHejg/gvGZyu0N4GKVN2UAve82HxK4Pvwe1oHrE
OzG2jPt64Yc1w000144P4An2D4b7Ci+pbTWP8kOixAKJGLoHzGb1Aal4beIr9O95
QBwPfXu5wJgiG6iik1Z6fVS3/+FUCAHvOsxE3N66eL0qJ3rFAADjKdHv6AplFzwQ
/4Z++BE1VPxRCp9DBMiVZKl4VrgfWewWuRZoBes01+sw4/ejUKMCQc9YmK5g6qG9
N45lP7saq2609GU/FXUc61AmHQVwPz3fuqXvZj8Qn1dQ6CtE1uVHlpIm0N/u8xFA
hix8uwDMGF6ZI9mYIiOBI+u3QHUqQG2wI3CovceqTYUKI/m9+/xib0qxhFkogEVl
p50LgqR5lxv7YBIWxt3hYo3fRp0YRJK5gWcljjM74vwvYybfrIB2oMH6jgErp3jO
BoUWAYmwOIjFnNywGtuhwaAuu3ujJ7w365DdyEZgoJ+eLdmS4Ky8fJGZ7p0zMYu2
YNLIq+UioBDtdA0zDFNnivgZsTq3bgDchnhEhf4cP0ZAv810VWZdYIB1/uW3pcCX
qd9kIITN/Gfu5PmzHsP69mcmkExhNUiD3+pIagohc+3LImdj1LTgWEpFyhPfKYZN
tOVseCoMSNCi0avWsowPI6V1mk8HodxwALDTPoA59KgOEqOZAe49yeDLb7HOwU0E
XyqFjQEQAM+e+hpUUNmjWJiqjt/FxRQgYRh4owUI682ehwtiwZkgfS77Dh+tQDAl
aNabFobTdH2xxRYLZ6rEw8lpiT6eBhWbflii1EeLBjjXir2fKGpvLCcmtwKXKi53
FX9sCtihcVxc0x8nnlC74BIx7Tg9if1gv1bj0YW3uF4j/zX9zhUQlrk1Ik1zToeZ
ezEbAGI0oYZ7zD5wghYRUW3X6tugRIDiZMfaIwTlOlD2Ha99d7sIAuQy8i2/T6sv
YqQ4CjZiYh69g/9ag7/RG1nDB9djFTeaoHtbMeoy09+F0fTAPagcwv9UNjE1tiKM
bjbbgrrkbD+EkKON0ZEUQJUi4glNCPndEsitn3zJYvMNplP3CDcfo2ja63eoNdE5
JbpxS3wG/mKegcl7SRSCcNR1UQytSiWnChsUEjPhg2UAK3ESyPjzQ7ykoEndAm7G
kMZ0/lE+MK4CRq7K7wTVxn7qMVAvdE8r6QqIOdOB8jJUTjrlRkDBirLXWq/J/TNu
L4YwBlbphPYIf7n1pR/mIwOSfrvKzJ7b8g6Y6BZdrIhMgnUf0EDJTWr17JGAD5zb
T6WwsfaioJxzTISctjcF2+Do4qDugOvsXztKucQOO2wrGRHEr/2aQKU5jqJligb0
nyojQgK1VMWI4oDYZ0YLmV6YXGUOBb9dmeo+44uEFkwQUDcYlH3HABEBAAHCwV8E
GAEJABMFAl8qhY0JEOCZpfLrcVt4AhsMAAA2ZRAAKXMKRkAsNIVDkD5GmQlYsJjQ
MKGwkQhbwTmugIm0Y3JsQO0x8aZeM6Oi9KyKXqAiH/P3KeXKSGaHqJJmrexD2/OX
YN3u5MKrOPfqH2tW+hu9fUgLIs5H3ANoLRQ8G99Ndqmr2UCE3EBAJ4quo/rZfOZj
cPOi4byCpFqK17KRLPGHMXE5OL7jDTYRjOmsMH2NqvZhfeJRlec3DnumznGj3jKi
0qlA+t8jCUHHCopGdWLbCuUm1XjvcnEcYZx2hAFuFA8/691ur3ErTi02yxLm2n3w
rC4pK4IwCzEzBKK16k+FsfbXEYNBUbmeCvNNkDILh40nA/6rsvfilZ2zKLRW1oQ5
fH8noYeTqnBy7fGzNTrAbfEuY3AmKL/86ixNb74yPeL58Bz+H3Smpuz8w9708pq7
/ykWNMDTyDHObnaTioMHxnHri3zwO3hTSOoEpjyIYjOgQw7Tb/C8XWsxxbwquKE7
91A9JTP7As+MwUgyHNQZvaxiccmz/c5c9RwFbgcgFSW0eN1mWUlikBgmJEI3EnMJ
HSb7daXkoKsqHj23qF7Voe4Wus02jCXWB/nydE0ul8FnTgI0hSHszOxL6CPqyOtE
n4j3ZIYNXZf4DjMrSiFuoqQTDYcfQ7fnu6EBsY6T9V8aJNUYsn8qfAaZ/SFruX4u
vcZsQ9+9ZLKi1eYcH6s=
=sJzJ
-----END PGP PUBLIC KEY BLOCK-----
EOF
```

dtrudg:
Hi @cclerget - thanks for the comments. 

> `Fingerprint` keyword addition while convenient for users removes some flexibility. A keyword like `Verify` to specify verify command options could also bring `--fingerprint` option in `singularity verify` command which would be useful for scripting purpose. We would end with something like :
> ```
> bootstrap: library
> from: image
> verify: --url https://keys.example.com --fingerprint 3C9407D0E9B4BC7D3090A367E099A5F2EB715B78
>```
>
>The build code could just call singularity executable with those options.

I'm inclined to believe that a very simple `Fingerprint: xxx` in the build would be most obvious to the user, so more likely to be used - which we want to encourage. I had some chat with @GodloveD around this idea, which originated in an issue he posted, and the thinking was that it would be expected that the fingerprint check would work like a plain `verify xxx.sif`... so use the existing keyring(s) / remote config etc.

I really do strongly believe we need to favour simplicity for the 90% of use cases, vs flexibility for the 10%. More advanced users can script around it to handle key imports etc. before builds. We really want people who are less experienced to have a very simple way to add some additional checks in builds, when they feel it is useful, and is easy to reason about.

I will also say I am not very positive about continuing to add features through singularity calling itself. It becomes very hard to reason about the flow of execution when we do this, and debugging is much harder. Perhaps it points to the code not being modular enough to reuse pieces of the flow as needed, without an exec - which I'd prefer we tackled head on than exec-ing `singularity`. That doesn't preclude adding a `--fingerprint` option for verify on its own though.

> It would also be useful to allow injection of public keys from the definition file, we could start building with an empty local keyring dedicated to build without interference with the user local keyring by using SINGULARITY_SYPGPDIR, the approach of always using an empty keyring would structure verify header options, if an author set the --local option, it would force him to import the public key from the definition file, so a build could be reproduced by other people without worrying about the local keyring.
Globally for a reproducible build, it would be recommended to either provide the public key or specify the keyserver URL.

I personally don't believe builds should inject keys. Keys that are permitted for verification on the system are a property of the system / user's decisions, not the build. I know it's possible working with `%setup` etc. to manipulate that... but I don't think there should be an official route to it.

As there's a bit of difference in approach here, and the likelihood is that the best idea maybe somewhere in the middle, I'd like to ask others - @GodloveD / @tri-adam / @ikaneshiro / @EmmEff for their thoughts so we can hopefully get this worked through for 3.7.0.

Thanks again for looking at this so quickly!

GodloveD:
So, as I shared in #5150, I have a janky way of doing this already, but it makes assumptions about the cache.  Those assumptions broke with the last minor version bump and will likely break again in the future.  So I really just want a "blessed" implementation that will not change in future releases so that my def files don't break.  

For my own selfish needs I don't really care too much about the implementation details or def file syntax.  Anything would be simpler that what I have right now.  atm, I just copy and paste the exact same def file template at the beginning of each new container I build so difficult syntax is not an issue.  And I expect the def files to propagate via the containers themselves (through inspect) and in places like git repos so others probably won't trip over complicated syntax either.  

So any way you slice this is fine by me.  Just as along as there is a solution that is not likely to change in the future.  And if it were documented, that would also be a bonus.  😄

tri-adam:
> `Fingerprint` keyword addition while convenient for users removes some flexibility. A keyword like `Verify` to specify verify command options could also bring `--fingerprint` option in `singularity verify` command which would be useful for scripting purpose. We would end with something like :
> ```
> bootstrap: library
> from: image
> verify: --url https://keys.example.com --fingerprint 3C9407D0E9B4BC7D3090A367E099A5F2EB715B78
> ```

This actually raises a couple of security concerns for me:

1. Consider the case where Alice owns private key `3C9407D0E9B4BC7D3090A367E099A5F2EB715B78`. Mallory obtains Alice's private key, and starts publishing malicious images signed with it. Much to her dismay... Alice finds out that someone has been impersonating her. She revokes her key, and pushes that to the public and private keyservers she and her colleagues use. From this point, Mallory chooses to start distributing Singularity definition files, to continue to propogate her malicious base images. She had saved Mallory's non-revoked key, stands up a basic HTTP server to serve it, and includes the URL in the definition. Since the definition allows control over where the key material is sourced, she has nullified the revocation.
1. This is easily leveraged to inject a "tracker" into a build process. By supplying a key server that I control, I force your build machine to reach out to my key server every time that definition is executed.

All that to say... from a security point of view, I believe there is value in _not_ allowing someone to override the key server from a definition. It keeps the way keys are retrieved consistent and under the control of the user and/or administrator.

ikaneshiro:
Adding my opinion into the mix to move this forward: 

I think its best to keep the complexity of this feature low and only have something like `Fingerprint: xxx` to verify an image is signed by the key of the listed fingerprint. I agree with @tri-adam that the proposed calling of the `verify` command within the build system adds complexity that leaves room for misunderstood or unexpected behavior that could be cause for concern for some users or sites. I also agree with @dctrud that we shouldn't officially "bless" importing keys to a users keyring via a build definition, though it has always been possible.

If I could add one bit of feedback, it would be to allow multiple fingerprints to be specified for an image in the case a user or org is maintaining images with multiple signatures. So, something like `Fingerprints: xxx, yyy`, where the fingerprints form a strict whitelist.

dtrudg:
@ikaneshiro - that's a good point on the multiple fingerprints as a strict whitelist, to ensure multiple signatures that confer different meaning. Will work up that tomorrow and mark for review.

dtrudg:
Ready for a first review!

cmd/internal/cli/build_linux.go
```diff
@@ -267,6 +271,11 @@ func runBuildLocal(ctx context.Context, cmd *cobra.Command, dst, spec string) {
 		authToken = lc.AuthToken
 	}
 
+	kc, err := getKeyserverClientConfig(buildArgs.keyServerURL, endpoint.KeyserverVerifyOp)
+	if err != nil {
+		sylog.Fatalf("Unable to get library client configuration: %v", err)
```
tri-adam: Typo:

```suggestion
		sylog.Fatalf("Unable to get key server client configuration: %v", err)
```

cmd/internal/cli/build_darwin.go
```diff
@@ -31,12 +31,13 @@ func runBuild(cmd *cobra.Command, args []string) {
 		sylog.Fatalf("Only remote builds are supported on this platform")
 	}
 
-	bc, lc, err := getBuildAndLibraryClientConfig(buildArgs.builderURL, buildArgs.libraryURL)
+	bc, lc, kc, err := getServiceConfigs(buildArgs.builderURL, buildArgs.libraryURL, buildArgs.keyServerURL)
 	if err != nil {
 		sylog.Fatalf("Unable to get builder and library client configuration: %v", err)
```
tri-adam: Maybe just generalize this?

```suggestion
		sylog.Fatalf("Unable to get service configuration: %v", err)
```

internal/pkg/build/sources/verify_sif.go
```diff
@@ -0,0 +1,36 @@
...
+// verifySIF checks whether a bootstrap SIF image verifies
+func verifySIF(imagePath string, kc *scskeyclient.Config) error {
+	sylog.Infof("Verifying bootstrap image %s", imagePath)
+	opts := []singularity.VerifyOpt{}
+	if kc != nil {
+		opts = append(opts, singularity.OptVerifyUseKeyServer(kc))
+	}
+	ctx := context.TODO()
+	return singularity.Verify(ctx, imagePath, opts...)
+}
```
tri-adam: Might be worth opening a tech debt issue to addd contexts here? These are both using a key server some of the time, which is potentially a long-running operation.

pkg/build/types/parser/deffile_test.go
```diff
@@ -215,7 +217,8 @@ func TestParseDefinitionFile(t *testing.T) {
 			}
 
 			if !reflect.DeepEqual(defTest, defCorrect) {
-				t.Fatal("parsed definition did not match reference")
+				b, _ := json.Marshal(defTest)
+				t.Fatalf("parsed definition did not match reference, %s", string(b))
```
tri-adam: Perhaps consider using `(*testing.T).Log()` for `b`, to make it easier to read the actual failure?

tri-adam (changes requested):
Just a few minor requests/comments, thanks!

pkg/build/types/parser/deffile_test.go
dtrudg: Ahh - yeah, this was a quick debug change.... I will tidy it it up though rather than remove it as it is useful when working in this area.

internal/pkg/build/sources/verify_sif.go
dtrudg: Agreed - I'll put a // TODO comment and open issues. I think we quite a way below where we have a context... but we did some work in other area which would be good to reprise here.

dtrudg: On second look - I'll fix this properly.... it's not too far away.

dtrudg:
Should be good for another pass @tri-adam 

@ikaneshiro / @cclerget - Ad has had a good look through this. If you are able to take a look it'd be great so @tri-adam can then rebase #5689 once merged.

cmd/internal/cli/build_linux.go
```diff
@@ -115,7 +117,9 @@ func runBuildRemote(ctx context.Context, cmd *cobra.Command, dst, spec string) {
 		sylog.Fatalf("Building encrypted container with the remote builder is not currently supported.")
 	}
 
-	bc, lc, err := getBuildAndLibraryClientConfig(buildArgs.builderURL, buildArgs.libraryURL)
+	// TODO - the keyserver config needs to go to the remote builder for fingerprint verification at
+	// build time to be fully supported.
+	bc, lc, _, err := getServiceConfigs(buildArgs.builderURL, buildArgs.libraryURL, buildArgs.keyServerURL)
```
ikaneshiro, line 122: This should probably be tracked in a tech debt issue?

ikaneshiro (approved):

tri-adam (approved):
LGTM

cmd/internal/cli/build_linux.go
dtrudg: Will do.

dtrudg: https://github.com/hpcng/singularity/issues/5695

Issue #5615, opened 2020-10-14
New way to find free loop device


locula:
### Version of Singularity:

What version of Singularity are you using?

```
$ singularity version
3.5.2-1.1.el7.20.deshaw
```

### Expected behavior

I expect that when looking for a free loop device, we should always be able to find one if there are free loop devices available.

### Actual behavior

What actually happened? Why was it incorrect?

Currently, if we run two containers at the same time and both containers are looking for a free loop device, they may both pick the same loop device thinking that it is free. In this case, the second containers fails because when it tries to use the loop device, it is already taken by the first container. To resolve this, we want to use LOOP_CTL_GET_FREE to get a new loop device. We also want to increase the maximum number of loop devices from 256 to 10,000 to accommodate for greater loop device usage.

dtrudg:
> To resolve this, we want to use LOOP_CTL_GET_FREE to get a new loop device.

Hi @locula - the issue here is understood. However, as mentioned in comments on the #5614 that you opened, we are still currently supporting running on distros with a kernel <3.1, which is where `LOOP_CTL_GET_FREE` was introduced.

Also, the shared loop device cannot leverage `LOOP_CTL_GET_FREE`, and is a feature that a number of deployments are making use of, to share loop devices where they run many instances of the same container on a single machine.

If you would like to propose a change, please could you ensure the above is either accommodated, or you explain why you believe changes are appropriate.

> We also want to increase the maximum number of loop devices from 256 to 10,000 to accommodate for greater loop device usage.

As a default this would be a very large number, that probably doesn't make sense to have as a default. Most sites will not accumulate more than 256 loop devices on a machine, and larger than that number would likely indicate a problem somewhere. Please could you give more information, as an issue report, of the circumstances that require this very large number for your install.

Finally...

> $ singularity version
3.5.2-1.1.el7.20.deshaw

It looks like this is a version that has likely been customized in house at deshaw. Please could you make sure the issue occurs with the latest open source code, in version 3.6.4 without any modifications, and report here if true. It would be useful to show a simple way to replicate the problem with debug output too, so it can inform review of any code changes.

Thank you.

locula:
> Hi @locula - the issue here is understood. However, as mentioned in comments on the #5614 that you opened, we are still currently supporting running on distros with a kernel <3.1, which is where `LOOP_CTL_GET_FREE` was introduced.

[...]

Issue #2316, opened 2018-10-29
Add CI check for buildcfg dependency for any packages in pkg/


bauerm97:
In order to make sure that all packages in `pkg` will be usable to external projects, we'll add a CI check to enforce that `buildcfg` only be used in strictly internal packages. 

`buildcfg` cannot be a direct OR indirect dependency of *any* package in `pkg/`.

bauerm97:
Can be accomplished via the following command: 

`! go list -f '{{.Deps}}' ./pkg/... | grep "buildcfg"`

carterpeel:
Hello,

This is a templated response that is being sent out to all open issues.  We are working hard on 'rebuilding' the Singularity community, and a major task on the agenda is finding out what issues are still outstanding.


**Please consider the following:**

1. Is this issue a duplicate, or has it been fixed/implemented since being added?
2. Is the issue still relevant to the current state of Singularity's functionality?
3. Would you like to continue discussing this issue or feature request?


Thanks,
Carter

pedroalvesbatista:
@bauerm97 Still getting this problem ? Have this been solved already ? If yes, what work around have you followed and applied ?

We're looking into the issue carefully, soon will bring to community and discuss ways to better solve as well address this. Thankyou for keeping the interest in the subject.

kmuriki:
Moved to the new Apptainer repo https://github.com/apptainer/apptainer/issues/1157

PR #5660, opened 2020-10-26
Use MaxLoopDevices variable from config file

locula:

This PR makes use of the MaxLoopDevices variable defined in the config file in all the places where the maximum number of loop devices is referenced. This makes it easy for users to change this value as they choose.

 - Addresses #5615

### DIFF
--- a/internal/pkg/build/sources/packer_ext3_linux.go
+++ b/internal/pkg/build/sources/packer_ext3_linux.go
@@ -18,7 +18,7 @@ import (
 	"github.com/sylabs/singularity/pkg/image"
 	"github.com/sylabs/singularity/pkg/sylog"
 	"github.com/sylabs/singularity/pkg/util/loop"
-	"github.com/sylabs/singularity/pkg/util/singularityconf"
+	// "github.com/sylabs/singularity/pkg/util/singularityconf"
 )
 
 // Pack puts relevant objects in a Bundle!
@@ -42,8 +42,9 @@ func unpackExt3(b *types.Bundle, img *image.Image) error {
 
 	var number int
 	loopdev := &loop.Device{
-		MaxLoopDevices: int(singularityconf.GetCurrentConfig().MaxLoopDevices),
-		Info:           info,
+		MaxLoopDevices: 256,
+		// MaxLoopDevices: int(singularityconf.GetCurrentConfig().MaxLoopDevices),
+		Info: info,
 	}
 
 	if err := loopdev.AttachFromFile(img.File, os.O_RDONLY, &number); err != nil {
--- a/internal/pkg/build/sources/packer_ext3_linux.go
+++ b/internal/pkg/build/sources/packer_ext3_linux.go
@@ -14,12 +14,10 @@ import (
 	"os/exec"
 	"syscall"
 
-	"github.com/sylabs/singularity/internal/pkg/buildcfg"
 	"github.com/sylabs/singularity/pkg/build/types"
 	"github.com/sylabs/singularity/pkg/image"
 	"github.com/sylabs/singularity/pkg/sylog"
 	"github.com/sylabs/singularity/pkg/util/loop"
-	"github.com/sylabs/singularity/pkg/util/singularityconf"
 )
 
 // Pack puts relevant objects in a Bundle!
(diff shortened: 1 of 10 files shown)


pkg/util/loop/loop_linux_test.go
```diff
@@ -11,9 +11,27 @@ import (
 	"syscall"
 	"testing"
 
+	"github.com/sylabs/singularity/internal/pkg/buildcfg"
 	"github.com/sylabs/singularity/internal/pkg/test"
+	"github.com/sylabs/singularity/pkg/util/singularityconf"
 )
 
+func getMaxLoopDevices() int {
```
dtrudg: This `getMaxLoopDevices` is repeated in many locations - please instead define it once as an exported/Public function (probably in `pkg/util/loop` is a good place - and then call it from other places it is needed.

pkg/util/crypt/crypt_dev.go
```diff
@@ -36,11 +38,27 @@ var (
 	ErrInvalidPassphrase = errors.New("no key available with this passphrase")
 )
 
+func getMaxLoopDevices() int {
```
dtrudg: This `getMaxLoopDevices` is repeated in many locations - please instead define it once as an exported/Public function (probably in `pkg/util/loop` is a good place - and then call it from other places it is needed.

pkg/ocibundle/tools/loop.go
```diff
@@ -9,14 +9,32 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/sylabs/singularity/internal/pkg/buildcfg"
 	"github.com/sylabs/singularity/pkg/util/loop"
+	"github.com/sylabs/singularity/pkg/util/singularityconf"
 )
 
+func getMaxLoopDevices() int {
```
dtrudg: This `getMaxLoopDevices` is repeated in many locations - please instead define it once as an exported/Public function (probably in `pkg/util/loop` is a good place - and then call it from other places it is needed.

internal/pkg/build/sources/packer_ext3_linux.go
```diff
@@ -31,6 +33,22 @@ func (p *Ext3Packer) Pack(context.Context) (*types.Bundle, error) {
 	return p.b, nil
 }
 
+func getMaxLoopDevices() int {
```
dtrudg: This `getMaxLoopDevices` is repeated in many locations - please instead define it once as an exported/Public function (probably in `pkg/util/loop` is a good place - and then call it from other places it is needed.

dtrudg (changes requested):

dtrudg:
@locula - Apologies - in my comment before I didn't think through that these functions are bringing an `internal/` dependency into the supposedly reusable (from other projects) `pkg` code.

@cclerget - do you have any input on this? The general intent seems fine to me. However it does introduce the issue that things in `pkg/xxx` no depend on the `internal/pkg` config handling - so they aren't really independent of a Singularity install.

I think we probably need to look at how stuff could be shuffled around to avoid this hard `internal/` dep from `pkg/` if at all possible? I know we have #2316 which has been open for a long time to address this, and would hope we can make progress there rather than steps back.

cclerget:
@dctrud @locula In order to deal with `buildcfg` dependency from a public package, the easiest approach is to create two files in `pkg/loop` and use go build tags :

- loop_singularity.go

```
// +build singularity_engine

package loop

import (
	"github.com/sylabs/singularity/internal/pkg/buildcfg"
	"github.com/sylabs/singularity/pkg/util/singularityconf"
)

func GetMaxLoopDevices() int {
	// if the caller has set the current config use it
	// otherwise parse the default configuration file
	cfg := singularityconf.GetCurrentConfig()
	if cfg == nil {
		var err error

		configFile := buildcfg.SINGULARITY_CONF_FILE
		cfg, err = singularityconf.Parse(configFile)
		if err != nil {
			return 256
		}
	}
	return int(cfg.MaxLoopDevices)
}
```

- loop_no_singularity.go

```
// +build !singularity_engine

package loop

func GetMaxLoopDevices() int {
        // externally imported package, use the default value
	return 256
}
```

This way external project would use `loop_no_singularity.go`.

> I think we probably need to look at how stuff could be shuffled around to avoid this hard internal/ dep from pkg/ if at all possible? I know we have #2316 which has been open for a long time to address this, and would hope we can make progress there rather than steps back.

Currently two packages are using `buildcfg` as indirect dependency: `pkg/build/types` and `pkg/util/crypt` and would require some modification before adding this check

cclerget (changes requested):
Require to manage `buildcfg` internal dependency as described in https://github.com/hpcng/singularity/pull/5660#issuecomment-724054468

locula:
@dctrud I've removed the internal dependencies and updated the code. Is there any additional feedback for this?

dtrudg:
@locula - we'll come back to this soon after the 3.7.0 release. I'm afraid things are a bit busy in that regard at present.

locula:
Hi, I just wanted to follow up and check if there are any updates on this.

cclerget (approved):
LGTM, thanks @locula ! And sorry for the delay

dtrudg (approved):

Issue #5616, opened 2020-10-14
Refactor NVIDIA / ROCm lib handling code


dtrudg:
#5574 refined the behaviour of the GPU lib handling code, but during review it was clear that the prior existing code needs a refactor. It's far too deeply nested, with long functions that should be split up for clarity of intent and implementation.

Will tackle for 3.7.

PR #5671, opened 2020-10-28
Try /sbin/ldconfig for gpu libs if $PATH ldconfig fails

### LINKED ISSUE #5002: --nv --rocm won't find correct ldconfig when run from Nix userspace

dtrudg:

On systems with Nix/Guix type userland on top of a 'standard'
distribution there may be a non-working userland `ldconfig` on `$PATH`
that will fail to find the system installed CUDA libs. If we fail to
execute `ldconfig` on `$PATH` then fall back to trying the default
`/sbin/ldconfig` in case that works.

 - Fixes #5002

### DIFF
--- a/pkg/util/gpu/paths.go
+++ b/pkg/util/gpu/paths.go
@@ -162,10 +162,6 @@ func paths(gpuFileList []string) ([]string, []string, error) {
 	// walk through the ldconfig output and add entries which contain the filenames
 	// returned by nvidia-container-cli OR the nvliblist.conf file contents
 	ldConfig, err := exec.LookPath("ldconfig")
-	if ee, ok := err.(*exec.Error); ok && ee.Err == exec.ErrNotFound {
-		sylog.Debugf("Could not find ldconfig in PATH")
-		ldConfig = "ldconfig"
-	}
 	if err != nil {
 		return nil, nil, fmt.Errorf("could not lookup ldconfig: %v", err)
 	}
--- a/pkg/util/gpu/paths.go
+++ b/pkg/util/gpu/paths.go
@@ -19,6 +19,8 @@ import (
 	"github.com/sylabs/singularity/pkg/sylog"
 )
 
+const systemLdconfig = "/sbin/ldconfig"
+
 // nvidiaContainerCli runs `nvidia-container-cli list` and returns list of
 // libraries, ipcs and binaries for proper NVIDIA work. This may return duplicates!
 func nvidiaContainerCli(args ...string) ([]string, error) {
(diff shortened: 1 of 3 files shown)


pkg/util/gpu/paths.go
```diff
@@ -168,6 +170,11 @@ func paths(gpuFileList []string) ([]string, []string, error) {
 		return nil, nil, fmt.Errorf("could not lookup ldconfig: %v", err)
 	}
 	out, err := exec.Command(ldConfig, "-p").Output()
+	// #5002 If we failed and our ldconfig is not in standard POSIX location, try that
+	if err != nil && ldConfig != systemLdconfig {
```
cclerget, line 170: Not related to your changes but while we are here ... after looking at the code above, there is a strange path :
```
        ldConfig, err := exec.LookPath("ldconfig")
	if ee, ok := err.(*exec.Error); ok && ee.Err == exec.ErrNotFound {
		sylog.Debugf("Could not find ldconfig in PATH")
		ldConfig = "ldconfig"
	}
	if err != nil {
		return nil, nil, fmt.Errorf("could not lookup ldconfig: %v", err)
	}
```

which let think about a code artifact like :
```
        ldConfig, err := exec.LookPath("ldconfig")
	if ee, ok := err.(*exec.Error); ok && ee.Err == exec.ErrNotFound {
		sylog.Debugf("Could not find ldconfig in PATH")
		ldConfig = "ldconfig"
	} else if err != nil {
		return nil, nil, fmt.Errorf("could not lookup ldconfig: %v", err)
	}
``` 

Removing the useless statement `ldConfig = "ldconfig"` would avoid this confusion

dtrudg: ahh - good catch, that is an odd flow there.

dtrudg:
@cclerget - rebased and dropped the ineffective error check block. As mentioned on slack I have #5616 open as a 'refactor' task for this code, and also want to add some e2e there that will run on GPU equipped systems.

cclerget (approved):
LGTM

PR #5697, opened 2020-11-05
Add --no-mount to override (disable) singularity.conf mount options

### LINKED ISSUE #5081: Ability to disable "mount hostfs" in command line

dtrudg:
**NOTE**: This is proposed as an alternative to the single `--no-hostfs` in #5086, but I'm very open to any of:

1. Merging this `--no-mount` approach.
2. Adding the single requested `--no-hostfs` by re-opening #5086
3. Adding all of the individual `--no-xxx` flags for each `mount xxx` config option by modifying this PR.

My opinion r.e. the previous discussion in #5086 is that as we have an existing `--no-xxx` pattern for flags, the proposal in `#5086` comments of a `--mount ^xxxx` negation syntax is confusing. Also if we allow positive overrides via a `--mount` we need to worry about precedence / behavior when user binds are disabled etc. For positive overrides (mounting things disabled in `singularity.conf`) `--bind/-B` is already available, with understood semantics.

It has been requested that users can turn off mount options set to `yes`
in `singularity.conf` as they may interfere with some containers -
especially when `mount hostfs = yes` is set.

In #5086 a single `--no-hostfs` flag was introduced but concern was
raised that we should be able to disable all, and this would lead to a
proliferation of flags. Another viewpoint is that we should have all of
the `--no-xxxxx` flags.

This PR proposes a middle ground. Add a `--no-mount` flag that can take
one or more 'mounts' to disable, overriding the `singularity.conf`
value. I feel this is more clear / friendlier than using `--mount ^hostfs`
negation syntax.

Implemented through individual NoXXXX booleans on the EngineConfig so it'd
be trivial to add the `--no-xxxx` CLI flags later if we change our
minds.

 - Fixes #5081

### DIFF
--- a/cmd/internal/cli/actions_linux.go
+++ b/cmd/internal/cli/actions_linux.go
@@ -138,6 +138,31 @@ func hidepidProc() bool {
 	return false
 }
 
+// Set engine flags to disable mounts, to allow overriding them if they are set true
+// in the singularity.conf
+func setNoMountFlags(c *singularityConfig.EngineConfig) {
+	for _, v := range NoMount {
+		switch v {
+		case "proc":
+			c.SetNoProc(true)
+		case "sys":
+			c.SetNoSys(true)
+		case "dev":
+			c.SetNoDev(true)
+		case "devpts":
+			c.SetNoDevPts(true)
+		case "home":
+			c.SetNoHome(true)
+		case "tmp":
+			c.SetNoTmp(true)
+		case "hostfs":
+			c.SetNoHostfs(true)
+		default:
+			sylog.Warningf("Ignoring unknown mount type '%s'", v)
+		}
+	}
+}
+
 // TODO: Let's stick this in another file so that that CLI is just CLI
 func execStarter(cobraCmd *cobra.Command, image string, args []string, name string) {
 	var err error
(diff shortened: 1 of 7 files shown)


cmd/internal/cli/actions_linux.go
```diff
@@ -138,6 +138,31 @@ func hidepidProc() bool {
 	return false
 }
 
+// Set engine flags to disable mounts, to allow overriding them if they are set true
+// in the singularity.conf
+func setNoMountFlags(c *singularityConfig.EngineConfig) {
```
cclerget, line 143: Adding `cwd` could be also useful

cclerget (commented):
Overall it looks good, disabling `cwd` could be nice too as actually it requires to execute singularity from a denied directory like `/` to disable it

dtrudg:
> Overall it looks good, disabling `cwd` could be nice too as actually it requires to execute singularity from a denied directory like `/` to disable it

Added a `cwd` exclude in ac2aa0c

cclerget (approved):
LGTM

PR #5763, opened 2020-12-16
Use -buildmode=default on ppc64le

### LINKED ISSUE #5762: Use -buildmode=default (not pie) on ppc64le
### Version of Singularity:

3.7

### Actual behavior

See: https://github.com/golang/go/issues/43228

dtrudg:
-buildmode=pie can cause SIGSEGV or unknown caller pc errors with
plugins on ppc64le. Use the default non-pie build mode which avoids
these errors.

Fixes #5762


tri-adam (approved):
LGTM

PR #5812, opened 2021-02-03
fix: allow pull with --library https://library.sylabs.io

### LINKED ISSUE #5808: Cannot override with `--library` *back* to default library URL
### Version of Singularity:

3.7.x

### Expected behavior

If using a non-default remote endpoint it should be possible to pull from the default Sylabs Cloud by setting `--library https://library.sylabs.io` 

### Actual behavior

The pull will be made against the configured remote library, not the library.sylabs.io URL specified with `--library`.

This is due to logic in: https://github.com/hpcng/singularity/blob/01a985fc5c39e022304a597862df060326216451/cmd/internal/cli/singularity.go#L612

... where the library URL passed downward is cleared if it matches the default Sylabs cloud URL.

```
isDefault := uri == endpoint.SCSDefaultLibraryURI
...
if isDefault {
	uri = ""
}
```

It should not be cleared if we are using a different remote, and then requesting the default library URI by flag.

dtrudg:
If the current remote is not the default Sylabs cloud endpoint we should be able to use `--library https://library.sylabs.io` to pull from the default library. This worked prior to remote changes in 3.7.

Refactored the flag handling for the `--library` `--builder` and keyserver `--url` flags to accomplish this goal, and improve clarity of the code.

Previously these flags defaulted to the URI of the relevant service at the sylabs cloud endpoint, making it superficially look like those URIs would be used for subsequent operations. Within `getXXXClientConfig`, however, we reset `uri` to `""` if the uri had not been modified from the default. The `""` which is then passed lower down means that the URLs for the _configured endpoint_ are used for operations.

Refactoring so that `""` is the default value for the flags makes it clearer that they are for overriding the current endpoint, and by default we are not overriding the current remote endpoint. It also allows us to override with `--library https://library.sylabs.io` if we are *not* using the default endpoint. This was impossible with the previous logic, as we always cleared the URL if it matched the production service, e.g.

```
isDefault := uri == endpoint.SCSDefaultLibraryURI
...
if isDefault {
	uri = ""
}
```

With the above code in place, the uri could never be overridden to the default URI for Sylabs cloud.

    
Fixes #5808

### DIFF
--- a/cmd/internal/cli/build.go
+++ b/cmd/internal/cli/build.go
@@ -135,7 +135,7 @@ var buildDetachedFlag = cmdline.Flag{
 var buildBuilderFlag = cmdline.Flag{
 	ID:           "buildBuilderFlag",
 	Value:        &buildArgs.builderURL,
-	DefaultValue: endpoint.SCSDefaultBuilderURI,
+	DefaultValue: "",
 	Name:         "builder",
 	Usage:        "remote Build Service URL, setting this implies --remote",
 	EnvKeys:      []string{"BUILDER"},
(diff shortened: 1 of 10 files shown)


EmmEff (approved):
LGTM

PR #5888, opened 2021-03-24
Implement support for host in library reference

EmmEff:
Implement support for using library ref "library://&lt;hostname&gt;/&lt;entity&gt;/&lt;collection&gt;/&lt;container&gt;[:tag]" while maintaining legacy support for "library://&lt;entity&gt;/&lt;collection&gt;/&lt;container&gt;[:tag]".

This allows the end-user to override the currently selected remote (similar to the `--library <hostname>` argument on `singularity pull`), for example.

Exclusive setting is honoured to disable host name override if system-wide exclusive flag is set.

### DIFF
--- a/e2e/pull/pull.go
+++ b/e2e/pull/pull.go
@@ -197,6 +197,21 @@ var tests = []testStruct{
 		force:            true,
 		expectedExitCode: 0,
 	},
+
+	// pulling with library URI containing host name
+	{
+		desc:             "library URI containing bad host name",
+		srcURI:           "library://notlibrary.sylabs.io/library/default/busybox:1.31.1",
+		library:          "https://notlibrary.sylabs.io",
+		expectedExitCode: 255,
+	},
+	{
+		desc:             "library URI containing host name",
+		srcURI:           "library://library.sylabs.io/library/default/busybox:1.31.1",
+		library:          "https://library.sylabs.io",
+		force:            true,
+		expectedExitCode: 0,
+	},
 }
 
 func (c *ctx) imagePull(t *testing.T, tt testStruct) {
--- a/e2e/pull/pull.go
+++ b/e2e/pull/pull.go
@@ -198,17 +198,23 @@ var tests = []testStruct{
 		expectedExitCode: 0,
 	},
 
+	// pulling with library URI containing host name and library argument
+	{
+		desc:             "library URI containing host name and library argument",
+		srcURI:           "library://notlibrary.sylabs.io/library/default/busybox:1.31.1",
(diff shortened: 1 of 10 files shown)


dtrudg (commented):
Edit - sorry jumped the gun on the LGTM when I remembered about the shortening options. Left a separate comment below on that.

Would be great if you could...

- Open an issue on https://github.com/hpcng/singularity-userdocs/issues to remind us to document this.
- Add e2e test cases that mirror the ones for the `--library` option, but using the remote in ref format: https://github.com/hpcng/singularity/blob/1eba636703fd1531c94b3f40e7822f302f21b9f0/e2e/pull/pull.go#L186

internal/pkg/client/library/library.go
```diff
@@ -24,27 +24,37 @@ const defaultTag = "latest"
...
+	slashes := strings.Count(ref, "/")
 
-	// parse library ref
-	r, err := scslibrary.Parse(validLibraryRef)
-	if err != nil {
-		return fmt.Errorf("error parsing library ref: %v", err)
+	var elem, tags []string
+	var host, pathref string
+
+	if slashes == 0 || slashes == 2 {
```
dtrudg: I think this changes behavior? At present you can have slashes ==1 when you ommit the entity (in this example `library`) but specify `collection/container`

```
12:34 PM $ singularity pull library://default/ubuntu
743.6KiB / 26.8MiB [=>-----------------------------------------------------------------------] 3 % 413.1 KiB/s 1m4s
```

With this PR...

```
12:38 PM $ singularity pull library://default/ubuntu
FATAL:   While pulling library image: error fetching image: error making request to server:
	Get "https://default/v1/images/ubuntu:latest?arch=amd64": dial tcp: lookup default: no such host
```

e2e/pull/pull.go
```diff
@@ -197,6 +197,21 @@ var tests = []testStruct{
...
+	{
+		desc:             "library URI containing bad host name",
+		srcURI:           "library://notlibrary.sylabs.io/library/default/busybox:1.31.1",
+		library:          "https://notlibrary.sylabs.io",
+		expectedExitCode: 255,
+	},
+	{
+		desc:             "library URI containing host name",
+		srcURI:           "library://library.sylabs.io/library/default/busybox:1.31.1",
+		library:          "https://library.sylabs.io",
```
dtrudg: I think this is going to error due to having a hostname in the URI *and* the `--library` flag?

dtrudg (changes requested):
I think the e2e is going to error - otherwise LGTM

dtrudg (approved):

PR #5913, opened 2021-04-09
Add overlay command to ease the creation of ext3 writable overlay

cclerget:

Add overlay command to ease the creation of ext3 writable overlay
- allow to create an overlay as a single ext3 image file.
- allow to create and add an ext3 writable overlay to an existing SIF image.

SIF restrictions:
- SIF image must not contain a writable overlay
- SIF image must not be signed
- SIF image must not be encrypted

### DIFF
--- a/cmd/internal/cli/overlay_create.go
+++ b/cmd/internal/cli/overlay_create.go
@@ -1,0 +1,64 @@
+package cli
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/spf13/cobra"
+	"github.com/sylabs/singularity/docs"
+	"github.com/sylabs/singularity/internal/app/singularity"
+	"github.com/sylabs/singularity/pkg/cmdline"
+	"github.com/sylabs/singularity/pkg/sylog"
+)
+
+var (
+	overlaySize int
+	overlayDirs []string
+)
+
+// -s|--size
+var overlaySizeFlag = cmdline.Flag{
+	ID:           "overlaySizeFlag",
+	Value:        &overlaySize,
+	DefaultValue: 64,
+	Name:         "size",
+	ShortHand:    "s",
+	Usage:        "size of the EXT3 writable overlay in MiB",
+}
+
+// --create-dir
+var overlayCreateDirFlag = cmdline.Flag{
+	ID:           "overlayCreateDirFlag",
+	Value:        &overlayDirs,
+	DefaultValue: []string{},
+	Name:         "create-dir",
+	Usage:        "directory to create as part of the overlay layout",
+}
+
+func setPath(cmd *cobra.Command, args []string) {
+	userPath := os.Getenv("PATH")
+	path := []string{defaultPath}
+	if userPath != "" {
+		path = append(path, userPath)
+	}
+	os.Setenv("PATH", strings.Join(path, string(filepath.ListSeparator)))
+}
+
+// OverlayCreateCmd is the 'overlay create' command that allows to create writable overlay.
+var OverlayCreateCmd = &cobra.Command{
+	Args:   cobra.ExactArgs(1),
+	PreRun: setPath,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if err := singularity.OverlayCreate(overlaySize, args[0], overlayDirs...); err != nil {
+			sylog.Fatalf(err.Error())
+		}
+		return nil
+	},
+	DisableFlagsInUseLine: true,
+
+	Use:     docs.OverlayCreateUse,
+	Short:   docs.OverlayCreateShort,
+	Long:    docs.OverlayCreateLong,
+	Example: docs.OverlayCreateExample,
+}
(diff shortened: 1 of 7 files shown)


dtrudg:
Hi @cclerget - I'll probably get around to a proper review on Monday.

A quick thought though...

> SIF restrictions:
> 
>     * SIF image must not contain a writable overlay
> 
>     * SIF image must not be signed

Should there be some restriction or warning for SIFs where the rootfs is encrypted? Our overlay will not be, and that may not be intuitive / expected to the user?

cclerget:
Ah right forgot encryption case, thanks !

cmd/internal/cli/overlay_create.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+// -s|--size
+var overlaySizeFlag = cmdline.Flag{
+	ID:           "overlaySizeFlag",
+	Value:        &overlaySize,
+	DefaultValue: 64,
+	Name:         "size",
+	ShortHand:    "s",
+	Usage:        "size of the EXT3 writable overlay in MB",
```
dtrudg: Presumably this is MiB? If so can we put MiB here as we have had some confusion previously with powers of 10 MB vs powers of 2 MiB in some code.

docs/content.go
```diff
@@ -1039,4 +1039,26 @@ Enterprise Performance Computing (EPC)`
 
   To display the resulting configuration instead of writing it to file:
   $ singularity config global --dry-run --set "bind path" /etc/resolv.conf`
+
+	OverlayUse   string = `overlay`
+	OverlayShort string = `Manage EXT3 writable overlay image`
```
dtrudg: Either 'Manage **an** EXT3 image....' or ' Manage..... image**s**'

docs/content.go
```diff
@@ -1039,4 +1039,26 @@ Enterprise Performance Computing (EPC)`
...
+	OverlayLong  string = `
+  The overlay command allows to manage EXT3 writable overlay image.`
```
dtrudg: 'The overlay command allows management of EXT3 writable overlay images' ?

docs/content.go
```diff
@@ -1039,4 +1039,26 @@ Enterprise Performance Computing (EPC)`
...
+  $ singularity overlay create --help`
+
+	OverlayCreateUse   string = `create <options> image`
+	OverlayCreateShort string = `Create EXT3 writable overlay image`
+	OverlayCreateLong  string = `
+  The overlay create command allows to create EXT3 writable overlay image either
+  as a single EXT3 image or by adding it automatically to an existing SIF image.`
+	OverlayCreateExample string = `
+  To create and add a writable overlay to an existing SIF image:
+  $ singularity overlay --size 1024 /tmp/image.sif
```
dtrudg: Needs to be `overlay create` I think

docs/content.go
```diff
@@ -1039,4 +1039,26 @@ Enterprise Performance Computing (EPC)`
...
+
+  To create a single EXT3 writable overlay image:
+  $ singularity overlay --size 1024 /tmp/my_overlay.img`
```
dtrudg: Needs to be `overlay create` I think

internal/app/singularity/overlay_create.go
```diff
@@ -0,0 +1,205 @@
...
+			sifArch = arch
+
+			img.File.Close()
+
+			for _, overlay := range overlays {
+				if overlay.Type != image.EXT3 {
+					continue
+				}
+				delCmd := fmt.Sprintf("singularity sif del %d %s", overlay.ID, imgPath)
+				return fmt.Errorf("an overlay writable partition already exist in %s (ID: %d), delete it first with %q", imgPath, overlay.ID, delCmd)
```
dtrudg:
```suggestion
				return fmt.Errorf("a writable overlay partition already exists in %s (ID: %d), delete it first with %q", imgPath, overlay.ID, delCmd)
```

dtrudg (changes requested):
Looks mostly good to me. A few minor docs / messages things commented here.

Couple of questions....

- Why not e.g. `truncate --size 1024M myfile` instead of `dd` as that will be faster and less likely to exhaust disk as it creates a sparse file?
- `mkfs.ext3` is not in the standard user PATH on at least debian 10+, as that doesn't include`/sbin`. Worth looking in `/sbin` by default?

cclerget:
Thanks for the review @dtrudg , I applied your suggestions.

>     * Why not e.g. `truncate --size 1024M myfile` instead of `dd` as that will be faster and less likely to exhaust disk as it creates a sparse file?

IIRC sparse files have been source of bugs in v2, and it doesn't apply when adding overlay in a SIF image.
Also this is the way it is done in the documentation.

> 
>     * `mkfs.ext3` is not in the standard user PATH on at least debian 10+, as that doesn't include`/sbin`. Worth looking in `/sbin` by default?

Added `setPath` which is blinldy prepending the singularity default paths to the current PATH

dtrudg (approved):
LGTM

ikaneshiro (approved):

PR #5917, opened 2021-04-09
Disable -race for tests on ppc64le only

### LINKED ISSUE #5914: Cannot use -race when running tests on ppc64le
### Version of Singularity:

```
master
```

### Actual behavior

On ppc64le the e2e tests will fail if run with -race under Go 1.14 or 1.15.

They worked correctly with Go 1.13, but newer versions cause the CGO init() method to be called twice, the latter time unprivileged... so it fails. Not clear why this happens. Only on ppc64le, not amd64 or arm64.

Needs a reproducer worked up to raise it as an upstream issue in Go.

dtrudg:
Closes #5914

On ppc64le the e2e tests will fail if run with -race under Go 1.14 or
1.15.

They worked correctly with Go 1.13, but newer versions cause the CGO
init() method to be called twice if -race is used, the latter time unprivileged... so
it fails. Not clear why this happens. Only on ppc64le, not amd64 or
arm64.


cclerget (approved):
LGTM

PR #5953, opened 2021-04-22
fix: wait for cryptsetup /dev/mapper device to appear

### LINKED ISSUE #5950: Race condition exists between cryptsetup and mounting of /dev/mapper/<GUID>
### Version of Singularity:

```
$ singularity version
singularity version 3.7.3-1.el7.SN.1
```

### Expected behavior

```
[tyson@gra-ccsudo1:/tmp]$ singularity run --passphrase master.sif 
Enter encryption passphrase: 
bash-4.4$ 
```

### Actual behavior

```
[tyson@gra-ccsudo1:/tmp]$ singularity -d shell --passphrase master.sif
...
DEBUG   [U=3001792,P=10345]create()                      Mount all
DEBUG   [U=3001792,P=10345]mountGeneric()                Mounting tmpfs to /opt/software/singularity-3.7.3/var/singularity/mnt/session
DEBUG   [U=3001792,P=10345]mountImage()                  Mounting loop device /dev/loop7 to /opt/software/singularity-3.7.3/var/singularity/mnt/session/rootfs of type encryptfs
DEBUG   [U=3001792,P=10372]func1()                       Using cryptsetup at "/sbin/cryptsetup"
DEBUG   [U=3001792,P=10372]Open()                        Running /sbin/cryptsetup /sbin/cryptsetup open --batch-mode --type luks2 --key-file - /dev/loop7 5c0135f8-159d-4f4c-b1e9-ace83a5b1495
DEBUG   [U=3001792,P=10372]Open()                        Successfully opened encrypted device /dev/loop7
DEBUG   [U=3001792,P=10345]umount()                      Umount /opt/software/singularity-3.7.3/var/singularity/mnt/session/final
DEBUG   [U=3001792,P=10345]umount()                      Umount /opt/software/singularity-3.7.3/var/singularity/mnt/session/rootfs
ERROR   [U=3001792,P=10345]CleanupContainer()            could not cleanup crypt: invalid argument
FATAL   [U=3001792,P=10345]Master()                      container creation failed: mount /proc/self/fd/3->/opt/software/singularity-3.7.3/var/singularity/mnt/session/rootfs error: while mounting image /proc/self/fd/3: failed to mount squashfs filesystem: no such file or directory
[tyson@gra-ccsudo1:/tmp]$ 
```

It's a race condition so it may be tricky to trigger. It happens very reliably on our cluster login nodes though. Possibly because they usually sit at a load average >> 1. Or maybe because of other items. I don't know. What I do know is if you wrapper the `cryptsetup` executable with one that sleeps for 5 seconds before returning that resolves the issue.

```
$ cat /etc/os-release
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
ANSI_COLOR="0;31"

```

Custom RPM build (i.e., from source).

twhitehead:

The device `/dev/mapper/...` does not show up before `cryptsetup` returns. This creates a race condition as, if singularity attempts to mount `/dev/mapper/...`, before it shows up, it fails.

This pull request causes singularity to wait up to 25.5s for `/dev/mapper/...` to show up, through a series of exponential back off sleeps, before continuing. If it doesn't show up in that time it will abort with an error.

 - Fixes #5950

- [X] Read the [Guidelines for Contributing](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- [X] Added changes to the [CHANGELOG](https://github.com/sylabs/singularity/blob/master/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- [X] Added tests to validate this PR, linted with `make check`  and tested this PR locally with a `make test`, and `make testall` if possible (see CONTRIBUTING.md).
- [X] Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- [X] Added myself as a contributor to the [Contributors File](https://github.com/sylabs/singularity/blob/master/CONTRIBUTORS.md)

### DIFF
--- a/pkg/util/crypt/crypt_dev.go
+++ b/pkg/util/crypt/crypt_dev.go
@@ -306,6 +306,9 @@ func (crypt *Device) Open(key []byte, path string) (string, error) {
 			if err == nil {
 				break
 			}
+			if !errors.Is(err, os.ErrNotExist) {
+				return "", err
+			}
 			delay := 100 * (1 << attempt) * time.Millisecond
 			if delay > 12800*time.Millisecond {
 				return "", fmt.Errorf("device /dev/mapper/%s did not show up within %d seconds", nextCrypt, (delay-1)/time.Second)
--- a/pkg/util/crypt/crypt_dev.go
+++ b/pkg/util/crypt/crypt_dev.go
@@ -309,11 +309,12 @@ func (crypt *Device) Open(key []byte, path string) (string, error) {
 			if !errors.Is(err, os.ErrNotExist) {
 				return "", err
 			}
-			delay := 100 * (1 << attempt) * time.Millisecond
-			if delay > 12800*time.Millisecond {
-				return "", fmt.Errorf("device /dev/mapper/%s did not show up within %d seconds", nextCrypt, (delay-1)/time.Second)
+			delayNext := 100 * (1 << attempt) * time.Millisecond // power of two exponential back off means
+			delaySoFar := delayNext - 1                          // total delay so far is next delay - 1
+			if delaySoFar >= 25500*time.Millisecond {
+				return "", fmt.Errorf("device /dev/mapper/%s did not show up within %d seconds", nextCrypt, delaySoFar/time.Second)
 			}
-			time.Sleep(delay)
+			time.Sleep(delayNext)
 		}
 
 		sylog.Debugf("Successfully opened encrypted device %s", path)
--- a/pkg/util/crypt/crypt_dev.go
+++ b/pkg/util/crypt/crypt_dev.go
@@ -14,6 +14,7 @@ import (
 	"os/exec"
 	"strings"
 	"syscall"
+	"time"
 
 	uuid "github.com/satori/go.uuid"
 	"github.com/sylabs/singularity/internal/pkg/util/bin"
@@ -299,6 +300,19 @@ func (crypt *Device) Open(key []byte, path string) (string, error) {
 
 			return "", fmt.Errorf("cryptsetup open failed: %s: %v", string(out), err)
 		}
+
+		for attempt := 0; true; attempt++ {
+			_, err := os.Stat("/dev/mapper/" + nextCrypt)
+			if err == nil {
+				break
+			}
+			delay := 100 * (1 << attempt) * time.Millisecond
+			if delay > 12800*time.Millisecond {
+				return "", fmt.Errorf("device /dev/mapper/%s did not show up within %d seconds", nextCrypt, (delay-1)/time.Second)
+			}
+			time.Sleep(delay)
+		}
+
 		sylog.Debugf("Successfully opened encrypted device %s", path)
 		return nextCrypt, nil
 	}
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -85,6 +85,7 @@
     - Thomas Hamel <hmlth@t-hamel.fr>
     - Tim Wright <7im.Wright@protonmail.com>
     - Tru Huynh <tru@pasteur.fr>
+    - Tyson Whitehead <twhitehead@gmail.com>
     - Vanessa Sochat <vsochat@stanford.edu>
     - Westley Kurtzer <westley@sylabs.io>, <westleyk@nym.hush.com>
     - Yannick Cote <y@sylabs.io>, <yhcote@gmail.com>


twhitehead:
If you are happy with this, I can cherry pick it against the the release-3.7 branch too.

gmkurtzer:
This looks good to me, but I'll let someone more qualified do the review. With that said, I would think that 25 seconds is probably excessive. How long are you seeing that it typically takes for `/dev/mapper` devices to show up?

twhitehead:
I don't actually know other than in my initial debugging a 5s delay to cryptsetup was sufficient to resolve it, so probably doesn't have to be long at all. I had just figured it never showing up after `cyrptsetup` succeeds is such an unlikely error path though, that that a longer pause to be 100% certain was safe. I'm happy to force push a different amount if you have a specific number in mind.

pkg/util/crypt/crypt_dev.go
```diff
@@ -299,6 +300,19 @@ func (crypt *Device) Open(key []byte, path string) (string, error) {
 
 			return "", fmt.Errorf("cryptsetup open failed: %s: %v", string(out), err)
 		}
+
+		for attempt := 0; true; attempt++ {
+			_, err := os.Stat("/dev/mapper/" + nextCrypt)
+			if err == nil {
```
dtrudg, line 306: In checks for the existence of a file/directory/device, it's usual to check if `os.IsNotExist(err)` is true, and fail immediately if the err is not an IsNotExist.

pkg/util/crypt/crypt_dev.go
```diff
@@ -299,6 +300,19 @@ func (crypt *Device) Open(key []byte, path string) (string, error) {
...
+				break
+			}
+			delay := 100 * (1 << attempt) * time.Millisecond
+			if delay > 12800*time.Millisecond {
```
dtrudg: If you can make the conditionals more explicit that the max wait here is 25.5s then that would be useful for future readers skimming over the code. Some concise comments about the exponential backoff progression / max would work also.

dtrudg (changes requested):
Thanks for contributing this. I have a couple of fairly minor requests.

dtrudg:
Hi @twhitehead - given we are about 2 weeks out from an RC for 3.8 it's not clear there will be another 3.7 release, so you don't need to cherry-pick to the release-3.7 branch. If we issue one then this PR is nice and contained that we can cherry-pick easily.

pkg/util/crypt/crypt_dev.go
twhitehead: Good idea. I've add another variable and comment to make the total delay so far calculation clear.

pkg/util/crypt/crypt_dev.go
twhitehead: Done.

twhitehead:
I just force pushed a new version. I'm not sure why it hasn't shown up here yet.

BTW, I'm always a bit confused about what it the best practice for integrating feedback into a pull request

- force pushing corrections based on feedback the best approach, or
- push additional correction commits

and in the last case, should you squash them squash them in the end when everyone agrees it is good?

dtrudg:
@twhitehead - for Singularity the core developers tend to force push updated commits with corrections, but it doesn't always happen like that, and there is no hard rule here. If there are relatively small PRs like this one that have fixup commits we can do a 'squash and merge' when we merge the PR to clean the history up a bit when it's appropriate.

twhitehead:
Thanks. As it still didn't seem to be picking up on my force push, I decided to change it over to a fixups commit for each of the requests as suggested [here](https://github.com/marketplace/actions/block-fixup-commit-merge#why) and force pushed that, which it seems to have picked up on.

From [this community discussion](https://github.community/t/merge-option-with-rebase-autosquash/2242/4) it simply sounds like there isn't currently a good way to do this in github (or at least for something more complex than this one which can just be squashed as you pointed out).

dtrudg (approved):

dtrudg:
Thanks @twhitehead!

twhitehead:
NP. Thank you for all the hard work on this project and the quick handling and feedback on the pull request. :+1:

PR #5977, opened 2021-05-10
Rename package github.com/sylabs/singularity to github.com/hpcng/singularity

cclerget:


INSTALL.md
```diff
@@ -83,11 +83,11 @@ to build Singularity from source:
 ```
 $ mkdir -p ${GOPATH}/src/github.com/sylabs && \
```
DrDaveD: This should be  hpcng, not sylabs.

go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/sylabs/singularity
+module github.com/hpcng/singularity
 
 go 1.13
```
DrDaveD, line 3: How about the copy of golang.org/x/crypto in github.com/sylabs/golang-x-crypto, should we move that too?

DrDaveD (changes requested):
A little more is needed in INSTALL.md, and perhaps the golang-x-crypto should be coming from hpcng.

INSTALL.md
cclerget: fixed

go.mod
cclerget: Yes, change applied

DrDaveD (approved):
Looks good now

ikaneshiro (approved):

kmuriki (approved):
Looks good to me. Thanks

PR #5979, opened 2021-05-10
Use github actions in replacement of CircleCI

cclerget:
Motivations in favor of github actions:

- Integrated CI for contributors
- Github actions are not subject to Docker Hub rate limit which is convenient for e2e tests
- More performant than CircleCI free plan


ikaneshiro (approved):
Lgtm

.github/workflows/ci.yml
```diff
@@ -0,0 +1,204 @@
...
+      - name: Check code
+        run: make -C ./builddir check
+  
+  rpmbuild-centos7:
+    runs-on: ubuntu-20.04
+    name: rpmbuild-centos7
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set OS type and version
```
DrDaveD: This step does more than set OS type and version, it actually runs the build.  That was confusing when I went to look at the output.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,204 @@
...
+          OS_VERSION: 7
+        run: ./scripts/ci-docker-run
+
+  rpmbuild-centos8:
+    runs-on: ubuntu-20.04
+    name: rpmbuild-centos8
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set OS type and version
```
DrDaveD: Same here

DrDaveD (changes requested):
Please change the name of the rpmbuild build step, otherwise looks good.

.github/workflows/ci.yml
cclerget: fixed

DrDaveD (approved):

PR #5989, opened 2021-05-20
Added Probot-Stale to clean up stale issues automatically

carterpeel:
With the current configuration, the bot will wait for 60 days of inactivity before sending a message to the issue discussion. If no response is received within 7 days, it will automatically close the issue.


cclerget (commented):
Just a usability question, does it re-open issues on comment or when the stale label is removed ?

carterpeel:
> Just a usability question, does it re-open issues on comment or when the stale label is removed ?

From what I recall, it will remove the stale label within the grace period if there is a response or comment.

cclerget (approved):
LGTM

PR #6161, opened 2021-09-01
Merge release-3.8 into master after tagging v3.8.2

kmuriki:
## Description of the Pull Request (PR):

This merges release-3.8 branch into master branch.

PR #6164, opened 2021-09-01
Try merging release-3.8 again

DrDaveD:
This is another attempt at merging the commits from release-3.8 into the master branch, the same as #6161 but after #6162 which pulled in the only commit that actually made a change.  For some reason I don't understand this PR thinks it needs to merge the same one-line change again, but when I tested this on my own repo it passed the CI tests so I guess it's OK.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,7 +83,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.16.4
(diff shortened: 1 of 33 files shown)


kmuriki:
@DrDaveD  Well this worked!

DrDaveD:
I think we might as well merge this to hopefully avoid future issues with merging the release branch back into master.  @ikaneshiro do you agree?

ikaneshiro (approved):
This looks OK to me

Issue #108, opened 2016-07-04
Get singularity running on mac/in docker?


cjprybol:
My school recently installed singularity on a cluster I have access to, and I would like to try it, but I'm unable to create any images because I don't have sudo privileges.

To get around that, I'm trying to get an image going on my personal computer (mac) on which I do have sudo privileges, because it seems as though once an image is ready to be run, it can be run anywhere that has singularity installed (so I can build an image on my laptop and transfer it to the cluster to run).

When trying to install singularity on my mac, I run

**mac code**

``` bash
git clone https://github.com/gmkurtzer/singularity.git
cd singularity
./autogen.sh
./configure --prefix=/usr/local
```

**truncated output**

```
ERROR!!!!!!

This host does not support the CLONE_NEWNS (mount) namespace flag! You
really really really don't want to run Singularity containers without a
Separate mount name namespace!
```

From that, I take it installing on mac isn't a good idea, so my next idea was to try and run it in a sandboxed linux VM on my laptop. I tried docker first. I'm using the docker for mac beta [here](https://docs.docker.com/docker-for-mac/). This is the minimum reproducible example I could come up with.

**install starting in a mac bash session with docker installed**

``` bash
# pull image from dockerhub
docker pull debian
# run and enter into a debian shell
docker run -it debian /bin/bash
# update package database
apt-get update
# get packages needed to install singularity
apt-get -y install build-essential curl git sudo man vim autoconf libtool
# add user so we can get out of root
useradd -m user
# set password for user
echo 'user:password'|chpasswd
# give sudo privileges to user
usermod -aG sudo user
# change to user
su user
# go to home directory of user
cd ~
# pull master repo
git clone https://github.com/gmkurtzer/singularity.git
# enter, configure, install
cd singularity
./autogen.sh
./configure --prefix=/usr/local
make
echo "password" | sudo --stdin make install
bash test.sh
```

**output from test.sh**

```
Building test container...

 + sudo singularity create -s 15 container.img                  (retval=1) ERROR
Creating a sparse image with a maximum size of 15MiB...
INFO   : Using given image size of 15
ERROR: Failed binding image: container.img
Full output in: /tmp/tmp.01BQhVNvLh
```

If I ignore the message, and try and bootstrap the container (or any other container, they give the same results) using the centos.def or debain.def files, I get a segmentation fault.

So I'm asking both
1. What's the easiest way to get singularity up and running for someone without sudo access on a cluster?
[...]

Issue #91, opened 2016-06-24
bootstrap RunCmd and RunScript need clarifying


loveshack:
It seems as if RunCmd is meant only to take a single arg.
Also RunScript potentially will have problems with shell expansion/word splitting unless the args are appropriately quoted.
I think that needs documenting.

gmkurtzer:
Yeah, I was messing with this a bit last night. Notice my change to the
example/debian.def.

Because this is all happening and being passed through shell I don't see a
way to do it that doesn't loose some quoting or use <<< which alpine
Linux's implementation of Bash does not support.

If it must be a single argument, so be it and it will be documented
appropriately.... But I dislike that.

Any other ideas?

On Friday, June 24, 2016, Dave Love notifications@github.com wrote:

> It seems as if RunCmd is meant only to take a single arg.
> Also RunScript potentially will have problems with shell expansion/word
> splitting unless the args are appropriately quoted.
> I think that needs documenting.
> 
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> https://github.com/gmkurtzer/singularity/issues/91, or mute the thread
> https://github.com/notifications/unsubscribe/ANVkxYNbp5tFJOgxpFeuRwvS8PYlxngGks5qO9RDgaJpZM4I9ve7
> .

## 

Gregory M. Kurtzer
High Performance Computing Services (HPCS)
University of California
Lawrence Berkeley National Laboratory
One Cyclotron Road, Berkeley, CA 94720

loveshack:
I removed this when I saw the RunCmd change.  It's clearly more
difficult quoting-wise if you use sh -c; is that necessary?

```
--- a/libexec/mods/linux_build.smod
+++ b/libexec/mods/linux_build.smod
@@ -157,7 +157,11 @@ RunScript() {
         chmod +x "$SINGULARITY_BUILD_ROOT/singularity"
     fi

-    echo "$@" >> "$SINGULARITY_BUILD_ROOT/singularity"
+    for arg in "$@"; do
+        local qq="'\''"
+        printf "'%s' " "${arg//\'/$qq}" >> "$SINGULARITY_BUILD_ROOT/singularity"
+    done
+    echo >> "$SINGULARITY_BUILD_ROOT/singularity"

     return 0
 }
```

I don't know why <<< is necessary, but it's been in bash since the last
v2 release which can't be in any current distribution.  Even Solaris 10
has v3.

[...]

Issue #88, opened 2016-06-23
mpirun example fails


loveshack:
As reported on the list, the mpirun example from http://singularity.lbl.gov/#hpc fails for me.  It's en el6 host with a recent trunk singularity and any of OMPI 1.6 (our native one), 1.8, or 1.10.  I used a cut-and-paste from the example (fixed for non-ascii).  The ranks all fail with:

```
    ompi_mpi_init: ompi_rte_init failed
    --> Returned "(null)" (-43) instead of "Success" (0)
```

rhc54:
that is correct - you need the OMPI master, or release 2.1 once that occurs (probably this fall?)

gmkurtzer:
Thanks for the clarification Ralph!

loveshack:
This needs to be clarified in the documentation, especially as that
wasn't what I understood before, though I didn't see how it could work.

It could use explanation of what the integration does too.

gmkurtzer:
I agree and will update/clarify the documentation. Reopening issue as a reminder to do that.

loveshack:
Maybe note that it will work with compatible MPIs (including older OMPI) inside and outside the container, at least with no pid namespacing. 

Is there some way to use the container's mpirun to start the container MPI processes?  I've forgotten what went wrong when I tried.

gmkurtzer:
It is possible to use the container's mpirun to start the contained MPI process... BUT..... Then each container has to have ssh and sshd with sshd's running within the container (or having the scheduler automatically spawn the orted processes within the container). Starting the MPI process from within the container increases complexities significantly without gaining an equal amount... As a matter of fact, I don't think it gains anything, and may even loose more then it gains. -- This will make a good FAQ as well.

I added some verbiage on the HPC page at the bottom noting the OMPI version used.

gmkurtzer:
BTW, I added an FAQ about the mpirun inside/outside container. Thanks!

PR #6185, opened 2021-09-28
Fix make check warnings, and do lint check only once in CI, from sylabs 91

DrDaveD:
This PR does does equivalent things from
- #sylabs/singularity#91
which eliminated a warning described in
- #sylabs/singularity#88
by getting rid of the check/run-linter script.  I have seen that warning so I think it would be good to get rid of it.  It does require anybody who is running it interactively to install golangci-lint themselves.

The PR also refactored CI to only do the make check once in a separate CI test, and I did that here as well in a new test called check_source.

Also included is
- #sylabs/singularity#108
which simply bumped the golangci-lint version from 1.40 to 1.41


ikaneshiro (approved):

gmkurtzer (approved):
This looks good to me.

PR #5805, opened 2021-01-29
add `SINGULARITY_MKSQUASHFS_PROCS` and `SINGULARITY_MKSQUASHFS_MEM` env vars

kaczmarj:
## Description of the Pull Request (PR):

These environment variables can be used to set the -processors and -mem options to `mksquashfs`. These environment variables take precedence over the configuration file.

### This fixes or addresses the following GitHub issues:

 - related to the conversation in https://github.com/hpcng/singularity/pull/5003#issuecomment-764937035

#### Before submitting a PR, make sure you have done the following:

- Read the [Guidelines for Contributing](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/sylabs/singularity/blob/master/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- Added tests to validate this PR and tested this PR locally with a `make testall`
- Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/master/CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](https://github.com/sylabs/singularity/blob/master/CONTRIBUTORS.md)

Attn: @singularity-maintainers

PR #6186, opened 2021-09-28
Adding environment parameters to override mksquashfs procs/mem.

soichih:

This adds 2 new ENV parameters SINGULARITY_MKSQUASHFS_PROCS and SINGULARITY_MKSQUASHFS_MEM. These environment variables can be used to set the -processors and -mem options to mksquashfs. These environment variables take precedence over the configuration file.

This is necessary because when I submit a job on a batch scheduler with a specific amount of cpus/mem available for the job, singularity must only use up to that number of cpu/mem available or the job will be killed while caching the SIF image. This can not be done by singularity.conf alone (and it's not the right place for that - as cluster admin won't know how much a cpu/proc each job would actually request)

This is similar to this PR #5805  (Thanks @kaczmarj !!)

I didn't add env/override for `mksquashfs path` as I am not sure if it's safe to be overridden.

https://github.com/hpcng/singularity/pull/5003#issuecomment-764937035

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## v3.8.2 - [2021-08-31]
 
+### New features / functionalities
+
+   - Add SINGULARITY_MKSQUASHFS_PROCS and SINGULARITY_MKSQUASHFS_MEM to override
+     mksquashfs options.
 ### Bug fixes
 
   - Fix regression when files `source`d from `%environment` contain `\` escaped
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
    - Add SINGULARITY_MKSQUASHFS_PROCS and SINGULARITY_MKSQUASHFS_MEM to override
      mksquashfs options.
+     
 ### Bug fixes
 
   - Fix regression when files `source`d from `%environment` contain `\` escaped
--- a/internal/pkg/util/fs/squashfs/squashfs.go
+++ b/internal/pkg/util/fs/squashfs/squashfs.go
@@ -7,8 +7,10 @@ package squashfs
 
 import (
 	"fmt"
+	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/hpcng/singularity/internal/pkg/buildcfg"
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## v3.8.2 - [2021-08-31]
 
+### New features / functionalities
```
DrDaveD, line 5: This section should be above the 3.8.2 label, not below it.

DrDaveD (changes requested):

DrDaveD (changes requested):
I don't see that you addressed the [comment](https://github.com/hpcng/singularity/pull/5805#issuecomment-790205610) from @ikaneshiro asking for a corresponding cli option to `pull`, `build`, and the `actions` command group.

soichih:
@DrDaveD Sorry I didn't see that comment. I think that, singularity using mksquashfs to build container is an implementation detail so I don't think it should be exposed via CLI, in my opinion. If we do that, maybe it's better to call these options a bit more generic like "--max-build-mem/procs"? I guess I can say the same for the ENV options.. so maybe rename them to `SINGULARITY_BUILD_MAXMEM"? 

Anyway, I just looked at those cli options stuff and I couldn't quite figure how I can access those CLI parameters in squashfs.go. Do I need to pass those parameters down the callstack as a function arguments? (yikes!)

DrDaveD:
> @DrDaveD Sorry I didn't see that comment. I think that, singularity using mksquashfs to build container is an implementation detail so I don't think it should be exposed via CLI, in my opinion. If we do that, maybe it's better to call these options a bit more generic like "--max-build-mem/procs"? I guess I can say the same for the ENV options.. so maybe rename them to `SINGULARITY_BUILD_MAXMEM"?

That makes sense to me.

> Anyway, I just looked at those cli options stuff and I couldn't quite figure how I can access those CLI parameters in squashfs.go. Do I need to pass those parameters down the callstack as a function arguments? (yikes!)

Ian pointed to some example code.  It looks like it just becomes a global flag name in the cli module.

DrDaveD:
@soichih Does it look like the cli module global flag is going to do it?

DrDaveD:
@soichih Now this PR has conflicts.  Can you take another look?

internal/pkg/util/fs/squashfs/squashfs.go
```diff
@@ -60,6 +62,16 @@ func GetProcs() (uint, error) {
 	// proc is either "" or the string value in the conf file
 	proc := c.MksquashfsProcs
 
+	// let user override via ENV
+	procEnv := os.Getenv("SINGULARITY_MKSQUASHFS_PROCS")
```
PaulCharlton, line 66: @soichih as for naming ... perhaps consider that the naming would be cleaner and closer to a canonical form with the suffix "_PROCESSORS" instead of "_PROCS" ... the reason being is that the mksquashfs command uses "-processors N" as its command line option

DrDaveD:
@soichih If you want this please resubmit it in https://github.com/apptainer/apptainer.

PR #6188, opened 2021-09-28
Reapply sylabs pr 152: Inherit Docker/OCI LABELs during build

DrDaveD:
This re-applies the pr
- sylabs/singularity#152
which was a fix for
- sylabs/singularity#146

It had been applied previously in #6105 but then was un-applied in #6119 because it was the only thing in the master branch that we didn't want in release 3.8.

Below is the original pr description:

When building / pulling from a Docker/OCI source, retrieve LABELs from the config, and ensure they are set in the resulting Singularity container.

This fixes a longstanding regression from Singularity 2.x. Note that you will now need to use `--force` in a build to override a label that already exists in the source Docker/OCI container.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@
     the `%test` section of the build with a ephemeral tmpfs overlay,
     permitting tests that write to the container filesystem.
 
+### Changed defaults / behaviours
+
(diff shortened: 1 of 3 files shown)


DrDaveD:
Force-pushed to fix conflicts in CHANGELOG.md

gmkurtzer (approved):
This looks good to me.

Issue #183, opened 2016-08-04
Question about here documents


nl384:
Hi Greg,

I just have a couple questions about the nature of using here documents in Singularity. So I noticed that if I try to use RunCmd with 'cat' and a here document, it doesn't seem to actually create a file:

```
RunCmd cat <<'HERE' > /usr/local/setup.sh 
#!/bin/bash

conda env create -f /usr/local/environment.yml
echo "You have successfully set up your environment”
HERE
```

However, it worked when I put a here document with /bin/bash around my 'cat' here document.

```
RunCmd /bin/bash <<EOF 
cat <<'HERE' > /usr/local/setup.sh 
#!/bin/bash

conda env create -f /usr/local/environment.yml
echo "You have successfully set up your environment. You can now run your Python interpreter using the 'python-tensorflow-build' command."
HERE
EOF
```

Is this similar to the error that I got here (https://github.com/gmkurtzer/singularity/issues/107) where the output was probably being put onto my host computer? Also, it seems strange but I found that single quoting my limit string didn't seem to actually work to escape my variables. For instance, something like the following would create a file with "" instead of "$@":

```
RunCmd /bin/bash <<EOF
cat <<'HERE' >> /singularity
#!/bin/bash
source activate ~/.conda/envs/python-tensorflow
python "$@"
HERE
EOF

```

To create a file that actually had "$@" I have to manually escape with "\$@" despite single quoting 'HERE'. It this something related to singularity, and is there a way to fix it?

Thanks!
Nathan

gmkurtzer:
Hi @nl384,

In the first example, you actually did something kinda interesting, you ran the HERE doc through the container's /bin/cat and then you outputted it to the host file system. In your second example, that is a better way to do it, and while there is nothing technically wrong with multiple levels of HERE docs, they are hard to follow.

In the last example, as you see variables still need to be escaped because the definition file itself is being parsed by the shell. It is a major shortcoming in the current definition design that will hopefully soon be rectified. I have some ideas on how to do this, and they should be coming soon! But for now, there is no easy fix.

HTH!

nl384:
Thanks, Greg. What is the better way to do what I did in the second example? Thanks!

gmkurtzer:
Well, there is technically nothing wrong with it but I will keep the use-case of creating a script/file/text via this method in my head as I work out the next version of the definition file syntax (the old one will still be supported as well).

Greg

nl384:
Thanks, Greg. I'll close the issue.

[...]

PR #6191, opened 2021-09-29
Split pkg/util/crypt key and device functionality, from sylabs prs 182 & 184

DrDaveD:
This pr pulls in sylabs prs
- sylabs/singularity#182
- sylabs/singularity#184
which fixes issues
- sylabs/singularity#181
- sylabs/singularity#183

The original PR descriptions were:

`pkg/util/crypt` contained code that calls out to `cryptsetup`, which is found via routines that rely on a valid buildcfg (only present when SingularityCE is built via `mconfig`). This, in turn, prevents `pkg/build/types` being imported from other projects as `bundle.go` imports `pkg/util/crypt`.

Move the device handling portions of the crypt code to `internal/pkg/util/crypt`.

Keep the remaining key handling portions of the code, in `pkg/util/cryptkey`. These now have no internal or buildcfg dependencies.

Ensure no package in pkg/.. depends directly or indirectly on buildcfg, as that would prevent use from another project.


gmkurtzer (approved):
This looks good to me.

ikaneshiro (approved):

PR #188, opened 2016-08-06
WiP ENH: try testing under different shells (to demo #187)

yarikoptic:
most probably wouldn't work right away -- just want to see what the effect would be atm

Issue #187, opened 2016-08-06
[2.1.2] test.sh fails if my shell is not avail inside the container


yarikoptic:
```
 + singularity shell -w container.img -c true                                          (retval=0) ERROR
ERROR: Shell does not exist in container: /bin/zsh
ERROR: Using /bin/sh instead...
Full output in: /tmp/tmp.SCbtE7htxK
```

I guess it should check SHELL, which is in my case /bin/zsh obviously ... overriding SHELL env var is not sufficient -- I guess since inside it populates with the same user record, so to use /bin/zsh in my case, which doesn't exist inside the container... should get installed may be?  or reset it to /bin/bash or ... ?

gmkurtzer:
So if I'm understanding correctly... /bin/sh is calling for /bin/zsh? hrmm, I'll have to check on that and try to replicate.

yarikoptic:
I didn't check but thought that it relates to the fact that my user has zsh as the shell configured

gmkurtzer:
Is zsh configured as your default shell in /etc/passwd?

Also what version of Singularity is this?

yarikoptic:
On Wed, 10 Aug 2016, Gregory M. Kurtzer wrote:

>    Is zsh configured as your default shell in /etc/passwd?

yes

>    Also what version of Singularity is this?

2.1.2 (as in the subject line/title ;) )

gmkurtzer:
I don't believe this is a problem anymore with the massive updates to master:

```
[gmk@centos7-x64]~/git/singularity-devel% grep gmk /etc/passwd                               
gmk:x:1000:1000:Gregory M. Kurtzer:/home/gmk:/bin/zsh
[gmk@centos7-x64]~/git/singularity-devel% singularity exec /tmp/Centos-7 grep gmk /etc/passwd
gmk:x:1000:1000:Gregory M. Kurtzer:/home/gmk:/bin/zsh
[gmk@centos7-x64]~/git/singularity-devel% singularity shell /tmp/Centos-7                    
Singularity: Invoking an interactive shell within container...

Singularity.Centos-7> echo $SHELL
/bin/sh
Singularity.Centos-7> exit
[gmk@centos7-x64]~/git/singularity-devel% 
```

Can you retest?

Thanks!

gmkurtzer:
Closing, let me know if the issue is still a problem for you and I will reopen.

Thanks!

PR #6200, opened 2021-10-05
Replace cp (except -L) calls with docker CopyWithTar, from sylabs 188

DrDaveD:
This pulls in sylabs PR
- sylabs/singularity#188
which fixed issue
- sylabs/singularity#187

Original pr description:

In order to reduce our calls to external binaries, and open up the
ability for further work to handle effective-ownership preserving
uid mapping during copies, use the CopyWithTar function
from docker's pkg/archive. This copies symlinks as-is, so cannot
be used at present for Singularity's host->container copies, which
dereference all symlinks with cp -L (we should revisit this for 4.0
and probably drop the full dereferencing in line with Docker).

We have to wrap CopyWithTar for unprivileged use, as it will try
to create the top-level of the copy destination as the 'container root'
user. Single entry UID/GID maps are used to set this to the current
user. The low level Untar function is overridden so that it always
extracts as the effective UID/GID, also.

This wrapping maintains the current implicit ownership squashing in
unprivileged Singularity sandbox builds. Given sandboxes can be created
/ used with fakeroot / userns preserving effective ownership, we
may want to revisit for 4.0 whether squashing silently is always
appropriate, and think about how we can use the UID/GID mapping
functionality of CopyWithTar productively


DrDaveD:
The first attempt at e2e failed with errors that were unfamiliar to me, so I started to investigate but after getting nowhere I just re-ran the job and then it worked.  This is ready to go, @ikaneshiro and @cclerget please review.

ikaneshiro (approved):

Issue #193, opened 2016-08-19
Image-based bind mounts


bbockelm:
I would like to use one image for a the operating system environment (say, CentOS) and another image for a bind mount (say, a HEP-specific runtime).

Why not simply combine the two into a single image?  The HEP runtime can be quite large (a test squashfs for CMS is 77GB): it's really expensive in CPU and disk quota to make a merged version.  If I have two slightly different OS versions, I don't want to keep two copies of the runtime.

gmkurtzer:
Is it possible to throw it into a local directory and make it a bind mount?

bbockelm:
Yeah - that's the other issue: there's a huge number of files, causing inode exhaustion (or at least stress on the metadata server!).  Unfortunately, unprivileged users also cannot mount it into the local directory.

gmkurtzer:
I am thinking the implementation would be a list of image files and a bind point location inside the container (either envar or configuration file). Would that meet your desire?

Is this an urgent need? If so I can push to get it into the 2.2 release.

gmkurtzer:
This wasn't brought into 2.2, nor will it be part of 2.3. I am hoping we can get it in 2.4. LOL

GodloveD:
@bbockelm and @gmkurtzer does the `--overlay` option in development do what you need?  Should we close this issue?

GodloveD:
@bbockelm This should be implemented now correct? 

http://singularity.lbl.gov/docs-overlay

I'm going to close this, but feel free to re-open if you were actually asking for something different.

Issue #190, opened 2016-08-10
Make rpmbuild create rpm with Version matching release tag


tcooper:
## Summary

Related to #131 and ultimately the choice of tag format please consider modifying inputs to the RPM build process so that the Version and/or Release of the generated RPM increment appropriately.

Currently all 2.x tagged releases produce an RPM with same Version and Release which prevents a simple update.
### singularity-2.1

```
[tcooper@dev singularity-2.1]$ rpm -q --queryformat %{NAME}-%{VERSION}-%{RELEASE}.%{ARCH} -p /home/tcooper/rpmbuild/RPMS/x86_64/singularity-2.1-0.1.el6.x86_64.rpm
singularity-2.1-0.1.el6.x86_64
```
### singularity-2.1.1

```
[tcooper@dev singularity-2.1.1]$ rpm -q --queryformat %{NAME}-%{VERSION}-%{RELEASE}.%{ARCH} -p /home/tcooper/rpmbuild/RPMS/x86_64/singularity-2.1-0.1.el6.x86_64.rpm
singularity-2.1-0.1.el6.x86_64
```
### singularity-2.1.2

```
 [tcooper@dev singularity-2.1.2]$ rpm -q --queryformat %{NAME}-%{VERSION}-%{RELEASE}.%{ARCH} -p /home/tcooper/rpmbuild/RPMS/x86_64/singularity-2.1-0.1.el6.x86_64.rpm
singularity-2.1-0.1.el6.x86_64
```

A minimal change (assuming the git tag _is_ the Version you want to use) that produces an RPM with an updated Version is...

```
[tcooper@dev singularity-2.1.2]$ diff -Naur configure.ac.orig configure.ac
--- configure.ac.orig   2016-08-10 11:40:18.031028845 -0700
+++ configure.ac    2016-08-10 11:30:27.362370439 -0700
@@ -1,5 +1,5 @@
AC_PREREQ(2.59)
-AC_INIT([singularity],[2.1],[gmkurtzer@lbl.gov])
+AC_INIT([singularity],[2.1.2],[gmkurtzer@lbl.gov])


 if test -z "$prefix" -o "$prefix" = "NONE" ; then
```

Followed by...

```
[tcooper@dev singularity-2.1.2]$ autoconf configure.ac > configure
```

This will update the `singularity.spec` file appropriately...

```
$ egrep "Version:|Release:" singularity.spec
Version: 2.1.2
Release: %{_rel}%{?dist}
```

[...]

PR #6206, opened 2021-10-06
Use Go filepath.Glob instead of calling /bin/sh, from sylabs 193

DrDaveD:
This pulls in sylabs pr
- sylabs/singularity#193
which fixed issue
- sylabs/singularity#190

As a legacy of the time Singularity was implemented in C & shell,
copying in the `%files` section uses a call out to shell to handle
globbing of the copy source. This results in odd behaviour if environment
vars are substituted in the source, but won't be for the dest etc.

It seems like a good time to clean up this behaviour. Switch to using the
Go `filepath.Glob` function to perform the globbing. Only do globbing,
not any other shell substitutions. This will bring us to the same set of
globbing patterns as Dockerfile COPY supports.

Note - we still don't properly handle globbing through absolute symlinks
in a stage->stage copy here. For that we need to emulate globbing in a
chroot, by writing our own globbing function


ikaneshiro (commented):
It is worth noting that these changes will lead to globbing patterns including files and directories with a `.` prefix that  shell globbing normally omits

CHANGELOG.md
```diff
@@ -14,6 +14,12 @@
   regression from Singularity 2.x. Note that you will now need to use
   `--force` in a build to override a label that already exists in the source
   Docker/OCI container.
+- The source paths for `%files` lines in a definition file are no longer
```
PaulCharlton, line 17: @DrDaveD would this be considered a "breaking change" by anybody?

DrDaveD: It's possible but I doubt it is likely to affect anyone because people shouldn't be expecting to be able to use environment variables in %files, I don't think.  I'm more concerned about the change in globbing behavior that Ian mentioned, and he is looking into an alternative.

DrDaveD:
Replacing this with issue apptainer/apptainer#19.

PR #6217, opened 2021-10-11
Use gofumpt for stricter formatting, from sylabs 242

DrDaveD:
This pulls in sylabs pr
- sylabs/singularity#242

I ran `go get mvdan.cc/gofumpt@latest` and `gofumpt -w .` to make the big commit touching many files, I didn't cherry-pick that commit.

The original PR description was:

Use `gofumpt` for stricter formatting. Mainly clarifies octal literals and does some `var x =` to `x :=` simplification in our codebase.

PR #6219, opened 2021-10-12
Allow `remote add --insecure` for http endpoints, from sylabs 245

DrDaveD:
This pulls in sylabs pr
- sylabs/singularity#245
which fixes
- sylabs/singularity#237

The original pr description was:

Previously, `https://` was always prefixed to a remote endpoint's
configured `URI`. This meant that remote service discovery was not
possible against an http-only server.

Patches have been circulated by the community to allow use of http for
development etc.

This PR allows the `--insecure` flag to be specified to `remote add`.
An insecure endpoint configuration will perform service discovery over
http.

Additionally, the filename for caching remote configs is changed to a
safe 'slug' version of the remote URI. This allows URIs with a '/' or
other special characters in them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
   network namespaces as these may not be supported on many installations.
 - `--no-https` now applies to connections made to library services specified
   in `--library://<hostname>/...` URIs.
+- `remote add --insecure` may be used to configure endpoints that are only
+  accessible via http.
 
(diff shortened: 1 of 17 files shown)


DrDaveD:
The source check is failing I think because it is assumes #6217 is merged.

DrDaveD:
@cclerget told me that we won't be needing this.

PR #6221, opened 2021-10-12
Bump mvdan.cc/sh/v3 and fix script sourcing from PATH plus minor fixes

cclerget:
### Description

It was fixed upstream by https://github.com/mvdan/sh/pull/742

### This fixes or addresses the following GitHub issues:

 - Fixes #6220 

#### Before submitting a PR, make sure you have done the following:

- Read the [Guidelines for Contributing](https://github.com/hpcng/singularity/blob/master/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/hpcng/singularity/blob/master/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/hpcng/singularity/blob/master/CONTRIBUTING.md)
- Added tests to validate this PR, linted with `make check`  and tested this PR locally with a `make test`, and `make testall` if possible (see CONTRIBUTING.md).
- Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/hpcng/singularity/blob/master/CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](https://github.com/hpcng/singularity/blob/master/CONTRIBUTORS.md)

Issue #366, opened 2016-11-29
[trivial bug] add_http should change http to https


vsoch:
Currently, if the url provided is `http`, and the default is to return `https`, the regular expression searches for http and then doesn't change the url. We should check for the specific prefix, and fix if necessary.

flx42:
Err, why? Are you saying `http://example.com` should always be changed to `https://example.com`? 'https' should only be added if the scheme was not specified.

flx42:
Note however that the current approach can't handle the convoluted case of domains actually starting with `http`. For instance try `add_http("httpie.org")`.
Really, we should probably be using something like `urlparse.urlparse("httpie.org", scheme="https")`

vsoch:
I like that idea, +1.

vsoch:
I basically want to say that we need to be sure that (most things) passed to this function return a proper url.

vsoch:
This has been finished. Good work!

PR #6227, opened 2021-10-15
Require go 1.16

DrDaveD:
This PR updates the minimum required go version to 1.16.

This is especially challenging for rpm builds on RHEL systems, because their current version is go 1.15.  This PR includes the ability for rpm builds to compile the go toolchain from source using a too-old go version.  It adds an option `--only-rpm` to `mconfig` which skips the go version check and instead configures `singularity.spec` to require the newer go source tarball and build it.  That go source tarball then becomes part of the source rpm.

This increase in minimum go version is needed for
- #6221 
It takes a little bit from Sylabs prs
- sylabs/singularity#366 and sylabs/singularity#371
but unlike those, it doesn't break the ability to do the rpm build on systems such as the EPEL/Fedora koji that do not allow rpm builds to download anything from the internet.


DrDaveD:
Replaced by #6229

Issue #292, opened 2016-11-02
Installing with multijob make fails


markeganfuller:
Installing via `make install` fails if `-j` is used to specify multiple jobs, for instance a common setting on Arch is `$(nproc)`.

Steps to reproduce:
```bash
git clone git@github.com:singularityware/singularity.git
cd singularity
./autogen.sh
./configure
make -j 20
sudo make install -j 20
```
Note: I used twenty here as an example and to try and force the bug to appear.

Error message
```no-highlight
make[4]: Leaving directory `/home/vagrant/singularity/src/lib'
make[3]: Leaving directory `/home/vagrant/singularity/src/lib'
make[2]: Leaving directory `/home/vagrant/singularity/src/lib'
Making install in util
make[2]: Entering directory `/home/vagrant/singularity/src/util'
make[3]: Entering directory `/home/vagrant/singularity/src/util'
make[3]: Nothing to be done for `install-exec-am'.
make[3]: Nothing to be done for `install-data-am'.
make[3]: Leaving directory `/home/vagrant/singularity/src/util'
make[2]: Leaving directory `/home/vagrant/singularity/src/util'
make[2]: Entering directory `/home/vagrant/singularity/src'
make[3]: Entering directory `/home/vagrant/singularity/src'
make  install-data-hook
make[4]: Entering directory `/home/vagrant/singularity/src'
 /bin/chown root:root /usr/local/libexec/singularity/sexec-suid
/bin/chown: cannot access ‘/usr/local/libexec/singularity/sexec-suid’: No such file or directory
 /bin/chmod 4755 /usr/local/libexec/singularity/sexec-suid
/bin/chmod: cannot access ‘/usr/local/libexec/singularity/sexec-suid’: No such file or directory
make[4]: *** [make_suid] Error 1
make[4]: Leaving directory `/home/vagrant/singularity/src'
make[3]: *** [install-data-am] Error 2
make[3]: Leaving directory `/home/vagrant/singularity/src'
make[2]: *** [install-am] Error 2
make[2]: Leaving directory `/home/vagrant/singularity/src'
make[1]: *** [install-recursive] Error 1
make[1]: Leaving directory `/home/vagrant/singularity/src'
make: *** [install-recursive] Error 1
```

Looks like the chmod / chown is not correctly dependent on the file creation.

gmkurtzer:
Fixed in [master 46578d8]. Can you test and if it looks good go ahead and close the ticket.

Thanks!

[...]

PR #6241, opened 2021-10-27
cgroups v2 support (apply from v1 / OCI spec), from sylabs 292

edytuk:
This pulls in sylabs PR

- sylabs/singularity#292
which fixed
- sylabs/singularity#60

> Support execution on a cgroups v2 unified hierarchy system, by using the `containerd/cgroups/v2` translation of the `opencontainers/runtime-spec` `LinuxResources` structure (which we use for cgroups config toml files) into v2 Resources. Device limits are implemented with the eBPF translation.
> 
> This will mean that existing v1 / OCI style cgroups configuration files can be used on a system configured for the unified hierarchy without an ugly failure. Also fixes failures with the `oci` commands on these systems.
> 
> Note that we don't yet support the Unified attribute from LinuxResources, which is a free-form map[string][string] for expressing cgroups v2 configuration natively.
> 
> _NOTE_ CircleCI will not test the new cgroups v2 code as it running Ubuntu 20.04 with the cgroups v1 hierarchy. I have run the tests on Debian 11 and Fedora 34, both of which use the cgroups v2 unified hierarchy.


cclerget (approved):

CHANGELOG.md
```diff
@@ -58,6 +58,8 @@
 
 - Fix regression when files `source`d from `%environment` contain `\` escaped
   shell builtins (fixes issue with `source` of conda profile.d script).
+- The `oci` commands will operate on systems that use the v2 unified cgroups
```
DrDaveD, line 61: Noticed after merge: this was put into the wrong place, under 3.8.2 instead of changes since last release.  Restored to proper place with issue apptainer/apptainer#11.

Issue #302, opened 2016-11-09
Container execution while restricting direct access (i.e. copy of container img)


serious-steve:
Hi folks,

 I've been bringing this up in the google group but this might be the better place to discuss my feature request.

I'd like to contribute the following functionality "group containers" to singularity:

- "Group Containers" are meant to be provided to users for execution only by our admin team
- Container image files must not be readable/copyable (due to licensed SW)
- Desired functionality: singularity shell / singularity exec in ro-mode

Greg has already proposed a way to implement this here: 
https://groups.google.com/a/lbl.gov/d/msg/singularity/VLVnK04joEw/fzKIgV2eBgAJ

I implemented the functionality proposed by Greg:
- Try to read the container file as calling user
- On fail: check if the singularity/singularity.conf has set 'execute group containers' enabled. Default is off
- If disabled -> Fail
- Else: check if root owns the container image. If no -> Fail
- Else: Obtain all groups of the calling user. Compare whether the container file is in any of those.
- If no -> FAIL 
- Else: allow privileged fopen (readonly)

The SUID feature must be enabled for this to work.
I will issue a pull request and combine that with this issue somehow...

serious-steve:
I've not written tests yet as first I'd like feedback on the code/the way I implemented this.
This is the link to the branch: https://github.com/c1t4r/singularity/tree/group_containers

bauerm97:
So I've taken a look at your implementation and I think that for the most part the logic looks good. However, I'm curious about the reason why you decided to split the logic for this feature into two files. It looks like this feature is only going to be used when calling `rootfs_image_init`, therefore it might make more sense to move all the logic into `rootfs/image/image.c` for now.

The other thing is, right now when you use sudo singularity create to make an image, it has read permissions for all users and groups, even though it's owned by root. This means you're going to have to edit permissions on the images in order for them to cooperate with this feature anyway. My question is: would it be enough for you to simply `chmod 0750 container.img` and `chgrp containergroup container.img`? All users not in the group will fail to open the container for reading and error out, while users in the group will be able to open it for reading and will continue on as normal. See my testing below (note that I am a member of docker usergroup on my computer):

```
mibauer@depc403:~/devel_singularity/testgroup$ ls -l group*
-rwxr-x--- 1 root docker 805306399 Nov  9 12:16 group2.img
-rwxr-x--- 1 root root   805306399 Nov  9 12:16 group.img
[...]

Issue #10, opened 2016-03-10
bug importing directory


kforner:
Hello,
Using singularity 0f5452bcf97109c6562e3390a2fa7e2d32e096ec, 
the build imports only a subset of the files and put them in wrong places.

For instance, it installs /usr/local/R-3.0.2/lib/R/library/base/help/scoping.R which does not exist, 
it is actually /usr/local/R-3.0.2/lib/R/library/base/demo/scoping.R

If I change the %files spec to /usr/local/R-3.0.2/lib/R/library/base/*, it imports all files but flatten the directory.

```
r.spec:
Name: R302
Exec: /usr/local/bin/R
%files
/usr/local/R-3.0.2/lib/R/library/base
```

the directory base looks like: 

```
drwxr-xr-x   7 root root  4096 Feb  5  2014 ./
drwxr-xr-x 258 root root 12288 May 15  2015 ../
-rw-r--r--   1 root root   954 Feb  5  2014 CITATION
drwxr-xr-x   2 root root  4096 Feb  5  2014 demo/
-rw-r--r--   1 root root   267 Feb  5  2014 DESCRIPTION
drwxr-xr-x   2 root root  4096 Feb  5  2014 help/
drwxr-xr-x   2 root root  4096 Feb  5  2014 html/
-rw-r--r--   1 root root 22491 Feb  5  2014 INDEX
drwxr-xr-x   2 root root  4096 Feb  5  2014 Meta/
drwxr-xr-x   2 root root  4096 Feb  5  2014 R/

/usr/local/R-3.0.2/lib/R/library/base/demo:
total 28
drwxr-xr-x 2 root root 4096 Feb  5  2014 ./
drwxr-xr-x 7 root root 4096 Feb  5  2014 ../
-rw-r--r-- 1 root root 1179 Feb  5  2014 error.catching.R
-rw-r--r-- 1 root root 4252 Feb  5  2014 is.things.R
-rw-r--r-- 1 root root 2115 Feb  5  2014 recursion.R
-rw-r--r-- 1 root root 1560 Feb  5  2014 scoping.R

/usr/local/R-3.0.2/lib/R/library/base/help:
total 1892
drwxr-xr-x 2 root root    4096 Feb  5  2014 ./
drwxr-xr-x 7 root root    4096 Feb  5  2014 ../
-rw-r--r-- 1 root root   11185 Feb  5  2014 aliases.rds
-rw-r--r-- 1 root root   31346 Feb  5  2014 AnIndex
[...]

PR #6248, opened 2021-10-29
fix: check remote build bootstrap valid before submitting, from sylabs 302

edytuk:
This pulls in sylabs PR

- sylabs/singularity#302
which fixed
- sylabs/singularity#10

> When we submit a remote build we should check that the 'bootstrap' header is valid before we submit the remote build.
> 
> Also ensure it is not a 'localimage' source, as this is not supported at present.
> 
> Example...
> 
> ```
> 04:51 PM $ singularity build -r test.sif libary://alpine
> FATAL:   Unable to build from libary://alpine: invalid build source "libary"
> ```


DrDaveD:
We won't be needing this fix.

Issue #313, opened 2016-11-13
Support for image metadata


vsoch:
For example, in a Dockerfile you can set a key value pair, called a [LABEL](https://docs.docker.com/engine/reference/builder/#/label) to describe the image. This would be easy to have integrated into Singularity Hub, and the questions are:

- How/if we want to integrate this into the image itself
- Given yes - how parsed with singularity command line tool?

vsoch:
DONE! In development branch. The user can now add `%labels` to bootstrap spec file!

PR #6251, opened 2021-11-02
e2e: Disable busybox bootstrap due to frequent dl failures, from sylabs 313

edytuk:
This pulls in sylabs PR

- sylabs/singularity#313

> Disable busybox bootstrap from tgz, as the download from the busybox site has been failing often, disrupting CI / merges.


DrDaveD (approved):
I agree, the busybox test fails too often.  

FYI: @kmuriki @ikaneshiro @cclerget

PR #6254, opened 2021-11-02
WIP: fix: Invalid path for --pwd is ignored instead of producing an error

### LINKED ISSUE #6086: Invalid path for --pwd is ignored instead of producing an error
### Version of Singularity:

```
$ singularity --version
singularity version 3.8.0-1.el7
```

### Expected behavior

When I specify the `--pwd` command to `singularity exec`, I expect that the working directory is set to that location. When the location does not exist, I expect an error, in the same way that `cd` behaves (and any similar function in all interpreted programming languages I know, `setwd()` in R, `os.chdir()` in Python, ...)

### Actual behavior

If the path to `--pwd`  does not exist inside the container, it defaults to `$HOME`. I can not imagine a situation where this default would be useful if I give `--pwd` explicitly. Also this might lead to files being overwritten by accident if singularity for example was called by a script which assumes that the working directory is set successfully because there was no error by singularity.

```
singularity pull library://default/alpine
singularity exec --pwd /etc alpine_latest.sif
# /etc
singularity exec --pwd /no/such/path alpine_latest.sif
# /home/user
```

centos-release-7-9.2009.1.el7.centos.x86_64

Via `yum`.

kadern0:
Signed-off-by: Pablo Caderno <kaderno@gmail.com>

Changed default behavior of switching to a different directory if ` os.Chdir(e.EngineConfig.OciConfig.Process.Cwd)` fails when starting the container. 

Whilst this approach should fix the issue, it might be "too strict" for other cases. I think some discussion about that would be beneficial. 

Fixes #6086

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@
 
 ### Changed defaults / behaviours
 
+- When starting a container, in case of problem with the cwd, istead of trying to
+  default to a different directory, return error to prevent unexpected behaviors #6086
 - LABELs from Docker/OCI images are now inherited. This fixes a longstanding
   regression from Singularity 2.x. Note that you will now need to use
   `--force` in a build to override a label that already exists in the source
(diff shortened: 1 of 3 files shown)


DrDaveD:
@kadern0 If you would like this to get in please resubmit it to https://github.com/apptainer/apptainer.

PR #6258, opened 2021-11-03
Add seccomp, cryptsetup, devscripts, correct go version test to debian packaging

### LINKED ISSUE #6236: debian build fails on ubuntu 20.04, mising seccomp and cryptsetup support
### Version of Singularity:

3.8.3

### Expected behavior

build on ubuntu 20.04 with debuild
build with seccomp and cryptsetup support

### Actual behavior

fails due to missing dependency on golang-go (>= 2:1.13) but golang-go (2:1.13~1ubuntu2) is installed.
Reason is failing version comparison, because  2:1.13 > 2:1.13~1ubuntu2. 
seccomp and cryptsetup not detected since missing in debian/control Build-Depends

follow the instructions here for debian build on ubuntu 20.04 

ubuntu 20.04

it is about build

Tested fix: 

modify debian/control like this

Build-Depends:
...
 golang-go (>= 2:1.13~~),
 libseccomp-dev,
 cryptsetup
...

With  libseccomp-dev and cryptsetup installed the debuild works fine. The version comparison 2:1.13~~ < 2:1.13~1ubuntu2) now yields the expected result on ubuntu 20.04.

skluth:

Useful application of e.g. singularity build --fakeroot needs seccomp support in the build.

For creation of encrypted containers cryptsetup is needed.

Devscripts was a missing dependency.

Use of "~~" in the golang version test avoids failure with ubuntu golang packages

 - Fixes #6236

### DIFF
--- a/dist/debian/control
+++ b/dist/debian/control
@@ -15,7 +15,10 @@ Build-Depends:
  libssl-dev,
  python,
  uuid-dev,
- golang-go (>= 2:1.13)
+ devscripts,
+ libseccomp-dev,
+ cryptsetup,
+ golang-go (>= 2:1.13~~)
 Standards-Version: 3.9.8
 Homepage: http://gmkurtzer.github.io/singularity
 Vcs-Git: https://github.com/hpcng/singularity.git


skluth:
e2e_tests seem to fail because of docker access trouble.

DrDaveD (approved):

skluth:
Thanks! Just noticed the rpm build (dist/rpm/singularity.spec.in) is also missing the seccomp support, the needed line `BuildRequires:  libseccomp-devel` is not there. I did test that on OBS, which accepted it as an OBS patch, but having it upstream here would be nicer. Should we open another PR?

DrDaveD:
First please make an issue describing exactly what goes wrong with missing libseccomp and --fakeroot.  I would like to see the problem for myself.  I didn't ask it for debian but I think the rpm gets a lot more use.

skluth:
Turns out that dist/rpm/singularity.spec.in contains
`%if ! 0%{?el6}`
`%if "%{sles_version}" != "11"`
`BuildRequires: libseccomp-devel`
`%endif`
`%endif`
so unless its RHEL6 or SLES11 seccomp is in for rpms, and since a while. No idea why this did not appear in the OBS spec file, but it is a "spec.in" file processed by some config step later.

DrDaveD:
Oh right, libseccomp-devel didn't exist on those older OSes.

skluth:
yes, but seccomp is in the rpm builds since a long time, I should have discovered that earlier :)

PR #399, opened 2016-12-08
updating cache creation to fix #334

vsoch:
Fixes #334 

Changes proposed in this pull request

 - Change the get_cache to use a function, mkdir_p, that will use `os.makedirs()` to create the path instead of checking if it exists, and then creating, for each.

@flx42 could you take a look?

@singularityware-admin

PR #398, opened 2016-12-08
updating configure_transform.py executable to use optparse and not ar…

vsoch:
This PR is the final bit to fully remove argparse from the configuration, meaning that older systems (RHEL 6) don't have the unmet dependency.

- I wrote a custom function as a workaround for requiring the arguments
- I changed the exit code from 0 to 1, given that something is missing
- argparse is gone!

@singularityware-admin

PR #6288, opened 2021-11-11
Split SIF/bare types w.r.t allow container $type config, from sylabs 399

edytuk:
This pulls in sylabs PR

- sylabs/singularity#399
 which fixed
- sylabs/singularity#398

> The `allow container $type` directives are confusing as they allow
> limiting use of SquashFS and EXT3 containers... but will also apply to
> those filesystems inside of a SIF.
> 
> We commonly present SIF as a separate container format, and users
> don't generally need to be aware that it's a wrapper format for other
> filesystem types. We generally talk about a `SquashFS` image in the
> context of a raw Singularity 2.x image, not the contents of a
> SIF. With this in mind it makes more sense to separately allow /
> disallow SIF or bare SquashFS etc.
> 
> This PR:
> 
> - Modifies the logic applying the `allow container squashfs/ext3` directive so they only apply to _bare_ images
>       in those formats, not SIFs using those formats internally.
> 
> - Modifies `allow container encrypted` so that it only controls use of
>       SIF containers where the RootFS is an encrypted SquashFS
>       filesystem.
> 
> - Adds a new `allow container sif` directive, controlling use of
>       unencrypted SIF containers, with _any_ internal FS, except an
>       encrypted RootFS (see above).

### DIFF
--- a/pkg/image/image.go
+++ b/pkg/image/image.go
@@ -253,6 +253,40 @@ func (i *Image) GetDataPartitions() ([]Section, error) {
 	return i.getPartitions(DataUsage)
 }
 
+// HasEncryptedRootFs returns true if the image contains an encrypted
+// rootfs partition.
+func (i *Image) HasEncryptedRootFs() (encrypted bool, err error) {
+	rootFsParts, err := i.GetRootFsPartitions()
+	if err != nil {
+		return false, fmt.Errorf("while getting root FS partitions: %v", err)
+	}
+
+	for _, p := range rootFsParts {
+		if p.Type == ENCRYPTSQUASHFS {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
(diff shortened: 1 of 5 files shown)


pkg/image/image.go
```diff
@@ -253,6 +253,40 @@ func (i *Image) GetDataPartitions() ([]Section, error) {
...
+		if p.Type == ENCRYPTSQUASHFS {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+// initFile ensures file descriptor is associated to a file handle.
+func (i *Image) initFile() error {
```
cclerget: This is a dead function, it has been removed by fix in https://github.com/hpcng/singularity/pull/6054

cclerget (changes requested):

cclerget (approved):

PR #6289, opened 2021-11-11
fix: Don't truncate config file until we have valid output to write, from sylabs 410

edytuk:
This pulls in sylabs PR

- sylabs/singularity#410
 which fixed
- sylabs/singularity#409

> When using `config global --set` an invalid directive value could previously
> lead to an empty `singularity.conf`, as the file was opened with
> `O_TRUNC` before the new config was generated/validated.
> 
> Generate the config output to an in memory buffer, and only create /
> truncate the `singularity.conf` file once we have generated known valid
> output.

### DIFF
--- a/internal/app/singularity/config_global_linux.go
+++ b/internal/app/singularity/config_global_linux.go
@@ -6,7 +6,9 @@
 package singularity
 
 import (
+	"bytes"
 	"fmt"
+	"io"
 	"os"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


DrDaveD (approved):

DrDaveD:
Note: this missed the CHANGELOG entry.  Added back with apptainer/apptainer#11

PR #6295, opened 2021-11-12
Issue(s): #6294 - update test case which is returning a false negative condition

### LINKED ISSUE #6294: update test case which is returning a false negative condition
update test case which is returning a false negative condition in `pkg/util/unix/socket_test.go`

```go
		// close socket implies to delete file automatically
		os.Chdir(dir)
		ln.Close()

		// socket file is deleted by net package at close
		if err := os.Remove(path); err == nil {
			t.Errorf("unexpected success during socket removal")
		}
```

note that os.Remove should succeed if the directory is empty

per https://pkg.go.dev/net#TCPListener.File, closing the listener is separate from closing/deleting the file

PaulCharlton:
update test case which is returning a false negative condition
1) remove extraneous chdir;
2) test that socket file removal does not present an error
3) report actual underlying error if socket file removal does not succeed

 - Fixes #6294

### DIFF
--- a/pkg/util/unix/socket_test.go
+++ b/pkg/util/unix/socket_test.go
@@ -80,13 +80,12 @@ func TestCreateSocket(t *testing.T) {
 
 		<-syncCh
 
-		// close socket implies to delete file automatically
-		os.Chdir(dir)
+		// close socket does not delete the os file
 		ln.Close()
 
-		// socket file is deleted by net package at close
-		if err := os.Remove(path); err == nil {
-			t.Errorf("unexpected success during socket removal")
+		// socket file is NOT deleted by net package at close
+		if err := os.Remove(path); err != nil {
+			t.Errorf("unexpected failure [%s] during socket file removal", err.Error())
 		}
 	}
 }


cclerget (changes requested):
Hey @PaulCharlton , from https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/net/unixsock_posix.go;l=219;drc=refs%2Ftags%2Fgo1.17.3 and https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/net/unixsock_posix.go;l=174;drc=refs%2Ftags%2Fgo1.17.3, the unix socket listener is removing the socket automatically at close, this behavior can be disabled with `SetUnlinkOnClose` to false, the call to `os.Remove` ensure the socket was indeed deleted explaining the `err == nil` which mean in this case that the socket was present at the time of removal otherwise `os.Remove` call is expected to return a `file does not exist` error, admittedly the original condition is confusing and can miss other errors, you can probably turn it into :

```
if err := os.Remove(path); !os.IsNotExist(err)
```

Since a nil error here is not expected, it would also be handled with the condition above

DrDaveD:
@PaulCharlton If you would like this to get in please resubmit it at https://github.com/apptainer/apptainer

Issue #449, opened 2017-01-09
Bind host /tmp during bootstrap


PauloSouza7:
To build some apps we need a lot of temporary space to compile sources.
For example OpenCV takes more than 4GB.
It will be great to have access to host /tmp during bootstrap process
Also, can you make `/tmp/` the current dir of bootstrap instead of `/`

bauerm97:
What version of Singularity (version number or commit) are you running on? In our current master branch (after 2.2) we have added a new bootstrapping process which should be mounting the `/tmp/` directory for access during all bootstrap scripts.

PauloSouza7:
Hi,

I'm using the stable 2.2 from download page. I'll test the git master. 

How about making /tmp the current directory for bootstrap instead of / ?

Sure I can cd /tmp on the first line but I think if /tmp is the default directory it will avoid people downloading and keeping temp files on /

Thanks,
Paulo.

gmkurtzer:
This feature has been added in [development e9ba16c].

PR #6322, opened 2021-11-29
e2e: check more output of plugin commands, from sylabs 449

edytuk:
This pulls in sylabs PR

- sylabs/singularity#449
 which fixed
- sylabs/singularity#448

> Check output of plugin commands in addition to return code. Check state reported by plugin list across install/disable/enable/uninstall.


DrDaveD:
This was resubmitted at apptainer/apptainer#14

PR #439, opened 2017-01-03
commenting out singularity hub tests until new endpoints are added

vsoch:
This is a quick pause of the shub tests, to be re-added with additional functionality after the final hub is ready to go! For https://github.com/singularityware/singularity/pull/438#issuecomment-270138126

@singularityware-admin

PR #6323, opened 2021-11-29
Switch to containers/common from seccomp/containers-golang, from sylabs 438

edytuk:
This pulls in sylabs PR

- sylabs/singularity#438
 which fixed
- sylabs/singularity#437

> seccomp/containers-golang has been archived and will not receive fixes,
> so we need to transition to containers/common/pkg/seccomp.
> 
> Update our example seccomp profile to the output of `make seccomp.json`
> on the new module.
> 
> Add handling of errnoRet & defaultErrnoRet in seccomp profiles, and use
> this rather than passing the err number in function calls.
> 
> Skip over any seccomp rules that are redundant, i.e. the action is the
> same as the default action. These now cause AddRule to error with our updated dependencies.
> 
> I've opened sylabs/singularity#439 as an issue to improve unit and e2e testing here, so that we are checking the _effect_ of a broader variety of filters is as expected. I've added some tests in this PR that cover parts of the internal image-spec -> libseccomp translations, but a lot of paths here aren't well covered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,9 @@
 - Perform concurrent multi-part downloads for `library://` URIs. Uses 3
   concurrent downloads by default, and is configurable in `singularity.conf` or
   via environment variables.
+- Updated seccomp support allows use of seccomp profiles that set an error
+  return code with `errnoRet` and `defaultErrnoRet`. Previously EPERM was hard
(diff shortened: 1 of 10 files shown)


DrDaveD:
This was resubmitted at apptainer/apptainer#15

PR #6368, opened 2022-02-06
Fix mount ordering between image and user binds, from apptainer 188

DrDaveD:
This pulls in
- apptainer/apptainer#188
which addressed
- apptainer/apptainer#187


DrDaveD:
This is not serious enough to backport, plus it has a change in behavior.  Closing.

PR #6374, opened 2022-02-08
provide a local debian package builder script for testing

d-a-v:
A proposed user script `dist/debian-makepackage` provides a single command to build an installable debian package.

This has been tested with ubuntu-20.04 and relies on the [recommended golang PPA](https://github.com/golang/go/wiki/Ubuntu) as stated in the user script.

### DIFF
--- a/dist/debian-makepackage
+++ b/dist/debian-makepackage
@@ -1,0 +1,22 @@
+#!/bin/sh -ex
+
+# This script has been tested on ubuntu-20.04 using golang from this PPA:
+#   https://github.com/golang/go/wiki/Ubuntu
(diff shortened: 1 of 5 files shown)


d-a-v:
... It seems I am confused by default branch `master` which is way older than `release-3.8`.

DrDaveD:
Thanks for the contribution, but the master branch is no longer active in this repository.  The master development branch is now the main branch in https://github.com/apptainer/apptainer.  The release-3.8 branch is still active for bug fixes in the singularity-3.8 series.  All the active branches do now have debian package builders that work, and in fact all future releases will include .deb packages in the release assets.

d-a-v:
Thanks for the input

PR #6396, opened 2022-03-01
New Issue and PR creation lockdown

kmuriki:
This is an attempt to lockdown creation of new Issues and PRs in the Apptainer/Singularity repo.

### DIFF
--- a/.github/workflows/lockdown.yml
+++ b/.github/workflows/lockdown.yml
@@ -1,5 +1,4 @@
-name: 'Repo Lockdown'
-
+name: RepoLockdown
 on:
   issues:
     types: opened
@@ -21,4 +20,3 @@ jobs:
           pr-comment: >
             This repository does not accept pull requests,
             see the README for details.
-
--- a/.github/workflows/lockdown.yml
+++ b/.github/workflows/lockdown.yml
@@ -14,6 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: dessant/repo-lockdown@v2
+        with:
           issue-comment: >
             This repository does not accept bug reports,
             see the README for details.
--- a/.github/workflows/lockdown.yml
+++ b/.github/workflows/lockdown.yml
@@ -2,12 +2,9 @@ name: RepoLockdown
 on:
   issues:
     types: opened
-  pull_request_target:
-    types: opened
 
 permissions:
   issues: write
-  pull-requests: write
 
 jobs:
   action:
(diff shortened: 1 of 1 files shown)


.github/workflows/lockdown.yml
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+  action:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: dessant/repo-lockdown@v2
+       with:
+          issue-comment: >
+            This repository does not accept bug reports,
+            see the README for details.
```
DrDaveD: I suggest the following text:

New issues are no longer accepted in this repository.  If `singularity --version` says `singularity-ce`, submit instead to https://github.com/sylabs/singularity, otherwise submit to https://github.com/apptainer/apptainer.

.github/workflows/lockdown.yml
```diff
@@ -0,0 +1,23 @@
...
+          pr-comment: >
+            This repository does not accept pull requests,
+            see the README for details.
```
DrDaveD: We need it to not close new PRs, because we will continue to make release-3.8 PRs.  Unless you can set it to only close PRs on the master branch?

We do not want it to close existing open issues and PRs, not to add comments on closed issues and PRs.  Make sure those options are off (I'm not sure what the defaults are but I saw mentions of them).

DrDaveD (changes requested):

.github/workflows/lockdown.yml
kmuriki: OK the text has been updated to this new text. Thanks.

.github/workflows/lockdown.yml
kmuriki: Yes for now I removed auto handling of the PRs and put it back to the old state where new PRs can be submitted and they will stay. 

This plugin will not impact the old issues that have been sitting there. It will auto close only the new PRs from the point this PR gets merged into master. Turns out its not sufficient of the plugin config is in a branch, it needs to be in the master branch only to work. Not sure if there is some other config which controlling this but for now we need to merge this into the master branch.

kmuriki:
Should we force merge this PR as its failing the lint_markdown and rpmbuild-centos7 ?

DrDaveD (approved):

DrDaveD:
Yes, the CI tests don't all pass on the master branch anymore and it doesn't matter.  I changed the settings to not require CI tests to pass before merging.

